PR #81, opened 2025-06-17
.Net: Added Azure AI Persistent Agents

dmytrostruk:
Added Azure AI Persistent Agents including:
- Implementation from @stephentoub
- Integration tests
- Sample usage

Created `Microsoft.Agents.AzureAIAgentsPersistent` project to temporarily store `PersistentAgentsChatClient` implementation of `IChatClient` interface.

All integration tests and samples are verified: 
image

### DIFF
--- a/dotnet/tests/AzureAIAgentsPersistent.IntegrationTests/AzureAIAgentsPersistentFixture.cs
+++ b/dotnet/tests/AzureAIAgentsPersistent.IntegrationTests/AzureAIAgentsPersistentFixture.cs
@@ -9,8 +9,8 @@ using Azure;
 using Azure.AI.Agents.Persistent;
 using Azure.Identity;
 using Microsoft.Agents;
-using Microsoft.Extensions.AI.AzureAIAgentsPersistent;
 using Microsoft.Extensions.AI;
+using Microsoft.Extensions.AI.AzureAIAgentsPersistent;
 using Shared.IntegrationTests;
 
 namespace AzureAIAgentsPersistent.IntegrationTests;


dotnet/src/Microsoft.Agents.AzureAIAgentsPersistent/PersistentAgentsChatClient.cs
```diff
@@ -0,0 +1,431 @@
...
+        {
+            runOptions.MaxCompletionTokens ??= options.MaxOutputTokens;
+            runOptions.OverrideModelName ??= options.ModelId;
+            runOptions.TopP ??= options.TopP;
+            runOptions.Temperature ??= options.Temperature;
+            runOptions.ParallelToolCalls ??= options.AllowMultipleToolCalls;
+            // Ignored: options.TopK, options.FrequencyPenalty, options.Seed, options.StopSequences
+
+            if (options.Tools is { Count: > 0 } tools)
+            {
```
stephentoub: When this is actually moved over into Azure.AI.Agents.Persistent, this will likely need to do something like this:
https://github.com/dotnet/extensions/blob/694b95ef75c6bd9de00ef761dadae4e70ee8739f/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIAssistantChatClient.cs#L263-L279
Might be worth a TODO. Without that, options.Tools become overrides rather than augments. And unlike with OpenAI Assistants, where the assistants client provides the method to retrieve an Assistant and thus get access to any tools configured on it, the A.AI.A.Persistent library splits that functionality onto a separate type, such that external to the library, this code doesn't have access to it.

dotnet/src/Microsoft.Agents.AzureAIAgentsPersistent/PersistentAgentsChatClient.cs
```diff
@@ -0,0 +1,431 @@
...
+                                aiFunction.Description,
+                                BinaryData.FromBytes(JsonSerializer.SerializeToUtf8Bytes(aiFunction.JsonSchema, AgentsChatClientJsonContext.Default.JsonElement))));
+                            break;
+
+                        case HostedCodeInterpreterTool:
+                            toolDefinitions.Add(new CodeInterpreterToolDefinition());
+                            break;
+
+                        case HostedWebSearchTool webSearch when webSearch.AdditionalProperties?.TryGetValue("connectionId", out object? connectionId) is true:
+                            toolDefinitions.Add(new BingGroundingToolDefinition(new BingGroundingSearchToolParameters([new BingGroundingSearchConfiguration(connectionId!.ToString())])));
```
stephentoub: I wasn't sure what the difference was between the Bing "Grounding" tool and the Bing "Custom" tool, so I arbitrarily chose the grounding one to map HostedWebSearchTool to.

dotnet/src/Microsoft.Agents.AzureAIAgentsPersistent/PersistentAgentsChatClient.cs
dmytrostruk: Makes sense, added TODO

dotnet/samples/GettingStarted/Providers/ChatClientAgent_With_AzureAIAgentsPersistent.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        // Start a new thread for the agent conversation.
+        AgentThread thread = agent.GetNewThread();
+
+        // Respond to user input
+        await InvokeAgentAsync("Tell me a joke about a pirate.");
+        await InvokeAgentAsync("Now add some emojis to the joke.");
+
+        // Local function to invoke agent and display the conversation messages for the thread.
+        async Task InvokeAgentAsync(string input)
```
markwallace-microsoft: Rename to `RunAgentAsync`

stephentoub:
> temporarily 

How temporary is this? I was hoping we'd just prioritize getting the IChatClient implementation added to Azure.AI.Agents.Persistent.

dmytrostruk:
> I was hoping we'd just prioritize getting the IChatClient implementation added to Azure.AI.Agents.Persistent.

@stephentoub This is still in progress. I believe the main purpose of this PR is to double-check if current abstraction works well with `IChatClient` implementation and if integration tests are passing. We don't even need to merge this PR, we can keep it open until the `IChatClient` implementation will be added to `Azure.AI.Agents.Persistent`, and then just remove the implementation from here and leave only tests and samples.

stephentoub:
Sounds good. Thanks!

dotnet/src/Microsoft.Agents.AzureAIAgentsPersistent/Microsoft.Agents.AzureAIAgentsPersistent.csproj
```diff
@@ -0,0 +1,22 @@
...
+  <Import Project="$(RepoRoot)/dotnet/nuget/nuget-package.props" />
+
+  <ItemGroup>
+    <PackageReference Include="Azure.AI.Agents.Persistent" />
+    <PackageReference Include="Microsoft.Extensions.AI" />
+  </ItemGroup>
+
+  <PropertyGroup>
+    <!-- NuGet Package Settings -->
+    <Title>Microsoft.Agents.AzureAIAgentsPersistent</Title>
```
westey-m: Not terribly important since this will go away anyway, but this project probably makes more sense as Microsoft.Extensions.AI.AzureAIAgentsPersistent, since it's really an implementation of MEAI.

dotnet/tests/AgentConformance.IntegrationTests/AgentConformance.IntegrationTests.csproj
```diff
@@ -4,6 +4,7 @@
     <TargetFrameworks>$(ProjectsTargetFrameworks)</TargetFrameworks>
     <TargetFrameworks Condition="'$(Configuration)' == 'Debug'">$(ProjectsDebugTargetFrameworks)</TargetFrameworks>
     <IsTestProject>false</IsTestProject>
+    <UserSecretsId>5ee045b0-aea3-4f08-8d31-32d1a6f8fed0</UserSecretsId>
```
westey-m: The user secrets id from SK integration tests is already in the directory.build.props for tests, so should be reusable here too.

dotnet/tests/AzureAIAgentsPersistent.IntegrationTests/AzureAIAgentsPersistent.IntegrationTests.csproj
```diff
@@ -0,0 +1,15 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>$(ProjectsTargetFrameworks)</TargetFrameworks>
+    <TargetFrameworks Condition="'$(Configuration)' == 'Debug'">$(ProjectsDebugTargetFrameworks)</TargetFrameworks>
+    <InjectSharedIntegrationTestCode>True</InjectSharedIntegrationTestCode>
+    <UserSecretsId>5ee045b0-aea3-4f08-8d31-32d1a6f8fed0</UserSecretsId>
```
westey-m: Same as above, the SK integration tests secrets id is already in the directory props.

dmytrostruk:
Merging this PR to add tools support later.

PR #84, opened 2025-06-18
Add PoC CopilotStudio agent

westey-m:
This is just a proof of concept to show that it's possible to do this with our abstractions.
If CopilotStudio becomes available over A2A we may just be able to have an A2AAgent instead.
One thing that doesn't really make sense, for CopilotStudio or any other similar A2A-type agents, is the `AdditionalInstructions` option at invocation time.  We cannot use it, and the user may have a reasonable expectation that it gets used.
I'll send a separate PR to remove that from the abstraction.

#33


dotnet/src/Microsoft.Agents.CopilotStudio/CopilotStudioAgent.cs
```diff
@@ -0,0 +1,131 @@
...
+
+        // Enumerate the response messages
+        var reponseMessagesList = new List<ChatMessage>();
+        await foreach (ChatMessage message in reponseMessages.ConfigureAwait(false))
+        {
+            // Add the message to the list
+            reponseMessagesList.Add(message);
+        }
+
+        return new ChatResponse(reponseMessagesList);
```
stephentoub: This is a neat PoC and cool to see.

If we were doing this "for real", we'd want to populate additional properties on ChatResponse and ChatMessage, in paritcular the various Id properties like ResponseId and MessageId.

Those end up being particularly important for streaming consumers so that they can tell things like response boundaries. The ToChatResponse helpers do that, for example.

dotnet/src/Microsoft.Agents.CopilotStudio/CopilotStudioAgent.cs
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    /// <inheritdoc/>
+    public override async Task<ChatResponse> RunAsync(
+        IReadOnlyCollection<ChatMessage> messages,
+        AgentThread? thread = null,
+        AgentRunOptions? options = null,
+        CancellationToken cancellationToken = default)
+    {
+        if (messages is not { Count: > 0 })
```
stephentoub: Presumably we'd actually want to throw ArgumentNullException for null.

For Count == 0, we don't want to still submit the request?

dotnet/src/Microsoft.Agents.CopilotStudio/CopilotStudioAgent.cs
```diff
@@ -0,0 +1,131 @@
...
+        // Ensure that we have a valid thread to work with.
+        CopilotStudioAgentThread copilotStudioAgentThread = base.ValidateOrCreateThreadType(thread, () => new CopilotStudioAgentThread());
+        if (copilotStudioAgentThread.Id is null)
+        {
+            // If the thread ID is null, we need to start a new conversation and set the thread ID accordingly.
+            copilotStudioAgentThread.Id = await this.StartNewConversationAsync(cancellationToken).ConfigureAwait(false);
+        }
+
+        // Invoke the Copilot Studio agent with the provided messages.
+        string question = string.Join("\n", messages.Select(m => m.Text));
```
stephentoub: It only supports a single string input?

dotnet/src/Microsoft.Agents.CopilotStudio/CopilotStudioAgent.cs
```diff
@@ -0,0 +1,131 @@
...
+        var reponseMessages = ActivityProcessor.ProcessActivityAsync(this.Client.AskQuestionAsync(question, copilotStudioAgentThread.Id, cancellationToken), streaming: false, this._logger);
+
+        // Enumerate the response messages
```
stephentoub: What is it typically sending back for response messages? I've not used the API. Does it typically return one message? Many messages?

dotnet/src/Microsoft.Agents.CopilotStudio/ActivityProcessor.cs
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    private static IEnumerable<AIContent> GetMessageItems(IActivity activity)
+    {
+        yield return new TextContent(activity.Text);
+
+        // TODO: figure out how we want to support CardActions publicly on the abstraction.
+        foreach (CardAction action in activity.SuggestedActions?.Actions ?? [])
+        {
+            yield return new TextContent(action.Title);
```
stephentoub: You probably want to store CardAction as the TextContent's RawRepresentation.

Are there other interesting aspects of CardAction? What else is on it?

dotnet/src/Microsoft.Agents.CopilotStudio/ActivityProcessor.cs
```diff
@@ -0,0 +1,55 @@
...
+            switch (activity.Type)
+            {
+                case "message":
+                    yield return
+                        new(ChatRole.Assistant, contents: [.. GetMessageItems(activity)])
+                        {
+                            RawRepresentation = activity,
+                        };
+                    break;
+                case "typing" when streaming:
```
stephentoub: Why only when streaming?

dotnet/src/Microsoft.Agents.CopilotStudio/CopilotStudioAgent.cs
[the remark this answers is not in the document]
westey-m: Fixed

dotnet/src/Microsoft.Agents.CopilotStudio/CopilotStudioAgent.cs
```diff
@@ -0,0 +1,131 @@
...
+            // If the thread ID is null, we need to start a new conversation and set the thread ID accordingly.
+            copilotStudioAgentThread.Id = await this.StartNewConversationAsync(cancellationToken).ConfigureAwait(false);
+        }
+
+        // Invoke the Copilot Studio agent with the provided messages.
+        string question = string.Join("\n", messages.Select(m => m.Text));
+        var reponseMessages = ActivityProcessor.ProcessActivityAsync(this.Client.AskQuestionAsync(question, copilotStudioAgentThread.Id, cancellationToken), streaming: true, this._logger);
+
+        // Enumerate the response messages
+        await foreach (ChatMessage message in reponseMessages.ConfigureAwait(false))
```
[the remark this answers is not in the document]
westey-m: Fixed

dotnet/tests/CopilotStudio.IntegrationTests/Support/CopilotStudioTokenHandler.cs
```diff
@@ -0,0 +1,143 @@
...
+    private const string KeyChainServiceName = "copilot_studio_client_app";
+    private const string KeyChainAccountName = "copilot_studio_client";
+
+    private readonly CopilotStudioConnectionSettings _settings;
+    private readonly string[] _scopes;
+
+    private IConfidentialClientApplication? _clientApplication;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="CopilotStudioTokenHandler"/> class with the specified connection settings.`
```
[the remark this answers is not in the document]
westey-m: Fixed

dotnet/tests/CopilotStudio.IntegrationTests/CopilotStudioFixture.cs
```diff
@@ -0,0 +1,75 @@
...
+            services
+                .BuildServiceProvider()
+                .GetRequiredService<IHttpClientFactory>();
+
+        CopilotClient client = new(settings, httpClientFactory, NullLogger.Instance, CopilotStudioHttpClientName);
+
+        CopilotStudioAgent agent = new(client);
+
+        this._agent =
+            new CopilotStudioAgent(client);
```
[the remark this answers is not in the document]
westey-m: Fixed

dotnet/src/Microsoft.Agents.CopilotStudio/CopilotStudioAgent.cs
westey-m: > Presumably we'd actually want to throw ArgumentNullException for null.

Good point, I'll fix that.

> For Count == 0, we don't want to still submit the request?

I am consistently getting back an empty response from the service, so thought I'd add a shortcut to avoid wasting money and time, but of course they may fix it so that it does return a response, so I could always submit in the hope that the result improves.  I suppose if the user invoked with no message, they probably did it for a good reason.

dotnet/src/Microsoft.Agents.CopilotStudio/CopilotStudioAgent.cs
westey-m: Yeah, this is the method signature:
```csharp
IAsyncEnumerable<IActivity> AskQuestionAsync(
    string question,
    string? conversationId = null,
    CancellationToken ct = default(CancellationToken))
```

dotnet/src/Microsoft.Agents.CopilotStudio/ActivityProcessor.cs
westey-m: Yeah, we definitely need a better approach than what I have there.  I have a todo to look into it more.
I was also actually planning on pinging you to get your thoughts on CardActions.  They are pretty nice, in that they prompt the user to pick from a list of options.  I don't know whether it makes sense to support that directly in the abstraction.  @markwallace-microsoft was mentioning that there may be something similar for A2A.

They do have quite a few properties though that I'm not sure how broadly they may be supported, like an Image URL to display on the button and alt text for the image:
https://learn.microsoft.com/en-us/dotnet/api/microsoft.agents.core.models.cardaction?view=m365-agents-sdk

SuggestedActions is fairly light, although it does have a list of IDs of recipients to which the actions should be shown, which adds some complexity too.

Below is a serialized SuggestedActions property when I asked the agent to send a message for me on teams and it asked me back "Where would you like to post the message? Please select a location."

With this type of prompt from the agent, i.e. "Please select a location" there seems to be an expectation that the client would implement a selection experience of some sort instead of just expecting the end user to type in what they want.

```json
    "SuggestedActions": {
        "To": [],
        "Actions": [
            {
                "Type": "messageBack",
                "Title": "Channel",
                "Image": null,
                "ImageAltText": null,
                "Text": "Channel",
                "DisplayText": "Channel",
                "Value": {
                    "property": "location",
                    "value": "Channel"
                },
                "ChannelData": null
            },
            {
                "Type": "messageBack",
                "Title": "Group chat",
                "Image": null,
                "ImageAltText": null,
                "Text": "Group chat",
                "DisplayText": "Group chat",
                "Value": {
                    "property": "location",
                    "value": "Group chat"
                },
                "ChannelData": null
            }
        ]
    },
```

dotnet/src/Microsoft.Agents.CopilotStudio/CopilotStudioAgent.cs
westey-m: It's sending back a lot of "activities" with both streaming and non-streaming style "activities" being returned.  There is no options to switch between these two.  If the "activities" are of type "typing" they are typically streaming style messages, with anything from empty text to words.  "activities" with type "message" contains fully formed results and I've typically seen one of those per response.

"activities" with type "event" are also interesting in that I believe they convey information about what the agent is actually doing in the background.  These could be valuable to return on OnIntermediateMessage if we have an appropriate AIContext representation for them.

dotnet/src/Microsoft.Agents.CopilotStudio/ActivityProcessor.cs
westey-m: See my explanation above for the types of messages returned:
https://github.com/microsoft/agent-framework/pull/84#discussion_r2155220959

Also, I need to review this further.  I ported this from what we have in the SK repo, but I added the `when streaming` restriction, since I don't think these "typing" updates really make sense in the non-streaming scenario where the caller is expecting a final result.  At the same time I'm questioning whether TextReasoningContent makes sense here, because for streaming we expect TextContent items too.

dotnet/src/Microsoft.Agents.CopilotStudio/CopilotStudioAgent.cs
stephentoub: Seems then like the shared ProcessActivityAsync should actually be split into one for non-streaming that produces just a single message and one for streaming that produces a stream of ChatResponseUpdate instances directly?

dotnet/src/Microsoft.Agents.CopilotStudio/ActivityProcessor.cs
stephentoub: > I don't know whether it makes sense to support that directly in the abstraction. 

We'd want multiple examples from different providers first in order to figure out what the common denominator is.

In the meantime, the two options are:
1) Just store the content as the RawRepresentation of an AIContent type, either the base or the closest derived one.
2) Add a custom AIContent-derived type that's specific to the target library.

dotnet/src/Microsoft.Agents.CopilotStudio/CopilotStudioAgent.cs
westey-m: Yeah, thanks for the callout.  I've added a few more mappings, and TODO's around these, so that we don't forget to review further.

stephentoub:
High level question...

With the agent thread impl being empty, isn't this just an IChatClient implementation? What benefits do we get doing this by deriving from Agent?

dotnet/src/Microsoft.Agents.CopilotStudio/ActivityProcessor.cs
```diff
@@ -0,0 +1,59 @@
...
+internal static class ActivityProcessor
+{
+    public static async IAsyncEnumerable<ChatMessage> ProcessActivityAsync(IAsyncEnumerable<IActivity> activities, bool streaming, ILogger logger)
+    {
+        await foreach (IActivity activity in activities.ConfigureAwait(false))
+        {
+            switch (activity.Type)
+            {
+                case "message":
+                    yield return CreateChatMessageFromActivity(activity, GetMessageItems(activity));
```
rogerbarreto: I would keep consistency just setting the `Activity` into the `chatMessage.RawRepresentation` scope. For now avoiding some messages to have the activity in the `AIContent.RawRepresentation`.

```suggestion
                    yield return CreateChatMessageFromActivity(activity, GetMessageItems(activity)) { RawRepresentation = activity };
```

dotnet/src/Microsoft.Agents.CopilotStudio/ActivityProcessor.cs
rogerbarreto: If they return `typing` from their non-streaming flag I would just proxy whatever comes from the `SDK` down to the caller without transformations, this could have useful metadata information we still don't know.

dotnet/src/Microsoft.Agents.CopilotStudio/CopilotStudioAgent.cs
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    private async Task<string> StartNewConversationAsync(CancellationToken cancellationToken)
+    {
+        string? conversationId = null;
+        await foreach (IActivity activity in this.Client.StartConversationAsync(emitStartConversationEvent: true, cancellationToken).ConfigureAwait(false))
+        {
+            if (activity.Conversation is not null)
+            {
+                conversationId = activity.Conversation.Id;
```
rogerbarreto: You might be able to leave as soon you get the Id, leaving the AsyncEnumerable sooner may not trigger an abort.

```suggestion
                conversationId = activity.Conversation.Id;
                return conversationId;
```

dotnet/src/Microsoft.Agents.CopilotStudio/CopilotStudioAgent.cs
```diff
@@ -0,0 +1,137 @@
...
+            }
+        }
+
+        if (string.IsNullOrEmpty(conversationId))
+        {
+            throw new System.InvalidOperationException("Failed to start a new conversation.");
+        }
+
+        return conversationId!;
```
rogerbarreto: If you are able to leave sooner this can be simplified.

```suggestion
        throw new System.InvalidOperationException("Failed to start a new conversation.");
```

dotnet/src/Microsoft.Agents.CopilotStudio/CopilotStudioAgent.cs
```diff
@@ -0,0 +1,137 @@
...
+        await foreach (ChatMessage message in responseMessages.ConfigureAwait(false))
+        {
+            // TODO: Review list of ChatResponse properties to ensure we set all availble values.
+            // Setting ResponseId and MessageId end up being particularly important for streaming consumers
+            // so that they can tell things like response boundaries.
+            yield return new ChatResponseUpdate(message.Role, message.Contents)
+            {
+                ResponseId = message.MessageId,
+            };
+        }
```
rogerbarreto: You might consider looking into other properties here:

Updates also have `MessageId`, `ResponsesId` and `ConversationId`.

ConversationId
AdditionalProperties
AuthorName
MessageId

dotnet/src/Microsoft.Agents.CopilotStudio/CopilotStudioAgentThread.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+namespace Microsoft.Agents.CopilotStudio;
+
+/// <summary>
+/// Represents a thread for interacting with a Copilot Studio agent.
+/// </summary>
+public class CopilotStudioAgentThread : AgentThread
+{
+}
```
rogerbarreto: If not implementation is required for the thread, we can just go without it, right?

dotnet/tests/CopilotStudio.IntegrationTests/Support/CopilotStudioTokenHandler.cs
```diff
@@ -0,0 +1,143 @@
...
+
+#pragma warning disable CA1812 // Internal class that is apparently never instantiated.
+
+/// <summary>
+/// A <see cref="DelegatingHandler"/> that adds an authentication token to the request headers for Copilot Studio API calls.
+/// </summary>
+/// <remarks>
+///  For more information on how to setup various authentication flows, see the Microsoft Identity documentation at https://aka.ms/msal.
+/// </remarks>
+internal sealed class CopilotStudioTokenHandler : DelegatingHandler
```
rogerbarreto: you might just be able to set this as an `HttpClientHandler` and don't need to set the `InnerHandler` in the ctor.

```suggestion
internal sealed class CopilotStudioTokenHandler : HttpClientHandler
```

dotnet/src/Microsoft.Agents.CopilotStudio/CopilotStudioAgent.cs
westey-m: It's unclear to me what would happen if we don't stream the async enumerable to the end.  I would hope that it completes the operation, but I don't think we can make that assumption, so I think it's safer to stream to the end.

dotnet/src/Microsoft.Agents.CopilotStudio/CopilotStudioAgent.cs
westey-m: I'm hesitant to avoid streaming the result to the end, since it may not complete the operation before we return if we don't.  E.g. it may have returned an id, but there may still be steps pending for completely setting up the conversation that we need to wait for.  It is likely just be a badly designed API, but having an AsyncEnumerable returned by an update means that it may returned multiple results before the operation is complete.

dotnet/tests/CopilotStudio.IntegrationTests/Support/CopilotStudioTokenHandler.cs
westey-m: Thanks, will update.  Note that this is just a clone from what we have in the SK repo, so I didn't review this in detail.

dotnet/src/Microsoft.Agents.CopilotStudio/ActivityProcessor.cs
rogerbarreto: CardActions might be a good scenario that we can abstract as a `toolCall` (`FunctionCallContent`) from the Model.

dotnet/src/Microsoft.Agents.CopilotStudio/ActivityProcessor.cs
westey-m: Thanks, yeah, makes sense.  Will update.

dotnet/src/Microsoft.Agents.CopilotStudio/CopilotStudioAgentThread.cs
westey-m: Yeah, I was thinking about this as well, but I'm unsure whether we should just accept `AgentThread`.  Accepting subclasses of `AgentThread` would certainly be problematic since they may support other thread storage capabilities, like in the case of `ChatClientAgentThread` and users may expect those to be supported.  We can though restrict it to just `AgentThread` itself with no subclasses allowed.
On the other hand I do like the consistency of saying that you have a matching thread type for your agent.  It is easy for consumers to understand.

@stephentoub, any preference from your end?

dotnet/src/Microsoft.Agents.CopilotStudio/CopilotStudioAgent.cs
westey-m: I've been digging into it more and trying out more scenarios, and it's actually less consistent than during my first observations.  The "typing" activities are not always returned and some times returned twice, so we cannot rely on them consistently.  

E.g. when I used a question that would invoke a server side tool, and it returned a follow up ask with options, no "typing" activities containing the response text were produced.

I also observed cases where the final text was returned twice as two sequences of "typing" activities, with ids that did not differentiate between the two iterations.  The "typing" message ids have a pattern of `{guid}-{i}` with each new message incrementing `i`, but the `guid` staying the same.  The second repeat of the same answer via "tying" messages had the same `guid` and just kept on incrementing `i`.  Only one "message" activity was returned though, so it was more reliable.

I've changed the code to cater for these inconsistencies and added some comments around it.

dotnet/src/Microsoft.Agents.CopilotStudio/ActivityProcessor.cs
westey-m: FYI: the API has no support for separate streaming and non-streaming cases.
Also see my post on our chat in teams around how we should return these types of messages to callers.

dotnet/src/Microsoft.Agents.CopilotStudio/CopilotStudioAgentThread.cs
stephentoub: I'd like to first understand an answer to https://github.com/microsoft/agent-framework/pull/84#issuecomment-2985495916

dotnet/src/Microsoft.Agents.CopilotStudio/CopilotStudioAgent.cs
westey-m: Yeah, it was still TODO.  I've added a few more, but the available data is definitely more sparse here.

westey-m:
> High level question...
> 
> With the agent thread impl being empty, isn't this just an IChatClient implementation? What benefits do we get doing this by deriving from Agent?

@stephentoub.  I was pondering this myself too, and I think while the two are similar, CopilotStudio supports very few of the features available on IChatClient.  The only option on ChatOptions for example that would be supported is ConversationId.  I realise that of course not all options are supported by all implementations today, but typically they do support a broader set than 1.

That said, I am concerned about any proliferation in agent types, especially one-off agent types for specific services.  I was asking myself, do we need an abstraction for clients to high level agents like these so that we can just expose a single high level Agent implementation for these?  However, this is essentially what A2A is, right.  @markwallace-microsoft, has been chatting with the CopilotStudio Agent folks and it sounds like they will be supporting an A2A protocol for CopilotStudio agents, which means if we have an `A2AAgent` that uses an `A2AClient` that could serve as the basis for communicating with many of these types of agents.  Note that Mark has also called attention to the fact that there are many competing A2A protocols at the moment, so there is some additional complexity there.

We also chatted more about this in our standup earlier, and @markwallace-microsoft mentioned that in conversations with the CopilotStudio folks they also referred to an "Activity protocol", and we realized that this is a published protocol and is the one returned by CopilotStudio today.  See https://github.com/microsoft/Agents/blob/main/specs/activity/protocol-activity.md
Very helpful in deciphering the different activity types, but also yet another agent protocol.

I think it's useful to consider the high level agents like CopilotStudio or agents exposed via A2A as a class of agents.  Let's call these black box agents, since their inner workings can be opaque and they typically support a very simple request in / response out interface without any further options or manual intervention supported.  In my mind our agent abstraction also matches this level in the stack.  It is essentially an abstraction for black box agents.  We can of course implement our black box agent abstraction using the very open and flexible IChatClient, but based on that it seems like IChatClient sits lower in the stack than our agent abstractions.

Of course, none of this necessarily means that we couldn't expose these black box agents as IChatClient implementations.  And conversely it also doesn't prove that we can, without some more analysis/POCs, especially around some of the A2A protocols available.

dotnet/src/Microsoft.Agents.CopilotStudio/CopilotStudioAgent.cs
stephentoub: The string-based overload appears to be a simple wrapper for the IActivity-based overload, which is much richer in what it can accept, e.g. IActivity has Attachments which supports arbitrary payloads with a MIME type.

https://learn.microsoft.com/en-us/dotnet/api/microsoft.agents.copilotstudio.client.copilotclient.askquestionasync?view=m365-agents-sdk#microsoft-agents-copilotstudio-client-copilotclient-askquestionasync(microsoft-agents-core-models-iactivity-system-threading-cancellationtoken)

https://github.com/microsoft/Agents-for-net/blob/e7c554707bac6b554cb00e0072affc01ffd8df86/src/libraries/Client/Microsoft.Agents.CopilotStudio.Client/CopilotClient.cs#L164-L173

stephentoub:
> The only option on ChatOptions for example that would be supported is ConversationId.

I don't think that's a strong reason not to implement IChatClient. There are plenty of implementations that only support a subset, possibly small subset, of what's available.

> I think it's useful to consider the high level agents like CopilotStudio or agents exposed via A2A as a class of agents.

That's quite possible, but with how the abstractions are currently arranged, I'm missing why this is relevant. Agent effectively accepts a collection of ChatMessages in and produces ChatMessages out. This PoC has the copilot studio-based agent doing that and little else. If that's all it's going to do, it would seem it's "just" an IChatClient that can be used with a ChatClientAgent.

Is there copilot-studio-specific behavior that would need to live at the agent level that shouldn't be encapsulated into an IChatClient? The _current_ distinguishing factor seems to be whether the AgentThread and whether there's any custom interaction required there. But this implementation has none, with an empty custom AgentThread-derived type.

I can certainly imagine there being an Agent-derived type for representing a remote A2A agent that's accessed via an A2A library / HttpClient. Presumably it would need a custom AgentThread-derived type? Does the agent / agent-thread abstraction have what it needs to represent such a thing, especially with long-running tasks?

Similarly, I imagine we'll need some kind of AgentThread-derived type for working with the shared agent runtime, and that AgentThread would need to work with ChatClientAgent (I'm not sure what that looks like today or how that would be made to work, if the ChatClientAgent needs to be responsible for creating the thread).

But as written in this PR, I'm missing what about copilot studio makes it desirable to expose at the agent level rather than the chat client level.

markwallace-microsoft:
One of the goals of implementing for Copilot Studio was to clarify our thinking around different types of agent implementations and how we would build them. We may not move forward with this implementation.

I think we are discussing when we should and should not use `IChatClient` as the basis for integrating a new agent type.

For Agents like OpenAI Assistants, Responses, Foundry, ... using `IChatClient` seems to work, although we still need to demonstrate we can support all of the different tool types. So my current thinking is that if the Agent looks like Inference we should use `IChatClient` and otherwise we should consider another option.

Copilot Studio and A2A seem different to me, right now. Supporting A2A would give us Copilot Studio and lot's of other agent types with a single implementation so that seems like a good direction to go. So we may have a dedicated A2A Agent type.

stephentoub:
> One of the goals of implementing for Copilot Studio was to clarify our thinking

Makes sense.

> Supporting A2A would give us Copilot Studio and lot's of other agent types with a single implementation so that seems like a good direction to go.

I agree. Anywhere we can use a single implementation to support a whole category is goodness.

My main observation is having a dedicated derivation of Agent for copilot studio doesn't look right.

westey-m:
> My main observation is having a dedicated derivation of Agent for copilot studio doesn't look right.

Agreed! As I also said above `I am concerned about any proliferation in agent types, especially one-off agent types for specific services.`

The thing I think we need to address though is whether everything becomes an IChatClient, because they can, or if we have an Agent and Thread type for IChatClient and another for black box type agents.  While we haven't explored this yet, there may be a good chance that A2A can be represented as an IChatClient too after all.

One thing that bugs me about with the everything-is-an-ChatClientAgent approach is the fact that black box type agents typically don't support per request functions or instructions.  Therefore AIContextProviders are unlikely to work with these, and we may want to limit attaching AIContextProviders to those agents/threads that will work.  E.g. ChatClientAgentThread allows attaching AIContextProviders but Agent implementaitons that support black box agents do not have this on their public API surface.

dotnet/src/Microsoft.Agents.CopilotStudio/ActivityProcessor.cs
westey-m: I changed the behavior that was there, so that it doesn't returning typing for streaming, since the text that was present in 'typing' was inconsistently being returned.  It's just returning 'message' as TextContent now and the other types are returned as TextReasoningContent.  I've got a TODO in there to revisit this, and ensure that the most appropriate content type is used.

dotnet/src/Microsoft.Agents.CopilotStudio/CopilotStudioAgentThread.cs
westey-m: Ultimately this agent is a POC and should be replaced with an A2A agent when that is available and the service supports A2A.  I don't expect us to ship this agent as it is, but it is useful right now to test what is possible with an agent that is quite different to the inference based agents.
To that end, I think for the thread anyway, what we have is OK for now.

dotnet/src/Microsoft.Agents.CopilotStudio/ActivityProcessor.cs
westey-m: For the time being the activity is stored as the raw representation of the ChatMessage, allowing users to extract the suggested actions.  I've got a todo to investigate adding a custom AI Content derived type for CopilotStudio, since I think it'll be a good POC.

Rather than a tool call response, I think a general purpose user input request or approval request may be useful where the user needs to input / approve something before the agent can continue.  I think it's something we need to investigate separately anyway.

Issue #86, opened 2025-06-19
Agent Abstraction Run Responses


westey-m:
## Goal

We need to decide what structure the output from the agent will take, taking into account the various types of data that can be produced and the use cases for agents.

## Background

Agents may produce many different types of output during a run including
1. General response messages to the user
2. Structured confirmation requests to the user
3. Tool invocation activities executed (both local and remote).  For information only.
4. Reasoning/Thinking output. For information only.
6. Handoffs / transitions from agent to agent where an agent contains sub agents.
7. An indication that the agent is responding (i.e. typing) as if it's a real human.
8. Complete messages in addition to updates, when streaming
9. and more...

When an agent is in conversation with one or more humans or agents, some of this output would typically become part of this higher level conversation (1 & 2), while other parts are secondarily useful for various reasons: e.g. to display to the user so that they can see what the agent is doing while they wait for it to respond, or so that troubleshooting can be done on agent behavior if it does something wrong.  Whether to display the secondary information will depend on the context.  I'll refer to the two types as Primary and Secondary.

We also typically want to constrain the amount of data from an agent that bubbles up into higher level conversations to avoid exceeding the context window and avoid leaking internal details.

## Decisions Drivers

- Most common use case is default and simple (most common to be agreed)
- Complex use cases supported (use cases to be listed)
- Satisfies needs of orchestration layer (to be clarified and expanded on)
- Solution is competitive when compared with alternative SDKs.

## Potential options (Work in progress)

- Return information as various response message types and let the caller filter the responses as needed.
- Return information as various response messages types and provide a filter option to return `Primary-Only` or `All`.
- Return `Primary-Only` for non-streaming and `All` for streaming.
- Return `Primary-Only` and provide a per-invocation subscription mechanism for the rest.
- Return `Primary-Only` and provide an agent level subscription mechanism for the rest.

## Comparison with other SDKs / Protocols (Work in progress)

Approaches observed from the compared SDKs:
1. Response object with separate properties for Primary and Secondary
2. Response stream that contains Primary and Secondary entries and callers need to filter.
3. Response containing just Primary.

| SDK | Non-Streaming | Streaming |
|-|-|-|
[...]

PR #104, opened 2025-06-25
Run Response ADR & Updates

westey-m:
Generates options for #86

There are a few gaps remaining in the doc and no decisions yet, but it should serve as a good starting point for discussion around this topic.

### DIFF
--- a/dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatResponseExtensions.cs
+++ b/dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatResponseExtensions.cs
@@ -44,7 +44,7 @@ internal static class ChatResponseExtensions
     {
         _ = Throw.IfNull(chatResponseUpdate);
 
-        return new AgentRunResponseUpdate()
+        return new()
         {
             AgentId = agentId,
             Role = chatResponseUpdate.Role,


docs/decisions/00NN-agent-run-response.md
```diff
@@ -0,0 +1,342 @@
...
+| AWS (Strands) | **Approach 3** Returns an [AgentResult](https://strandsagents.com/latest/api-reference/agent/#strands.agent.agent_result.AgentResult) (Primary) with messages and a reason for the run's completion. | **Approach 2** [Streams events](https://strandsagents.com/latest/api-reference/agent/#strands.agent.agent.Agent.stream_async) (Primary+Secondary) including, response text, current_tool_use, even data from "callbacks" (strands plugins) |
+| LangGraph | TBD | TBD |
+| Agno | **Approach 3** Returns a [RunResponse](https://docs.agno.com/reference/agents/run-response) object with content, messages, reasoning and thinking properties, but no secondary events.  | **Approach 2** Returns [RunResponseEvent](https://docs.agno.com/reference/agents/run-response#runresponseevent-types-and-attributes) objects including tool call, memory update, etc, information, where the [RunResponseCompletedEvent](https://docs.agno.com/reference/agents/run-response#runresponsecompletedevent) has similar properties to RunResponse|
```
rogerbarreto: Does `Agno` in the `RunResponse.Messages` return Secondary messages? Like Function Call/Response ?

docs/decisions/00NN-agent-run-response.md
```diff
@@ -0,0 +1,342 @@
...
+## Response Type Options
+
+- **Option 1** Run: Messages List contains mix of Primary and Secondary content, RunStreaming: Stream of Primary + Secondary
+  - **Option 1.1** Use ChatClient response types
+  - **Option 1.2** Return derived ChatClient response types
+- **Option 2** Run: Container with Primary and Secondary Properties, RunStreaming: Stream of Primary + Secondary
+  - **Option 2.1** Response types extend MEAI types
+  - **Option 2.2** New Response types
+- **Option 3** Run: Primary-only, RunStreaming: Stream of Primary + Secondary
+
```
rogerbarreto: ### Option 4 - <strike>Non-Streaming</strike>
Non-Streaming messages can be represented as streaming updates and the caller will have a unified implementation.

The reason for this suggestion relies on the fact that Streaming will always have Secondary and the caller needs to implement their code-base considering/ignoring the Secondary messages. 

By having a single API we might simplify the user experience, providing Extension methods that can eventually build the ChatResponse result, similar on how we see here.

https://github.com/microsoft/agent-framework/blob/e7b559fa62f2114fea0cac34add0e66d82431854/dotnet/src/Microsoft.Extensions.AI.AzureAIAgentsPersistent/PersistentAgentsChatClient.cs#L67

```suggestion
- **Option 4** RunStreaming Only API. Stream of Primary + Secondary.

```

docs/decisions/00NN-agent-run-response.md
westey-m: Messages more or less maps to chat history, in that it contains a system message that has the agent instructions in it, the input message for what the user asked, plus tools, pus a response message, so I suppose you could also say that it's sort of like 1 or 2 as well, but it doesn't fit well into any category exactly.

docs/decisions/00NN-agent-run-response.md
westey-m: Thanks, added another option to the list.

docs/decisions/00NN-agent-run-response.md
westey-m: I've updated the doc with more info.

docs/decisions/00NN-agent-run-response.md
```diff
@@ -0,0 +1,386 @@
...
+- Solutions provides an easy to use experience for users who are getting started and just want the answer to a question.
+- Solution must be extensible to future requirements, e.g. long running agent processes.
+- Experience is in line or better than the best in class experience from other SDKs
+
+## Response Type Options
+
+- **Option 1** Run: Messages List contains mix of Primary and Secondary content, RunStreaming: Stream of Primary + Secondary
+  - **Option 1.1** Updates do not use `TextContent`
+  - **Option 1.2** Use ChatClient response types
+  - **Option 1.3** Return derived ChatClient response types
```
eavanvalkenburg:
```suggestion
  - **Option 1.3** Return derived ChatClient response types
  - **Option 1.4** Presence of Secondary Content is determined by a runtime parameter
```

westey-m: Thanks, great point.  I had that option in mind originally too, but it slipped my mind since, ha ha.  Will add it.

docs/decisions/00NN-agent-run-response.md
```diff
@@ -0,0 +1,386 @@
...
+Run returns a `Task<ChatResponse>` and RunStreaming returns a `IAsyncEnumerable<ChatResponseUpdate>`.
+For Run, the returned `ChatResponse.Messages` contains an ordered list of messages that contain both the updates and the final response.
+The last message should be considered the final response.
+
+`ChatResponse.Text` automatically aggregates all text from any `TextContent` items in all `ChatMessage` items in the response.
+If we can ensure that no updates ever contain `TextContent`, this will mean that `ChatResponse.Text` will always contain
+the final response text. See option 1.1
+If we cannot ensure this, either the solution or usage becomes more complex, see 1.2 and 1.3.
+
+#### Option 1.1 Updates do not use `TextContent`
```
dmytrostruk: It would be also nice to briefly summarize pros and cons for each option including user experience, implementation complexity etc. This should provide even more insights to make a decision.

westey-m: Makes sense, done.

docs/decisions/00NN-agent-run-response.md
```diff
@@ -0,0 +1,468 @@
...
+
+# Agent Run Responses Design
+
+## Context and Problem Statement
+
+Agents may produce lots of output during a run including
+
+1. General response messages to the user
+2. Structured confirmation requests to the user
+3. Tool invocation activities executed (both local and remote).  For information only.
```
stephentoub: What does "for information only" mean?

docs/decisions/00NN-agent-run-response.md
```diff
@@ -0,0 +1,468 @@
...
+
+A potential third category is:
+
+3. **Long Running**: A response that does not contain a final response or updates, but rather a reference to a long running task.
+
+### Different use cases for Primary and Secondary output
+
+To solve complex problems, many agents must be used together. These agents typically have their own capabilities and responsibilities and communicate via input messages and final responses/handoff calls, while the internal workings of each agent is not of interest to the other agents participating in solving the problem.
+
+When an agent is in conversation with one or more humans, the information that may be displayed to the user(s) can vary. E.g. When an agent is part of a conversation with multiple humans it may be asked to perform tasks by the humans, and they may not want a stream of distracting updates posted to the conversation, but rather just a final response.  On the other hand, if an agent is being used by a single human to perform a task, the human may be waiting for the agent to complete the task.  Therefore, they may be interested in getting updates of what the agent is doing.
```
stephentoub: In either case, it can be very desirable to audit everything that was done, e.g. I'm frequently looking at the logs of the copilot coding agent, both while it's running and after it's run, to understand what it did and why it did it, in part to understand how I can better help guide it in the next go around.

docs/decisions/00NN-agent-run-response.md
```diff
@@ -0,0 +1,468 @@
...
+## Decision Drivers
+
+- Solutions provides an easy to use experience for users who are getting started and just want the answer to a question.
+- Solution must be extensible to future requirements, e.g. long running agent processes.
+- Experience is in line or better than the best in class experience from other SDKs
+
+## Response Type Options
+
+- **Option 1** Run: Messages List contains mix of Primary and Secondary content, RunStreaming: Stream of Primary + Secondary
+  - **Option 1.1** Updates do not use `TextContent`
```
stephentoub: I don't see anything here about message ordering being important, and the last message always being the primary content / response. That's effectively how it works today with IChatClient.

docs/decisions/00NN-agent-run-response.md
```diff
@@ -0,0 +1,468 @@
...
+
+// Callers can still get access to all updates too.
+foreach (var update in response.Messages)
+{
+    Console.WriteLine(update.Contents.FirstOrDefault()?.GetType().Name);
+}
+```
+
+- **PROS**: Easy and familiar user experience, reuse response types from IChatClient.
+- **CONS**: Requires all implementations to avoid using `TextContent` for anything but the final response.
```
stephentoub: This is not feasible. It's already the case today that TextContent is used, because it's how textual results from the model are represented. You'll see interactions like:
- Message 1: User: Text: "What is 2 + 3?"
- Message 2: Assistant: Text: "To compute 2 + 3, I need to use the Add method.", Function Call: "Add(2, 3)"
- Message 3: Tool: Function Result: "5"
- Message 4: Assistant: Text: "The result of adding 2 and 3 is 5."

What is feasible in these situations is only using the final message if that's all you care about.
```csharp
response.Messages[^1].Text
```

In the case of streaming, the ChatResponseUpdates have a message ID on them, and ToChatResponse uses that to appropriately split up the updates into messages.

docs/decisions/00NN-agent-run-response.md
```diff
@@ -0,0 +1,468 @@
...
+
+#### Option 1.2 Presence of Secondary Content is determined by a runtime parameter
+
+We can allow callers to optionally include secondary content in the list of messages.
+Open Question: Do we allow secondary content to use `TextContent` types?
```
stephentoub: If "Secondary" means not final, then for streaming it's impossible to know (for either the dev or the agent) whether the content is primary or not. And I think there's a lot of value in consistency between streaming and non-streaming, especially in terms of the content types being used.

If this is all based on a desire to write:
```
response.Text
```
instead of
```
response.Message[^1].Text
```
we could add a ChatResponse.FinalText or LastText or LastMessageText or whatever. I'm personally not seeing the concern, but we could do so if others think it's imperative.

docs/decisions/00NN-agent-run-response.md
```diff
@@ -0,0 +1,468 @@
...
+
+```csharp
+// By default the response only has the final response messages, so text
+// contains the final response, and it's a good starting experience.
```
stephentoub: I'm not convinced only seeing text from final message is a better getting started experience. What scenarios are we basing that on? Do you have examples you can share where seeing all text from all messages is problematic?

I also continue to believe there's a lot of benefit in keeping the non-streaming and streaming as close as possible, where we strive to have the only difference being whether you see incremental results as they're available or have to wait until the end to see everything.

The main place where I'd expect response.Text vs response.Messages[^1].Text to have a significant impact would be when the final response contains something meant to be consumed programmatically, e.g. JSON that needs to be deserialized. It's why the `GetResponseAsync<T>` method only looks at response.Messages[^1].Text.

docs/decisions/00NN-agent-run-response.md
```diff
@@ -0,0 +1,468 @@
...
+
+// Callers can still get access to all updates too.
+foreach (var update in response.Messages)
+{
+    Console.WriteLine(update.Contents.FirstOrDefault()?.GetType().Name);
+}
+```
+
+- **PROS**: Easy getting started experience.
+- **CONS**: Requires custom response types.
```
stephentoub: "types"... plural?

I could see wanting an AgentChatResponse (or AgentRunResponse), especially if there's other things we think we might want to hand back. Would we want that to have a reference to the Agent instance? The AgentThread instance? Are there other things in the future we think we might want to be able to add to the response that are only relevant at the agent level?

docs/decisions/00NN-agent-run-response.md
```diff
@@ -0,0 +1,468 @@
...
+#### Option 1.3 Use ChatClient response types
+
+```csharp
+// Since text contains the aggregate output of everything that happened, the following
+// would not get the caller a nice final response text.
+var response = agent.RunAsync("Do Something");
+Console.WriteLine(response.Text);
+
+// Instead they would need to do the following:
+Console.WriteLine(response.Messages.Last().Text);
```
stephentoub: The equivalent of Last().Text could be added as an additonal property on ChatResponse, or if all of this agent stuff is in a different package and we thought Last().Text was only relevant to agents, as an extension method on ChatResponse provided by that package. But if we think it's meaningful, I'd be more inclined to just add it to the base. Let's not forget we do have the power to augment M.E.AI.Abstractions :smile:

docs/decisions/00NN-agent-run-response.md
```diff
@@ -0,0 +1,468 @@
...
+    Console.WriteLine(update.Contents.FirstOrDefault()?.GetType().Name);
+}
+```
+
+- **PROS**: Easy getting started experience.
+- **CONS**: Requires custom response types.
+
+### Option 2 Run: Container with Primary and Secondary Properties, RunStreaming: Stream of Primary + Secondary
+
+Run and RunStreaming return new types.
```
stephentoub: I think we might want this, anyway, if for no other reason than to be able to add additional agent-specific things to it, like Agent or AgentId or AgentThread or some such things.

I love IChatClient being the main workhorse behind most implementations, but not all agents will be based on it, and forcing their Run methods to still return ChatResponse doesn't sound right.

I still want them using ChatMessage, all the same AIContent types, AITool, etc., but for ChatResponse, it does seem like it makes sense to have the concrete types returned from RunAsync be specific to the agents layer.

docs/decisions/00NN-agent-run-response.md
```diff
@@ -0,0 +1,468 @@
...
+    public static AgentRunResponseReason Length { get; } = new AgentRunResponseReason("length");
+    public static AgentRunResponseReason ContentFilter { get; } = new AgentRunResponseReason("content_filter");
+}
+
+// Not included in AgentRunResponseReason compared to ChatFinishReason
+public static ChatFinishReason ToolCalls { get; } = new ChatFinishReason("tool_calls");
+```
+
+- **PROS**: Final response messages and Updates are categorised and therefore easy to distinguish and this design popular SDKs like AutoGen and OpenAI SDK. Properties that don't make that sense on the agent API surface, like ConversationId, can be removed.
+- **CONS**: Requires custom response types and design differs significantly between streaming and non-streaming.
```
stephentoub: > design differs significantly between streaming and non-streaming.

How so?

docs/decisions/00NN-agent-run-response.md
```diff
@@ -0,0 +1,468 @@
...
+    public string? AuthorName { get; set; }
+    public DateTimeOffset? CreatedAt { get; set; }
+    public UsageDetails? Usage { get; set; }
+    public object? RawRepresentation { get; set; }
+    public AdditionalPropertiesDictionary? AdditionalProperties { get; set; }
+}
+
+// Not Included in AgentRunResponseUpdate compared to ChatResponseUpdate
+public string? ConversationId { get; set; }
+public string? ModelId { get; set; }
```
stephentoub: Why isn't ModelId relevant to an update?

docs/decisions/00NN-agent-run-response.md
```diff
@@ -0,0 +1,468 @@
...
+
+public class AgentRunResponseReason // Compare with ChatFinishReason
+{
+    public static AgentRunResponseReason Stop { get; } = new AgentRunResponseReason("stop");
+    public static AgentRunResponseReason Length { get; } = new AgentRunResponseReason("length");
+    public static AgentRunResponseReason ContentFilter { get; } = new AgentRunResponseReason("content_filter");
```
stephentoub: Are reasons like Length relevant at the agent level? Is such a reason even necessary at the agent level? What do other frameworks do?

docs/decisions/00NN-agent-run-response.md
```diff
@@ -0,0 +1,468 @@
...
+- **PROS**: Single API for streaming/non-streaming
+- **CONS**: More complex to for inexperienced users.
+
+## Long Running Processes Options
+
+Some agent protocols, like A2A, support long running agentic processes. When invoking the agent
+in the non-streaming case, the agent may respond with an id of a process that was launched.
+
+The caller is then expected to poll the service to get status updates using the id.
+The caller may also subscribe to updates from the process using the id.
```
stephentoub: Separate from A2A, some inference services are starting to support the concept of long-running, e.g. Responses now supports a background mode. We should figure out how we integrate that with this, when IChatClient is what's being used to represent that service. Maybe we need an additional Cursor/SequenceNumber/etc. ID on ChatResponse{Update} / ChatOptions, and/or a new FinishReason that's pending and some way to specify in a subsequent ChatOptions that the ConversationId should be treated as a sequence number, or something like that.

We don't need to design that here, but whatever we do, it should integrate well with what we do at the agents level, in particular for ChatClientAgent.

docs/decisions/00NN-agent-run-response.md
```diff
@@ -0,0 +1,468 @@
...
+
+- **PROS**: Single API for streaming/non-streaming
+- **CONS**: More complex to for inexperienced users.
+
+## Long Running Processes Options
+
+Some agent protocols, like A2A, support long running agentic processes. When invoking the agent
+in the non-streaming case, the agent may respond with an id of a process that was launched.
+
+The caller is then expected to poll the service to get status updates using the id.
```
stephentoub: What APIs would exist at the agent level for this? I'd imagine we'd update AgentRunOptions with the ID or something like that?

It's hard to reason about any of this without seeing what the end-to-end consuming code looks like. Have you tried writing that?

docs/decisions/00NN-agent-run-response.md
```diff
@@ -0,0 +1,468 @@
...
+
+## Tool result options (Work in progress)
+
+We need to consider abstractions for `AIContent` derived types for tool call results for common tool types beyond Function calls, e.g. CodeInterpreter, WebSearch, etc.
+
+## StructuredOutputs (Work in progress)
+
+Open Questions:
+
+1. Should our agent abstraction support Structured Outputs.
```
stephentoub: Seems fairly important, no?

docs/decisions/00NN-agent-run-response.md
```diff
@@ -0,0 +1,468 @@
...
+2. If it does, it is the agent that decides what the output structure should be, or can either caller and agent decide?
```
stephentoub: What do other frameworks do here?

docs/decisions/00NN-agent-run-response.md
westey-m: It's a reference to the lack of manual tool calling capabilities with an agent.  The caller is informed that a tool was invoked, but only after the invocation has already taken place.

docs/decisions/00NN-agent-run-response.md
westey-m: Yeah, I think that makes sense. The point I'm trying to make though is mainly around there being a definite separation between the two types of output, since the intermediate updates may be treated quite differently in the UI or only shown to users on demand.
I cannot imagine it being the best default experience in a group chat where someone asks an agent that is part of the conversation to perform a task, and it starts adding lots of detailed updates to the chat.  This could be quite disruptive to the humans in the conversation, who I imagine may want to continue their conversation while the agent executes.

docs/decisions/00NN-agent-run-response.md
```diff
@@ -0,0 +1,468 @@
...
+---
+
+# Agent Run Responses Design
+
+## Context and Problem Statement
+
+Agents may produce lots of output during a run including
+
+1. General response messages to the user
+2. Structured confirmation requests to the user
```
SteveSandersonMS: Is this something that would really be part of the output stream? Ideally, requesting confirmation should be part of the internal implementation of a tool call that does something sensitive. That way the agent/LLM has no choice about whether to get confirmation and can't possibly choose to skip it, since invoking the tool is what triggers the confirmation request.

This can be true even for non-sensitive times when the agent/LLM wants to get structured info from a user. For example it could have a "RequestBookingSlot" tool that triggers a UI that lets the user pick a date.

I know that the rest of this doc isn't contingent on this, so feel free to ignore this if you like. Just trying to ensure clarity.

docs/decisions/00NN-agent-run-response.md
```diff
@@ -0,0 +1,468 @@
...
+3. Tool invocation activities executed (both local and remote).  For information only.
+4. Reasoning/Thinking output. For information only.
+5. Handoffs / transitions from agent to agent where an agent contains sub agents.
+6. An indication that the agent is responding (i.e. typing) as if it's a real human.
+7. Complete messages in addition to updates, when streaming
+8. Id for long running process that is launched
+9. and more
```
SteveSandersonMS: What about structured output as a final response? Will it be possible to have a strongly-typed agent that always takes a particular type of input and produces a particular type of final result?

docs/decisions/00NN-agent-run-response.md
```diff
@@ -0,0 +1,468 @@
...
+
+Where agents are nested, consumers would also likely would want to constrain the amount of data from an agent that bubbles up into higher level conversations to avoid exceeding the context window, therefore limiting it to final response only.
+
+### Comparison with other SDKs / Protocols
+
+Approaches observed from the compared SDKs:
+
+1. Response object with separate properties for Primary and Secondary
+2. Response stream that contains Primary and Secondary entries and callers need to filter.
+3. Response containing just Primary.
```
SteveSandersonMS: For terminology, what about just "result" plus "progress" messages/events? There are a lot of JS Promise types/libraries that have that concept.

docs/decisions/00NN-agent-run-response.md
westey-m: When I prototyped the CopilotStudio agent I observed many intermediate pieces of text being returned as `typing` activities, and a final result message as a `message` activity.  The answer was repeated three times in total and resembled something like this:
typing: The
typing: Capital
typing: of
typing: France
typing: is
typing: Paris
typing: .
typing: The
typing: Capital
typing: of
typing: France
typing: is
typing: Paris
typing: .
message: The Capital of France is Paris.

All `typing` activities shared a common base id with an ascending sub id, so there was no clear differentiation between the first typing iteration and the second.

The overall behavior was inconsistent though.  I observed when a tool was invoked, it was simply returning no typing messages. I just reran some invocations and I am not receiving typing messages right now, so this behavior may be unintended.

Regardless, observing the above led me to believe that we should be careful around assuming that all intermediate output should be part of the final response.

stephentoub: > Regardless, observing the above led me to believe that we should be careful around assuming that all intermediate output should be part of the final response.

Is "typing" here the equivalent of TextReasoningContent (which is not included in Text)? Or possibly "typing" would be surfaced in streaming and "message" would be surfaced in non-streaming, or something like that.

Clearly there's a distinction being made between typing and message.

We need to really understand the underlying protocol being employed in order to know how to correctly translate it, and we should be careful not to make designs around only observed behaviors, especially when those observations aren't consistent / conclusive.

Is there someone you can talk to in order to understand what's actually intended for those?

docs/decisions/00NN-agent-run-response.md
SteveSandersonMS: Some of this comes down to how multi-agent nested/handoff/round-robin orchestrations work, and how to know which subset of the internal conversation to bubble up through each layer and ultimately into a UI.

Being very prescriptive about ordering (e.g., "only final output bubbles up") would risk being pretty inflexible. TBH I don't know if we even have good UX patterns for multi-agent operations yet.

I'd be interested to know if this team has concepts of how nested/handoff operations will be surfaced and how the API will allow developers to pick out the correct nested information to surface in the UI (not dumping out sensitive internal implementation details, but allowing nested agents to surface user-facing progress updates without all the ancestor agents needing to manually know about and forward specific child agent message types).

docs/decisions/00NN-agent-run-response.md
SteveSandersonMS: Would be good to see some modelling of multi-agent scenarios (in different orchestration modes) and what sorts of output would be flowing between them, and which subset needs to get back to the user, to make a judgement.

docs/decisions/00NN-agent-run-response.md
westey-m: Thanks for bringing this up.  It's actually quite fundamental to the design of this Agent abstraction.  Specifically, the abstraction must allow a very broad set of agents to be implemented, including agents accessed via the A2A protocol and CopilotStudio agents.  Both of these are remote and have no tool calling support in their protocols.  The only way for these agents to request user confirmation is via a regular response to the user.  Of course, the internal implementations of these agents can still use custom code in their tools to interrupt their execution loop and return a confirmation request to the user.

I think it's really important that we don't think of agents as just an LLM with tools configured.  Agent implementations can include complex processes with multiple different LLMs being used for different tasks.  E.g. consider an agent that does deep research on a topic by browsing the web.  It may be implemented using a process with searching, fan out to browse many sites, consolidation, filtering and summarization of the results it found, etc.  Something like this is also likely to be packaged as an opaque service that can be reused easily.

docs/decisions/00NN-agent-run-response.md
westey-m: Thanks, that's good context.  I expect that for streaming the final message would be streamed last without there necessarily being any progress updates intermixed.  That said, your point about the caller not knowing that it's the final message still stands.  The agent should know if it's the final message though, and should be able to mark it as such if needed.

docs/decisions/00NN-agent-run-response.md
westey-m: > Clearly there's a distinction being made between typing and message.

Exactly, and we should ensure that we can pass on that differentiation to callers so that they have the necessary information to decide what to do with that part of the response.  Not everything returned is equally valuable.  Of course we can only pass on based on the differentiation we get from the underlying system, but that should be enough for us to determine what the final response is and what is not.

> Would be good to see some modelling of multi-agent scenarios (in different orchestration modes) and what sorts of output would be flowing between them, and which subset needs to get back to the user, to make a judgement.

This is the design doc for multi-agent-orchestration that was done as part of SK.  The new one would be based on the same design:
https://github.com/microsoft/semantic-kernel/blob/main/docs/decisions/0071-multi-agent-orchestration.md

This is the PR for converting it to be part of the new agent framework:
https://github.com/microsoft/agent-framework/pull/107

stephentoub: > Exactly, and we should ensure that we can pass on that differentiation to callers

Yes, but that does not mean one is necessarily intermediate and one is necessarily final.

docs/decisions/00NN-agent-run-response.md
westey-m: Yes, I think we do need to support this (see open questions at the bottom of the doc around structured outputs).

We need to consider what this means for opaque, non-customizable, remote agents though.  These may still of course support a specific output type, but it would be fixed rather than customizable, and whether the output matches the requested user model, may only be verifiable after a call to the remote agent.  Considering that some of these are also stateful, and their server side chat history would already be updated during the call, failing only after changes are persisted isn't the best, but this should be considered a bug that doesn't ship to production.

docs/decisions/00NN-agent-run-response.md
westey-m: Revisiting this after our separate discussion about cases where a caller is explicitly asking an LLM to explain it's actions.  In this case, if using ChatClientAgent, we may receive multiple `TextContent` items from the LLM, and they should all be part of the Primary content.  So I think this option, where we simply consider all `TextContent` to be Primary content, is still reasonable.  I think it's preferable to assuming that only the last message is Primary content, since that would simply be inaccurate in some cases (with an exception for structured output of course).

docs/decisions/00NN-agent-run-response.md
westey-m: > If "Secondary" means not final

Secondary means 'not result', rather than 'not final'.  While some agent SDKs use the term final content for Primary, it is just another way of saying: the answer to the user's question.  As discussed around the case where a user explicitly asks for the agent to explain what it is doing, and this may be returned in multiple messages, all those messages should be considered Primary content since all should be returned to the user as the user's answer.

docs/decisions/00NN-agent-run-response.md
westey-m: I think the final and intermediate terminology is problematic (I'm fixing up the doc to remove the usage of these terms to try and make this clearer).  The distinction I have in mind is:
1. content that is the response to the caller's question.
2. content that shows what the agent is doing.

docs/decisions/00NN-agent-run-response.md
westey-m: As discussed separately, looks like we have to support cases where part of the response is produced before the last message, e.g. where models are producing text explaining what they are doing, because of the request asking them to do this.

docs/decisions/00NN-agent-run-response.md
westey-m: Thanks, I've added a proposal to the doc around adding custom AgentRunResponse types.

docs/decisions/00NN-agent-run-response.md
westey-m: Thanks, makes sense.  Since we are leaning towards having a custom response type, we probably don't need to change MEAI at the moment.  Also considering the case where we need to support models that produce the result in multiple messages, the last wouldn't be that useful after all.

docs/decisions/00NN-agent-run-response.md
westey-m: Makes sense.  I've added a section for adding a custom response type.

docs/decisions/00NN-agent-run-response.md
westey-m: The difference is mainly around there being a split between the Response and Updates for non-streaming and and they are combined in a single stream for streaming.

docs/decisions/00NN-agent-run-response.md
westey-m: There may be no model involved for certain internal operations, e.g. reporting a rag search, and overall the agent may use multiple models.  Of course we could still keep the property and just fill it in when it makes sense.  It's important to note though that there isn't always the same 1 to 1 mapping between an invocation and a model as there would be for inference.

docs/decisions/00NN-agent-run-response.md
westey-m: Yes, makes sense.
As discussed, we'll revisit and design this in more detail after this update is in.

docs/decisions/00NN-agent-run-response.md
```diff
@@ -0,0 +1,447 @@
...
+
+To solve complex problems, many agents must be used together. These agents typically have their own capabilities and responsibilities and communicate via input messages and final responses/handoff calls, while the internal workings of each agent is not of interest to the other agents participating in solving the problem.
+
+When an agent is in conversation with one or more humans, the information that may be displayed to the user(s) can vary. E.g. When an agent is part of a conversation with multiple humans it may be asked to perform tasks by the humans, and they may not want a stream of distracting updates posted to the conversation, but rather just a final response.  On the other hand, if an agent is being used by a single human to perform a task, the human may be waiting for the agent to complete the task.  Therefore, they may be interested in getting updates of what the agent is doing.
+
+Where agents are nested, consumers would also likely would want to constrain the amount of data from an agent that bubbles up into higher level conversations to avoid exceeding the context window, therefore limiting it to the Primary response only.
```
eavanvalkenburg:
```suggestion
Where agents are nested, consumers would also likely want to constrain the amount of data from an agent that bubbles up into higher level conversations to avoid exceeding the context window, therefore limiting it to the Primary response only.
```

docs/decisions/00NN-agent-run-response.md
```diff
@@ -0,0 +1,447 @@
...
+    public UsageDetails? Usage { get; set; }
+    public object? RawRepresentation { get; set; }
+    public AdditionalPropertiesDictionary? AdditionalProperties { get; set; }
+}
+
+// Not Included in AgentRunResponseUpdate compared to ChatResponseUpdate
+public string? ConversationId { get; set; }
+public string? ModelId { get; set; }
+
+public class AgentRunResponseReason // Compare with ChatFinishReason
```
eavanvalkenburg:
```suggestion
public class AgentRunFinishReason // Compare with ChatFinishReason
```

docs/decisions/00NN-agent-run-response.md
westey-m: Yes, definitely worth supporting IMO.

docs/decisions/00NN-agent-run-response.md
westey-m: I've added a section to compare with other frameworks, and I've added some proposals around how to support this.

docs/decisions/00NN-agent-run-response.md
westey-m: I've filled out the section on structured outputs after comparing with other frameworks.  Please take a look.

docs/decisions/00NN-agent-run-response.md
westey-m: Thanks, fixed everwhere.

docs/decisions/00NN-agent-run-response.md
westey-m: Thanks, fixed.

docs/decisions/00NN-agent-run-response.md
westey-m: I've added a comparison with other frameworks, see the "Response Reason Support" section.  Only two of them support something like this, and the one is freeform text, while the other is really just what the model produced.

I'd be reluctant to support this unless we had a clear use case for it.  I don't have one right now.

stephentoub: So... it'll be removed, right? That was really the intent of my question. I rarely see use for this at the inference level, nevermind at the agents level.

westey-m: Just removed, thanks!

docs/decisions/00NN-agent-run-response.md
westey-m: I haven't tried.  As discussed offline, we'll revisit this in a separate design after this initial change is in.

docs/decisions/00NN-agent-run-response.md
westey-m: Yeah, I like this terminology as well.  I still think Result might be misunderstood but it's probably one of the clearest options.
I tried to update the document to be a little bit clearer anyway, although I may switch to this terminology more consistently everywhere when I get a sec.
Note that with the chosen direction this differentiation won't show up in the actual code so at least no naming is required in the code.

docs/decisions/00NN-agent-run-response.md
markwallace-microsoft, line 1: Change NN to the appropriate number

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AgentRunResponse.cs
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    /// <summary>
+    /// Tries to parse the response as structured output using the specified serializer options.
+    /// </summary>
+    /// <typeparam name="T">The type of structured output to parse.</typeparam>
+    /// <param name="serializerOptions">The JSON serialization options to use.</param>
+    /// <param name="structuredOutput">The parsed structured output.</param>
+    /// <returns><see langword="true" /> if parsing was successful; otherwise, <see langword="false" />.</returns>
+    public bool TryParseAsStructuredOutput<T>(JsonSerializerOptions serializerOptions, [NotNullWhen(true)] out T? structuredOutput)
```
markwallace-microsoft: Do we need both methods?

docs/decisions/00NN-agent-run-response.md
```diff
@@ -0,0 +1,515 @@
...
+
+## Addendum 1: AIContext Derived Types for different response types / Gap Analysis (Work in progress)
+
+We need to decide what AIContent types, each agent response type will be mapped to.
+
+| Number | DataType | AIContent Type |
+|-|-|-|
+| 1. | General response messages to the user | TextContent + DataContent + UriContent |
+| 2. | Structured confirmation requests to the user | ? |
+| 3. | Function invocation activities executed (both local and remote). For information only. | FunctionCallContent + FunctionResultContent |
```
eavanvalkenburg: Since ChatResponse has multiple ChatMessages and each ChatMessage can have multiple contents, is this supposed to go into a single ChatMessage, or will the ChatResponse have one ChatMessage with function calls, and one ChatMessage with FRC?

docs/decisions/00NN-agent-run-response.md
```diff
@@ -0,0 +1,515 @@
...
+| 4. | Tool invocation activities executed (both local and remote). For information only. | FunctionCallContent/FunctionResultContent/Custom ? |
```
eavanvalkenburg: I noticed OpenAI uses Annotations for this (and not a set of FCC + FRC), should we add (or something similar) that as a prime AIContent, so that we can expose that as a secondary content for an agent?

docs/decisions/00NN-agent-run-response.md
```diff
@@ -0,0 +1,515 @@
...
+| 5. | Reasoning/Thinking output. For information only. | TextReasoningContent |
+| 6. | Handoffs / transitions from agent to agent. | ? |
+| 7. | An indication that the agent is responding (i.e. typing) as if it's a real human. | ? |
+| 8. | Complete messages in addition to updates, when streaming | TextContent |
+| 9. | Id for long running process that is launched | ? |
+| 10. | Memory storage / lookups (are these just traces?) | ? |
```
eavanvalkenburg: This could also be AnnotationContent...

docs/decisions/00NN-agent-run-response.md
stephentoub: Yes, we should be figuring out the right set of AIContent-derived types to add to express this, coupled with possibly adding additional properties to the existing ones.
https://github.com/dotnet/extensions/issues/6057

docs/decisions/00NN-agent-run-response.md
stephentoub: You typically end up with one message per request or response. So if I make a request, the model responds with three parallel tool requests, automatic function invocation invokes the tools and issues another intermediate request with the three tool responses, and another response comes back, you'll have four messages total. All the ones that were created as part of the logical operation will be in the ChatResponse object. So I send my single initial message, and I get back a ChatResponse containing three messages; the first contains three FCCs, the second contains three FRCs, and the last contains a TextContent or whatever.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AgentRunResponse.cs
westey-m: Strictly speaking no, but it is a fairly common pattern in .net to have a Try method that doesn't throw and an equivalent non-try method that does.

docs/decisions/00NN-agent-run-response.md
westey-m: I think for the agent output, it makes sense to replicate the structure that the underlying service produced. The reason why a user may be interested in all these messages is to reconstruct and understand the process leading up to the agent result.  Consolidation would potentially be lossy.

docs/decisions/00NN-agent-run-response.md
westey-m: Agreed, this is something that is still outstanding.  It will require more investigation, but proposing to complete what we have here first and then investigate what additional types we require separately.

docs/decisions/00NN-agent-run-response.md
```diff
@@ -0,0 +1,515 @@
+---
+# These are optional elements. Feel free to remove any of them.
+status: proposed {proposed | rejected | accepted | deprecated | … | superseded by [ADR-0001](0001-madr-architecture-decisions.md)}
+contact: westey-m {person proposing the ADR}
+date: 2025-06-24 {YYYY-MM-DD when the decision was last updated}
+deciders: {list everyone involved in the decision}
+consulted: {list everyone whose opinions are sought (typically subject-matter experts); and with whom there is a two-way communication}
+informed: {list everyone who is kept up-to-date on progress; and with whom there is a one-way communication}
```
rogerbarreto: update the header.

docs/decisions/00NN-agent-run-response.md
```diff
@@ -0,0 +1,515 @@
...
+
+// Callers can also iterate through all updates if needed
+await foreach (var update in responses)
+{
+    Console.WriteLine(update.Contents.FirstOrDefault()?.GetType().Name);
+}
+```
+
+- **PROS**: Single API for streaming/non-streaming
+- **CONS**: More complex to for inexperienced users.
```
rogerbarreto: My initial idea would be that we do still have a `RunAsync` but that is a extension method instead of the caller having to transforming RunStreamingAsync output.

docs/decisions/00NN-agent-run-response.md
```diff
@@ -0,0 +1,515 @@
...
+        CancellationToken cancellationToken = default);
+
+    public abstract IAsyncEnumerable<AgentRunResponseUpdate> RunStreamingAsync(
+        IReadOnlyCollection<ChatMessage> messages,
+        AgentThread? thread = null,
+        AgentRunOptions? options = null,
+        CancellationToken cancellationToken = default);
+}
+
+class AgentRunResponse // Compare with ChatResponse
```
eavanvalkenburg: we could add `AgentId` to allow one to track which agent did what, across a longer run.

docs/decisions/00NN-agent-run-response.md
```diff
@@ -0,0 +1,515 @@
...
+    public UsageDetails? Usage { get; set; }
+    public AdditionalPropertiesDictionary? AdditionalProperties { get; set; }
+}
+
+// Not Included in AgentRunResponse compared to ChatResponse
+public ChatFinishReason? FinishReason { get; set; }
+public string? ConversationId { get; set; }
+public string? ModelId { get; set; }
+
+public class AgentRunResponseUpdate // Compare with ChatResponseUpdate
```
eavanvalkenburg: same here

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AgentRunResponse.cs
dmytrostruk: nit: I'm wondering if `Parse<T>` and `TryParse<T>` or `Deserialize<T>` and `TryDeserialize<T>` would be sufficient names, since the logic is all about deserializing response to specific type and I'm not sure if `StructuredOutput` part actually means something here.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AgentRunResponse.cs
```diff
@@ -0,0 +1,240 @@
...
+
+    /// <summary>Gets or sets any additional properties associated with the run response.</summary>
+    public AdditionalPropertiesDictionary? AdditionalProperties { get; set; }
+
+    /// <inheritdoc />
+    public override string ToString() => this.Text;
+
+    /// <summary>Creates an array of <see cref="AgentRunResponseUpdate" /> instances that represent this <see cref="AgentRunResponse" />.</summary>
+    /// <returns>An array of <see cref="AgentRunResponseUpdate" /> instances that may be used to represent this <see cref="AgentRunResponse" />.</returns>
+    public AgentRunResponseUpdate[] ToAgentRunResponseUpdates()
```
dmytrostruk, line 106: I understand the use-case for `AgentRunResponseUpdate[]` to `AgentRunResponse` conversion, when you want to aggregate multiple updates into a single response in order to handle it all together, but what would be the use-case for `AgentRunResponse` to `AgentRunResponseUpdate[]` conversion?

stephentoub: When the thing you're talking to doesn't actually support streaming, so you implement the streaming entrypoint in terms of the non-streaming.

docs/decisions/00NN-agent-run-response.md
westey-m: I've logged issue #158 to investigate this further from an Agent framework perspective.

docs/decisions/00NN-agent-run-response.md
westey-m: I've logged issue #158 to investigate this further from an Agent framework perspective.

docs/decisions/00NN-agent-run-response.md
westey-m: Thanks, Makes sense, added.

docs/decisions/00NN-agent-run-response.md
westey-m: Thanks, Makes sense, added.

docs/decisions/00NN-agent-run-response.md
westey-m: I think this suggestion makes more sense as an implementation detail of option 1.1.  E.g. an implementer can still implement RunAsync to just be a transformation on top of the RunStreamingAsync.  For cases where the service does support a non-streaming api though, they can do the extra work and map the non-streaming api directly.
Only having RunAsync as an extension method, blocks the second option, so having RunAsync on the Agent class is more flexible.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AgentRunResponse.cs
westey-m: Yeah, I like the deserialize option.  I think it's pretty clear that it would deserialize the response into the given type, so StructuredOutput is not necessary.  I'll switch to that naming.

docs/decisions/00NN-agent-run-response.md
westey-m: Done.

docs/decisions/00NN-agent-run-response.md
westey-m: Done, thanks.

dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatResponseExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+
+    /// <summary>
+    /// Converts a <see cref="ChatResponseUpdate"/> instance to an <see cref="AgentRunResponseUpdate"/>.
+    /// </summary>
+    /// <param name="chatResponseUpdate">The <see cref="ChatResponseUpdate"/> to convert. Cannot be <see langword="null"/>.</param>
+    /// <param name="agentId">The ID of the agent that generated the response. Cannot be <see langword="null"/>.</param>
+    /// <returns>An <see cref="AgentRunResponseUpdate"/> containing the properties from the specified <see cref="ChatResponseUpdate"/>.</returns>
+    public static AgentRunResponseUpdate ToAgentRunResponseUpdate(this ChatResponseUpdate chatResponseUpdate, string agentId)
+    {
+        _ = Throw.IfNull(chatResponseUpdate);
```
rogerbarreto, line 45: If agentId is not required, we might reflect it in the property or make it required as the `ToAgentRunResponse`.

```suggestion
    public static AgentRunResponseUpdate ToAgentRunResponseUpdate(this ChatResponseUpdate chatResponseUpdate, string? agentId = null)
    {
        _ = Throw.IfNull(chatResponseUpdate);
```

dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatResponseExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        return new AgentRunResponseUpdate()
```
rogerbarreto: nit 

```suggestion
        return new()
```

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIContentExtensions.cs
```diff
@@ -0,0 +1,118 @@
...
+#endif
+
+namespace Microsoft.Extensions.AI.Agents;
+
+// TODO: Consolidate with same internal class in Microsoft.Extensions.AI.Abstractions when both are available in the same repository.
+/// <summary>Internal extensions for working with <see cref="AIContent"/>.</summary>
+internal static class AIContentExtensions
+{
+    /// <summary>Concatenates the text of all <see cref="TextContent"/> instances in the list.</summary>
+    public static string ConcatText(this IEnumerable<AIContent> contents)
```
rogerbarreto, line 21: Do we already have this extension in MEAI right? Should we just assume the consumers will have to rely also in the `Extensions.AI` namespace already?

```suggestion
```

westey-m: It's actually internal to MEAI right now, and it's internal here too.  If this ends up in the MEAI library then we can merge the two.
See the todo at the top of the class that reminds around this:
```chsarp
// TODO: Consolidate with same internal class in Microsoft.Extensions.AI.Abstractions when both are available in the same repository.
```

dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatResponseExtensions.cs
westey-m: While agent id is optional on the response, it is never optional for ChatClientAgent, since we always have an ID here.  And this method is specific to ChatClientAgent.

dotnet/src/Microsoft.Agents.Orchestration/AgentActor.cs
```diff
@@ -111,54 +111,46 @@ protected async ValueTask<ChatMessage> InvokeAsync(IEnumerable<ChatMessage> inpu
...
+            this.Context.Cancellation.ThrowIfCancellationRequested();
 
-            if (this.Context.ResponseCallback is not null)
-            {
-                await this.Context.ResponseCallback.Invoke(messages).ConfigureAwait(false);
-            }
+            await HandleStreamedMessage(lastStreamedResponse, isFinal: false).ConfigureAwait(false);
+
+            lastStreamedResponse = streamedResponse;
         }
```
stephentoub, line 139: FYI, this now appears to be dropping the last update.

I'm fixing it in my PR https://github.com/microsoft/agent-framework/pull/164 as I'm resolving conflicts with this one.

westey-m: Oops, thanks for fixing.

PR #105, opened 2025-06-25
.Net: Agent Samples - Adding AzureAIAgentsPersistent to Steps

rogerbarreto:
# Motivation

This change is a step forward how we expect the caller to use agents without modifying much of the experience, extracting the specifics from the steps, while keeping the `Providers` folder with the full implementation.

- This change adds the option for using AzureAIAgentsPersistent provider to the Samples.
- Updates the Samples_01 to actually target the `ChatClientAgent` component, renamed to `Step01_ChatClientAgent_Running`.
- Give another round of consistency across the different samples.

### DIFF
--- a/dotnet/samples/GettingStarted/AgentSample.cs
+++ b/dotnet/samples/GettingStarted/AgentSample.cs
@@ -27,12 +27,12 @@ public class AgentSample(ITestOutputHelper output) : BaseSample(output)
         AzureAIAgentsPersistent
     }
 
-    protected Task<IChatClient> GetChatClientAsync(ChatClientProviders provider, ChatClientAgentOptions options)
+    protected Task<IChatClient> GetChatClientAsync(ChatClientProviders provider, ChatClientAgentOptions options, CancellationToken cancellationToken = default)
         => provider switch
         {
             ChatClientProviders.OpenAI => GetOpenAIChatClientAsync(),
             ChatClientProviders.AzureOpenAI => GetAzureOpenAIChatClientAsync(),
-            ChatClientProviders.AzureAIAgentsPersistent => GetAzureAIAgentPersistentClientAsync(options),
+            ChatClientProviders.AzureAIAgentsPersistent => GetAzureAIAgentPersistentClientAsync(options, cancellationToken),
             ChatClientProviders.OpenAIResponses or
             ChatClientProviders.OpenAIResponses_InMemoryMessageThread or
             ChatClientProviders.OpenAIResponses_ConversationIdThread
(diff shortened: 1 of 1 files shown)


dotnet/samples/GettingStarted/AgentSample.cs
```diff
@@ -20,45 +22,113 @@ public enum ChatClientProviders
...
-            ChatClientProviders.AzureOpenAI => GetAzureOpenAIChatClient(),
+            ChatClientProviders.OpenAI => GetOpenAIChatClientAsync(),
+            ChatClientProviders.AzureOpenAI => GetAzureOpenAIChatClientAsync(),
+            ChatClientProviders.AzureAIAgentsPersistent => GetAzureAIAgentPersistentClientAsync(options),
             ChatClientProviders.OpenAIResponses or
-            ChatClientProviders.OpenAIResponses_InMemoryMessage or
-            ChatClientProviders.OpenAIResponses_ConversationId
-            => GetOpenAIResponsesClient(),
+            ChatClientProviders.OpenAIResponses_InMemoryMessageThread or
+            ChatClientProviders.OpenAIResponses_ConversationIdThread
```
markwallace-microsoft, line 38: Why do we need 3 ChatClientProviders for OpenAIResponses if they all return the same IChatClient?

dotnet/samples/GettingStarted/AgentSample.cs
```diff
@@ -20,45 +22,113 @@ public enum ChatClientProviders
...
+                => GetOpenAIResponsesClientAsync(),
             _ => throw new NotSupportedException($"Provider {provider} is not supported.")
         };
 
     protected ChatOptions? GetChatOptions(ChatClientProviders? provider)
```
markwallace-microsoft, line 43: Would this be better: `GetChatOptions(ChatClientProviders? provider, bool storeEnabled)`

dotnet/samples/GettingStarted/AgentSample.cs
```diff
@@ -20,45 +22,113 @@ public enum ChatClientProviders
...
+        };
+    }
+
+    #region Private GetChatClient
+
+    private Task<IChatClient> GetOpenAIChatClientAsync()
+        => Task.FromResult(
+                new OpenAIClient(TestConfiguration.OpenAI.ApiKey)
+                    .GetChatClient(TestConfiguration.OpenAI.ChatModelId)
+                    .AsIChatClient());
```
markwallace-microsoft, line 79: The `.AsIChatClient()` seems odd, is there a way to eliminate this?

dotnet/samples/GettingStarted/AgentSample.cs
rogerbarreto: There are three different ways of using this same provider in the samples, and part of the thread configuration is related how a specific provider configuration needed for the `ChatOptions`.

- Using without threads
- Using with Threads but IN-Memory
- Using with Threads but Server-Side (ConversationId).

dotnet/samples/GettingStarted/AgentSample.cs
rogerbarreto: Not really because `storeEnable` is a feature specific for `OpenAIResponses` provider. Trying to hide this from the Samples as much as possible, showing that all they really need is a `IChatClient` from a specific provider to get the sample code working.

dotnet/samples/GettingStarted/AgentSample.cs
rogerbarreto: This will be removed once the `OpenAI SDK` implements `IChatClient`. Otherwise we need to convert their `OpenAI.ChatClient` into a `MEAI.IChatClient`.

PR #107, opened 2025-06-25
.NET Port Agent Orchestration

crickman:
Port agent orchestrations from SK repo.

Migrates code "as-is" except where patterns must be adjusted due to:

- Dependencies on new agent framework
- Dependency on M.E.AI (instead of SK)

> Note: Referencing agent runtime packages published from SK repo until an improved version becomes available.


dotnet/Directory.Packages.props
```diff
@@ -23,8 +23,15 @@
     <PackageVersion Include="Microsoft.Extensions.AI.Abstractions" Version="9.6.0" />
     <PackageVersion Include="Microsoft.Extensions.DependencyInjection" Version="8.0.1" />
     <PackageVersion Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="9.0.5" />
+    <PackageVersion Include="Microsoft.Extensions.Hosting" Version="8.0.1" />
+    <PackageVersion Include="Microsoft.Extensions.Logging" Version="8.0.1" />
     <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="9.0.6" />
+    <!-- Microsoft.SemanticKernel.Agents.Runtime.* -->
+    <PackageVersion Include="Microsoft.SemanticKernel.Agents.Runtime.Abstractions" Version="1.58.0-preview" />
+    <PackageVersion Include="Microsoft.SemanticKernel.Agents.Runtime.Core" Version="1.58.0-preview" />
+    <PackageVersion Include="Microsoft.SemanticKernel.Agents.Runtime.InProcess" Version="1.58.0-preview" />
```
stephentoub, line 32: We shouldn't have any dependencies here to existing libraries that are necessary to enable what's being added her. What are these?

dotnet/samples/GettingStarted/Orchestration/HandoffOrchestration_With_StructuredInput.cs
```diff
@@ -0,0 +1,124 @@
...
+        OrchestrationMonitor monitor = new();
+
+        // Define the orchestration
+        HandoffOrchestration<GithubIssue, string> orchestration =
+            new(OrchestrationHandoffs
+                    .StartWith(triageAgent)
+                    .Add(triageAgent, dotnetAgent, pythonAgent),
+                triageAgent,
+                pythonAgent,
+                dotnetAgent)
```
stephentoub, line 55: Is there a way we could avoid the developer needing to name all the agents, again, after they've already specified what the handoffs are? Why are the agents needed separately?

dotnet/samples/GettingStarted/Orchestration/HandoffOrchestration_With_StructuredInput.cs
```diff
@@ -0,0 +1,124 @@
...
+            {
+                LoggerFactory = this.LoggerFactory,
+                ResponseCallback = monitor.ResponseCallback,
```
stephentoub, line 58: This is all optional and used only for the samples, right? i.e. a developer doesn't need to provide such callbacks, they only do so if they want observability into the inner workings of the orchestration? How does that relate to any configuration they may have done on the individual agents?

dotnet/samples/GettingStarted/Orchestration/HandoffOrchestration_With_StructuredInput.cs
```diff
@@ -0,0 +1,124 @@
...
+
+                    Additional context
+                    Normal DBContext logging shows only normal context queries. Queries run by VectorizedSearchAsync() don't appear in those logs and I could not find a way to enable logging in semantic search so that I could actually see the exact query that is failing. It would have been very useful to see the failing semantic query.                    
+                    """,
+                Labels = []
+            };
+
+        // Start the runtime
+        await using InProcessRuntime runtime = new();
+        await runtime.StartAsync();
```
stephentoub, line 88: What does starting the "in process" runtime do?

dotnet/src/Microsoft.Agents.Orchestration/OrchestrationActor.cs
```diff
@@ -0,0 +1,44 @@
...
+using Microsoft.Extensions.Logging;
+using Microsoft.SemanticKernel.Agents.Runtime;
+using Microsoft.SemanticKernel.Agents.Runtime.Core;
+
+namespace Microsoft.Agents.Orchestration;
+
+/// <summary>
+/// Base abstractions for any actor that participates in an orchestration.
+/// </summary>
+public abstract class OrchestrationActor : BaseAgent
```
stephentoub, line 14: What is BaseAgent?

We have the Agent type that's part of the core layer; we really shouldn't have two, nevermind a separate layer of "actor"s on top. Does this all need to be public, or is it an implementation detail?

If it needs to be public, we should instead be looking at what functionality is necessary from all of this to add to Agent so we don't have duplicative concepts. There seems to be problematic overlap here with Agent, AgentThread, etc.

@westey-m

dotnet/src/Microsoft.Agents.Orchestration/AgentOrchestration.cs
```diff
@@ -0,0 +1,256 @@
...
+    /// <param name="runtime">The runtime associated with the orchestration.</param>
+    /// <param name="cancellationToken">A cancellation token that can be used to cancel the operation.</param>
+    public async ValueTask<OrchestrationResult<TOutput>> InvokeAsync(
+        TInput input,
+        IAgentRuntime runtime,
+        CancellationToken cancellationToken = default)
+    {
+        Throw.IfNull(input, nameof(input));
+
+        TopicId topic = new($"{this.OrchestrationLabel}_{Guid.NewGuid().ToString().Replace("-", string.Empty)}");
```
stephentoub, line 116:
```suggestion
        TopicId topic = new($"{this.OrchestrationLabel}_{Guid.NewGuid():N}");
```

dotnet/src/Microsoft.Agents.Orchestration/AgentOrchestration.cs
```diff
@@ -0,0 +1,256 @@
...
+                this.LoggerFactory,
+                cancellationToken);
+
+        ILogger logger = this.LoggerFactory.CreateLogger(this.GetType());
+
+        TaskCompletionSource<TOutput> completion = new();
+
+        AgentType orchestrationType = await this.RegisterAsync(runtime, context, completion, handoff: null).ConfigureAwait(false);
+
+        cancellationToken.ThrowIfCancellationRequested();
```
stephentoub, line 134: Nit: it'd be better to do this at the beginning of the method, after argument validation, rather than here

dotnet/src/Microsoft.Agents.Orchestration/AgentOrchestration.cs
```diff
@@ -0,0 +1,256 @@
...
+                                runtime,
+                                context,
+                                resultTransform,
+                                outputTransform,
+                                completion,
+                                context.LoggerFactory.CreateLogger<ResultActor<TResult>>());
+#if !NETCOREAPP
+                        return new ValueTask<IHostableAgent>(actor);
+#else
+                        return ValueTask.FromResult<IHostableAgent>(actor);
```
stephentoub, line 250: Nit: there's no benefit to using FromResult; it's just a nicety that in turn delegates to the ctor. You can keep the code simpler and just always use the ctor

dotnet/src/Microsoft.Agents.Orchestration/GroupChat/GroupChatAgentActor.cs
```diff
@@ -0,0 +1,73 @@
...
+    /// <inheritdoc/>
+    public ValueTask HandleAsync(GroupChatMessages.Reset item, MessageContext messageContext)
+    {
+        this.ResetThread();
+
+#if !NETCOREAPP
+        return new ValueTask();
+#else
+        return ValueTask.CompletedTask;
+#endif
```
stephentoub, line 59:
```suggestion
        return default;
```

dotnet/src/Microsoft.Agents.Orchestration/GroupChat/GroupChatOrchestration.cs
```diff
@@ -0,0 +1,99 @@
...
+
+namespace Microsoft.Agents.Orchestration.GroupChat;
+
+/// <summary>
+/// An orchestration that coordinates a group-chat.
+/// </summary>
+public class GroupChatOrchestration<TInput, TOutput> :
+    AgentOrchestration<TInput, TOutput>
+{
+    internal const string DefaultAgentDescription = "A helpful agent.";
```
stephentoub, line 21: Why do we need to manufacture a description?

dotnet/src/Microsoft.Agents.Orchestration/GroupChat/GroupChatOrchestration.cs
```diff
@@ -0,0 +1,99 @@
...
+            ++agentCount;
+            AgentType agentType = await RegisterAgentAsync(agent, agentCount).ConfigureAwait(false);
+            string name = agent.Name ?? agent.Id ?? agentType;
+            string? description = agent.Description;
+
+            team[name] = (agentType, description ?? DefaultAgentDescription);
+
+            logger.LogRegisterActor(this.OrchestrationLabel, agentType, "MEMBER", agentCount);
+
+            await runtime.SubscribeAsync(agentType, context.Topic).ConfigureAwait(false);
```
stephentoub, line 66: What's the difference between registering and subscribing?

dotnet/src/Microsoft.Agents.Orchestration/GroupChat/GroupChatTeam.cs
```diff
@@ -0,0 +1,32 @@
...
+
+/// <summary>
+/// Describes a team of agents participating in a group chat.
+/// </summary>
+public class GroupChatTeam : Dictionary<string, (string Type, string Description)>;
+
+/// <summary>
+/// Extensions for <see cref="GroupChatTeam"/>.
+/// </summary>
+public static class ChatGroupExtensions
```
stephentoub, line 17: Can things like this be internal?

In general I'm concerned about the breadth / complexity of everything. I'd much prefer to start as minimalist as possible.

dotnet/src/Microsoft.Agents.Orchestration/GroupChat/RoundRobinGroupChatManager.cs
```diff
@@ -0,0 +1,55 @@
...
+
+namespace Microsoft.Agents.Orchestration.GroupChat;
+
+/// <summary>
+/// A <see cref="GroupChatManager"/> that selects agents in a round-robin fashion.
+/// </summary>
+/// <remarks>
+/// Subclass this class to customize filter and user interaction behavior.
+/// </remarks>
+public class RoundRobinGroupChatManager : GroupChatManager
```
stephentoub, line 17: Have all of these different modes of orchestration proved useful? Are there any we can cull?

dotnet/samples/GettingStarted/Orchestration/ConcurrentOrchestration_Intro.cs
```diff
@@ -0,0 +1,61 @@
...
+            };
+
+        // Start the runtime
+        await using InProcessRuntime runtime = new();
+        await runtime.StartAsync();
+
+        // Run the orchestration
+        string input = "What is temperature?";
+        Console.WriteLine($"\n# INPUT: {input}\n");
+        OrchestrationResult<string[]> result = await orchestration.InvokeAsync(input, runtime);
```
westey-m, line 52: It would be nice to have an overload that produces the output and also constructs an InProcess runtime if none is provided, so callers can go from input to answer in one call, E.g.:
```csharp
        string[] output = await orchestration.RuntToCompletionAsync(input);
```
This would help simplify the getting started experience for developers.

dotnet/samples/GettingStarted/Orchestration/HandoffOrchestration_Intro.cs
```diff
@@ -0,0 +1,109 @@
...
+            new(OrchestrationHandoffs
+                    .StartWith(triageAgent)
+                    .Add(triageAgent, statusAgent, returnAgent, refundAgent)
+                    .Add(statusAgent, triageAgent, "Transfer to this agent if the issue is not status related")
+                    .Add(returnAgent, triageAgent, "Transfer to this agent if the issue is not return related")
+                    .Add(refundAgent, triageAgent, "Transfer to this agent if the issue is not refund related"),
+                triageAgent,
+                statusAgent,
+                returnAgent,
+                refundAgent)
```
westey-m, line 72: We should be able to build this list from the agents passed into `.StartWith` and `.Add`, so that we don't have to pass them in a second time here.

dotnet/samples/GettingStarted/Orchestration/HandoffOrchestration_Intro.cs
```diff
@@ -0,0 +1,109 @@
...
+            {
+                InteractiveCallback = () =>
```
westey-m, line 74: How would the InteractiveCallback work when the orchestration is running in a service and we need user input from a user using a web client?

stephentoub:
There's a lot to do here. I'm going to merge it so we can iterate.

PR #108, opened 2025-06-26
Python package design

eavanvalkenburg:
Design for python package

### DIFF
--- a/docs/design/python-package-setup.md
+++ b/docs/design/python-package-setup.md
@@ -4,9 +4,9 @@
 * Developer experience is key
     * the components needed for a basic agent with tools and a runtime should be importable from `agent_framework` without having to import from subpackages. This will be referred to as _tier 0_ components.
     * for more advanced components, _tier 1_ components, such as context providers, guardrails, vector data, text search, exceptions, evaluation, utils, telemetry and workflows, they should be importable from `agent_framework.<component>`, so for instance `from agent_framework.vector_data import vectorstoremodel`.
-    * for connectors (_tier 2_), a developer should never have to import from more than 2 levels deep
-        * i.e.: `from agent_framework.connectors.openai import OpenAIClient`
-        * Question: should we shorten this even further: either `from agent_framework.openai import OpenAIClient` or maybe something like `from agent_framework.ext.openai import OpenAIClient`?
+    * for parts of the package that are either additional functionality or integrations with other services (connectors) (_tier 2_), we use the term _tier 2_, however they should also be importable from `agent_framework.<component>`, so for instance `from agent_framework.openai import OpenAIClient`.
+        * this means that the package structure is flat, and the components are grouped by functionality, not by type, so for instance `from agent_framework.openai import OpenAIChatClient` will import the OpenAI chat client, but also the OpenAI tools, and any other OpenAI related functionality.
+        * There should not be a need for deeper imports from those packages, unless a good case is made for that, so the internals of the extensions packages should always be a folder with the name of the package, a `__init__.py` and one or more `_files.py` file, where the `_files.py` file contains the implementation details, and the `__init__.py` file exposes the public interface.
     * if a single file becomes too cumbersome (files are allowed to be 1k+ lines) it should be split into a folder with an `__init__.py` that exposes the public interface and a `_files.py` that contains the implementation details, with a `__all__` in the init to expose the right things, if there are very large dependencies being loaded it can optionally using lazy loading to avoid loading the entire package when importing a single component.
     * as much as possible, related things are in a single file which makes understanding the code easier.
     * simple and straightforward logging and telemetry setup, so developers can easily add logging and telemetry to their code without having to worry about the details.
@@ -14,7 +14,7 @@
     * To allow connectors to be treated as independent packages, we will use namespace packages for connectors, in principle this only includes the packages that we will develop in our repo, since that is easy to manage and maintain.
     * further advantages are that each package can have a independent lifecycle, versioning, and dependencies.
     * and this gives us insights into the usage, through pip install statistics, especially for connectors to services outside of Microsoft.
-    * the goal is to group related connectors based on vendors, not on types, so for instance doing: `import agent_framework.connectors.google` will import connectors for all Google services, such as `GoogleChatClient` but also `BigQueryCollection`, etc.
+    * the goal is to group related connectors based on vendors, not on types, so for instance doing: `import agent_framework.google` will import connectors for all Google services, such as `GoogleChatClient` but also `BigQueryCollection`, etc.
     * All dependencies for a subpackage should be required dependencies in that package, and that package becomes a optional dependency in the main package as an _extra_ with the same name, so in the main `pyproject.toml` we will have:
         ```toml
         [project.optional-dependencies]
@@ -28,7 +28,7 @@
 ```python
 from typing import Annotated
 from agent_framework import Agent, ai_tool
-from agent_framework.connectors.openai import OpenAIChatClient
+from agent_framework.openai import OpenAIChatClient
 
 @ai_tool(description="Get the current weather in a given location")
 async def get_weather(location: Annotated[str, "The location as a city name"]) -> str:
@@ -65,15 +65,17 @@ Overall the following structure is proposed:
         * utils (optional)
         * telemetry (could also be observability or monitoring)
         * workflows (includes multi-agent orchestration)
-    * tier 2; extensions
-        * connectors (namespace package)
+    * tier 2; extensions 
+        * Extensions are any additional functionality that is useful for a user, to reduce friction they will imported in a similar way as tier 1, however the code for them will be in a separate package, so that they can be installed separately, they must have everything in a folder with the same name as the package, and without a `__init__.py` file in the root, so that they can be used as a namespace package.
+        * Some examples are:
             * openai
             * azure
-            * will be exposed through i.e. `agent_framework.connectors.openai` and `agent_framework.connectors.azure`
-        * extensions (namespace package)
+            * will be exposed through i.e. `agent_framework.openai` and `agent_framework.azure`
             * anything other then a connector that we want to expose as a separate package, for instance:
                 * mem0 (memory management)
-                * would be exposed through i.e. `agent_framework.extensions.mem0`
+                * would be exposed through i.e. `agent_framework.mem0`
+        * A package name cannot overlap with each other or with components in the main package, so `guardrails` cannot be used as a package name, since it is already used in the main package.
+            * A package can implement additional guardrails functionality, but that would become something like `from agent_framework.azure import ContentSafetyGuardrail`, which would contain a guardrail implementation using Azure Content Safety, but it is not the same as the `agent_framework.guardrails` component.
 * tests
 * samples
 * extensions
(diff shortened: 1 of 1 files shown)


docs/design/python-package-setup.md
```diff
@@ -0,0 +1,303 @@
+# Python Package design for Agent Framework
+
+## Design goals
+* Developer experience is key
+    * the components needed for a basic agent with tools and a runtime should be importable from `agent_framework` without having to import from subpackages. This will be referred to as _tier 0_ components.
+    * for more complex pieces, a developer should never have to import from more than 2 levels deep for connectors and 1 level deep for everything else
+        * i.e.: `from agent_framework.connectors.openai import OpenAIClient` or `from agent_framework.exceptions import ToolCallError`
```
dmytrostruk: Is it going to be `from agent_framework` or `from microsoft.agents`?

docs/design/python-package-setup.md
```diff
@@ -0,0 +1,303 @@
...
+I propose we start using `ty` as it is the fastest and can therefore be used in pre-commit hooks.
+
+#### Task runner
+AG already has experience with poe the poet, so let's start there, removing the MAKE file setup that SK uses.
+
+### Unit test coverage
+The goal is to have at least 80% unit test coverage for all code under both the main package and the subpackages.
+
+### Telemetry and logging
+Telemetry and logging are handled by the `agent_framework.telemetry` and `agent_framework._logging` packages.
```
dmytrostruk, line 273:
```suggestion
Telemetry and logging are handled by the `agent_framework.telemetry` and `agent_framework.logging` packages.
```

docs/design/python-package-setup.md
eavanvalkenburg: no, in Python there is no such thing as a Microsoft namespace that we hook into, so just `agent_framework`

docs/design/python-package-setup.md
eavanvalkenburg: I want to expose the logger directly (`from agent_framework import get_logger` or something similar), hence the code for that will be in a underscore file, while telemetry will be exposed from `agent_framework.telemetry`

docs/design/python-package-setup.md
```diff
@@ -0,0 +1,303 @@
...
+* All other parameters should be supplied as keyword parameters, this is especially important to configure correctly when using Pydantic or dataclasses.
+* If there are multiple required parameters, and they do not have a order that is common sense, then they will all use keyword parameters.
+* If we use `kwargs` we will document how and what we use them for, this might be a reference to a outside package's documentation or an explanation of what the `kwargs` are used for.
+* If we want to combine `kwargs` for multiple things, such as partly for a external client constructor, and partly for our own use, we will try to keep those separate, by adding a parameter, such as `client_kwargs` with type `dict[str, Any]`, and then use that to pass the kwargs to the client constructor (by using `Client(**client_kwargs)`), while using the `**kwargs` parameters for our own use.
+
+
+# Open questions
+
+* Do we need filters? and what about filters vs guardrails?
+* What do we want to do with templates? Or is context providers the new way of making "instructions" dynamic?
```
ekzhu: Why can't we do both? Templates with hooks for dynamic instruction.

docs/design/python-package-setup.md
```diff
@@ -0,0 +1,303 @@
...
+* Do we want to separate other packages out into subpackages, like maybe telemetry, workflows, multi-agent orchestration, etc?
+* what should be included when doing just `pip install agent-framework`, how can we minimize external dependencies?
+    * in other words, which batteries should be included?
+* What versioning scheme do we want to use, SemVer or CalVer?
+* How do we want to generate API docs?
+* Do we want to have a single logger for all AF, or should we have separate loggers for each component?
+* What do we prefer, when the parameters are the same except for one, a Subclass or a Parameter, take:
```
ekzhu: Using types allows better type-checking support and in cases when specific message type is required -- for example, UserMessage may contain image type by AssistantMessage cannot, and external data must be convereted to UserMessage only. 

Genrally, it is going to be case-by-case so I don't think we should make a decision for all cases here.

stephentoub: > for example, UserMessage may contain image type by AssistantMessage cannot, and external data must be convereted to UserMessage only

FWIW, over time we've seen such differences evaporate, eg openai's responses api can now generate images, Gemini with gemini-2.0-flash-exp now supports both text and image output modalities together, etc, with a natural way to model that being an assistant message containing an image.

docs/design/python-package-setup.md
eavanvalkenburg: Clarified that I was thinking about SK templates, and whether they should be ported over, and if so, we have three types of templates currently, so probably makes sense to stick to one since it is less central in building an agent then it was in the past in SK.

docs/design/python-package-setup.md
eavanvalkenburg: Thanks @stephentoub, that is good context, we'll discuss this tomorrrow.

docs/design/python-package-setup.md
ekzhu: Agree on narrowing down to one design of template.

eavanvalkenburg: this was captured here https://github.com/microsoft/agent-framework/issues/121

docs/design/python-package-setup.md
```diff
@@ -0,0 +1,376 @@
...
+We currently can choose between mypy, pyright, ty and pyrefly for static type checking.
+I propose we run `mypy` and `pyright` in GHA, similar to what AG already does. We might explore newer tools as a later date.
+
+#### Task runner
+AG already has experience with poe the poet, so let's start there, removing the MAKE file setup that SK uses.
+
+### Unit test coverage
+The goal is to have at least 80% unit test coverage for all code under both the main package and the subpackages.
+
+### Telemetry and logging
```
TaoChenOSU, line 272: We should probably consider auto instrumentation and provide an implementation of it to the OTel community, just like LangChain and other frameworks: https://github.com/open-telemetry/opentelemetry-python-contrib/pull/3600

docs/design/python-package-setup.md
```diff
@@ -0,0 +1,384 @@
...
+            "agent-framework-google == 1.0.0"
+        ]
+        ```
+    * this means developers can use `pip install agent-framework[google]` to get AF with all Google connectors and dependencies, as well as manually installing the subpackage with `pip install agent-framework-google`.
+
+### Sample getting started code
+```python
+from typing import Annotated
+from agent_framework import Agent, ai_tool
+from agent_framework.connectors.openai import OpenAIChatClient
```
moonbox3: I watched the meeting recording (thank you for recording it). I think we're decided, and to chime in: I'm a fan of the shorter import: `from agent_framework.openai import OpenAIChatClient`. If we go with the longer one, would `agent_framework.client.openai import OpenAIChatClient` be better?

docs/design/python-package-setup.md
```diff
@@ -0,0 +1,384 @@
...
+
+logger = logging.getLogger("agent_framework")
+```
+
+#### Telemetry
+Telemetry will be based on OpenTelemetry (OTel), and will be implemented in the `agent_framework.telemetry` package.
+
+We should consider auto-instrumentation and provide an implementation of it to the OTel community.
+
+### Function definitions
```
moonbox3, line 337: It doesn't feel like this "Function definitions" (design-worthy versus package structure) should live in this doc?

docs/design/python-package-setup.md
```diff
@@ -0,0 +1,384 @@
...
+* Only parameters that are fully expected to be passed and only if there are a very limited number of them, let's say 3 or less, can they be supplied as positional parameters (still with a keyword, _almost_ never positional only).
+* All other parameters should be supplied as keyword parameters, this is especially important to configure correctly when using Pydantic or dataclasses.
+* If there are multiple required parameters, and they do not have a order that is common sense, then they will all use keyword parameters.
+* If we use `kwargs` we will document how and what we use them for, this might be a reference to a outside package's documentation or an explanation of what the `kwargs` are used for.
+* If we want to combine `kwargs` for multiple things, such as partly for a external client constructor, and partly for our own use, we will try to keep those separate, by adding a parameter, such as `client_kwargs` with type `dict[str, Any]`, and then use that to pass the kwargs to the client constructor (by using `Client(**client_kwargs)`), while using the `**kwargs` parameters for other uses, which are then also well documented.
+
+### Attributes vs inheritance
```
moonbox3, line 346: Same with this? We can have a more general "design principles" doc or something similar.

docs/design/python-package-setup.md
```diff
@@ -0,0 +1,384 @@
...
+#### Task runner
+AG already has experience with poe the poet, so let's start there, removing the MAKE file setup that SK uses.
+
+### Unit test coverage
+The goal is to have at least 80% unit test coverage for all code under both the main package and the subpackages.
+
+### Telemetry and logging
+Telemetry and logging are handled by the `agent_framework.telemetry` and `agent_framework._logging` packages.
+
+#### Logging
```
moonbox3, line 275: Why is this needed in the "package design" doc?

docs/design/python-package-setup.md
```diff
@@ -0,0 +1,384 @@
...
+            ...
+    extensions/
+        azure/
+            __init__.py
+            _context_safety.py
+```
+
+## Coding standards
+
+* We use google docstyles for docstrings.
```
moonbox3, line 200: We should be open to AG's experience here - I feel like their docs are much stronger, and appear to be better generated (automatically) than what we've been doing.

docs/design/python-package-setup.md
```diff
@@ -0,0 +1,384 @@
...
+    description="An agent that can get the current weather.",
+)
+response = await agent.run("What is the weather in Amsterdam?")
+print(response)
+```
+
+## Global Package structure
+Overall the following structure is proposed:
+
+* agent-framework
```
moonbox3, line 52: I don't see the "Open Questions" section any more, that had a question about which dependencies should be installed by default. The feedback I continuously receive (and eventually acted on for something like the AzureAIAgent) was to remove the need to do `pip install semantic-kernel[azure]` which installed the dependencies like `azure-ai-agents` or `azure-ai-projects`. The overall feedback was: "This [SK] is a Microsoft-backed SDK, why don't we get Azure dependencies by default? Why do I need to install an extra package like [azure], it's confusing."

eavanvalkenburg: That question has indeed not been resolved, I'm also leaning towards making sure that OpenAI and Azure are installed by default, I moved the open questions to issues, this is the one for this: https://github.com/microsoft/agent-framework/issues/123

docs/design/python-package-setup.md
eavanvalkenburg: AG also uses google docstyle, the bigger question is how to generate API docs sustainably, moved to this issue: https://github.com/microsoft/agent-framework/issues/124

docs/design/python-package-setup.md
eavanvalkenburg: I'll split that into a code guidance doc soon, including some other pieces that I worked on earlier, for now we'll leave it here.

docs/design/python-package-setup.md
eavanvalkenburg: ok, I'm in, updated accordingly

docs/design/python-package-setup.md
moonbox3: Added some notes in #124. Would like to hear AG's perspective on how they're actually generating the docs (is it the same tool we use and have issues with?).

Issue #91, opened 2025-06-23
Agent Tools Support


markwallace-microsoft:
Enable all supported tool types:

- [ ] Create samples demonstrating each tool type (see old repository)
- [ ] Create any missing `AITool` implementations
- [ ] Create proposal for how Foundry specific tool properties are handled

stephentoub:
We have a few options here from least work / worst experience to most work / best experience:
1. Don't add anything. Using Foundry-specific tools for a specific backend can be done via ChatOptions.RawRepresentationFactory, e.g.
```C#
ChatOptions options = new()
{
    RawRepresentationFactory = _ => new ResponseCreationOptions()
    {
        Tools = { ... }, // backend-specific tools
    },
};
```
The upside here is it's already possible and nothing special needs to be done with AITools, no additional types need to be added, etc. The downsides are that there's a separate mechanism required to add in the tools and you need to know which backend you might be talking to (the argument to the callback is an IChatClient, so you can have the callback do different things based on what you're talking to, but you still need to directly code against the backing thing.)

2. Add Foundry-specific AITool-derived types in the Foundry package. Anyone can derive from AITool, so the Foundry package would provide such tools, and the Foundry IChatClient implementation would recognize them. Compared to (2), the consumer no longer needs to use a special mechanism to register the tools, but you still need to know you're using that backend in order to know what tools to use. To some extent in the limit, such knowledge is unavoidable, of course: if you know you want to employ a service-specific tool configured with service-specific configuration, then you're already breaking glass for those specific aspects of the interactions.

3. Create an AITool-derived abstraction in M.E.AI.Abstractions that will map to that Foundry tool. Existing examples include HostedCodeInterpreterTool and HostedWebSearchTool. We should absolutely be looking to add more where there is a corresponding concept in other services. A good example is server-side use of MCP: https://github.com/dotnet/extensions/issues/6492.

Note that a combination of these is also possible: we don't need the same answer for every tool. The most common tools might fall into (3), rare might be (2) (possibly implicitly by having the existing tool types in the foundry package just add AITool as a base type), and (1) is always possible as a fallback.

[...]

PR #110, opened 2025-06-27
.Net: Code interpreter tool abstraction and implementation examples

dmytrostruk:
Related: https://github.com/microsoft/agent-framework/issues/91

This PR contains an example of Code Interpreter tool abstraction and a couple of implementations (Anthropic, Google and Amazon are in progress). 

As for now, it looks simple, but may evolve once we add more implementation examples (e.g. .NET OpenAI Responses SDK doesn't have a support for code interpreter yet, so we even a little bit ahead of time):
```csharp
/// <summary>
/// Proposal for abstraction updates based on the common code interpreter tool properties.
/// Based on the decision, the abstraction can be updated in M.E.AI or specific SDK if some properties are not common.
/// </summary>
public class NewHostedCodeInterpreterTool : HostedCodeInterpreterTool
{
    public IList<string>? FileIds { get; set; }
}
```

In PR there are also implementations, that should be added to related SDK later once the abstraction will be updated on M.E.AI level or the tool type can be added directly to related SDK if some properties are not common.

It's not necessary to merge this code now, since it serves investigation purposes only a lot of logic will be moved to the related SDKs once the updates to abstractions will be released.

More tool types are in progress.

### DIFF
--- a/dotnet/samples/GettingStarted/Tools/CodeInterpreterTools.cs
+++ b/dotnet/samples/GettingStarted/Tools/CodeInterpreterTools.cs
@@ -4,8 +4,8 @@ using System.Text;
 using Azure.AI.Agents.Persistent;
 using GettingStarted.Tools.Abstractions;
 using GettingStarted.Tools.Extensions;
-using Microsoft.Agents;
 using Microsoft.Extensions.AI;
+using Microsoft.Extensions.AI.Agents;
 using OpenAI;
 using OpenAI.Files;
 


dotnet/samples/GettingStarted/Tools/Abstractions/NewHostedCodeInterpreterTool.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace GettingStarted.Tools.Abstractions;
+
+/// <summary>
+/// Proposal for abstraction updates based on the common code interpreter tool properties.
+/// Based on the decision, the abstraction can be updated in M.E.AI or specific SDK if it's not common.
+/// </summary>
+public class NewHostedCodeInterpreterTool : HostedCodeInterpreterTool
+{
+    public IList<string>? FileIds { get; set; }
```
stephentoub, line 13: I assume this whole PR is just for validation purposes and the intent would be that we'd add such a property (and corresponding ctor or factory methods) to HostedCodeInterpreterTool, yes?

stephentoub:
> a couple of implementations (Anthropic, Google and Amazon are in progress).

Those will be more interesting to see, as they're not based on OpenAI (all of the ones in this PR appear to be).

dotnet/samples/GettingStarted/Tools/Abstractions/NewHostedCodeInterpreterTool.cs
dmytrostruk: Correct, because I need to validate if everything is working correctly and that common properties (if such exists) are actually useful for the abstraction.

dmytrostruk:
> Those will be more interesting to see, as they're not based on OpenAI (all of the ones in this PR appear to be).

Yes, I'm working on that and also working on getting API keys to test these integrations locally.

dotnet/samples/GettingStarted/Tools/Abstractions/NewHostedCodeInterpreterTool.cs
dmytrostruk: I think this PR should also help us understand how to design agent output and what common properties are returned from agent run when using tools (web search link, interpreter generated code, etc.) cc: @westey-m

dotnet/samples/GettingStarted/AgentSample.cs
```diff
@@ -19,8 +22,9 @@ public class AgentSample(ITestOutputHelper output) : BaseSample(output)
     /// </summary>
     public enum ChatClientProviders
     {
-        OpenAI,
         AzureOpenAI,
```
rogerbarreto, line 25: For consistency, suggest adding also the Azure AIAssistant

```suggestion
        AzureOpenAIChatCompletion,
        AzureOpenAIAssistant,
```

dotnet/src/Microsoft.Extensions.AI.AzureAIAgentsPersistent/PersistentAgentsChatClient.cs
```diff
@@ -21,19 +21,22 @@ public sealed partial class PersistentAgentsChatClient : IChatClient
     private const string ProviderName = "azure";
 
     /// <summary>The underlying <see cref="PersistentAgentsClient" />.</summary>
-    private readonly PersistentAgentsClient _client;
+    private readonly PersistentAgentsClient? _client;
```
westey-m: Why is client nullable if we are setting it in the constructor?

dmytrostruk: Oh good catch, it's just an update from Persistent SDK PR, since that code also includes protected parameterless constructor for mocking. Going to update it here.

dotnet/samples/GettingStarted/Steps/Step01_ChatClientAgent_Running.cs
```diff
@@ -20,8 +20,8 @@ public sealed class Step01_ChatClientAgent_Running(ITestOutputHelper output) : A
     /// a unique interaction with no conversation history between them.
     /// </summary>
     [Theory]
-    [InlineData(ChatClientProviders.OpenAI)]
```
markwallace-microsoft, line 23: Why are removing OpenAI?

dmytrostruk: It was just renamed to `OpenAIChatCompletion` in line 24, because OpenAI also has Assistants API and Responses API, so I decided to rename this `OpenAI` enum value for more clarity.

PR #111, opened 2025-06-27
Exploring Workflow Languages and APIs

lokitoth:
Exploring various high-level workflow APIs with separation from a lower-level generic API.
See: https://github.com/microsoft/agent-framework/pull/111/files?short_path=05f4b39#diff-05f4b39d5ac942d4bad7dfabd5243de4d5b3fbe26c06aae08a67e2934cd34c3a

Currently:
* Implements a naive low-level graph API (in principle this could instead target AutoGen 0.4, or dapr, or Orleans, or etc.)
* Implements previous add_edge-based GraphBuilder idea (graph_med)
* Implements Algebraic dataflow approach
* Implements initial set of Flow Combinators

TODO:
* Debug initial set of Flow Combinators
* Refactor mid and low-level APIs to provide more information to upper layers when in the middle of building
* Replatform Lowering to target a more realistic "runtime" (async, proper .run() method type)
* Replatform Lowering to source a more realistic generified Workflow (e.g. graph=..., routing=...)

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -6,6 +6,7 @@ version = "0.0.0"
 license = {file = "LICENSE"}
 urls.source = "https://github.com/microsoft/agent-framework/tree/main/python"
 dependencies = [
+    "pytest>=8.4.0",
 ]
 
 [dependency-groups]
--- a/python/src/agent_framework/graph/__init__.py
+++ b/python/src/agent_framework/graph/__init__.py
@@ -1,0 +1,63 @@
+# Copyright (c) Microsoft. All rights reserved.
+from __future__ import annotations  # noqa: I001
+
+from ._graph_shared import (  # this must be imported first
+    EdgeCondition,
+    EpsilonAction,
+    GraphEdge,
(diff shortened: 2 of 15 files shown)


lokitoth:
Added the more fleshed out "Combinators" approach to workflow building (not yet tested/debugged, but should work in principle, based on GraphAlgebra working):

```python
# In a sequence, input some data from the user, then process it, and finally, output the result.
    loop = While(
        lambda ctx: ctx.remaining_retries > 0,
        [
            get_user_request,  # since this is StepT, we could just as easily have it b an agent with .run() method
                               # or a "built" ExecutableGraph instance
            Switch({
                case_(RequestType.PARENS):
                    match_parens,
                case_(RequestType.QUIT):
                    write_quit_response,
            })
        ]
    )

    flow = create_context + loop + extract_output
```

See https://github.com/microsoft/agent-framework/pull/111/files#diff-c4a0454c6e85cb837b485416349678406ef4377f99ebe44927ba286f48718bcaR161-R175

PR #125, opened 2025-07-02
Python: package setup with logger

eavanvalkenburg:
Initial setup of the package in line with the design.

Includes the logger and a base exception.

Closes: #126


python/.vscode/settings.json
```diff
@@ -0,0 +1,46 @@
+{
```
TaoChenOSU, line 1: Should we include this file?

PR #127, opened 2025-07-02
Af demo

westey-m:


dotnet/samples/GettingStarted/AgentTypesDemo.cs
```diff
@@ -0,0 +1,63 @@
...
+            new()
+            {
+                Name = "Joker",
+                Description = "An agent that tells jokes.",
+                Instructions = "You are good at telling jokes.",
+                Model = TestConfiguration.AzureAI.DeploymentName!,
+                Temperature = 0.1f,
+            });
+
+        // We can invoke the agent with foundry specific parameters.
```
stephentoub, line 30: Interesting. I understand creating / configuring it with foundry-specific arguments. But invoking it with foundry-specific arguments... won't most invocations actually be performed indirectly against the abstraction?

dotnet/src/Microsoft.Extensions.AI.AzureAIAgentsPersistent/FoundryAgent.cs
```diff
@@ -0,0 +1,188 @@
...
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>A <see cref="ChatResponse"/> containing the list of <see cref="ChatMessage"/> items.</returns>
+    public async Task<ChatResponse> RunAsync(string message, AgentThread? thread = null, ThreadAndRunOptions? options = null, CancellationToken cancellationToken = default)
+    {
+        if (this._chatClientAgent == null)
+        {
+            throw new InvalidOperationException("The agent has been deleted.");
+        }
+
+        return await this._chatClientAgent.RunAsync(message, thread, chatOptions: new() { RawRepresentationFactory = _ => options }, cancellationToken: cancellationToken).ConfigureAwait(false);
```
stephentoub: RawRepresentationFactory is a delegate because it's expected that each invocation returns a new, mutable instance. Here, though, it's returning the same instance on each call, which means that options instance is likely to be mutated. If it's reused for a subsequent request, that's likely to cause problems.

dotnet/samples/GettingStarted/AgentTypesDemo.cs
westey-m: I would expect that too when users are interested in using this agent amongst a set of heterogeneous agents. That said, I think there is value in being able to use the same instance in an abstract way or to access Foundry specific options without having to switch to a completely different experience (i.e. via PersistentAgentsClient)

dotnet/src/Microsoft.Extensions.AI.AzureAIAgentsPersistent/FoundryAgent.cs
westey-m: Oh, thanks for the pointer.  I've added a shallow clone for the options.  Not sure if we may need to do a deep clone?

Issue #115, opened 2025-06-30
Move Agent abstractions and implementations from the semantigen repo


markwallace-microsoft:
Once the code has been refactored to remove Kernel etc. we need to move the code to the agent-framework repository

- [x] Move only necessary code
- [x] All unit tests are moved and passing
- [x] All integration tests are moved and passing
- [x] All samples are moved and work
- [ ] Organise a bug bash for the moved code

dmytrostruk:
Agent abstractions were moved to Agent Framework repository. Bug bash session will be organized once all implementations will be ported.

PR #130, opened 2025-07-04
.Python: Added Agent and AgentThread abstractions

dmytrostruk:
Related: https://github.com/microsoft/agent-framework/issues/115

In this PR:
- Added `Agent` and `AgentThread` abstractions from semantigen repository.
- Added unit tests.

### DIFF
--- a/python/agent_framework/_agents.py
+++ b/python/agent_framework/_agents.py
@@ -76,7 +76,7 @@ class AgentThread(ABC):
 class Agent(Protocol):
     """A protocol for an agent that can be invoked."""
 
-    async def get_response(
+    async def run(
         self,
         messages: str | ChatMessage | list[str | ChatMessage] | None = None,
         *,
@@ -90,7 +90,7 @@ class Agent(Protocol):
         as a single ChatResponse object. The caller is blocked until
         the final result is available.
 
-        Note: For streaming responses, use the invoke_stream method, which returns
+        Note: For streaming responses, use the run_stream method, which returns
         intermediate steps and the final result as a stream of ChatResponseUpdate
         objects. Streaming only the final result is not feasible because the timing of
         the final result's availability is unknown, and blocking the caller until then
@@ -107,39 +107,7 @@ class Agent(Protocol):
         """
         ...
 
-    async def invoke(
-        self,
-        messages: str | ChatMessage | list[str | ChatMessage] | None = None,
-        *,
-        arguments: dict[str, Any] | None = None,
-        thread: AgentThread | None = None,
-        on_intermediate_message: Callable[[ChatMessage], Awaitable[None]] | None = None,
-        **kwargs: Any,
-    ) -> AsyncIterable[ChatResponse]:
-        """Invoke the agent.
-
-        This invocation method will return the final results of the agent's execution as a
-        stream of ChatResponse objects to the caller. The reason for returning a stream
-        is to allow for future extensions to the agent's capabilities, such as multi-modality.
-
-        To get the intermediate steps of the agent's execution, use the on_intermediate_message callback
-        to handle those messages.
-
-        Note: A ChatResponse object contains an entire message.
-
-        Args:
-            messages: The message(s) to send to the agent.
-            arguments: Additional arguments to pass to the agent.
-            thread: The conversation thread associated with the message(s).
-            on_intermediate_message: A callback function to handle intermediate steps of the agent's execution.
-            kwargs: Additional keyword arguments.
-
-        Yields:
-            An agent response item.
-        """
-        ...
-
-    async def invoke_stream(
+    async def run_stream(
         self,
         messages: str | ChatMessage | list[str | ChatMessage] | None = None,
         *,
@@ -148,9 +116,9 @@ class Agent(Protocol):
         on_intermediate_message: Callable[[ChatMessage], Awaitable[None]] | None = None,
         **kwargs: Any,
     ) -> AsyncIterable[ChatResponseUpdate]:
-        """Invoke the agent as a stream.
+        """Run the agent as a stream.
 
-        This invocation method will return the intermediate steps and final results of the
+        This method will return the intermediate steps and final results of the
         agent's execution as a stream of ChatResponseUpdate objects to the caller.
 
         To get the intermediate steps of the agent's execution as fully formed messages,
--- a/python/tests/unit/test_agents.py
+++ b/python/tests/unit/test_agents.py
@@ -25,7 +25,7 @@ class MockAgentThread(AgentThread):
 
 # Mock Agent implementation for testing
 class MockAgent:
-    async def get_response(
+    async def run(
         self,
         messages: str | ChatMessage | list[str | ChatMessage] | None = None,
         *,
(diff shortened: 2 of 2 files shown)


python/agent_framework/_agents.py
```diff
@@ -0,0 +1,288 @@
...
+    def __init__(self) -> None:
+        """Initialize the agent thread."""
+        self._is_deleted: bool = False  # type: ignore
+        self._id: str | None = None  # type: ignore
+
+    @property
+    def id(self) -> str | None:
+        """Returns the ID of the current thread (if any)."""
+        if self._is_deleted:
+            raise RuntimeError("Thread has been deleted; call `create()` to recreate it.")
```
eavanvalkenburg: these two messages are counter each other, one say, _is_deleted -> call create, but create says if _is_deleted: cannot create....

python/agent_framework/_agents.py
```diff
@@ -0,0 +1,288 @@
...
+# region AgentThread
+
+
+class AgentThread(ABC):
+    """Base class for agent threads."""
+
+    def __init__(self) -> None:
+        """Initialize the agent thread."""
+        self._is_deleted: bool = False  # type: ignore
+        self._id: str | None = None  # type: ignore
```
eavanvalkenburg, line 20: I assume a implementations `create` method needs to add the id?

python/agent_framework/_agents.py
```diff
@@ -0,0 +1,288 @@
...
+    @abstractmethod
+    async def _on_new_message(
+        self,
+        new_message: ChatMessage,
+    ) -> None:
+        """Invoked when a new message has been contributed to the chat by any participant."""
+        raise NotImplementedError
+
+
+# endregion
```
eavanvalkenburg: nit: these are not strictly necessary

python/agent_framework/_agents.py
```diff
@@ -0,0 +1,288 @@
+# Copyright (c) Microsoft. All rights reserved.
+
+import uuid
```
eavanvalkenburg:
```suggestion
from uuid import uuid4
```

python/agent_framework/_agents.py
```diff
@@ -0,0 +1,288 @@
...
+        description: The description of the agent
+        id: The unique identifier of the agent  If no id is provided,
+            a new UUID will be generated.
+        instructions: The instructions for the agent (optional)
+        name: The name of the agent
+    """
+
+    arguments: dict[str, Any] | None = None
+    description: str | None = None
+    id: str = Field(default_factory=lambda: str(uuid.uuid4()))
```
eavanvalkenburg:
```suggestion
    id: str = Field(default_factory=lambda: str(uuid4()))
```

python/agent_framework/_agents.py
```diff
@@ -0,0 +1,288 @@
...
+
+    # endregion
+
+    # region Thread Management
+
+    async def _ensure_thread_exists_with_messages(
+        self,
+        *,
+        messages: str | ChatMessage | Sequence[str | ChatMessage] | None = None,
+        thread: AgentThread | None,
```
eavanvalkenburg: shouldn't this be:
```suggestion
        thread: TThreadType | None,
```

python/agent_framework/_agents.py
```diff
@@ -0,0 +1,288 @@
...
+        raise NotImplementedError
+
+
+# endregion
+
+
+# region Agent
+
+
+class Agent(AFBaseModel, ABC):
```
eavanvalkenburg: I'm thinking about actually splitting this into a Agent protocol (just the methods that are expected for use, and runtime_checkable) and a AgentBase which is this class and that acts as the base class that we use internally that has some implementations for pieces like the `_ensure_thread_exists_with_messages` message.

python/agent_framework/_agents.py
dmytrostruk: In implementations, `create` method should return the id, it can be either server-side thread id or some locally generated id.

python/agent_framework/_agents.py
dmytrostruk: Is there any example available how it should look like?

eavanvalkenburg: the protocol for ChatClient (now in PR, pretty similar to ModelClient) will have a similar way of working, so the protocol side you can look at there, a base class will come prob as part of Peters work

dmytrostruk: Thanks, updated!

python/agent_framework/_agents.py
```diff
@@ -0,0 +1,186 @@
...
+        """Initialize the agent thread."""
+        self._is_deleted: bool = False  # type: ignore
+        self._id: str | None = None  # type: ignore
+
+    @property
+    def id(self) -> str | None:
+        """Returns the ID of the current thread (if any)."""
+        if self._is_deleted:
+            raise RuntimeError(
+                "Thread has been deleted; Create a new AgentThread instance and call `create()` to recreate it."
```
markwallace-microsoft: Why does it matter if the thread has been deleted? 

This agent thread could have been deleted on the server and we wouldn't know but in that case the method would work.

python/agent_framework/_agents.py
```diff
@@ -0,0 +1,186 @@
...
+            )
+        return self._id
+
+    async def create(self) -> str | None:
+        """Starts the thread and returns the thread ID."""
+        # A thread should not be recreated after it has been deleted.
+        if self._is_deleted:
+            raise RuntimeError("Thread has already been deleted. For new thread, create a new AgentThread instance.")
```
markwallace-microsoft: If both errors are needed they should have a consistent error message.

python/agent_framework/_agents.py
```diff
@@ -0,0 +1,186 @@
...
+            arguments: Additional arguments to pass to the agent.
+            thread: The conversation thread associated with the message(s).
+            kwargs: Additional keyword arguments.
+
+        Returns:
+            An agent response item.
+        """
+        ...
+
+    async def invoke(
```
markwallace-microsoft: Should naming be consistent with .Net?

python/agent_framework/_agents.py
dmytrostruk: I didn't change this logic from semantigen repo since I'm not sure how the implementations are working at the moment. As for now, I'm trying to port already existing code to get a working version. But I do agree that this logic is probably not needed. Will remove it in scope of this PR.

python/agent_framework/_agents.py
dmytrostruk: @eavanvalkenburg What do you think about `invoke -> run` and `invoke_stream -> run_streaming` method name updates?

eavanvalkenburg: Let's do it!

python/agent_framework/_agents.py
```diff
@@ -0,0 +1,172 @@
...
+        self,
+        messages: str | ChatMessage | list[str | ChatMessage] | None = None,
+        *,
+        arguments: dict[str, Any] | None = None,
+        thread: AgentThread | None = None,
+        on_intermediate_message: Callable[[ChatMessage], Awaitable[None]] | None = None,
+        **kwargs: Any,
+    ) -> AsyncIterable[ChatResponse]:
```
eavanvalkenburg: This is also being redesigned, so I I'm not sure what we do with this for now, since we will likely redo both run and run_stream and get rid of get_response...

python/agent_framework/_agents.py
dmytrostruk: Renamed to `run` and `run_stream`.

python/agent_framework/_agents.py
dmytrostruk: Removed `get_response` method, made `run` method to return `ChatResponse` instead of `AsyncIterable[ChatResponse]`.

Issue #14, opened 2025-05-28
Embedding Client interface in Python


ekzhu:
Follow the [`IEmbeddingGenerator` interace in MEAI](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.ai.iembeddinggenerator-2?view=net-9.0-pp), modify and adapt for Python language.

Issue #13, opened 2025-05-28
ChatClient interface and base class in Python


ekzhu:
Follow the MEAI `IChatClient` interface -- while making necessary modifications and adaptation for the Python language. 

https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.ai.ichatclient?view=net-9.0-pp

markwallace-microsoft:
@eavanvalkenburg Please review what comes over from the semantigen repo and create tasks for any additional work to align with M.E.AI

eavanvalkenburg:
Need to add ChatClientBase as well, including function calling logic

PR #132, opened 2025-07-07
Python: added embeddingsclient and redid chatclient

eavanvalkenburg:
Added EmbeddingGenerator Protocol, also slightly updated the ModelClient to align with ChatClient interface in MEAI

Closes: #14 
Improves: #13

### DIFF
--- a/python/agent_framework/_types.py
+++ b/python/agent_framework/_types.py
@@ -1401,7 +1401,7 @@ class ChatOptions(AFBaseModel):
 class GeneratedEmbeddings(AFBaseModel, MutableSequence[TEmbedding], Generic[TEmbedding]):
     """A model representing generated embeddings."""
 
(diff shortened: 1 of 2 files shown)


python/agent_framework/_clients.py
```diff
@@ -0,0 +1,82 @@
...
+# region: ChatClient Protocol
+
+
+@runtime_checkable
+class ChatClient(Protocol):
+    """A protocol for a chat client that can generate responses."""
+
+    async def get_response(
+        self,
+        messages: ChatMessage | Sequence[ChatMessage],
```
moonbox3, line 20: This doesn't look accurate to me. This should be able to accept a `str`, a `Sequence[str | ChatMessage]` or `None`. An agent, with an existing thread, doesn't always need a message - the instructions may be enough to generate a response, especially with the thread context. We should also allow for strings to be passed in, positionally, and normalized to ChatMessage on behalf of the caller.

PR #144, opened 2025-07-07
Python: OpenAI Connector

peterychang:
resolves https://github.com/microsoft/agent-framework/issues/37

### DIFF
--- a/python/packages/openai/agent_framework/openai/_chat_completion.py
+++ b/python/packages/openai/agent_framework/openai/_chat_completion.py
@@ -104,7 +104,6 @@ class OpenAIChatCompletionBase(OpenAIHandler, ChatClientBase):
 
     # region content creation
 
-    # TODO(peterychang): Usage?
     def _create_chat_message_content(
         self, response: ChatCompletion, choice: Choice, response_metadata: dict[str, Any]
     ) -> "ChatResponse":


python/extensions/agent-framework-openai/agent_framework/openai/openai_chat_completion.py
```diff
@@ -0,0 +1,98 @@
...
+from .openai_chat_completion_base import OpenAIChatCompletionBase
+from .openai_config_base import OpenAIConfigBase
+from .openai_model_types import OpenAIModelTypes
+from .openai_settings import OpenAISettings
+from agent_framework.exceptions import ServiceInitializationError
+
+logger: logging.Logger = logging.getLogger(__name__)
+
+
+class OpenAIChatCompletion(OpenAIConfigBase, OpenAIChatCompletionBase):
```
eavanvalkenburg: in the new implementation, I would like the pattern to follow the main, so a single lazy loading init in the openai folder, where pretty much everything is exposed, and all other files, in the `_openai_chat_completion.py` pattern. I also think we should combine files that are related, so chat_completion_base should move into a single chat_completion file and we also don't need the openai prefix since this is it's own package!

python/agent_framework/_prompt_execution_settings.py
```diff
@@ -0,0 +1,105 @@
...
+from pydantic import Field
+
+from agent_framework import AFBaseModel
+
+logger = logging.getLogger(__name__)
+
+_T = TypeVar("_T", bound="PromptExecutionSettings")
+
+
+class PromptExecutionSettings(AFBaseModel):
```
eavanvalkenburg: could we start without this one and just say that every connector that get's ported needs to make the change to Options?

python/agent_framework/exceptions.py
```diff
@@ -17,3 +17,48 @@ class AgentExecutionException(AgentException):
...
+    pass
+
+
+class ServiceContentFilterException(ServiceResponseException):
+    """An error was raised by the content filter of the service."""
+
+    pass
+
+
+class ServiceInvalidExecutionSettingsError(ServiceResponseException):
```
eavanvalkenburg: this seems unnecessary, since we will not need Execution settings anymore, in general I think it would be good to reduce the number of specific exceptions, not always needed for clarity.

python/agent_framework/_ai_service_client_base.py
```diff
@@ -0,0 +1,36 @@
...
+    Has an ai_model_id and service_id, any other fields have to be defined by the subclasses.
+
+    The ai_model_id can refer to a specific model, like 'gpt-35-turbo' for OpenAI,
+    or can just be a string that is used to identify the model in the service.
+
+    The service_id is used in Semantic Kernel to identify the service, if empty the ai_model_id is used.
+    """
+
+    ai_model_id: Annotated[str, StringConstraints(strip_whitespace=True, min_length=1)]
+    service_id: str = ""
```
eavanvalkenburg: service_id is no longer needed, so maybe we don't need this whole base class anymore...

python/agent_framework/_ai_service_client_base.py
```diff
@@ -0,0 +1,36 @@
+# Copyright (c) Microsoft. All rights reserved.
+
+from abc import ABC
+from typing import Annotated, Any
+
+from pydantic.types import StringConstraints
+
+from agent_framework import AFBaseModel
```
eavanvalkenburg:
```suggestion
from ._pydantic import AFBaseModel
```

python/agent_framework/_prompt_execution_settings.py
peterychang: I can do that and just leave the code that relies on it commented out for now

python/agent_framework/exceptions.py
peterychang: I'll leave it for now until we figure out what to do with the exceptions. Maybe just reduce it down to the top level ones that were defined in SK?

python/extensions/agent-framework-openai/agent_framework/openai/_openai_handler.py
```diff
@@ -0,0 +1,198 @@
...
+]
+
+
+class OpenAIHandler(AFBaseModel, ABC):
+    """Internal class for calls to OpenAI API's."""
+
+    client: AsyncOpenAI
+    ai_model_id: Annotated[str, StringConstraints(strip_whitespace=True, min_length=1)]
+    ai_model_type: OpenAIModelTypes = OpenAIModelTypes.CHAT
+    prompt_tokens: int = 0
```
eavanvalkenburg: I think by now the usage details should just be in the response, let's remove from this class and the associated funcs

python/agent_framework/__init__.py
```diff
@@ -10,6 +11,8 @@
 
 _IMPORTS = {
     "get_logger": "._logging",
+    "AFBaseModel": "._pydantic",
+    "AFBaseSettings": "._pydantic",
```
dmytrostruk: Do we need to add these classes here? Same comment for `__init__.pyi` file.

python/agent_framework/_pydantic.py
```diff
@@ -1,10 +1,69 @@
...
+from pydantic_settings import BaseSettings, SettingsConfigDict
 
 
 class AFBaseModel(BaseModel):
     """Base class for all pydantic models in the Agent Framework."""
 
     model_config = ConfigDict(populate_by_name=True, arbitrary_types_allowed=True, validate_assignment=True)
+
+
+T = TypeVar("T", bound="AFBaseSettings")
```
dmytrostruk: nit: `T` name looks very generic, maybe `TSettings` will work better, especially in the places where this type var is used.

python/extensions/agent-framework-openai/agent_framework/openai/_chat_completion.py
```diff
@@ -0,0 +1,331 @@
...
+# from agent_framework.contents.streaming_text_content import StreamingTextContent
+
+# from agent_framework.utils.telemetry.model_diagnostics.decorators import (
+#     trace_chat_completion,
+#     trace_streaming_chat_completion,
+# )
+
+
+# Implements agent_framework.ChatClient protocol
+class OpenAIChatCompletionBase(OpenAIHandler):
```
dmytrostruk: Since it implements `ChatClient` protocol, I'm wondering if `OpenAIChatClientBase` name will work better. So, we will have `OpenAIChatClientBase` and `OpenAIChatClient` classes. If we do this renaming, we can also update file name to `openai/_chat_client.py`.

In similar way, the class that will implement `EmbeddingGenerator` protocol will have a name `OpenAIEmbeddingGenerator` and so on.

python/extensions/agent-framework-openai/agent_framework/openai/exceptions.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class ContentFilterCodes(Enum):
+    """Content filter codes."""
+
+    RESPONSIBLE_AI_POLICY_VIOLATION = "ResponsibleAIPolicyViolation"
+
+
+@dataclass
+class ContentFilterAIException(ServiceContentFilterException):
```
dmytrostruk: Another naming option:
```suggestion
class OpenAIContentFilterException(ServiceContentFilterException):
```

python/agent_framework/__init__.py
peterychang: `AFBaseModel` is necessary since components of the clients will inherit from the class. `AFBaseSettings` is a helper class to automagically grab environment variables for each model client. Not strictly necessary, but its quite helpful. This is a class thats shared among most(all?) of the connectors, but isn't used outside of that. The problem is that `agent_framework` is the only common place I can put this

python/extensions/agent-framework-openai/agent_framework/openai/_chat_completion.py
peterychang: Eduard has a PR that changes this. Either he or I will have to make some modifications to this class (depending on who wins the race)

python/packages/main/agent_framework/const.py
eavanvalkenburg, line 1: This should be _const.py

python/packages/openai/agent_framework/openai/_chat_completion.py
```diff
@@ -0,0 +1,314 @@
...
+                yield next(
+                    self._create_streaming_chat_message_content(chunk, choice, chunk_metadata)
+                    for choice in chunk.choices
+                )
+
+    # endregion
+
+    # region content creation
+
+    # TODO(peterychang): Usage?
```
eavanvalkenburg: Still relevant? I see usage details going in?

python/packages/main/agent_framework/const.py
peterychang: I figured it would be good to separate out the consts since we might end up with a fair number of them, similar to how exceptions work

python/packages/openai/agent_framework/openai/_openai_handler.py
```diff
@@ -0,0 +1,148 @@
...
+            assert isinstance(options, ChatOptions)  # nosec # noqa: S101
+            return await self._send_completion_request(options, messages)
+        # TODO(evmattso): move other PromptExecutionSettings to a common options class
+        if self.ai_model_type == OpenAIModelTypes.EMBEDDING:
+            raise NotImplementedError("Embedding generation is not yet implemented in OpenAIHandler")
+        if self.ai_model_type == OpenAIModelTypes.TEXT_TO_IMAGE:
+            raise NotImplementedError("Text to image generation is not yet implemented in OpenAIHandler")
+        if self.ai_model_type == OpenAIModelTypes.SPEECH_TO_TEXT:
+            assert isinstance(options, SpeechToTextOptions)  # nosec # noqa: S101
+            return await self._send_audio_to_text_request(options)
```
moonbox3, line 67:
```suggestion
            return await self._send_speech_to_text_request(options)
```

python/packages/openai/agent_framework/openai/_openai_handler.py
```diff
@@ -0,0 +1,148 @@
...
+        if self.ai_model_type == OpenAIModelTypes.TEXT_TO_SPEECH:
+            assert isinstance(options, TextToSpeechOptions)  # nosec # noqa: S101
+            return await self._send_text_to_audio_request(options)
```
moonbox3, line 70:
```suggestion
            return await self._send_text_to_speech_request(options)
```

PR #147, opened 2025-07-08
Python: added ChatClientBase with function calling

eavanvalkenburg:
Create a Base class for internal implementations of ChatClients, it is consistent with ChatClient, but has the function calling logic in it.

This is a first rough implementation without things like filters, once that has been figured out we will need to update.

Closes: #13

### DIFF
--- a/python/packages/main/agent_framework/__init__.pyi
+++ b/python/packages/main/agent_framework/__init__.pyi
@@ -1,7 +1,5 @@
 # Copyright (c) Microsoft. All rights reserved.
 
-import agent_framework.openai as openai
-
 from . import __version__  # type: ignore[attr-defined]
 from ._agents import Agent, AgentThread
 from ._clients import ChatClient, ChatClientBase, EmbeddingGenerator, use_tool_calling
@@ -63,6 +61,5 @@ __all__ = [
     "__version__",
     "ai_function",
     "get_logger",
-    "openai",
     "use_tool_calling",
 ]


python/extensions/agent-framework-openai/tests/test_test.py
dmytrostruk, line 1: Is this file required? If yes, should we update file name for more clarity?

python/src/agent_framework/_clients.py
```diff
@@ -0,0 +1,505 @@
...
+            An async iterable of chat response updates containing the content of the response messages
+            generated by the client.
+
+        Raises:
+            ValueError: If the input message sequence is `None`.
+        """
+        ...
+
+
+class ChatClientBase(AFBaseModel, ABC):
```
dmytrostruk: In M.E.AI, function calling logic is implemented using decorator pattern, basically wrapping another instance of `ChatClient` to enable function calling. This allows to add not only function calling, but other capabilities to the `ChatClient`, like telemetry, caching and so on. I was wondering if you considered to follow the same pattern in Python? If it's doable of course. If not, I'm worried that `ChatClientBase` will contain a lot of logic when we will decide to add telemetry etc.

python/tests/unit/test_clients.py
```diff
@@ -91,3 +143,53 @@ async def test_embedding_generator_generate(embedding_generator: ImplementedEmbe
...
+    assert response.messages[0].role == ChatRole.ASSISTANT
+    assert response.messages[0].text == "test response - Hello"
+
+
+async def test_base_client_get_streaming_response(chat_client_base: MockChatClientBase):
+    async for update in chat_client_base.get_streaming_response(ChatMessage(role="user", text="Hello")):
+        assert update.text == "update - Hello" or update.text == "another update"
+
+
+async def test_base_client_with_function_calling(chat_client_base: MockChatClientBase):
```
dmytrostruk: Should we add test for function calling with streaming?

python/packages/main/tests/unit/test_clients.py
```diff
@@ -0,0 +1,233 @@
...
+        if not self.streaming_responses:
+            yield ChatResponseUpdate(text=f"update - {messages[0].text}", role="assistant")
+            return
+        response = self.streaming_responses.pop(0)
+        for update in response:
+            yield update
+        await asyncio.sleep(0)
+
+
+class ImplementedEmbeddingGenerator:
```
dmytrostruk: nit:
```suggestion
class MockEmbeddingGenerator:
```

python/packages/main/agent_framework/_clients.py
```diff
@@ -0,0 +1,511 @@
...
+        chat_options.tools = None
+        chat_options.tool_choice = ChatToolMode.NONE.mode
+        return
+    chat_options.tools = [
+        (_tool_to_json_schema_spec(t) if isinstance(t, AITool) else t) for t in chat_options.tools or []
+    ]
+    chat_options.tool_choice = chat_tool_mode.mode
+
+
+def tool_call_enabled(
```
dmytrostruk: I'm wondering, how this will look like from user perspective? If we can add some examples/tests, that would be great. 

Another small suggestion: in .NET we have `Use` keyword when we want to use/add multiple decorators to the `ChatClient`, like `UseFunctionInvocation`, `UseLogging` and so on. So, we can have `use_tool_call`, `use_logging` etc in Python. But I think the name depends on how it's actually going to be used by customers.

eavanvalkenburg: it's now a single class level decorator, the tests have example cases. It is opinionated about being applied only to subclasses of ChatClientBase, not the ChatClient protocol

PR #148, opened 2025-07-08
.Net: Update Code Interpreter Sample as a Step.

### LINKED ISSUE #140: .Net: Add Code Interpreter Sample
## Add Code Interpreter

Similar to SK, add a samples for code interpreter that can be reused across different providers (AzureAIAgent / OpenAIAssistant / Bedrock).

rogerbarreto:
## Motivation and Context

- Resolves #140 

This change updates current proposal for Code Interpreter Tooling as a step sample where a similar and simple code is used from the caller perspective.

This update also simplifies the code and attempts to use the same patterns previously used in other existing samples.

### Observation

Having to handle the usage of a the Raw Representation Factory in the samples adds some considerable complexity in the callers hand by having to generate the expected internal representation of the code to be called within a ChatClient.


dmytrostruk:
@rogerbarreto Note: 
The existing examples are not meant to be used by end-customers (I probably had to put it in another place rather than `GettingStarted` project). The purpose of current examples is to verify current tool abstraction, find new commonalities across multiple providers and tools and propose new tool abstractions (or re-use existing ones if they are sufficient) that potentially can be moved to M.E.AI or provider SDKs later).

Classes under `Tools/Abstractions` folder are candidates to be moved to either M.E.AI or specific provider SDK based on its characteristics, here is XML doc from current code:
```csharp
/// <summary>
/// Proposal for abstraction updates based on the common code interpreter tool properties.
/// Based on the decision, the <see cref="HostedCodeInterpreterTool"/> abstraction can be updated in M.E.AI or specific SDK if some properties are not common.
/// </summary>
```

Classes under `Tools/Extensions` folder are candidates to be moved to specific provider SDK to add support for the new tool abstraction when it's publicly available. Here is XML doc from current code:
```csharp
/// <summary>
/// <see cref="ChatOptions"/> conversion for Azure AI Persistent Agent.
/// When abstraction is in place, this logic should go to Azure AI Persistent Agents SDK.
/// </summary>
```

With existing structure, it should be easier to migrate the new code to the related destinations when we are ready. 

Regarding the observation:
> Having to handle the usage of a the Raw Representation Factory in the samples adds some considerable complexity in the callers hand

In all cases, the usage of `RawRepresentationFactory` is required, if we want to be able to use tool types, that are not currently supported by provider SDKs (this is the purpose of current code). But ideally, that code should be located in provider SDKs repositories, so then `RawRepresentationFactory` won't be needed.

In this PR, samples code doesn't use `RawRepresentationFactory` directly, because it is handled by `base.GetChatOptions(provider);` method, but `RawRepresentationFactory` is still used inside. In another words, it won't be possible to copy-paste example methods and run them without copying additional helper methods from base test class. 

That being said, if you want to add tool samples under `Steps` folder, I think it would be better to add samples that can be used with abstractions that are available today, in this case the code will be runnable without additional internal logic required. For these examples, we can use already existing [HostedCodeInterpreterTool](https://github.com/dotnet/extensions/blob/294a33df47d48f39038ebed1c82a8a50e0345e50/src/Libraries/Microsoft.Extensions.AI.Abstractions/HostedCodeInterpreterTool.cs) and [HostedWebSearchTool](https://github.com/dotnet/extensions/blob/294a33df47d48f39038ebed1c82a8a50e0345e50/src/Libraries/Microsoft.Extensions.AI.Abstractions/HostedWebSearchTool.cs) from M.E.AI. Note that these abstractions are already introduced in some providers, but not all of them (see [comment](https://github.com/microsoft/agent-framework/issues/91#issuecomment-3002272157)). Which means that the usage of `RawRepresentationFactory` is still required, if we want to add more agent providers to our samples.

Hope it makes sense!

dotnet/samples/GettingStarted/Steps/Step03_ChatClientAgent_UsingCodeInterpreterTools.cs
```diff
@@ -0,0 +1,130 @@
...
+        var fileId = await UploadFileAsync("Resources/groceries.txt", provider);
+
+        chatOptions!.Tools = [new NewHostedCodeInterpreterTool { FileIds = [fileId] }];
+
+        var agentOptions = new ChatClientAgentOptions
+        {
+            Name = "HelpfulAssistant",
+            Instructions = "You are a helpful assistant.",
+            // Transformation is required until the abstraction will be added to either SDK provider or M.E.AI and
+            // implementations will handle new properties/classes.
```
dmytrostruk: nit: The `ChatOptions` transformation is not present in this code, so the comment can be probably removed.
```suggestion
```

rogerbarreto:
@dmytrostruk 

> The existing examples are not meant to be used by end-customers (I probably had to put it in another place rather than GettingStarted project). The purpose of current examples is to verify current tool abstraction, find new commonalities across multiple providers and tools and propose new tool abstractions (or re-use existing ones if they are sufficient) that potentially can be moved to M.E.AI or provider SDKs later).

Yes, thanks for the added context, this change doubles down on the "_finding new commonalities across multiple providers_". By showing how the base code can work by using the same helper methods, and is valid as a sample following that we have in SK, samples for Code Interpreter in agents getting started steps.

> But ideally, that code should be located in provider SDKs repositories, so then RawRepresentationFactory won't be needed.

Agree on this, will work on moving the RawFactory impl to another folder to make this intention clear (another SDK responsibility).

> That being said, if you want to add tool samples under Steps folder, I think it would be better to add samples that can be used with abstractions that are available today, in this case the code will be runnable without additional internal logic required. For these examples, we can use already existing [HostedCodeInterpreterTool](https://github.com/dotnet/extensions/blob/294a33df47d48f39038ebed1c82a8a50e0345e50/src/Libraries/Microsoft.Extensions.AI.Abstractions/HostedCodeInterpreterTool.cs) and [HostedWebSearchTool](https://github.com/dotnet/extensions/blob/294a33df47d48f39038ebed1c82a8a50e0345e50/src/Libraries/Microsoft.Extensions.AI.Abstractions/HostedWebSearchTool.cs) from M.E.AI. Note that these abstractions are already introduced in some providers, but not all of them (see https://github.com/microsoft/agent-framework/issues/91#issuecomment-3002272157). Which means that the usage of RawRepresentationFactory is still required, if we want to add more agent providers to our samples.

The samples as of now actually already abstract away any specific's from the providers into a rather solid/fixed AF code, in this case this proves that the Agents abstractions also work for Code Interpreter tooling for multiple providers without big changes compared to other samples.

Will investagate later on how we can consider adding the providers mentioned above that already implements the `Hosted___Tool`.

dotnet/samples/GettingStarted/External/MEAI.OpenAI/NewOpenAIAssistantChatClient.cs
```diff
@@ -0,0 +1,644 @@
...
+                            {
+                                var threadInitializationMessage = new ThreadInitializationMessage(OpenAI.Assistants.MessageRole.User, [OpenAI.Assistants.MessageContent.FromText("attachments")]);
+
+                                foreach (var fileId in fileIds)
+                                {
+                                    threadInitializationMessage.Attachments.Add(new(fileId, [codeInterpreterToolDefinition]));
+                                }
+
+                                runOptions.AdditionalMessages.Add(threadInitializationMessage);
+                            }
```
rogerbarreto, line 330: This proposed change is the only modification to the current OpenAIAssistantChatClient implementation within the MEAI.OpenAI package.

It addresses the current limitation where the caller is responsible for implementing custom `RawRepresentation` logic to handle file attachments. With this change, the responsibility is shifted to the ChatClient, which will now handle any file attachments by default when they are included as part of the Code Interpreter tool input.

dotnet/samples/GettingStarted/External/MEAI.OpenAI/NewOpenAIAssistantChatClient.cs
```diff
@@ -0,0 +1,644 @@
+﻿#pragma warning disable IDE0005 // Using directive is unnecessary.
```
stephentoub, line 1: We're going to delete all of this External and duplicated stuff before this repo is made public, right? i.e. all of this is just temporary?

dotnet/samples/GettingStarted/Steps/Step03_ChatClientAgent_UsingCodeInterpreterTools.cs
```diff
@@ -0,0 +1,129 @@
...
+        await foreach (var update in agent.RunStreamingAsync(Prompt, thread))
+        {
+            if (!string.IsNullOrWhiteSpace(update.Text))
+            {
+                assistantOutput.Append(update.Text);
+            }
+
+            if (update.RawRepresentation is not null)
+            {
+                codeInterpreterOutput.Append(GetCodeInterpreterOutput(update.RawRepresentation, provider));
```
stephentoub, line 55: Is there a proposal for what AIContent-derived type should be added to represent the output of tools like code interpreter?

dotnet/samples/GettingStarted/External/MEAI.OpenAI/NewOpenAIAssistantChatClient.cs
rogerbarreto: Correct

dotnet/samples/GettingStarted/Steps/Step03_ChatClientAgent_UsingCodeInterpreterTools.cs
rogerbarreto: Will dig more on this in subsequent PR's with other providers, want to have at least 3 examples.

dotnet/samples/GettingStarted/External/MEAI.Abstractions/HostedFileCodeInterpreterTool.cs
```diff
@@ -0,0 +1,31 @@
...
+
+using Microsoft.Extensions.AI;
+
+namespace OpenAI.Assistants;
+
+/// <summary>
+/// Proposal for abstraction updates based on the common code interpreter tool properties.
+/// Based on the decision, the <see cref="HostedCodeInterpreterTool"/> abstraction can be updated in M.E.AI directly.
+/// </summary>
+public class NewHostedCodeInterpreterTool : HostedCodeInterpreterTool
```
markwallace-microsoft: Class name doesn't match the file name

Maybe an extension method for `HostedCodeInterpreterTool` which allows you to add the fileIds would be a better approach for now

dotnet/samples/GettingStarted/External/MEAI.OpenAI/NewOpenAIAssistantChatClient.cs
```diff
@@ -0,0 +1,644 @@
...
+#pragma warning disable S1067 // Expressions should not be too complex
+#pragma warning disable S1751 // Loops with at most one iteration should be refactored
+#pragma warning disable S3011 // Reflection should not be used to increase accessibility of classes, methods, or fields
+#pragma warning disable S4456 // Parameter validation in yielding methods should be wrapped
+#pragma warning disable S4457 // Parameter validation in "async"/"await" methods should be wrapped
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>Represents an <see cref="IChatClient"/> for an Azure.AI.Agents.Persistent <see cref="AssistantClient"/>.</summary>
+public sealed class NewOpenAIAssistantChatClient : IChatClient
```
markwallace-microsoft, line 42: Why do we need `New` in the name?

dotnet/samples/GettingStarted/Steps/Step03_ChatClientAgent_UsingCodeInterpreterTools.cs
```diff
@@ -0,0 +1,129 @@
...
+    {
+        var fileId = await UploadFileAsync("Resources/groceries.txt", provider);
+
+        var agentOptions = new ChatClientAgentOptions
+        {
+            Name = "HelpfulAssistant",
+            Instructions = "You are a helpful assistant.",
+            ChatOptions = new ChatOptions()
+            {
+                Tools = [new NewHostedCodeInterpreterTool([fileId])]
```
markwallace-microsoft: This pattern would be better as it is something a developer could do in their application if they want to pass additional items not support by the M.E.AI tool.

```
HostedCodeInterpreterTool codeInterpreterTool = new();
codeInterpreterTool.AddFileIds([fileId])
```

dotnet/src/Microsoft.Extensions.AI.Agents.AzureAI/PersistentAgentsChatClient.cs
```diff
@@ -287,6 +288,15 @@ public void Dispose() { }
 
                         case HostedCodeInterpreterTool:
                             toolDefinitions.Add(new CodeInterpreterToolDefinition());
+
+                            // Once available, HostedCodeInterpreterTool.FileIds property will be used instead of the AdditionalProperties.
+                            if (tool.AdditionalProperties.TryGetValue("fileIds", out object? fileIdsObject) && fileIdsObject is IEnumerable<string> fileIds)
```
markwallace-microsoft, line 293: Can we use the raw representation?

We need a PR to update the code that has been merged into the Foundry Agent SDK.

stephentoub: I'd prefer we focus on getting the desired changes into meai and then updating foundry with the new version.

dotnet/samples/GettingStarted/External/MEAI.OpenAI/NewOpenAIAssistantChatClient.cs
rogerbarreto: This is a proposal, the new actually is a way to avoid clashing with the original type in the OpenAI package.

dotnet/samples/GettingStarted/External/MEAI.Abstractions/HostedFileCodeInterpreterTool.cs
rogerbarreto: Currently there's no way to add or change the `AdditionalProperties` in the `HostedCodeInterpreterTool` using extension methods as this property is a `virtual IReadOnlyDictionary`, the only way is thru a specialization.

There's a not plan on modifying the original, an issue was raised about this before and seems there's expected to have specializations on this class instead.

- https://github.com/dotnet/extensions/issues/6570

CC: @stephentoub

stephentoub: > There's a not plan on modifying the original

What do you mean, @rogerbarreto? I've stated multiple times we can and should augment HostedCodeInterpreterTool to accommodate additional needs. I need you to tell me what the right strongly-typed surface area that needs to be added is, such that it'll make sense with all the relevant providers, and then we'll just do it :)

dotnet/samples/GettingStarted/Steps/Step03_ChatClientAgent_UsingCodeInterpreterTools.cs
rogerbarreto: According with stephen's comment on this Issue: 

- https://github.com/dotnet/extensions/issues/6570

The intent for the tools is to be readonly and immutable after creation, given that, I followed the ctor approach to not diverge from this. 

In theory since the proposal is an `IList<string>` this can be achieved using the `codeTool.FileIds.Add(string)`.

stephentoub: > The intent for the tools is to be readonly and immutable after creation

For the base AITool. I'm fine with specific derived tools being mutable.

My expectation is we'd have something like:
```diff
public class HostedCodeInterpreterTool
{
    ...
+    public IList<whatever>? FileIds { get; set; }
}
```
or something like that... assuming FileIds actually makes sense for multiple providers and won't be conflicting/confusing for the rest.

I'd also be ok if we wanted to keep it immutable and instead have it be:
```diff
public class HostedCodeInterpreterTool
{
    ...
+    public IReadOnlyList<whatever>? FileIds { get; init; }
}
```
though in general I shy away from `init`.

rogerbarreto: > I'm fine with specific derived tools being mutable.

Perfect, thanks for added thoughts

dotnet/samples/GettingStarted/External/MEAI.OpenAI/NewOpenAIAssistantChatClient.cs
markwallace-microsoft: How are we tracking this proposal? Who needs to agree to the changes and when will they be made?

rogerbarreto: Added the issue below for tracking.

- https://github.com/microsoft/agent-framework/issues/163

dotnet/samples/GettingStarted/External/MEAI.Abstractions/HostedFileCodeInterpreterTool.cs
rogerbarreto: > What do you mean, @rogerbarreto? I've stated multiple times we can and should augment HostedCodeInterpreterTool

Yes, sure, I'm referring to the issue targeting the `AITool` type (where the `AdditionalProperties` is defined and it's readonly. Due to that it would not be able to be modified from an extension method.

Just to clarify, this doesn't mean we can't leverage the higher level abstractions `Hosted...` have dedicated properties like `FileIds`.

PR #150, opened 2025-07-08
Python: Added ChatClientAgentThread and ChatClientAgent implementations

dmytrostruk:
This PR contains `ChatClientAgentThread` and `ChatClientAgent` implementations with unit tests.

### DIFF
--- a/python/packages/main/tests/unit/test_agents.py
+++ b/python/packages/main/tests/unit/test_agents.py
@@ -1,8 +1,7 @@
 # Copyright (c) Microsoft. All rights reserved.
 
-from collections.abc import AsyncIterable, Sequence
+from collections.abc import AsyncIterable, MutableSequence, Sequence
 from typing import Any
-from collections.abc import MutableSequence
 from uuid import uuid4
 
 from pytest import fixture, raises
--- a/python/packages/main/agent_framework/_clients.py
+++ b/python/packages/main/agent_framework/_clients.py
@@ -256,7 +256,7 @@ class ChatClient(Protocol):
 
     async def get_response(
         self,
-        messages: str | ChatMessage | Sequence[ChatMessage],
+        messages: str | ChatMessage | list[ChatMessage],
         **kwargs: Any,
     ) -> ChatResponse:
         """Sends input and returns the response.
@@ -276,7 +276,7 @@ class ChatClient(Protocol):
 
     def get_streaming_response(
         self,
-        messages: str | ChatMessage | Sequence[ChatMessage],
+        messages: str | ChatMessage | list[ChatMessage],
         **kwargs: Any,
     ) -> AsyncIterable[ChatResponseUpdate]:
         """Sends input messages and streams the response.
@@ -355,7 +355,7 @@ class ChatClientBase(AFBaseModel, ABC):
 
     async def get_response(
         self,
-        messages: str | ChatMessage | Sequence[ChatMessage],
+        messages: str | ChatMessage | list[ChatMessage],
         *,
         model: str | None = None,
         max_tokens: int | None = None,
@@ -430,7 +430,7 @@ class ChatClientBase(AFBaseModel, ABC):
 
     async def get_streaming_response(
         self,
-        messages: str | ChatMessage | Sequence[ChatMessage],
+        messages: str | ChatMessage | list[ChatMessage],
         *,
         model: str | None = None,
         max_tokens: int | None = None,


python/agent_framework/_agents.py
```diff
@@ -146,3 +147,229 @@ def run_stream(
...
+
+        Raises:
+            ValueError: If both id and messages are provided, or if id is empty/whitespace.
+
+        Notes:
+            - If id is set, _id is assigned and _chat_messages remains empty (service-managed).
+            - If messages is set, _chat_messages is populated and _id is None (local).
+            - If neither is provided, creates an empty local thread.
+        """
+        super().__init__()
```
moonbox3: Do each agent's client implementations then provide a concrete thread type to use with that agent? For example, there's special handling required for an OpenAI responses agent thread - one can choose to store the previous response id, which is used to chain responses (simulated a server-side thread id). Or they can disable that and it turns into more of the "chat history" behavior where all messages are sent during a request.

python/agent_framework/_agents.py
```diff
@@ -146,3 +147,229 @@ def run_stream(
...
+    description: str | None = None
+    instructions: str | None = None
+
+    def __init__(self, chat_client: ChatClient, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+        self._chat_client = chat_client
+
+    async def run(
+        self,
+        messages: str | ChatMessage | list[ChatMessage] | None = None,
```
moonbox3:
```suggestion
        messages: str | ChatMessage | list[str | ChatMessage] | None = None,
```

python/agent_framework/_agents.py
```diff
@@ -146,3 +147,229 @@ def run_stream(
...
+            input_messages=messages,
+            construct_thread=lambda: ChatClientAgentThread(),
+            expected_type=ChatClientAgentThread,
+        )
+
+        response_updates: list[ChatResponseUpdate] = []
+
+        async for update in self._chat_client.get_streaming_response(thread_messages):
+            response_updates.append(update)
+            yield update
```
moonbox3: Shouldn't we be only yielding non-AITool call content? AITool-related content gets sent to the caller via a callback. Initial agent design choice in SK was that an invoke method (in this case `run` or `run_stream`) should return the final answer to the caller. Any other interim content is sent via a callback that one can choose to handle, if they want.

python/agent_framework/_agents.py
```diff
@@ -146,3 +147,229 @@ def run_stream(
...
+        # Only notify the thread of new messages if the chatResponse was successful
+        # to avoid inconsistent messages state in the thread.
+        await self._notify_thread_of_new_messages(thread, thread_messages)
+        await self._notify_thread_of_new_messages(thread, response.messages)
+
+        return response
+
+    async def run_stream(
+        self,
+        messages: str | ChatMessage | list[ChatMessage] | None = None,
```
moonbox3:
```suggestion
        messages: str | ChatMessage | list[str | ChatMessage] | None = None,
```

python/agent_framework/_agents.py
dmytrostruk: > Do each agent's client implementations then provide a concrete thread type to use with that agent?

Correct, and each thread type implementation will depend on the specifics of the agent type.

> For example, there's special handling required for an OpenAI responses agent thread - one can choose to store the previous response id, which is used to chain responses (simulated a server-side thread id). Or they can disable that and it turns into more of the "chat history" behavior where all messages are sent during a request.

Yes, this is how current `ChatClientAgentThread` is implemented, it stores either conversation/thread id, which means that all the messages are stored on a server side, or a collection of chat messages/chat history which will be included during a request. And this is applicable to `ChatClientAgent` only, but if there will be an agent type that will have another thread/conversation logic, then it will be possible to implement a concrete thread type with that logic.

python/agent_framework/_agents.py
dmytrostruk: Good question, at this point we expect that all tool invocations will be already performed inside `_chat_client` by default and it will return final results only. But, taking into account that it will be possible to pass any custom `ChatClient` implementation to the `ChatClientAgent`, I think it should be possible to pass the kind of `ChatClient` which will also return any other interim content, so users can decide what to do with it (e.g. call manually, skip etc). I think this approach should provide more flexibility. In other words, whatever inner `ChatClient` is going to return - we will return it back to the user.

There are also tool types that can't be invoked without user intervention, like remote calls with authentication and things like that, so it should be possible to return this data back to the user as well.

python/agent_framework/_agents.py
```diff
@@ -146,3 +147,229 @@ def run_stream(
...
+
+class ChatClientAgentThread(AgentThread):
+    """Chat client agent thread.
+
+    This class manages chat threads either locally (in-memory) or via a service based on initialization.
+    """
+
+    _storage_location: ChatClientAgentThreadType | None = None
+
+    def __init__(self, id: str | None = None, messages: Iterable[ChatMessage] | None = None):
```
eavanvalkenburg:
```suggestion
    def __init__(self, id: str | None = None, messages: Sequence[ChatMessage] | None = None):
```

python/agent_framework/_agents.py
```diff
@@ -146,3 +147,229 @@ def run_stream(
...
+                raise ValueError("ID cannot be empty or whitespace")
+            self.id = id
+            self._storage_location = ChatClientAgentThreadType.CONVERSATION_ID
+        elif messages:
+            self._chat_messages.extend(messages)
+            self._storage_location = ChatClientAgentThreadType.IN_MEMORY_MESSAGES
+
+    async def get_messages(self) -> AsyncIterable[ChatMessage]:
+        """Get all messages in the thread."""
+        for message in self._chat_messages:
```
eavanvalkenburg: this will raise when conversation id is used instead of messages. I think it makes sense to just have the self._chat_messages as `list[ChatMessage] | None` as a class attribute, then we can keep this more predictable

python/agent_framework/_agents.py
```diff
@@ -146,3 +147,229 @@ def run_stream(
...
+    """
+
+    id: str = Field(default_factory=lambda: str(uuid4()))
+    name: str = Field(default="UnnamedAgent")
+    description: str | None = None
+    instructions: str | None = None
+
+    def __init__(self, chat_client: ChatClient, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+        self._chat_client = chat_client
```
eavanvalkenburg: this should be a attribute, and doesn't have to be private, then Pydantic will also validate the type

python/agent_framework/_agents.py
```diff
@@ -146,3 +147,229 @@ def run_stream(
...
+       description: The description of the agent
+       instructions: The instructions for the agent (optional)
+    """
+
+    id: str = Field(default_factory=lambda: str(uuid4()))
+    name: str = Field(default="UnnamedAgent")
+    description: str | None = None
+    instructions: str | None = None
+
+    def __init__(self, chat_client: ChatClient, **kwargs: Any) -> None:
```
eavanvalkenburg: also having `kwargs` undefined is not the best experience.

python/agent_framework/_agents.py
```diff
@@ -146,3 +147,229 @@ def run_stream(
...
+                messages.append(message)
+
+        if input_messages is None:
+            input_messages = []
+
+        if isinstance(input_messages, (str, ChatMessage)):
+            input_messages = [input_messages]
+
+        normalized_messages = [
+            ChatMessage(role=ChatRole.USER, contents=[TextContent(msg)]) if isinstance(msg, str) else msg
```
eavanvalkenburg:
```suggestion
            ChatMessage(role=ChatRole.USER, text=msg]) if isinstance(msg, str) else msg
```

python/agent_framework/_agents.py
```diff
@@ -146,3 +147,229 @@ def run_stream(
...
+            for msg in input_messages
+        ]
+
+        messages.extend(normalized_messages)
+
+        return thread, messages
```
eavanvalkenburg: should input_messages be added to the thread in case it is a remote thread?

python/agent_framework/_agents.py
```diff
@@ -146,3 +147,229 @@ def run_stream(
...
+        super().__init__(**kwargs)
+        self._chat_client = chat_client
+
+    async def run(
+        self,
+        messages: str | ChatMessage | list[str | ChatMessage] | None = None,
+        *,
+        thread: AgentThread | None = None,
+        **kwargs: Any,
+    ) -> ChatResponse:
```
eavanvalkenburg: this will be replaced with the new AgentRunResponse

python/agent_framework/_agents.py
```diff
@@ -146,3 +147,229 @@ def run_stream(
...
+    def __init__(self, chat_client: ChatClient, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+        self._chat_client = chat_client
+
+    async def run(
+        self,
+        messages: str | ChatMessage | list[str | ChatMessage] | None = None,
+        *,
+        thread: AgentThread | None = None,
+        **kwargs: Any,
```
eavanvalkenburg: I think we need somethign with the new AgentRunOptions (or whatever it's called here...)

python/agent_framework/_agents.py
```diff
@@ -146,3 +147,229 @@ def run_stream(
...
+        # Only notify the thread of new messages if the chatResponse was successful
+        # to avoid inconsistent messages state in the thread.
+        await self._notify_thread_of_new_messages(thread, thread_messages)
+        await self._notify_thread_of_new_messages(thread, response.messages)
+
+    def get_new_thread(self) -> AgentThread:
+        return ChatClientAgentThread()
+
+    def _update_thread_with_type_and_conversation_id(
+        self, chatClientThread: ChatClientAgentThread, responseConversationId: str | None
```
eavanvalkenburg: dotnet shining through :D
```suggestion
        self, chat_client_thread: ChatClientAgentThread, response_conversation_id: str | None
```

python/agent_framework/_agents.py
eavanvalkenburg: also when this is called before a first run, this will throw, and since it is a public method people might do that!

python/agent_framework/_agents.py
eavanvalkenburg: the decision was made to also return "secondary content" on the streaming api and do away with the callback, see https://github.com/microsoft/agent-framework/pull/104

python/agent_framework/_agents.py
```diff
@@ -15,49 +22,43 @@ class AgentThread(AFBaseModel):
...
 
-    @abstractmethod
-    async def _on_new_message(
-        self,
-        new_messages: ChatMessage | Sequence[ChatMessage],
+
+# region AgentBase
+
+
+class AgentBase(AFBaseModel):
```
eavanvalkenburg: please move this after the Agent Protocol, and I think it makes sense to have the `id`, `name` and `description` fields in this class, similar to how the ChatClientBase has all the pieces to complete the protocol, with one or more abstractmethods that need to be implemented.

python/agent_framework/_agents.py
dmytrostruk: Yes, and it is done in `run` and `run_stream` methods after actual request and only if it's successful, to avoid inconsistent messages state in the thread.

python/agent_framework/_agents.py
dmytrostruk: Yes, just need to merge another PR first.

python/agent_framework/_agents.py
dmytrostruk: In .NET version it's currently an empty class, created for reservation purposes only, just so we don't break the API later. Is it going to be a problem to add this in Python later in case we will need such type?

python/agent_framework/_agents.py
dmytrostruk: Oops 😄

python/packages/main/agent_framework/_agents.py
```diff
@@ -146,3 +113,241 @@ def run_stream(
...
+            ValueError: If both id and messages are provided, or if id is empty/whitespace.
+
+        Notes:
+            - If id is set, _id is assigned and _chat_messages remains empty (service-managed).
+            - If messages is set, _chat_messages is populated and _id is None (local).
+            - If neither is provided, creates an empty local thread.
+        """
+        super().__init__()
+
+        if id and messages:
```
eavanvalkenburg, line 203: I would expect this check to happen before the init call, and I think there is a slight performance benefit when passing in the arguments to the init rather then doing it afterwards like below

python/packages/main/agent_framework/_agents.py
```diff
@@ -146,3 +113,241 @@ def run_stream(
...
+
+    chat_client: ChatClient
+
+    async def run(
+        self,
+        messages: str | ChatMessage | list[str | ChatMessage] | None = None,
+        *,
+        thread: AgentThread | None = None,
+        **kwargs: Any,
+    ) -> ChatResponse:
```
eavanvalkenburg: Shouldn't this be AgentRunResponse?

python/packages/main/tests/unit/test_agents.py
```diff
@@ -1,39 +1,54 @@
...
+    ChatClient,
+    ChatClientAgent,
+    ChatClientAgentThread,
+    ChatMessage,
+    ChatResponse,
+    ChatResponseUpdate,
+    ChatRole,
+    TextContent,
+)
+from agent_framework._agents import ChatClientAgentThreadType
```
eavanvalkenburg: This should be without_agents

dmytrostruk: At least for now, the purpose of `ChatClientAgentThreadType` is to be used internally to differentiate between server-side thread and local messages, but the details are hidden from users, and they only operate with `ChatclientAgentThread`.

python/packages/main/agent_framework/_agents.py
```diff
@@ -67,24 +54,9 @@ async def _on_new_message(
 class Agent(Protocol):
     """A protocol for an agent that can be invoked."""
 
-    @property
-    def id(self) -> str:
```
eavanvalkenburg: Why where these removed?

python/packages/main/agent_framework/_agents.py
```diff
@@ -138,3 +110,273 @@ def run_stream(
...
+
+
+class ChatClientAgentThread(AgentThread):
+    """Chat client agent thread.
+
+    This class manages chat threads either locally (in-memory) or via a service based on initialization.
+    """
+
+    chat_messages: list[ChatMessage] | None = None
+    _storage_location: ChatClientAgentThreadType | None = None
```
eavanvalkenburg: Why can't this be public?

python/packages/main/agent_framework/_agents.py
dmytrostruk: Reverted

python/packages/main/agent_framework/_agents.py
dmytrostruk: Converted to public as per internal discussion.

Issue #92, opened 2025-06-23
Review design of Agent API surface to improve developer UX


markwallace-microsoft:
We want the developer UX to be as good as or better than the main competitors e.g. Agno, OpenAI ADK, Google ADK, ... Currently developers need to write a lot of unintuitive code to create an Agent and we want to simplify this.

Goal is to create a POC with some options and demonstrate how this would work for different types of agents i.e. Foundry, ChatClient, OpenAI Responses. 

Specifically address:
- [x] Name, Description and Instructions on the agent base class
- [x] Naming of Agent class
- [x] Add name with fallback to id property
- [x] Determine if Name and Id are both required on Agent.  Check what agents have ID/Name.

westey-m:
## Agent naming

For reference existing SDKs use the following names for base classes:

| SDK | Name |
|-|-|
| M.SK.A | Agent |
| Autogen | [ChatAgent](https://microsoft.github.io/autogen/stable/reference/python/autogen_agentchat.base.html#autogen_agentchat.base.ChatAgent) |
| Google ADK | [BaseAgent](https://google.github.io/adk-docs/api-reference/java/com/google/adk/agents/BaseAgent.html) |
| AWS (Strands) | [Agent](https://strandsagents.com/latest/api-reference/agent/#strands.agent.agent) |
| LangGraph | No equivalent, since agents are [graphs](https://langchain-ai.github.io/langgraph/reference/agents/#langgraph.prebuilt.chat_agent_executor.create_react_agent) |
| Agno | [Agent](https://docs.agno.com/reference/agents/agent) |

westey-m:
## Service Agent Id/Name support

| Service | HasId | HasName | Notes |
|-|-|-|-|
| Foundry Persistent | Y | Y ||
| OpenAI Assistant | Y | Y ||
| CopilotStudio | N | N | Has SchemaName which is used to identify the agent in communication which could map to either |
| A2A | N | Y | Name is on card, URL is potentially a unique identifier |
| Bedrock Agent | Y | Y |  |

PR #154, opened 2025-07-09
Remove instructions from base agent type, and ensure description is only required where necessary.

westey-m:
Not all agent types require Instructions on the agent class to function, since some are proxies for service agents that have their instructions configured server side.  It may not even be possible to retrieve instructions in all cases.
The agent may also have been built to perform a specific task, and therefore instructions are not applicable, since it cannot be configured with instructions.  It might just do one thing.

Orchestrations were all throwing when agents didn't have a description, but it should only really throw if the description is absolutely required.

- Removing the Instructions property from the base agent type and adding it to ChatClientAgent instead.
- Ensuring that description is only required where necessary.

#92

### DIFF
--- a/dotnet/src/Microsoft.Agents.Orchestration/Handoff/Handoffs.cs
+++ b/dotnet/src/Microsoft.Agents.Orchestration/Handoff/Handoffs.cs
@@ -85,10 +85,12 @@ public static class OrchestrationHandoffsExtensions
 
         foreach (Agent target in targets)
         {
-            agentHandoffs[target.Name ?? target.Id] =
-                !string.IsNullOrWhiteSpace(target.Description) ?
-                target.Description! :
-                throw new InvalidOperationException($"The provided target agent with Id '{target.Id}' and name '{target.Name}' has no description, and no handoff description has been provided. At least one of these are required to register a handoff.");
+            if (string.IsNullOrWhiteSpace(target.Description) && string.IsNullOrWhiteSpace(target.Name))
+            {
+                throw new InvalidOperationException($"The provided target agent with Id '{target.Id}' has no description or name, and no handoff description has been provided. At least one of these are required to register a handoff so that the appropriate target agent can be chosen.");
+            }
+
+            agentHandoffs[target.Name ?? target.Id] = target.Description ?? target.Name!;
         }
 
         return handoffs;
--- a/dotnet/tests/Microsoft.Agents.Orchestration.UnitTests/HandoffsTests.cs
+++ b/dotnet/tests/Microsoft.Agents.Orchestration.UnitTests/HandoffsTests.cs
@@ -201,7 +201,7 @@ public class HandoffsTests
     }
 
     [Fact]
-    public void AddWithAgentWithNoDescriptionThrows()
+    public void AddWithAgentWithNoDescriptionUsesName()
     {
         // Arrange
         OrchestrationHandoffs handoffs = new("source");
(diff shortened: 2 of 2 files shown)


dotnet/src/Microsoft.Agents.Orchestration/Handoff/Handoffs.cs
```diff
@@ -84,7 +85,10 @@ public static OrchestrationHandoffs Add(this OrchestrationHandoffs handoffs, Age
 
         foreach (Agent target in targets)
         {
-            agentHandoffs[target.Name ?? target.Id] = target.Description ?? string.Empty;
+            agentHandoffs[target.Name ?? target.Id] =
+                !string.IsNullOrWhiteSpace(target.Description) ?
+                target.Description! :
+                throw new InvalidOperationException($"The provided target agent with Id '{target.Id}' and name '{target.Name}' has no description, and no handoff description has been provided. At least one of these are required to register a handoff.");
```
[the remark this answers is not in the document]
westey-m: This should show the name or null if it's not there, so won'tfix.

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Abstractions/AgentMetadata.cs
```diff
@@ -7,7 +7,7 @@ namespace Microsoft.Extensions.AI.Agents.Runtime;
 /// <summary>
 /// Represents metadata associated with an agent, including its type, unique key, and description.
 /// </summary>
-public readonly struct AgentMetadata(string type, string key, string description) : IEquatable<AgentMetadata>
+public readonly struct AgentMetadata(string type, string key, string? description) : IEquatable<AgentMetadata>
```
rogerbarreto: nit
```suggestion
public readonly struct AgentMetadata(string type, string key, string? description = null) : IEquatable<AgentMetadata>
```

dotnet/tests/Microsoft.Agents.Orchestration.UnitTests/HandoffsTests.cs
```diff
@@ -201,23 +201,19 @@ public void AddWithAgentsWithNoNameUsesId()
...
-        handoffs.Add(sourceAgent, targetAgent);
+        InvalidOperationException ex = Assert.Throws<InvalidOperationException>(() => handoffs.Add(sourceAgent, targetAgent1));
 
         // Assert
-        Assert.Single(handoffs);
-        Assert.Equal("source", handoffs.FirstAgentName);
-        AgentHandoffs sourceHandoffs = handoffs["source"];
-        Assert.Single(sourceHandoffs);
-        Assert.Equal(string.Empty, sourceHandoffs["target"]);
+        Assert.Equal("The provided target agent with Id 'target1' and name 'target 1' has no description, and no handoff description has been provided. At least one of these are required to register a handoff.", ex.Message);
```
rogerbarreto: Not sure if we should be strict to the full error description.

dotnet/src/Microsoft.Agents.Orchestration/AgentActor.cs
```diff
@@ -30,7 +29,7 @@ protected AgentActor(AgentId id, IAgentRuntime runtime, OrchestrationContext con
             id,
             runtime,
             context,
-            VerifyDescription(agent),
+            agent.Description,
```
rogerbarreto, line 32: Having `AgentId` type and `agent.Id` can be quite confusing here. If that's still applicable what happens when Agent.Id is provided , how do `AgentId.Key` equals to the `Agent.Id`, should that be verified ? 

Should internally a `AgentId` be generated from the `Agent.Id` instead?

```diff
-    protected AgentActor(AgentId id, IAgentRuntime runtime, OrchestrationContext context, Agent agent, ILogger? logger = null)
-        : base(
-            id,
+    protected AgentActor(IAgentRuntime runtime, OrchestrationContext context, Agent agent, ILogger? logger = null)
+        : base(
+            new AgentId(Throw.IfNull(Agent.Id)),

westey-m: AgentId here doesn't match exactly to an agent id and is part of the Agent Runtime, the design of which is being reviewed separately right now.  Consider giving this feedback as part of that design work instead, since it wouldn't make sense for me to start making design changes for that in this unrelated PR.

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Abstractions/AgentMetadata.cs
westey-m: Updated.

dotnet/tests/Microsoft.Agents.Orchestration.UnitTests/HandoffsTests.cs
westey-m: I think we should.  I wouldn't expect this to change very often, so the stricter the better.

dotnet/src/Microsoft.Agents.Orchestration/Handoff/Handoffs.cs
stephentoub: Could we infer a description, albeit a bad one, from other things, like Name?

westey-m: Yeah, name might work, but it'll depend on the naming users use.  Seems reasonable to support it.  Added.

PR #157, opened 2025-07-09
Python: Added AgentRunResponse and AgentRunResponseUpdate types

dmytrostruk:
This PR introduces two new classes, `AgentRunResponse` and `AgentRunResponseUpdate`, to enhance the handling of agent responses and updates in the framework.

### DIFF
--- a/python/packages/main/tests/unit/test_types.py
+++ b/python/packages/main/tests/unit/test_types.py
@@ -1,7 +1,6 @@
 # Copyright (c) Microsoft. All rights reserved.
 
 from collections.abc import MutableSequence
-from typing import cast
 
 from pydantic import BaseModel, ValidationError
 from pytest import fixture, mark, raises
@@ -550,12 +549,6 @@ def test_agent_run_response_str_method(chat_message: ChatMessage) -> None:
 
 
 # region AgentRunResponseUpdate
-
-
-def test_agent_run_response_update_init_string_content() -> None:
-    update = AgentRunResponseUpdate(contents="Test")
-    assert len(update.contents) == 1
-    assert cast(TextContent, update.contents[0]).text == "Test"
 
 
 def test_agent_run_response_update_init_content_list(text_content: TextContent) -> None:


python/agent_framework/_types.py
```diff
@@ -1495,3 +1497,156 @@ def __iadd__(self, values: Iterable[TEmbedding] | Self) -> Self:
...
+
+class AgentRunResponse(AFBaseModel):
+    """Represents the response to an Agent run request.
+
+    Provides one or more response messages and metadata about the response.
+    A typical response will contain a single message, but may contain multiple
+    messages in scenarios involving function calls, RAG retrievals, or complex logic.
+    """
+
+    messages: list[ChatMessage] = Field(default_factory=lambda: [])
```
eavanvalkenburg:
```suggestion
    messages: list[ChatMessage] = Field(default_factory=list)
```

python/agent_framework/_types.py
```diff
@@ -1495,3 +1497,156 @@ def __iadd__(self, values: Iterable[TEmbedding] | Self) -> Self:
...
+        if messages is not None:
+            if isinstance(messages, ChatMessage):
+                self.messages.append(messages)
+            elif isinstance(messages, list):
+                self.messages.extend(messages)
+
+    @property
+    def text(self) -> str:
+        """Get the concatenated text of all messages."""
+        return "".join(msg.text for msg in self.messages) if self.messages else ""
```
eavanvalkenburg: would raise when there is not `.text` present, and we also don't want to include TextReasoningContent
```suggestion
        return "".join(msg.text for msg in self.messages if isinstance(msg, TextContent) if self.messages else ""
```

python/agent_framework/_types.py
```diff
@@ -1495,3 +1497,156 @@ def __iadd__(self, values: Iterable[TEmbedding] | Self) -> Self:
...
+
+# region AgentRunResponseUpdate
+
+
+class AgentRunResponseUpdate(AFBaseModel):
+    """Represents a single streaming response chunk from an Agent."""
+
+    author_name: str | None = None
+    role: ChatRole | None = None
+    contents: list[AIContents] = Field(default_factory=lambda: [])
```
eavanvalkenburg:
```suggestion
    contents: list[AIContents] = Field(default_factory=list)
```

python/agent_framework/_types.py
```diff
@@ -1495,3 +1497,156 @@ def __iadd__(self, values: Iterable[TEmbedding] | Self) -> Self:
...
+            elif isinstance(contents, list):
+                self.contents.extend(contents)
+
+        self.role = role
+
+    @field_validator("author_name", mode="before")
+    @classmethod
+    def clean_author_name(cls, value: str | None) -> str | None:
+        """Ensure author_name is None if empty or whitespace."""
+        return None if value is None or value.strip() == "" else value
```
eavanvalkenburg: why do we need to do this? we could also do `author_name: Annotated[str | None, StringConstraints(strip_whitespace=True)]` in this case, only a emptry string is kept as a string, but I don't know if that is an issue.

eavanvalkenburg: and we should prob keep this consistent between this and the ChatResponse class

python/agent_framework/_types.py
```diff
@@ -1495,3 +1497,156 @@ def __iadd__(self, values: Iterable[TEmbedding] | Self) -> Self:
...
+            "".join(content.text for content in self.contents if isinstance(content, TextContent))
+            if self.contents
+            else ""
+        )
+
+    def __str__(self) -> str:
+        return self.text
+
+
+def _process_agent_run_update(response: "AgentRunResponse", update: "AgentRunResponseUpdate") -> None:
```
eavanvalkenburg: could we generalize this one, for AgentRunResponse and ChatResponse?

python/agent_framework/_types.py
```diff
@@ -1495,3 +1497,156 @@ def __iadd__(self, values: Iterable[TEmbedding] | Self) -> Self:
...
+        response.response_id = update.response_id
+    if update.created_at is not None:
+        response.created_at = update.created_at
+    if update.additional_properties is not None:
+        if response.additional_properties is None:
+            response.additional_properties = {}
+        response.additional_properties.update(update.additional_properties)
+
+
+def _finalize_agent_run_response(response: "AgentRunResponse") -> None:
```
eavanvalkenburg: this is exactly the same for both, so just change:
```suggestion
def _finalize_response(response: "AgentRunResponse | ChatResponse") -> None:
```

python/agent_framework/_types.py
dmytrostruk: It was my initial version but IDE doesn't like that:
image

python/agent_framework/_types.py
dmytrostruk: We are iterating here over `self.messages` which is `list[ChatMessage]`, so `msg` variable is never going to be an instance of `TextContent`. But here we are accessing `msg.text`, where `text` property comes from `ChatMessage` and this property already does the filtering for `TextContent`:
https://github.com/microsoft/agent-framework/blob/fef4fd2c18862538a46f028eb8d8fd43caa1fc9d/python/packages/main/agent_framework/_types.py#L885

python/agent_framework/_types.py
dmytrostruk: Removed it.

python/packages/main/agent_framework/_types.py
```diff
@@ -1552,3 +1557,107 @@ def __iadd__(self, values: Iterable[TEmbedding] | Self) -> Self:
         else:
             self.embeddings += values
         return self
+
+
+# region AgentRunResponse
+
+
+class AgentRunResponse(AFBaseModel):
```
peterychang, line 1565: Would it make sense to have `ChatResponse` and `AgentRunResponse` inherit from the same base type? There seems to be a lot of overlap between the two. Same for the update versions

dmytrostruk: From code reusability perspective - yes, base type would definitely make things simpler. But from user perspective - not sure, `ChatResponse` should be used together with `ChatClient`, while `AgentRunResponse` should be used together with `Agent`, and both of them should be decoupled from each other. `ChatClientAgent` is just an adapter here. 

As for now, I'm not sure what would be the case when user will want to abstract from response type and use base class, but we can always add more abstraction if needed.

python/agent_framework/_types.py
eavanvalkenburg: yeah, I saw this as well, pyright is unnecessarily strict with those, add `# type: ignore[reportUnknownVariableType]`

python/packages/main/agent_framework/_types.py
eavanvalkenburg: in the design, one of the arguments for adding this AgentRunResponse is that we can independently evolve this without worrying about ChatResponse, so good to start with a separate model, because otherwise we or users might make assumptions with `isinstance(item, ResponseBase)` or similar.

python/packages/main/agent_framework/_types.py
```diff
@@ -1552,3 +1557,107 @@ def __iadd__(self, values: Iterable[TEmbedding] | Self) -> Self:
...
+    """
+
+    messages: list[ChatMessage] = Field(default_factory=list)
+    response_id: str | None = None
+    created_at: CreatedAtT | None = None  # use a datetimeoffset type?
+    usage_details: UsageDetails | None = None
+    raw_representation: Any | None = None
+    additional_properties: dict[str, Any] | None = None
+
+    def __init__(self, messages: ChatMessage | list[ChatMessage] | None = None, **kwargs: Any) -> None:
```
eavanvalkenburg: if we do a custom init, make sure that the expected fields are in the init, this helps in discoverability, now people would have to go into the code to understand which other fields are present, and preferably the messages update is done before the init and then passed to `super().__init__(**args)` with the other args.

python/packages/main/agent_framework/_types.py
```diff
@@ -1552,3 +1557,107 @@ def __iadd__(self, values: Iterable[TEmbedding] | Self) -> Self:
...
+    role: ChatRole | None = None
+    contents: list[AIContents] = Field(default_factory=list)
+    raw_representation: Any | None = None
+    additional_properties: dict[str, Any] | None = None
+    response_id: str | None = None
+    message_id: str | None = None
+    created_at: CreatedAtT | None = None  # use a datetimeoffset type?
+
+    def __init__(
+        self, role: ChatRole | None = None, contents: str | list[AIContents] | None = None, **kwargs: Any
```
eavanvalkenburg: same as for the AgentRunResponse

PR #164, opened 2025-07-11
Round 2 of cleanup for agent runtime and orchestrations

stephentoub:
- Moved InProcessRuntime type into abstractions package and deleted InProcess package.
- Moved several members of IAgentRuntime to be extension methods instead, e.g. multiple GetActorAsync overloads.
- Added synchronous RegisterMessageHandler overloads and used them to avoid unnecessary async usage at call sites.
- Removed unnecessary surface area from InProcessRuntime, e.g. StopAsync, RunUntilIdleAsync, etc.
- Fixed spin loop in InProcessRuntime that would consume an entire core for the duration of the orchestration's operation.
- Removed a bunch of allocation from InProcessRuntime.
- Made a runtime optional for orchestrations, defaulting to using a temporary InProcessRuntime if none is provided.
- Removed custom delegate types from orchestrations.
- Consolidated namespaces.
- Used records to simplify message classes.
- Tweaked naming on AgentActor to make purpose of protected methods more clear.
- Removed invocation in AgentActor.InvokeAsync of empty update / isFinal parameter.
- Changed OrchestrationHandoffs to avoid needing to pass in agents duplicatively.
- Made various extension methods, such as those on OrchestrationHandoffsExtensions, into instance methods.

cc: @crickman, @reubenbond, @westey-m, @TaoChenOSU


dotnet/src/Microsoft.Agents.Orchestration/AgentOrchestration.cs
```diff
@@ -129,17 +126,18 @@ public async ValueTask<OrchestrationResult<TOutput>> InvokeAsync(
 
         TaskCompletionSource<TOutput> completion = new();
 
-        ActorType orchestrationType = await this.RegisterAsync(runtime, context, completion, handoff: null).ConfigureAwait(false);
+        InProcessRuntime? temporaryRuntime = null;
+        runtime ??= temporaryRuntime = InProcessRuntime.StartNew();
```
westey-m, line 131: Why do we need the temporaryRuntime variable?

stephentoub: To know whether to dispose of it.

PR #165, opened 2025-07-11
.Net: Introduce Dependency Injection Samples

### LINKED ISSUE #136: .Net: Add Dependency Injection Sample
## Add Dependency Injection Sample

Similar to SK, add a dependency injection sample to the repo.

rogerbarreto:
## Motivation Context

- Resolves #136 

This change add Dependency Injection use-case for the samples

As part of this update the `GetChatClientAsync` helpers proven to not be DI friendly (no support for `Singleton<T>(Func<Task<IServiceCollection, T>>()`).

After further investigation, the `Async` was there to create asynchronously the Agent in the server-side "hidding" this responsibility from the caller which is not ideal as the caller needs to handle the Agent Creation and CleanUp in the current AF API design.

Given the above the samples were updated also to make clear to the caller when applicable (for server-side agent providers) the need to create agents ahead of time prior to invocation.

image

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,44 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  # Maintain dependencies for nuget
+  - package-ecosystem: "nuget"
+    directory: "dotnet/"
+    schedule:
+      interval: "cron"
+      cronjob: "0 8 * * 4,0" # Every Thursday(4) and Sunday(0) at 8:00 UTC
+    ignore:
+      # For all System.* and Microsoft.Extensions/Bcl.* packages, ignore all major version updates
+      - dependency-name: "System.*"
+        update-types: ["version-update:semver-major"]
+      - dependency-name: "Microsoft.Extensions.*"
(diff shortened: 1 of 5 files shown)


dotnet/samples/GettingStarted/GettingStarted.csproj
```diff
@@ -17,7 +17,9 @@
   <ItemGroup>
     <PackageReference Include="Azure.AI.OpenAI" />
     <PackageReference Include="Azure.Identity" />
-    <PackageReference Include="Microsoft.Extensions.AI.OpenAI" />
+    <PackageReference Include="OpenAI" VersionOverride="2.2.0-beta.4" />
+    <PackageReference Include="Microsoft.Extensions.AI.OpenAI" VersionOverride="9.6.0-preview.1.25310.2"  />
```
markwallace-microsoft, line 21: Why do we need the version overrides?

rogerbarreto: `Azure.AI.OpenAI` is still tied to the the pre-release `OpenAI.beta-4` (we do have a new 2.2.0 GA release from OpenAI), attempting to use the new OpenAI with `AzureOpenAI` breaks the AzureOpenAI samples in runtime.

stephentoub: I'm still trying to get them to release a new Azure.AI.OpenAI.

But, I'm not clear on why we need version overrides. What's pulling these packages forward? When I updated the versions in the main versions file yesterday, I explicitly left these on the cited version numbers rather than bringing them forward.

PR #166, opened 2025-07-11
Set CopilotStudio integration tests to manual run only.

westey-m:


dotnet/tests/CopilotStudio.IntegrationTests/CopilotStudioRunStreamingTests.cs
```diff
@@ -7,9 +7,36 @@ namespace CopilotStudio.IntegrationTests;
...
+    public override Task RunWithNoMessageDoesNotFailAsync()
+    {
+        return base.RunWithNoMessageDoesNotFailAsync();
+    }
+
+    [Fact(Skip = ManualVerification)]
+    public override Task RunWithStringReturnsExpectedResultAsync()
+    {
+        return base.RunWithStringReturnsExpectedResultAsync();
+    }
```
rogerbarreto, line 41: nit

```suggestion
    public override Task RunWithChatMessageReturnsExpectedResultAsync()
        => base.RunWithChatMessageReturnsExpectedResultAsync();

    [Fact(Skip = ManualVerification)]
    public override Task RunWithChatMessagesReturnsExpectedResultAsync()
        => base.RunWithChatMessagesReturnsExpectedResultAsync();

    [Fact(Skip = ManualVerification)]
    public override Task RunWithNoMessageDoesNotFailAsync()
        => base.RunWithNoMessageDoesNotFailAsync();

    [Fact(Skip = ManualVerification)]
    public override Task RunWithStringReturnsExpectedResultAsync()
        => base.RunWithStringReturnsExpectedResultAsync();
```

dotnet/tests/CopilotStudio.IntegrationTests/CopilotStudioRunStreamingTests.cs
```diff
@@ -7,9 +7,36 @@ namespace CopilotStudio.IntegrationTests;
 
 public class CopilotStudioRunStreamingTests() : RunStreamingTests<CopilotStudioFixture>(() => new())
```
rogerbarreto, line 8: I wonder if the whole class is skiped we might try something like this?

```suggestion
public class CopilotStudioRunStreamingTests : RunStreamingTests<CopilotStudioFixture>
{
    public CopilotStudioRunStreamingTests ()
        : base(() => new())
    {
        throw SkipException.ForSkip(ManualVerification)
    }
```

dotnet/tests/CopilotStudio.IntegrationTests/CopilotStudioRunStreamingTests.cs
```diff
@@ -7,9 +7,36 @@ namespace CopilotStudio.IntegrationTests;
...
 {
+    // Set to null to run the tests.
+    private const string ManualVerification = "For manual verification";
+
     [Fact(Skip = "Copilot Studio does not support thread history retrieval, so this test is not applicable.")]
```
rogerbarreto, line 13: Not supported, I would rather remove the `[Fact]` then skipping.

```suggestion
    // Copilot Studio does not support thread history retrieval, so this test is not applicable.
```

dotnet/tests/CopilotStudio.IntegrationTests/CopilotStudioRunStreamingTests.cs
westey-m: I'm not a big fan of this, since I constantly see people having to switch between these different styles when the code changes just a little bit.  E.g. you need to add another line of code, then you have to switch back to a block, but when the line is removed it's back to inline.

dotnet/tests/CopilotStudio.IntegrationTests/CopilotStudioRunStreamingTests.cs
westey-m: We'd need to upgrade to V3 of xunit first (probably a good thing to do anyway, but we should add a work item for it).
With v2 that we are on it still causes a test failure.

dotnet/tests/CopilotStudio.IntegrationTests/CopilotStudioRunStreamingTests.cs
westey-m: Fact is still present on the base method, and therefore the test still get's executed.  The main difference is how it shows up in the test results.  Without Fact(Skip), where we have to modify the internals to do nothing, it just returns passed.
I think skipped is more appropriate though, since we are actually skipping it.  My preference would of course be that it doesn't show up in the test explorer at all, but removing Fact doesn't do that.

PR #175, opened 2025-07-11
Basic agent invocation sample

westey-m:
### Motivation and Context

Since we have many different agents with various requirements, our samples aren't always the simplest to get started with, so adding a sample that deals with inference based agents only, and shows clearly how to get the text result from the response.

#92

### Description

Adding a basic agent invocation sample.

### Contribution Checklist


dotnet/samples/GettingStarted/Steps/Step01_ChatClientAgent_Running.cs
```diff
@@ -15,6 +15,34 @@ public sealed class Step01_ChatClientAgent_Running(ITestOutputHelper output) : A
...
+
+        // Define the agent
+        var agent = new ChatClientAgent(chatClient, options: new()
+        {
+            Name = ParrotName,
+            Instructions = ParrotInstructions,
+        });
+
+        // Invoke the agent and output the text result.
+        var response = await agent.RunAsync("Fortune favors the bold.");
```
stephentoub: Do we override ToString like ChatResponse does? If so, this could just be:
```C#
Console.WriteLine(await agent.RunAsync("Fortune favors the bold."));
```

dotnet/samples/GettingStarted/Steps/Step01_ChatClientAgent_Running.cs
```diff
@@ -15,6 +15,34 @@ public sealed class Step01_ChatClientAgent_Running(ITestOutputHelper output) : A
...
+    [InlineData(ChatClientProviders.OpenAIChatCompletion)]
+    [InlineData(ChatClientProviders.OpenAIResponses)]
+    public async Task RunBasic(ChatClientProviders provider)
+    {
+        // Get the chat client to communicate with the inference service backing our agent.
+        // Any implementation of Microsoft.Extensions.AI.Agents.IChatClient can be used with the ChatClientAgent.
+        // See the Providers folder for examples on how to create chat clients for some sample providers.
+        using var chatClient = base.GetChatClient(provider);
+
+        // Define the agent
```
stephentoub, line 35: Instead of `var`, maybe use `Agent` to highlight the rest of the sample works regardless of the concrete type of agent being used?

dotnet/samples/GettingStarted/Steps/Step01_ChatClientAgent_Running.cs
westey-m: Great suggestion, I'll update.

dotnet/samples/GettingStarted/Steps/Step01_ChatClientAgent_Running.cs
westey-m: Agreed, I'll also update chatClient to be IChatClient to make that clearer too.

dotnet/samples/GettingStarted/Steps/Step01_ChatClientAgent_Running.cs
```diff
@@ -15,6 +15,34 @@ public sealed class Step01_ChatClientAgent_Running(ITestOutputHelper output) : A
...
+        var agent = new ChatClientAgent(chatClient, options: new()
```
stephentoub: While IChatClient inherits from IDisposable, it's rare it actually does anything, and even if it did, for the simplest example purposes, you could get away without calling it and just let finalization clean up as needed. If you remove the need for the using, then this could be `new ChatClientAgent(GetChatClient(provider), new() { ...})`

westey-m: @stephentoub Should we perhaps make ChatClientAgent (and Agent I suppose) disposable, so that it takes care of disposing the chat client?

dotnet/samples/GettingStarted/AgentSample.cs
```diff
@@ -31,13 +31,13 @@ public enum ChatClientProviders
         AzureAIAgentsPersistent
     }
 
-    protected IChatClient GetChatClient(ChatClientProviders provider, ChatClientAgentOptions options)
+    protected IChatClient GetChatClient(ChatClientProviders provider, ChatClientAgentOptions? options = null)
         => provider switch
         {
             ChatClientProviders.OpenAIChatCompletion => GetOpenAIChatClient(),
-            ChatClientProviders.OpenAIAssistant => GetOpenAIAssistantChatClient(options),
+            ChatClientProviders.OpenAIAssistant => GetOpenAIAssistantChatClient(options ?? throw new InvalidOperationException("Options are required for this provider.")),
```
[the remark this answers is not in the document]
westey-m: Fixed.

dotnet/samples/GettingStarted/Steps/Step01_ChatClientAgent_Running.cs
stephentoub: We probably should.

westey-m: OK, I'll create a separate PR for that and as part of that PR return to this sample to clean it up further.

rogerbarreto: I think `ChatClientAgent` should be disposable as well, mostly for scenarios where you don't provide an external `ChatClient`, we don't have yet (AFAIK), but this paves the way for that.

westey-m: If I remove `using` I'd need some ugly suppressions to avoid a Code Analysis warning, which I fear is worse than the using clause.  It'll probably call more attention to it than it deserves.

stephentoub: This analyzer rule is problematic in general, with lots of false positives. I'd have no qualms about disabling it for the whole repo, or just for samples. It's not on by default; it's getting turned on by https://github.com/microsoft/agent-framework/blob/407ed6de70ec46b59d8fadcef8f9a90b716f59ac/dotnet/Directory.Build.props#L6
If we want to disable it, we can do so in https://github.com/microsoft/agent-framework/blob/407ed6de70ec46b59d8fadcef8f9a90b716f59ac/dotnet/.editorconfig#L155

westey-m: OK, I've suppressed it for samples.  Probably better to keep it for the main code base, so we don't miss any important disposal cases, although I do agree that there's a lot of false positives.

PR #179, opened 2025-07-12
Allow InProcessRuntime subscriptions to be processed concurrently

stephentoub:


stephentoub:
How important is it that we throw an AggregateException with all the exceptions? I neglected to do so, such that right now tests are failing because only the first exception is propagating. I can fix it to be an aggregate if that's important...? Otherwise I'll fix the tests.

PR #181, opened 2025-07-14
Make Agent disposable.

westey-m:
### Motivation and Context

Agents may own disposable objects, e.g. ChatClientAgent owns an IChatClient instance, which is disposable.

#92

### Description

- Making Agent Disposable
- Disposing the IChatClient in ChatClientAgent
- Updating tests and samples to match

### Contribution Checklist

### DIFF
--- a/dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatClientAgent.cs
+++ b/dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatClientAgent.cs
@@ -172,7 +172,11 @@ public sealed class ChatClientAgent : Agent
             return;
         }
 
-        this.ChatClient.Dispose();
+        if (disposing)
+        {
+            this.ChatClient.Dispose();
+        }
+
         base.Dispose(disposing);
 
         this._disposed = true;


dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/Agent.cs
```diff
@@ -187,6 +187,22 @@ public abstract IAsyncEnumerable<AgentRunResponseUpdate> RunStreamingAsync(
...
+
+    /// <summary>
+    /// Provides a mechanism for releasing unmanaged resources.
+    /// </summary>
+    /// <param name="disposing"><see langword="true"/> if being called from <see cref="Dispose()"/>; otherwise, <see langword="false"/>.</param>
+    protected virtual void Dispose(bool disposing)
+    {
+        // Nothing to dispose in the base class, but allowing dispose for derived classes.
+    }
+
```
rogerbarreto, line 205: Can you just enforce the specializations to do the disposing implementation instead?

If `abstraction` takes this responsibility by default, this can induce derived to ignore this.

```suggestion
    /// <inheritdoc/>
    public abstract void Dispose();
```

dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatClientAgent.cs
```diff
@@ -163,6 +164,20 @@ public override async IAsyncEnumerable<AgentRunResponseUpdate> RunStreamingAsync
...
 
+    /// <inheritdoc/>
+    protected override void Dispose(bool disposing)
+    {
+        if (this._disposed)
+        {
+            return;
+        }
+
+        this.ChatClient.Dispose();
```
rogerbarreto: We should only dispose `IChatClient`s there were instantiated within the `ChatClietnAgent` creation, not the externally provided ones. Currently we only have scenarios for external, `IDispose` should not be disposing those.

Unless we provide a `bool disposeChatClient` to the ctor. Similar to [`HttpClient(HttpHandler, disposeHandler)`.](https://source.dot.net/#Microsoft.Extensions.Http/DefaultHttpClientFactory.cs,108)



```suggestion
```

dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatClientAgent.cs
```diff
@@ -163,6 +164,20 @@ public override async IAsyncEnumerable<AgentRunResponseUpdate> RunStreamingAsync
     /// <inheritdoc/>
     public override AgentThread GetNewThread() => new ChatClientAgentThread();
 
+    /// <inheritdoc/>
+    protected override void Dispose(bool disposing)
```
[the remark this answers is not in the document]
westey-m: Fixed.

dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatClientAgent.cs
westey-m: Don't we want the ChatClientAgent to own ChatClients passed in?  I would have expected that to be the case.
This is not an uncommon pattern, e.g. StreamReader closing the underlying Stream that was passed to its constructor.
We can of course make this configurable too.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/Agent.cs
westey-m: Derived classes only need to implement this if they have something to dispose, and in that case they would get a warning from code analysis saying that they have a field that should be disposed.  No need to force classes that have nothing to dispose to implement this.

dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatClientAgent.cs
rogerbarreto: I don't think this is a good practice, you may have same `IChatClient` used across different agents.

For cases where you don't want to handle the IChatClient yourself I would expect providers to have their own special `ProviderChatClientAgent` handling and diposing the ChatClient internally.

I may say that `IChatClient` differs very much from `Stream` scenarios where a stream in theory should be Read only once and this may be ok

Any thoughts @stephentoub, @SteveSandersonMS?

SergeyMenshykh:
I wonder if agents need to be disposable. Do we currently have or plan to have agents that own chat clients (an agent constructor creates an instance of a chat client)? If so, then agents need to be disposable. Otherwise, they don't necessarily need to be. There may be cases where the same chat client instance is used by more than one agent, and disposing it through one agent that no longer needed could break the other agent(s) using the same client.

stephentoub:
I think it probably comes down to how we expect agents to be used. Do we expect folks to publish IChatClient instances in DI (possibly / likely as singletons) and then have ChatClientAgents be transient? If so, ChatClientAgents shouldn't own the IChatClient (or we need to add some middleware that injects a delegating IChatClient that makes dispose a nop). Or do we expect folks to publish Agents in DI, in which case they probably should own the IChatClient, though if they're being published as long-lived, the ownership is likely less relevant as you won't have many of them being created.

We made IChatClient implement IDisposable because of the pipelining model, and for middleware to have any chance of disposing what they wrap, they would need to themselves be disposable. Thus far we haven't created a pipeline model for Agent. Assuming we stick with that, we could say that it's up to a derived type whether it's IDisposable or not, and consumers can query an arbitrary Agent for I{Async}Disposable to know whether to dispose it. Then for ChatClientAgent, we can start with it not owning the IChatClient, and if it becomes a problem, either switch it, or make it configurable.

westey-m:
Closing this PR for now, since there isn't a quorum on supporting `IDisposable` at the moment.  We can revisit later.

PR #182, opened 2025-07-14
.Net: Add OpenTelemetry Support and Samples

### LINKED ISSUE #137: .Net: Add Telemetry Sample
## Add Telemetry

Similar to SK, add a telemetry sample to the repo.

rogerbarreto:
### Motivation and Context

- Resolves #137 

OpenTelemetry Support and Proposal, using similar wrapper approach as `Microsoft.Extensions.AI.OpenTelemetryChatClient`.

1. OpenTelemetry Constants ( AgentOpenTelemetryConsts.cs)
    - Comprehensive constants following Microsoft.Extensions.AI patterns
    - Agent-specific telemetry attribute names and metric definitions
    - Proper XML documentation for public API
2. OpenTelemetryAgent Wrapper ( OpenTelemetryAgent.cs)
    - Delegating agent that wraps any  Agent with telemetry instrumentation
    - Follows the exact pattern of Microsoft.Extensions.AI's OpenTelemetryChatClient
    - Captures both activities/spans and metrics
3. Extension Methods ( AgentExtensions.cs)
    - Easy-to-use .WithOpenTelemetry() extension method
    - Consistent with Microsoft.Extensions.AI patterns
4. Updated Step05 Telemetry Sample
    - Uses the new OpenTelemetryAgent wrapper
    - Demonstrates proper TracerProvider configuration
    - Shows agent-level telemetry collection

Telemetry Data Captured:

Metrics:

- Operation duration histogram with proper buckets
- Token usage histogram (input/output tokens)
- Request count counter
- All metrics tagged with operation type and agent name

Integration:

The implementation integrates seamlessly with:

- AppContext.SetSwitch("Microsoft.Extensions.AI.Agents.EnableTelemetry", true) pattern
- OpenTelemetry SDK's CreateTracerProviderBuilder()
- Console exporter and any other OpenTelemetry exporters
- Existing Agent Framework abstractions without breaking changes

Test Results:

- The Step05 telemetry sample successfully demonstrates:

- Rich telemetry data collection at the agent level
- Proper OpenTelemetry semantic conventions
- No duplication of underlying ChatClient telemetry
- Performance metrics and usage tracking
- Error handling and status reporting
- The telemetry output shows detailed information like:

```
Activity.TraceId:            9d3a4454fbab21fdeb476b56d2fcdbf0
Activity.SpanId:             e3cccd9a75815155
Activity.TraceFlags:         Recorded
Activity.ActivitySourceName: Microsoft.Extensions.AI.Agents
Activity.DisplayName:        agent.run TelemetryAgent
Activity.Kind:               Client
Activity.StartTime:          2025-07-14T15:05:42.6188603Z
Activity.Duration:           00:00:03.8557817
Activity.Tags:
    agent.operation.name: agent.run
    agent.request.id: 5ff1bbdb-8507-4fcb-ba0b-ed180bd396c6
    agent.request.name: TelemetryAgent
    agent.request.message_count: 1
    agent.request.instructions: You are a helpful assistant.
    agent.response.message_count: 1
    agent.response.id: chatcmpl-BtF2G6Lz8uSKLDHsdRXuTzBQOhL80
    agent.usage.input_tokens: 25
    agent.usage.output_tokens: 331
Resource associated with Activity:
    telemetry.sdk.name: opentelemetry
    telemetry.sdk.language: dotnet
    telemetry.sdk.version: 1.9.0
    service.name: unknown_service:testhost
```

### DIFF
--- a/dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatClientAgent.cs
+++ b/dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatClientAgent.cs
@@ -104,7 +104,7 @@ public sealed class ChatClientAgent : Agent
         (ChatClientAgentThread chatClientThread, ChatOptions? chatOptions, List<ChatMessage> threadMessages) =
             await this.PrepareThreadAndMessagesAsync(thread, messages, options, cancellationToken).ConfigureAwait(false);
 
-        var agentName = this.GetAgentName();
+        var agentName = this.GetLoggingAgentName();
 
         this._logger.LogAgentChatClientInvokingAgent(nameof(RunAsync), this.Id, agentName, this._chatClientType);
 
@@ -388,6 +388,6 @@ public sealed class ChatClientAgent : Agent
         }
     }
 
-    private string GetAgentName() => this.Name ?? "UnnamedAgent";
+    private string GetLoggingAgentName() => this.Name ?? "UnnamedAgent";
     #endregion
 }


dotnet/src/Microsoft.Extensions.AI.Agents/OpenTelemetryAgent.cs
```diff
@@ -0,0 +1,311 @@
...
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Extensions.AI.Agents;
+
+/// <summary>
+/// Represents a delegating agent that implements OpenTelemetry instrumentation for agent operations.
+/// </summary>
+/// <remarks>
+/// This class provides telemetry instrumentation for agent operations including activities, metrics, and logging.
+/// The telemetry output follows OpenTelemetry semantic conventions and is subject to change as the conventions evolve.
```
stephentoub: Which convention is this implementing?

dotnet/src/Microsoft.Extensions.AI.Agents/OpenTelemetryAgent.cs
```diff
@@ -0,0 +1,311 @@
...
+/// </remarks>
+public sealed class OpenTelemetryAgent : Agent, IDisposable
```
stephentoub, line 21: If we're doing the delegation pattern, we should have a DelegatingAgent : Agent that OpenTelemetryAgent then derives from, only overriding the things it needs to specialize and allowing the base to delegate to the inner for everything else.

stephentoub, line 21: This delegation is exactly what I said in https://github.com/microsoft/agent-framework/pull/181#issuecomment-3069550171 that we weren't yet doing.

dotnet/src/Microsoft.Extensions.AI.Agents/OpenTelemetryAgent.cs
rogerbarreto: Actually I don't think there's a convention for Agent telemetry yet, will leave this as a proposal for now.

dmytrostruk: > Actually I don't think there's a convention for Agent telemetry yet

There are some properties and operation names like `invoke_agent` and `create_agent`:
https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/#gen-ai-agent-id

dotnet/src/Microsoft.Extensions.AI.Agents/OpenTelemetryAgent.cs
rogerbarreto: I will investigate this pattern in a future PR. 

Created the tracking task below:

- https://github.com/microsoft/agent-framework/issues/183

dotnet/src/Microsoft.Extensions.AI.Agents/OpenTelemetryAgent.cs
stephentoub: There's https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-agent-spans/.

I'm a little nervous this is going to end up fully duplicating OpenTelemetryChatClient, just at a different level.

rogerbarreto: I missed the agent convention, good to know, will update with the provided info.

rogerbarreto: > I'm a little nervous this is going to end up fully duplicating OpenTelemetryChatClient, just at a different level.

I'm not fully duplicating the ChatClient, this initially is focused on `Agent` level telemetry, and don't track internal messaging from the ChatClient perspective, almost all the attributes are being prefixed in the `gen_ai.agent.*` namespace, as the provided convention started to point as a place for agent stuff.

dotnet/src/Microsoft.Extensions.AI.Agents/AgentOpenTelemetryConsts.cs
```diff
@@ -0,0 +1,238 @@
...
+        {
+            /// <summary>
+            /// Invoke GenAI agent operation.
+            /// </summary>
+            public const string InvokeAgent = "invoke_agent";
+
+            /// <summary>
+            /// Create GenAI agent operation.
+            /// </summary>
+            public const string CreateAgent = "create_agent";
```
westey-m: Is this being used?

dotnet/src/Microsoft.Extensions.AI.Agents/OpenTelemetryAgent.cs
```diff
@@ -0,0 +1,324 @@
...
+
+            if (activity is not null)
+            {
+                _ = activity
+                    // Required attributes per OpenTelemetry semantic conventions
+                    .AddTag(AgentOpenTelemetryConsts.GenAI.OperationName, operationName)
+                    .AddTag(AgentOpenTelemetryConsts.GenAI.System, genAISystem ?? AgentOpenTelemetryConsts.GenAI.Systems.MicrosoftExtensionsAI)
+                    // Agent-specific attributes
+                    .AddTag(AgentOpenTelemetryConsts.GenAI.Agent.Id, this.Id)
+                    .AddTag(AgentOpenTelemetryConsts.GenAI.Agent.Name, this.Name)
```
westey-m: Agent name is optional, so this.Name could be null, similar to Description below.

dotnet/src/Microsoft.Extensions.AI.Agents/OpenTelemetryAgent.cs
```diff
@@ -0,0 +1,324 @@
...
+                if (!string.IsNullOrWhiteSpace(thread?.Id))
+                {
+                    _ = activity.AddTag(AgentOpenTelemetryConsts.GenAI.ConversationId, thread.Id);
+                }
+
+                // Add instructions if available (for ChatClientAgent)
+                if (!string.IsNullOrWhiteSpace(chatClientAgent?.Instructions))
+                {
+                    _ = activity.AddTag(AgentOpenTelemetryConsts.GenAI.Agent.Request.Instructions, chatClientAgent.Instructions);
+                }
```
westey-m, line 222: Shouldn't this be checking this.Instructions?  instructions may be valid for agents other than ChatClientAgent as well.

dotnet/samples/GettingStarted/Steps/Step05_ChatClientAgent_Telemetry.cs
```diff
@@ -0,0 +1,55 @@
...
+/// <summary>
+/// Demonstrates how to use telemetry with <see cref="ChatClientAgent"/> using OpenTelemetry.
+/// </summary>
+public sealed class Step05_ChatClientAgent_Telemetry(ITestOutputHelper output) : AgentSample(output)
+{
+    /// <summary>
+    /// Demonstrates OpenTelemetry tracing with Agent Framework.
+    /// </summary>
+    [Theory]
+    [InlineData(ChatClientProviders.AzureOpenAI)]
```
westey-m, line 19: Also consider adding foundry persistent and openai assistant providers.  Would be nice to show that it works with those too.

dotnet/src/Microsoft.Extensions.AI.Agents/OpenTelemetryAgent.cs
```diff
@@ -0,0 +1,324 @@
...
+    private readonly Histogram<int> _tokenUsageHistogram;
+    private readonly Counter<int> _requestCounter;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="OpenTelemetryAgent"/> class.
+    /// </summary>
+    /// <param name="innerAgent">The underlying agent to wrap with telemetry.</param>
+    /// <param name="logger">An optional logger for emitting events.</param>
+    /// <param name="sourceName">An optional source name that will be used on the telemetry data.</param>
+    public OpenTelemetryAgent(Agent innerAgent, ILogger? logger = null, string? sourceName = null)
```
westey-m: I'm not seeing logger being used anywhere.  Are we missing some textual logging?

dotnet/src/Microsoft.Extensions.AI.Agents/OpenTelemetryAgent.cs
rogerbarreto: `Agent` base class doesn't have instructions property.

dotnet/src/Microsoft.Extensions.AI.Agents/AgentOpenTelemetryConsts.cs
rogerbarreto: Is not, removed.

dotnet/src/Microsoft.Extensions.AI.Agents/OpenTelemetryAgent.cs
rogerbarreto: Good point added in the beginning but wasn't necessary, as this is a OTEL wrapper. I saw that the OTEL wrapper for MEAI does use the ILogger, 

CC: @stephentoub, any context for adding the logging in the OTEL if you have the LoggingChatClient already?

docs/decisions/0002-agent-opentelemetry-instrumentation.md
```diff
@@ -0,0 +1,151 @@
...
+# Agent OpenTelemetry Instrumentation
+
+## Context and Problem Statement
+
+Currently, the Agent Framework lacks comprehensive observability and telemetry capabilities, making it difficult for developers to monitor agent performance, track usage patterns, debug issues, and gain insights into agent behavior in production environments. While the underlying ChatClient implementations may have their own telemetry, there is no standardized way to capture agent-specific metrics and traces that provide visibility into agent operations, token usage, response times, and error patterns at the agent abstraction level.
+
+## Decision Drivers
+
+- **Observability Requirements**: Developers need comprehensive telemetry to monitor agent performance, track usage patterns, and debug issues in production environments.
+- **Standardization**: The solution must follow established OpenTelemetry semantic conventions and integrate seamlessly with existing .NET telemetry infrastructure.
```
dmytrostruk, line 19: nit: It would be good to include a link to OpenTelemetry semantic conventions for agents as a reference.

dotnet/src/Microsoft.Extensions.AI.Agents/AgentExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+/// </summary>
+public static class AgentExtensions
+{
+    /// <summary>
+    /// Wraps the agent with OpenTelemetry instrumentation.
+    /// </summary>
+    /// <param name="agent">The agent to wrap.</param>
+    /// <param name="sourceName">An optional source name that will be used on the telemetry data.</param>
+    /// <returns>An <see cref="OpenTelemetryAgent"/> that wraps the original agent with telemetry.</returns>
+    public static OpenTelemetryAgent WithOpenTelemetry(this Agent agent, string? sourceName = null)
```
dmytrostruk, line 16: A couple of things regarding MEAI patterns: 
- For each client interface, there is related `Builder` class which allows not only to apply any custom or existing decorator but also include service provider so that each `IChatClient` instance can get required services to operate: https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/ChatClientBuilder.cs
- In MEAI, such methods named as `UseLogging`, `UseOpenTelemetry`, with `Use` keyword instead of `With`. But it is also associated with `Builder` logic, so if we want to introduce `AgentBuilder` in the future, in the same way as `ChatClientBuilder`, I think it makes sense to use the same naming conventions here.

dotnet/src/Microsoft.Extensions.AI.Agents/AgentOpenTelemetryConsts.cs
```diff
@@ -0,0 +1,233 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+namespace Microsoft.Extensions.AI.Agents;
+
+/// <summary>
+/// Provides constants used by agent telemetry services.
+/// </summary>
+internal static class AgentOpenTelemetryConsts
```
dmytrostruk, line 9: It would be great to provide a reference link to OpenTelemetry naming conventions in XML doc.

dotnet/src/Microsoft.Extensions.AI.Agents/OpenTelemetryAgent.cs
```diff
@@ -0,0 +1,319 @@
...
+
+            if (activity is not null)
+            {
+                _ = activity
+                    // Required attributes per OpenTelemetry semantic conventions
+                    .AddTag(AgentOpenTelemetryConsts.GenAI.OperationName, operationName)
+                    .AddTag(AgentOpenTelemetryConsts.GenAI.System, genAISystem ?? AgentOpenTelemetryConsts.GenAI.Systems.MicrosoftExtensionsAI)
+                    // Agent-specific attributes
+                    .AddTag(AgentOpenTelemetryConsts.GenAI.Agent.Id, this.Id)
+                    .AddTag(AgentOpenTelemetryConsts.GenAI.Agent.Name, !string.IsNullOrWhiteSpace(this.Name) ? this.Name : "UnnamedAgent")
```
dmytrostruk: `UnnamedAgent` is also used as default value in `ChatClientAgent` name:
https://github.com/microsoft/agent-framework/blob/2e6f6a4008fabb5c907ec9dfb781c1dfe5b89ac0/dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatClientAgent.cs#L365

I'm not saying that we should use the same constant for `OpenTelemetryAgent` and `ChatClientAgent` name, since these 2 classes should be unrelated to each other and they may have their own names, but since we use some default name already in these two agent types, maybe `Name` property in `Agent.cs` should be revisited and defined as non-nullable with some default value, which should be easy to configure if needed. It looks like `Name` property is important in order to be able to differentiate these agents on UIs and telemetry tools. cc: @westey-m

dotnet/src/Microsoft.Extensions.AI.Agents/OpenTelemetryAgent.cs
```diff
@@ -0,0 +1,319 @@
...
+        int inputMessageCount,
+        bool isStreaming)
+    {
+        // Record operation duration metric
+        if (this._operationDurationHistogram.Enabled && stopwatch is not null)
+        {
+            TagList tags = new()
+            {
+                { AgentOpenTelemetryConsts.GenAI.OperationName, AgentOpenTelemetryConsts.GenAI.Operations.InvokeAgent },
+                { AgentOpenTelemetryConsts.GenAI.Agent.Name, this.Name ?? "UnnamedAgent" }
```
dmytrostruk: But I do think that if we use `UnnamedAgent` default name in multiple places in the same `OpenTelemetryAgent` class, we should put it into a separate constant on class level.

dotnet/src/Microsoft.Extensions.AI.Agents/AgentExtensions.cs
rogerbarreto: I will add this pattern as part of the Delegating strategy update track in the issue below, [previously discussed in this pr](https://github.com/microsoft/agent-framework/pull/182#discussion_r2205399873)

- https://github.com/microsoft/agent-framework/issues/183

dotnet/src/Microsoft.Extensions.AI.Agents/OpenTelemetryAgent.cs
rogerbarreto: I think I had the same question a while ago, for now I'm moving all the "UnnamedAgent" to a const, for consistency.

dotnet/src/Microsoft.Extensions.AI.Agents/OpenTelemetryAgent.cs
rogerbarreto: Im using the OpenTelemetryConsts one for now but I rather prefer having this in the Agent.Name default setting.

dotnet/src/Microsoft.Extensions.AI.Agents/OpenTelemetryAgent.cs
westey-m: From an Agent abstraction perspective I think we have three options:
1. Make Name non-nullable with a default
I would expect Name to be something that is useful for a human to identify the agent, vs Id, which may just be a Guid and therefore harder use to quickly identify an agent.  If we default to some value like "UnnamedAgent" on the Agent itself, there is not really that much value in Name when it's not specified, since all unnamed agents will have this same name.  Consumers also then don't have the opportunity to fall back to a sensible default for their use case, e.g. falling back to id where appropriate, since they just see that name is set and don't know whether it's a default or not.

2. Make name required
This complicates the basic case where a user just wants to use a simple standalone ChatClientAgent, because it forces them to provide a name, even though it's really not required for the simple scenario.

3. Keep as is, i.e. Name is optional and nullable.
If we do this, we would still need to have some fallback here for tags of course, since it is a dimension that we need a value for.  Is there any official convention in OpenTelemetry for cases like these?  E.g. using special formatting to denote that it's a placeholder like "<unknown>" or similar similar.

Based on my concerns around options 1 & 2, my preference is still option 3.

rogerbarreto: I agree in keeping the 3. also for the sake of having a way to know if the Name was set or not. From a coding perspective woudn't be elegant checking if my agent name is "XYZ" or a constant I need to know prior. `null` gives the best message that something is not set, but is rather not required for all purposes.

dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatClientAgent.cs
```diff
@@ -146,14 +146,14 @@ public override async IAsyncEnumerable<AgentRunResponseUpdate> RunStreamingAsync
             await this.PrepareThreadAndMessagesAsync(thread, inputMessages, options, cancellationToken).ConfigureAwait(false);
 
         int messageCount = threadMessages.Count;
-        var agentName = this.GetAgentName();
+        var loggingAgentName = this.GetAgentName();
```
westey-m: Consider also renaming the method.
```suggestion
        var loggingAgentName = this.GetLoggingAgentName();
```

Issue #114, opened 2025-06-30
Move AzureOpenAI model client to agent-framework


markwallace-microsoft:
Once the code has been refactored to remove Kernel etc. we need to move the code to the agent-framework repository

- [ ] Move only necessary code and adhere to new patterns
- [ ] All unit tests are moved and passing
- [ ] All integration tests are moved and passing
- [ ] All samples are moved and work

PR #185, opened 2025-07-14
Python: Azure chat client

peterychang:
Adds Azure chat client and tests for azure/openai chat clients

resolves https://github.com/microsoft/agent-framework/issues/114


python/packages/main/tests/unit/test_openai_chat_completion.py
eavanvalkenburg, line 1: should be test_openai_chat_client.py

python/packages/main/agent_framework/openai/__init__.py
```diff
@@ -1,10 +1,13 @@
...
 
 
-from ._chat_client import OpenAIChatClient
-from ._shared import OpenAISettings
+from ._chat_client import OpenAIChatClient, OpenAIChatClientBase
+from ._shared import OpenAIHandler, OpenAIModelTypes, OpenAISettings
 
 __all__ = [
     "OpenAIChatClient",
+    "OpenAIChatClientBase",
```
eavanvalkenburg: this might be confusing for people, I think from the azure package we should be able to do `from agent_framework.openai._chat_clients.py import OpenAIChatClientBase`

python/packages/azure/agent_framework/azure/__init__.py
```diff
@@ -2,9 +2,14 @@
...
+
 try:
     __version__ = importlib.metadata.version(__name__)
 except importlib.metadata.PackageNotFoundError:
     __version__ = "0.0.0"  # Fallback for development mode
 
-__all__ = ["__version__"]
+__all__ = [
+    "AzureChatClient",
+    "__version__",
```
eavanvalkenburg: we should probably expose the entra get token method here as well

python/packages/azure/agent_framework/azure/_azure_config_base.py
eavanvalkenburg, line 1: let's replicate the structure of the openai folder, so a file called `_shared.py` for the configbase, settings and consts and then `_chat_client.py` for the chat client (and eventually other clients in their own files.

python/packages/azure/agent_framework_azure/_chat_client.py
```diff
@@ -0,0 +1,237 @@
...
+            api_version=settings.get("api_version"),
+            ad_token=settings.get("ad_token"),
+            ad_token_provider=settings.get("ad_token_provider"),
+            default_headers=settings.get("default_headers"),
+            env_file_path=settings.get("env_file_path"),
+        )
+
+    def get_chat_options_class(self) -> type["ChatOptions"]:
+        """Create a request settings object."""
+        return ChatOptions
```
dmytrostruk: Does it need to be public? I also see that this method is not used in the codebase.

python/packages/azure/agent_framework_azure/_chat_client.py
```diff
@@ -0,0 +1,237 @@
...
+                call_id=str(uuid4()),
+                name="Azure-OnYourData",
+                arguments={"query": tool_message.get("intent", [])},
+            )
+            result = FunctionResultContent(
+                call_id=function_call.call_id,
+                result=tool_message["citations"],
+                exception=function_call.exception,
+                additional_properties=function_call.additional_properties,
+            )
```
dmytrostruk, line 178: Do we need this?

python/packages/azure/agent_framework_azure/_chat_client.py
```diff
@@ -0,0 +1,237 @@
...
+        content = choice.message if isinstance(choice, Choice) else choice.delta
+        # When you enable asynchronous content filtering in Azure OpenAI, you may receive empty deltas
+        if content and content.model_extra is not None:
+            return content.model_extra.get("context", None)
+        # openai allows extra content, so model_extra will be a dict, but we need to check anyway, but no way to test.
+        return None  # pragma: no cover
+
+    @staticmethod
+    def split_message(message: "ChatResponse") -> ChatResponse:
+        """Split an Azure On Your Data response into separate ChatMessages within the ChatResponse.
```
dmytrostruk, line 197: Since it's public, do we expect this method to be used by developers? My concern is that with such specific public methods we will lose the meaning of ChatClient abstraction. Ideally, the logic should fit into ChatClient protocol, which is `get_response/get_streaming_response` methods. This will allow developers to easily change the AI provider without a need to change business logic, and it will be used together with `ChatClientAgent` in the same way.

Once specific methods are used, such as `azure_chat_client.split_message(content)` etc, it won't be possible to use another AI provider. I think there are cases when we can't prevent this (e.g. when very specific tool is used which is supported by only that provider), but it feels like Azure-On-Your-Data functionality should fit into current ChatClient protocol, since it's just a regular chat completion request to Azure but with grounded data. Except specific properties like `citations`, which can be part of response object metadata, I don't see any other specifics that will require us to expose additional methods to handle it, unless I'm missing something. Please let me know what you think.

python/packages/azure/tests/integration/test_azure_openai_chat_client_responses.py
dmytrostruk, line 1: nit: I would name this file `test_azure_chat_client.py`.

python/packages/azure/tests/integration/test_azure_openai_chat_client_responses.py
```diff
@@ -0,0 +1,63 @@
...
+    """Returns a story about Emily and David."""
+    return (
+        "Emily and David, two passionate scientists, met during a research expedition to Antarctica. "
+        "Bonded by their love for the natural world and shared curiosity, they uncovered a "
+        "groundbreaking phenomenon in glaciology that could potentially reshape our understanding "
+        "of climate change."
+    )
+
+
+async def test_azure_openai_chat_client_response() -> None:
```
dmytrostruk: nit: It would be also good to add an integration test for streaming case.

python/packages/main/tests/integration/test_openai_chat_completion_responses.py
dmytrostruk, line 1: nit: I would name this file `test_openai_chat_client.py`

python/packages/azure/agent_framework_azure/_chat_client.py
eavanvalkenburg: Good catch, that's indeed not needed anymore with the new wow with chatoptions

python/packages/azure/agent_framework_azure/_chat_client.py
peterychang: I'm not really sure what the original intention or requirement behind it was. I assume its still necessary for some component of azure though?

python/packages/azure/agent_framework_azure/_chat_client.py
peterychang: AFAIK this function was required to reformat the responses into something openai would be happy with. I'm not sure what the workflow around it would have been under SK.

@eavanvalkenburg do you have an opinion on this one?

PR #188, opened 2025-07-15
Add ChatClientAgent constructor overload

westey-m:
### Motivation and Context

There are some common settings which are expected to be used very often with a ChatClientAgent.  One of these is tools, but to provide it the user has to provide a ChatClientOptions wrapped in a ChatClientAgentOptions, which is not intuitive.

This PR adds a ChatClientAgent constructor overload with a list of optional parameters, which contain just the most commonly used settings.  This signature is not expected to grow over time with new parameters, as it should be catering for the 90% use case.

It allows for collapsing a constructor like this:
```csharp
        var agent = new ChatClientAgent(
            chatClient,
            options: new()
            {
                Instructions = "Answer questions about the menu.",
                ChatOptions = new()
                {
                    Tools = [AIFunctionFactory.Create(menuTools.GetSpecials)]
                }
            });
```

into this:
```csharp
        var agent = new ChatClientAgent(
            chatClient,
            instructions: "Answer questions about the menu.",
            tools: [AIFunctionFactory.Create(menuTools.GetSpecials)]);
```

The new signature has optional instructions, name, description, tools and loggerFactory.
Id is not there since for the significant majority of IChatClient implementations there is no external id, and the agent will fall back to a Guid.
Passing an Id would only be applicable for service based agents used via IChatClient, which is currently limited to Foundry Persistent (which we plan to replace with a dedicated Agent implementation) and OpenAI Assistant (which is being deprecated).

#92 

### Description

- Added a constructor overload with the most common options
- Refactored some samples to make use of the new signature

### Contribution Checklist

### DIFF
--- a/dotnet/tests/Microsoft.Extensions.AI.Agents.UnitTests/ChatCompletion/ChatClientAgentTests.cs
+++ b/dotnet/tests/Microsoft.Extensions.AI.Agents.UnitTests/ChatCompletion/ChatClientAgentTests.cs
@@ -620,6 +620,28 @@ public class ChatClientAgentTests
 
     #endregion
 
+    #region Options Constructor Tests
+
+    /// <summary>
+    /// Checks that the various properties on <see cref="ChatClientAgent"/> are null or defaulted when not provided to the constructor.
+    /// </summary>
+    [Fact]
+    public void OptionsPropertiesNullOrDefaultWhenNotProvidedToConstructor()
+    {
+        // Arrange
+        var chatClient = new Mock<IChatClient>().Object;
+        ChatClientAgent agent = new(chatClient, options: null!);
+
+        // Act & Assert
+        Assert.NotNull(agent.Id);
+        Assert.Null(agent.Instructions);
+        Assert.Null(agent.Name);
+        Assert.Null(agent.Description);
+        Assert.Null(agent.ChatOptions);
+    }
+
+    #endregion
+
     #region ChatOptions Property Tests
 
     /// <summary>


dotnet/samples/GettingStarted/Providers/ChatClientAgent_With_OpenAIResponseChatCompletion.cs
```diff
@@ -6,7 +6,7 @@
 using OpenAI;
 using OpenAI.Responses;
 
-#pragma warning disable CS8524 // The switch expression does not handle some values of its input type (it is not exhaustive) involving an unnamed enum value.
+#pragma warning disable OPENAI001 // Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.
```
stephentoub: Consider just putting OPENAI001 into a NoWarn in the project or in the editorconfig

westey-m: Sounds good, done.

dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatClientAgent.cs
```diff
@@ -25,9 +25,41 @@ public sealed class ChatClientAgent : Agent
...
+            }
+        };
+
+        // Get the type of the chat client before wrapping it as an agent invoking chat client.
+        this._chatClientType = chatClient.GetType();
+
+        this.ChatClient = chatClient.AsAgentInvokingChatClient();
+
+        this._logger = (loggerFactory ?? chatClient.GetService<ILoggerFactory>() ?? NullLoggerFactory.Instance).CreateLogger<ChatClientAgent>();
+    }
```
rogerbarreto: nit. Defer to the other ctor.

```suggestion
    public ChatClientAgent(IChatClient chatClient, string? instructions = null, string? name = null, string? description = null, IList<AITool>? tools = null, ILoggerFactory? loggerFactory = null)
        : this(chatClient, options: new ChatClientAgentOptions()
        {
            Name = name,
            Description = description,
            Instructions = instructions,
            ChatOptions = tools is null ? null : new() { Tools = tools }
        }, loggerFactory)
    {
    }
```

westey-m: Thanks, updated.

PR #193, opened 2025-07-16
Python: Implemented FoundryChatClient

dmytrostruk:
This pull request introduces a new Python package, `agent-framework-foundry`, which integrates Azure AI Foundry with the Agent Framework. It includes the implementation of `FoundryChatClient`, the initial setup of the package, unit tests for the FoundryChatClient, and examples.

### DIFF
--- a/python/packages/foundry/agent_framework_foundry/_chat_client.py
+++ b/python/packages/foundry/agent_framework_foundry/_chat_client.py
@@ -494,8 +494,11 @@ class FoundryChatClient(ChatClientBase):
 
         additional_messages: list[ThreadMessageOptions] | None = None
 
+        # System/developer messages are turned into instructions, since there is no such message roles in Foundry.
+        # All other messages are added 1:1, treating assistant messages as agent messages
+        # and everything else as user messages.
         for chat_message in messages:
-            if chat_message.role == ChatRole.SYSTEM or chat_message.role.value == "developer":
+            if chat_message.role.value in ["system", "developer"]:
                 for text_content in [content for content in chat_message.contents if isinstance(content, TextContent)]:
                     instructions.append(text_content.text)
 
--- a/python/packages/foundry/tests/unit/test_foundry_chat_client.py
+++ b/python/packages/foundry/tests/unit/test_foundry_chat_client.py
@@ -112,7 +112,6 @@ def test_foundry_chat_client_from_dict(mock_ai_project_client: MagicMock) -> Non
     assert chat_client.thread_id == "test-thread-id"
 
 
-@pytest.mark.asyncio
 async def test_foundry_chat_client_get_agent_id_or_create_existing_agent(
     mock_ai_project_client: MagicMock,
 ) -> None:
@@ -125,7 +124,6 @@ async def test_foundry_chat_client_get_agent_id_or_create_existing_agent(
     assert not chat_client._should_delete_agent  # type: ignore
 
 
-@pytest.mark.asyncio
 async def test_foundry_chat_client_get_agent_id_or_create_create_new(
     mock_ai_project_client: MagicMock,
     foundry_unit_test_env: dict[str, str],
@@ -142,7 +140,6 @@ async def test_foundry_chat_client_get_agent_id_or_create_create_new(
     assert chat_client._should_delete_agent  # type: ignore
 
 
-@pytest.mark.asyncio
 async def test_foundry_chat_client_get_agent_id_or_create_missing_model(
     mock_ai_project_client: MagicMock,
 ) -> None:
@@ -153,7 +150,6 @@ async def test_foundry_chat_client_get_agent_id_or_create_missing_model(
         await chat_client._get_agent_id_or_create()  # type: ignore
 
 
-@pytest.mark.asyncio
 async def test_foundry_chat_client_cleanup_agent_if_needed_should_delete(
     mock_ai_project_client: MagicMock,
 ) -> None:
@@ -169,7 +165,6 @@ async def test_foundry_chat_client_cleanup_agent_if_needed_should_delete(
     assert not chat_client._should_delete_agent  # type: ignore
 
 
-@pytest.mark.asyncio
 async def test_foundry_chat_client_cleanup_agent_if_needed_should_not_delete(
     mock_ai_project_client: MagicMock,
 ) -> None:
@@ -185,7 +180,6 @@ async def test_foundry_chat_client_cleanup_agent_if_needed_should_not_delete(
     assert not chat_client._should_delete_agent  # type: ignore
 
 
-@pytest.mark.asyncio
 async def test_foundry_chat_client_cleanup_agent_if_needed_exception_handling(
     mock_ai_project_client: MagicMock,
 ) -> None:
@@ -199,7 +193,6 @@ async def test_foundry_chat_client_cleanup_agent_if_needed_exception_handling(
         await chat_client._cleanup_agent_if_needed()  # type: ignore
 
 
-@pytest.mark.asyncio
 async def test_foundry_chat_client_aclose(mock_ai_project_client: MagicMock) -> None:
     """Test aclose method calls cleanup."""
     chat_client = create_test_foundry_chat_client(
@@ -212,7 +205,6 @@ async def test_foundry_chat_client_aclose(mock_ai_project_client: MagicMock) -> 
     mock_ai_project_client.agents.delete_agent.assert_called_once_with("agent-to-delete")
 
 
-@pytest.mark.asyncio
 async def test_foundry_chat_client_async_context_manager(mock_ai_project_client: MagicMock) -> None:
     """Test async context manager functionality."""
     chat_client = create_test_foundry_chat_client(
--- a/python/packages/main/agent_framework/_agents.py
+++ b/python/packages/main/agent_framework/_agents.py
@@ -1,9 +1,15 @@
 # Copyright (c) Microsoft. All rights reserved.
 
+import sys
 from collections.abc import AsyncIterable, Callable, MutableMapping, Sequence
 from enum import Enum
 from typing import Any, Literal, Protocol, TypeVar, runtime_checkable
 from uuid import uuid4
+
+if sys.version_info >= (3, 11):
+    from typing import Self  # pragma: no cover
+else:
+    from typing_extensions import Self  # pragma: no cover
 
 from pydantic import BaseModel, Field
 
@@ -359,7 +365,7 @@ class ChatClientAgent(AgentBase):
 
         super().__init__(**args)
 
-    async def __aenter__(self) -> "ChatClientAgent":
+    async def __aenter__(self) -> "Self":
         """Async context manager entry.
 
         If the chat_client supports async context management, enter its context.
--- a/python/samples/getting_started/agents/foundry/foundry_basic.py
+++ b/python/samples/getting_started/agents/foundry/foundry_basic.py
@@ -21,6 +21,8 @@ def get_weather(
 async def main() -> None:
     print("=== Basic Foundry Chat Client Example ===")
 
+    # Since no Agent ID is provided, the agent will be automatically created
+    # and deleted after getting a response
     async with ChatClientAgent(
         chat_client=FoundryChatClient(),
         instructions="You are a helpful weather agent.",
--- a/python/samples/getting_started/agents/foundry/foundry_with_existing_agent.py
+++ b/python/samples/getting_started/agents/foundry/foundry_with_existing_agent.py
@@ -24,7 +24,7 @@ def get_weather(
 async def main() -> None:
     print("=== Foundry Chat Client with Existing Agent ===")
 
-    # Create the client and agent yourself
+    # Create the client
     client = AIProjectClient(endpoint=os.environ["FOUNDRY_PROJECT_ENDPOINT"], credential=AzureCliCredential())
 
     # Create an agent that will persist
--- a/python/samples/getting_started/agents/foundry/foundry_with_existing_client.py
+++ b/python/samples/getting_started/agents/foundry/foundry_with_existing_client.py
@@ -24,9 +24,11 @@ def get_weather(
 async def main() -> None:
     print("=== Foundry Chat Client with Existing AIProjectClient ===")
 
-    # Create the client yourself
+    # Create the client
     client = AIProjectClient(endpoint=os.environ["FOUNDRY_PROJECT_ENDPOINT"], credential=AzureCliCredential())
 
+    # Since no Agent ID is provided, the agent will be automatically created
+    # and deleted after getting a response
     async with ChatClientAgent(
         chat_client=FoundryChatClient(
             client=client,
--- a/python/samples/getting_started/agents/foundry/foundry_with_explicit_settings.py
+++ b/python/samples/getting_started/agents/foundry/foundry_with_explicit_settings.py
@@ -23,6 +23,8 @@ def get_weather(
 async def main() -> None:
     print("=== Foundry Chat Client with Explicit Settings ===")
 
+    # Since no Agent ID is provided, the agent will be automatically created
+    # and deleted after getting a response
     async with ChatClientAgent(
         chat_client=FoundryChatClient(
             project_endpoint=os.environ["FOUNDRY_PROJECT_ENDPOINT"],


python/packages/main/agent_framework/_types.py
```diff
@@ -1421,6 +1440,8 @@ def serialize_model(self) -> str:
 class ChatOptions(AFBaseModel):
     """Common request settings for AI services."""
 
+    conversation_id: str | None = None
+    allow_multiple_tool_calls: bool | None = None
```
eavanvalkenburg: I've put all of these in alphabetical order, please do the same!

python/packages/foundry/agent_framework_foundry/_chat_client.py
```diff
@@ -0,0 +1,354 @@
...
+    ThreadMessageOptions,
+    ThreadRun,
+    ToolOutput,
+)
+from azure.ai.projects.aio import AIProjectClient
+
+
+@use_tool_calling
+class FoundryChatClient(ChatClientBase):
+    client: AIProjectClient
```
eavanvalkenburg: I think we also want to provide a way to set up without the dev needing to create the client, because Shawn wants the foundry agent to be a first class citizen!

python/packages/main/agent_framework/foundry/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+            ) from exc
+    raise AttributeError(f"module {__name__} has no attribute {name}")
+
+
+def __dir__() -> list[str]:
+    return list(_IMPORTS.keys())
+
+
+if TYPE_CHECKING:
+    from agent_framework_foundry import __version__  # noqa: F401
```
eavanvalkenburg: This also should include foundrychatclient to make sure type checkers can resolve them

python/samples/getting_started/agents/foundry_chat_client_agent.py
```diff
@@ -0,0 +1,43 @@
+# Copyright (c) Microsoft. All rights reserved.
+
+import asyncio
+import os
+from random import randint
+from typing import Annotated
+
+from agent_framework import ChatClientAgent
+from agent_framework_foundry import FoundryChatClient
```
eavanvalkenburg: This should be `from agent_framework.foundry import...`

python/packages/main/agent_framework/foundry/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+def __getattr__(name: str) -> Any:
+    if name in _IMPORTS:
+        submod_name = _IMPORTS[name]
+        try:
+            module = importlib.import_module(submod_name, package=__name__)
+            return getattr(module, name)
+        except ModuleNotFoundError as exc:
+            raise ModuleNotFoundError(
+                f"The '{PACKAGE_EXTRA}' extra is not installed, "
+                f"please do `pip install agent-framework[{PACKAGE_EXTRA}]`"
```
eavanvalkenburg, line 18: the extra does need to be defined, in the pyproject of main, same as azure.

python/packages/foundry/agent_framework_foundry/_chat_client.py
dmytrostruk: @eavanvalkenburg I provided a way to set up without the dev needing to create the client, but there is another problem - if user doesn't provide client, it means that most probably they don't have already created remote agent, but in order to make requests we need to create one. If we are going to create an agent on our side - we also need to delete it once the request is finished. 

For these purposes, I've implemented `FoundryChatClient` as disposable with `__aenter__/__aexit__` methods:
1. If agent id is not provided and user wants us to handle the creation/deletion - `FoundryChatClient` should be used within `async with`. 
2. If agent id is provided, we won't delete the agent because we didn't create it, so user can use `FoundryChatClient` directly.

I've added examples for these cases, please take a look. If you have alternative ideas how to implement it, please let me know.

python/packages/foundry/agent_framework_foundry/_chat_client.py
```diff
@@ -0,0 +1,550 @@
...
+            default_thread_id: Default thread ID to use for conversations.
+            project_endpoint: The Azure AI Foundry project endpoint URL. Used if client is not provided.
+            model_deployment_name: The model deployment name to use for agent creation.
+            credential: Azure async credential to use for authentication. If not provided,
+                DefaultAzureCredential will be used.
+            env_file_path: Path to environment file for loading settings.
+            env_file_encoding: Encoding of the environment file.
+            **kwargs: Additional keyword arguments passed to the parent class.
+        """
+        try:
```
eavanvalkenburg, line 124: we can probably skip this if `client` is passed in?

python/packages/foundry/agent_framework_foundry/_chat_client.py
```diff
@@ -0,0 +1,550 @@
...
+            if credential is None:
+                from azure.identity.aio import DefaultAzureCredential
+
+                credential = DefaultAzureCredential()
+
+            client = AIProjectClient(endpoint=foundry_settings.project_endpoint, credential=credential)
+
+        super().__init__(
+            client=client,  # type: ignore[reportCallIssue]
+            agent_id=agent_id,  # type: ignore[reportCallIssue]
```
eavanvalkenburg, line 153: what is someone doesn't pass in a `client`, but they do pass in a `agent_id`? wouldn't that be enough for the scenario of a existing agent?

python/packages/foundry/agent_framework_foundry/_chat_client.py
```diff
@@ -0,0 +1,550 @@
...
+
+    async def __aenter__(self) -> "FoundryChatClient":
+        """Async context manager entry."""
+        return self
+
+    async def __aexit__(self, exc_type: type[BaseException] | None, exc_val: BaseException | None, exc_tb: Any) -> None:
+        """Async context manager exit - clean up any agents we created."""
+        await self.aclose()
+
+    async def aclose(self) -> None:
```
eavanvalkenburg: we don't use any async markers in method names
```suggestion
    async def close(self) -> None:
```

python/packages/foundry/agent_framework_foundry/_chat_client.py
```diff
@@ -0,0 +1,550 @@
...
+    ) -> None:
+        """Initialize a FoundryChatClient.
+
+        Args:
+            client: An existing AIProjectClient to use. If not provided, one will be created.
+            agent_id: The ID of an existing agent to use. If not provided and client is provided,
+                a new agent will be created (and deleted after the request). If neither client
+                nor agent_id is provided, both will be created and managed automatically.
+            agent_name: The name to use when creating new agents.
+            default_thread_id: Default thread ID to use for conversations.
```
eavanvalkenburg: should this just be called `thread_id` so the semantics are the same as with agent_id: if not provided, we create, if provided we (try to) use existing?

python/packages/foundry/agent_framework_foundry/_chat_client.py
```diff
@@ -0,0 +1,550 @@
...
+            client = AIProjectClient(endpoint=foundry_settings.project_endpoint, credential=credential)
+
+        super().__init__(
+            client=client,  # type: ignore[reportCallIssue]
+            agent_id=agent_id,  # type: ignore[reportCallIssue]
+            default_thread_id=default_thread_id,  # type: ignore[reportCallIssue]
+            **kwargs,
+        )
+
+        self._should_delete_agent = False
```
eavanvalkenburg, line 158: shouldn't this be determined by the presence of `agent_id`?
```suggestion
        self._should_delete_agent = not agent_id
```

python/packages/foundry/agent_framework_foundry/_chat_client.py
eavanvalkenburg: and add the note that if the `conversation_id` is provided to the run that that is used.

python/packages/foundry/agent_framework_foundry/_chat_client.py
```diff
@@ -0,0 +1,550 @@
...
+    async def _ensure_agent_exists(self) -> str:
+        """Ensure an agent exists, creating one if necessary.
+
+        Returns:
+            str: The agent_id to use
+        """
+        # If no agent_id is provided, create a temporary agent
+        if self.agent_id is None:
+            if not self._foundry_settings.model_deployment_name:
+                raise ServiceInitializationError("Model deployment name is required for agent creation.")
```
eavanvalkenburg, line 239: can we do this check at creation time, when a client is not provided?

python/packages/foundry/agent_framework_foundry/_chat_client.py
```diff
@@ -0,0 +1,550 @@
...
+
+            agent_name = self._foundry_settings.agent_name
+            created_agent = await self.client.agents.create_agent(
+                model=self._foundry_settings.model_deployment_name, name=agent_name
+            )
+            self.agent_id = created_agent.id
+            self._should_delete_agent = True
+
+        return self.agent_id
```
eavanvalkenburg, line 248: should we also actually check if the agent_id exists in the service? `ensure exists` kind of implies that

python/packages/foundry/agent_framework_foundry/_chat_client.py
```diff
@@ -0,0 +1,550 @@
...
+
+        return stream, final_thread_id
+
+    async def _get_active_thread_run(self, thread_id: str | None) -> ThreadRun | None:
+        """Get any active run for the given thread."""
+        if thread_id is None:
+            return None
+
+        async for run in self.client.agents.runs.list(thread_id=thread_id, limit=1, order=ListSortOrder.DESCENDING):
+            if (
```
eavanvalkenburg:
```suggestion
            if (run.status not in [RunStatus.COMPLETED, RunStatus.CANCELLED, ...])
```

python/packages/foundry/agent_framework_foundry/_chat_client.py
```diff
@@ -0,0 +1,550 @@
...
+        stream: AsyncAgentRunStream[AsyncAgentEventHandler[Any]] | AsyncAgentEventHandler[Any],
+        thread_id: str,
+    ) -> AsyncIterable[ChatResponseUpdate]:
+        """Process events from the agent stream and yield ChatResponseUpdate objects."""
+        response_id: str | None = None
+
+        if stream is not None:
+            # Use 'async with' only if the stream supports async context management (main agent stream).
+            # Tool output handlers only support async iteration, not context management.
+            if hasattr(stream, "__aenter__") and hasattr(stream, "__aexit__"):  # type: ignore
```
eavanvalkenburg: you can probably do this:
```suggestion
            if isinstance(stream, contextlib.AbstractAsyncContextManager):
```

python/packages/foundry/agent_framework_foundry/_chat_client.py
```diff
@@ -0,0 +1,550 @@
...
+                async with stream as response_stream:  # type: ignore
+                    stream_iter = response_stream
```
eavanvalkenburg: after this statement is done, the context is closed already, is that not too soon?

python/samples/getting_started/agents/foundry/foundry_basic.py
```diff
@@ -0,0 +1,37 @@
...
+    conditions = ["sunny", "cloudy", "rainy", "stormy"]
+    return f"The weather in {location} is {conditions[randint(0, 3)]} with a high of {randint(10, 30)}°C."
+
+
+async def main() -> None:
+    print("=== Basic Foundry Chat Client Example ===")
+
+    async with FoundryChatClient() as chat_client:
+        agent = ChatClientAgent(
+            chat_client,
```
eavanvalkenburg: I'm open to adding a setup in ChatClientAgent that checks if the chat_client is a context manager and does the __aenter__ and __aexit__ itself, making the ChatClientAgent a ContextManager itself... what do you think?

basically you can then always do, regardless of ChatClient type:
```python
async with ChatClientAgent(chat_client=FoundryChatClient(), instructions=...., ) as agent:
    response = await agent.run("What's ....")
```

python/samples/getting_started/agents/chat_client_agent.py
```diff
@@ -17,11 +18,12 @@ def get_weather(
...
 
-async def main():
+async def main() -> None:
     instructions = "You are a helpful assistant, you can help the user with weather information."
     agent = ChatClientAgent(OpenAIChatClient(), instructions=instructions, tools=get_weather)
     print(str(await agent.run("What's the weather in Amsterdam?")))
 
 
 if __name__ == "__main__":
+    load_dotenv()
```
eavanvalkenburg, line 28: this isn't needed the pydantic settings takes care of that, it's only needed when you manually want to call `sys.getenv` and expect things from a .env file.

python/packages/foundry/agent_framework_foundry/_chat_client.py
dmytrostruk: > wouldn't that be enough for the scenario of a existing agent?

If `agent_id` is provided but `client` is not provided, users will still need to pass settings, so we can initialize the client on our side. Because we need to use a correct Foundry endpoint etc, otherwise it's not clear where this existing agent is hosted.

python/packages/foundry/agent_framework_foundry/_chat_client.py
dmytrostruk: It can be, although I decided to put this logic in `_ensure_agent_exists` where we actually create an agent and only then we set this property to `True` if this operation was successful. Because if `agent_id` is incorrect (for example it doesn't exist) and we set this property to `True`, the entire state is a little bit incorrect, because we can't delete an agent that doesn't exist etc. Also, if we are not performing any request and we just initialized `FoundryChatClient`, we shouldn't delete an agent, because it is created only when we actually perform a request.

python/packages/foundry/agent_framework_foundry/_chat_client.py
dmytrostruk: I wouldn't do that, because it's an additional round-trip to the server. In all cases, we will perform a request to get a response, and if agent doesn't exist, we will report it back to the user during the request. Also, if user passes an invalid agent_id, we can't simply ignore it and create a new one, because this situation looks like something is wrong in user's setup, so it's better to inform about it. 

But I agree that in this case method name `_ensure_agent_exists` is a little bit incorrect, because we are not handling already created agents, so I renamed it to `_get_agent_id_or_create`.

python/packages/foundry/agent_framework_foundry/_chat_client.py
dmytrostruk: To be honest, I took this logic from semantigen repository:
https://github.com/microsoft/semantigen/blob/991626c82d8f9b6aedcff5391abdf7317461c070/python/semantic_kernel/agents/foundry/foundry_agent_thread_actions.py#L486-L490

I tried without `async with` and it's not working in this case.

python/samples/getting_started/agents/chat_client_agent.py
dmytrostruk: In this case, it's something wrong either with pydantic settings or my local setup, because if I remove `load_dotenv()` call, the settings are not pulled from `.env` file.

python/packages/foundry/agent_framework_foundry/_chat_client.py
dmytrostruk: Actually, we are also using `model_deployment_name` and `agent_name` properties which are not related to the client, but to the agent.

python/packages/foundry/agent_framework_foundry/_chat_client.py
dmytrostruk: Refactored the code.

python/packages/foundry/agent_framework_foundry/_chat_client.py
```diff
@@ -0,0 +1,566 @@
...
+                    )
+                )
+
+        instructions: list[str] = []
+        tool_results: list[FunctionResultContent] | None = None
+
+        additional_messages: list[ThreadMessageOptions] | None = None
+
+        for chat_message in messages:
+            if chat_message.role == ChatRole.SYSTEM or chat_message.role.value == "developer":
```
eavanvalkenburg: This is a bit weird, and we can also do `chat_message.role in`

python/packages/foundry/tests/unit/test_foundry_chat_client.py
```diff
@@ -0,0 +1,280 @@
...
+        thread_id=settings["thread_id"],  # type: ignore
+        foundry_settings=foundry_settings,
+    )
+
+    assert chat_client.client is mock_ai_project_client
+    assert chat_client.agent_id == "test-agent-id"
+    assert chat_client.thread_id == "test-thread-id"
+
+
+@pytest.mark.asyncio
```
eavanvalkenburg: We don't need this, we auto run with async if needed

python/packages/main/agent_framework/_agents.py
```diff
@@ -349,6 +349,23 @@ def __init__(
 
         super().__init__(**args)
 
+    async def __aenter__(self) -> "ChatClientAgent":
```
eavanvalkenburg: This should return Self, otherwise a potential subclass will not have correct typing

python/packages/main/agent_framework/_types.py
```diff
@@ -252,7 +252,7 @@ def _coalesce_text_content(
                 first_new_content = None
             coalesced_contents.append(content)
     if current_texts:
-        coalesced_contents.append(type_(text=" ".join(current_texts)))
+        coalesced_contents.append(type_(text="".join(current_texts)))
```
eavanvalkenburg, line 283: I just tweaked all of these to make some kind of sense and not have too many extra spaces or lack of spaces...

python/samples/getting_started/agents/foundry/foundry_basic.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+async def main() -> None:
+    print("=== Basic Foundry Chat Client Example ===")
+
+    async with ChatClientAgent(
+        chat_client=FoundryChatClient(),
+        instructions="You are a helpful weather agent.",
+        tools=get_weather,
+    ) as agent:
```
eavanvalkenburg, line 30: Maybe add a comment that this setup will create and delete the agent in foundry

python/packages/foundry/agent_framework_foundry/_chat_client.py
dmytrostruk: Added comment and updated the code.

python/packages/main/agent_framework/_types.py
dmytrostruk: The problem is that I was getting a response from Foundry with spaces, while we were adding our own spaces on top of it, so the result looked weird. I think that on our side we shouldn't do any modifications and just return it as-is, because only AI model can decide where to put a space or any other character in word or sentence to have some meaning. For example, if the word is long enough, it can be returned back to the user in 2 chunks, and we can't put a space between it, because it will break the word.

PR #197, opened 2025-07-17
Initial draft of actor runtime abstractions

ReubenBond:
This PR adds an initial draft of the new actor runtime abstractions which are intended to replace IAgentRuntime.


dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Abstractions/ActorMessage.cs
```diff
@@ -0,0 +1,109 @@
...
+            if (reader.TokenType != JsonTokenType.StartObject)
+            {
+                throw new JsonException("Expected StartObject token");
+            }
+
+            using var doc = JsonDocument.ParseValue(ref reader);
+            var root = doc.RootElement;
+
+            // All ActorMessage messages must have a type property
+            if (!root.TryGetProperty("type", out var typeProperty))
```
stephentoub: Just curious, does the built-in polymorphic support not suffice?

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Abstractions/ActorResponse.cs
```diff
@@ -0,0 +1,65 @@
...
+    public RequestStatus Status { get; set; }
+
+    /// <summary>
+    /// Waits for the request to complete.
+    /// </summary>
+    /// <param name="cancellationToken">A token to cancel the wait operation.</param>
+    /// <returns>A task that completes when the request is finished.</returns>
+    public ValueTask WaitAsync(CancellationToken cancellationToken) 
+    {
+        throw new NotImplementedException("The method WaitAsync is not yet implemented.");
```
stephentoub: Is your intent that these NIE methods will end up being abstract or virtual, or it's really just not implemented yet? Trying to understand the shape.

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Abstractions/ActorWriteOperationBatch.cs
```diff
@@ -0,0 +1,26 @@
...
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Extensions.AI.Agents.Runtime.Abstractions;
+
+/// <summary>
+/// Represents a batch of write operations to be performed atomically on an actor.
+/// </summary>
+/// <param name="ETag">The ETag for optimistic concurrency control.</param>
+/// <param name="Operations">The collection of write operations to perform.</param>
+public class ActorWriteOperationBatch(string ETag, IReadOnlyCollection<ActorWriteOperation> Operations)
```
stephentoub: Nit: primary ctor parameters should still be cased like any other parameter :)

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Abstractions/IActorContext.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.AI.Agents.Runtime.Abstractions;
+
+// Implemented by the runtime
+/// <summary>
+/// Provides the runtime context for an actor, enabling it to interact with the actor system.
+/// </summary>
+public interface IActorContext
```
stephentoub: This is the new IAgentRuntime, right? This is the thing that a higher-level library would have in its public API to be parameterized with the runtime to target? I suggest this should be named accordingly, maybe even IAgentRuntime (or maybe IRuntimeContext). IActorContext makes it sound to me like it's context about the actor, not context for the actor.

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Abstractions/ActorMessage.cs
ReubenBond: It does suffice, I am fairly new to the more advanced STJ bits. I've updated all of the code to use the built-in polymorphism

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Abstractions/ActorResponse.cs
ReubenBond: Yep, I have split an `ActorResponseHandle` out from the payload data the class, and `ActorResponseHandle` is abstract and contains the 3 unimplemented methods.

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Abstractions/ActorWriteOperationBatch.cs
ReubenBond: Will fix. I did it this way because the analyzers preferred it when I did and otherwise complained about missing doc comments for the params.

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Abstractions/IActorContext.cs
ReubenBond: Need to revisit this to see how cleanly a higher-level abstraction like an Orchestration or Process can be built so that it supports an optional runtime.
I agree on the need for rename (eg to IRuntimeContext), too.

stephentoub: Maybe look at my pr, and in particular at the commented out lines? It's how I imagined they'd plug together, though I'm likely missing some things.

Issue #184, opened 2025-07-14
Overhaul orchestrations and their interaction with the agent runtime


stephentoub:
The current orchestration layer uses IAgentRuntime as an integral part of its operation. A runtime actor is created for each agent, 1:1, with message passing via the runtime as the way that all communication between agents is achieved. For example, for a sequential orchestration, the equivalent of (pseudo-code):
```C#
var input = ...;
Agent[] agents = ...;
AgentThread[] threads = ...;
for (int i = 0; i < agents.Length; i++)
{
    input = await agents[i].InvokeAsync(input);
}
return input;
```
this gets implemented today as something like:
```C#
var input = ...;
Agent[] agents = ...;
AgentThread[] threads = ...;
IAgentRuntime runtime = ...;
Actor[] actors = new Actor[agents.Length];
for (int i = 0; i < agents.Length; i++)
{
    int id = i;
    actors[id] = new();
    RegisterMessageHandler(id, actors[id], async m =>
    {
        var response = await actors[id].InvokeAsync(m.Input);
        SendMessage(runtime, id + 1, response);
    }
}

TaskCompletionSource r = new();
Actor final = new();
RegisterMessageHadler(agents.Length, final, async m =>
{
    r.SetResult(m.Input);
});

SendMessage(runtime, 0, input);
return r.Task;
```

This means you _must_ use a runtime, even for simple orchestrations, because the runtime inserted in the middle of every communication between agents. That also makes it very chatty and, as is evident from the code above, makes it more complicated.

The desired new direction is to instead just have a single runtime actor to represent the state of the whole orchestration. The orchestration ends up being written much more like the simple flow, checkpointing its state at whatever frequency is desired (the natural/simple answer is after each agent interaction, e.g.
```C#
var input = ...;
Agent[] agents = ...;
AgentThread[] threads = ...;

for (int i = 0; i < agents.Length; i++)
{
    Checkpoint(agents, threads, input, i);
    input = await agents[i].InvokeAsync(input);
}

Checkpoint(agents, threads, input, i);
return input;
```

Orchestrations can be implemented as a state machine, as well, where when checkpointed state is later restored, the implementation can effectively rehydrate its state and jump to the right point to continue execution.

When run without a runtime, the checkpointing can simply be elided.

We will want the orchestration.InvokeAsync method to accept an optional request ID. When not supplied, it'll be created automatically be the implementation (e.g. a guid); this ID can be used to later resume, and will be returned as part of the orchestration result instance.

[...]

PR #199, opened 2025-07-17
Overhaul orchestration library with new approach

stephentoub:
This effectively reimplements the orchestration library following https://github.com/microsoft/agent-framework/issues/184.

This redoes sequential, concurrent, handoff, and group chat.

- A runtime is now optional rather than being required. Someone just doing simple orchestration (which I think will be common) do not need any pub sub / runtime / etc. The implementation in such cases is also much cheaper.
- If a runtime is provided, state is saved, and resumption is now possible (but untested as of yet), and we'll want to revisit the API (in part as we bring in Reuben's new runtime APIs).
- The code implemented an orchestration is significantly simplified and centralized, e.g. sequential is just a for loop, handoff is just a while loop, etc. I've encountered multiple of what I think are bugs that stem from that complexity and that are now easily root caused and addressed.
- Orchestrations now _are_ agents, so they can be composed and used polymorphically with anything else working with agents.
- This should integrate nicely with the runtime drafted in https://github.com/microsoft/agent-framework/pull/197. The interaction is currently commented out, and I expect there will likely need to be some tweaks, but I believe this represents the general flow.

cc: @ReubenBond, @adityamandaleeka, @crickman, @TaoChenOSU, @ekzhu

@markwallace-microsoft, @alliscode, I understand there were some concerns about my doing this? Happy to chat more about it, and if you don't like the direction. If we end up deciding to reimplement these again (or throw them away completely) in favor of process framework as the underpinnings, that's fine, but as the change highlights, there's actually not that much to one of these orchestrations now.


ekzhu:
I really like the direction this is going. We should definitely align the orchestration interface with agent interface. I think we may need to revisit the implementation as we iterate on the workflow <--> runtime interface, and may change the runtime interface.

dotnet/samples/GettingStarted/Orchestration/ConcurrentOrchestration_With_StructuredOutput.cs
```diff
@@ -32,21 +32,16 @@ public async Task RunOrchestrationAsync()
...
 
         // Run the orchestration
         const string resourceId = "Hamlet_full_play_summary.txt";
         string input = Resources.Read(resourceId);
         Console.WriteLine($"\n# INPUT: @{resourceId}\n");
-        OrchestrationResult<Analysis> result = await orchestration.InvokeAsync(input);
+        AgentRunResponse result = await orchestration.RunAsync(input);
 
-        Analysis output = await result;
+        StructuredOutputTransform<Analysis> outputTransform = new(this.CreateChatClient());
```
TaoChenOSU: One of the initial requirements we got for the orchestrations was to have them spit out structured data directly. It was primarily for the benefits of readability and the debuggability for nested orchestrations, which were never realized. I don't think we have gotten too much feedback on the feature. That being said, I think it's ok to drop it. My question is if we should still maintain this sample and others that show case the structured outputs capability?

dotnet/samples/GettingStarted/Orchestration/GroupChatOrchestration2_Intro.cs
```diff
@@ -0,0 +1,78 @@
...
+                LoggerFactory = this.LoggerFactory,
+                ResponseCallback = monitor.ResponseCallback,
+                StreamingResponseCallback = streamedResponse ? monitor.StreamingResultCallback : null,
+            };
+
+        string input = "Create a slogan for a new electric SUV that is affordable and fun to drive.";
+        Console.WriteLine($"\n# INPUT: {input}\n");
+        AgentRunResponse result = await orchestration.RunAsync(input);
+
+        Console.WriteLine($"\n# RESULT:\n{string.Join("\n\n", result.Messages.Select(r => $"{r.Text}"))}");
```
TaoChenOSU: How is this sample different than the `GroupChatOrchestration_Intro.cs` except this line?

dotnet/src/Microsoft.Agents.Orchestration/GroupChat/GroupChatOrchestration.cs
```diff
@@ -1,91 +1,134 @@
...
+
+            // Run the selected agent with all messages (similar to how GroupChatAgentActor._cache works)
+            this.LogOrchestrationRunningAgent(context, nextAgent);
+            AgentRunResponse response = await RunAsync(nextAgent, context, allMessages, options: null, cancellationToken).ConfigureAwait(false);
+            this.LogOrchestrationAgentCompleted(context, nextAgent);
+
+            // Add the agent's response to the conversation
+            if (response.Messages.Count > 0)
+            {
+                allMessages.AddRange(response.Messages);
```
TaoChenOSU: One distinction between this approach and the one in SK is that not all messages from the agents are shared. In SK, only the "last" message from an agent, which is considered as the final response from the agent, is shared. It excludes the tool call/result messages that lead to the agent's final response.

We don't have data backing the effectiveness of the current approach in SK. The argument, however, is that hiding the agent's "thinking" processing will make the group chat more effective as it reduces context length.

Regardless of what we do, I think we should make it clear with comments and documentations.

dotnet/src/Microsoft.Agents.Orchestration/OrchestratingAgent.cs
```diff
@@ -0,0 +1,313 @@
...
+    /// <summary>
+    /// Runs the agent with input messages and respond with both streamed and regular messages.
+    /// </summary>
+    /// <param name="agent">The agent being run</param>
+    /// <param name="context">The associated orchestration context for this run.</param>
+    /// <param name="input">The list of chat messages to send.</param>
+    /// <param name="options">Options to use when invoking the agent.</param>
+    /// <param name="cancellationToken">A cancellation token that can be used to cancel the operation.</param>
+    /// <returns>A task that returns the response <see cref="ChatMessage"/>.</returns>
+    protected static async ValueTask<AgentRunResponse> RunAsync(AIAgent agent, OrchestratingAgentContext context, IReadOnlyCollection<ChatMessage> input, AgentRunOptions? options = null, CancellationToken cancellationToken = default)
```
TaoChenOSU, line 171: Is each invocation based on a new thread?

dotnet/src/Microsoft.Agents.Orchestration/Handoffs/HandoffOrchestration.cs
```diff
@@ -0,0 +1,227 @@
...
+                nextAgent = null;
+                await CheckpointAsync().ConfigureAwait(false);
+                this.LogOrchestrationAgentCompleted(context, agent);
+                break;
+            }
+
+            // Create the options for the next agent request, including handoff functions.
+            HandoffState handoffState = new(handoffs);
+            ChatClientAgentRunOptions? options = null;
+            List<AITool> handoffTools = handoffState.CreateHandoffFunctions(this.InteractiveCallback is not null);
```
TaoChenOSU: This is not yet addressed in SK, but I am just putting it here:

For agents that support parallel tool calls, it may try to hand off to multiple agents at the same time: https://github.com/microsoft/semantic-kernel/issues/12756. I don't have a good solution for this corner case yet.

stephentoub: > For agents that support parallel tool calls, it may try to hand off to multiple agents at the same time

Does it matter? Depending on the implementation, only one of them will take effect, and ultimately the model asked for whichever one happens (in addition to erroneously asking for more).

dotnet/src/Microsoft.Agents.Orchestration/OrchestratingAgent.cs
stephentoub: Right now, yes. It could be changed if that's desirable.

dotnet/src/Microsoft.Agents.Orchestration/GroupChat/GroupChatOrchestration.cs
stephentoub: > We don't have data backing the effectiveness of the current approach in SK. The argument, however, is that hiding the agent's "thinking" processing will make the group chat more effective as it reduces context length.

We can certainly change what I've done here if we want to say the last message is special. I do worry, though, that it's losing important data, especially in cases where the answer or part of it is not in the last message. I realize this example is a bit extreme, as I'm explicitly telling it to split things up, but when I run this code:
```C#
using OpenAI.Responses;
using System.ComponentModel;

IChatClient client = new OpenAIResponseClient("gpt-4.1", Environment.GetEnvironmentVariable("AI:OpenAI:ApiKey")!)
    .AsIChatClient()
    .AsBuilder()
    .UseFunctionInvocation()
    .Build();

var response = await client.GetStreamingResponseAsync(
    "First respond to me to tell me how old Alice is. Then respond to me to tell me how old Bob is.",
    new()
    {
        Tools = [AIFunctionFactory.Create([Description("The name of the person.")] (string name) =>
        {
            return name switch
            {
                "Alice" => "Alice is 30 years old.",
                "Bob" => "Bob is 25 years old.",
                _ => throw new ArgumentException("Unknown person")
            };
        }, "GetPersonAge")]
    }).ToChatResponseAsync();

foreach (var message in response.Messages)
{
    string text = message.Text;
    if (!string.IsNullOrWhiteSpace(text))
    {
        Console.WriteLine($"{message.MessageId}: {message.Text}");
    }
}
```
I get this:
```
msg_687e89776ef0819e844e3b8247fbe65401ea43f1a7683b09: I will first find out how old Alice is, and then I will let you know how old Bob is. Please wait a moment.
msg_687e89791274819eb39f89d46ada8ae201ea43f1a7683b09: Alice is 30 years old. Now, I will tell you how old Bob is.
msg_687e897a4b70819ebc18cb28557f258f01ea43f1a7683b09: Bob is 25 years old.
```
If we only used the last message, it would not include Alice's age.

dotnet/samples/GettingStarted/Orchestration/GroupChatOrchestration2_Intro.cs
stephentoub: I didn't actually look at what the examples were doing, just made them work. I can look more deeply and delete one if it's duplicative.

dotnet/samples/GettingStarted/Orchestration/ConcurrentOrchestration_With_StructuredOutput.cs
stephentoub: > That being said, I think it's ok to drop it. My question is if we should still maintain this sample and others that show case the structured outputs capability?

If we think we need that convenience layer, I think we should expose structured output as a helper on top of any Agent, not just orchestrations, e.g. we can have an extension method like this (pseudo-code/design):
```C#
public static async Task<T> RunAsync<T>(this Agent agent, IEnumerable<ChatMessage> messages, IChatClient chatClient)
{
    var response = await agent.RunAsync(input);
    var result = await chatClient.GetResponseAsync<T>([..messages, ..response.Messages], new()
    {
        Instructions = "Extract relevant data from the supplied messages and return it according to the provided JSON schema.",
    });
    return result.Result;
}
```

cc: @westey-m

dotnet/src/Microsoft.Agents.Orchestration/GroupChat/GroupChatOrchestration.cs
TaoChenOSU: Perhaps we can exclude messages that contain function calls/results? Does the first message contain text content and function call contents?

dotnet/src/Microsoft.Agents.Orchestration/Handoffs/HandoffOrchestration.cs
TaoChenOSU: Yes. Only one of them takes effect but the model will "think" all of them take effect.

stephentoub: > but the model will "think" all of them take effect.

Only if it's sent back the function call content... which is being deleted.

dotnet/src/Microsoft.Agents.Orchestration/GroupChat/GroupChatOrchestration.cs
stephentoub: > Does the first message contain text content and function call contents?

In this case, with Responses and gpt-4.1, it typically looks like this:
```
msg_687ee724c3c0819fbf179c0ab87c3aad044dfdaf75e60e94
        Microsoft.Extensions.AI.TextContent. First, I'll find out how old Alice is.

Let me check that for you.

fc_687ee72551cc819fbcd4b3a57f1bf974044dfdaf75e60e94
        Microsoft.Extensions.AI.FunctionCallContent.

b81380bbe6554b5ea79f608adf46b192
        Microsoft.Extensions.AI.FunctionResultContent.

msg_687ee7260000819f8675376c055b127e044dfdaf75e60e94
        Microsoft.Extensions.AI.TextContent. Alice is 30 years old.

Now, let me check how old Bob is.

fc_687ee7267e38819fab8a7b0e4d8f3d17044dfdaf75e60e94
        Microsoft.Extensions.AI.FunctionCallContent.

c5c1a47a6bc24140b4073e74c24e54d7
        Microsoft.Extensions.AI.FunctionResultContent.

msg_687ee7270590819fa69c3ef48744b124044dfdaf75e60e94
        Microsoft.Extensions.AI.TextContent. Bob is 25 years old.
```
but there are definitely cases where it will send both text content and function call requests in the same message.

> Perhaps we can exclude messages that contain function calls/results?

We have plenty of flexibility. We could exclude all function call-related content. We could just include a subset of messages. We could include all messages. Etc.

stephentoub:
@reubenbond, I've rebased this on your merged PR, wiring it up as I think we discussed and deleting the old runtime files. I also updated your sample to use a handoff orchestration.
image

markwallace-microsoft:
@stephentoub I chatted with Ben about this and we agreed that we want to merge this and then refactor this on top of process framework but maintain the developer UX and ideally the overall simplicity of the implementation. Having less code to refactor will ultimately save us time later.

dotnet/src/Microsoft.Agents.Orchestration/OrchestratingAgentContext.cs
```diff
@@ -0,0 +1,54 @@
...
+using Microsoft.Extensions.AI.Agents.Runtime;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+
+namespace Microsoft.Agents.Orchestration;
+
+/// <summary>
+/// Provides contextual information for an orchestration operation, including logging, and response callback.
+/// </summary>
+public sealed class OrchestratingAgentContext
```
markwallace-microsoft, line 14: Curious as to why the properties are mutable?

stephentoub: ETag needed to be, as it gets updated on every checkpointing operation. I made the others mutable just for consistency with that. I could make them init instead if that's preferable. To remove the setter entirely, I'd need a constructor, which is what was there previously, but the ctor was internal, which makes it harder for others to test, and making the ctor public then makes it harder to extend with additional members in the future. I don't have a strong opinion about this, though. For now I could add back the internal ctor or make the setters (other than ETag) internal.

PR #201, opened 2025-07-18
Rename Agent to AIAgent and add DisplayName prop

westey-m:
### Motivation and Context

To avoid naming clashes we are renaming Agent to AIAgent.
This matches naming in MEAI.

### Description

- Rename Agent to AIAgent
- Add new DisplayName property on AIAgent that returns Name and falls back to ID.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIAgent.cs
+++ b/dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIAgent.cs
@@ -12,7 +12,7 @@ namespace Microsoft.Extensions.AI.Agents;
 /// Base abstraction for all agents. An agent instance may participate in one or more conversations.
 /// A conversation may include one or more agents.
 /// </summary>
-public abstract class Agent
+public abstract class AIAgent
 {
     /// <summary>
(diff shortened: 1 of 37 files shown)


dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIAgent.cs
```diff
@@ -27,6 +27,14 @@ public abstract class Agent
     /// </summary>
     public virtual string? Name { get; }
 
+    /// <summary>
+    /// Gets a display name for the agent, which is either the <see cref="Name"/> or <see cref="Id"/> if the name is not set.
+    /// </summary>
+    public virtual string DisplayName
```
rogerbarreto, line 33: Suggest changing the `LoggingAgentName()` to use this property as well. maybe removing the extension class altogether.

westey-m: LoggingAgentName doesn't currently do the same thing as DisplayName.  It doesn't fall back to ID, and I think that's the right thing, since there is a separate ID field for logging already.

PR #206, opened 2025-07-21
Python: move all tests under tests and initial work on int tests

### LINKED ISSUE #143: Enable integration test runs for Python
Integration tests should run when PR's are merged to main

eavanvalkenburg:
### Motivation and Context

Moves all tests under tests, instead of seperation between unit and integration
Adapts current monkeypath to deal with this
Adds the two markers for azure and openai

Closes #143

### Contribution Checklist

- [x] The PR follows the [SK Contribution Guidelines](https://github.com/microsoft/agent-framework/blob/main/CONTRIBUTING.md)


.github/workflows/label-issues.yml
```diff
@@ -0,0 +1,48 @@
...
+            let title = context.payload.issue.title
+
+            // Define the labels array
+            let labels = ["triage"]
+
+            // Check if the body or the title contains the word 'python' (case-insensitive)
+            if ((body != null && body.match(/python/i)) || (title != null && title.match(/python/i))) {
+              // Add the 'python' label to the array
+              labels.push("python")
+            }/Users/edvan/Work/semantic-kernel/.github/workflows/label-pr.yml
```
dmytrostruk: What does this line mean?

.github/workflows/python-merge-tests.yml
```diff
@@ -0,0 +1,172 @@
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: true
+      matrix:
+        python-version: ["3.10"]
+        os: [ubuntu-latest]
+    env:
+      UV_PYTHON: ${{ matrix.python-version }}
+      AZURE_OPENAI_CHAT_DEPLOYMENT_NAME: ${{ vars.AZUREAI__DEPLOYMENTNAME }}
+      AZURE_OPENAI_ENDPOINT: ${{ secrets.AZUREAI__ENDPOINT }}
```
dmytrostruk: Can we use FOUNDRY prefix for these settings? 
- FOUNDRY_PROJECT_ENDPOINT
- FOUNDRY_MODEL_DEPLOYMENT_NAME

.github/workflows/python-merge-tests.yml
```diff
@@ -0,0 +1,172 @@
...
+        if: always()
+        uses: pmeier/pytest-results-action@v0.7.2
+        with:
+          path: ./python/**.xml
+          summary: true
+          display-options: fEX
+          fail-on-empty: true
+          title: Test results
+
+  python-tests-azure:
```
dmytrostruk, line 70: Is it necessary to have a dedicated step for each package? Because every time we add a new package, we have to modify this file, which is not a big problem, but I was wondering if new packages can be automatically included into pipeline if possible.

.github/workflows/python-tests.yml
```diff
@@ -0,0 +1,111 @@
+name: Python - Tests - No Env
```
dmytrostruk: Are these unit tests?

python/packages/azure/tests/test_azure_chat_client.py
```diff
@@ -32,6 +36,12 @@
 
 # region Service Setup
 
+skip_if_no_real_azure_endpoint = pytest.mark.skipif(
```
dmytrostruk: nit: Since it also depends on `RUN_INTEGRATION_TESTS` boolean, I would rename it to something like:
- `skip_unless_valid_integration_config`
- `skip_if_no_valid_integration_config`

Same in other files.

python/packages/azure/tests/test_azure_chat_client.py
```diff
@@ -618,3 +628,123 @@ async def test_cmc_streaming(
...
+    """Returns a story about Emily and David."""
+    return (
+        "Emily and David, two passionate scientists, met during a research expedition to Antarctica. "
+        "Bonded by their love for the natural world and shared curiosity, they uncovered a "
+        "groundbreaking phenomenon in glaciology that could potentially reshape our understanding "
+        "of climate change."
+    )
+
+
+@skip_if_no_real_azure_endpoint
```
dmytrostruk: Taking into account new integration tests for Foundry, should we do the same for Foundry tests?

.github/workflows/label-issues.yml
eavanvalkenburg: a mistake, removed

.github/workflows/python-merge-tests.yml
eavanvalkenburg: since each package will have slightly different setup, in terms of env settings and potentially extra steps like for Auth, which I still need to add, we will need these to be seperate

.github/workflows/python-tests.yml
eavanvalkenburg: yes, but since we don't distinguish between unit and integration other then at runtime, I don't want to call them that anymore, I'm open to other names...

python/packages/azure/tests/test_azure_chat_client.py
eavanvalkenburg: yeah, added now

PR #210, opened 2025-07-22
.Net: FileSearch Tool Sample + API Proposal Updates for MEAI.Abstractions/Azure.AI.Agents

### LINKED ISSUE #139: .Net: Add File Search Sample
## Add File Search

Similar to SK, add a samples for File Search that can be reused across different providers (AzureAIAgent / OpenAIAssistant / Bedrock).

rogerbarreto:
# Motivation and Context

- Resolves #139 

This pull introduce important updates including the addition of a new file search tool abstraction, updates to support tool-specific resources in chat clients, and a new example demonstrating file search functionality. Below is a breakdown of the most important changes:

### File Search Tool Enhancements
* Added `NewHostedFileSearchTool` class in `MEAI.Abstractions` to standardize file search functionality across providers. This includes support for managing vector store IDs and additional properties.

### Chat Client Updates
* Updated `NewOpenAIAssistantChatClient` to support tool-specific resources:
  - Introduced `CreateToolResources` method to handle resources for tools like file search and code interpreter.
  - Modified `GetStreamingResponseAsync` to include `ToolResources` in thread creation options.
  - Enhanced `ToOpenAIAssistantsFunctionToolDefinition` to handle file search tool definitions and their associated file IDs.

### New Example for File Search
* Added `Step04_ChatClientAgent_UsingFileSearchTools` example to demonstrate uploading files, creating vector stores, and utilizing file search tools for document analysis.

### Code Refinements
* Updated method signatures and logic in `Step03_ChatClientAgent_UsingCodeInterpreterTools` to fix a bug handling `RawRepresentation` in agent streaming updates.

### Minor Updates
* Adjusted `GetAzureAIAgentPersistentClient` in `AgentSample.cs` to use `NewPersistentAgentsChatClient` for improved compatibility with updated tool integrations.


dotnet/samples/GettingStarted/External/Azure.AI.Agents.Persistent/NewPersistentAgentsChatClient.cs
rogerbarreto, line 1: For development celerity, I'm bringing back the source code from `PersistentChatClient` that are dependent on MEAI abstractions changes, until we can push or create a PR to their repo using the decided/published MEAI API.

dotnet/samples/GettingStarted/External/MEAI.OpenAI/NewOpenAIAssistantChatClient.cs
```diff
@@ -89,6 +89,63 @@ public Task<ChatResponse> GetResponseAsync(
...
+                        break;
+
+                    case NewHostedFileSearchTool fileSearchTool:
+
+                        // Handle file IDs for file search tool
+                        if (fileSearchTool.AdditionalProperties.TryGetValue("vectorStoreIds", out object? vectorStoreIds) && vectorStoreIds is IEnumerable<string> enumerableVectorStoreIds)
+                        {
+                            fileSearchResources ??= new();
+
+                            foreach (var vectorStoreId in enumerableVectorStoreIds)
```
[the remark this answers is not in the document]
rogerbarreto: This cant be done as there's already a veriable using `vectorStoreIds`.

dotnet/samples/GettingStarted/External/MEAI.Abstractions/NewHostedFileSearchTool.cs
```diff
@@ -0,0 +1,35 @@
...
+/// This provides a standardized interface for file search functionality across providers.
+/// </summary>
+public class NewHostedFileSearchTool : AITool
+{
+    // Usage of an internal dictionary is temporary and only used here because the MEAI.Abstractions does not have this specialization yet and the
+    // ChatClients must rely on the AdditionalProperties to check and set correctly the File Search Resource avoiding a customized RawRepresentationFactory implementation.
+    private readonly Dictionary<string, object?> _additionalProperties = [];
+
+    /// <summary>Gets or sets the list of vector store IDs that the file search tool can access.</summary>
+    public IList<string> VectorStoreIds
```
stephentoub: Why `string` rather than `AIContent`? The argument made with CodeInterpreterTool was that different forms of AIContent could be used, like a HostedFileContent, a HostedVectorStoreContent, a DataContent when the data is included in the request, etc.

dotnet/samples/GettingStarted/External/MEAI.Abstractions/NewHostedFileSearchTool.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        get
+        {
+            // Only create the property in the dictionary when it is actually used
+            if (!this._additionalProperties.TryGetValue("vectorStoreIds", out var value) || value is null)
```
stephentoub: Why are you using AdditionalProperties?

rogerbarreto: Good point, before ChatClient proposals were in a different project, that was required. Since they are sharing the GettingStarted, I can use directly the property. Will update.

dotnet/samples/GettingStarted/External/MEAI.Abstractions/NewHostedFileSearchTool.cs
rogerbarreto: Yeah, I will take this opportunity to update to the latest considerations we discussed in the MEAI repo.

stephentoub: Ok, let me know when I should take another look.

dotnet/samples/GettingStarted/External/MEAI.Abstractions/HostedFileContent.cs
```diff
@@ -0,0 +1,41 @@
...
+/// Represents a file that is hosted by the AI service.
+/// </summary>
+/// <remarks>
+/// Unlike <see cref="DataContent"/> which contains the data for a file or blob, this class represents a file that is hosted
+/// by the AI service and referenced by an identifier. Such identifiers are specific to the provider.
+/// </remarks>
+[DebuggerDisplay("FileId = {FileId}")]
+public sealed class HostedFileContent : AIContent
+{
+    private string? _fileId;
```
markwallace-microsoft: Does this need to be optional given the only way to construct a `HostedFileContent` is by providing a valid `fileId`?

dotnet/samples/GettingStarted/External/MEAI.OpenAI/NewOpenAIAssistantChatClient.cs
```diff
@@ -80,6 +80,74 @@ public Task<ChatResponse> GetResponseAsync(
...
+            // The caller can provide tools in the supplied ThreadAndRunOptions. Augment it with any supplied via ChatOptions.Tools.
+            foreach (AITool tool in tools)
+            {
+                switch (tool)
+                {
+                    case NewHostedCodeInterpreterTool codeTool:
+
+                        if (codeTool.Inputs is { Count: > 0 })
+                        {
+                            codeInterpreterResources ??= new();
```
markwallace-microsoft, line 103: Can `codeInterpreterResources` be anything other than null here?

dotnet/samples/GettingStarted/External/MEAI.OpenAI/NewOpenAIAssistantChatClient.cs
```diff
@@ -80,6 +80,74 @@ public Task<ChatResponse> GetResponseAsync(
...
+                        }
+
+                        break;
+
+                    case NewHostedFileSearchTool fileSearchTool:
+
+                        // Handle file IDs for file search tool
+                        if (fileSearchTool.Inputs is { Count: > 0 })
+                        {
+                            fileSearchResources ??= new();
```
markwallace-microsoft, line 123: Can this be non null here?

dotnet/samples/GettingStarted/Steps/Step03_ChatClientAgent_UsingCodeInterpreterTools.cs
```diff
@@ -51,9 +53,9 @@ public async Task RunningWithFileReferenceAsync(ChatClientProviders provider)
                 assistantOutput.Append(update.Text);
             }
 
-            if (update.RawRepresentation is not null)
+            if (update.RawRepresentation is ChatResponseUpdate chatUpdate && chatUpdate.RawRepresentation is not null)
             {
-                codeInterpreterOutput.Append(GetCodeInterpreterOutput(update.RawRepresentation, provider));
+                codeInterpreterOutput.Append(GetCodeInterpreterOutput(chatUpdate.RawRepresentation, provider));
```
markwallace-microsoft, line 58: Can we create an extension method here `chatUpdate.GetCodeInterpreterOutput()`?

dotnet/samples/GettingStarted/External/MEAI.Abstractions/HostedFileContent.cs
```diff
@@ -0,0 +1,41 @@
...
+
+/// <summary>
+/// Represents a file that is hosted by the AI service.
+/// </summary>
+/// <remarks>
+/// Unlike <see cref="DataContent"/> which contains the data for a file or blob, this class represents a file that is hosted
+/// by the AI service and referenced by an identifier. Such identifiers are specific to the provider.
+/// </remarks>
+[DebuggerDisplay("FileId = {FileId}")]
+public sealed class HostedFileContent : AIContent
```
westey-m, line 16: We also have [UriContent](https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/UriContent.cs). Is the fileId typically a Uri, or just an ID?

dotnet/samples/GettingStarted/External/MEAI.Abstractions/HostedVectorStoreContent.cs
```diff
@@ -0,0 +1,41 @@
...
+namespace Microsoft.Extensions.AI;
+
+/// <summary>
+/// Represents a file that is hosted by the AI service.
+/// </summary>
+/// <remarks>
+/// Unlike <see cref="DataContent"/> which contains the data for a file or blob, this class represents a file that is hosted
+/// by the AI service and referenced by an identifier. Such identifiers are specific to the provider.
+/// </remarks>
+[DebuggerDisplay("FileId = {FileId}")]
```
westey-m:
```suggestion
[DebuggerDisplay("VectorStoreId= {VectorStoreId}")]
```

dotnet/samples/GettingStarted/External/MEAI.Abstractions/HostedVectorStoreContent.cs
```diff
@@ -0,0 +1,41 @@
...
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>
+/// Represents a file that is hosted by the AI service.
+/// </summary>
+/// <remarks>
+/// Unlike <see cref="DataContent"/> which contains the data for a file or blob, this class represents a file that is hosted
+/// by the AI service and referenced by an identifier. Such identifiers are specific to the provider.
+/// </remarks>
```
westey-m: Needs to be updated for vector store.

dotnet/samples/GettingStarted/External/MEAI.OpenAI/NewOpenAIAssistantChatClient.cs
rogerbarreto: No, In the current logic if I initialize this variable, it will always flag to the backend that I want to use code interpreter resource, I'm just doing there when I have this `codeTool` specified in the options.

dotnet/samples/GettingStarted/External/MEAI.OpenAI/NewOpenAIAssistantChatClient.cs
rogerbarreto: Not for the same reason above, to only flag the resource to be used when the search tool is provided.

dotnet/samples/GettingStarted/External/MEAI.Abstractions/HostedFileContent.cs
rogerbarreto: Agree, passed your observation forward in the proposal here.

- https://github.com/dotnet/extensions/pull/6620#discussion_r2228725241

dotnet/samples/GettingStarted/External/MEAI.Abstractions/HostedFileContent.cs
rogerbarreto: This is an ID, very specific to different cloud providers.

dotnet/samples/GettingStarted/Steps/Step03_ChatClientAgent_UsingCodeInterpreterTools.cs
rogerbarreto: I'm afraid not, because this is provider's specific.

dotnet/samples/GettingStarted/External/MEAI.Abstractions/NewHostedFileSearchTool.cs
```diff
@@ -0,0 +1,18 @@
...
+/// Proposal for abstraction updates based on the common file search tool properties.
+/// This provides a standardized interface for file search functionality across providers.
+/// </summary>
+public class NewHostedFileSearchTool : AITool
+{
+    /// <summary>Gets or sets a collection of <see cref="AIContent"/> to be used as input to the code interpreter tool.</summary>
+    /// <remarks>
+    /// Services support different varied kinds of inputs. Most support the IDs of vector stores that are hosted by the service,
+    /// represented via <see cref="HostedVectorStoreContent"/>. Some also support binary data, represented via <see cref="DataContent"/>.
+    /// Unsupported inputs will be ignored by the <see cref="IChatClient"/> to which the tool is passed.
```
westey-m, line 15: Have you considered throwing instead of ignoring?  Since this is not typically used via the agent abstraction but via the agent constructor, the developer should know which type of service they are using, and therefore should be passing supported types.  If they didn't, I expect it to be a bug.  Throwing would help them identify the bug during testing.

dotnet/samples/GettingStarted/Steps/Step03_ChatClientAgent_UsingCodeInterpreterTools.cs
westey-m: Shouldn't we have an AIContent type for the code interpreter output, which the different raw representations are mapped to by the ChatClient?

dotnet/samples/GettingStarted/External/MEAI.Abstractions/NewHostedFileSearchTool.cs
stephentoub: > Have you considered throwing instead of ignoring?

There are lots of options that are desirable to have at the abstraction level because lots of providers support it, but not all, so in general with M.E.AI we've adopted a "the provider uses what it can and ignores everything else" attitude, e.g. if a developer sets ChatOptions.TopK, we don't throw when using OpenAI, we just ignore it. If a developer includes a FooBarContent in their request and a particular provider doesn't know how to translate FooBarContent, it's just ignored. If a developer includes a HostedWebSearchTool in their request, and the service isn't currently capable of invoking such a tool, it's ignored, just as functions may be entirely ignored by a service/model that _is_ capable of doing function invocation. I think this falls into the same camp.

> Since this is not typically used via the agent abstraction but via the agent constructor

This is applicable per request, too. If separately we want an agent constructor to throw if it doesn't support a particular tool, we could try, but I'm not convinced that's a good direction.

dotnet/samples/GettingStarted/External/MEAI.Abstractions/NewHostedFileSearchTool.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>
+/// Proposal for abstraction updates based on the common file search tool properties.
+/// This provides a standardized interface for file search functionality across providers.
+/// </summary>
+public class NewHostedFileSearchTool : AITool
```
stephentoub, line 9: Various file search tools have various configuration options that might need to be exposed here?

OpenAI Responses' has filters, max numbers of results, and ranking options, and Assistants has max results and rankings.

AWS Bedrock has filters, number of results, search strategy, ranking.

Mistral's document_library tool appears to currently only support the list of ids.

Gemini's support seems to be nascent but likely coming.

Is such a HostedFileSearchTool going to be meaningfully usable without various knobs?

dotnet/samples/GettingStarted/External/MEAI.Abstractions/NewHostedFileSearchTool.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    /// <summary>Gets or sets a collection of <see cref="AIContent"/> to be used as input to the code interpreter tool.</summary>
+    /// <remarks>
+    /// Services support different varied kinds of inputs. Most support the IDs of vector stores that are hosted by the service,
+    /// represented via <see cref="HostedVectorStoreContent"/>. Some also support binary data, represented via <see cref="DataContent"/>.
+    /// Unsupported inputs will be ignored by the <see cref="IChatClient"/> to which the tool is passed.
+    /// </remarks>
+    public IList<AIContent>? Inputs { get; set; }
+}
```
stephentoub, line 18: Do we need to represent output generated by such a tool? e.g. some kind of message or content indicating that the tool was called and with what input, some kind of message or content indicating any results it produced if that's enabled, etc.?

dotnet/samples/GettingStarted/External/MEAI.Abstractions/NewHostedFileSearchTool.cs
rogerbarreto: > Is such a HostedFileSearchTool going to be meaningfully usable without various knobs? 

My original idea is, to be fast, that we already see the meaning having this abstraction in place with `FileStores`.

We can always add options/knobs as we go and we see fit, I'm happy to do a deeper investigation to add more knobs. Will create an issue to investigate this options further.

> Various file search tools have various configuration options that might need to be exposed here?
> 
> OpenAI Responses' has filters, max numbers of results, and ranking options, and Assistants has max results and rankings.
> 
> AWS Bedrock has filters, number of results, search strategy, ranking.
> 
> Mistral's document_library tool appears to currently only support the list of ids.
> 
> Gemini's support seems to be nascent but likely coming.
> 

Sounds very reasonable to have extra knobs definitely for `Ranking` and `Max Results`.

dotnet/samples/GettingStarted/External/MEAI.Abstractions/NewHostedFileSearchTool.cs
rogerbarreto: Analogous to the proposal in the Code Interpreter I see room to represent tooling output for search.

OpenAI does have an option for it.

- [Open AI Responses - include=["file_search_call.results"]](https://platform.openai.com/docs/guides/tools-file-search#include-search-results-in-the-response)
- [OpenAI Assistants - include=["step_details.tool_calls[*].file_search.results[*].content"]](https://platform.openai.com/docs/assistants/tools/file-search#improve-file-search-result-relevance-with-chunk-ranking)
- Mistral / Google / Amazon I coudn't figure (thru docs)

stephentoub: Yup! Anthropic provides web search results as well.
https://docs.anthropic.com/en/docs/agents-and-tools/tool-use/web-search-tool

dotnet/samples/GettingStarted/Steps/Step03_ChatClientAgent_UsingCodeInterpreterTools.cs
rogerbarreto: This was [same question raised by @stephentoub](https://github.com/microsoft/agent-framework/pull/210#discussion_r2229115980). 

We will consider the output for the original proposal in the MEAI repo.

- https://github.com/dotnet/extensions/pull/6620

dotnet/samples/GettingStarted/External/MEAI.Abstractions/NewHostedFileSearchTool.cs
rogerbarreto: Important to note that this `Outputs` here has a complete different meaning as the `Inputs`, basically the `Outputs` is not a `Content` but rather a configuration provided to the tool how to respond.

This might not be very well represented by a list of `AIContent`, and we might confuse the audience on why Output something while configuring the tool.

For this specific case, the knob can be simple, `bool? GenerateOutputs { get; set; }`.

PR #211, opened 2025-07-22
.Net: Add missing providers to tool samples

rogerbarreto:
This pull request enhances the functionality of the `Step02_ChatClientAgent_UsingFunctionTools` class by adding support for new chat client providers and improving the handling of server-side agents. The key changes include expanding the list of supported providers, introducing server-side agent creation and cleanup, and refactoring the agent initialization process for better modularity and clarity.


dotnet/samples/GettingStarted/Steps/Step02_ChatClientAgent_UsingFunctionTools.cs
```diff
@@ -14,28 +14,46 @@ public sealed class Step02_ChatClientAgent_UsingFunctionTools(ITestOutputHelper
 {
     [Theory]
     [InlineData(ChatClientProviders.AzureOpenAI)]
+    [InlineData(ChatClientProviders.AzureAIAgentsPersistent)]
+    [InlineData(ChatClientProviders.OpenAIAssistant)]
     [InlineData(ChatClientProviders.OpenAIChatCompletion)]
+    [InlineData(ChatClientProviders.OpenAIResponses)]
     public async Task RunningWithToolsBasic(ChatClientProviders provider)
```
westey-m: The point of this sample is to show a really basic 4 line case to show how simple function invocation can be.
Unfortunately with some underlying services the basic case is just not possible.  I don't think we should make the basic sample more complicated to support the other services though, we should just say that they are not supported in this simple case.

westey-m: The more complex samples below are fine to feature all underlying services though.

rogerbarreto: > Unfortunately with some underlying services the basic case is just not possible. I don't think we should make the basic sample more complicated

I'm a bit thorn, as the whole point of the Framework is to be abstract and cross provider compatible, if part of our abstractions only work for X, Y and Z, it needs to be re-considered.

> I don't think we should make the basic sample more complicated to support the other services though, we should just say that they are not supported in this simple case.

The sample should be unique and similar across all, and we may need to add the `agentId` to the ChatClientAgent/Agent ctor, while keeping the original simplicity.

rogerbarreto: Actually I already see an issue in this attempt (adding the Id), this won't change much our samples as the `Managed Agent Providers` expect instructions and tools to be provided ahead of time, and that's handy using the `ChatCLientAgentOptions`.

westey-m: If we have to show service agents everywhere I'd say you might as well just remove this sample, since it becomes just as complex as the others then.

PR #217, opened 2025-07-22
Python: Added fixes and more examples for Foundry agent

dmytrostruk:
### Motivation and Context

Fixed following issues:
- `AIProjectClient` and `AsyncTokenCredential` resource management - close instances if these resources were created automatically.
- Pass `conversation_id` to `ChatOptions` when invoking `ChatClient` from `Agent`.

Added following examples:
- Example with thread management.
- Example with function tools both on agent and run levels.
- Example with code interpreter.

### Contribution Checklist


python/packages/foundry/agent_framework_foundry/_chat_client.py
```diff
@@ -145,17 +152,22 @@ def __init__(
...
             thread_id=thread_id,  # type: ignore[reportCallIssue]
             **kwargs,
         )
 
         self._should_delete_agent = False
+        self._should_close_client = should_close_client
+        self._should_close_credential = should_close_credential
         self._foundry_settings = foundry_settings
 
     async def __aenter__(self) -> "FoundryChatClient":
```
eavanvalkenburg:
```suggestion
    async def __aenter__(self) -> Self:
```

python/packages/foundry/agent_framework_foundry/_chat_client.py
```diff
@@ -455,14 +479,15 @@ def _create_run_options(
             run_options["parallel_tool_calls"] = chat_options.allow_multiple_tool_calls
 
             if chat_options.tools is not None:
```
eavanvalkenburg: we should rewrite this to use the self._prepare_tools... function, because the function calling decorator also leverages that, so now we touch this twice (which was the issue with the old setup anyway), that's also why I added the TODO.

python/packages/main/agent_framework/_tools.py
```diff
@@ -159,3 +159,35 @@ def wrapper(f: Callable[..., ReturnT | Awaitable[ReturnT]]) -> AIFunction[Any, R
...
+        """
+        self.name = name
+        self.description = description
+        self.additional_properties = additional_properties
+
+    def __str__(self) -> str:
+        """Return a string representation of the tool."""
+        return f"HostedCodeInterpreterTool(name={self.name})"
+
+    def parameters(self) -> dict[str, Any]:
```
eavanvalkenburg: I think we can remove this both here and in AITool, because when I made this I didn't understand the purpose of the additional_properties so I renamed it to parameters, so let's delete here and in AITool!

python/samples/getting_started/agents/foundry/foundry_with_code_interpreter.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+async def main() -> None:
+    """Example showing how to use the HostedCodeInterpreterTool with Foundry."""
+    print("=== Foundry Chat Client with Code Interpreter Example ===")
+
+    async with ChatClientAgent(
+        chat_client=FoundryChatClient(),
+        instructions="You are a helpful assistant that can write and execute Python code to solve problems.",
+        tools=HostedCodeInterpreterTool(),
```
eavanvalkenburg, line 41: if this is going to be without params more often, then we should think about allowing `HostedCodeInterpreterTool` or even `"HostedCodeInterpreterTool"`...

dmytrostruk: I'm not sure about this, because as for now, this is without parameters, but a lot of providers support things like file references for code interpreter, so this abstraction is going to be extended in the future.

python/packages/foundry/agent_framework_foundry/_chat_client.py
```diff
@@ -455,14 +485,15 @@ def _create_run_options(
...
-                # TODO (eavanvalkenburg): replace with _prepare_tools_and_tool_choice overload
                 tool_definitions: list[MutableMapping[str, Any]] = []
 
                 for tool in chat_options.tools:
```
eavanvalkenburg: should we check if tool_choice == None and ifso keep tool_definitions empty?

python/packages/main/agent_framework/_tools.py
```diff
@@ -159,3 +155,32 @@ def wrapper(f: Callable[..., ReturnT | Awaitable[ReturnT]]) -> AIFunction[Any, R
...
+        name: str = "code_interpreter",
+        description: str | None = None,
+        additional_properties: dict[str, Any] | None = None,
+    ):
+        """Initialize a HostedCodeInterpreterTool.
+
+        Args:
+            name: The name of the tool. Defaults to "code_interpreter".
+            description: A description of the tool.
+            additional_properties: Additional properties associated with the tool.
```
eavanvalkenburg:
```suggestion
            additional_properties: Additional properties associated with the tool, specific to the service used.
```

python/packages/main/agent_framework/_tools.py
```diff
@@ -159,3 +155,32 @@ def wrapper(f: Callable[..., ReturnT | Awaitable[ReturnT]]) -> AIFunction[Any, R
         return wrapper(func)
 
     return decorator(func) if func else decorator  # type: ignore[reportReturnType, return-value]
+
+
+class HostedCodeInterpreterTool(AITool):
```
eavanvalkenburg, line 160:
```suggestion
class HostedCodeInterpreterTool(AFBaseModel, AITool):
```
and then just add the attributes, like all pydantic models, should shorten the code.

dmytrostruk: Using `AFBaseModel` is giving me this error because `HostedCodeInterpreterTool` also implements AITool protocol:
```
metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
```

Issue #113, opened 2025-06-30
Create release pipeline Python


markwallace-microsoft:
- [  ] Build and test
- [  ] Security checks
- [  ] Code signing
- [  ] Publish to pypi.org

PR #219, opened 2025-07-23
updated builds and release pipeline setup (without pypi publish)

eavanvalkenburg:
### Motivation and Context

Adds release action for python packages, accepts:
`python-<package short name>-<version> as release tags and then creates the relevant dist package
and then adds that package to the GH release, for public preview, we should then start the ADO job that uses the ESRP process.

Also updated readme's

Closes: #113 

### Contribution Checklist

### DIFF
--- a/.github/workflows/python-merge-tests.yml
+++ b/.github/workflows/python-merge-tests.yml
@@ -13,6 +13,7 @@ env:
   # Configure a constant location for the uv cache
   UV_CACHE_DIR: /tmp/.uv-cache
   RUN_INTEGRATION_TESTS: "true"
+  UV_SEMVER: "0.8.x"
 
 jobs:
   python-tests-main:
@@ -38,7 +39,7 @@ jobs:
       - name: Set up uv
         uses: astral-sh/setup-uv@v6
         with:
-          version: "0.8.x"
+          version: ${{ env.UV_SEMVER }}
           enable-cache: true
           cache-suffix: ${{ runner.os }}-${{ matrix.python-version }}
           cache-dependency-glob: "**/uv.lock"
@@ -91,7 +92,7 @@ jobs:
       - name: Set up uv
         uses: astral-sh/setup-uv@v6
         with:
-          version: "0.8.x"
+          version: ${{ env.UV_SEMVER }}
           enable-cache: true
           cache-suffix: ${{ runner.os }}-${{ matrix.python-version }}
           cache-dependency-glob: "**/uv.lock"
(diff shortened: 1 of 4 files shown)


.github/workflows/python-merge-tests.yml
```diff
@@ -38,7 +38,7 @@ jobs:
       - name: Set up uv
         uses: astral-sh/setup-uv@v6
         with:
-          version: "0.7.x"
+          version: "0.8.x"
```
peterychang: It seems like we need to update this in a lot of places. Any way we can make this a variable so we can just change it in one place?

python/packages/azure/pyproject.toml
```diff
@@ -24,44 +24,7 @@ classifiers = [
...
-    "autodoc_pydantic~=2.2",
-    "pygments",
-    "sphinxext-rediraffe",
-    "opentelemetry-instrumentation-openai",
-    "markdown-it-py[linkify]",
-    # Documentation tooling
-    "diskcache",
-    "redis",
-    "sphinx-autobuild",
+    "azure-identity >= 1.13"
```
peterychang, line 27: we might want to continue to depend on openai since the package references it directly

.github/workflows/python-merge-tests.yml
eavanvalkenburg: good idea, done

python/packages/azure/pyproject.toml
eavanvalkenburg: but Agent Framework already depends on it, so it will get installed through that and otherwise we would have to maintain two versions

README.md
```diff
@@ -1,33 +1,235 @@
...
-- Remove this section from the README
+```bash
+pip install agent-framework
+# Optional: Add Azure integration
+pip install agent-framework[azure]
+# Optional: Add Foundry integration
+pip install agent-framework[foundry]
+# Optional: Both
+pip install agent-framework[azure,foundry]
+```
```
dmytrostruk, line 22: Should we put this README in `python` folder? Because it's specific to Python, while we also need similar documentation for .NET. But if we list all of the information about Python and .NET in the same root file with examples, it can be a little bit overwhelming. I would keep just a general information in root README with a reference to dedicated Python and .NET README files that are located in respective folders. 

UPD: I see this content is already in `python/packages/main` folder, so should we remove it from root README?

eavanvalkenburg: unfortunately we need this here in the main folder and it is duplicate from the one in the python root, the python packaging works by copying the main folder into a temp location so it can't refer to a file in a different folder or a link to something else, but it is shown on the Pypi page so quite important, the one in root is important for people who visit the repo, so hence twice the same content...

PR #225, opened 2025-07-23
Add some OpenAI and Foundry extension methods

markwallace-microsoft:
### Motivation and Context

Proposed extension methods to make it easier to create an `AIAgent` instance from an `OpenAIClient` or `PersistentAgentsClient`.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Microsoft.Extensions.AI.Agents.OpenAI/OpenAIAssistantClientExtensions.cs
+++ b/dotnet/src/Microsoft.Extensions.AI.Agents.OpenAI/OpenAIAssistantClientExtensions.cs
@@ -24,13 +24,13 @@ public static class OpenAIAssistantClientExtensions
     /// Creates an AI agent from an <see cref="AssistantClient"/> using the OpenAI Assistant API.
     /// </summary>
     /// <param name="client">The OpenAI <see cref="AssistantClient" /> to use for the agent.</param>
-    /// <param name="model">The model identifier to use for responses (e.g., "gpt-4", "gpt-3.5-turbo").</param>
+    /// <param name="model">The model identifier to use (e.g., "gpt-4").</param>
     /// <param name="instructions">Optional system instructions that define the agent's behavior and personality.</param>
     /// <param name="name">Optional name for the agent for identification purposes.</param>
     /// <param name="description">Optional description of the agent's capabilities and purpose.</param>
     /// <param name="tools">Optional collection of AI tools that the agent can use during conversations.</param>
     /// <param name="loggerFactory">Optional logger factory for enabling logging within the agent.</param>
-    /// <returns>An <see cref="AIAgent"/> instance backed by the OpenAI Response service.</returns>
+    /// <returns>An <see cref="AIAgent"/> instance backed by the OpenAI Assistant service.</returns>
     /// <exception cref="ArgumentNullException">Thrown when <paramref name="client"/> or <paramref name="model"/> is <see langword="null"/>.</exception>
     /// <exception cref="ArgumentException">Thrown when <paramref name="model"/> is empty or whitespace.</exception>
     public static AIAgent CreateAIAgent(this AssistantClient client, string model, string? instructions = null, string? name = null, string? description = null, IList<AITool>? tools = null, ILoggerFactory? loggerFactory = null)
@@ -54,16 +54,16 @@ public static class OpenAIAssistantClientExtensions
     /// Creates an AI agent from an <see cref="AssistantClient"/> using the OpenAI Assistant API.
     /// </summary>
     /// <param name="client">The OpenAI <see cref="AssistantClient" /> to use for the agent.</param>
-    /// <param name="model">The model identifier to use for responses (e.g., "gpt-4", "gpt-3.5-turbo").</param>
+    /// <param name="model">The model identifier to use (e.g., "gpt-4").</param>
     /// <param name="options">Full set of options to configure the agent.</param>
     /// <param name="loggerFactory">Optional logger factory for enabling logging within the agent.</param>
-    /// <returns>An <see cref="AIAgent"/> instance backed by the OpenAI Response service.</returns>
-    /// <exception cref="ArgumentNullException">Thrown when <paramref name="client"/> or <paramref name="model"/> is <see langword="null"/>.</exception>
+    /// <returns>An <see cref="AIAgent"/> instance backed by the OpenAI Assistant service.</returns>
+    /// <exception cref="ArgumentNullException">Thrown when <paramref name="client"/> or <paramref name="model"/> or <paramref name="options"/> is <see langword="null"/>.</exception>
     /// <exception cref="ArgumentException">Thrown when <paramref name="model"/> is empty or whitespace.</exception>
     public static AIAgent CreateAIAgent(this AssistantClient client, string model, ChatClientAgentOptions options, ILoggerFactory? loggerFactory = null)
     {
         Throw.IfNull(client);
-        Throw.IfNull(model);
+        Throw.IfNullOrEmpty(model);
         Throw.IfNull(options);
 
         var assistantOptions = new AssistantCreationOptions()
@@ -116,13 +116,13 @@ public static class OpenAIAssistantClientExtensions
     /// Creates an AI agent from an <see cref="AssistantClient"/> using the OpenAI Assistant API.
     /// </summary>
     /// <param name="client">The OpenAI <see cref="AssistantClient" /> to use for the agent.</param>
-    /// <param name="model">The model identifier to use for responses (e.g., "gpt-4", "gpt-3.5-turbo").</param>
+    /// <param name="model">The model identifier to use (e.g., "gpt-4").</param>
     /// <param name="instructions">Optional system instructions that define the agent's behavior and personality.</param>
     /// <param name="name">Optional name for the agent for identification purposes.</param>
     /// <param name="description">Optional description of the agent's capabilities and purpose.</param>
     /// <param name="tools">Optional collection of AI tools that the agent can use during conversations.</param>
     /// <param name="loggerFactory">Optional logger factory for enabling logging within the agent.</param>
-    /// <returns>An <see cref="AIAgent"/> instance backed by the OpenAI Response service.</returns>
+    /// <returns>An <see cref="AIAgent"/> instance backed by the OpenAI Assistant service.</returns>
     /// <exception cref="ArgumentNullException">Thrown when <paramref name="client"/> or <paramref name="model"/> is <see langword="null"/>.</exception>
     /// <exception cref="ArgumentException">Thrown when <paramref name="model"/> is empty or whitespace.</exception>
     public static async Task<AIAgent> CreateAIAgentAsync(this AssistantClient client, string model, string? instructions = null, string? name = null, string? description = null, IList<AITool>? tools = null, ILoggerFactory? loggerFactory = null)
@@ -146,10 +146,10 @@ public static class OpenAIAssistantClientExtensions
     /// Creates an AI agent from an <see cref="AssistantClient"/> using the OpenAI Assistant API.
     /// </summary>
     /// <param name="client">The OpenAI <see cref="AssistantClient" /> to use for the agent.</param>
-    /// <param name="model">The model identifier to use for responses (e.g., "gpt-4", "gpt-3.5-turbo").</param>
+    /// <param name="model">The model identifier to use (e.g., "gpt-4").</param>
     /// <param name="options">Full set of options to configure the agent.</param>
     /// <param name="loggerFactory">Optional logger factory for enabling logging within the agent.</param>
-    /// <returns>An <see cref="AIAgent"/> instance backed by the OpenAI Response service.</returns>
+    /// <returns>An <see cref="AIAgent"/> instance backed by the OpenAI Assistant service.</returns>
     /// <exception cref="ArgumentNullException">Thrown when <paramref name="client"/> or <paramref name="model"/> is <see langword="null"/>.</exception>
     /// <exception cref="ArgumentException">Thrown when <paramref name="model"/> is empty or whitespace.</exception>
     public static async Task<AIAgent> CreateAIAgentAsync(this AssistantClient client, string model, ChatClientAgentOptions options, ILoggerFactory? loggerFactory = null)
--- a/dotnet/src/Microsoft.Extensions.AI.Agents.OpenAI/OpenAIChatClientExtensions.cs
+++ b/dotnet/src/Microsoft.Extensions.AI.Agents.OpenAI/OpenAIChatClientExtensions.cs
@@ -54,10 +54,11 @@ public static class OpenAIChatClientExtensions
     /// <param name="options">Full set of options to configure the agent.</param>
     /// <param name="loggerFactory">Optional logger factory for enabling logging within the agent.</param>
     /// <returns>An <see cref="AIAgent"/> instance backed by the OpenAI Chat Completion service.</returns>
-    /// <exception cref="ArgumentNullException">Thrown when <paramref name="client"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentNullException">Thrown when <paramref name="client"/> or <paramref name="options"/> is <see langword="null"/>.</exception>
     public static AIAgent CreateAIAgent(this ChatClient client, ChatClientAgentOptions options, ILoggerFactory? loggerFactory = null)
     {
         Throw.IfNull(client);
+        Throw.IfNull(options);
 
         var chatClient = client.AsIChatClient();
         ChatClientAgent agent = new(chatClient, options, loggerFactory);
--- a/dotnet/src/Microsoft.Extensions.AI.Agents.OpenAI/OpenAIResponseClientExtensions.cs
+++ b/dotnet/src/Microsoft.Extensions.AI.Agents.OpenAI/OpenAIResponseClientExtensions.cs
@@ -54,10 +54,11 @@ public static class OpenAIResponseClientExtensions
     /// <param name="options">Full set of options to configure the agent.</param>
     /// <param name="loggerFactory">Optional logger factory for enabling logging within the agent.</param>
     /// <returns>An <see cref="AIAgent"/> instance backed by the OpenAI Response service.</returns>
-    /// <exception cref="ArgumentNullException">Thrown when <paramref name="client"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentNullException">Thrown when <paramref name="client"/> or <paramref name="options"/> is <see langword="null"/>.</exception>
     public static AIAgent CreateAIAgent(this OpenAIResponseClient client, ChatClientAgentOptions options, ILoggerFactory? loggerFactory = null)
     {
         Throw.IfNull(client);
+        Throw.IfNull(options);
 
         var chatClient = client.AsIChatClient();
         ChatClientAgent agent = new(chatClient, options, loggerFactory);


dotnet/samples/GettingStarted/External/MEAI.OpenAI/AgentRunResponseExtensions.cs
```diff
@@ -0,0 +1,100 @@
...
+    /// OpenAI API (such as the Responses API), use <see cref="AsOpenAIResponse"/> instead.
+    /// </para>
+    /// </remarks>
+    public static ChatCompletion AsChatCompletion(this AgentRunResponse agentResponse)
+    {
+        if (agentResponse.RawRepresentation is ChatCompletion chatCompletion)
+        {
+            return chatCompletion;
+        }
+        throw new ArgumentException("ChatResponse.RawRepresentation must be a ChatCompletion");
```
stephentoub: We can expose conversions from M.E.AI.OpenAI if desirable. At least for ChatCompletion; last I tried, OpenAIResponse can't actually be constructed from outside of the library. We've already exposed the other direction, e.g. https://github.com/dotnet/extensions/blob/006a5696a88e4c8761017949dc1ea0a809a4feff/src/Libraries/Microsoft.Extensions.AI.OpenAI/MicrosoftExtensionsAIChatExtensions.cs#L31

dotnet/samples/GettingStarted/External/MEAI.OpenAI/OpenAIAgentExtensions.cs
```diff
@@ -0,0 +1,331 @@
...
+    /// <list type="bullet">
+    /// <item><description><see cref="AssistantChatMessage"/></description></item>
+    /// <item><description><see cref="DeveloperChatMessage"/></description></item>
+    /// <item><description><see cref="FunctionChatMessage"/> (obsolete)</description></item>
+    /// <item><description><see cref="SystemChatMessage"/></description></item>
+    /// <item><description><see cref="ToolChatMessage"/></description></item>
+    /// <item><description><see cref="UserChatMessage"/></description></item>
+    /// </list>
+    /// </remarks>
+    public static IEnumerable<Microsoft.Extensions.AI.ChatMessage> AsChatMessages(this IEnumerable<OpenAI.Chat.ChatMessage> messages)
```
stephentoub: This exists:
https://github.com/dotnet/extensions/blob/006a5696a88e4c8761017949dc1ea0a809a4feff/src/Libraries/Microsoft.Extensions.AI.OpenAI/MicrosoftExtensionsAIChatExtensions.cs#L24

dotnet/samples/GettingStarted/External/MEAI.OpenAI/AgentRunResponseExtensions.cs
markwallace-microsoft: Yes, anything we put in this External\MEAI.* folder is intended to go into MEAI once we have agreed on the patterns. I'll be creating issues to move this stuff over.

dotnet/samples/GettingStarted/External/MEAI.OpenAI/OpenAIAgentExtensions.cs
markwallace-microsoft: That's the other way around right? From `MEAI ChatMessage` to `OpenAI ChatMessage`.

stephentoub: Ah, yes. What's the use case?

markwallace-microsoft: One of the samples shows how to create chat messages using the OpenAI SDK and then use them with an `AIAgent`. This PR is one part of the proposal for how to reduce friction for developers using the OpenAI SDK.

dotnet/samples/GettingStarted/External/MEAI.OpenAI/AgentRunResponseExtensions.cs
rogerbarreto, line 1: @stephentoub @markwallace-microsoft 
Are we considering adding a dependency to `Microsoft.Extensions.AI.Agents` to the existing `Microsoft.Extensions.AI.OpenAI` package?

Or should we have a `Microsoft.Extensions.AI.Agents.OpenAI` for this? (Until it is implemented in the official `OpenAI` package or another official `OpenAI.Agents` .NET package that don't load all our dependencies.

dotnet/src/Microsoft.Extensions.AI.Agents.AzureAI/PersistentAgentsClientExtensions.cs
```diff
@@ -41,4 +42,55 @@ public static async Task<ChatClientAgent> GetRunnableAgentAsync(
...
+    /// <param name="instructions">The instructions for the agent.</param>
+    /// <param name="tools">The tools to be used by the agent.</param>
+    /// <param name="toolResources">The resources for the tools.</param>
+    /// <param name="temperature">The temperature setting for the agent.</param>
+    /// <param name="topP">The top-p setting for the agent.</param>
+    /// <param name="responseFormat">The response format for the agent.</param>
+    /// <param name="metadata">The metadata for the agent.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>A <see cref="ChatClientAgent"/> instance that can be used to perform operations on the newly created agent.</returns>
+    public static async Task<ChatClientAgent> CreateAIAgentAsync(
```
westey-m, line 58: Should we have a version of this that uses MEAI types?
E.g. for tools, we can provide AIFunction objects, and for ResponseFormat we can provide the MEAI config types.

dotnet/samples/GettingStarted/Providers/ChatClientAgent_With_AzureAIAgentsPersistent.cs
```diff
@@ -43,15 +40,15 @@ public async Task RunWithAzureAIAgentsPersistent()
         // Local function to run agent and display the conversation messages for the thread.
         async Task RunAgentAsync(string input)
         {
-            this.WriteUserMessage(input);
+            Console.WriteLine(input);
 
             var response = await agent.RunAsync(input, thread);
 
-            this.WriteResponseOutput(response);
+            Console.WriteLine(response.Messages.Last().Text);
```
westey-m: This should also just work:
```suggestion
            Console.WriteLine(response);
```

Even better, you could also switch to `Console.WriteLine(await agent.RunAsync(input, thread));`

dotnet/samples/GettingStarted/Providers/ChatClientAgent_With_OpenAIClient.cs
```diff
@@ -0,0 +1,93 @@
...
+{
+    private const string JokerName = "Joker";
+    private const string JokerInstructions = "You are good at telling jokes.";
+
+    [Fact]
+    public async Task RunWithChatCompletion()
+    {
+        // Get the agent directly from OpenAIClient.
+        AIAgent agent = new OpenAIClient(TestConfiguration.OpenAI.ApiKey)
+            .CreateChatClientAgent(TestConfiguration.OpenAI.ChatModelId, JokerInstructions, JokerName);
```
rogerbarreto: Instead of getting the Agent from the ChatClient what about getting the Agent from the `AIAgent` itself, so we centralize/hook the caller experience in the `AIAgent` abstraction, similarly how we did with `KernelBuilder`.

So the overall caller experience would be something like:

```csharp
var agent1 = AIAgent.CreateOpenAIChatCompletionAgent(new OpenAIClient(TestConfiguration.OpenAI.ApiKey), TestConfiguration.OpenAI.ChatModelId, JokerInstructions, JokerName);

var agent2 = AIAgent.CreateOpenAIResponsesAgent(new OpenAIClient(TestConfiguration.OpenAI.ApiKey), TestConfiguration.OpenAI.ChatModelId, JokerInstructions, JokerName);

var agent3 = AIAgent.CreateOpenAIAssistantsAgent(new OpenAIClient(TestConfiguration.OpenAI.ApiKey), TestConfiguration.OpenAI.ChatModelId, JokerInstructions, JokerName);
```

dotnet/samples/GettingStarted/External/MEAI.OpenAI/OpenAIClientExtensions.cs
```diff
@@ -0,0 +1,216 @@
...
+    /// <param name="instructions">Optional system instructions that define the agent's behavior and personality.</param>
+    /// <param name="name">Optional name for the agent for identification purposes.</param>
+    /// <param name="description">Optional description of the agent's capabilities and purpose.</param>
+    /// <param name="tools">Optional collection of AI tools that the agent can use during conversations.</param>
+    /// <param name="loggerFactory">Optional logger factory for enabling logging within the agent.</param>
+    /// <returns>An <see cref="AIAgent"/> instance backed by the OpenAI Chat Completion service.</returns>
+    /// <exception cref="ArgumentNullException">Thrown when <paramref name="client"/> or <paramref name="model"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentException">Thrown when <paramref name="model"/> is empty or whitespace.</exception>
+    public static AIAgent CreateChatClientAgent(this OpenAIClient client, string model, string? instructions = null, string? name = null, string? description = null, IList<AITool>? tools = null, ILoggerFactory? loggerFactory = null)
```
westey-m: Some alternative naming suggestions:

CreateChatAIAgent
CreateResponseAIAgent
CreateAssistantAIAgent

I think `client` is more of an implementation detail here, so shouldn't matter to folks as much.  The important thing is they are getting an `AIAgent` for Chat, Responses, or Assistants.

dotnet/samples/GettingStarted/External/MEAI.OpenAI/OpenAIClientExtensions.cs
```diff
@@ -0,0 +1,216 @@
...
+    /// <param name="instructions">Optional system instructions that define the agent's behavior and personality.</param>
+    /// <param name="name">Optional name for the agent for identification purposes.</param>
+    /// <param name="description">Optional description of the agent's capabilities and purpose.</param>
+    /// <param name="tools">Optional collection of AI tools that the agent can use during conversations.</param>
+    /// <param name="loggerFactory">Optional logger factory for enabling logging within the agent.</param>
+    /// <returns>An <see cref="AIAgent"/> instance backed by the OpenAI Response service.</returns>
+    /// <exception cref="ArgumentNullException">Thrown when <paramref name="client"/> or <paramref name="model"/> is <see langword="null"/>.</exception>
+    /// <exception cref="ArgumentException">Thrown when <paramref name="model"/> is empty or whitespace.</exception>
+    public static async Task<AIAgent> CreateAssistantClientAgentAsync(this OpenAIClient client, string model, string? instructions = null, string? name = null, string? description = null, IList<AITool>? tools = null, ILoggerFactory? loggerFactory = null)
```
westey-m: Should we have `Get` versions of these, for agents that already exist?

dotnet/samples/GettingStarted/External/MEAI.OpenAI/OpenAIClientExtensions.cs
```diff
@@ -0,0 +1,216 @@
...
+
+        var agentOptions = new ChatClientAgentOptions()
+        {
+            Id = assistantId,
+            Name = options.Name,
+            Description = options.Description,
+            Instructions = options.Instructions,
+            ChatOptions = options.ChatOptions?.Tools is null ? null : new ChatOptions()
+            {
+                Tools = options.ChatOptions.Tools,
```
westey-m: It feels weird to me that we are advertising the tools when we create the agent, and then also passing them as the default ChatOptions to the Agent, which means they'll get advertised again on each invocation.
I suppose the main (only?) value prop of doing it when you create the agent is to avoid passing them for each invocation, saving some small amount of network bandwidth.
Not that I have a great alternative though...

dotnet/samples/GettingStarted/External/MEAI.OpenAI/AIAgentWithOpenAIExtensions.cs
```diff
@@ -0,0 +1,356 @@
...
+    /// <list type="bullet">
+    /// <item><description><see cref="AssistantChatMessage"/></description></item>
+    /// <item><description><see cref="DeveloperChatMessage"/></description></item>
+    /// <item><description><see cref="FunctionChatMessage"/> (obsolete)</description></item>
+    /// <item><description><see cref="SystemChatMessage"/></description></item>
+    /// <item><description><see cref="ToolChatMessage"/></description></item>
+    /// <item><description><see cref="UserChatMessage"/></description></item>
+    /// </list>
+    /// </remarks>
+    public static IEnumerable<Microsoft.Extensions.AI.ChatMessage> AsChatMessages(this IEnumerable<OpenAI.Chat.ChatMessage> messages)
```
rogerbarreto: Make those internal?

dotnet/samples/GettingStarted/External/MEAI.OpenAI/OpenAIClientExtensions.cs
stephentoub: If the tools are AIFunctions, providing them only at construction will be problematic when using FunctionInvokingChatClient, as it won't have access to the requested functions to be able to invoke them. If this is important, we should look at imbuing FICC with its own tools collection that it also consults when looking up functions to invoke, e.g.
```C#
public class FunctionInvokingChatClient
{
    /// <summary>Gets or sets a collection of additional functions the client is able to invoke</summary>
    /// <remarks>
    /// These will not impact the requests sent by the FunctionInvokingChatClient, which will pass through the
    /// <see cref="ChatOptions.Tools" /> unmodified. However, if the service requests the invocation of a tool
    /// that was not in <see cref="ChatOptions.Tools" />, this collection will also be consulted to look for a
    /// corresponding tool to invoke. This is useful when the service may have been pre-configured to be aware
    /// of certain tools that aren't then sent on each individual request.
    /// </remarks>
    public IList<AITool> AdditionalTools { get; set; }

    ...
}
```

dotnet/samples/GettingStarted/Providers/ChatClientAgent_With_AzureOpenAIChatCompletion.cs
```diff
@@ -29,7 +29,7 @@ public async Task RunWithChatCompletion()
                 .AsIChatClient();
 
         // Define the agent
-        ChatClientAgent agent = new(chatClient, JokerInstructions, JokerName);
+        AIAgent agent = new ChatClientAgent(chatClient, JokerInstructions, JokerName);
```
westey-m: AzureOpenAIClient inherits from OpenAIClient.  Can we use the same Create extension methods that we added for OpenAI here?

dotnet/samples/GettingStarted/Providers/ChatClientAgent_With_OpenAIClient.cs
```diff
@@ -0,0 +1,93 @@
...
+        // Respond to user input.
+        await RunAgentAsync("Tell me a joke about a pirate.");
+        await RunAgentAsync("Now add some emojis to the joke.");
+
+        // Local function to invoke agent and display the conversation messages for the thread.
+        async Task RunAgentAsync(string input)
+        {
+            Console.WriteLine(input);
+
+            var chatMessage = new OpenAI.Chat.UserChatMessage(input);
```
westey-m: I don't think there are any clashes for UserChatMessage right?  If so, it might be better to clarify in a comment that we are using the OpenAI types here, rather than fully qualify, to avoid any assumption that qualifying is required.

dotnet/samples/GettingStarted/External/MEAI.OpenAI/OpenAIClientExtensions.cs
stephentoub: https://github.com/dotnet/extensions/pull/6661

dotnet/samples/GettingStarted/External/MEAI.OpenAI/AgentRunResponseExtensions.cs
stephentoub: > Are we considering adding a dependency to Microsoft.Extensions.AI.Agents to the existing Microsoft.Extensions.AI.OpenAI package?

We will need to decide where the AIAgent-dependent methods ship. I'm not sure yet if we'd want that to be yet another assembly, or just M.E.AI.OpenAI.

dotnet/samples/GettingStarted/Providers/ChatClientAgent_With_OpenAIClient.cs
stephentoub: We do not want to add OpenAI-specific static methods to AIAgent. Am I misunderstanding the proposal?

rogerbarreto: > We do not want to add OpenAI-specific static methods to AIAgent.

Why?

This works very well in Semantic Kernel and `DependencyInjection` namespaces with the `IKernelBuilder` and `IServiceCollection` interfaces. Would make a lot of sense in my opinion to have for `Agent Framework` the same, in this case `AIAgent.Create____`.

dotnet/samples/GettingStarted/External/MEAI.OpenAI/AgentRunResponseExtensions.cs
markwallace-microsoft: Anything that is in the `External\MEAI.OpenAI` folder is intended to go to M.E.AI.OpenAI

dotnet/samples/GettingStarted/Providers/ChatClientAgent_With_OpenAIClient.cs
markwallace-microsoft: The proposal is to add extension methods to `OpenAIClient` so that a developer can create an `AiAgent` directly which could be using the OpenAI Chat Completion or Assistant or Response API. These extension methods would live in the MEAI.OpenAI package which already has a dependency on the OpenAI SDK and would need a dependency on MEAI.Agents abstractions (or we may add the Agents abstractions to MEAI abstractions).

I think the extension method proposal will be more discoverable.

dotnet/samples/GettingStarted/External/MEAI.OpenAI/OpenAIClientExtensions.cs
markwallace-microsoft: The thinking behind the `CreateChatClientAgent` naming is that the method uses `OpenAIClient.CreateChatClient()` internally. So I am trying to align the naming with a concept that users of the `OpenAIClient` will be familiar with. 

@KrzysztofCwalina what are your thoughts?

dotnet/samples/GettingStarted/External/MEAI.OpenAI/OpenAIClientExtensions.cs
markwallace-microsoft: Yes, created this task: https://github.com/microsoft/agent-framework/issues/265

If we get agreement from the OpenAI and Foundry SDK folks on this approach we will have to add a full set of extension methods.

dotnet/samples/GettingStarted/External/MEAI.OpenAI/OpenAIClientExtensions.cs
markwallace-microsoft: This is how AF currently works but it's worth revisiting this approach to make sure we don't have any gaps or will cause any confusion.
 
Created a follow up task: https://github.com/microsoft/agent-framework/issues/266

dotnet/samples/GettingStarted/External/MEAI.OpenAI/AIAgentWithOpenAIExtensions.cs
markwallace-microsoft: The intent is that these will be public and live in the MEAI.OpenAI package so that developers using the OpenAI SDK can use the OpenAI types together with AF. Consider the case where a developer is already using the OpenAI SDK and now wants to start using AF, we want them to be able to continue using the types they are familiar with.

dotnet/src/Microsoft.Extensions.AI.Agents.AzureAI/PersistentAgentsClientExtensions.cs
markwallace-microsoft: Created a follow up issue: https://github.com/microsoft/agent-framework/issues/268

If this proposal is approved then we can extend the extension methods as suggested.

dotnet/samples/GettingStarted/Providers/ChatClientAgent_With_OpenAIClient.cs
stephentoub: > Why?

Because we're not going to take a dependency on OpenAI, either in name or in actual dependency, from the core abstractions library containing AIAgent.

> This works very well in Semantic Kernel and DependencyInjection namespaces with the IKernelBuilder and IServiceCollection interfaces. Would make a lot of sense in my opinion to have for Agent Framework the same, in this case AIAgent.Create____.

Those are instance extension methods. Those types do not have static methods baked into them.

Are you actually instead suggesting using extension statics, which require C# 14?

dotnet/samples/GettingStarted/External/MEAI.OpenAI/AgentRunResponseExtensions.cs
```diff
@@ -0,0 +1,101 @@
...
+    {
+        Throw.IfNull(agentResponse);
+
+        if (agentResponse.RawRepresentation is ChatResponse chatResponse)
+        {
+            return chatResponse.RawRepresentation is OpenAIResponse openAIResponse
+                ? openAIResponse
+                : throw new ArgumentException("ChatResponse.RawRepresentation must be an OpenAIResponse");
+        }
+        throw new ArgumentException("AgentRunResponse.RawRepresentation must be a ChatResponse");
```
stephentoub: We wouldn't actually want to throw here. The OpenAI library needs to add a factory for OpenAIResponse, and then this should do the conversion. We'll want to add the helper but only once it's possible to do it correctly, as otherwise the majority case will fail.

dotnet/samples/GettingStarted/External/MEAI.OpenAI/OpenAIClientExtensions.cs
stephentoub: (Alternatively, we could have the extensions on ChatClient, OpenAIResponseClient, etc., and then the name doesn't need to distinguish. A developer gets the client instance however they otherwise would, and then create an agent from it.)

dotnet/samples/GettingStarted/Providers/ChatClientAgent_With_OpenAIClient.cs
rogerbarreto: > Are you actually instead suggesting using extension statics, which require C# 14?

Actually I was, but after reading your answer yeah, didn't realized my proposal was something that only the new C# 10 supports moving forward. Never mind. 

This also raises some questions about adding DI helpers for creating agents, which might align a bit with this idea. `services.AddOpenAIChatAgent`, `services.AddOpenAIResponsesAgent`... etc

dotnet/samples/GettingStarted/External/MEAI.OpenAI/AIAgentWithOpenAIExtensions.cs
rogerbarreto: I see. IMHO, the public conversion doesn't seems necessary if all the needed public Agent API's now have extensions that supports the native OpenAI types (using the current converters as `internal`).

dotnet/samples/GettingStarted/Providers/ChatClientAgent_With_OpenAIResponseClient.cs
```diff
@@ -52,27 +48,23 @@ async Task RunAgentAsync(string input)
...
-        // Define the agent
-        ChatClientAgent agent =
-            new(chatClient, options: new()
-            {
-                Name = JokerName,
-                Instructions = JokerInstructions,
-                ChatOptions = new ChatOptions
+        // Get the agent directly from OpenAIClient.
+        AIAgent agent = new OpenAIClient(TestConfiguration.OpenAI.ApiKey)
+            .CreateResponseClientAgent(TestConfiguration.OpenAI.ChatModelId,
```
rogerbarreto: Suggest simplifying the name without `Client`.

```suggestion
            .CreateResponsesAIAgent(TestConfiguration.OpenAI.ChatModelId,
```

dotnet/src/Microsoft.Extensions.AI.Agents.AzureAI/PersistentAgentsClientExtensions.cs
```diff
@@ -41,4 +42,55 @@ public static async Task<ChatClientAgent> GetRunnableAgentAsync(
         var persistentAgentResponse = await persistentAgentsClient.Administration.GetAgentAsync(agentId, cancellationToken).ConfigureAwait(false);
         return persistentAgentResponse.AsRunnableAgent(persistentAgentsClient, chatOptions);
```
rogerbarreto:
```suggestion
        return persistentAgentResponse.AsAIAgent(persistentAgentsClient, chatOptions);
```

dotnet/src/Microsoft.Extensions.AI.Agents.AzureAI/PersistentAgentsClientExtensions.cs
```diff
@@ -41,4 +42,55 @@ public static async Task<ChatClientAgent> GetRunnableAgentAsync(
...
+            tools: tools,
+            toolResources: toolResources,
+            temperature: temperature,
+            topP: topP,
+            responseFormat: responseFormat,
+            metadata: metadata,
+            cancellationToken: cancellationToken).ConfigureAwait(false);
+
+        // Get a local proxy for the agent to work with.
+        return await persistentAgentsClient.GetAIAgentAsync(createPersistentAgentResponse.Value.Id, cancellationToken: cancellationToken).ConfigureAwait(false);
```
rogerbarreto, line 90:
```suggestion
        return await persistentAgentsClient.AsAIAgentAsync(createPersistentAgentResponse.Value.Id, cancellationToken: cancellationToken).ConfigureAwait(false);
```

dotnet/src/Microsoft.Extensions.AI.Agents.AzureAI/PersistentAgentsClientExtensions.cs
```diff
@@ -22,7 +23,7 @@ public static class PersistentAgentsClientExtensions
     /// <param name="chatOptions">Options that should apply to all runs of the agent.</param>
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
     /// <returns>A <see cref="ChatClientAgent"/> instance that can be used to perform operations on the persistent agent.</returns>
-    public static async Task<ChatClientAgent> GetRunnableAgentAsync(
+    public static async Task<ChatClientAgent> GetAIAgentAsync(
```
rogerbarreto, line 22: For consistency, suggest using `Create___AIAgentAsync` (root level clients) and `AsAIAgentAsync` (leaf clients).

`openAIClient.CreateResponsesAIAgentAsync()`
`openAIClient.CreateChatAIAgentAsync()`
`openAIClient.GetChatClient().AsAIAgentAsync()`
`persistentClient.AsAIAgentAsync()`
`persistentResponsesClient.AsAIAgentAsync()`

```suggestion
    public static async Task<ChatClientAgent> AsAIAgentAsync(
```

dotnet/samples/GettingStarted/External/MEAI.OpenAI/AIAgentWithOpenAIExtensions.cs
KrzysztofCwalina: hmmm, I like the extension methods above. the openai users does not need to know that they have to adapt. they just call APIs that accept openai types. Calling these adapters all over the codebase is not ideal.

dotnet/samples/GettingStarted/External/MEAI.OpenAI/OpenAIClientExtensions.cs
KrzysztofCwalina: I would extend this method off the ChatCLient, not OpenAIClient. We are thinking about refactoring ChatClient to a separate package from the OpenAIClient. 

I would then call the method CreateAIAgent

dotnet/samples/GettingStarted/External/MEAI.OpenAI/OpenAIClientExtensions.cs
KrzysztofCwalina: I am not sure we want/need this one. Assistants are deprecated and we will probably phase them out (including removing/frowarding-out them from the OpenAI package)

dotnet/samples/GettingStarted/External/MEAI.OpenAI/AgentRunResponseExtensions.cs
KrzysztofCwalina: what do you mean by "The OpenAI library needs to add a factory for OpenAIResponse"?

stephentoub: > what do you mean by "The OpenAI library needs to add a factory for OpenAIResponse"?

Last I checked, external to the OpenAI library there was no way to construct a new OpenAIResponse instance, e.g. an equivalent of OpenAIChatModelFactory for Responses. Has that changed? Or am I just missing it?

KrzysztofCwalina: Ah, makes sense. We recently added some missing factories to the package, but I just checked and apparently we still miss Responses. Not sure how that happened. I will do a PR today to add it.

dotnet/samples/GettingStarted/External/MEAI.OpenAI/AIAgentWithOpenAIExtensions.cs
markwallace-microsoft: OK, I get the point you are both making. I agree developers shouldn't have to call the converter explicitly if we provide extension methods for the `Run` methods that take and return the appropriate types.

dotnet/samples/GettingStarted/External/MEAI.OpenAI/OpenAIClientExtensions.cs
markwallace-microsoft: OK, that makes sense. I added it for completeness. I expect we will refactor a good bit once all of the feedback has landed.

dotnet/samples/GettingStarted/External/MEAI.OpenAI/OpenAIChatClientAgent.cs
```diff
@@ -0,0 +1,70 @@
...
+namespace OpenAI;
+
+/// <summary>
+/// OpenAI chat completion based implementation of <see cref="AIAgent"/>.
+/// </summary>
+public class OpenAIChatClientAgent : AIAgent
+{
+    private readonly ChatClientAgent _chatClientAgent;
+
+    public OpenAIChatClientAgent(OpenAIClient client, string model, string? instructions = null, string? name = null, string? description = null, ILoggerFactory? loggerFactory = null)
```
KrzysztofCwalina: this should take ChatClient, not OpenAIClient

dotnet/samples/GettingStarted/Providers/AIAgent_With_OpenAIClient.cs
```diff
@@ -0,0 +1,95 @@
...
+        await RunAgentAsync("Now add some emojis to the joke.");
+
+        // Local function to invoke agent and display the conversation messages for the thread.
+        async Task RunAgentAsync(string input)
+        {
+            Console.WriteLine(input);
+
+            var response = await agent.RunAsync(input, thread);
+
+            Console.WriteLine(response.Messages.Last().Text);
```
westey-m, line 40:
```suggestion
            Console.WriteLine(response);
```

dotnet/src/Microsoft.Extensions.AI.Agents.AzureAI/PersistentAgentsClientExtensions.cs
westey-m: I'm not sure about this.  Having something that makes a service call to retrieve the agent metadata be called As*, seems wrong.  Also, the input here isn't an agent type object, but rather a client.  If it was an agent type object, that we're just converting to a different agent type object, As* would make more sense to me.

dotnet/src/Microsoft.Extensions.AI.Agents.AzureAI/PersistentAgentsClientExtensions.cs
westey-m: +1.  I forgot to update this when I renamed Agent to AIAgent.

dotnet/src/Microsoft.Extensions.AI.Agents.OpenAI/OpenAIChatClientAgent.cs
```diff
@@ -0,0 +1,94 @@
...
+    /// <summary>
+    /// Initialize an instance of <see cref="OpenAIChatClientAgent"/>
+    /// </summary>
+    /// <param name="client">Instance of <see cref="OpenAIChatClient"/></param>
+    /// <param name="model">Id of the model.</param>
+    /// <param name="instructions">Optional instructions for the agent.</param>
+    /// <param name="name">Optional name for the agent.</param>
+    /// <param name="description">Optional description for the agent.</param>
+    /// <param name="loggerFactory">Optional instance of <see cref="ILoggerFactory"/></param>
+    public OpenAIChatClientAgent(OpenAIClient client, string model, string? instructions = null, string? name = null, string? description = null, ILoggerFactory? loggerFactory = null)
```
westey-m: Also consider adding tools to this constructor.  This would match the ChatClientAgent constructor.  An agent without tools is not that useful after all, and passing tools via ChatClientAgentOptions is not the best experience, since you have to pass them to the ChatOptions inside of ChatCilentAgentOptions.

westey-m: Would it make sense to also add a constructor overload that takes ChatCompletionsOptions or the relevant properties from that as input?  This would match the OpenAI specific run methods with regards to argument types.

stephentoub: I think these ctors should accept an OpenAI.Chat.ChatClient rather than an OpenAI.OpenAIClient.

dotnet/src/Microsoft.Extensions.AI.Agents.OpenAI/OpenAIChatClientAgent.cs
```diff
@@ -0,0 +1,94 @@
...
+
+    /// <summary>
+    /// Run the agent with the provided message and arguments.
+    /// </summary>
+    /// <param name="messages">The messages to pass to the agent.</param>
+    /// <param name="thread">The conversation thread to continue with this invocation. If not provided, creates a new thread. The thread will be mutated with the provided messages and agent response.</param>
+    /// <param name="options">Optional parameters for agent invocation.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>A <see cref="ChatCompletion"/> containing the list of <see cref="ChatMessage"/> items.</returns>
+    public async Task<ChatCompletion> RunAsync(IEnumerable<ChatMessage> messages, AgentThread? thread = null, AgentRunOptions? options = null, CancellationToken cancellationToken = default)
```
stephentoub: I was imagining that this would be virtual and that it would also seal the base RunAsync. That way, if someone did want to use OpenAI.Chat types to implement an agent, they'd derive from OpenAIChatClientAgent and override this method, at which point they're only aware of the OpenAI.Chat object model. We'd separately do the same with an OpenAIResponsesClientAgent.

dotnet/src/Microsoft.Extensions.AI.Agents.OpenAI/AIAgentWithOpenAIExtensions.cs
```diff
@@ -0,0 +1,356 @@
...
+    /// <param name="chatMessage">The OpenAI chat message to convert.</param>
+    /// <returns>A <see cref="Microsoft.Extensions.AI.ChatMessage"/> equivalent of the input OpenAI message.</returns>
+    /// <exception cref="ArgumentNullException">Thrown when <paramref name="chatMessage"/> is <see langword="null"/>.</exception>
+    /// <exception cref="NotSupportedException">Thrown when the <paramref name="chatMessage"/> type is not supported for conversion.</exception>
+    /// <remarks>
+    /// This method provides a bridge between OpenAI SDK message types and Microsoft Extensions AI message types.
+    /// It handles the conversion by switching on the concrete type of the OpenAI message and calling the appropriate
+    /// specialized conversion method.
+    /// </remarks>
+    public static Microsoft.Extensions.AI.ChatMessage AsChatMessage(this OpenAI.Chat.ChatMessage chatMessage)
```
stephentoub: Does this need to be public? What's the scenario where someone has a single message that needs to be converted to a single message?

dotnet/src/Microsoft.Extensions.AI.Agents.OpenAI/AIAgentWithOpenAIExtensions.cs
```diff
@@ -0,0 +1,356 @@
...
+    /// <list type="bullet">
+    /// <item><description><see cref="AssistantChatMessage"/></description></item>
+    /// <item><description><see cref="DeveloperChatMessage"/></description></item>
+    /// <item><description><see cref="FunctionChatMessage"/> (obsolete)</description></item>
+    /// <item><description><see cref="SystemChatMessage"/></description></item>
+    /// <item><description><see cref="ToolChatMessage"/></description></item>
+    /// <item><description><see cref="UserChatMessage"/></description></item>
+    /// </list>
+    /// </remarks>
+    public static IEnumerable<Microsoft.Extensions.AI.ChatMessage> AsChatMessages(this IEnumerable<OpenAI.Chat.ChatMessage> messages)
```
stephentoub: This is now available in the M.E.AI.OpenAI impl. It'll be in the next build.

dotnet/src/Microsoft.Extensions.AI.Agents.OpenAI/AgentRunResponseExtensions.cs
```diff
@@ -0,0 +1,101 @@
...
+    /// OpenAI API (such as the Responses API), use <see cref="AsOpenAIResponse"/> instead.
+    /// </para>
+    /// </remarks>
+    public static ChatCompletion AsChatCompletion(this AgentRunResponse agentResponse)
+    {
+        Throw.IfNull(agentResponse);
+
+        return agentResponse.RawRepresentation is ChatCompletion chatCompletion
+            ? chatCompletion
+            : throw new ArgumentException("ChatResponse.RawRepresentation must be a ChatCompletion");
```
stephentoub: This shouldn't need to throw. But an implementation for converting a ChatResponse to a ChatCompletion is already merged in M.E.AI.OpenAI, so it can use that once the build is available.

dotnet/src/Microsoft.Extensions.AI.Agents.OpenAI/AgentRunResponseExtensions.cs
```diff
@@ -0,0 +1,101 @@
...
+    {
+        Throw.IfNull(agentResponse);
+
+        if (agentResponse.RawRepresentation is ChatResponse chatResponse)
+        {
+            return chatResponse.RawRepresentation is OpenAIResponse openAIResponse
+                ? openAIResponse
+                : throw new ArgumentException("ChatResponse.RawRepresentation must be an OpenAIResponse");
+        }
+        throw new ArgumentException("AgentRunResponse.RawRepresentation must be a ChatResponse");
```
stephentoub, line 50: The change was merged into OpenAI to add a model factory for OpenAIResponse. Once that's available on nuget, we'll add the corresponding conversion methods to M.E.AI.OpenAI.

dotnet/samples/GettingStarted/External/MEAI.OpenAI/OpenAIChatClientAgent.cs
markwallace-microsoft: Done

dotnet/samples/GettingStarted/Providers/ChatClientAgent_With_OpenAIResponseClient.cs
markwallace-microsoft: These are all renamed to just `CreateAIAgent`

dotnet/src/Microsoft.Extensions.AI.Agents.AzureAI/PersistentAgentsClientExtensions.cs
markwallace-microsoft: This method performs a GET operation i.e. `persistentAgentsClient.Administration.GetAgentAsync` so I think the name is OK.

dotnet/src/Microsoft.Extensions.AI.Agents.AzureAI/PersistentAgentsClientExtensions.cs
markwallace-microsoft: I have using `CreateAIAgent` or `CreateAIAgentAsync`

dotnet/src/Microsoft.Extensions.AI.Agents.OpenAI/OpenAIChatClientAgent.cs
markwallace-microsoft: I agree. Will add this in a follow up PR as I have to also add an Agent implementation for OpenAIResponseClient. https://github.com/microsoft/agent-framework/issues/293

markwallace-microsoft: I agree. Will add this in a follow up PR as I have to also add an Agent implementation for OpenAIResponseClient. https://github.com/microsoft/agent-framework/issues/293

dotnet/samples/GettingStarted/External/MEAI.OpenAI/AIAgentWithOpenAIExtensions.cs
markwallace-microsoft: Methods are now internal

dotnet/samples/GettingStarted/Providers/ChatClientAgent_With_OpenAIResponseClient.cs
markwallace-microsoft: Renamed to CreateAIAgent

dotnet/src/Microsoft.Extensions.AI.Agents.OpenAI/OpenAIChatClientAgent.cs
markwallace-microsoft: Done

dotnet/src/Microsoft.Extensions.AI.Agents.OpenAI/AIAgentWithOpenAIExtensions.cs
markwallace-microsoft: Made this internal

dotnet/src/Microsoft.Extensions.AI.Agents.OpenAI/OpenAIAssistantClientExtensions.cs
```diff
@@ -20,24 +18,24 @@ namespace OpenAI;
...
-    /// <param name="model">The model identifier to use for chat completions (e.g., "gpt-4", "gpt-3.5-turbo").</param>
+    /// <param name="client">The OpenAI <see cref="AssistantClient" /> to use for the agent.</param>
+    /// <param name="model">The model identifier to use for responses (e.g., "gpt-4", "gpt-3.5-turbo").</param>
     /// <param name="instructions">Optional system instructions that define the agent's behavior and personality.</param>
     /// <param name="name">Optional name for the agent for identification purposes.</param>
     /// <param name="description">Optional description of the agent's capabilities and purpose.</param>
     /// <param name="tools">Optional collection of AI tools that the agent can use during conversations.</param>
     /// <param name="loggerFactory">Optional logger factory for enabling logging within the agent.</param>
-    /// <returns>An <see cref="AIAgent"/> instance backed by the OpenAI Chat Completion service.</returns>
+    /// <returns>An <see cref="AIAgent"/> instance backed by the OpenAI Response service.</returns>
```
westey-m:
```suggestion
    /// <returns>An <see cref="AIAgent"/> instance backed by the OpenAI Assistants service.</returns>
```

dotnet/src/Microsoft.Extensions.AI.Agents.OpenAI/OpenAIAssistantClientExtensions.cs
```diff
@@ -53,79 +51,71 @@ public static AIAgent CreateChatClientAgent(this OpenAIClient client, string mod
...
+    /// Creates an AI agent from an <see cref="AssistantClient"/> using the OpenAI Assistant API.
     /// </summary>
-    /// <param name="client">The OpenAI client to use for the agent.</param>
-    /// <param name="model">The model identifier to use for chat completions (e.g., "gpt-4", "gpt-3.5-turbo").</param>
+    /// <param name="client">The OpenAI <see cref="AssistantClient" /> to use for the agent.</param>
+    /// <param name="model">The model identifier to use for responses (e.g., "gpt-4", "gpt-3.5-turbo").</param>
     /// <param name="options">Full set of options to configure the agent.</param>
     /// <param name="loggerFactory">Optional logger factory for enabling logging within the agent.</param>
-    /// <returns>An <see cref="AIAgent"/> instance backed by the OpenAI Chat Completion service.</returns>
+    /// <returns>An <see cref="AIAgent"/> instance backed by the OpenAI Response service.</returns>
```
westey-m:
```suggestion
    /// <returns>An <see cref="AIAgent"/> instance backed by the OpenAI Assistants service.</returns>
```

dotnet/src/Microsoft.Extensions.AI.Agents.OpenAI/OpenAIAssistantClientExtensions.cs
```diff
@@ -53,79 +51,71 @@ public static AIAgent CreateChatClientAgent(this OpenAIClient client, string mod
     }
 
     /// <summary>
-    /// Creates an AI agent from an <see cref="OpenAIClient"/> using the OpenAI Chat Completion API.
+    /// Creates an AI agent from an <see cref="AssistantClient"/> using the OpenAI Assistant API.
     /// </summary>
-    /// <param name="client">The OpenAI client to use for the agent.</param>
-    /// <param name="model">The model identifier to use for chat completions (e.g., "gpt-4", "gpt-3.5-turbo").</param>
+    /// <param name="client">The OpenAI <see cref="AssistantClient" /> to use for the agent.</param>
+    /// <param name="model">The model identifier to use for responses (e.g., "gpt-4", "gpt-3.5-turbo").</param>
```
westey-m:
```suggestion
    /// <param name="model">The model identifier to use for assistants (e.g., "gpt-4", "gpt-3.5-turbo").</param>
```

dotnet/src/Microsoft.Extensions.AI.Agents.OpenAI/OpenAIAssistantClientExtensions.cs
```diff
@@ -20,24 +18,24 @@ namespace OpenAI;
...
+public static class OpenAIAssistantClientExtensions
 {
     /// <summary>
-    /// Creates an AI agent from an <see cref="OpenAIClient"/> using the OpenAI Chat Completion API.
+    /// Creates an AI agent from an <see cref="AssistantClient"/> using the OpenAI Assistant API.
     /// </summary>
-    /// <param name="client">The OpenAI client to use for the agent.</param>
-    /// <param name="model">The model identifier to use for chat completions (e.g., "gpt-4", "gpt-3.5-turbo").</param>
+    /// <param name="client">The OpenAI <see cref="AssistantClient" /> to use for the agent.</param>
+    /// <param name="model">The model identifier to use for responses (e.g., "gpt-4", "gpt-3.5-turbo").</param>
```
westey-m:
```suggestion
    /// <param name="model">The model identifier to use for assistants (e.g., "gpt-4", "gpt-3.5-turbo").</param>
```

Issue #173, opened 2025-07-11
Move OpenAI Responses model client to agent-framework


markwallace-microsoft:
Once the code has been refactored to remove Kernel etc. we need to move the code to the agent-framework repository

- [ ] Move only necessary code and adhere to new patterns
- [ ] All unit tests are moved and passing
- [ ] All integration tests are moved and passing
- [ ] All samples are moved and work

stephentoub:
Samples I get. But what else is necessary? Why can't it just use the IChatClient from M.E.AI.OpenAI?

dmytrostruk:
@stephentoub Same here, this is for Python :)

Issue #238, opened 2025-07-24
Python: OpenAI Responses client structured output does not work with streaming responses


peterychang:
Currently not working due to https://github.com/openai/openai-python/issues/2305

PR #239, opened 2025-07-24
Python: OpenAI responses client

peterychang:
Resolves https://github.com/microsoft/agent-framework/issues/173

Structured outputs do not currently work due to a bug in the openai apis
Opens https://github.com/microsoft/agent-framework/issues/238

### DIFF
--- a/python/packages/main/agent_framework/openai/_responses_client.py
+++ b/python/packages/main/agent_framework/openai/_responses_client.py
@@ -127,7 +127,7 @@ class OpenAIResponsesClient(OpenAIConfigBase, ChatClientBase, OpenAIHandler):
         parallel_tool_calls: bool | None = None,
         model: str | None = None,
         previous_response_id: str | None = None,
-        reasoning: str | None = None,
+        reasoning: dict[str, str] | None = None,
         service_tier: str | None = None,
         response_format: type[BaseModel] | None = None,
         seed: int | None = None,
@@ -148,7 +148,35 @@ class OpenAIResponsesClient(OpenAIConfigBase, ChatClientBase, OpenAIHandler):
         additional_properties: dict[str, Any] | None = None,
         **kwargs: Any,
     ) -> ChatResponse:
-        """Get a response from the OpenAI API."""
+        """Get a response from the OpenAI API.
+
+        Args:
+            messages: the message or messages to send to the model
+            include: additional output data to include in the model response.
+            instruction: a system (or developer) message inserted into the model's context.
+            max_tokens: The maximum number of tokens to generate.
+            parallel_tool_calls: Whether to enable parallel tool calls.
+            model: The model to use for the agent.
+            previous_response_id: The ID of the previous response.
+            reasoning: The reasoning to use for the response.
+            service_tier: The service tier to use for the response.
+            response_format: The format of the response.
+            seed: The random seed to use for the response.
+            store: whether to store the response.
+            temperature: the sampling temperature to use.
+            tool_choice: the tool choice for the request.
+            tools: the tools to use for the request.
+            top_p: the nucleus sampling probability to use.
+            user: the user to associate with the request.
+            truncation: the truncation strategy to use.
+            timeout: the timeout for the request.
+            additional_properties: additional properties to include in the request.
+            kwargs: any additional keyword arguments,
+                will only be passed to functions that are called.
+
+        Returns:
+            A chat response from the model.
+        """
         filtered_options = self._filter_options(
             background=False,
             include=include,
@@ -192,7 +220,7 @@ class OpenAIResponsesClient(OpenAIConfigBase, ChatClientBase, OpenAIHandler):
         parallel_tool_calls: bool | None = None,
         model: str | None = None,
         previous_response_id: str | None = None,
-        reasoning: str | None = None,
+        reasoning: dict[str, str] | None = None,
         service_tier: str | None = None,
         response_format: type[BaseModel] | None = None,
         seed: int | None = None,
@@ -213,7 +241,35 @@ class OpenAIResponsesClient(OpenAIConfigBase, ChatClientBase, OpenAIHandler):
         additional_properties: dict[str, Any] | None = None,
         **kwargs: Any,
     ) -> AsyncIterable[ChatResponseUpdate]:
-        """Get a response from the OpenAI API."""
+        """Get a streaming response from the OpenAI API.
+
+        Args:
+            messages: the message or messages to send to the model
+            include: additional output data to include in the model response.
+            instruction: a system (or developer) message inserted into the model's context.
+            max_tokens: The maximum number of tokens to generate.
+            parallel_tool_calls: Whether to enable parallel tool calls.
+            model: The model to use for the agent.
+            previous_response_id: The ID of the previous response.
+            reasoning: The reasoning to use for the response.
+            service_tier: The service tier to use for the response.
+            response_format: The format of the response.
+            seed: The random seed to use for the response.
+            store: whether to store the response.
+            temperature: the sampling temperature to use.
+            tool_choice: the tool choice for the request.
+            tools: the tools to use for the request.
+            top_p: the nucleus sampling probability to use.
+            user: the user to associate with the request.
+            truncation: the truncation strategy to use.
+            timeout: the timeout for the request.
+            additional_properties: additional properties to include in the request.
+            kwargs: any additional keyword arguments,
+                will only be passed to functions that are called.
+
+        Returns:
+            A stream representing the response(s) from the LLM.
+        """
         filtered_options = self._filter_options(
             background=False,
             include=include,


python/packages/main/agent_framework/openai/_chat_client.py
```diff
@@ -31,7 +45,10 @@
 from ..exceptions import ServiceInitializationError, ServiceInvalidResponseError
 from ._shared import OpenAIConfigBase, OpenAIHandler, OpenAIModelTypes, OpenAISettings
 
-__all__ = ["OpenAIChatClient"]
+__all__ = ["OpenAIChatClient", "OpenAIResponsesClient"]
```
dmytrostruk: I would probably keep these in separate files, even though they all inherit from `ChatClientBase`. We are going to add OpenAI Assistants Client as well, and having all of them in one file will be probably hard to maintain.

python/packages/main/agent_framework/openai/_chat_client.py
```diff
@@ -345,3 +360,397 @@ def from_dict(cls, settings: dict[str, Any]) -> "OpenAIChatClient":
...
+        )
+
+    def _filter_options(self, **kwargs: Any) -> dict[str, Any]:
+        """Filter options for the responses call."""
+        # The responses call does not support all the options that the chat completion call does.
+        # We filter out the unsupported options.
+        return {key: value for key, value in kwargs.items() if value is not None}
+
+    # The responses create call takes very different parameters than the chat completion call,
+    # so we override the get_response method to handle the specific parameters for responses.
```
dmytrostruk: Should we handle different parameters with `kwargs` in this case? This client is going to be used by `ChatClientAgent`, and that class won't know about specific parameters of Responses API, so one of the ways how `ChatClientAgent` can pass specific parameters to this client is through `kwargs`.

python/packages/main/agent_framework/openai/_chat_client.py
```diff
@@ -345,3 +360,397 @@ def from_dict(cls, settings: dict[str, Any]) -> "OpenAIChatClient":
...
+            items.append(ChatMessage(role="assistant", contents=parsed_tool_calls))
+        if isinstance(item, ResponseOutputMessage):
+            for content in item.content:
+                # TODO(peterychang): Add annotations when available
+                if isinstance(content, ResponseOutputText):
+                    items.append(ChatMessage(role=item.role, text=content.text))
+                    metadata.update(self._get_metadata_from_response(content))
+                elif isinstance(content, ResponseOutputRefusal):
+                    items.append(ChatMessage(role=item.role, text=content.refusal))
+        return ChatResponse(
```
dmytrostruk: We can also set `raw_representation` property here and assign it with `response: OpenAIResponse` for breaking-glass scenarios, so user will be able to see actual response and additional details if needed.

python/packages/main/agent_framework/openai/_chat_client.py
```diff
@@ -345,3 +360,397 @@ def from_dict(cls, settings: dict[str, Any]) -> "OpenAIChatClient":
...
+        elif isinstance(event, ResponseTextDeltaEvent):
+            items.append(TextContent(text=event.delta))
+            metadata.update(self._get_metadata_from_response(event))
+        elif isinstance(event, ResponseCompletedEvent):
+            # Tool calls are available in the completed event
+            if parsed_tool_calls := [tool for tool in self._get_tool_calls_from_response(event.response)]:
+                items.extend(parsed_tool_calls)
+        else:
+            return None
+        return ChatResponseUpdate(
```
dmytrostruk: We can also set `raw_representation` property here and assign it with `event: OpenAIResponseStreamEvent` for breaking-glass scenarios, so user will be able to see actual response and additional details if needed.

python/packages/main/tests/openai/test_openai_responses_client.py
```diff
@@ -0,0 +1,315 @@
+# Copyright (c) Microsoft. All rights reserved.
```
dmytrostruk, line 1: Please also add an example how to use Responses chat client under `samples/getting_started/chat_client` folder. Thanks!

python/packages/main/agent_framework/openai/_chat_client.py
peterychang: I did this to clarify what this function was expecting. If the agent passes in `**kwargs`, then everything should work fine since all the additional parameters default to false and the kwargs value will catch anything else.

dmytrostruk: If we do override for non-streaming scenario, should we do the same override for streaming scenario?

peterychang: Oh, I thought I did add that. Yeah, it should definitely have one as well

python/packages/main/agent_framework/openai/_responses_client.py
```diff
@@ -0,0 +1,506 @@
...
+
+# region ResponsesClient
+
+
+@use_tool_calling
+class OpenAIResponsesClient(OpenAIConfigBase, ChatClientBase, OpenAIHandler):
+    """OpenAI Responses client class."""
+
+    MODEL_PROVIDER_NAME: ClassVar[str] = "openai"
+    SUPPORTS_FUNCTION_CALLING: ClassVar[bool] = True
```
eavanvalkenburg: this isn't needed anymore

python/packages/main/agent_framework/openai/_responses_client.py
```diff
@@ -0,0 +1,506 @@
...
+            env_file_path (str | None): Use the environment settings file as a fallback
+                to environment variables. (Optional)
+            env_file_encoding (str | None): The encoding of the environment settings file. (Optional)
+            instruction_role (str | None): The role to use for 'instruction' messages, for example,
+        """
+        try:
+            openai_settings = OpenAISettings(
+                api_key=SecretStr(api_key) if api_key else None,
+                org_id=org_id,
+                chat_model_id=ai_model_id,
```
eavanvalkenburg:
```suggestion
                responses_model_id=ai_model_id,
```

python/packages/main/agent_framework/openai/_responses_client.py
```diff
@@ -0,0 +1,506 @@
...
+                env_file_path=env_file_path,
+                env_file_encoding=env_file_encoding,
+            )
+        except ValidationError as ex:
+            raise ServiceInitializationError("Failed to create OpenAI settings.", ex) from ex
+
+        if not async_client and not openai_settings.api_key:
+            raise ServiceInitializationError("The OpenAI API key is required.")
+        if not openai_settings.chat_model_id:
```
eavanvalkenburg:
```suggestion
        if not openai_settings.responses_model_id:
```

python/packages/main/agent_framework/openai/_responses_client.py
```diff
@@ -0,0 +1,506 @@
...
+            raise ServiceInitializationError("The OpenAI model ID is required.")
+
+        super().__init__(
+            ai_model_id=openai_settings.chat_model_id,
```
eavanvalkenburg:
```suggestion
            ai_model_id=openai_settings.responses_model_id,
```

python/packages/main/agent_framework/openai/_responses_client.py
```diff
@@ -0,0 +1,506 @@
...
+        | Callable[..., Any]
+        | list[Callable[..., Any]]
+        | MutableMapping[str, Any]
+        | list[MutableMapping[str, Any]]
+        | None = None,
+        top_p: float | None = None,
+        user: str | None = None,
+        truncation: str | None = None,
+        timeout: float | None = None,
+        additional_properties: dict[str, Any] | None = None,
```
eavanvalkenburg, line 148: the point of this additional_properties dict is to allow the not-common parameters to be passed, because doing it this way breaks the abstraction, since you can't fully do the same call against this connector vs others

python/packages/main/agent_framework/openai/_responses_client.py
```diff
@@ -0,0 +1,506 @@
...
+            chat_options.ai_model_id = self.ai_model_id
+        if chat_options.tools:
+            chat_options.additional_properties.update({
+                "response_tools": self._chat_to_response_tool_spec(chat_options.tools)
+            })
+        response = await self._send_request(chat_options, messages=self._prepare_chat_history_for_request(messages))
+        assert isinstance(response, OpenAIResponse)  # nosec  # noqa: S101
+        return next(self._create_response_content(response, item) for item in response.output)
+
+    # @trace_streaming_chat_completion(MODEL_PROVIDER_NAME)
```
eavanvalkenburg: remove this

python/packages/main/agent_framework/openai/_responses_client.py
peterychang: This is one-way compatible. You can make a call to this client the same way you could make to any other. But if you write specifically for the responses client, its not necessarily compatible with other clients.

I'm not a fan of just leaving the default `get_responses` function though, since most of the parameters won't do anything and there's no indication of what the expected parameters actually are

python/packages/main/agent_framework/openai/_responses_client.py
eavanvalkenburg: in the settings there is already a attribute for the responses model is, so we should use that since you might want to use different models for that vs chat

peterychang: Yeah, sorry. I didn't realize the responses model id had already been added to the settings :P

PR #261, opened 2025-07-28
.Net: WIP: feat: Define Workflow and Executor APIs

lokitoth:
Implements the Workflow APIs for .NET

TODOs:

- [x] Executor and Message Handling
- [x] Workflow Builder
- [x] Higher Order Edges
  - [x] Loop
  - [x] Chain
  - [x] Fan In/Out
- [x] Execution on Runtime
- [x] State/~~Checkpointing~~ (will be in folloup PR)
- [x] AoT Support
  - [x] Factor out explicit handler registration
  - [x] Fix ILTrim issues
  - [x] Fix DynamicCodeExecution issues
  - [ ] ~~Source Generator for ConfigureRoutes()~~ (will be in folloup PR, if needed)
- [x] Source Documentation XML Comments
- [x] Address Remaining feedback

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,13 @@
 # Microsoft Agent Framework
 
 Highlights
-- Flexible Agent Framework: build, orchestrate, and deploy AI agents and multi-agent systems
-- Multi-Agent Orchestration: Group chat, sequential, concurrent, and handoff patterns
-- Plugin Ecosystem: Extend with native functions, OpenAPI, Model Context Protocol (MCP), and more
+- Flexible Agent Framework: build, orchestrate, and deploy AI agents and workflows
+- Multi-Agent Orchestration: group chat, sequential, concurrent, and handoff patterns
+- Graph-based Workflows: connect agents and deterministic functions using data flows with streaming, checkpointing, time-travel, and Human-in-the-loop.
+- Plugin Ecosystem: extend with native functions, OpenAPI, Model Context Protocol (MCP), and more
 - LLM Support: OpenAI, Azure OpenAI, Azure AI Foundry, and more
-- Runtime Support: In-process and distributed agent execution
-- Multimodal: Text, vision, and function calling
+- Runtime Support: in-process and distributed agent execution
+- Multimodal: text, vision, and function calling
 - Cross-Platform: .NET and Python implementations
 
 Below are the basics for each language implementation. For more details on python see [here](./python/README.md) and for .NET see [here](./dotnet/README.md).
@@ -14,14 +15,16 @@ Below are the basics for each language implementation. For more details on pytho
 ## More Examples & Samples
 
 ### Python
-- [Getting Started with Agents](./python/samples/getting_started/agents): Basic agent creation and tool usage
-- [Chat Client Examples](./python/samples/getting_started/chat_client): Direct chat client usage patterns
+- [Getting Started with Agents](./python/samples/getting_started/agents): basic agent creation and tool usage
+- [Chat Client Examples](./python/samples/getting_started/chat_client): direct chat client usage patterns
 - [Azure Integration](./python/packages/azure): Azure OpenAI and AI Foundry integration
+- [Getting Started with Workflows](./python/samples/getting_started/workflows): basic workflow creation and integration with agents
 
 ### .Net
-- [Getting Started with Agents](./dotnet/samples/GettingStarted/Steps): Basic agent creation and tool usage
-- [Agent Provider Samples](./dotnet/samples/GettingStarted/Providers): Samples showing different agent providers
-- [Orchestration Samples](./dotnet/samples/GettingStarted/Orchestration): Advanced multi-agent patterns
+- [Getting Started with Agents](./dotnet/samples/GettingStarted/Steps): basic agent creation and tool usage
+- [Agent Provider Samples](./dotnet/samples/GettingStarted/Providers): samples showing different agent providers
+- [Orchestration Samples](./dotnet/samples/GettingStarted/Orchestration): advanced multi-agent patterns
+- [Getting Started with Workflows](./dotnet/samples/GettingStarted/Workflows): (Coming soon) basic workflow creation and integration with agents
 
 ## Agent Framework Documentation
 


dotnet/src/Microsoft.Agents.Orchestration/Workflows/Core/MessageRouting.cs
```diff
@@ -0,0 +1,342 @@
...
+        Microsoft.Agents.Orchestration.Workflows.Core.MessageHandlerInfo
+    >;
+
+namespace Microsoft.Agents.Orchestration.Workflows.Core;
+
+/// <summary>
+/// This class represents the result of a call to a <see cref="IMessageHandler{TMessage}"/>
+/// or <see cref="IMessageHandler{TMessage,TResult}"/>.
+/// </summary>
+public sealed class CallResult
```
crickman: Can we please have one class/struct per file to aid in discoverability?

dotnet/src/Microsoft.Agents.Orchestration/Workflows/WorkflowBuilderExtensions.cs
```diff
@@ -0,0 +1,156 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
+using Microsoft.Agents.Orchestration.Workflows.Core;
+
+namespace Microsoft.Agents.Orchestration.Workflows;
+
+internal static class Check
```
crickman: I believe the repo already has shared utilities for this: https://github.com/microsoft/agent-framework/blob/main/dotnet/src/Shared/Throw/Throw.cs

dotnet/src/Microsoft.Agents.Orchestration/Workflows/WorkflowBuilder.cs
```diff
@@ -0,0 +1,353 @@
...
+    where TExecutor : Executor;
+
+internal struct EdgeKey : IEquatable<EdgeKey>
+{
+    public string SourceId { get; init; }
+    public string TargetId { get; init; }
+
+    public EdgeKey(string sourceId, string targetId)
+    {
+        this.SourceId = sourceId ?? throw new ArgumentNullException(nameof(sourceId));
```
crickman: `Throw.IfNull`?

dotnet/src/Microsoft.Agents.Orchestration/Workflows/WorkflowBuilder.cs
```diff
@@ -0,0 +1,353 @@
...
+        this._idValue = id ?? throw new ArgumentNullException(nameof(id));
+    }
+
+    public bool IsUnbound => this.ExecutorType == Type.Unbound;
+
+    public string Id => this.ExecutorType switch
+    {
+        Type.Unbound => this._idValue ?? throw new InvalidOperationException("This ExecutorIsh is unbound and has no ID."),
+        Type.Executor => this._executorValue!.Id,
+        //Type.Function => throw new NotImplementedException("Function type is not yet implemented."),
```
crickman: Should incomplete code be associated with a task?

dotnet/src/Microsoft.Agents.Orchestration/Workflows/Core/Message.cs
```diff
@@ -0,0 +1,117 @@
...
+using ExecutorId = string;
+// TODO: Unclear whether this should be forcibly a serializable type.
+using MetadataValueT = object;
+using RetryExceptionT = System.InvalidOperationException;
+using TopicId = string;
+
+namespace Microsoft.Agents.Orchestration.Workflows.Core;
+
+/// <summary>
+/// .
```
crickman: Comments?

dotnet/src/Microsoft.Agents.Orchestration/Workflows/Core/Executor.cs
```diff
@@ -0,0 +1,275 @@
...
+public interface IIdentified
+{
+    /// <summary>
+    /// The unique identifier.
+    /// </summary>
+    string Id { get; }
+}
+
+/// <summary>
+/// .
```
crickman: Comments?

dotnet/src/Microsoft.Agents.Orchestration/Workflows/Core/DisposableObject.cs
```diff
@@ -0,0 +1,59 @@
...
+    /// <summary>
+    /// Implements invocation of the DisposeAsync method when the object is finalized to
+    /// dispose unmanaged resources properly.
+    /// </summary>
+    ~DisposableObject()
+    {
+        // Finalizer calls DisposeAsync to ensure resources are released.
+        // This is a safety net in case DisposeAsync was not called.
+#pragma warning disable CA2012 // Use ValueTasks correctly: Uses OnCompleted to properly handle the ValueTask return.
+        ValueTask disposeTask = this.DisposeAsync(false);
```
crickman: Object state/references may not be valid during finalization

dotnet/src/Microsoft.Agents.Orchestration/Workflows/Core/ExecutionContext.cs
```diff
@@ -0,0 +1,17 @@
...
+/// <summary>
+/// Provides services for <see cref="Executor"/> subclasses.
+/// </summary>
+public interface IExecutionContext
+{
+    /// <summary>
+    /// .
+    /// </summary>
+    /// <returns></returns>
+    Task MagicAsync();
```
crickman: Is this a placeholder?

crickman:
Should this be its own project independent from the specific orchestrations?

dotnet/src/Microsoft.Agents.Orchestration/Workflows/Core/CompletedValueTaskSource.cs
```diff
@@ -0,0 +1,27 @@
...
+/// Helper class to work around lack of proper ValueTask support in .NET Framework.
+/// </summary>
+internal static class CompletedValueTaskSource
+{
+    internal static ValueTask Completed =>
+#if NET5_0_OR_GREATER
+        ValueTask.CompletedTask;
+#else
+        new(Task.CompletedTask);
+#endif
```
stephentoub: You can use `default` or `default(ValueTask)` everywhere. That's all CompletedTask is:
https://github.com/dotnet/runtime/blob/a8fee605084efb04f3c300ecb317f6b0fcdf3f99/src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/ValueTask.cs#L113
This CompletedValueTaskSource.Completed should not be necessary.

dotnet/src/Microsoft.Agents.Orchestration/Workflows/Core/CompletedValueTaskSource.cs
```diff
@@ -0,0 +1,27 @@
...
+
+    internal static ValueTask<T> FromResult<T>(T result)
+    {
+#if NET5_0_OR_GREATER
+        return new ValueTask<T>(result);
```
stephentoub: This exists on all TFMs. You shouldn't need the else block.

The only benefit of this helper then is generic type inference... if that's not meaningful for how it's being consumed, I'd suggest just having call sites stick with `new<T>(result)`.

dotnet/src/Microsoft.Agents.Orchestration/Workflows/Core/DisposableObject.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Agents.Orchestration.Workflows.Core;
+
+/// <summary>
+/// Provides a base class implementing the <see cref="IAsyncDisposable"/> interface using
+/// the virtual Dispose pattern.
+/// </summary>
+public class DisposableObject : IAsyncDisposable
```
stephentoub: This shouldn't be public.

dotnet/src/Microsoft.Agents.Orchestration/Workflows/Core/DisposableObject.cs
```diff
@@ -0,0 +1,59 @@
...
+{
+    /// <summary>
+    /// Implements invocation of the DisposeAsync method when the object is finalized to
+    /// dispose unmanaged resources properly.
+    /// </summary>
+    ~DisposableObject()
```
stephentoub: Why does this need a finalizer?

dotnet/src/Microsoft.Agents.Orchestration/Workflows/Core/DisposableObject.cs
```diff
@@ -0,0 +1,59 @@
...
+#pragma warning restore CA2012 // Use ValueTasks correctly
+
+        if (!disposeTask.IsCompleted)
+        {
+            using (ManualResetEvent barrier = new(false))
+            {
+                disposeTask.GetAwaiter().OnCompleted(() => barrier.Set());
+
+                // Wait for the DisposeAsync to complete.
+                barrier.WaitOne(); // TODO: Timeout?
```
stephentoub: You definitely do not want to be doing this in a finalizer.

dotnet/src/Microsoft.Agents.Orchestration/Workflows/Core/DisposableObject.cs
```diff
@@ -0,0 +1,59 @@
...
+        return CompletedValueTaskSource.Completed;
+    }
+
+    /// <inheritdoc/>
+    public async ValueTask DisposeAsync()
+    {
+        await this.DisposeAsync(true).ConfigureAwait(false);
+        GC.SuppressFinalize(this);
+    }
+}
```
stephentoub: I think this whole class should be deleted.

dotnet/src/Microsoft.Agents.Orchestration/Workflows/Core/Events.cs
```diff
@@ -0,0 +1,152 @@
...
+/// <summary>
+/// .
+/// </summary>
+public record ExecutorEvent : WorkflowEvent
+{
+    /// <summary>
+    /// The identifier of the executor that generated this event.
+    /// </summary>
+#if NET9_0_OR_GREATER
+    required
```
stephentoub: We shouldn't change the required nature based on TFM. If we want to use required, we should polyfill in the relevant System.Runtime.CompilerServices types as internal. You can see examples of that being done in various repos, like dotnet/extensions.

dotnet/src/Microsoft.Agents.Orchestration/Workflows/Core/Events.cs
```diff
@@ -0,0 +1,152 @@
...
+{
+    /// <summary>
+    /// .
+    /// </summary>
+    public ExecutorInvokeEvent(string executorId, object? data = null) : base(data)
+    {
+        this.ExecutorId = executorId ?? throw new ArgumentNullException(nameof(executorId), "Executor ID cannot be null.");
+    }
+
+#if NET9_0_OR_GREATER
```
stephentoub: Why?

dotnet/src/Microsoft.Agents.Orchestration/Workflows/Core/Events.cs
```diff
@@ -0,0 +1,152 @@
...
+    public ExecutorCompleteEvent()
+    { }
+#endif
+}
+
+// TODO: This is a placeholder for streaming chat message content.
+/// <summary>
+/// .
+/// </summary>
+public class StreamingChatMessageContent
```
stephentoub: This should be using the M.E.AI types for messages/content, assuming that's what this represents.

dotnet/src/Microsoft.Agents.Orchestration/Workflows/Core/ExecutionContext.cs
```diff
@@ -0,0 +1,55 @@
...
+
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Microsoft.Agents.Orchestration.Workflows.Core;
+
+/// <summary>
+/// Provides services for <see cref="Executor"/> subclasses.
+/// </summary>
+public interface IExecutionContext
```
stephentoub: Nit: we should pick a different name. This looks too similar to ExecutionContext in .NET, which has a very different meaning.

dotnet/src/Microsoft.Agents.Orchestration/Workflows/Core/MessageRouting.cs
```diff
@@ -0,0 +1,363 @@
...
+internal struct MessageHandlerInfo
+{
+    public Type InType { get; init; }
+    public Type? OutType { get; init; } = null;
+
+    public MethodInfo HandlerInfo { get; init; }
+    public Func<object, ValueTask<object?>>? Unwrapper { get; init; } = null;
+
+    [SuppressMessage("AOT", "IL3050:Calling members annotated with 'RequiresDynamicCodeAttribute' may break functionality " +
+        "when AOT compiling.", Justification = "<Pending>")]
```
stephentoub: You'll want to address these sooner rather than later, to ensure we're not designing in problems. We shouldn't merge with these suppressions.

dotnet/src/Microsoft.Agents.Orchestration/Workflows/Core/TypeErasure.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            throw new InvalidOperationException($"Expected ValueTask or ValueTask<{typeof(TResult).Name}>, but got {maybeValueTask.GetType().Name}.");
+        }
+    }
+
+#if NET5_0_OR_GREATER
+    // This suppression is qualified because for some reason VS is not recognizing the attribute's presence, treating the
+    // import as an error (due to unnecessary using).
+    [System.Diagnostics.CodeAnalysis.RequiresDynamicCode("Calls System.Reflection.MethodInfo.MakeGenericMethod(params Type[])")]
+#endif
```
stephentoub: There are ways to do this without using MakeGenericMethod. You can see, for example, that AIFunctionFactory M.E.AI.Abstractions is able to support `ValueTask`/`ValueTask<T>`-returning methods without using MakeGenericMethod.

dotnet/src/Microsoft.Agents.Orchestration/Workflows/WorkflowBuilderExtensions.cs
```diff
@@ -0,0 +1,155 @@
...
+}
+
+internal enum Activation
+{
+    WhenAll,
+}
+
+internal static class WorkflowBuilderExtensions
+{
+    public static WorkflowBuilder AddLoop(this WorkflowBuilder builder, ExecutorIsh source, ExecutorIsh loopBody, Func<object?, bool>? condition = null)
```
ekzhu: See my comment on the Python PR regarding add_loop. I think it is not very useful to have it in the builder.https://github.com/microsoft/agent-framework/pull/272#discussion_r2241739079

dotnet/src/Microsoft.Agents.Orchestration/Workflows/WorkflowBuilderExtensions.cs
```diff
@@ -0,0 +1,155 @@
...
+    private class FanInUnwrapper : Executor,
+                                  IMessageHandler<FanInMessage, IEnumerable<object>>
+    {
+        public ValueTask<IEnumerable<object>> HandleAsync(FanInMessage message, IExecutionContext context)
+        {
+            return CompletedValueTaskSource.FromResult(message.Result!);
+        }
+    }
+
+    public static WorkflowBuilder AddFanIn(this WorkflowBuilder builder, ExecutorIsh target, Activation activation = Activation.WhenAll, params ExecutorIsh[] sources)
```
ekzhu: See my other comment on the Python PR regarding fan in. https://github.com/microsoft/agent-framework/pull/272#discussion_r2241755202. I think we can remove `activation` option from the parameters for now. So the target executor will always wait for the sources to finish before starting.

stephentoub:
What is the relationship between what's in this PR and "Process Framework"? Is the intent that this replaces PF, or is the substrate on top of which it's implemented, or something else?

crickman:
> What is the relationship between what's in this PR and "Process Framework"? Is the intent that this replaces PF, or is the substrate on top of which it's implemented, or something else?

This is a replacement of process framework (called "workflow framework") that adopts AutoGen style message handling to support PF concepts...but divorced from its dependencies on `Kernel` and `KernelFunction`. Here's the python version: https://github.com/microsoft/agent-framework/pull/272

Eric has a good doc he's written up...unfortunately I don't have the link.

dotnet/src/Microsoft.Agents.Orchestration/Workflows/Core/MessageHandler.cs
```diff
@@ -0,0 +1,37 @@
...
+public interface IMessageHandler<TMessage>
+{
+    /// <summary>
+    /// Handles the incoming message asynchronously.
+    /// </summary>
+    /// <param name="message">The message to handle.</param>
+    /// <param name="context">The execution context.</param>
+    /// <returns>A task that represents the asynchronous operation.</returns>
+    ValueTask HandleAsync(TMessage message, IExecutionContext context);
+}
```
stephentoub: This design a class implementing an interface automatically handling messages for it will make it fairly hard I expect to be NativeAOT compliant. It also makes it more cumbersome to have synchronous-only handlers, as implementations are forced to have ValueTask returns. I get the mixin appeal of this approach, but I think it'd be preferable to instead do something where the base class exposes protected methods that can be called with delegates, e.g.
https://github.com/stephentoub/agent-framework/blob/83582a0d9dac012954cf938b065d6b0d81e8c35d/dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Abstractions/RuntimeActor.cs#L69-L140
Any number of variations of such methods can be exposed, and then implementations can call them with lambdas or dedicated methods, e.g.
https://github.com/stephentoub/agent-framework/blob/83582a0d9dac012954cf938b065d6b0d81e8c35d/dotnet/src/Microsoft.Agents.Orchestration/GroupChat/GroupChatAgentActor.cs#L33-L48
https://github.com/stephentoub/agent-framework/blob/83582a0d9dac012954cf938b065d6b0d81e8c35d/dotnet/src/Microsoft.Agents.Orchestration/Handoff/HandoffActor.cs#L57-L59
That also significantly simplifies all of the reflection goop that's currently implemented in this PR to support hooking up these interfaces, dealing with generic value tasks, etc.

dotnet/tests/Microsoft.Agents.Workflow.UnitTests/Sample/02_Simple_Workflow_Sequential.cs
```diff
@@ -0,0 +1,42 @@
...
+    public static ValueTask RunAsync()
+    {
+        UppercaseExecutor uppercase = new();
+        ReverseTextExecutor reverse = new();
+
+        WorkflowBuilder builder = new(uppercase);
+        builder.AddEdge(uppercase, reverse);
+
+        Workflow<string> workflow = builder.Build<string>();
+        // async foreach (var event in workflow.RunAsync("hello world"))
```
v1212: how to run the sample in local? there is no implement of workflow.RunAsync("hello world") so far.

there are also errors regarding ExecutorCompleteEvent(string executorId, object? data = null) : base(data) --> Argument 1: cannot convert from 'object' to 'Microsoft.Agents.Orchestration.Workflows.Core.ExecutorEvent'

lokitoth:
Resolved some of the feedback comments.

Summary:
* Pulled most of multi-type files apart
* Removed async DisposableObject base
* Standardized on Throw.IfNull where appropriate

dotnet/tests/Microsoft.Agents.Workflow.UnitTests/Sample/02_Simple_Workflow_Sequential.cs
lokitoth: To run a Workflow, one wraps it in a Runner instance - in this case LocalRunner. Once we have it running on top of the Actor runtime, there will be a choice of runners.

See updated Samples.

I believe as of last commit there should not be this build break, but please let me know if it is still triggering for you.

dotnet/src/Microsoft.Agents.Orchestration/Workflows/Core/MessageRouting.cs
lokitoth: Did a pass to pull out most of them; will do another pass once state stuff and Agent hosting is implemented.

dotnet/src/Microsoft.Agents.Orchestration/Workflows/WorkflowBuilderExtensions.cs
lokitoth: I can see there being value in various strategies around activation. In the short term, will remove FanInTrigger from the Builder, but I'll keep the underlying mechanism in place for now.

lokitoth:
More updates for Routing to isolate the reflection bit, and prepare for SourceGenerators

dotnet/src/Microsoft.Agents.Workflow/Core/CallResult.cs
```diff
@@ -0,0 +1,76 @@
...
+    public bool IsSuccess => this.Exception == null;
+
+    private CallResult(bool isVoid = false)
+    {
+        // Private constructor to enforce use of static methods.
+        this.IsVoid = isVoid;
+    }
+
+    /// <summary>
+    /// Create a <see cref="CallResult"/> indicating a successful that returned a result (non-void).
```
TaoChenOSU:
```suggestion
    /// Create a <see cref="CallResult"/> indicating a successful call that returned a result (non-void).
```

dotnet/src/Microsoft.Agents.Workflow/Core/Edge.cs
```diff
@@ -0,0 +1,164 @@
...
+/// <param name="SourceId">The id of the source executor node.</param>
+/// <param name="SinkIds">A list of ids of the target executor nodes.</param>
+/// <param name="Partitioner">A function that maps an incoming message to a subset of the target executor nodes.</param>
+public record FanOutEdgeData(
+    string SourceId,
+    List<string> SinkIds,
+    PartitionerT? Partitioner = null)
+{
+    /// <summary>
+    /// .
```
TaoChenOSU: nit: summary

dotnet/src/Microsoft.Agents.Workflow/Core/Edge.cs
```diff
@@ -0,0 +1,164 @@
...
+/// <param name="Trigger">The <see cref="FanInTrigger"/> that determines when the fan-in edge is activated.</param>
+public record FanInEdgeData(
+    IEnumerable<string> SourceIds,
+    string SinkId,
+    FanInTrigger Trigger = FanInTrigger.WhenAll)
+{
+    internal Guid UniqueKey { get; } = Guid.NewGuid();
+
+    /// <summary>
+    /// .
```
TaoChenOSU: nit: summary

dotnet/src/Microsoft.Agents.Workflow/Core/Events.cs
```diff
@@ -0,0 +1,135 @@
...
+/// <remarks>
+/// The user is expected to raise this event from a terminating <see cref="Executor"/>, or to build
+/// the workflow with output capture using <see cref="WorkflowBuilderExtensions.BuildWithOutput"/>.
+/// </remarks>
+public record WorkflowCompletedEvent : WorkflowEvent;
+
+/// <summary>
+/// Event triggered when a workflow executor request external information.
+/// </summary>
+public record RequestInputEvent(ExternalRequest Request) : WorkflowEvent;
```
TaoChenOSU: nit: In Python and the design doc, this is called `RequestInfoEvent`

dotnet/src/Microsoft.Agents.Workflows/Core/CallResult.cs
```diff
@@ -0,0 +1,76 @@
...
+    public bool IsSuccess => this.Exception == null;
+
+    private CallResult(bool isVoid = false)
+    {
+        // Private constructor to enforce use of static methods.
+        this.IsVoid = isVoid;
+    }
+
+    /// <summary>
+    /// Create a <see cref="CallResult"/> indicating a successful that returned a result (non-void).
```
TaoChenOSU:
```suggestion
    /// Create a <see cref="CallResult"/> indicating a successful call that returned a result (non-void).
```

dotnet/src/Microsoft.Agents.Workflows/Core/Events.cs
```diff
@@ -0,0 +1,135 @@
...
+/// <remarks>
+/// The user is expected to raise this event from a terminating <see cref="Executor"/>, or to build
+/// the workflow with output capture using <see cref="WorkflowBuilderExtensions.BuildWithOutput"/>.
+/// </remarks>
+public record WorkflowCompletedEvent : WorkflowEvent;
+
+/// <summary>
+/// Event triggered when a workflow executor request external information.
+/// </summary>
+public record RequestInputEvent(ExternalRequest Request) : WorkflowEvent;
```
TaoChenOSU: nit: In the design doc and Python, this is called `RequestInfoEvent`

dotnet/src/Microsoft.Agents.Workflows/Specialized/AIAgentHostExecutor.cs
```diff
@@ -0,0 +1,31 @@
...
+using Microsoft.Extensions.AI;
+using Microsoft.Extensions.AI.Agents;
+
+namespace Microsoft.Agents.Workflows.Specialized;
+
+internal class AIAgentHostExecutor : Executor, IMessageHandler<IList<ChatMessage>>
+{
+    private AIAgent Agent { get; set; }
+
+    public AIAgentHostExecutor(AIAgent agent)
```
TaoChenOSU, line 16: nit: we may want to allow a thread object to be passed in.

dotnet/src/Microsoft.Agents.Orchestration/Workflows/Core/MessageHandler.cs
lokitoth: I got trimming to a functional level, so we may be in good shape there. If there is still an issue, now that we have a baseline manual registration approach we can switch to Source Generators in the future.

dotnet/src/Microsoft.Agents.Workflows/Core/Edge.cs
```diff
@@ -0,0 +1,164 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Shared.Diagnostics;
+
+using PartitionerT = System.Func<object?, int, System.Collections.Generic.IEnumerable<int>>;
+using PredicateT = System.Func<object?, bool>;
+
+namespace Microsoft.Agents.Workflows.Core;
```
stephentoub: Nit: we generally try to avoid subnamespaces like "Core". For typical consumption, what namespaces is a developer going to need to import? The answer should ideally be "1".

dotnet/src/Microsoft.Agents.Workflows/Core/Edge.cs
```diff
@@ -0,0 +1,164 @@
...
+}
+
+/// <summary>
+/// Represents a connection from a single node to a set of nodes, optionally associated with a paritition selector
+/// function which maps incoming messages to a subset of the target set.
+/// </summary>
+/// <param name="SourceId">The id of the source executor node.</param>
+/// <param name="SinkIds">A list of ids of the target executor nodes.</param>
+/// <param name="Partitioner">A function that maps an incoming message to a subset of the target executor nodes.</param>
+public record FanOutEdgeData(
```
stephentoub: What drove the choice of using records for these types? We like the generated ToString implementations? Are these being put into dictionaries and need specialized hash code implementations? Are we expecting developers to need value equality? Collections in records don't behave the way some developers expect them to? From a NativeAOT perspective, records also bring along a fair amount of non-trimmable code (they implement a bunch of interfaces that will generally not be trimmed away, and thus their implementations of those interfaces generally won't be trimmed away).

dotnet/src/Microsoft.Agents.Workflows/Core/Edge.cs
```diff
@@ -0,0 +1,164 @@
...
+    string SourceId,
+    List<string> SinkIds,
+    PartitionerT? Partitioner = null)
+{
+    /// <summary>
+    /// Converts a <see cref="FanOutEdgeData"/> instance to an <see cref="Edge"/>.
+    /// </summary>
+    /// <param name="data">The <see cref="FanOutEdgeData"/> to convert.</param>
+    public static implicit operator Edge(FanOutEdgeData data)
```
stephentoub: Nit: We generally try to avoid implicit conversions that allocate.

dotnet/src/Microsoft.Agents.Workflows/Core/Edge.cs
```diff
@@ -0,0 +1,164 @@
...
+        return new Edge(data);
+    }
+}
+
+/// <summary>
+/// Specifies the condition under which a fan-in operation is triggered in a workflow.
+/// Use <see cref="FanInTrigger.WhenAll"/> to trigger the operation when all incoming edges have data, or
+/// <see cref="FanInTrigger.WhenAny"/> to trigger when any incoming edge has data.
+/// </summary>
+public enum FanInTrigger
```
stephentoub: Are all and any sufficient, or will there be cases where the condition is more nuanced (like when a specific subset of edges complete, or a specific number of them complete)? As an enum, this will be very limited in what can be expressed in the future.

dotnet/src/Microsoft.Agents.Workflows/Core/Edge.cs
```diff
@@ -0,0 +1,164 @@
...
+
+/// <summary>
+/// Represents a connection from a set of nodes to a single node. It can trigger either when all edges have data
+/// or when any of them have data.
+/// </summary>
+/// <param name="SourceIds">An enumeration of ids of the source executor nodes.</param>
+/// <param name="SinkId">The id of the target executor node.</param>
+/// <param name="Trigger">The <see cref="FanInTrigger"/> that determines when the fan-in edge is activated.</param>
+public record FanInEdgeData(
+    IEnumerable<string> SourceIds,
```
stephentoub: What happens if each enumeration of source IDs changes the results, or if the enumeration is expensive? Should this accept an IEnumerable in order to be generalized but then convert it to a list to actually be stored, so that the contents are guaranteed to be immutable for future accesses?

dotnet/src/Microsoft.Agents.Workflows/Core/Edge.cs
```diff
@@ -0,0 +1,164 @@
...
+/// <summary>
+/// Represents a connection or relationship between nodes, characterized by its type and associated data.
+/// </summary>
+/// <remarks>
+/// An <see cref="Edge"/> can be of type <see cref="Type.Direct"/>, <see cref="Type.FanOut"/>, or <see
+/// cref="Type.FanIn"/>, as specified by the <see cref="EdgeType"/> property. The <see cref="Data"/> property holds
+/// additional information relevant to the edge, and its concrete type depends on the value of <see
+/// cref="EdgeType"/>, functioning as a tagged union.
+/// </remarks>
+public class Edge
```
stephentoub: We might want to seal types like this, so as to better keep our future options open.

dotnet/src/Microsoft.Agents.Workflows/Core/Executor.cs
```diff
@@ -0,0 +1,169 @@
...
+
+    /// <summary>
+    /// A set of <see cref="Type"/>s, representing the messages this executor can handle.
+    /// </summary>
+    public ISet<Type> InputTypes => this.Router.IncomingTypes;
+
+    /// <summary>
+    /// A set of <see cref="Type"/>s, representing the messages this executor can produce as output.
+    /// </summary>
+    public virtual ISet<Type> OutputTypes => new HashSet<Type>([typeof(object)]);
```
stephentoub: Properties shouldn't be returning a new collection on each access. Did you mean for this to be:
```csharp
public virtual ISet<Type> OutputTypes { get; } = new HashSet<Type>() { typeof(object) };
```
?

dotnet/src/Microsoft.Agents.Workflows/Core/Executor.cs
```diff
@@ -0,0 +1,169 @@
...
+    /// Checks if the executor can handle a specific message type.
+    /// </summary>
+    /// <param name="messageType"></param>
+    /// <returns></returns>
+    public bool CanHandle(Type messageType) => this.Router.CanHandle(messageType);
+
+    /// <inheritdoc cref="IAsyncDisposable.DisposeAsync"/>
+    protected virtual async ValueTask DisposeAsync()
+    {
+        this._initialized = false;
```
stephentoub: Does this imply it can be re-initialized after being disposed?

dotnet/src/Microsoft.Agents.Workflows/Core/Executor.cs
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    /// <inheritdoc cref="IAsyncDisposable.DisposeAsync"/>
+    ValueTask IAsyncDisposable.DisposeAsync()
+    {
+        GC.SuppressFinalize(this); // Should we be suppressing the finalizer here? CodeAnalysis seems to want it (CA1816)
```
stephentoub: It _can_, but it's not really warranted. And I would hope that no derived types actually need a finalizer.

dotnet/src/Microsoft.Agents.Workflows/Core/IMessageHandler.cs
```diff
@@ -0,0 +1,37 @@
...
+
+using System.Threading.Tasks;
+
+namespace Microsoft.Agents.Workflows.Core;
+
+/// <summary>
+/// A message handler interface for handling messages of type <typeparamref name="TMessage"/>.
+/// </summary>
+/// <typeparam name="TMessage"></typeparam>
+public interface IMessageHandler<TMessage>
```
stephentoub: Even without NativeAOT issues, I still personally would prefer if these were addressed via Register methods on the base class rather than using reflection. The user experience with such methods can be just as good if not better (you can more easily have different overloads with delegate types that bind to more signatures), there's less "magic", and it's more efficient.

dotnet/src/Microsoft.Agents.Workflows/Core/Message.cs
```diff
@@ -0,0 +1,117 @@
...
+    /// </summary>
+    public ExecutorId? TargetId { get; init; }
+    /// <summary>
+    /// .
+    /// </summary>
+    public DateTimeOffset Timestamp { get; init; } = DateTimeOffset.UtcNow;
+    /// <summary>
+    /// .
+    /// </summary>
+    public string IsoTimestamp => this.Timestamp.ToString("o");
```
stephentoub: Why is this needed? If the caller wants this particular format, they can just do it themselves, and when they do it themselves, it can be more efficient, e.g. if they're using string interpolation, $"... {data.Timestamp:o}" will be more efficient than $"... {data.IsoTimestamp".

dotnet/src/Microsoft.Agents.Workflows/Core/Message.cs
```diff
@@ -0,0 +1,117 @@
...
+    public TimeSpan? Timeout { get; init; } = null;
+
+    /// <summary>
+    /// .
+    /// </summary>
+    public int Retries { get; init; } = 0;
+    /// <summary>
+    /// .
+    /// </summary>
+    public int MaxRetries { get; init; } = 3;
```
stephentoub: What is the difference between Retries and MaxRetries?

dotnet/src/Microsoft.Agents.Workflows/Core/MessageHandlerInfo.cs
```diff
@@ -0,0 +1,134 @@
...
+using System;
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using System.Reflection;
+using System.Threading.Tasks;
+
+namespace Microsoft.Agents.Workflows.Core;
+
+internal struct MessageHandlerInfo
```
stephentoub: None of this file would be needed if you use the Register delegate approach... just sayin ;-)

dotnet/src/Microsoft.Agents.Workflows/Core/MessageRouter.cs
```diff
@@ -0,0 +1,58 @@
...
+    private readonly Dictionary<Type, MessageHandlerF> _typedHandlers;
+    private readonly bool _hasCatchall;
+
+    internal MessageRouter(Dictionary<Type, MessageHandlerF> handlers)
+    {
+        this._typedHandlers = Throw.IfNull(handlers);
+        this._hasCatchall = this._typedHandlers.ContainsKey(typeof(object));
+    }
+
+    public HashSet<Type> IncomingTypes => [.. this._typedHandlers.Keys];
```
stephentoub: This is going to construct a new hash set on every access. A property shouldn't do that.

dotnet/src/Microsoft.Agents.Workflows/Core/RouteBuilder.cs
```diff
@@ -0,0 +1,134 @@
...
+/// <remarks>
+/// Override the <see cref="ExecutorBase.ConfigureRoutes"/> method to customize the routing of messages to handlers. By
+/// default, uses reflection to find implementations of <see cref="IMessageHandler{TMessage}"/> and
+/// <see cref="IMessageHandler{TMessage, TResult}"/>.
+/// </remarks>
+public class RouteBuilder
+{
+    private readonly Dictionary<Type, MessageHandlerF> _typedHandlers = new();
+
+    internal RouteBuilder AddHandler(Type messageType, MessageHandlerF handler, bool overwrite = false)
```
stephentoub: Interesting. Are these the "Register" methods I've been alluding to? Why do we still need those extra handler interfaces and all that reflection goop?

dotnet/src/Microsoft.Agents.Workflows/Core/RouteBuilderExtensions.cs
```diff
@@ -0,0 +1,101 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+using System.Reflection;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Agents.Workflows.Core;
+
+internal static class IMessageHandlerReflection
```
stephentoub: Another file I'd love to see be deleted :)

dotnet/src/Microsoft.Agents.Workflows/Core/ValueTaskTypeErasure.cs
```diff
@@ -0,0 +1,76 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+using System.Diagnostics;
+using System.Reflection;
+using System.Threading.Tasks;
+
+namespace Microsoft.Agents.Workflows.Core;
+
+internal static class ValueTaskReflection
```
stephentoub: And another :)

dotnet/src/Microsoft.Agents.Workflows/Execution/StreamingExecutionHandle.cs
```diff
@@ -0,0 +1,184 @@
...
+    /// Asynchronously streams workflow events as they occur during workflow execution.
+    /// </summary>
+    /// <remarks>This method yields <see cref="WorkflowEvent"/> instances in real time as the workflow
+    /// progresses. The stream completes when a <see cref="WorkflowCompletedEvent"/> is encountered. Events are
+    /// delivered in the order they are raised.</remarks>
+    /// <param name="cancellation">A <see cref="CancellationToken"/> that can be used to cancel the streaming operation. If cancellation is
+    /// requested, the stream will end and no further events will be yielded.</param>
+    /// <returns>An asynchronous stream of <see cref="WorkflowEvent"/> objects representing significant workflow state changes.
+    /// The stream ends when the workflow completes or when cancellation is requested.</returns>
+    public async IAsyncEnumerable<WorkflowEvent> WatchStreamAsync([EnumeratorCancellation] CancellationToken cancellation = default)
```
stephentoub: From the name, I infer this is only an observer, watching something that's already happening, but from the implementation, it looks like this is actually performing run operations?

dotnet/src/Microsoft.Agents.Workflows/Specialized/OutputCollectorExecutor.cs
```diff
@@ -0,0 +1,43 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.Agents.Workflows.Core;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Agents.Workflows.Specialized;
```
stephentoub, line 7: Nit: back to the namespace comment, we should revisit all the namespace all up and reduce it to the minimum amount necessary.

dotnet/src/Microsoft.Agents.Workflows/Specialized/AIAgentHostExecutor.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.Agents.Workflows.Core;
+using Microsoft.Extensions.AI;
+using Microsoft.Extensions.AI.Agents;
+
+namespace Microsoft.Agents.Workflows.Specialized;
+
+internal class AIAgentHostExecutor : Executor<AIAgentHostExecutor>, IMessageHandler<IList<ChatMessage>>
```
stephentoub: Is there the inverse as well, treating an executor as an agent?

dotnet/src/Microsoft.Agents.Workflows/Core/Edge.cs
lokitoth: It was primarily around hinting that they are "immutable", and I was expecting to use equality more. Now that they are a bit more stabilized, can turn them into manually defined simpler types.

dotnet/src/Microsoft.Agents.Workflows/Core/Edge.cs
lokitoth: The goal was to make Edge behave like a type union.

dotnet/src/Microsoft.Agents.Workflows/Core/Edge.cs
lokitoth: In Python we even got right of Any (precisely because we started imaging we might want to support more complex cases). Will probably drop this outright for now.

dotnet/src/Microsoft.Agents.Workflows/Core/Edge.cs
lokitoth: Should be List<> or array. Good catch.

dotnet/src/Microsoft.Agents.Workflows/Core/Executor.cs
lokitoth: No, this is vestigial. Can remove it all outright now that Init() does not have a protected hook.

dotnet/src/Microsoft.Agents.Workflows/Core/Executor.cs
lokitoth: Our warning set requires it. In this case can all be removed since Initialize is no longer a formal concept.

dotnet/src/Microsoft.Agents.Workflows/Core/IMessageHandler.cs
lokitoth: I disagree about the user experience, so I went with a best-of-both-worlds approach. Those that prefer to manually register can do so, but I fully expect the majority of users to rely on the reflected variant.

dotnet/src/Microsoft.Agents.Workflows/Core/Message.cs
lokitoth: Vestigial type. I forgot to delete it.

dotnet/src/Microsoft.Agents.Workflows/Core/MessageRouter.cs
lokitoth: Good catch

dotnet/src/Microsoft.Agents.Workflows/Core/RouteBuilder.cs
lokitoth: Because the user experience of having to invoke AddHandler to list the set of methods already on the class is not good.

dotnet/src/Microsoft.Agents.Workflows/Execution/StreamingExecutionHandle.cs
lokitoth: The contract is that the run operations could be proceeding even without the user calling this. That they do not is an implementation detail. 

With that said, I am not very happy with this name, definitely looking for suggestions.

dotnet/src/Microsoft.Agents.Workflows/Specialized/OutputCollectorExecutor.cs
lokitoth: Agreed. What should probably happen is the types in Core get pulled to the root namespace, and the Runner-related types go into a separate namespace/assembly.

dotnet/src/Microsoft.Agents.Workflows/Specialized/AIAgentHostExecutor.cs
lokitoth: I think that would be more about treating a Workflow as an Agent: It feels odd to have a user create a lightweight wrapper only to then wrap it again, especially since the executor would have to speak "agent" types anyways.

dotnet/src/Microsoft.Agents.Workflows/Core/RouteBuilder.cs
stephentoub: > set of methods already on the class

I'm not following. In both cases you need to express the method should be used, either by adding the interface to the declaration or by adding the register call in the ctor. It's a single thing in both cases.

lokitoth: When it is done in interface form the binding is declarative, rather than imperative. 

It limits the potential set of places one might need to look to understand how routing is set up and makes it more amenable to tooling.

It also is more akin to how C# types generally behave in terms of declaring their adherence to a contract.

stephentoub: And when it's done via a delegate, it doesn't need to advertise the implementation detail of what it's listening for as part of its signature (which changing would be a breaking change for any public type), it doesn't require that the message types it's listening for have the same or more visibility, it can conditionally change what it listens for rather than having it decided statically at build time, it's not limited in expression to only the set of interface methods exposed, it doesn't require additional interfaces to be exposed incurring the additional concept count, it doesn't pay for expensive and complicated reflection-based lookup for each instance, it doesn't force a whole bunch of metadata to bloat the size of the aot compilation, and what registration capabilities are available are more easily discoverable in Intellisense on the instance vs needing to know which interfaces are automagically found via reflection.

dotnet/src/Microsoft.Agents.Workflows/Core/Executor.cs
```diff
@@ -0,0 +1,168 @@
...
+using System.Reflection;
+using System.Threading.Tasks;
+
+namespace Microsoft.Agents.Workflows.Core;
+
+/// <summary>
+/// A component that processes messages in a <see cref="Workflow"/>.
+/// </summary>
+[DebuggerDisplay("{GetType().Name}{Id}")]
+public abstract class ExecutorBase : IIdentified, IAsyncDisposable
```
crickman: Can we please define one class per file?  Scrolling down to locate the namesake class for this file is just extra work/time each instance this file is opened.

dotnet/src/Microsoft.Agents.Workflows/Core/Message.cs
```diff
@@ -0,0 +1,117 @@
...
+using MetadataValueT = object;
+using RetryExceptionT = System.InvalidOperationException;
+using TopicId = string;
+
+namespace Microsoft.Agents.Workflows.Core;
+
+/// <summary>
+/// .
+/// </summary>
+public record MessageMetadata
```
crickman: Can we please define one class per file to aid in discoverability?

dotnet/src/Microsoft.Agents.Workflows/Specialized/OutputCollectorExecutor.cs
crickman: +1 - `Workflow` feels more natural in the root namespace.

dotnet/src/Microsoft.Agents.Workflows/Core/IWorkflowContext.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Threading.Tasks;
+
+namespace Microsoft.Agents.Workflows.Core;
+
+/// <summary>
+/// Provides services for an <see cref="ExecutorBase"/> during the execution of a workflow.
+/// </summary>
+public interface IWorkflowContext
```
crickman: Should this include `CancellationToken` (or perhaps `IMessageHandler.HandleAsync` might)?

dotnet/src/Microsoft.Agents.Orchestration/Workflows/WorkflowBuilderExtensions.cs
lokitoth: Removed all Trigger types for now.

dotnet/tests/Microsoft.Agents.Workflows.UnitTests/Sample/01_Simple_Workflow_Sequential.cs
```diff
@@ -0,0 +1,55 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.IO;
+using System.Threading.Tasks;
+using Microsoft.Agents.Workflows.Core;
+using Microsoft.Agents.Workflows.Execution;
+
+namespace Microsoft.Agents.Workflows.Sample;
+
+internal static class Step1EntryPoint
```
crickman, line 9: Do we really want these samples to be buried under a unit-test sub-folder?  Perhaps we can move these to `GettingStarted` or `GetStartedWithWorkflows`?

lokitoth: Yes, but I do not want to block this PR on it; tracked by #411

dotnet/src/Microsoft.Agents.Workflows/Specialized/AIAgentHostExecutor.cs
lokitoth: Work to complete wrapping workflow in agent at #408

dotnet/src/Microsoft.Agents.Orchestration/Workflows/Core/MessageHandler.cs
lokitoth: I've isolated the Reflection stuff into its own namespace. Going to punt choosing whether we want to remove it to post this PR.

dotnet/src/Microsoft.Agents.Workflows/Specialized/AIAgentHostExecutor.cs
lokitoth: Agreed, but there is more work in general to be done here, plus reconciling the .NET and Python implementations. Tracked by #409

dotnet/src/Microsoft.Agents.Workflows/Core/Executor.cs
lokitoth: Yes, good catch. Fixed.

dotnet/src/Microsoft.Agents.Workflows/Core/RouteBuilder.cs
lokitoth: >  it doesn't need to advertise the implementation detail of what it's listening for as part of its signature (which changing would be a breaking change for any public type),

Consumers of the type still need to know that because they need to understand how to wire executors up together, and which types will flow across edges. 

With the interfaces it lets it be declared on the type, rather than somewhere available only at runtime; in the ideal world, if I could make a requirement that a type only have public signatures of a specific genericized constraint and automatically glued to all of them that would be even better (no interfaces at all), i.e. define a MetaClass that only allows public APIs (besides .ctor) that look like `void|T|ValueTask|ValueTask<T> Method(T input, IWorkflowCtx context)`.

I think we are going to need to agree to disagree here, at least for the time-being, though I am happy to replace this (after this PR is in) with a SourceGenerator-based approach.

dotnet/src/Microsoft.Agents.Workflows/Specialized/OutputCollectorExecutor.cs
lokitoth: I've done a pass collapsing the namespaces. All the public types are either in the root namespace or in .Reflection to keep that out of the default toolbox.

dotnet/src/Microsoft.Agents.Workflows/Core/IWorkflowContext.cs
lokitoth: Yes, but I need to do a bit of a cleanup in terms of the meanings of various `CancellationToken`s, so going to handle that outside of this PR (it is getting long/big). Tracked by #410

dotnet/src/Microsoft.Agents.Workflows/Core/RouteBuilder.cs
lokitoth: Will be addressed further after, tracked by #412

PR #262, opened 2025-07-29
.NET: CosmosDB Actor State Storage

adityamandaleeka:
This change adds an IActorStateStorage implementation that uses CosmosDB, and a set of tests for it that use Aspire and can run either using the real service or in the emulator.

I've just rebased everything onto current main (was previously based on a personal feature branch) and had to disable a few warnings to get things building and running. I suspect we'll be able to resolve some of those soon but I didn't want to delay getting this PR out.

### DIFF
--- a/dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/ActorDocuments.cs
+++ b/dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/ActorDocuments.cs
@@ -7,6 +7,17 @@ namespace Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB;
 
 /// <summary>
 /// Root document for each actor that provides actor-level ETag semantics.
+/// Every write operation updates this document to ensure a single ETag represents
+/// the entire actor's state for optimistic concurrency control.
+/// This document contains no actor state data. It only serves to track last modified
+/// time and provide a single ETag for the actor's state.
+///
+/// Example structure:
+/// {
+///   "id": "rootdoc",                       // Root document ID (constant per actor partition)
+///   "actorId": "actor-123",                // Partition key (actor ID)
+///   "lastModified": "2024-...",            // Timestamp
+/// }
 /// </summary>
 public sealed class ActorRootDocument
 {
@@ -28,6 +39,13 @@ public sealed class ActorRootDocument
 
 /// <summary>
 /// Actor state document that represents a single key-value pair in the actor's state.
+/// Document Structure (one per actor key):
+/// {
+///   "id": "state_sanitizedkey",            // Unique document ID for the state entry
+///   "actorId": "actor-123",                // Partition key (actor ID)
+///   "key": "foo",                          // Logical key for the state entry
+///   "value": { "bar": 42, "baz": "hello" } // Arbitrary JsonElement payload
+/// }
 /// </summary>
 public sealed class ActorStateDocument
 {


dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/CosmosActorStateStorage.cs
```diff
@@ -0,0 +1,323 @@
...
+    public async ValueTask<WriteResponse> WriteStateAsync(
+       ActorId actorId,
+       IReadOnlyCollection<ActorStateWriteOperation> operations,
+       string etag,
+       CancellationToken cancellationToken = default)
+    {
+        if (operations.Count == 0)
+        {
+            // No operations to perform - return success with current ETag or generate new one
+            string resultEtag = !string.IsNullOrEmpty(etag) ? etag : Guid.NewGuid().ToString("N");
```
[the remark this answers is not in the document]
ReubenBond: Bad copilot... there are 3 options:
- Return the input etag
- Throw InvalidOperationException (no operations specified)
- Perform a read instead and get the eTag (but... reading what)

I prefer the second option - throwing.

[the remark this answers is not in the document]
adityamandaleeka: I agree that the second is the best from the storage impl perspective. I did have to modify InProcessActorContext as well since right now that can submit an empty set of operations in cases where there are no ActorStateWriteOperations in its operation batch. For now, I just made that code treat this case as success and keep the same etag. I guess that code will change soon anyway once we turn more things into storage writes.

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/CosmosActorStateStorage.cs
```diff
@@ -0,0 +1,301 @@
...
+    /// <summary>
+    /// Writes state changes to the actor's persistent storage.
+    /// </summary>
+    public async ValueTask<WriteResponse> WriteStateAsync(
+       ActorId actorId,
+       IReadOnlyCollection<ActorStateWriteOperation> operations,
+       string etag,
+       CancellationToken cancellationToken = default)
+    {
+        var container = await this._lazyContainer.GetContainerAsync().ConfigureAwait(false);
```
stephentoub: Nit: move to after the below argument validation?

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/LazyCosmosContainer.cs
```diff
@@ -0,0 +1,80 @@
...
+
+    /// <summary>
+    /// Gets the Container, initializing it if necessary.
+    /// </summary>
+    public Task<Container> GetContainerAsync() => this._lazyContainer.Value;
+
+    private async Task<Container> InitializeContainerAsync()
+    {
+        // Create database if it doesn't exist
+        var database = await this._cosmosClient!.CreateDatabaseIfNotExistsAsync(this._databaseName!).ConfigureAwait(false);
```
ReubenBond, line 57: I think this is fine for now but worth noting that if there is a transient failure when first resolving the container, it will become a permanent failure and require an app restart to rectify. Retries at this level can help there. If the class implements `IAsyncDisposable`, then it can retry indefinitely (with appropriate backoff, etc) and use a `CancellationTokenSource` that is canceled in `DisposeAsync` to tell it when to quit trying.

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/LazyCosmosContainer.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        var containerProperties = new ContainerProperties(this._containerName!, "/actorId")
+        {
+            Id = this._containerName!,
+            IndexingPolicy = new IndexingPolicy
+            {
+                IndexingMode = IndexingMode.Consistent,
+                Automatic = true
+            },
+            PartitionKeyPaths = ["/actorId"]
```
ReubenBond, line 67: I thought we were going to separate the partition key into separate `actorType` and `actorKey` properties and make the partition hierarchy `["/actorType", "/actorKey"]` so that users can potentially query for all actors of a given type.

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/CosmosActorStateStorage.cs
```diff
@@ -0,0 +1,301 @@
...
+
+        // Read root document first to get actor-level ETag
+        string actorETag = await this.GetActorETagAsync(container, actorId, cancellationToken).ConfigureAwait(false);
+
+        foreach (var op in operations)
+        {
+            switch (op)
+            {
+                case GetValueOperation get:
+                    var id = GetDocumentId(actorId, get.Key);
```
ReubenBond: This can just be 'get.Key', possibly with some known prefix like `$"state_{get.Key}"` so that we can add system keys later without clashing with application keys for a given actor. Document ids are unique within a given partition already, and there is a logical partition per actor

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/CosmosActorStateStorage.cs
```diff
@@ -0,0 +1,301 @@
...
+
+        [JsonPropertyName("key")]
+        public string Key { get; set; } = default!;
+
+        [JsonPropertyName("value")]
+        public JsonElement Value { get; set; } = default!;
+    }
+
+    private static string Sanitize(string input)
+    => Uri.EscapeDataString(input).Replace("%2F", "__");
```
ReubenBond: Perhaps our sanitization routine in Orleans is overly complex. IIRC it creates ids where the original input can be recovered (unsanitized). Here it is for reference: https://github.com/dotnet/orleans/blob/dd821c52d0a1123df7391461341592bd33fafbbf/src/Azure/Shared/Cosmos/CosmosIdSanitizer.cs#L15

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/CosmosActorStateStorage.cs
ReubenBond: Similarly for the root document id - it can be a well-known id

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/CosmosActorStateStorage.cs
adityamandaleeka: I am happy to just copy that over, especially since it's been battle-tested.

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/LazyCosmosContainer.cs
adityamandaleeka: Yea I left that for a follow up given that Aspire 9.4 added support for HPK and was about to come out (it's out now).

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/CosmosActorStateStorage.cs
adityamandaleeka: Good point.

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/LazyCosmosContainer.cs
ReubenBond: ok, sgtm. We should log it as an issue though since it would be a breaking change. I.e, we should do it before this is released and used in production.

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime/InProcessActorContext.cs
```diff
@@ -239,11 +239,23 @@ public async ValueTask<WriteResponse> WriteAsync(ActorWriteOperationBatch operat
...
-            operations.ETag,
-            cancellationToken).ConfigureAwait(false);
+        IReadOnlyCollection<ActorStateWriteOperation> writeOps =
+            [.. operations.Operations.OfType<ActorStateWriteOperation>()];
+
+        WriteResponse result;
+        if (writeOps.Count == 0)
+        {
+            // Nothing to write
+            result = new WriteResponse(operations.ETag, success: true);
```
ReubenBond, line 249: I think this should be an error instead. i.e, InvalidOperationException

dotnet/tests/CosmosDB.IntegrationTests/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB.Tests.AppHost/CosmosDBTestConstants.cs
```diff
@@ -0,0 +1,13 @@
...
+namespace Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB.Tests;
+
+internal static class CosmosDBTestConstants
+{
+    public static string TestCosmosDbName = "ActorStateStorageTests";
+    public static string TestCosmosDbDatabaseName = "state-database";
+
+    // Set to use the CosmosDB emulator for testing.
+    // Warning: Using the emulator may cause test flakiness.
+    public static bool UseEmulatorForTesting = false;
```
ReubenBond: This should probably be set via an env var or similar.
See https://github.com/dotnet/orleans/blob/6039bacc10d1257eb7effaf7bfb2906ab90abdeb/test/Extensions/Tester.Cosmos/CosmosTestUtils.cs#L7
and
https://github.com/dotnet/orleans/blob/6039bacc10d1257eb7effaf7bfb2906ab90abdeb/test/TestInfrastructure/TestExtensions/TestDefaultConfiguration.cs#L10
for prior "art" (if you can call it that)

dotnet/tests/CosmosDB.IntegrationTests/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB.Tests.AppHost/AppHost.cs
```diff
@@ -0,0 +1,23 @@
...
+var cosmosDb = builder.AddAzureCosmosDB(CosmosDBTestConstants.TestCosmosDbName);
+
+if (CosmosDBTestConstants.UseEmulatorForTesting)
+{
+    cosmosDb.RunAsEmulator(emulator => emulator.WithLifetime(ContainerLifetime.Persistent));
+}
+else
+{
+    var cosmosDbResource = builder.AddParameterFromConfiguration("CosmosDbName", "CosmosDb:Name");
+    var cosmosDbResourceGroup = builder.AddParameterFromConfiguration("CosmosDbResourceGroup", "CosmosDb:ResourceGroup");
```
ReubenBond, line 14: If these params are not set, then we can use the emulator instead, for ex, or we can use different launch profiles.

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime/InProcessActorContext.cs
adityamandaleeka: We run into this case today with the HelloHttpApi sample; there are other ops here, just not ActorStateWrite ones. I think in that case we still want the code below to execute which is why I didn't make it throw here. Do you think we should refactor this?

ReubenBond: Yeah, we should fix the sample - it should avoid calling this API when it would be a no-op

ReubenBond: It can be fixed in a later PR

dotnet/tests/CosmosDB.IntegrationTests/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB.Tests.AppHost/AppHost.cs
adityamandaleeka: I kinda prefer it to fail here actually... if you're running the tests and you intended non-emulator I don't think you want them to fall back to the emulator.

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/LazyCosmosContainer.cs
adityamandaleeka: Good call, will fix in a follow up PR.

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/LazyCosmosContainer.cs
adityamandaleeka: Done. https://github.com/microsoft/agent-framework/issues/305

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime/InProcessActorContext.cs
adityamandaleeka: Logged an issue.

dotnet/Directory.Packages.props
```diff
@@ -21,6 +24,9 @@
     <PackageVersion Include="OpenTelemetry.Instrumentation.AspNetCore" Version="1.12.0" />
     <PackageVersion Include="OpenTelemetry.Instrumentation.Http" Version="1.12.0" />
     <PackageVersion Include="OpenTelemetry.Instrumentation.Runtime" Version="1.12.0" />
+    <PackageVersion Include="Microsoft.Azure.Cosmos" Version="3.52.0" />
+    <!-- Newtonsoft (Required by CosmosClient) -->
+    <PackageVersion Include="Newtonsoft.Json" Version="13.0.3" />
```
stephentoub, line 29: Ouch. Even if the Azure library is using Newtonsoft, why do we need to reference it? Does it use Newtonsoft types in its public surface area?

dotnet/samples/HelloHttpApi/HelloHttpApi.ApiService/HostApplicationBuilderAgentExtensions.cs
```diff
@@ -26,8 +28,16 @@ public static IHostApplicationBuilder AddAIAgent(this IHostApplicationBuilder bu
...
         });
+
         var actorBuilder = builder.AddActorRuntime();
 
+        // Add CosmosDB state storage to override default storage
+        builder.Services.AddCosmosActorStateStorage(serviceProvider =>
+        {
+            var cosmosClient = serviceProvider.GetRequiredService<CosmosClient>();
+            return new LazyCosmosContainer(cosmosClient, "actor-state-db", "ActorState");
+        });
```
stephentoub: Previously (I think) the sample didn't require any other resources, beyond an Azure OpenAI resource, in order to run. Does this change that?

dotnet/Directory.Packages.props
adityamandaleeka: Yea I was surprised by this too: https://github.com/Azure/azure-cosmos-dotnet-v3/issues/4900#issuecomment-2491774097

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/LazyCosmosContainer.cs
```diff
@@ -0,0 +1,80 @@
...
+namespace Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB;
+
+#pragma warning disable VSTHRD011 // Use AsyncLazy<T>
+
+/// <summary>
+/// A lazy wrapper around a Cosmos DB Container.
+/// This avoids performing async I/O-bound operations (i.e. Cosmos DB setup) during
+/// DI registration, deferring them until first access.
+/// </summary>
+public sealed class LazyCosmosContainer
```
stephentoub: Does this have to be public?

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/CosmosActorStateStorage.cs
```diff
@@ -0,0 +1,297 @@
...
+        {
+            switch (op)
+            {
+                case SetValueOperation set:
+                    var docId = GetDocumentId(set.Key);
+
+                    var item = new ActorStateDocument
+                    {
+                        Id = docId,
+                        ActorId = actorId.ToString(),
```
stephentoub: Do this once before the loop?

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/CosmosActorStateStorage.cs
```diff
@@ -0,0 +1,297 @@
...
+
+        [JsonPropertyName("key")]
+        public string Key { get; set; } = default!;
+
+        [JsonPropertyName("value")]
+        public JsonElement Value { get; set; } = default!;
+    }
+
+    private static string GetDocumentId(string key) => $"state_{CosmosIdSanitizer.Sanitize(key)}";
+    private static string GetRootDocumentId() => "rootdoc";
```
stephentoub: Why isn't this just a private const? I was confused at call sites about code calling GetRootDocumentId() and using it, even in the case where there wasn't one yet

dotnet/samples/HelloHttpApi/HelloHttpApi.ApiService/HostApplicationBuilderAgentExtensions.cs
adityamandaleeka: Correct, and I think we should still have a sample that doesn't require anything other than AI resources in order to run (and maybe have the storage be configurable). There's another PR that's changing this sample around so we can sort that out when that goes in.

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/CosmosActorStateStorage.cs
```diff
@@ -0,0 +1,297 @@
...
+            // eTag provided - replace existing root document with eTag check
+            batch.ReplaceItem(rootDocId, newRoot, new TransactionalBatchItemRequestOptions { IfMatchEtag = etag });
+        }
+
+        try
+        {
+            var result = await batch.ExecuteAsync(cancellationToken).ConfigureAwait(false);
+            if (!result.IsSuccessStatusCode)
+            {
+                return new WriteResponse(eTag: string.Empty, success: false);
```
stephentoub, line 106: Is an empty etag the right answer here? We don't want to, for example, return the original etag that was passed in? Or is rootResult.ETag not valid in the failure case?

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/CosmosActorStateStorage.cs
```diff
@@ -0,0 +1,297 @@
...
+            }
+
+            // Get the ETag from the root document operation (last operation in batch)
+            var rootResult = result[result.Count - 1];
+            return new WriteResponse(eTag: rootResult.ETag, success: true);
+        }
+        catch (CosmosException)
+        {
+            // If any operation in the batch fails, we return failure
+            return new WriteResponse(eTag: string.Empty, success: false);
```
stephentoub, line 116: Same etag question here

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/CosmosActorStateStorage.cs
```diff
@@ -0,0 +1,297 @@
...
+        }
+    }
+
+    /// <summary>
+    /// Reads state data from the actor's persistent storage.
+    /// </summary>
+    public async ValueTask<ReadResponse> ReadStateAsync(
+    ActorId actorId,
+    IReadOnlyCollection<ActorStateReadOperation> operations,
+    CancellationToken cancellationToken = default)
```
stephentoub: nit: indentation

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/CosmosActorStateStorage.cs
```diff
@@ -0,0 +1,297 @@
...
+                    else
+                    {
+                        query = new QueryDefinition("SELECT c.key FROM c WHERE c.actorId = @actorId AND c.key != null")
+                            .WithParameter("@actorId", actorId.ToString());
+                    }
+
+                    var requestOptions = new QueryRequestOptions
+                    {
+                        PartitionKey = GetPartitionKey(actorId),
+                        MaxItemCount = 100 // TODO Fix
```
stephentoub: Did you intend to leave this TODO here?

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/CosmosActorStateStorage.cs
```diff
@@ -0,0 +1,297 @@
...
+                            id,
+                            GetPartitionKey(actorId),
+                            cancellationToken: cancellationToken)
+                            .ConfigureAwait(false);
+
+                        results.Add(new GetValueResult(response.Resource.Value));
+                    }
+                    catch (CosmosException ex) when (ex.StatusCode == System.Net.HttpStatusCode.NotFound)
+                    {
+                        results.Add(new GetValueResult(null));
```
stephentoub, line 157: I'm curious about this error handling... we use null to indicate that the item wasn't found?

And then other CosmosExceptions are allowed to propagate?

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/CosmosActorStateStorage.cs
```diff
@@ -0,0 +1,297 @@
...
+                    }
+                    break;
+
+                case ListKeysOperation list:
+                    var keys = new List<string>();
+                    string? continuationToken = null;
```
stephentoub: nit: you could move both of these down about 20 lines closer to where they're used

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/CosmosActorStateStorage.cs
```diff
@@ -0,0 +1,297 @@
...
+
+                    var iterator = container.GetItemQueryIterator<KeyProjection>(
+                        query,
+                        list.ContinuationToken,
+                        requestOptions);
+
+                    while (iterator.HasMoreResults)
+                    {
+                        var page = await iterator.ReadNextAsync(cancellationToken).ConfigureAwait(false);
+                        foreach (var projection in page)
```
stephentoub, line 192: It's interesting that this is surfaced as enumerable pages. Is there no other API that returns a single async enumerable and that handles all the paging under the covers?

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/CosmosActorStateStorage.cs
```diff
@@ -0,0 +1,297 @@
...
+        }
+
+        return new ReadResponse(actorETag, results);
+    }
+
+    [SuppressMessage("Performance", "CA1812", Justification = "This is a projection class for Cosmos DB queries.")]
+    private sealed class KeyProjection
+    {
+        [JsonPropertyName("key")]
+        public string Key { get; set; } = default!;
```
stephentoub: Is Key guaranteed to be set to a non-null string by GetItemQueryIterator?

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/CosmosActorStateStorage.cs
```diff
@@ -0,0 +1,297 @@
...
+            var rootResponse = await container.ReadItemAsync<ActorRootDocument>(
+                rootDocId,
+                GetPartitionKey(actorId),
+                cancellationToken: cancellationToken).ConfigureAwait(false);
+            return rootResponse.ETag;
+        }
+        catch (CosmosException ex) when (ex.StatusCode == System.Net.HttpStatusCode.NotFound)
+        {
+            // No root document means no actor state exists
+            return Guid.NewGuid().ToString("N");
```
stephentoub, line 234: It's ok for this to manufacture an etag? I was under the impression only the runtime should be doing that. Are these state stores delegated that responsibility by the runtime for the state they manage?

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/CosmosIdSanitizer.cs
```diff
@@ -0,0 +1,133 @@
...
+        int count = 0;
+        foreach (var c in input)
+        {
+            if (SanitizedCharacters.IndexOf(c) >= 0)
+            {
+                count++;
+            }
+        }
+        return count;
+    }
```
stephentoub, line 72: We don't yet build net10.0 assets, but for future reference, on .NET 10 this could just be:
```C#
return input.CountAny(SanitizedCharacters);
```
which would then end up being even better if SanitizedCharacters were a `SearchValues<char>` instead of a `ReadOnlySpan<char>`.

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/CosmosIdSanitizer.cs
```diff
@@ -0,0 +1,133 @@
...
+// not allowed in Cosmos DB IDs with a safe escape sequence. This implementation was
+// inspired heavily by the one in Orleans, with some modifications made to enable
+// targeting NET472.
+internal static class CosmosIdSanitizer
+{
+    private const char EscapeChar = '~';
+    public const char SeparatorChar = '_';
+
+    private static ReadOnlySpan<char> SanitizedCharacters => new[] { '/', '\\', '?', '#', SeparatorChar, EscapeChar };
+    private static ReadOnlySpan<char> ReplacementCharacters => new[] { '0', '1', '2', '3', '4', '5' };
```
stephentoub: Nit: these can use collection expression syntax... becomes more obvious it's non-allocating on each property access

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/CosmosIdSanitizer.cs
stephentoub: And on other core builds before 10 it could be a loop around Count

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,75 @@
...
+    {
+        // Register CosmosClient as singleton
+        services.AddSingleton<CosmosClient>(serviceProvider =>
+        {
+            var cosmosClientOptions = new CosmosClientOptions
+            {
+                ApplicationName = "AgentFramework",
+                ConnectionMode = ConnectionMode.Direct,
+                ConsistencyLevel = ConsistencyLevel.Session,
+                UseSystemTextJsonSerializerWithOptions = new JsonSerializerOptions()
```
stephentoub: What's getting serialized? Should we be using a JSO for a source generated JSON context for whatever types we're passing off to be serialized? I don't know if the cosmos client is NativeAOT safe yet, but it'd be nice if we're not introducing additional debt there.

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime/InProcessActorContext.cs
stephentoub, line 249: IIRC from earlier in the PR, the implementation was throwing an exception if passed an empty collection? Should that exception be allowed to propagate? Or if we want this to be non-exceptional, should that exception in the implementation be changed to just do what this is doing?

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime/InProcessActorContext.cs
```diff
@@ -239,11 +239,23 @@ public async ValueTask<WriteResponse> WriteAsync(ActorWriteOperationBatch operat
...
+        }
+        else
+        {
+            result = await this.Storage.WriteStateAsync(
+                this.ActorId,
+                [.. operations.Operations.OfType<ActorStateWriteOperation>()],
```
stephentoub:
```suggestion
                [.. writeOps],
```
?

dotnet/tests/CosmosDB.IntegrationTests/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB.Tests.AppHost/AppHost.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+using System.Runtime.CompilerServices;
+using Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB.Tests;
+
+[assembly: InternalsVisibleTo("Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB.Tests")]
```
stephentoub: Is this necessary? IVT and I are enemies

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime/InProcessActorContext.cs
adityamandaleeka: 🤦

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime/InProcessActorContext.cs
adityamandaleeka: Yea this is a temporary thing until this is fixed: https://github.com/microsoft/agent-framework/issues/306

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/CosmosActorStateStorage.cs
adityamandaleeka: Yes, the state stores are responsible for etag creation and management. The runtime just regurgitates what it's been told for optimistic concurrency control.

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/CosmosActorStateStorage.cs
adityamandaleeka: We control the write path and always write a key, and the queries also include a key != null check so it should be safe.

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/CosmosActorStateStorage.cs
adityamandaleeka: I suppose we could return the eTag that was passed in. Or maybe it would be better to return the current known state of the actor (if the failure was indeed because of an eTag mismatch), but I don't know if there's a way to get that without doing another read.

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/CosmosActorStateStorage.cs
adityamandaleeka: It can be now, updated.

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/LazyCosmosContainer.cs
adityamandaleeka: I think we can keep it internal with some refactoring. Will push an update soon.

adityamandaleeka: Done.

dotnet/tests/CosmosDB.IntegrationTests/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB.Tests.AppHost/AppHost.cs
adityamandaleeka: Not really... the constants we use from the other test project can just be made public.

dotnet/tests/CosmosDB.IntegrationTests/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB.Tests.AppHost/CosmosDBTestConstants.cs
adityamandaleeka: I made it an env var now but left it explicit (i.e. no fallbacks).

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/CosmosActorStateStorage.cs
adityamandaleeka: Yes, that's the intended behavior. GetValueResult is expected to give you null when the value isn't found.

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/CosmosActorStateStorage.cs
adityamandaleeka: AFAICT they don't have one. It's fairly straightforward to wrap if we want to I suppose.

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/CosmosActorStateStorage.cs
adityamandaleeka: I think we can leave it at the default (-1) for now and let it decide dynamically what the size should be. If it becomes a perf problem later we can sort it out.

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/ServiceCollectionExtensions.cs
adityamandaleeka: Done.

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/CosmosActorStateStorage.cs
adityamandaleeka: I don't think there's a way to get the current eTag without doing another read. I'll leave this as is for now.

dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB/CosmosActorStateStorage.cs
adityamandaleeka: Same answer :) I don't think there's a way to get the current eTag without doing another read. I'll leave this as is for now.

PR #271, opened 2025-07-29
Adding draft version of workflows design doc.

alliscode:
Adding initial draft of updated workflows design document. This document outlines the goals, design principles, and current thinking on coding patterns.


docs/design/workflows_updated.md
```diff
@@ -0,0 +1,636 @@
...
+    .add_fan_in_edges(mappers, shuffler, activation=Activation.WhenAll)
+    .add_fan_out_edges(shuffler, reducers)
+    .add_fan_in_edges(reducers, aggregator, activation=Activation.WhenAll)
+    .build()
+)
+```
+
+### 4. Conditional Branching
+
+Dynamic routing based on message content.
```
alliscode: We will need conditional branching based on shared state (user variables) as well.

docs/design/workflows_updated.md
```diff
@@ -0,0 +1,636 @@
...
+- **Direct routing**: Message flows if type matches and condition passes
+- **Fan-out**: Single source sends to multiple targets
+- **Fan-in**: Multiple sources synchronize at single target
+- **Conditional**: Routing based on message content
+
+### Concurrency Model
+
+- Executors within a superstep run concurrently
+- Messages from different sources are delivered in parallel
+- Shared state access is protected by locks
```
alliscode: We should also have transactions between super step boundaries for state changes including user variables.

docs/design/workflows_updated.md
```diff
@@ -0,0 +1,636 @@
...
+# Conditional routing
+workflow = (
+    WorkflowBuilder()
+    .add_edge(router, executor_a, lambda msg: msg.type == "A")
+    .add_edge(router, executor_b, lambda msg: msg.type == "B")
+    .set_start_executor(router)
+    .build()
+)
+
+# Fan-out/Fan-in pattern
```
alliscode: This seems more like map-reduce then fan-out/in. I tend to think of fan out as sending the same message to multiple destinations e.g. taking a single document and getting multiple viewpoints on it from different agents in parallel.

Map-reduce starts with a collection of items like List<T> which is then split up into batches and sent to different nodes for processing in parallel. The main difference being that each processor is getting different instances of type T rather than the same instance. An example might be getting a list of tasks and then having a different agent work on each of them in parallel.

docs/design/workflows_updated.md
```diff
@@ -0,0 +1,636 @@
...
+### Event Handling
+
+Events are emitted during execution and streamed to consumers:
+
+```python
+async for event in workflow.run_stream(message):
+    match event:
+        case ExecutorCompleteEvent(executor_id=id, data=result):
+            logger.info(f"Executor {id} completed with result: {result}")
+        case HumanInTheLoopEvent(executor_id=id):
```
alliscode: Are these event types defined by the framework or the user of both?

docs/design/workflows_updated.md
```diff
@@ -0,0 +1,636 @@
...
+### Usage in Executors
+
+```python
+class StatefulExecutor(Executor[str]):
+    async def _execute(self, data: str, ctx: ExecutorContext) -> None:
+        # Read from shared state
+        counter = await ctx.get_shared_state("counter") or 0
+        
+        # Update shared state
+        await ctx.set_shared_state("counter", counter + 1)
```
alliscode: What happens if another executor mutates `counter` between the get and set? Or is the `counter` state locked once it's read?

docs/design/workflows_updated.md
```diff
@@ -0,0 +1,636 @@
...
+### 1. Edge Groups and Synchronization
+
+Fan-in edges can be grouped for synchronized delivery:
+
+```python
+# All three workers must complete before aggregator runs
+workflow = (
+    WorkflowBuilder()
+    .add_fan_in_edges([worker1, worker2, worker3], aggregator,
+                      activation=Activation.WhenAll)
```
alliscode: We will need to have configurable stream fusion interpolators as well. See https://github.com/microsoft/psi/wiki/Stream-Fusion-and-Merging

Update: I no longer think we need interpolators. Fan-in with WhenAll queues all messages until there is at least one message from each edge, and then it collects them all into a list and calls the executor. So the framework does not need to pair elements from multiple streams.

docs/design/workflows_updated.md
```diff
@@ -0,0 +1,636 @@
...
+
+### Integration Pattern
+
+```python
+# Main execution loop with HIL support
+hil_event = None
+while True:
+    if hil_event:
+        # Resume at specific executor with human input
+        events = workflow.run_stream(human_input, executor=hil_event.executor_id)
```
alliscode: It feels weird to target an executor within the workflow, even though it's via an opaque identifier. Is there a world where we have an event type that understands it is a request and needs an associated response, and then instead of `workflow.run_stream(human_input, executor=hil_event.executor_id)` you would do something like:

`workflow.run_stream(response=human_input)`

docs/design/workflows_updated.md
```diff
@@ -0,0 +1,636 @@
...
+    .add_edge(router, executor_b, lambda msg: msg.type == "B")
+    .set_start_executor(router)
+    .build()
+)
+
+# Fan-out/Fan-in pattern
+workflow = (
+    WorkflowBuilder()
+    .set_start_executor(splitter)
+    .add_fan_out_edges(splitter, [worker1, worker2, worker3])
```
ekzhu: This sends the same value to all the workers. What does it look like if we want to send different values to different workers?

docs/design/workflows_updated.md
```diff
@@ -0,0 +1,636 @@
...
+
+### 4. WorkflowBuilder (`workflow.py`)
+
+Fluent API for constructing workflows.
+
+```python
+class WorkflowBuilder:
+    """Builder pattern for workflow construction"""
+    
+    def add_edge(self, source: Executor, target: Executor,
```
ekzhu: We also need to be able to add condition groups, in which there are multiple edges and their conditions, but only one edge is going to be triggered. It also can have a default edge.

ekzhu: `add_conditional_edges` ?

docs/design/workflows_updated.md
alliscode: Jacob suggests using a tool calling structure

docs/design/workflows_updated.md
ekzhu: Make the condition a union type to support `Callable[[Message, ExecutionContext], bool]`

docs/design/workflows_updated.md
lokitoth: If we add a way to set a conditional-per-target, or have an explicit "scheduler" Callable[[msgT], int] which returns the index of the target, it would let us split the outgoing message stream.

In either case, the implementor would add a PartitionKey (or similar) on the outgoing message, and the individual conditionals or the combined scheduler would key off it.

docs/design/workflows_updated.md
```diff
@@ -0,0 +1,636 @@
...
+
+### 1. Edge Groups and Synchronization
+
+Fan-in edges can be grouped for synchronized delivery:
+
+```python
+# All three workers must complete before aggregator runs
+workflow = (
+    WorkflowBuilder()
+    .add_fan_in_edges([worker1, worker2, worker3], aggregator,
```
alliscode: More thought needed, how do we simplify for v1

docs/design/workflows_updated.md
eavanvalkenburg: do we need some way of deciding which executor can go first/has priority?

docs/design/workflows_updated.md
```diff
@@ -0,0 +1,636 @@
...
+## API Design
+
+### Creating Executors
+
+```python
+@output_message_types(str)
+class UpperCaseExecutor(Executor[str]):
+    async def _execute(self, data: str, ctx: ExecutorContext) -> str:
+        result = data.upper()
+        await ctx.send_message(result)
```
eavanvalkenburg: what is the purpose of the two returns with the same value? couldn't we smooth that out in the `executor` method?

docs/design/workflows_updated.md
```diff
@@ -0,0 +1,636 @@
...
+        return result
+```
+
+### Building Workflows
+
+```python
+# Sequential workflow
+workflow = (
+    WorkflowBuilder()
```
eavanvalkenburg: we should be able to do (by using class methods):
```suggestion
    WorkflowBuilder
```

docs/design/workflows_updated.md
```diff
@@ -0,0 +1,636 @@
...
+    .build()
+)
+
+# Fan-out/Fan-in pattern
+workflow = (
+    WorkflowBuilder()
+    .set_start_executor(splitter)
+    .add_fan_out_edges(splitter, [worker1, worker2, worker3])
+    .add_fan_in_edges([worker1, worker2, worker3], aggregator,
+                      activation=Activation.WhenAll)
```
eavanvalkenburg: what happens with `WhenAny` to the rest of the messages? are they passed to the aggregator in subsequent supersteps?

docs/design/workflows_updated.md
```diff
@@ -0,0 +1,636 @@
...
+    .build()
+)
+```
+
+### Running Workflows
+
+```python
+# Stream execution with event handling
+async for event in workflow.run_stream(initial_message):
```
eavanvalkenburg: we use run_streaming (already in dotnet, PR open for that change in Py) in Agents, so not sure if the same verbiage should be used here.

docs/design/workflows_updated.md
```diff
@@ -0,0 +1,636 @@
...
+# Workflow lifecycle events
+WorkflowStartedEvent    # Workflow execution begins
+WorkflowCompletedEvent  # Workflow reaches completion
+
+# Executor events
+ExecutorInvokeEvent     # Executor starts processing
+ExecutorCompleteEvent   # Executor finishes processing
+
+# Agent-specific events
+AgentRunEvent          # Agent produces final response
```
eavanvalkenburg: there has been no other mentions of Agents in this design, so why a AgentRunEvent?

docs/design/workflows_updated.md
alliscode: I believe that if two executors within the same super step need to coordinate the order of execution, then the workflow has been incorrectly defined. In theory, the dependencies of an executor should be defined by its input edges and once those are satisfied it should have everything it needs to run.

If you can think of a scenario where this is not good enough, please let me know.

docs/design/workflows_updated.md
alliscode: Great question. If there are multiple messages queued for a fan_in with WhenAny, the runtime will iterate over them and run the executor one for each sequentially. See sequence diagram I've just added.

docs/design/workflows_updated.md
alliscode: Yes, I think we should align on the same verbiage

docs/design/workflows_updated.md
alliscode: After more thought, I think this might be simple enough

docs/design/workflows_updated.md
```diff
@@ -0,0 +1,636 @@
...
+    .build()
+)
+```
+
+### 2. Type System Integration
+
+The framework leverages Python's type system for safety:
+
+```python
+def _is_instance_of(data: Any, target_type: type) -> bool:
```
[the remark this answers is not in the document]
alliscode: This design doc will discuss internal workings of the workflow framework.

docs/design/workflows_updated.md
TaoChenOSU: They need to `hold` the state if they need consistency for a sequence of operations.

docs/design/workflows_updated.md
TaoChenOSU: Remodeled as `send_reponses_stream`.

docs/design/workflows_updated.md
TaoChenOSU: The builder needs to store properties of the workflow for building so an instance of the builder is needed.

docs/design/workflows_updated.md
TaoChenOSU: Removed

docs/design/workflows_updated.md
TaoChenOSU: This feels weird. If we pass the execution context to the conditions, then in theory, the conditions can act as executors. 

I think we can pass an immutable copy of the shared states to the conditions.

docs/design/workflows_updated.md
TaoChenOSU: Moving the discussion to https://github.com/microsoft/agent-framework/pull/271#discussion_r2240998838

docs/design/workflows_updated.md
TaoChenOSU: @ekzhu Could you clarify on if we need `condition` groups where only one edge will be triggered or `switch` groups where multiple edges can be triggered?

docs/design/workflows_updated.md
TaoChenOSU: If we want to send different values to different workers, the executor can specify the target id when calling the send_message API, in which case the message will only be routed to the specified target.

docs/design/workflows_updated.md
```diff
@@ -0,0 +1,421 @@
...
+If an executor fails during a superstep, messages processed during that superstep will be rolled back. Updates to shared state will not be committed. This ensures that the workflow remains consistent and avoids partial updates.
+
+## API Design
+
+### Creating Executors
+
+```python
+class SampleExecutor(Executor):
+
+    @handles_message(output_types=[str])
```
ekzhu: Suggest renaming to `handler`, for a decorator is a noun rather than a verb. Also, in the design doc we specifically called out "handler" as a core concept, so the name should match.

docs/design/workflows_updated.md
```diff
@@ -0,0 +1,421 @@
...
+
+### 1. Templatized Workflows
+
+- Support for reusable workflow templates
+- High-level workflow definitions with parameterization and templatized WorkflowBuilder to allow for easy instantiation of common patterns
+
+### 2. Declarative Workflow Definitions
+
+- CSDL (Copilot Studio Definition Language) support for defining workflows
+- Visual workflow designer with drag-and-drop interface
```
ekzhu: The designer will be part of Foundry UI and VS Code Foundry extension, not part of this framework.

docs/design/workflows_updated.md
moonbox3: In our current implementation we have it named as `message_handler`.

TaoChenOSU: We can update the PR

docs/design/workflows_updated.md
ekzhu: conditions are in fact built-in executors. You can model it as an executor.

docs/design/workflows_updated.md
ekzhu: I think I meant a group of edges only one of them will be triggered.

docs/design/workflows_updated.md
alliscode: Can we pass the message and a read-only copy of the global state dictionary rather than the context?

docs/design/workflows_updated.md
TaoChenOSU: Is the following similar to what you are thinking?
```
switch(data):
  case condition_1:
     edge_1
     break
  case condition_2:
     edge_2
     break
  case condition_3:
     edge_3
     break
  default:
     edge_4
     break
```
or similarly
```
if condition_1:
  edge_1
elif condition_2:
  edge_2
elif condition_3:
  edge_3
else:
  edge_4
```
And the conditions don't have to be mutually exclusive.

ekzhu: Exactly. Both are equivalent.

PR #272, opened 2025-07-29
Python: Define Workflow and Executor APIs

### LINKED ISSUE #317: Workflow validation
We should be able to validate a workflow graph. The following should be validated:
1. Edge duplication
2. Edge compatibility (data type)

If validation fails, the workflow builder should throw an exception.

TaoChenOSU:
### Motivation and Context

First PR to bring workflow POC to Agent Framework Python.

### Description

The following concepts are introduced:
- Executor
- Edge
- Workflow
- WorkflowBuilder
- WorkflowEvent
- Closes #317

Samples are included in this initial PR.

### Contribution Checklist

### DIFF
--- a/python/packages/workflow/agent_framework_workflow/__init__.py
+++ b/python/packages/workflow/agent_framework_workflow/__init__.py
@@ -20,7 +20,7 @@ from ._executor import (
     Executor,
     RequestInfoExecutor,
     RequestInfoMessage,
-    message_handler,
+    handler,
 )
 from ._validation import (
     EdgeDuplicationError,
@@ -67,6 +67,6 @@ __all__ = [
     "WorkflowStartedEvent",
     "WorkflowValidationError",
     "__version__",
-    "message_handler",
+    "handler",
     "validate_workflow_graph",
 ]
--- a/python/packages/workflow/agent_framework_workflow/_executor.py
+++ b/python/packages/workflow/agent_framework_workflow/_executor.py
@@ -33,13 +33,13 @@ class Executor:
         """
         self._id = id or str(uuid.uuid4())
 
-        self._message_handlers: dict[type, Callable[[Any, WorkflowContext], Any]] = {}
+        self._handlers: dict[type, Callable[[Any, WorkflowContext], Any]] = {}
         self._discover_handlers()
 
-        if not self._message_handlers:
+        if not self._handlers:
             raise ValueError(
-                f"Executor {self.__class__.__name__} has no message handlers defined. "
-                "Please define at least one message handler using the @message_handler decorator."
+                f"Executor {self.__class__.__name__} has no handlers defined. "
+                "Please define at least one handler using the @handler decorator."
             )
 
     async def execute(self, message: Any, context: WorkflowContext) -> None:
@@ -53,9 +53,9 @@ class Executor:
             An awaitable that resolves to the result of the execution.
         """
         handler: Callable[[Any, WorkflowContext], Any] | None = None
-        for message_type in self._message_handlers:
+        for message_type in self._handlers:
             if is_instance_of(message, message_type):
-                handler = self._message_handlers[message_type]
+                handler = self._handlers[message_type]
                 break
 
         if handler is None:
@@ -76,12 +76,11 @@ class Executor:
             attr = getattr(self, attr_name)
             if callable(attr) and hasattr(attr, "_handler_spec"):
                 handler_spec = attr._handler_spec  # type: ignore
-                if self._message_handlers.get(handler_spec["message_type"]) is not None:
+                if self._handlers.get(handler_spec["message_type"]) is not None:
                     raise ValueError(
-                        f"Duplicate message handler for type {handler_spec['message_type']} "
-                        f"in {self.__class__.__name__}"
+                        f"Duplicate handler for type {handler_spec['message_type']} in {self.__class__.__name__}"
                     )
-                self._message_handlers[handler_spec["message_type"]] = attr
+                self._handlers[handler_spec["message_type"]] = attr
 
     def can_handle(self, message: Any) -> bool:
         """Check if the executor can handle a given message type.
@@ -92,25 +91,25 @@ class Executor:
         Returns:
             True if the executor can handle the message type, False otherwise.
         """
-        return any(is_instance_of(message, message_type) for message_type in self._message_handlers)
+        return any(is_instance_of(message, message_type) for message_type in self._handlers)
 
 
 # endregion: Executor
 
-# region: Message Handler Decorator
+# region: Handler Decorator
 
 
 ExecutorT = TypeVar("ExecutorT", bound="Executor")
 
 
 @overload
-def message_handler(
+def handler(
     func: Callable[[ExecutorT, Any, WorkflowContext], Awaitable[Any]],
 ) -> Callable[[ExecutorT, Any, WorkflowContext], Awaitable[Any]]: ...
 
 
 @overload
-def message_handler(
+def handler(
     func: None = None,
     *,
     output_types: list[type] | None = None,
@@ -120,7 +119,7 @@ def message_handler(
 ]: ...
 
 
-def message_handler(
+def handler(
     func: Callable[[ExecutorT, Any, WorkflowContext], Awaitable[Any]] | None = None,
     *,
     output_types: list[type] | None = None,
@@ -131,7 +130,7 @@ def message_handler(
         Callable[[ExecutorT, Any, WorkflowContext], Awaitable[Any]],
     ]
 ):
-    """Decorator to register a message handler for an executor.
+    """Decorator to register a handler for an executor.
 
     Args:
         func: The function to decorate. Can be None when using with parameters.
@@ -141,11 +140,11 @@ def message_handler(
         The decorated function with handler metadata.
 
     Example:
-        @message_handler
+        @handler
         async def handle_string(self, message: str, ctx: WorkflowContext) -> None:
             ...
 
-        @message_handler(output_types=[str, int])
+        @handler(output_types=[str, int])
         async def handle_data(self, message: dict, ctx: WorkflowContext) -> None:
             ...
     """
@@ -153,20 +152,20 @@ def message_handler(
     def decorator(
         func: Callable[[ExecutorT, Any, WorkflowContext], Awaitable[Any]],
     ) -> Callable[[ExecutorT, Any, WorkflowContext], Awaitable[Any]]:
-        # Extract the message type from a message handler function.
+        # Extract the message type from a handler function.
         sig = inspect.signature(func)
         params = list(sig.parameters.values())
 
         if len(params) != 3:  # self, message, ctx
-            raise ValueError(f"Message handler must have exactly 3 parameters, got {len(params)}")
+            raise ValueError(f"Handler must have exactly 3 parameters, got {len(params)}")
 
         message_type = params[1].annotation
         if message_type is inspect.Parameter.empty:
-            raise ValueError("Message handler's second parameter must have a type annotation")
+            raise ValueError("Handler's second parameter must have a type annotation")
 
         @functools.wraps(func)
         async def wrapper(self: ExecutorT, message: Any, ctx: WorkflowContext) -> Any:
-            """Wrapper function to call the message handler."""
+            """Wrapper function to call the handler."""
             return await func(self, message, ctx)
 
         wrapper._handler_spec = {  # type: ignore
@@ -182,7 +181,7 @@ def message_handler(
     return decorator(func)
 
 
-# endregion: Message Handler Decorator
+# endregion: Handler Decorator
 
 # region: Agent Executor
 
@@ -240,7 +239,7 @@ class AgentExecutor(Executor):
         self._streaming = streaming
         self._cache: list[ChatMessage] = []
 
-    @message_handler(output_types=[AgentExecutorResponse])
+    @handler(output_types=[AgentExecutorResponse])
     async def run(self, request: AgentExecutorRequest, ctx: WorkflowContext) -> None:
         """Run the agent executor with the given request."""
         self._cache.extend(request.messages)
@@ -300,7 +299,7 @@ class RequestInfoExecutor(Executor):
         super().__init__(id=self.EXECUTOR_ID)
         self._request_events: dict[str, RequestInfoEvent] = {}
 
-    @message_handler
+    @handler
     async def run(self, message: RequestInfoMessage, ctx: WorkflowContext) -> None:
         """Run the RequestInfoExecutor with the given message."""
         source_executor_id = ctx.get_source_executor_id()
--- a/python/packages/workflow/agent_framework_workflow/_validation.py
+++ b/python/packages/workflow/agent_framework_workflow/_validation.py
@@ -172,7 +172,7 @@ class WorkflowGraphValidator:
                     logger.warning(
                         f"Executor '{source_executor.id}' has no output type annotations. "
                         f"Type compatibility validation will be skipped for edges from this executor. "
-                        f"Consider adding output_types to @message_handler decorators for better validation."
+                        f"Consider adding output_types to @handler decorators for better validation."
                     )
                 if not target_input_types:
                     logger.warning(
@@ -245,9 +245,9 @@ class WorkflowGraphValidator:
         """
         input_types: list[type[Any]] = []
 
-        # Access the private _message_handlers attribute to get input types
-        if hasattr(executor, "_message_handlers"):
-            input_types.extend(executor._message_handlers.keys())  # type: ignore
+        # Access the private _handlers attribute to get input types
+        if hasattr(executor, "_handlers"):
+            input_types.extend(executor._handlers.keys())  # type: ignore
 
         return input_types
 
--- a/python/packages/workflow/agent_framework_workflow/_workflow.py
+++ b/python/packages/workflow/agent_framework_workflow/_workflow.py
@@ -260,12 +260,12 @@ class WorkflowBuilder:
         types of the source executors. For example:
 
             class Target(Executor):
-                @message_handler
+                @handler
                 def handle_messages(self, messages: list[Message]) -> None:
                     # Process the aggregated messages from all sources
 
             class Source(Executor):
-                @message_handler(output_type=[Message])
+                @handler(output_type=[Message])
                 def handle_message(self, message: Message) -> None:
                     # Send a message to the target executor
                     self.send_message(message)
--- a/python/samples/getting_started/workflow/step_04_simple_group_chat.py
+++ b/python/samples/getting_started/workflow/step_04_simple_group_chat.py
@@ -13,7 +13,7 @@ from agent_framework.workflow import (
     WorkflowBuilder,
     WorkflowCompletedEvent,
     WorkflowContext,
-    message_handler,
+    handler,
 )
 
 """
@@ -32,7 +32,7 @@ class RoundRobinGroupChatManager(Executor):
         self._max_round = max_round
         self._current_round = 0
 
-    @message_handler(output_types=[AgentExecutorRequest])
+    @handler(output_types=[AgentExecutorRequest])
     async def start(self, task: str, ctx: WorkflowContext) -> None:
         """Execute the task by sending messages to the next executor in the round-robin sequence."""
         initial_message = ChatMessage(ChatRole.USER, text=task)
@@ -52,7 +52,7 @@ class RoundRobinGroupChatManager(Executor):
             target_id=self._get_next_member(),
         )
 
-    @message_handler(output_types=[AgentExecutorRequest])
+    @handler(output_types=[AgentExecutorRequest])
     async def handle_agent_response(self, response: AgentExecutorResponse, ctx: WorkflowContext) -> None:
         """Execute the task by sending messages to the next executor in the round-robin sequence."""
         # Send the response to the other members
--- a/python/samples/getting_started/workflow/step_05_simple_group_chat_with_hil.py
+++ b/python/samples/getting_started/workflow/step_05_simple_group_chat_with_hil.py
@@ -15,7 +15,7 @@ from agent_framework.workflow import (
     WorkflowBuilder,
     WorkflowCompletedEvent,
     WorkflowContext,
-    message_handler,
+    handler,
 )
 
 """
@@ -34,7 +34,7 @@ class CriticGroupChatManager(Executor):
         self._current_round = 0
         self._chat_history: list[ChatMessage] = []
 
-    @message_handler(output_types=[AgentExecutorRequest])
+    @handler(output_types=[AgentExecutorRequest])
     async def start(self, task: str, ctx: WorkflowContext) -> None:
         """Handler that starts the group chat with an initial task."""
         initial_message = ChatMessage(ChatRole.USER, text=task)
@@ -51,7 +51,7 @@ class CriticGroupChatManager(Executor):
         # Update the cache with the initial message
         self._chat_history.append(initial_message)
 
-    @message_handler(output_types=[AgentExecutorRequest, RequestInfoMessage])
+    @handler(output_types=[AgentExecutorRequest, RequestInfoMessage])
     async def handle_agent_response(self, response: AgentExecutorResponse, ctx: WorkflowContext) -> None:
         """Handler that processes the response from the agent."""
         # Update the chat history with the response
@@ -74,7 +74,7 @@ class CriticGroupChatManager(Executor):
         selection = self._get_next_member()
         await ctx.send_message(AgentExecutorRequest(messages=[], should_respond=True), target_id=selection)
 
-    @message_handler(output_types=[AgentExecutorRequest])
+    @handler(output_types=[AgentExecutorRequest])
     async def handle_request_response(self, response: list[ChatMessage], ctx: WorkflowContext) -> None:
         """Handler that processes the response from the RequestInfoExecutor."""
         # Update the chat history with the response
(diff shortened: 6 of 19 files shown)


crickman:
Will unit-tests be included?

python/samples/getting_started/workflow/step_01_simple_executor.py
```diff
@@ -0,0 +1,31 @@
+# Copyright (c) Microsoft. All rights reserved.
+
+import asyncio
+import sys
+
+from agent_framework.workflow import Executor, ExecutorContext
```
ekzhu: Should be `WorkflowContext`. And how come it is not triggering type checking error?

python/packages/workflow/agent_framework_workflow/__init__.py
```diff
@@ -0,0 +1,42 @@
...
+    "ExecutorContext",
+    "ExecutorEvent",
+    "ExecutorInvokeEvent",
+    "HumanInTheLoopEvent",
+    "WorkflowBuilder",
+    "WorkflowCompletedEvent",
+    "WorkflowEvent",
+    "WorkflowStartedEvent",
+    "__version__",
+    "events",
```
eavanvalkenburg: do we need to expose both `events` and the individual event types like `WorkflowCompletedEvent`? and depending on this, events.py could be `_events.py` to keep things consistent

python/packages/workflow/agent_framework_workflow/core/__init__.py
eavanvalkenburg, line 1: Do we need this core folder, flat is better then nested...

python/samples/getting_started/workflow/step_01_simple_executor.py
eavanvalkenburg, line 1: the samples need a lot more explanation!

python/samples/getting_started/workflow/step_02a_simple_workflow_condition.py
```diff
@@ -0,0 +1,90 @@
...
+    """An executor that responds to a message based on spam detection."""
+
+    @override
+    async def _execute(self, data: bool, ctx: ExecutorContext) -> None:
+        """Respond with a message based on whether the input is spam."""
+        if data is True:
+            raise RuntimeError("Input is spam, cannot respond.")
+
+        # Simulate processing delay
+        await asyncio.sleep(1)
```
eavanvalkenburg: would be good to show how this step can access the actual input of the workflow!

python/samples/getting_started/workflow/step_02b_simple_workflow_loop.py
```diff
@@ -0,0 +1,124 @@
...
+            self._guess = (self._lower + self._upper) // 2
+            await ctx.send_message(self._guess)
+            return self._guess
+
+        # The previous guess was too high.
+        # Update the upper bound to the previous guess.
+        # Generate a new number that is between the new bounds.
+        self._upper = self._guess - 1
+        self._guess = (self._lower + self._upper) // 2
+        await ctx.send_message(self._guess)
```
eavanvalkenburg: why do I need to both use the ctx and the return?

python/samples/getting_started/workflow/step_03_simple_group_chat.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+@output_message_types(list[ChatMessage])
+class FakeAgentExecutor(Executor[AgentSelectionDecision]):
+    """An executor that simulates a group chat agent A."""
+
+    @override
+    async def _execute(self, data: AgentSelectionDecision, ctx: ExecutorContext) -> list[ChatMessage]:
+        """Simulate a response."""
+        response = ChatResponse(
```
eavanvalkenburg: why not put a actual ChatClientAgent in here?

python/packages/workflow/agent_framework_workflow/core/workflow.py
```diff
@@ -0,0 +1,211 @@
...
+                group_ids: list[str] = []
+                group_ids.extend([e.id for e in edges[0:i]])
+                group_ids.extend([e.id for e in edges[i + 1 :]])
+                edge.set_edge_group(group_ids)
+
+        self._edges.extend(edges)
+
+        return self
+
+    def add_loop(
```
ekzhu: I am ambivalent about this method. I feel it doesn't simplify much -- only remove 1 line of code, but make the code harder to reason about. When you read the code 

```python
.add_loop(source, target, condition=...)
```

It is not very clear where is this condition applied to, without understanding the method's definition. So it is increasing the human cognitive load as it requires user to read the method definition, and the code becomes not self-explanatory. 

I would rather stick with 

```python
.add_edge(source, target, condition=...)
.add_edge(target, source)
```

for it expresses the loop using very foundational concepts that is known to everyone. Also it allows me to customize where to add the condition. I can choose to add condition on both edges if I want to.

ekzhu: Furthermore, `add_loop` as it is only supports adding loop between two nodes. If a loop involves more than two nodes we can't actually use this. So this method is limited in its utility.

python/packages/workflow/agent_framework_workflow/core/workflow.py
```diff
@@ -0,0 +1,211 @@
...
+        self._workflow_context = workflow_context
+        return self
+
+    def build(self) -> Workflow:
+        """Build and return the constructed workflow.
+
+        Returns:
+            A Workflow instance with the defined edges and starting executor.
+        """
+        if not self._start_executor:
```
eavanvalkenburg: could we do a convenience, maybe take the first executer from the first edge as the default start executor (with a log message stating doing that)?

python/packages/workflow/agent_framework_workflow/core/workflow.py
```diff
@@ -0,0 +1,211 @@
...
+            source: The source executor of the loop edge.
+            target: The target executor of the loop edge.
+            condition: An optional condition function that be applied on only the source -> target edge.
+        """
+        self._edges.append(Edge(source, target, condition))
+        self._edges.append(Edge(target, source))
+
+        return self
+
+    def add_chain(
```
ekzhu: This is great! I love this.

python/packages/workflow/agent_framework_workflow/core/workflow.py
```diff
@@ -0,0 +1,211 @@
...
+
+        self._edges.extend(edges)
+
+        return self
+
+    def add_loop(
+        self,
+        source: Executor[Any],
+        target: Executor[Any],
+        condition: Callable[[Any], bool] | None = None,
```
eavanvalkenburg: optionally conditions on both directions?

python/packages/workflow/agent_framework_workflow/core/workflow.py
```diff
@@ -0,0 +1,211 @@
...
+        """
+        for target in targets:
+            self._edges.append(Edge(source, target))
+        return self
+
+    def add_fan_in_edges(
+        self,
+        sources: list[Executor[Any]],
+        target: Executor[Any],
+        activation: Activation = Activation.WhenAll,
```
eavanvalkenburg:
```suggestion
        activation: Activation | Literal['all', 'any'] = Activation.WhenAll,
```

python/packages/workflow/agent_framework_workflow/core/workflow.py
```diff
@@ -0,0 +1,211 @@
...
+            target: The target executor of the edge.
+            condition: An optional condition function that determines whether the edge
+                       should be traversed based on the message type.
+        """
+        # TODO(@taochen): Support executor factories for lazy initialization
+        self._edges.append(Edge(source, target, condition))
+        return self
+
+    def add_fan_out_edges(self, source: Executor[Any], targets: list[Executor[Any]]) -> "Self":
+        """Add multiple edges to the workflow.
```
ekzhu: We will need to make it clear what this method does to the message from the source: does it copy the message and send it to all the targets?

python/packages/workflow/agent_framework_workflow/core/workflow.py
```diff
@@ -0,0 +1,211 @@
...
+        activation: Activation = Activation.WhenAll,
+    ) -> "Self":
+        """Add multiple edges from sources to a single target executor.
+
+        Args:
+            sources: A list of source executors for the edges.
+            target: The target executor for the edges.
+            activation: The activation condition for the edges, either WhenAll or WhenAny.
+        """
+        edges = [Edge(source, target) for source in sources]
```
eavanvalkenburg: The above combined with something like this...
```python
if isinstance(activation, str):
    activation = Activation(activation)
```

python/packages/workflow/agent_framework_workflow/core/workflow_context.py
```diff
@@ -0,0 +1,129 @@
...
+        events = self._events.copy()
+        self._events.clear()
+        return events
+
+    async def has_events(self) -> bool:
+        """Check if there are any events in the context."""
+        return bool(self._events)
+
+
+class NoopWorkflowContext(WorkflowContext):
```
eavanvalkenburg: when would someone use this?

python/packages/workflow/agent_framework_workflow/core/executor.py
```diff
@@ -0,0 +1,127 @@
...
+
+class NoopExecutorContext(ExecutorContext):
+    """A no-operation executor context that does nothing."""
+
+    def __init__(self):
+        """Initialize the noop executor context."""
+        super().__init__(executor_id="", shared_state=SharedState(), workflow_context=NoopWorkflowContext())
+
+
+class Executor(Generic[T], ABC):
```
eavanvalkenburg: should we do something where one can turn a Agent into a Executor with one line?
Could be AgentExecutor class that takes a agent or a method on the ChatClientBase class (but that would cause some difficulties in importing)

python/packages/workflow/agent_framework_workflow/core/workflow.py
ekzhu: Can we remove this option, and make the behavior always waiting for all the sources to send message before invoking the target? 

If user wants to have `WhenAny` behavior, they should just add individual edges. 

This makes the semantic of `add_fan_out_edges` unambiguous, and we no longer need user to learn about the concept of activation when using this method. Activation can remain low-level detail.

python/packages/workflow/agent_framework_workflow/core/_shared_state.py
```diff
@@ -0,0 +1,68 @@
...
+        Usage:
+            async with shared_state.hold():
+                await shared_state.set_within_hold("key", value)
+                value = await shared_state.get_within_hold("key")
+        """
+        async with self._shared_state_lock:
+            yield self
+
+    # Unsafe methods that don't acquire locks (for use within hold() context)
+    async def set_within_hold(self, key: str, value: Any) -> None:
```
eavanvalkenburg: would it make sense to make the within_hold methods non-async?

eavanvalkenburg: and maybe use similar names as this: https://docs.python.org/3/library/asyncio-queue.html

python/packages/workflow/agent_framework_workflow/core/executor.py
eavanvalkenburg, line 1: this can just be called _executor.py

python/packages/workflow/agent_framework_workflow/core/workflow.py
eavanvalkenburg, line 1: and _workflow.py

python/packages/workflow/agent_framework_workflow/core/workflow_context.py
eavanvalkenburg, line 1: and _workflow_context.py

python/samples/getting_started/workflow/step_03_simple_group_chat.py
ekzhu: Yes, let's use an actual ChatClientAgent, it is already in the repo now we can finally break each other's code !

python/packages/workflow/agent_framework_workflow/core/workflow.py
alliscode: Totally agree

TaoChenOSU:
> Will unit-tests be included?

Yes, we'll include unit tests

python/samples/getting_started/workflow/step_01_simple_executor.py
TaoChenOSU: We have two contexts currently. One is the `WorkflowContext` and the other is the `ExecutorContext`. The `ExecutorContext` exposes only the APIs that an executor would need. Underneath, it's using the `WorkflowContext` to realize message passing and event streaming.

Now that I am looking at it, the name `ExecutorContext` is still very convoluted. I think I will rename them to `RuntimeContext` and `WorkflowContext`.

TaoChenOSU:
> bunch of small comments and suggestions. And I'm wondering if this should be called `agent_framework.workflows`?

Interesting question! I think we are building the foundations for creating different workflows. LangGraph also does `langgraph.graph`. 

I am open to both `workflow` and `workflows`.

python/packages/workflow/agent_framework_workflow/core/workflow.py
TaoChenOSU: Make sense. Will remove

python/packages/workflow/agent_framework_workflow/core/executor.py
TaoChenOSU: Yes, we will have an `AgentExecutor`. I was going to do it in another PR. But since now we want to make the samples use actual agents, I can include it in this PR too.

python/samples/getting_started/workflow/step_01_simple_executor.py
TaoChenOSU: Removed

python/packages/workflow/agent_framework_workflow/core/workflow_context.py
TaoChenOSU: Removed

python/packages/workflow/agent_framework_workflow/core/workflow.py
TaoChenOSU: This is removed

python/samples/getting_started/workflow/step_04_simple_group_chat.py
```diff
@@ -0,0 +1,146 @@
...
+        """Determine if the group chat should terminate based on the current round."""
+        return self._current_round >= self._max_round
+
+    def _get_next_member(self) -> str:
+        """Get the next member in the round-robin sequence."""
+        return self._members[(self._current_round - 1) % len(self._members)]
+
+
+@output_message_types(list[ChatMessage])
+class FakeAgentExecutor(Executor[AgentSelectionDecision]):
```
alliscode: Is this a placeholder for a built-in agent executor or is this an example of how a user would implement their own. I see that it's emitting an AgentRunEvent so I'm wondering if this is something we would expect users to need to do themselves?

python/packages/workflow/agent_framework_workflow/_executor.py
```diff
@@ -0,0 +1,254 @@
...
+
+    def __init__(
+        self,
+        agent: AIAgent,
+        agent_thread: AgentThread | None = None,
+        streaming: bool = False,
+        id: str | None = None,
+    ):
+        """Initialize the executor with a unique identifier."""
+        super().__init__(id)
```
eavanvalkenburg: could the id just come from agent? (default is a guid)

```suggestion
        super().__init__(id or agent.id)
```

python/samples/getting_started/workflow/step_05_simple_group_chat_with_hil.py
```diff
@@ -0,0 +1,205 @@
...
+    Executor,
+    HumanInTheLoopEvent,
+    WorkflowBuilder,
+    WorkflowCompletedEvent,
+    WorkflowContext,
+    message_handler,
+)
+
+if sys.version_info >= (3, 12):
+    pass  # pragma: no cover
```
eavanvalkenburg: remove

python/packages/workflow/agent_framework_workflow/_executor.py
```diff
@@ -0,0 +1,254 @@
...
+    agent_run_response: AgentRunResponse
+
+
+class AgentExecutor(Executor):
+    """An executor that wraps an agent for handling messages."""
+
+    def __init__(
+        self,
+        agent: AIAgent,
+        agent_thread: AgentThread | None = None,
```
eavanvalkenburg, line 223: I think after agent it should be keyword args
```suggestion
        *,
        agent_thread: AgentThread | None = None,
```

python/samples/getting_started/workflow/step_05_simple_group_chat_with_hil.py
```diff
@@ -0,0 +1,205 @@
...
+    group_chat_manager = CriticGroupChatManagerWithHIL(
+        members=[writer.id, reviewer.id],
+        id="group_chat_manager",
+    )
+
+    # Step 2: Build the workflow with the defined edges.
+    workflow = (
+        WorkflowBuilder()
+        .set_start_executor(group_chat_manager)
+        .add_edge(group_chat_manager, hil_executor)
```
eavanvalkenburg: could we have a convenience method for bidirectional edges?

python/packages/workflow/agent_framework_workflow/_workflow.py
```diff
@@ -0,0 +1,205 @@
...
+        self._edges = edges
+        self._start_executor = start_executor
+        self._executors = {edge.source_id: edge.source for edge in edges} | {
+            edge.target_id: edge.target for edge in edges
+        }
+
+        self._shared_state = SharedState()
+        self._runner = Runner(self._edges, self._shared_state, runner_context)
+
+    async def run_stream(
```
eavanvalkenburg: if there is only one `run` method, can't we just call it `run` instead of run_stream? and for agents this function name was updated to `run_streaming` so would be good to keep in sync.

Other option could be `execute`, but looking at the samples, my preference would be `run`

python/packages/workflow/agent_framework_workflow/_workflow.py
```diff
@@ -0,0 +1,205 @@
...
+        self,
+        message: Any,
+        executor: Executor | str | None = None,
```
eavanvalkenburg: this would imply that a start_executor is optional, but it raises when that is not set in the workflow builder, I would always go for ease of use, so lift that restriction in WFB, and here do a check if either this is passed, or the start_executor is set

python/packages/workflow/agent_framework_workflow/_workflow.py
```diff
@@ -0,0 +1,205 @@
...
+    def set_start_executor(self, executor: Executor | str) -> "Self":
+        """Set the starting executor for the workflow.
+
+        Args:
+            executor: The starting executor, which can be an Executor instance or its ID.
+        """
+        self._start_executor = executor
+        return self
+
+    def set_runner_context(self, runner_context: RunnerContext) -> "Self":
```
eavanvalkenburg: this is not used in any of the samples, so I have no clue what this would look like

python/packages/workflow/agent_framework_workflow/_workflow.py
```diff
@@ -0,0 +1,205 @@
...
+            start_executor: The starting executor for the workflow, which can be an Executor instance or its ID.
+            runner_context: The RunnerContext instance to be used during workflow execution.
+        """
+        self._edges = edges
+        self._start_executor = start_executor
+        self._executors = {edge.source_id: edge.source for edge in edges} | {
+            edge.target_id: edge.target for edge in edges
+        }
+
+        self._shared_state = SharedState()
```
eavanvalkenburg, line 79: should there be a way to pass your own SharedState object in? so that you can do something like a DBBackedSharedState?

python/samples/getting_started/workflow/step_05_simple_group_chat_with_hil.py
```diff
@@ -0,0 +1,205 @@
...
+        .add_edge(hil_executor, group_chat_manager)
```
YusakuNo1: The definition of hil_executor seems to be missing from this revision?

python/samples/getting_started/workflow/step_04_simple_group_chat.py
TaoChenOSU: this is removed.

python/samples/getting_started/workflow/step_05_simple_group_chat_with_hil.py
TaoChenOSU: We are considering those to be higher level APIs. We will eventually add convenient methods for them.

python/packages/workflow/agent_framework_workflow/core/_shared_state.py
TaoChenOSU: Why do we want to make it not async?

python/packages/workflow/agent_framework_workflow/core/workflow.py
TaoChenOSU: This is not bi-directional.

python/packages/workflow/agent_framework_workflow/_shared_state.py
```diff
@@ -0,0 +1,68 @@
...
+    """A class to manage shared state in a workflow."""
+
+    def __init__(self):
+        """Initialize the shared state."""
+        self._state: dict[str, Any] = {}
+        self._shared_state_lock = asyncio.Lock()
+
+    async def set(self, key: str, value: Any) -> None:
+        """Set a value in the shared state."""
+        async with self._shared_state_lock:
```
ekzhu, line 18: just to be clear this is not sufficient for context state management because most of times we want to lock the state for a sequence of operations interlevaing some other code, rather than just one operation at a time.

python/samples/getting_started/workflow/step_05_simple_group_chat_with_hil.py
TaoChenOSU: Fixed

python/packages/workflow/agent_framework_workflow/_workflow.py
TaoChenOSU: Good question! I think with the initial version (local runtime), we will delegate this to a future improvement.

python/packages/workflow/agent_framework_workflow/core/workflow.py
TaoChenOSU: We will keep it explicit in the initial version.

YusakuNo1:
@TaoChenOSU  I triggered a sub-workflow from the main workflow setp, for example, I have `main_flow_executor_1`, `main_flow_executor_2`, `sub_flow_executor_1` and `sub_flow_executor_2`, I'll trigger the sub flow from `main_flow_executor_2` with the following code:

```python
class MainFlowExecutor2(Executor):
    @message_handler
    async def run(self, text: str, ctx: WorkflowContext) -> None:
        sub_workflow = await _start_subflow()
        completion_event = None
        async for event in sub_workflow.run_stream("[sub] initial message"):
            if not isinstance(event, WorkflowCompletedEvent):
                print(f"Event: {event}")
            else:
                completion_event = event

        subflow_result = completion_event.data

        result = text + "\n" + "from MainFlowExecutor2" + "\n" + subflow_result
        await ctx.add_event(WorkflowCompletedEvent(result))
```

I found the execution order is correct:
```
[main] initial message
from MainFlowExecutor1
from MainFlowExecutor2
[sub] initial message
from SubFlowExecutor1
from SubFlowExecutor2
```

But the order for the events are reverted, the user gets events from sub flow, and then from main flow, do I use sub-flow in a wrong way, or it's a bug?
```
Event: ExecutorInvokeEvent(executor_id=sub_flow_executor_1)
Event: ExecutorCompleteEvent(executor_id=sub_flow_executor_1)
Event: ExecutorInvokeEvent(executor_id=sub_flow_executor_2)
Event: ExecutorCompleteEvent(executor_id=sub_flow_executor_2)
Event: ExecutorInvokeEvent(executor_id=main_flow_executor_1)
Event: ExecutorCompleteEvent(executor_id=main_flow_executor_1)
Event: ExecutorInvokeEvent(executor_id=main_flow_executor_2)
Event: ExecutorCompleteEvent(executor_id=main_flow_executor_2)
```

python/packages/workflow/agent_framework_workflow/_shared_state.py
TaoChenOSU: If holding is required for a sequence of operation, we can call `hold` which will return an async context manager and lock the state within the context scope.

python/packages/workflow/agent_framework_workflow/_workflow.py
moonbox3: We will be adding a sync `run` method.

ekzhu:
@YusakuNo1 

> But the order for the events are reverted, the user gets events from sub flow, and then from main flow, do I use sub-flow in a wrong way, or it's a bug?

I believe this is because the way events are being piped out of the workflow runner happens after a super step finishes, while the print statements you put inside the handler executed during the superstep. 

If you change the print to ctx.add_event then it should follow the same sequence.

python/packages/workflow/agent_framework_workflow/_validation.py
```diff
@@ -0,0 +1,487 @@
...
+    """Convenience function to validate a workflow graph.
+
+    Args:
+        edges: list of edges in the workflow
+        start_executor: The starting executor (can be instance or ID)
+
+    Raises:
+        WorkflowValidationError: If any validation fails
+    """
+    validator = WorkflowGraphValidator()
```
TaoChenOSU, line 492: Do we need to use a class for validation?

moonbox3: The instance holds shared state (self._edges) and (self._executors) that's used across the 7+ validation methods. This allows us to avoid the need to pass these collections as parameters to every method. I think it's a cleaner design than threading the same data through multiple static functions.

PR #283, opened 2025-07-30
Python: add agent telemetry

eavanvalkenburg:
### Motivation and Context

Adds agent level telemetry in sync with Otel docs.

Closes: #241 

### Contribution Checklist


python/packages/main/agent_framework/_agents.py
```diff
@@ -525,7 +528,7 @@ async def run(
             response_id=response.response_id,
             created_at=response.created_at,
             usage_details=response.usage_details,
-            raw_representation=response.raw_representation,
+            raw_representation=response,
```
dmytrostruk, line 531: As far as I understand, the `response` here is of type `ChatResponse`. So, if we want to get some details coming from provider SDK directly, we will need to do `agent_response.raw_representation.raw_representation`. Just double checking if it's acceptable for us.

If the purpose of this change is to be able to get some properties from `ChatResponse`, maybe we should consider adding these properties also on `AgentRunResponse` level?

Otherwise, if we want to keep current PR approach, should we do the same change for streaming scenario?

python/packages/main/agent_framework/telemetry.py
```diff
@@ -103,12 +130,15 @@ class GenAIAttributes(str, Enum):
     PROMPT = "gen_ai.prompt"
 
     # Operation names
-    CHAT_COMPLETION_OPERATION = "chat.completions"
-    CHAT_STREAMING_COMPLETION_OPERATION = "chat.streaming_completions"
+    CHAT_COMPLETION_OPERATION = "chat"
     TOOL_EXECUTION_OPERATION = "execute_tool"
+    #    Describes GenAI agent creation and is usually applicable when working with remote agent services.
+    CREATE_AGENT_OPERATION = "create_agent"
+    AGENT_INVOKE_OPERATION = "invoke_agent"
```
dmytrostruk: nit: If we have `CREATE_AGENT_OPERATION`, should we do `INVOKE_AGENT_OPERATION` for the naming to be consistent?

python/packages/main/agent_framework/_agents.py
eavanvalkenburg: yeah, that's the goal, from the Agent perspective the ChatResponse is the raw representation, and this was also changed in dotnet.

eavanvalkenburg: updated the same for ...Update

python/packages/main/agent_framework/telemetry.py
eavanvalkenburg: good one, updated

PR #291, opened 2025-07-31
.NET: Clean / address some message warnings

rogerbarreto:
### Motivation and Context

Clean up and addressing of majority of "addressable" message warnings in the IDE.

### DIFF
--- a/dotnet/src/LegacySupport/DiagnosticAttributes/NullableAttributes.cs
+++ b/dotnet/src/LegacySupport/DiagnosticAttributes/NullableAttributes.cs
@@ -1,6 +1,6 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 
-#pragma warning disable CA1019, RCS1251
+#pragma warning disable CA1019, RCS1251, IDE0300
 
 namespace System.Diagnostics.CodeAnalysis;
 
@@ -81,7 +81,9 @@ internal sealed class NotNullIfNotNullAttribute : Attribute
 /// <summary>Applied to a method that will never return under any circumstance.</summary>
 [AttributeUsage(AttributeTargets.Method, Inherited = false)]
 [ExcludeFromCodeCoverage]
-internal sealed class DoesNotReturnAttribute : Attribute;
+internal sealed class DoesNotReturnAttribute : Attribute
+{
+}
 
 /// <summary>Specifies that the method will not return if the associated Boolean parameter is passed the specified value.</summary>
 [AttributeUsage(AttributeTargets.Parameter, Inherited = false)]
@@ -109,7 +111,7 @@ internal sealed class MemberNotNullAttribute : Attribute
     /// <param name="member">
     /// The field or property member that is promised to be not-null.
     /// </param>
-    public MemberNotNullAttribute(string member) => this.Members = [member];
+    public MemberNotNullAttribute(string member) => this.Members = new[] { member };
 
     /// <summary>Initializes the attribute with the list of field and property members.</summary>
     /// <param name="members">


dotnet/tests/Microsoft.Extensions.AI.Agents.UnitTests/ChatCompletion/ChatClientAgentThreadTests.cs
```diff
@@ -24,8 +24,8 @@ public void VerifyChatClientAgentThreadImplementsIMessagesRetrievableThread()
         var thread = new ChatClientAgentThread();
 
         // Assert
-        Assert.IsAssignableFrom<IMessagesRetrievableThread>(thread);
-        Assert.IsAssignableFrom<AgentThread>(thread);
+        Assert.IsType<IMessagesRetrievableThread>(thread, exactMatch: false);
+        Assert.IsType<AgentThread>(thread, exactMatch: false);
```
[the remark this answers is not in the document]
rogerbarreto: Wrong, when exact match is set to false it tests for Assignable. This was the recommendation message warning.

dotnet/src/Microsoft.Agents.Orchestration/ConcurrentOrchestration.cs
```diff
@@ -41,12 +41,17 @@ public Func<AgentRunResponse[], CancellationToken, Task<AgentRunResponse>> Aggre
                 return f;
             }
 
-            return static async (responses, cancellationToken) =>
-                new AgentRunResponse([.. responses.Where(r => r.Messages.Count > 0).Select(r =>
-                {
-                    var messages = r.Messages;
-                    return messages.Count > 0 ? messages[messages.Count - 1] : new();
-                })]);
+            return (AgentRunResponse[] responses, CancellationToken cancellationToken)
```
stephentoub: I know you're just trying to address the suggestions about async methods without awaits. I think we should just globally suppress that (and I want to pursue getting it changed in the compiler). There's nothing wrong with using it as was done here; it keeps the code simpler, and ensures that any exceptions are propagated as part of the task.

dotnet/tests/Microsoft.Extensions.AI.Agents.UnitTests/ChatCompletion/ChatClientAgentThreadTests.cs
stephentoub, line 28: Why did these change?

rogerbarreto: https://xunit.net/xunit.analyzers/rules/xUnit2032

## Reason for rule
There is confusing about which arguments is "from" and when is "to" when using Assert.IsAssignableFrom (and Assert.IsNotAssignableFrom). We have added an overload of Assert.IsType (and Assert.IsNotType) that allows users to pass a flag whether they want strict ("exact") type matching behavior (like Assert.IsType without the flag) or compatible ("inexact") type matching behavior (like Assert.IsAssignableFrom).

## How to fix violations
To fix a violation of this rule, convert the assertion to Assert.IsType (or Assert.IsNotType) using exactMatch: false.

dotnet/src/Microsoft.Agents.Orchestration/ConcurrentOrchestration.cs
rogerbarreto: If we should suppress globally I would assume that globally also there's no problem using an async without an await inside, right? 

We all had many of those warnings in the past would be interesting to understand how problematic a global suppression would be if we stop at all caring on returning synchronously on `async` methods.

dotnet/tests/Microsoft.Extensions.AI.Agents.Abstractions.UnitTests/Microsoft.Extensions.AI.Agents.Abstractions.UnitTests.csproj
```diff
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFrameworks>$(ProjectsTargetFrameworks)</TargetFrameworks>
+    <NoWarn>$(NoWarn);Moq1410;</NoWarn>
```
westey-m: Consider just putting this into the test folder's Directory.Build.Props.

dotnet/tests/Microsoft.Extensions.AI.Agents.UnitTests/OpenTelemetryAgentTests.cs
```diff
@@ -12,6 +12,8 @@
 using OpenTelemetry.Metrics;
 using OpenTelemetry.Trace;
 
+#pragma warning disable Moq1410 // Moq: Set MockBehavior to Strict
```
westey-m: If we move this suppression into the directory.build.props, we can remove this one too.

dotnet/src/LegacySupport/DiagnosticAttributes/NullableAttributes.cs
```diff
@@ -8,30 +8,22 @@ namespace System.Diagnostics.CodeAnalysis;
 /// <summary>Specifies that null is allowed as an input even if the corresponding type disallows it.</summary>
 [AttributeUsage(AttributeTargets.Field | AttributeTargets.Parameter | AttributeTargets.Property, Inherited = false)]
 [ExcludeFromCodeCoverage]
-internal sealed class AllowNullAttribute : Attribute
-{
-}
+internal sealed class AllowNullAttribute : Attribute;
```
westey-m: Also consider just suppressing any issues for the legacysupport classes.  They are all copied from the Microsoft.Extentions lib, which is really the source of truth for these.  So any changes here, make taking updates from there harder.

PR #296, opened 2025-08-01
.NET: Add agent hosting package and update sample

ReubenBond:
## Motivation and Context

We need a way to host agents so that we can expose them in a uniform way (eg via various HTTP-based protocols or in-process to local callers). This PR implements support for this. It:

- Introduces a new `Microsoft.Extensions.AI.Agents.Hosting` package that provides infrastructure for hosting AI agents. This includes:
  - Registering agents by name
  - Resolving a _proxy_ for the agent which uses the `IActorClient` to communicate with the actual agent instance. The reason for this indirection is to support a distributed implementation of `IActorClient` which communicates via a gateway. That lets us host agents in other languages, scale out, etc.
- Replaces the `HelloHttpApi` sample with a new `AgentWebChat` sample and updates it to use these new capabilities.
- Implements agent proxy pattern for remote agent invocation with thread-based conversation management.

## Changes
- New Agent Hosting Package
  - Added `Microsoft.Extensions.AI.Agents.Hosting` project with core agent hosting infrastructure
  - Implemented `AgentActor` for hosting agents as actors with proper lifecycle management
  - Created `AgentProxy` and `AgentProxyFactory` for creating proxies to remote agents
  - Added `AgentProxyThread` for managing threaded conversations with agents
  - Added extension methods for `HostApplicationBuilder` to simplify agent registration
- Updated Sample Application (Renamed `HelloHttpApi` sample to `AgentWebChat` and updated to use the new bits)

Registering an Agent looks like this:

```csharp
builder.AddAIAgent(
    "pirate",
    instructions: "You are a pirate. Speak like a pirate",
    description: "An agent that speaks like a pirate.",
    chatClientServiceKey: "chat-model");
```

A more advanced example (an orchestration, using dependency injection directly) looks like this

```csharp
builder.AddAIAgent("knights-and-knaves", (sp, key) =>
{
    var chatClient = sp.GetRequiredKeyedService<IChatClient>("chat-model");

    ChatClientAgent knight = new(
        chatClient,
        """
        You are a knight. This means that you must always tell the truth. Your name is Alice.
        Bob is standing next to you. Bob is a knave, which means he always lies.
        When replying, always start with your name (Alice). Eg, "Alice: I am a knight."
        """, "Alice");

    ChatClientAgent knave = new(
        chatClient,
        """
        You are a knave. This means that you must always lie. Your name is Bob.
        Alice is standing next to you. Alice is a knight, which means she always tells the truth.
        When replying, always include your name (Bob). Eg, "Bob: I am a knight."
        """, "Bob");

    ChatClientAgent narrator = new(
        chatClient,
        """
        You are are the narrator of a puzzle involving knights (who always tell the truth) and knaves (who always lie).
        The user is going to ask questions and guess whether Alice or Bob is the knight or knave.
        Alice is standing to one side of you. Alice is a knight, which means she always tells the truth.
        Bob is standing to the other side of you. Bob is a knave, which means he always lies.
        When replying, always include your name (Narrator).
        Once the user has deduced what type (knight or knave) both Alice and Bob are, tell them whether they are right or wrong.
        If the user asks a general question about their surrounding, make something up which is consistent with the scenario.
        """, "Narrator");

    return new ConcurrentOrchestration([knight, knave], name: key);
});
```

We register agents by name as keyed services in the Dependency Injection container and add them to a hash set so that we can implement basic agent discovery. The sample exposes a discovery endpoint and uses that endpoint to populate the dropdown of agent types.

### Contribution Checklist

### DIFF
--- a/dotnet/tests/Microsoft.Extensions.AI.Agents.Runtime.Abstractions.UnitTests/JsonSerializationTests.cs
+++ b/dotnet/tests/Microsoft.Extensions.AI.Agents.Runtime.Abstractions.UnitTests/JsonSerializationTests.cs
@@ -20,7 +20,7 @@ public class JsonSerializationTests
             WriteIndented = false, // Use compact JSON for easier testing
             DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull,
             Converters = { new JsonStringEnumConverter() },
-            TypeInfoResolver = ActorJsonContext.Default
+            TypeInfoResolver = AgentRuntimeAbstractionsJsonUtilities.JsonContext.Default
         };
     }
 


dotnet/src/Shared/Samples/OrchestrationSample.cs
```diff
@@ -27,7 +27,7 @@ public abstract class OrchestrationSample : BaseSample
     protected ChatClientAgent CreateAgent(string instructions, string? description = null, string? name = null, params AIFunction[] functions)
     {
         // Get the chat client to use for the agent.
-        using IChatClient chatClient = CreateChatClient();
+        IChatClient chatClient = CreateChatClient();
```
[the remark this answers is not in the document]
ReubenBond: That's not true. The using statement here is incorrect. I undid this change anyway because it's not related to this PR.

dotnet/src/Microsoft.Extensions.AI.Agents.Hosting/LocalAgentRegistry.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Collections.Generic;
+
+namespace Microsoft.Extensions.AI.Agents.Hosting;
+
+internal sealed class LocalAgentRegistry
+{
+    public HashSet<string> AgentNames { get; } = [];
```
DeagleGross, line 9: can there be multiple instances of same agent? should this collection be expanded to Dictionary<string, string> or smth like that then?

ReubenBond: Agents in this case are singletons. You can have multiple distinct threads with one agent. You can think of an Agent as a stateless function and an AgentThread as the input to that function.

dotnet/src/Microsoft.Extensions.AI.Agents.Hosting/LocalAgentCatalog.cs
```diff
@@ -0,0 +1,37 @@
...
+    public readonly HashSet<string> _registeredAgents;
+    private readonly IServiceProvider _serviceProvider;
+
+    public LocalAgentCatalog(LocalAgentRegistry agentHostBuilder, IServiceProvider serviceProvider)
+    {
+        this._registeredAgents = [.. agentHostBuilder.AgentNames];
+        this._serviceProvider = serviceProvider;
+    }
+
+    public override async IAsyncEnumerable<AIAgent> GetAgentsAsync([EnumeratorCancellation] CancellationToken cancellationToken = default)
```
stephentoub, line 24: I'm curious about this approach rather than just enumerating all AIAgents directly added to DI, e.g. `GetServices<AIAgent>()`?

dotnet/samples/AgentWebChat/AgentWebChat.AgentHost/InvocationHttpProcessor.cs
```diff
@@ -0,0 +1,112 @@
...
+                var thread = agent.GetThread(sessionId);
+                var updateTypeInfo = AgentAbstractionsJsonUtilities.DefaultOptions.GetTypeInfo(typeof(AgentRunResponseUpdate));
+                await foreach (var update in agent.RunStreamingAsync(messages, thread, cancellationToken: cancellationToken).ConfigureAwait(false))
+                {
+                    var eventData = JsonSerializer.Serialize(update, updateTypeInfo);
+                    var eventText = $"data: {eventData}\n\n";
+
+                    await response.WriteAsync(eventText, cancellationToken);
+                    await response.Body.FlushAsync(cancellationToken);
+                }
```
stephentoub: Could this use SseFormatter from System.Net.ServerSentEvents 10.x? If not, I'm wondering if there's something we need to add before it goes out the door in .NET 10

dotnet/samples/AgentWebChat/AgentWebChat.AgentHost/Program.cs
```diff
@@ -0,0 +1,79 @@
...
+
+// Add service defaults & Aspire client integrations.
+builder.AddServiceDefaults();
+builder.Services.AddOpenApi();
+
+// Add services to the container.
+builder.Services.AddProblemDetails();
+
+// Configure the chat model and our agent.
+builder.AddKeyedChatClient("chat-model");
```
stephentoub, line 36: Where is this extension method coming from? Is this Aspire?

dotnet/samples/AgentWebChat/AgentWebChat.AgentHost/Program.cs
```diff
@@ -0,0 +1,79 @@
...
+
+    ChatClientAgent knave = new(
+        chatClient,
+        """
+        You are a knave. This means that you must always lie. Your name is Bob.
+        Alice is standing next to you. Alice is a knight, which means she always tells the truth.
+        When replying, always include your name (Bob). Eg, "Bob: I am a knight."
+        """, "Bob");
+
+    ChatClientAgent narrator = new(
```
stephentoub, line 64: narrator isn't being used. How was it intended to be integrated? Maybe you'd intended for something like:
```C#
return new SequentialOrchestration(
    new ConcurrentOrchestration([knight, knave], name: key),
    narrator);
```
?

dotnet/samples/AgentWebChat/AgentWebChat.Web/AgentClient.cs
```diff
@@ -0,0 +1,91 @@
...
+            {
+                yield break;
+            }
+
+            if (sseLine.StartsWith("data: ", StringComparison.Ordinal) && TryParseUpdate(sseLine, logger, out var responseUpdate))
+            {
+                yield return responseUpdate;
+            }
+        }
+    }
```
stephentoub: This should be able to use SseParser from System.Net.ServerSentEvents

dotnet/src/Microsoft.Extensions.AI.Agents.Hosting/AgentHostingJsonUtilities.cs
```diff
@@ -0,0 +1,64 @@
...
+
+    // Keep in sync with CreateDefaultOptions above.
+    [JsonSourceGenerationOptions(JsonSerializerDefaults.Web,
+        DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull,
+        NumberHandling = JsonNumberHandling.AllowReadingFromString)]
+    [JsonSerializable(typeof(AgentRunRequest))]
+    [JsonSerializable(typeof(AgentProxyThread))]
+    [JsonSerializable(typeof(ChatClientAgentThread))]
+    [JsonSerializable(typeof(IReadOnlyCollection<ChatMessage>))]
+    [JsonSerializable(typeof(List<ChatMessage>))]
```
stephentoub: Are these necessary? I'd have hoped we would get those from AIJsonUtilities, which is chained into AgentAbstractionsJsonUtilities. Similarly for others here, like AgentRunResponse{Update} and ChatClientAgentThread, we'd hopefully be picking those up from the other utility types of the lower libraries. Ideally this JsonContext wouldn't need anything except for new types this library contributes.

dotnet/src/Microsoft.Extensions.AI.Agents.Hosting/AgentProxy.cs
```diff
@@ -0,0 +1,146 @@
...
+using System.Threading.Tasks;
+using Microsoft.Extensions.AI.Agents.Runtime;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Extensions.AI.Agents.Hosting;
+
+/// <summary>
+/// Represents a proxy for an AI agent that communicates with the agent runtime via an actor client.
+/// </summary>
+public sealed class AgentProxy : AIAgent
```
stephentoub, line 19: Does this need to be public? I see you've also added an AgentProxyFactory, so I'm curious for the need for both the factory and for this to be public, rather than either not having the factory or having this be internal and the factory returning instances of this as the base AIAgent.

dotnet/src/Microsoft.Extensions.AI.Agents.Hosting/AgentProxyThread.cs
```diff
@@ -0,0 +1,81 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Extensions.AI.Agents.Hosting;
+
+/// <summary>
+/// Represents an agent thread for a <see cref="AgentProxy"/>.
+/// </summary>
+public sealed partial class AgentProxyThread : AgentThread
```
stephentoub: Similarly here... does this need to be public?

dotnet/src/Microsoft.Extensions.AI.Agents.Hosting/AgentProxyThread.cs
```diff
@@ -0,0 +1,81 @@
...
+{
+#if NET7_0_OR_GREATER
+    [System.Diagnostics.CodeAnalysis.StringSyntax("Regex")]
+#endif
```
stephentoub, line 16: We could also bring this in as a polyfill downlevel, but we can do that subsequently.

dotnet/src/Microsoft.Extensions.AI.Agents.Hosting/AgentProxyThread.cs
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="AgentProxyThread"/> class with the specified identifier.
+    /// </summary>
+    public AgentProxyThread() : this(CreateId())
+    {
+    }
+
+    internal static string CreateId() => Guid.NewGuid().ToString("N");
```
stephentoub, line 58: This will only ever create hex characters, but the regex is allowing more than that... we expect IDs to come from elsewhere?

dotnet/src/Microsoft.Extensions.AI.Agents.Hosting/AgentProxyThread.cs
```diff
@@ -0,0 +1,81 @@
...
+        {
+            throw new ArgumentException($"Thread ID '{id}' is not valid. Thread IDs must contain only alphanumeric characters, hyphens, underscores, dots, and tildes.", nameof(id));
+        }
+    }
+
+    /// <inheritdoc/>
+    protected override Task OnNewMessagesAsync(IReadOnlyCollection<ChatMessage> newMessages, CancellationToken cancellationToken = default)
+    {
+        return base.OnNewMessagesAsync(newMessages, cancellationToken);
+    }
```
stephentoub: Is this necessary if it's just delegating to the base?

dotnet/src/Microsoft.Extensions.AI.Agents.Hosting/HostApplicationBuilderAgentExtensions.cs
```diff
@@ -0,0 +1,153 @@
...
+        return builder.AddAgentActor(name);
+    }
+
+    private static IHostApplicationBuilder AddAgentActor(this IHostApplicationBuilder builder, string name)
+    {
+        Throw.IfNull(builder);
+
+        // Register the agent by name for discovery.
+        var agentHostBuilder = GetAgentRegistry(builder);
+        agentHostBuilder.AgentNames.Add(name);
```
stephentoub, line 128: I see, with regards to my previous comment about why you have this intermediate builder rather than enumerating via GetServices, is it because there's no exposed enumeration mechanism that lets you not only enumerate the services but also their associated keyed names?

dotnet/src/Microsoft.Extensions.AI.Agents.Hosting/AgentProxy.cs
```diff
@@ -0,0 +1,146 @@
...
+{
+    private readonly IActorClient _client;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="AgentProxy"/> class with the specified agent name and actor client.
+    /// </summary>
+    /// <param name="name">The name of the agent.</param>
+    /// <param name="client">The actor client used to communicate with the agent.</param>
+    public AgentProxy(string name, IActorClient client)
```
stephentoub, line 28: I was confused yesterday when seeing you briefly demo things how it all plugs together, but after reading through the PR, I get it now.

The hosting layer wraps an arbitrary AIAgent with an runtime-aware actor. That actor monitors for new messages, including run requests, and forwards them to AIAgent.RunAsync.

This proxy then acts as the other side of the channel. It is an AIAgent and so can be used as one, but any requests to it are in turn sent as messages via the runtime to that actor wrapping the real AIAgent.

dotnet/samples/AgentWebChat/AgentWebChat.AgentHost/InvocationHttpProcessor.cs
```diff
@@ -0,0 +1,112 @@
...
+using System.Text.Json;
+using Microsoft.AspNetCore.Http.Features;
+using Microsoft.Extensions.AI;
+using Microsoft.Extensions.AI.Agents;
+using Microsoft.Extensions.AI.Agents.Hosting;
+using Microsoft.Extensions.AI.Agents.Runtime;
+
+namespace AgentWebChat.AgentHost;
+
+internal static class InvocationHttpProcessor
```
stephentoub: How much of the boilerplate in this file do you imagine we'll eventually want to package up as a reusable component?

dotnet/src/Microsoft.Extensions.AI.Agents.Hosting/LocalAgentCatalog.cs
ReubenBond: The rationale is that we want it to be async so we can query a remote server if needed... but on the other hand, this is a very rough sketch of what agent discovery could look like and it's probably too basic to be useful.

Worth noting that agents are registered using Keyed DI and that cannot be enumerated in a way that gives us all registered keys.

dotnet/samples/AgentWebChat/AgentWebChat.AgentHost/InvocationHttpProcessor.cs
ReubenBond: Yes, probably, if we have stable packages or are willing to use preview ones.

dotnet/samples/AgentWebChat/AgentWebChat.AgentHost/Program.cs
ReubenBond: Sort-of. It's defined in the Utilities folder. I'm asking for something like this to be added to Aspire (cc @luisquintanilla). I went with this because it was convenient for the sample.

dotnet/samples/AgentWebChat/AgentWebChat.AgentHost/Program.cs
ReubenBond: That might explain why it doesn't work as intended 😅

dotnet/samples/AgentWebChat/AgentWebChat.Web/AgentClient.cs
ReubenBond: I've pushed an update which does away with this and does use SseParser - thanks

dotnet/src/Microsoft.Extensions.AI.Agents.Hosting/AgentHostingJsonUtilities.cs
ReubenBond: Will fix

dotnet/src/Microsoft.Extensions.AI.Agents.Hosting/AgentProxy.cs
ReubenBond: Currently it does because it has a `GetThread(string id)` method. We could remove the factory altogether. It would make the surface simpler.

dotnet/src/Microsoft.Extensions.AI.Agents.Hosting/AgentProxyThread.cs
ReubenBond: I'll hide it. We can set the `Id` on the `AgentThread` base type instead of going through a constructor.

dotnet/src/Microsoft.Extensions.AI.Agents.Hosting/AgentProxyThread.cs
ReubenBond: I restricted the regex to characters which are suitable for a URL slug, but we don't need them all for our own generated identifiers.

dotnet/src/Microsoft.Extensions.AI.Agents.Hosting/AgentProxyThread.cs
ReubenBond: I'll delete it. I only added this for my own debugging purposes and forgot to remove it.

dotnet/src/Microsoft.Extensions.AI.Agents.Hosting/HostApplicationBuilderAgentExtensions.cs
ReubenBond: That's right. Services in Keyed DI cannot be enumerated without providing a key since it would break the `AnyKey` case.

dotnet/samples/AgentWebChat/AgentWebChat.AgentHost/InvocationHttpProcessor.cs
ReubenBond: In the latest revision, I've removed this and replaced it with a more general-purpose `HttpActorClient` and `HttpActorProcessor` which I believe we could make into a reusable component.

dotnet/src/Microsoft.Extensions.AI.Agents.Hosting/HostApplicationBuilderAgentExtensions.cs
stephentoub: They can, you just don't get the keys.

dotnet/samples/AgentWebChat/AgentWebChat.AgentHost/Program.cs
luisquintanilla: cc: @maddymontaquila @DamianEdwards

DamianEdwards: This is the method name already being used in the CommunityToolkit packages AFAIK, e.g. https://github.com/search?q=repo%3ACommunityToolkit%2FAspire%20AddKeyedChatClient&type=code

Issue #281, opened 2025-07-30
Python: Feature completeness for OpenAI Response Agent


markwallace-microsoft:
- [x] Should work with OpenAI and Azure OpenAI
- [x] Add tools must be supported and covered in integration tests
- [x] All options must be supported and covered in integration tests
- [x] Add getting started samples to cover all features
- [x] Create documentation on learn site to cover all features

dmytrostruk:
Existing examples here:
https://github.com/microsoft/agent-framework/tree/main/python/samples/getting_started/agents/openai_responses_client

PR #311, opened 2025-08-04
Python: Azure Responses client

peterychang:
https://github.com/microsoft/agent-framework/issues/281

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Get Started with Microsoft Agent Framework
+# Microsoft Agent Framework
 
 Highlights
 - Flexible Agent Framework: build, orchestrate, and deploy AI agents and multi-agent systems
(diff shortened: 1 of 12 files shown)


python/packages/azure/agent_framework_azure/_responses_client.py
```diff
@@ -0,0 +1,129 @@
...
+                prompts could use `developer` or `system`. (Optional)
+        """
+        try:
+            # Filter out any None values from the arguments
+            azure_openai_settings = AzureOpenAISettings(
+                api_key=SecretStr(api_key) if api_key else None,
+                base_url=AnyUrl(base_url) if base_url else None,
+                endpoint=AnyUrl(endpoint) if endpoint else None,
+                responses_deployment_name=deployment_name,
+                api_version=api_version or "preview",
```
eavanvalkenburg: this means that whatever is in the .env will be overwritten, either by `preview` or by `api_version`, I don't think that's ideal, because eventually people will be able to call this with something other then preview and we don't want to have to do a code change to acomodate. I think we just put in a log message if the resulting api_version is not 'preview' but without actually breaking, so that later on people can ignore and then we can remove the log message.

python/packages/azure/agent_framework_azure/_responses_client.py
```diff
@@ -0,0 +1,131 @@
...
+
+from ._shared import (
+    AzureOpenAIConfigBase,
+    AzureOpenAISettings,
+)
+
+
+@use_telemetry
+@use_tool_calling
+class AzureResponsesClient(AzureOpenAIConfigBase, OpenAIResponsesClientBase):
```
dmytrostruk, line 24: Please add an example of this client under `samples/getting_started/chat_client` folder.

PR #313, opened 2025-08-04
Python: Added author name logic to ChatClientAgent

dmytrostruk:
### Motivation and Context

This PR contains logic which initializes author name in chat message response as agent name in case it's not returned from chat client. 

### Contribution Checklist

### DIFF
--- a/python/packages/main/agent_framework/_agents.py
+++ b/python/packages/main/agent_framework/_agents.py
@@ -491,6 +491,7 @@ class ChatClientAgent(AgentBase):
         """
         input_messages = self._normalize_messages(messages)
         thread, thread_messages = await self._prepare_thread_and_messages(thread=thread, input_messages=input_messages)
+        agent_name = self._get_agent_name()
 
         response = await self.chat_client.get_response(
             messages=thread_messages,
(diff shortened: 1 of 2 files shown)


python/packages/main/agent_framework/_agents.py
```diff
@@ -595,6 +601,7 @@ async def run_streaming(
         """
         input_messages = self._normalize_messages(messages)
         thread, thread_messages = await self._prepare_thread_and_messages(thread=thread, input_messages=input_messages)
+        agent_name = self._get_agent_name()
```
eavanvalkenburg, line 604: why not use `display_name`?

dmytrostruk: Just to be consistent with .NET:
https://github.com/microsoft/agent-framework/blob/32f1132a5d1ac0af56ac5fb0b353db2fb24f8295/dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatClientAgent.cs#L388

Also, I'm not sure if `author_name` would be meaningful enough as GUID, but I'm happy to change the logic as needed.

eavanvalkenburg: Ok great, then let's leave as is

PR #347, opened 2025-08-05
Python: MCP support for locally defined MCP servers

eavanvalkenburg:
### Motivation and Context

Draft of MCP support.

Still needs test and design choice.

### Contribution Checklist


python/packages/main/agent_framework/_agents.py
```diff
@@ -381,6 +385,13 @@ def __init__(
             kwargs: any additional keyword arguments.
                 Unused, can be used by subclasses of this Agent.
         """
+        mcp_servers: list[MCPBase] | None = None
+        if isinstance(tools, MCPBase):
+            mcp_servers = [tools]
+            tools = None
+        elif isinstance(tools, list) and all(isinstance(tool, MCPBase) for tool in tools):  # type: ignore[reportUnknownVariableType]
+            mcp_servers = tools  # type: ignore[reportArgumentType]
+            tools = None
```
dmytrostruk: I like the approach of exposing MCP as part of `tools` array, since it is a tool for the agent. I'm not sure about exposing it as `mcp_servers` property in `ChatClientAgent`, because `ChatClientAgent` is an abstraction. If chat client doesn't support MCP as a tool, the property on `ChatClientAgent` level is not applicable. Ideally, `ChatClientAgent` should be unaware about specific tool types. It should be able to pass the tool to the concrete chat client implementation, and that implementation should handle the tool if it's supported.

python/packages/main/agent_framework/_agents.py
```diff
@@ -338,6 +340,8 @@ def __init__(
         tool_choice: ChatToolMode | Literal["auto", "required", "none"] | dict[str, Any] | None = "auto",
         tools: AITool
         | list[AITool]
+        | MCPBase
+        | list[MCPBase]
```
dmytrostruk: Can we make `MCPBase` inherit from `AITool`?

python/packages/main/agent_framework/_agents.py
```diff
@@ -381,6 +385,13 @@ def __init__(
             kwargs: any additional keyword arguments.
                 Unused, can be used by subclasses of this Agent.
         """
+        mcp_servers: list[MCPBase] | None = None
+        if isinstance(tools, MCPBase):
+            mcp_servers = [tools]
+            tools = None
+        elif isinstance(tools, list) and all(isinstance(tool, MCPBase) for tool in tools):  # type: ignore[reportUnknownVariableType]
```
dmytrostruk: This checks whether all tools in array are MCP tools. What will happen if I pass one MCP tool and one code interpreter tool for example? It feels like this scenario should be supported.

python/packages/main/agent_framework/_mcp.py
```diff
@@ -0,0 +1,752 @@
...
+        )
+    return DataContent(
+        uri=mcp_type.resource.blob,
+        raw_representation=mcp_type,
+        additional_properties=mcp_type.annotations.model_dump() if mcp_type.annotations else {},
+    )
+
+
+def _ai_content_to_mcp_content_types(
+    content: ChatMessage | TextContent | DataContent | UriContent,
```
dmytrostruk: I think `ChatMessage` should be handled separately, since it's not `AIContent`, it's a container for `AIContent`. There is also a separate definition of `AIContents`, which is `TextContent | DataContent | TextReasoningContent | UriContent...` and more, so maybe we can re-use it here.

python/packages/main/agent_framework/_mcp.py
```diff
@@ -0,0 +1,752 @@
...
+) -> Sequence[types.TextContent | types.ImageContent | types.AudioContent | types.EmbeddedResource]:
+    """Convert a kernel content type to a MCP type."""
```
dmytrostruk:
```suggestion
    """Convert an Agent Framework content type to an MCP content type."""
```

python/packages/main/agent_framework/_mcp.py
```diff
@@ -0,0 +1,752 @@
...
+        if content.media_type and content.media_type.startswith("audio/"):
+            return [types.AudioContent(type="audio", data=content.uri, mimeType=content.media_type)]
+        if content.media_type and content.media_type.startswith("application/"):
+            return [
+                types.EmbeddedResource(
+                    type="resource",
+                    resource=types.BlobResourceContents(
+                        blob=content.uri,
+                        mimeType=content.media_type,
+                        uri="sk://binary",  # type: ignore[reportArgumentType]
```
dmytrostruk: What does it mean?

python/packages/main/agent_framework/_mcp.py
```diff
@@ -0,0 +1,752 @@
...
+                    ),
+                )
+            ]
+    if isinstance(content, UriContent):
+        return [
+            types.EmbeddedResource(
```
dmytrostruk: In `_mcp_content_types_to_ai_content` method we map `types.ResourceLink` to `UriContent`, while in this method we map `UriContent` to `types.EmbeddedResource`. Should we use the same type for mapping in both ways?

python/packages/main/agent_framework/_mcp.py
```diff
@@ -0,0 +1,752 @@
...
+
+def _normalize_mcp_name(name: str) -> str:
+    """Normalize MCP tool/prompt names to allowed identifier pattern (A-Za-z0-9_.-)."""
+    return re.sub(r"[^A-Za-z0-9_.-]", "-", name)
+
+
+# region: MCP Plugin
+
+
+class MCPBase:
```
dmytrostruk: nit:
```suggestion
# region: MCP Tool


class MCPBase:
```

As already mentioned above, it would be great if `MCPBase` can inherit from `AITool`, so we can use it as any other tool, if possible.

python/packages/main/agent_framework/_mcp.py
```diff
@@ -0,0 +1,752 @@
...
+        self,
+        name: str,
+        description: str | None = None,
+        load_tools: bool = True,
+        load_prompts: bool = True,
+        session: ClientSession | None = None,
+        request_timeout: int | None = None,
+        chat_client: ChatClient | None = None,
+    ) -> None:
+        """Initialize the MCP Plugin Base."""
```
dmytrostruk, line 238:
```suggestion
        """Initialize the MCP Tool Base."""
```

python/packages/main/agent_framework/_mcp.py
```diff
@@ -0,0 +1,752 @@
...
+        self.functions: dict[str, AIFunction[Any, Any]] = {}
+
+    async def connect(self) -> None:
+        """Connect to the MCP server."""
+        if not self.session:
+            try:
+                transport = await self._exit_stack.enter_async_context(self.get_mcp_client())
+            except Exception as ex:
+                await self._exit_stack.aclose()
+                raise FunctionException("Failed to connect to the MCP server. Please check your configuration.") from ex
```
dmytrostruk: I'm wondering what is `FunctionException`? If MCP will be `AITool`, maybe we can introduce `AIToolException` or something like that.

python/packages/main/agent_framework/_mcp.py
```diff
@@ -0,0 +1,752 @@
...
+        )
+        if not mcp_content:
+            return types.ErrorData(
+                code=types.INTERNAL_ERROR,
+                message="Failed to get right content types from the response.",
+            )
+        return types.CreateMessageResult(
+            role="assistant",
+            content=mcp_content,
+            model=response.ai_model_id or "unknown",
```
dmytrostruk, line 344: Usage of `unknown` value looks like either `model` parameter should accept `None` or `ai_model_id` should not be `None`.

python/packages/main/agent_framework/_mcp.py
```diff
@@ -0,0 +1,752 @@
...
+        )
+
+    async def logging_callback(self, params: types.LoggingMessageNotificationParams) -> None:
+        """Callback function for logging.
+
+        This function is called when the MCP Server sends a log message.
+        By default it will log the message to the logger with the level set in the params.
+
+        Please subclass the MCP*Plugin and override this function if you want to adapt the behavior.
```
dmytrostruk, line 353: The documentation needs to be updated in all methods to use the latest terminology (`Plugin` -> `Tool`) etc.

python/packages/main/agent_framework/_mcp.py
```diff
@@ -0,0 +1,752 @@
...
+        """Load tools from the MCP server."""
+        if not self.session:
+            raise FunctionExecutionException(
+                "MCP server not connected, please call connect() before using this method."
+            )
+        try:
+            tool_list = await self.session.list_tools()
+        except Exception:
+            tool_list = None
+            # Create methods with the kernel_function decorator for each tool
```
dmytrostruk:
```suggestion
            # Create methods with the ai_function decorator for each tool
```

python/packages/main/agent_framework/_mcp.py
```diff
@@ -0,0 +1,752 @@
...
+        if self.sse_read_timeout:
+            args["sse_read_timeout"] = self.sse_read_timeout
+        if self.terminate_on_close is not None:
+            args["terminate_on_close"] = self.terminate_on_close
+        if self._client_kwargs:
+            args.update(self._client_kwargs)
+        return streamablehttp_client(**args)
+
+
+class MCPWebsocketTools(MCPBase):
```
dmytrostruk: All MCP child classes have `Tools` plural name, is there a specific reason for that? Can we name it as `MCPWebsocketTool`? Same for other classes.

python/packages/main/agent_framework/openai/_chat_client.py
```diff
@@ -236,7 +236,28 @@ def _openai_chat_message_parser(self, message: ChatMessage) -> list[dict[str, An
...
+                        else:
+                            if isinstance(content.result, BaseModel):
+                                results.append(
+                                    content.result.model_dump_json(
+                                        exclude_none=True, exclude={"raw_representation", "additional_properties"}
+                                    )
+                                )
+                            else:
+                                results.append(json.dumps(content.result))
+                    args["content"] = results if len(results) > 1 else results[0] if results else None
```
dmytrostruk: nit: Could we place this code in a separate method, so it's clearer what its responsibility?

python/packages/main/agent_framework/_agents.py
eavanvalkenburg: ChatClientAgent is not a abstraction, but I made a change and made these into private attributes. for me, a LocalMcpTool is nothing but a list of AIFunctions, so if a ChatClient supports functions, it supports this. We will also get `HostedMcpServerTool` and for that it will be dependent on the ChatClient to deal with that: https://github.com/dotnet/extensions/pull/6664/files#diff-1a7ee3fb848ff2598a6a035727bdde2b60eba7017d2f7a0f274bb8bee1fb73bb

python/packages/main/agent_framework/_mcp.py
```diff
@@ -0,0 +1,760 @@
...
+                    type="resource",
+                    resource=types.BlobResourceContents(
+                        blob=content.uri,
+                        mimeType=content.media_type,
+                        # uri's are not limited in MCP but they have to be set.
+                        # the uri of data content, contains the data uri, which
+                        # is not the uri meant here, UriContent would match this.
+                        uri=content.additional_properties.get("uri", "sk://binary")
+                        if content.additional_properties
+                        else "sk://binary",  # type: ignore[reportArgumentType]
```
dmytrostruk: Should it be `af://binary` in this case?

eavanvalkenburg:
replaced by two separate PR's

PR #363, opened 2025-08-06
Python: Added examples for Azure OpenAI Responses agent

dmytrostruk:
### Motivation and Context

Added following examples with Azure OpenAI Responses agent:
1. Example with streaming/non-streaming agent.
2. Example with agent and function tools.
3. Example with agent and thread management.
4. Example with agent and code interpreter.

### Contribution Checklist

### DIFF
--- a/dotnet/Directory.Packages.props
+++ b/dotnet/Directory.Packages.props
@@ -8,9 +8,9 @@
   <ItemGroup>
     <!-- Azure.* -->
     <PackageVersion Include="Aspire.Azure.AI.OpenAI" Version="9.3.1-preview.1.25305.6" />
-    <PackageVersion Include="Aspire.Hosting.AppHost" Version="9.3.1" />
+    <PackageVersion Include="Aspire.Hosting.AppHost" Version="9.4.0" />
     <PackageVersion Include="Aspire.Hosting.Azure.CognitiveServices" Version="9.3.1" />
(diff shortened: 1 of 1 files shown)


python/samples/getting_started/agents/azure_responses_client/azure_responses_client_basic.py
```diff
@@ -0,0 +1,63 @@
...
+
+async def non_streaming_example() -> None:
+    """Example of non-streaming response (get the complete result at once)."""
+    print("=== Non-streaming Response Example ===")
+
+    agent = ChatClientAgent(
+        chat_client=AzureResponsesClient(),
+        instructions="You are a helpful weather agent.",
+        tools=get_weather,
+    )
```
markwallace-microsoft, line 28: The pattern we want to adopt is

```python
client = AzureResponsesClient()
agent = client.create_agent(
        instructions="You are a helpful weather agent.",
        tools=get_weather,
    )
```

Longer term we want to create a shared Python package that just contain the AF agent abstractions. The Foundry and Azure OpenAI SDK teams can take a dependency on that so we get the best alignment. For 3P services we will still need to provide `agent_framework.<third-party>` packages.

dmytrostruk: Will be resolved in scope of separate PR as per agreement.

Issue #343, opened 2025-08-05
Single Agent: Long running agents support


markwallace-microsoft:

dmytrostruk:
Follow-up task for Python: https://github.com/microsoft/agent-framework/issues/2478

PR #364, opened 2025-08-06
Python: Draft: Prototype for one proposed design of background calls

peterychang:
Prototype for long running tasks
https://github.com/microsoft/agent-framework/issues/343

### DIFF
--- a/python/responses_background_test.py
+++ b/python/responses_background_test.py
@@ -2,7 +2,7 @@ import asyncio
 
 from typing import Annotated
 
-from agent_framework import ChatResponseUpdate, ChatMessage, FunctionResultContent, TextContent, ai_function
+from agent_framework import ChatResponseUpdate, ChatMessage, TextContent, ai_function
 from agent_framework.openai import OpenAIResponsesClient
 
 @ai_function
--- a/python/packages/main/agent_framework/openai/_responses_client.py
+++ b/python/packages/main/agent_framework/openai/_responses_client.py
@@ -577,7 +577,7 @@ class OpenAIResponsesClientBase(OpenAIHandler, RunnableChatClient):
                         )
                         response_tools.append(
                             WebSearchToolParam(
-                                type="web_search_preview",
+                                type="web_search",
                                 user_location=WebSearchUserLocation(
                                     type="approximate",
                                     city=location.get("city", None),
--- a/python/packages/main/pyproject.toml
+++ b/python/packages/main/pyproject.toml
@@ -23,7 +23,7 @@ classifiers = [
   "Typing :: Typed",
 ]
 dependencies = [
-    "openai>=1.94.0",
+    "openai>=1.103.0",
     "pydantic>=2.11.7",
     "pydantic-settings>=2.10.1",
     "typing-extensions>=4.14.0",
--- a/python/packages/main/tests/openai/test_openai_background.py
+++ b/python/packages/main/tests/openai/test_openai_background.py
@@ -8,7 +8,6 @@ import pytest
 from pydantic import BaseModel
 
 from agent_framework import ChatClient, ChatMessage, ChatResponse, ChatResponseUpdate, ResponseStatus, ai_function
-from agent_framework.exceptions import ServiceResponseException
 from agent_framework.openai import OpenAIResponsesClient
 
 skip_if_openai_integration_tests_disabled = pytest.mark.skipif(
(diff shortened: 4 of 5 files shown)


python/packages/main/agent_framework/_types.py
```diff
@@ -807,6 +809,59 @@ def __init__(
...
+    Attributes:
+        type: The type of content, which is always "async_message" for this class.
+        details: The usage information, including input and output token counts, and any additional counts.
+        additional_properties: Optional additional properties associated with the content.
+        raw_representation: Optional raw representation of the content.
+    """
+
+    message_id: str | None = None
+    type: Literal["async_message"] = "async_message"  # type: ignore[assignment]
+    _messages: Future[list["ChatMessage"]] = PrivateAttr()
```
eavanvalkenburg: Should we set a default factory?

python/packages/main/agent_framework/_clients.py
```diff
@@ -642,6 +644,28 @@ def service_url(self) -> str | None:
...
 
+class LongRunningChatClient(ChatClientBase):
+    """A chat client that supports long-running requests.
+
+    This is a base class for chat clients that can handle long-running requests.
+    It should be extended by specific implementations that support long-running requests.
+    """
+
+    @abstractmethod
+    def get_long_running_response(
```
eavanvalkenburg: we would have to extend the tracing and function calling decorators to this method

python/packages/main/agent_framework/_types.py
```diff
@@ -808,6 +809,40 @@ def __init__(
         )
 
 
+class AsyncMessageContent(AIContent):
```
eavanvalkenburg: wouldn't it make more sense to have a `AsyncChatResponse` or just extend the existing `ChatResponse` to allow no content and add the required fields to allow you to poll for updates?

peterychang:
My case for adding the sender to the `AsyncMessageContent`. This is based on an autogen feature that allows direct agent-to-agent messaging

```python

class WriterAgent(AIAgent):
    model_client: OpenAIResponsesClient = OpenAIResponsesClient()
    _id: str = uuid4().hex

    @property
    def id(self) -> str:
        return self._id

    async def run(
        self,
        messages: str | ChatMessage | list[str] | list[ChatMessage] | None = None,
        *,
        thread: AgentThread | None = None,
        **kwargs: Any,
    ) -> AgentRunResponse:
        # normalize messages so they are always list[ChatMessage]
        messages: list[ChatMessage] = normalize_messages(messages) # type: ignore
        
        # hypothetical new type for agents to retrieve async message content
        if isinstance(messages[0].contents[0], RetrieveAsyncMessageContent): # type: ignore
            message_id = messages[0].contents[0].message_id
            resp = self.model_client.get_long_running_response(message_id=message_id)
            return AgentRunResponse(
                messages=resp.result(),
                thread=thread,
            )
        # This contains an AsyncMessageContent object
        chat_response = await self.model_client.get_response(
            background=True,
            store=True,
            messages=messages,
            thread=thread,
            **kwargs
        )
        # We can smooth this out by passing an optional ID into the get_response method
        async_message_content = chat_response.messages[-1].contents[0]
        assert isinstance(async_message_content, AsyncMessageContent)
        async_message_content.source = self.id

        return AgentRunResponse(
            messages=[chat_response],
            thread=thread,
        )

class EditorAgent(AIAgent):
    async def run(
        self,
        messages: str | ChatMessage | list[str] | list[ChatMessage] | None = None,
        *,
        thread: AgentThread | None = None,
        **kwargs: Any,
    ) -> AgentRunResponse:
        # normalize messages so they are always list[ChatMessage]
        messages: list[ChatMessage] = normalize_messages(messages) # type: ignore
        
        if isinstance(messages[-1].contents[0], AsyncMessageContent):
            async_message_content = messages[-1].contents[0]
            # send_message is an autogen method that allows for direct agent-to-agent messaging
            self.send_message(message=ChatMessage( # type:ignore
                contents=[RetrieveAsyncMessageContent( # type:ignore
                    async_message_content.message_id)],
                recipient=async_message_content.source)
            )
```

peterychang:
> having had a look, I think what we could consider (and would love your thoughts) is that we add fields to ChatResponse and ChatResponseUpdate that indicate it is a long_running/background response and then we have a extra method for those services that support it, where you can pass the response to and ask it either to check or to wait (or two methods). and we can indicate with overloads on `get_response` and `get_streaming_response` that if `background` (or `long_running`) is True, that it will return a FutureChatResponse or something...

The `AsyncMessageContent` is the indicator that there is a long_running response. I don't know what extending `ChatResponse` would buy us over this. Alternatively, we could go with how OpenAI does this and have a `status` field in the response.

If we didn't need to worry about serializing/deserializing the chat history, this would be an ideal place to use a `Future`, since we could modify `ChatResponse` to handle it completely transparently outside of the client.

I've given a lot of thought about ways to make this change serializable while allowing all other components to be agnostic of whether a task is long running or not, or at the very least minimizing the amount of specialized code anybody would need to implement. So far, this is the best I've come up with

peterychang:
`get_response` and `get_streaming_response` both working. Currently testing with tools and attempting to add sane function overloads so that having a `message_id`/`conversation_id` OR `messages` is a valid call.

~OpenAI AsyncResponses client's `stream` call is broken for the background parameter. See https://github.com/openai/openai-python/issues/2579~
Bug is fixed as of package version 1.103.0

python/responses_background_test.py
```diff
@@ -0,0 +1,112 @@
+import asyncio
```
peterychang, line 1: This file is separate and distinct from the actual test file. This is the file I use as a playground for testing any changes made to the API

PR #365, opened 2025-08-06
.NET: Avoid double responses when streaming

adityamandaleeka:
Pirate was speaking double 😆. When `RunStreamingAsync` received a completed message, it was yielding all message content again via `ToAgentRunResponseUpdates()`, duplicating content that had already been streamed. Now it just breaks on completion since the completed status only signals the end of the stream.


DeagleGross:
tests need to be fixed - for example `RunStreamingAsync_NonFailedStatus_YieldsAgentRunResponseUpdateAsync`.

adityamandaleeka:
Yes, and I'll add a new test as well.

PR #366, opened 2025-08-07
Python: introduce workflow checkpointing

### LINKED ISSUE #316: Workflow checkpointing
Allow workflow checkpointing. Features including:
1. Save the states of all executors in a workflow after a super step has finished.
2. Save the shared states of a workflow after a super step has finished.
3. Save the pending messages to be delivered in the next super step.
4. API to checkpoint a workflow and return a checkpoint object.
5. API to resume a workflow from a checkpoint object.

moonbox3:
### Motivation and Context

As part of running a workflow, we want to provide the users with the option to enable checkpointing. At this point in time, if checkpointing is enabled, we will checkpoint at each superstep boundary. Users may then resume an existing workflow, or create a new workflow object and run it via:

- `run_streaming_from_checkpoint(...)`
- `run_from_checkpoint(...)`

Which are similar in naming to the `run_streaming(...)` or `run(...)` apis that we currently have.

Checkpointing is enabled in the following way:

```python
workflow = (
    WorkflowBuilder()
    .add_edge(upper_case_executor, reverse_text_executor)
    .set_start_executor(upper_case_executor)
    .with_checkpointing(checkpoint_storage=checkpoint_storage) # enabled here
    .build()
)
```

We currently support an `InMemoryCheckpointStorage` and a `FileCheckpointStorage` with potentially others to come soon.

When the user wants to resume from a particular checkpoint, they must provide the checkpoint_id (str) as well as the checkpoint_storage, so we can rehydrate the workflow and get it in the proper place for continued execution through the graph. The reason why we're requiring `checkpoint_storage` to be passed in to the `run_streaming_from_checkpoint` method is so that we don't tightly couple `.with_checkpointing()` as the only way to execute from a checkpoint - a user may not want checkpointing enabled for a workflow, but want to resume the workflow from a previous checkpoint.

### Description

Introduces checkpointing functionality.
- Adds a sample
- Adds unit tests
- Closes #316

### Contribution Checklist

- [X] I didn't break anyone :smile:

### DIFF
--- a/python/packages/main/agent_framework/_tools.py
+++ b/python/packages/main/agent_framework/_tools.py
@@ -4,13 +4,16 @@ import inspect
 from collections.abc import Awaitable, Callable
 from functools import wraps
 from time import perf_counter
-from typing import Annotated, Any, Generic, Protocol, TypeVar, get_args, get_origin, runtime_checkable
+from typing import TYPE_CHECKING, Annotated, Any, Generic, Protocol, TypeVar, get_args, get_origin, runtime_checkable
 
 from opentelemetry import metrics, trace
 from pydantic import BaseModel, Field, create_model
 
 from ._logging import get_logger
 from .telemetry import GenAIAttributes, start_as_current_span
+
+if TYPE_CHECKING:
+    from ._types import AIContents
 
 tracer: trace.Tracer = trace.get_tracer("agent_framework")
 meter: metrics.Meter = metrics.get_meter_provider().get_meter("agent_framework")
@@ -216,6 +219,45 @@ def ai_function(
     return decorator(func) if func else decorator  # type: ignore[reportReturnType, return-value]
 
 
+def _parse_inputs(
+    inputs: "AIContents | dict[str, Any] | str | list[AIContents | dict[str, Any] | str] | None",
+) -> list["AIContents"]:
+    """Parse the inputs for a tool, ensuring they are of type AIContents."""
+    if inputs is None:
+        return []
+
+    from ._types import AIContent, DataContent, HostedFileContent, HostedVectorStoreContent, UriContent
+
+    parsed_inputs: list["AIContents"] = []
+    if not isinstance(inputs, list):
+        inputs = [inputs]
+    for input_item in inputs:
+        if isinstance(input_item, str):
+            # If it's a string, we assume it's a URI or similar identifier.
+            # Convert it to a UriContent or similar type as needed.
+            parsed_inputs.append(UriContent(uri=input_item, media_type="text/plain"))
+        elif isinstance(input_item, dict):
+            # If it's a dict, we assume it contains properties for a specific content type.
+            # we check if the required keys are present to determine the type.
+            if "uri" in input_item:
+                parsed_inputs.append(
+                    UriContent(**input_item) if "media_type" in input_item else DataContent(**input_item)
+                )
+            elif "file_id" in input_item:
+                parsed_inputs.append(HostedFileContent(**input_item))
+            elif "vector_store_id" in input_item:
+                parsed_inputs.append(HostedVectorStoreContent(**input_item))
+            elif "data" in input_item:
+                parsed_inputs.append(DataContent(**input_item))
+            else:
+                raise ValueError(f"Unsupported input type: {input_item}")
+        elif isinstance(input_item, AIContent):
+            parsed_inputs.append(input_item)
+        else:
+            raise TypeError(f"Unsupported input type: {type(input_item).__name__}. Expected AIContents or dict.")
+    return parsed_inputs
+
+
 class HostedCodeInterpreterTool(AITool):
     """Represents a hosted tool that can be specified to an AI service to enable it to execute generated code.
 
@@ -226,6 +268,7 @@ class HostedCodeInterpreterTool(AITool):
     def __init__(
         self,
         name: str = "code_interpreter",
+        inputs: "AIContents | dict[str, Any] | str | list[AIContents | dict[str, Any] | str] | None" = None,
         description: str | None = None,
         additional_properties: dict[str, Any] | None = None,
     ):
@@ -233,10 +276,19 @@ class HostedCodeInterpreterTool(AITool):
 
         Args:
             name: The name of the tool. Defaults to "code_interpreter".
+            inputs: A list of contents that the tool can accept as input. Defaults to None.
+                This should mostly be HostedFileContent or HostedVectorStoreContent.
+                Can also be DataContent, depending on the service used.
+                When supplying a list, it can contain:
+                - AIContents instances
+                - dicts with properties for AIContents (e.g., {"uri": "http://example.com", "media_type": "text/html"})
+                - strings (which will be converted to UriContent with media_type "text/plain").
+                If None, defaults to an empty list.
             description: A description of the tool.
             additional_properties: Additional properties associated with the tool, specific to the service used.
         """
         self.name = name
+        self.inputs = _parse_inputs(inputs)
         self.description = description
         self.additional_properties = additional_properties
 
--- a/python/packages/main/agent_framework/_types.py
+++ b/python/packages/main/agent_framework/_types.py
@@ -99,6 +99,8 @@ __all__ = [
     "FunctionCallContent",
     "FunctionResultContent",
     "GeneratedEmbeddings",
+    "HostedFileContent",
+    "HostedVectorStoreContent",
     "SpeechToTextOptions",
     "StructuredResponse",
     "TextContent",
@@ -1036,6 +1038,68 @@ class UsageContent(AIContent):
         )
 
 
+class HostedFileContent(AIContent):
+    """Represents a hosted file content.
+
+    Attributes:
+        file_id: The identifier of the hosted file.
+        type: The type of content, which is always "hosted_file" for this class.
+        additional_properties: Optional additional properties associated with the content.
+        raw_representation: Optional raw representation of the content.
+
+    """
+
+    type: Literal["hosted_file"] = "hosted_file"  # type: ignore[assignment]
+    file_id: str
+
+    def __init__(
+        self,
+        file_id: str,
+        *,
+        additional_properties: dict[str, Any] | None = None,
+        raw_representation: Any | None = None,
+        **kwargs: Any,
+    ) -> None:
+        """Initializes a HostedFileContent instance."""
+        super().__init__(
+            file_id=file_id,  # type: ignore[reportCallIssue]
+            additional_properties=additional_properties,
+            raw_representation=raw_representation,
+            **kwargs,
+        )
+
+
+class HostedVectorStoreContent(AIContent):
+    """Represents a hosted vector store content.
+
+    Attributes:
+        vector_store_id: The identifier of the hosted vector store.
+        type: The type of content, which is always "hosted_vector_store" for this class.
+        additional_properties: Optional additional properties associated with the content.
+        raw_representation: Optional raw representation of the content.
+
+    """
+
+    type: Literal["hosted_vector_store"] = "hosted_vector_store"  # type: ignore[assignment]
+    vector_store_id: str
+
+    def __init__(
+        self,
+        vector_store_id: str,
+        *,
+        additional_properties: dict[str, Any] | None = None,
+        raw_representation: Any | None = None,
+        **kwargs: Any,
+    ) -> None:
+        """Initializes a HostedVectorStoreContent instance."""
+        super().__init__(
+            vector_store_id=vector_store_id,  # type: ignore[reportCallIssue]
+            additional_properties=additional_properties,
+            raw_representation=raw_representation,
+            **kwargs,
+        )
+
+
 AIContents = Annotated[
     TextContent
     | DataContent
@@ -1044,7 +1108,9 @@ AIContents = Annotated[
     | FunctionCallContent
     | FunctionResultContent
     | ErrorContent
-    | UsageContent,
+    | UsageContent
+    | HostedFileContent
+    | HostedVectorStoreContent,
     Field(discriminator="type"),
 ]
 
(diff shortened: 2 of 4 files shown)


python/packages/workflow/agent_framework_workflow/__init__.py
```diff
@@ -22,6 +28,13 @@
     RequestInfoMessage,
     handler,
 )
+from ._runner_context import (
+    CheckpointableInProcRunnerContext,
+    CheckpointableRunnerContext,
+    InProcRunnerContext,
```
TaoChenOSU, line 35: Do we need to expose this and the following classes?

python/packages/workflow/agent_framework_workflow/_checkpoint.py
```diff
@@ -0,0 +1,181 @@
...
+
+    # Metadata
+    metadata: dict[str, Any] = field(default_factory=dict)  # type: ignore[misc]
+    version: str = "1.0"
+
+
+class CheckpointStorage(Protocol):
+    """Protocol for checkpoint storage backends."""
+
+    def save_checkpoint(self, checkpoint: WorkflowCheckpoint) -> str:
```
TaoChenOSU: Should we make the methods async for future storage types?

python/packages/workflow/agent_framework_workflow/_checkpoint.py
```diff
@@ -0,0 +1,181 @@
...
+        self.storage_path.mkdir(parents=True, exist_ok=True)
+        logger.info(f"Initialized file checkpoint storage at {self.storage_path}")
+
+    def save_checkpoint(self, checkpoint: WorkflowCheckpoint) -> str:
+        """Save a checkpoint and return its ID."""
+        file_path = self.storage_path / f"{checkpoint.checkpoint_id}.json"
+        checkpoint_dict = asdict(checkpoint)
+
+        # Note: Using synchronous I/O for simplicity in this prototype
+        # In production, consider using aiofiles for truly async I/O
```
TaoChenOSU: I think we should get ready for production scenarios.

python/packages/workflow/agent_framework_workflow/_runner.py
```diff
@@ -22,51 +23,108 @@ def __init__(
...
         """Run the workflow until no more messages are sent."""
+        if self._running:
+            raise RuntimeError("Runner is already running.")
+
+        self._running = True
         try:
-            if self._is_running:
-                raise RuntimeError("Runner is already running.")
-            self._is_running = True
+            # Process any events from initial execution before checkpointing
```
TaoChenOSU, line 74: Question: should checkpoints contain events?

python/packages/workflow/agent_framework_workflow/_runner_context.py
```diff
@@ -74,6 +77,43 @@ async def has_events(self) -> bool:
         ...
 
 
+@runtime_checkable
+class CheckpointableRunnerContext(RunnerContext, Protocol):
```
TaoChenOSU: nit: could we combine these functionalities to the RunnerContext?

python/packages/workflow/agent_framework_workflow/_workflow.py
```diff
@@ -2,18 +2,23 @@
...
 from ._executor import Executor, RequestInfoExecutor
-from ._runner import DEFAULT_MAX_ITERATIONS, Runner
+from ._runner import Runner
 from ._runner_context import InProcRunnerContext, RunnerContext
 from ._shared_state import SharedState
 from ._validation import validate_workflow_graph
 from ._workflow_context import WorkflowContext
 
+# Default maximum iterations for workflow convergence
+DEFAULT_MAX_ITERATIONS = 100
```
TaoChenOSU: I am thinking we can create a module for constants. What do you think?

python/packages/workflow/agent_framework_workflow/_workflow.py
```diff
@@ -100,15 +113,24 @@ def executors(self) -> list[Executor]:
...
+    ) -> AsyncIterable[WorkflowEvent]:
+        """Run the workflow with a starting message and stream events.
 
         Args:
             message: The message to be sent to the starting executor.
 
         Yields:
             WorkflowEvent: The events generated during the workflow execution.
         """
+        # Reset context state for new workflow run if it's checkpointable
```
TaoChenOSU: Question: why is this necessary in this API?

python/packages/workflow/agent_framework_workflow/_workflow.py
```diff
@@ -347,4 +596,15 @@ def build(self) -> Workflow:
 
         validate_workflow_graph(self._edges, self._start_executor)
 
-        return Workflow(self._edges, self._start_executor, InProcRunnerContext(), self._max_iterations)
+        # Choose appropriate context based on checkpoint storage
```
TaoChenOSU: I don't think this is scalable. Do we need another type on context just for checkpointing?

python/packages/workflow/agent_framework_workflow/_workflow.py
```diff
@@ -186,7 +303,27 @@ async def send_responses(self, responses: dict[str, Any]) -> WorkflowRunResult:
         Returns:
             A WorkflowRunResult instance containing a list of events generated during the workflow execution.
         """
-        events = [event async for event in self.send_responses_streaming(responses)]
```
TaoChenOSU: Why couldn't we use `send_response_streaming`?

python/packages/workflow/agent_framework_workflow/_runner.py
```diff
@@ -22,51 +23,108 @@ def __init__(
...
 
+                # Update context with current iteration state immediately
+                await self._update_context_with_shared_state()
+
+                logger.info(f"Completed superstep {self._iteration}")
+
+                # Check what state we have before checkpointing
+                has_messages_before = await self._ctx.has_messages()
+                has_events_before = await self._ctx.has_events()
+                logger.debug(f"Before checkpoint: messages={has_messages_before}, events={has_events_before}")
```
TaoChenOSU: nit: can we remove these?

python/packages/workflow/agent_framework_workflow/_runner.py
```diff
@@ -22,51 +23,108 @@ def __init__(
...
                     events = await self._ctx.drain_events()
                     for event in events:
+                        logger.debug(f"Yielding event: {event}")
                         yield event
 
-                if not await self._ctx.has_messages():
+                has_messages = await self._ctx.has_messages()
+                has_events = await self._ctx.has_events()
+                logger.debug(f"Has messages after superstep {self._iteration}: {has_messages}")
+                logger.debug(f"Has events after superstep {self._iteration}: {has_events}")
```
TaoChenOSU: nit: can we remove these?

python/packages/workflow/agent_framework_workflow/_runner.py
```diff
@@ -22,51 +23,108 @@ def __init__(
...
                     break
-            else:
-                raise RuntimeError(f"Runner did not converge after {self._max_iterations} iterations.")
-        finally:
-            self._is_running = False
+
+            # Check if we reached max iterations without convergence
+            if self._iteration >= self._max_iterations:
+                has_messages = await self._ctx.has_messages()
+                if has_messages:
```
TaoChenOSU: This is not necessary because it must still have message otherwise it will break out of the loop.

python/packages/workflow/agent_framework_workflow/_checkpoint.py
```diff
@@ -0,0 +1,181 @@
...
+    def list_checkpoints(self, workflow_id: str | None = None) -> list[WorkflowCheckpoint]:
+        """List checkpoint objects. If workflow_id is provided, filter by that workflow."""
+        ...
+
+    def delete_checkpoint(self, checkpoint_id: str) -> bool:
+        """Delete a checkpoint by ID."""
+        ...
+
+
+class MemoryCheckpointStorage:
```
eavanvalkenburg: For vector stores we use `InMemory...` might be good to keep consistent

TaoChenOSU:
We probably don't need a separate context for checkpointing. The context only needs to know how to use the checkpoint storage for saving and loading.

python/packages/workflow/agent_framework_workflow/__init__.py
moonbox3: Checkpoint specific context removed.

python/packages/workflow/agent_framework_workflow/_runner.py
moonbox3: As we talked about: no. Removing any ref to events regarding checkpoints.

python/packages/workflow/agent_framework_workflow/_runner_context.py
moonbox3: Yes, as we talked about, they're now combined into one.

python/packages/workflow/agent_framework_workflow/_checkpoint.py
moonbox3: Thanks. Lost the naming change in a mess of merge conflicts. Back to `InMemoryCheckpointStorage`.

python/packages/workflow/agent_framework_workflow/_workflow.py
moonbox3: This got mucked up in the nasty merge conflicts I had. Will revert.

python/packages/workflow/agent_framework_workflow/_runner.py
```diff
@@ -105,6 +147,128 @@ async def _deliver_messages_inner(
         ]
         await asyncio.gather(*tasks)
 
+    async def _create_checkpoint_if_enabled(self, checkpoint_type: str) -> str | None:
```
TaoChenOSU, line 150: nit: should we have better-defined checkpoint types?

python/packages/workflow/agent_framework_workflow/_runner.py
```diff
@@ -105,6 +147,128 @@ async def _deliver_messages_inner(
...
+
+        try:
+            # Auto-snapshot executor states (opt-in via snapshot_state attribute or method)
+            await self._auto_snapshot_executor_states()
+            await self._update_context_with_shared_state()
+            label = checkpoint_type
+            metadata = {
+                "label": label,
+                "superstep": self._iteration,
+                "checkpoint_type": checkpoint_type,
```
TaoChenOSU: `label` and `checkpoint_type` are the same.

python/packages/workflow/agent_framework_workflow/_runner.py
TaoChenOSU, line 150: nit: This method probably doesn't need to return anything.

python/packages/workflow/agent_framework_workflow/_runner.py
```diff
@@ -105,6 +147,128 @@ async def _deliver_messages_inner(
...
+    async def restore_from_checkpoint(self, checkpoint_id: str) -> bool:
+        """Restore workflow state from a checkpoint.
+
+        Args:
+            checkpoint_id: The ID of the checkpoint to restore from
+
+        Returns:
+            True if restoration was successful, False otherwise
+        """
+        if not self._ctx.has_checkpointing():
```
TaoChenOSU, line 235: nit:  I see that `has_checkpointing` is determined by if the context has a checkpoint storage. I feel like a workflow should still be able to restart from a checkpoint even there isn't a checkpoint storage provided.

python/samples/getting_started/workflow/step_07_checkpoint_resume.py
```diff
@@ -0,0 +1,205 @@
...
+
+class UpperCaseExecutor(Executor):
+    @handler(output_types=[str])
+    async def to_upper_case(self, text: str, ctx: WorkflowContext) -> None:
+        result = text.upper()
+        print(f"UpperCaseExecutor: '{text}' -> '{result}'")
+        # Persist executor state into checkpointable context
+        prev = await ctx.get_state() or {}
+        count = int(prev.get("count", 0)) + 1
+        await ctx.set_state({
```
TaoChenOSU, line 48: nit: I think we shouldn't require users to set the states in the handlers. Instead, we can provide an abstract method that user must override to save the state for when the checkpointing logic happens.

moonbox3: Let's discuss. Not sure how I feel requiring user's to override a method just so it can set state.

python/packages/workflow/agent_framework_workflow/_runner.py
moonbox3: `has_checkpointing` intentionally signals persistent, context-managed checkpoints (ID-addressable) via a configured storage; without bound storage, the runner can’t fetch by ID. A workflow built without storage can still resume by passing `checkpoint_storage` to `run_streaming_from_checkpoint` (restores messages/shared_state/executor states); we can rename the flag to `has_persistent_checkpointing` to make this clearer.

python/packages/workflow/agent_framework_workflow/_checkpoint.py
```diff
@@ -0,0 +1,199 @@
...
+    messages: dict[str, list[dict[str, Any]]] = field(default_factory=dict)  # type: ignore[misc]
+    shared_state: dict[str, Any] = field(default_factory=dict)  # type: ignore[misc]
+    executor_states: dict[str, dict[str, Any]] = field(default_factory=dict)  # type: ignore[misc]
+
+    # Runtime state
+    iteration_count: int = 0
+    max_iterations: int = DEFAULT_MAX_ITERATIONS
+
+    # Metadata
+    metadata: dict[str, Any] = field(default_factory=dict)  # type: ignore[misc]
```
ekzhu, line 36: Use a TypedDict?

moonbox3: Thanks for the feedback. Can update in the next iteration.

PR #374, opened 2025-08-08
Python: added AIAnnotation types and extra tests

eavanvalkenburg:
### Motivation and Context

In line with MEAI, adding AIAnnotations that we can use to capture the annotations from different services.

### Contribution Checklist

### DIFF
--- a/python/packages/main/agent_framework/_types.py
+++ b/python/packages/main/agent_framework/_types.py
@@ -334,13 +334,13 @@ class AIAnnotation(AFBaseModel):
     """Base class for all AI Annotation types.
 
     Args:
-        type: The type of content, which is always "ai" for this class.
+        type: The type of content, which is always "ai_annotation" for this class.
         additional_properties: Optional additional properties associated with the content.
         raw_representation: Optional raw representation of the content from an underlying implementation.
 
     """
 
-    type: Literal["ai"] = "ai"
+    type: Literal["ai_annotation"] = "ai_annotation"
     annotated_regions: list[AnnotatedRegions] | None = None
     additional_properties: dict[str, Any] | None = None
     raw_representation: Any | None = Field(default=None, repr=False)
@@ -438,7 +438,7 @@ class TextContent(AIContent):
         The following things happen:
         The text is concatenated.
         The annotations are combined.
-        The additional properties are merged, with the values of shared keys of the 'self' instance taking precedence.
+        The additional properties are merged, with the values of shared keys of the first instance taking precedence.
         The raw_representations are combined into a list of them, if they both have one.
         """
         if not isinstance(other, TextContent):
@@ -473,7 +473,7 @@ class TextContent(AIContent):
         The following things happen:
         The text is concatenated.
         The annotations are combined.
-        The additional properties are merged, with the values of shared keys of the 'self' instance taking precedence.
+        The additional properties are merged, with the values of shared keys of the first instance taking precedence.
         The raw_representations are combined into a list of them, if they both have one.
         """
         if not isinstance(other, TextContent):
@@ -542,7 +542,7 @@ class TextReasoningContent(AIContent):
         The following things happen:
         The text is concatenated.
         The annotations are combined.
-        The additional properties are merged, with the values of shared keys of the 'self' instance taking precedence.
+        The additional properties are merged, with the values of shared keys of the first instance taking precedence.
         The raw_representations are combined into a list of them, if they both have one.
         """
         if not isinstance(other, TextReasoningContent):
@@ -574,7 +574,7 @@ class TextReasoningContent(AIContent):
         The following things happen:
         The text is concatenated.
         The annotations are combined.
-        The additional properties are merged, with the values of shared keys of the 'self' instance taking precedence.
+        The additional properties are merged, with the values of shared keys of the first instance taking precedence.
         The raw_representations are combined into a list of them, if they both have one.
         """
         if not isinstance(other, TextReasoningContent):


python/packages/main/agent_framework/_types.py
```diff
@@ -351,6 +432,48 @@ def __init__(
             **kwargs,
         )
 
+    def __add__(self, other: "TextContent") -> "TextContent":
+        """Concatenate two TextContent instances."""
+        if not isinstance(other, TextContent):
+            raise TypeError("Incompatible type")
+        if self.raw_representation is None:
```
peterychang, line 446: adding raw representations together doesn't seem to make much sense? It might be ok if all the contents have one, but if some do and others don't, i'm not sure what use this would have

python/packages/main/agent_framework/_types.py
```diff
@@ -351,6 +432,48 @@ def __init__(
...
+        if self.annotations is None:
+            annotations = other.annotations
+        elif other.annotations is None:
+            annotations = self.annotations
+        else:
+            annotations = self.annotations + other.annotations
+        return TextContent(
+            text=self.text + other.text,
+            annotations=annotations,
+            additional_properties={**(self.additional_properties or {}), **(other.additional_properties or {})},
```
peterychang: what happens if both have properties with the same name but different values?

python/packages/main/agent_framework/_types.py
```diff
@@ -351,6 +432,48 @@ def __init__(
...
+
+    def __iadd__(self, other: "TextContent") -> Self:
+        """In-place concatenation of two TextContent instances."""
+        if not isinstance(other, TextContent):
+            raise TypeError("Incompatible type")
+        self.text += other.text
+        if self.additional_properties is None:
+            self.additional_properties = {}
+        if other.additional_properties:
+            self.additional_properties.update(other.additional_properties)
```
peterychang: is it fine for `other`'s colliding keys to override?

python/packages/main/agent_framework/openai/_chat_client.py
```diff
@@ -177,11 +177,11 @@ def _get_tool_calls_from_chat_choice(self, choice: Choice | ChunkChoice) -> list
         content = choice.message if isinstance(choice, Choice) else choice.delta
         if content and (tool_calls := getattr(content, "tool_calls", None)) is not None:
             for tool in cast(list[ChatCompletionMessageToolCall] | list[ChoiceDeltaToolCall], tool_calls):
-                if tool.function:
+                if tool.function:  # type: ignore
```
peterychang: did something change that we need the `type: ignore` here?

python/packages/main/agent_framework/_types.py
eavanvalkenburg: I'm hesitant of throwing one away when we combine two pieces of content (and the scenario is mostly when combining chunks from a streaming response), that's why if both have a raw representation, we put them in a list so they are maintained

python/packages/main/agent_framework/_types.py
eavanvalkenburg: clarified and made changes that "self"s additional properties take precedence.

python/packages/main/agent_framework/_types.py
eavanvalkenburg: clarified and changed

python/packages/main/agent_framework/openai/_chat_client.py
eavanvalkenburg: not sure, but I limited the reach now, I'll check when I revisit this code

python/packages/main/agent_framework/_types.py
```diff
@@ -351,6 +432,69 @@ def __init__(
             **kwargs,
         )
 
+    def __add__(self, other: "TextContent") -> "TextContent":
+        """Concatenate two TextContent instances.
+
+        The following things happen:
+        The text is concatenated.
+        The annotations are combined.
+        The additional properties are merged, with the values of shared keys of the 'self' instance taking precedence.
```
peterychang: in this case, `self` would be the first object, right?

python/packages/main/agent_framework/_types.py
```diff
@@ -297,20 +301,96 @@ def _finalize_response(response: "ChatResponse | AgentRunResponse") -> None:
...
+    """Base class for all AI Annotation types.
+
+    Args:
+        type: The type of content, which is always "ai" for this class.
+        additional_properties: Optional additional properties associated with the content.
+        raw_representation: Optional raw representation of the content from an underlying implementation.
+
+    """
+
+    type: Literal["ai"] = "ai"
```
dmytrostruk: Can it be `ai_annotation`? `ai` seems too generic.
```suggestion
    type: Literal["ai_annotation"] = "ai_annotation"
```

python/packages/main/agent_framework/_types.py
```diff
@@ -297,20 +301,96 @@ def _finalize_response(response: "ChatResponse | AgentRunResponse") -> None:
...
+        url: The URL of the cited content.
+        file_id: The file identifier of the cited content, if applicable.
+        tool_name: The name of the tool that generated the citation, if applicable.
+        snippet: A snippet of the cited content, if applicable.
+        annotated_regions: A list of regions that have been annotated with this citation.
+        additional_properties: Optional additional properties associated with the content.
+        raw_representation: Optional raw representation of the content from an underlying implementation.
+    """
+
+    type: Literal["citation"] = "citation"  # type: ignore[assignment]
```
dmytrostruk, line 364: If `AIAnnotation` type will be `ai_annotation`, this one can be `citation_annotation`.

python/packages/main/agent_framework/_types.py
eavanvalkenburg: yes, updated

python/packages/main/agent_framework/_types.py
eavanvalkenburg: good one, fixed

python/packages/main/agent_framework/_types.py
eavanvalkenburg: this is the type discriminator used in MEAI, so following that: https://github.com/dotnet/extensions/blob/163820068e1762c574bedafaf8f759f42123293e/src/Libraries/Microsoft.Extensions.AI.Abstractions/Contents/AIAnnotation.cs#L13

PR #375, opened 2025-08-08
.NET: Expand Persistent AIAgent Extensions to support AITools .

rogerbarreto:
### Motivation and Context

To simplify the creation of `AIAgents` for the `Persistent` package, we added a new `MEAI.AITool` type argument extension for the creation of persistent agents.

As part of the PR I also change the persistent sample to make usage of the function calling feature.

This makes it easier to provide Function Tools instead of manually handling and creating `IEnumerable<ToolDefinitions>` lists and `ToolResources` types directly.

As part of this change I also identify a potential problem in the Backend, where if I provide the toolresources and tooldefinitions at the agent creation, they are not invoked in the runtime, due to that instead of converting the AITools into resources and definitions, I'm using the default behavior of the `ChatClientAgent` where when it get a ChatOptions with tools it uses across it's life, merging to any existing tools provided in a per-request basis.


dotnet/src/Microsoft.Extensions.AI.Agents.AzureAI/PersistentAgentsClientExtensions.cs
```diff
@@ -89,4 +89,57 @@ public static async Task<ChatClientAgent> CreateAIAgentAsync(
...
+             topP: topP,
+             responseFormat: responseFormat,
+             metadata: metadata,
+             cancellationToken: cancellationToken).ConfigureAwait(false);
+
+        // Get a local proxy for the agent to work with.
+        return await persistentAgentsClient.GetAIAgentAsync(
+            createPersistentAgentResponse.Value.Id,
+
+            // As of now, providing the tools as definitions and resources at the Agent creation didn't work (no functions were invoked).
```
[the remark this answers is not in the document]
rogerbarreto: Tracking issue created:

- https://github.com/microsoft/agent-framework/issues/376

dotnet/src/Microsoft.Extensions.AI.Agents.AzureAI/PersistentAgentsClientExtensions.cs
```diff
@@ -89,4 +89,58 @@ public static async Task<ChatClientAgent> CreateAIAgentAsync(
...
+    public static async Task<ChatClientAgent> CreateAIAgentAsync(
+         this PersistentAgentsClient persistentAgentsClient,
+        string model,
+        IEnumerable<AITool>? tools,
+        string? name = null,
+        string? description = null,
+        string? instructions = null,
+        float? temperature = null,
+        float? topP = null,
+        BinaryData? responseFormat = null,
```
westey-m, line 117: Consider also taking the MEAI response format type here instead of BinaryData.
The benefit of the MEAI types (and this overload) in my mind is to make integration with the rest of the code base easier, and the MEAI response types do that as well, as you can just point at a class, and get the matching output format.  Similar to tools, you can just point at a method, and get the right bits to advertise to the service.

dotnet/samples/GettingStarted/Providers/AIAgent_With_AzureAIAgentsPersistent.cs
```diff
@@ -30,13 +36,16 @@ public async Task GetWithAzureAIAgentsPersistent()
...
+        // Get a server side AI Agent.
+        AIAgent agent = await persistentAgentsClient.GetAIAgentAsync(
+            agentId: persistentAgent.Value.Id,
+            chatOptions: new() { Tools = [AIFunctionFactory.Create(GetWeather)] }); // Add a tool to the agent.
 
         // Start a new thread for the agent conversation.
         AgentThread thread = agent.GetNewThread();
 
         // Respond to user input
+        await RunAgentAsync("What's the weather in Amsterdam?"); // This will invoke the GetWeather tool.
```
westey-m, line 48: Perhaps not the best sample to ask a joker agent about the weather, unless the weather is a joke ;)
Maybe worth considering splitting the function calling bit into a separate function that doesn't use a joker agent, or changing this one to just be weather related, e.g:
"What's the weather in Amsterdam?"
"And Paris?"

dotnet/src/Microsoft.Extensions.AI.Agents.AzureAI/PersistentAgentsClientExtensions.cs
rogerbarreto: That's a valid point and as we just had the MEAI new release with the hosted and AdditionalTools support the hack in this PR is not valid anymore. 

Will put this PR into Draft mode until my proposed changes make into `Azure.AI.Agents` where the clients there, accepts natively the `MEAI` types.

rogerbarreto:
We can revisit this proposal later under community demand.

PR #379, opened 2025-08-08
Python: added HostedFileContent and HostedVectorStoreContent

eavanvalkenburg:
### Motivation and Context

Added hosted file content and vector store content to the types.
Includes shortcuts in the HostedCodeInterpreterTool definition. 

### Contribution Checklist

### DIFF
--- a/python/packages/main/tests/main/test_types.py
+++ b/python/packages/main/tests/main/test_types.py
@@ -227,7 +227,7 @@ def test_hosted_vector_store_content():
     content = HostedVectorStoreContent(vector_store_id="vs-789", additional_properties={"version": 1})
 
     # Check the type and content
-    assert content.type == "vector_store"
+    assert content.type == "hosted_vector_store"
     assert content.vector_store_id == "vs-789"
     assert content.additional_properties["version"] == 1
 
@@ -241,7 +241,7 @@ def test_hosted_vector_store_content_minimal():
     content = HostedVectorStoreContent(vector_store_id="vs-101112")
 
     # Check the type and content
-    assert content.type == "vector_store"
+    assert content.type == "hosted_vector_store"
     assert content.vector_store_id == "vs-101112"
     assert content.additional_properties is None
     assert content.raw_representation is None


python/packages/main/agent_framework/_tools.py
```diff
@@ -197,6 +200,54 @@ def wrapper(f: Callable[..., ReturnT | Awaitable[ReturnT]]) -> AIFunction[Any, R
...
+        if isinstance(input_item, str):
+            # If it's a string, we assume it's a URI or similar identifier.
+            # Convert it to a UriContent or similar type as needed.
+            from ._types import UriContent
+
+            parsed_inputs.append(UriContent(uri=input_item, media_type="text/plain"))
+        elif isinstance(input_item, dict):
+            # If it's a dict, we assume it contains properties for a specific content type.
+            # Convert it to the appropriate AIContents subclass.
+            if "uri" in input_item:
```
peterychang, line 242: are these keys defined somewhere?

python/packages/main/agent_framework/_types.py
```diff
@@ -808,6 +810,68 @@ def __init__(
...
+
+    Attributes:
+        file_id: The identifier of the hosted file.
+        type: The type of content, which is always "file" for this class.
+        additional_properties: Optional additional properties associated with the content.
+        raw_representation: Optional raw representation of the content.
+
+    """
+
+    type: Literal["file"] = "file"  # type: ignore[assignment]
```
peterychang: Are we certain this type won't end up conflicting later on?

python/packages/main/agent_framework/_tools.py
eavanvalkenburg: these are the main content fields of the different types

python/packages/main/agent_framework/_types.py
eavanvalkenburg: good point, I've raised this in the equivalent MEAI PR

python/packages/main/agent_framework/_tools.py
eavanvalkenburg: yes, I clarified and updated the code a bit, they are basically the required args to the different types, the whole goal of this is to allow you to just quickly add a dict, instead of having to import some extra content type.

python/packages/main/agent_framework/_types.py
eavanvalkenburg: both dotnet and here updated, thanks @peterychang

PR #380, opened 2025-08-08
Python: added create_agent and removed need for Field in annotations

eavanvalkenburg:
### Motivation and Context

Small quality of life improvements:

- added a `create_agent` method to the ChatClientBase
- added parsing of a string in Annotated to a Field so that that works as description, other annotations are kept as is, so for more complex field def, you can still supply Field.

### Contribution Checklist

### DIFF
--- a/python/packages/main/agent_framework/_tools.py
+++ b/python/packages/main/agent_framework/_tools.py
@@ -138,7 +138,11 @@ class AIFunction(AITool, Generic[ArgsT, ReturnT]):
 
 
 def _parse_annotation(annotation: Any) -> Any:
-    """Parse a type annotation and return the corresponding type."""
+    """Parse a type annotation and return the corresponding type.
+
+    If the second annotation (after the type) is a string, then we convert that to a pydantic Field description.
+    The rest are returned as-is, allowing for multiple annotations.
+    """
     origin = get_origin(annotation)
     if origin is not None:
         args = get_args(annotation)


python/packages/main/agent_framework/_clients.py
```diff
@@ -641,6 +644,36 @@ def service_url(self) -> str | None:
...
+            name: The name of the agent.
+            instructions: The instructions for the agent.
+            tools: Optional list of tools to associate with the agent.
+            **kwargs: Additional keyword arguments to pass to the agent.
+                See ChatClientAgent for all the available options.
+
+        Returns:
+            An instance of ChatClientAgent.
+        """
+        from ._agents import ChatClientAgent
```
dmytrostruk, line 673: nit: Should we put this method in separate file from the organization point of view? Because currently it looks like `_agents` file references `_clients` file (because `ChatClientAgent` depends on `ChatClient`) and with this method `_clients` file references `_agents` file, which looks like circular dependency. Since CI doesn't fail, I assume Python allows that, but maybe it will be better to store such methods separately for easier maintainability. Feel free to skip this comment if needed.

python/samples/getting_started/agents/openai_chat_client/openai_chat_client_basic.py
```diff
@@ -21,8 +19,8 @@ async def non_streaming_example() -> None:
     """Example of non-streaming response (get the complete result at once)."""
     print("=== Non-streaming Response Example ===")
 
-    agent = ChatClientAgent(
-        chat_client=OpenAIChatClient(),
+    agent = OpenAIChatClient().create_agent(
```
dmytrostruk, line 22: nit: In .NET, I think the naming is `CreateAIAgent`, I'm wondering if we should follow that naming pattern in Python as well.

python/packages/main/agent_framework/_clients.py
eavanvalkenburg: yeah, that's why at the top of the file, we use `if TYPE_CHECKING` this is a way to import for type checking purposes only, and then when we actually need to import we do it in the method, making sure we don't end up with a circular import. Since there is no notion of a Extension method in python, this is the only way to do it!

python/samples/getting_started/agents/openai_chat_client/openai_chat_client_basic.py
eavanvalkenburg: I kind of prefer this, because it returns a type of agent, which happens to also be a AIAgent, but just longer, we can update if you want.

dmytrostruk: If the type is `AIAgent`, it feels like it should be `create_ai_agent`, but I like shorter version :)

python/packages/main/agent_framework/_clients.py
```diff
@@ -641,6 +644,36 @@ def service_url(self) -> str | None:
         """
         return None
 
+    def create_agent(
```
peterychang, line 647: Maybe its my inner C++ dev talking, but this feels like its breaking encapsulation?

python/samples/getting_started/agents/openai_chat_client/openai_chat_client_basic.py
eavanvalkenburg: Let's keep it like this, and check in by public preview

python/packages/main/agent_framework/_clients.py
eavanvalkenburg: Since there is no equivalent to extensions in python, this is the only way to get this behavior...

PR #388, opened 2025-08-11
Python: openai updates

eavanvalkenburg:
### Motivation and Context

Redoes the logic of tool handling and requests for OpenAI and derived classes.

Also does some work on the entra auth, so that we default to `DefaultAzureCredential` as less as possible.

### Contribution Checklist

### DIFF
--- a/python/packages/azure/tests/test_azure_assistants_client.py
+++ b/python/packages/azure/tests/test_azure_assistants_client.py
@@ -369,7 +369,9 @@ async def test_azure_assistants_client_with_existing_assistant() -> None:
         assistant_id = temp_client.assistant_id
 
         # Now test using the existing assistant
-        async with AzureAssistantsClient(assistant_id=assistant_id) as azure_assistants_client:
+        async with AzureAssistantsClient(
+            assistant_id=assistant_id, ad_credential=DefaultAzureCredential()
+        ) as azure_assistants_client:
             assert isinstance(azure_assistants_client, ChatClient)
             assert azure_assistants_client.assistant_id == assistant_id
 


python/packages/azure/agent_framework_azure/_shared.py
```diff
@@ -149,10 +155,15 @@ def get_azure_openai_auth_token(self, token_endpoint: str | None = None) -> str
         Raises:
             ServiceInitializationError: If the token endpoint is not provided.
         """
-        endpoint_to_use = token_endpoint or self.token_endpoint
+        if not credential:
+            from azure.identity import DefaultAzureCredential
+
+            credential = DefaultAzureCredential()
```
dmytrostruk: Since we are already touching this place, should we remove the usage of `DefaultAzureCredential` and make `credential` property as mandatory on clients?

python/packages/main/agent_framework/_clients.py
```diff
@@ -674,6 +658,16 @@ def create_agent(
 
         return ChatClientAgent(chat_client=self, name=name, instructions=instructions, tools=tools, **kwargs)
 
+    def update_agent_name(self, agent_name: str | None) -> None:
```
dmytrostruk: I'm not sure if `ChatClient` should know anything about `agent` or be able to update `agent_name`. If we take a regular OpenAI chat completion or similar endpoint type, there is no `agent_name` in this API. The developers should be able to use `ChatClient` protocol even without agents, and the API could be confusing when `ChatClient` will have `update_agent_name` public method for the models that don't have this property available. 

It works in case of `create_agent` method, because this method is basically a wrapper that allows to move to another type `ChatClient -> ChatClientAgent`, but `update_agent_name` doesn't really fit into `ChatClient`. Can we place this method on `ChatClientAgent` level for example?

python/packages/main/agent_framework/exceptions.py
```diff
@@ -1,10 +1,24 @@
...
-    pass
+    Automatically logs the message as debug.
+    """
+
+    def __init__(self, message: str, inner_exception: Exception | None = None, *args: Any, **kwargs: Any):
+        """Create an AgentFrameworkException.
+
+        This emits a debug log, with the inner_exception if provided.
+        """
+        logger.debug(message, exc_info=inner_exception)
```
dmytrostruk, line 20: Why `debug` log but not `error` log?

python/packages/azure/agent_framework_azure/_entra_id_authentication.py
```diff
@@ -1,21 +1,27 @@
...
+
+if TYPE_CHECKING:
+    from azure.identity import ChainedTokenCredential
+    from azure.identity.aio import ChainedTokenCredential as AsyncChainedTokenCredential
 
 logger: logging.Logger = logging.getLogger(__name__)
 
 
-def get_entra_auth_token(token_endpoint: str) -> str | None:
+def get_entra_auth_token(credential: "ChainedTokenCredential", token_endpoint: str) -> str | None:
```
peterychang: If we're passing in a credential, we should take in optional params for the other `get_token` params

python/packages/azure/agent_framework_azure/_shared.py
```diff
@@ -149,10 +155,15 @@ def get_azure_openai_auth_token(self, token_endpoint: str | None = None) -> str
...
+
+        endpoint_to_use = token_endpoint or self.token_endpoint or self.default_token_endpoint
         if endpoint_to_use is None:  # type: ignore
```
peterychang: This case is impossible now (since `self.default_token_endpoint` is always a type str)

python/packages/azure/agent_framework_azure/_shared.py
eavanvalkenburg: yeah, I was thinking about that as well, removed it completely now, only in samples do we use DefaultAzureCred

python/packages/foundry/agent_framework_foundry/_chat_client.py
```diff
@@ -605,3 +612,15 @@ def _convert_function_results_to_tool_output(
                 tool_outputs.append(ToolOutput(tool_call_id=call_id, output=str(function_result_content.result)))
 
         return run_id, tool_outputs
+
+    @override
+    def update_agent_name(self, agent_name: str | None) -> None:
```
peterychang: seems a little unintuitive that the update will only happen if an agent name isn't already set

python/packages/main/agent_framework/_clients.py
peterychang: Agreed. Lets try to keep the concept of agents out of the clients

eavanvalkenburg: you're right, removed it from `ChatClientBase`, but can be implemented by clients and then it will be called, the reason we do need a function, is that not every client might use the same name for the `agent_name` concept, like assistants.

python/packages/main/agent_framework/exceptions.py
eavanvalkenburg: becomes very chatty, and often a uncaught error is logged anyway.

python/packages/main/agent_framework/_tools.py
```diff
@@ -22,6 +23,45 @@
...
+        inputs = [inputs]
+    for input_item in inputs:
+        if isinstance(input_item, str):
+            # If it's a string, we assume it's a URI or similar identifier.
+            # Convert it to a UriContent or similar type as needed.
+            parsed_inputs.append(UriContent(uri=input_item, media_type="text/plain"))
+        elif isinstance(input_item, dict):
+            # If it's a dict, we assume it contains properties for a specific content type.
+            # we check if the required keys are present to determine the type.
+            if "uri" in input_item:
```
peterychang, line 49: Document these keys and where they're from if they're not arbitrary

python/packages/foundry/agent_framework_foundry/_chat_client.py
eavanvalkenburg: yeah, but you can make multiple agents with a single client, so I didn't want just the last one to use it to own it, but it is a bit weird yes...

python/packages/main/agent_framework/_tools.py
eavanvalkenburg: I clarified the code comment on where they come from.

python/packages/main/agent_framework/openai/_chat_client.py
```diff
@@ -1,20 +1,24 @@
...
+from openai import AsyncOpenAI, BadRequestError
+from openai.lib._parsing._completions import type_to_response_format_param
 from openai.types import CompletionUsage
 from openai.types.chat.chat_completion import ChatCompletion, Choice
-from openai.types.chat.chat_completion_chunk import ChatCompletionChunk, ChoiceDeltaToolCall
+from openai.types.chat.chat_completion_chunk import ChatCompletionChunk
 from openai.types.chat.chat_completion_chunk import Choice as ChunkChoice
-from openai.types.chat.chat_completion_message_tool_call import ChatCompletionMessageToolCall
-from pydantic import SecretStr, ValidationError
+from openai.types.chat.chat_completion_message_custom_tool_call import ChatCompletionMessageCustomToolCall
```
peterychang, line 15: Is this a new type? The only documentation I can find on it is from langchain?

python/packages/main/agent_framework/openai/_responses_client.py
```diff
@@ -1,267 +1,79 @@
...
-    def _filter_options(self, **kwargs: Any) -> dict[str, Any]:
-        """Filter options for the responses call."""
-        # The responses call does not support all the options that the chat completion call does.
-        # We filter out the unsupported options.
-        return {key: value for key, value in kwargs.items() if value is not None}
-
-    # The responses create call takes very different parameters than the chat completion call,
-    # so we override the get_response method to handle the specific parameters for responses.
-    @override
-    async def get_response(
```
peterychang, line 65: I think its better to keep the customized interface here for clarity's sake. Anybody using the responses API would need to know that the default interface is meaningless and then look up the openai interface to know what parameters are available

python/packages/azure/agent_framework_azure/_assistants_client.py
```diff
@@ -35,6 +35,7 @@ def __init__(
         ad_token: str | None = None,
         ad_token_provider: AsyncAzureADTokenProvider | None = None,
         token_endpoint: str | None = None,
+        ad_credential: "ChainedTokenCredential | None" = None,
```
dmytrostruk, line 38: This parameter is optional, but if users don't provide it or any alternative authentication method, they will receive a runtime error when making a request. I'm wondering if we should make this parameter as required, so if users don't provide it, they will see compile-time error.

eavanvalkenburg: no they can provide API key as well, so there is no clear single thing that needs to be supplied, unless we put in a bunch of overrides, which might be nice...

python/packages/main/agent_framework/openai/_responses_client.py
eavanvalkenburg: readded with some updates

python/packages/main/agent_framework/openai/_chat_client.py
eavanvalkenburg: yeah looks new

PR #390, opened 2025-08-11
.NET: feat: `Microsoft.Extensions.AI.Agents.Hosting.A2A` package

DeagleGross:
## Motivation and Context
Attempt to test agentic-framework integration with the A2A (Agent-To-Agent) protocol and introduce the package which allows easy attachment for the users.

PR uses the official [a2aproject/a2a-dotnet](https://github.com/a2aproject/a2a-dotnet) SDK package for the inner A2A protocol implementation.

Related to #406

## Client usage

The PR has UI changes (selection of protocol is possible), and if `Protocol.A2A` is selected, frontend will be using a separate client to communicate via A2A. Client can resolve the `A2ACardResolver` or `A2AClient` to resolve the agent card (`/v1/card` endpoint) and to send message stream (`message:stream` endpoint) respectively. _This client is not implemented in generic-way and I dont think there is a point in moving it to A2A package, because every user has to implement their own handling._ 

## Backend usage

In order to enable A2A integration on a specific agent, user could use one of these 2 methods:
```csharp
builder.AddAIAgent("pirate", ...);
builder.AddAIAgent("knights-and-knaves", () => ...)

app.AttachA2AMessaging("/a2a/pirate", agentName: "pirate");
app.AttachA2AMessaging("/a2a/knights-and-knaves", agentName: "knights-and-knaves");
```
`AttachA2AMessaging` is the simplest form of A2A implementation, where only resolving agent card and responding with trivial `Message` structure is possible. Input requires the name of agent and the group-root endpoint for A2A communication for this agent.

I tried to implement it in such a way, that no other changes would be required to turn on A2A communication for agent, resulting in allowing both agentic-framework communication and a2a communication possible at the same time.

Registrations are automatically adding the new API surface to the AspNetCore app:
image

## Implementation

The implementation is different for A2A `Message` or A2A `AgentTask` communication. The first one is trivial and allows agent to respond with the `Message` only. The main class of A2A SDK is `TaskManager` which provides the API to 1) subscribe to the processing of different events; 2) operate with the `AgentTask`.

### A2A `Message` workflow support
For the message-only workflow, the implementation is basically parsing the incoming agentic-framework `MessageSendParams`, converting to A2A input (`IReadOnlyCollection<ChatMessage>?`) and then running the "innerAgent" `RunAsync` (where innerAgent is just the agent registered by the user). On the return-flow mapping of A2A output happens back to the agentic-framework output and that's it.

### A2A `AgentTask` workflow support
not supported for now and will be added in future - it requires some redesign of the agentic-framework `AIAgent` capabilities and can change in the future. **History of this PR shows implementation of AgentTask based message processing, but due to entropy it is dropped.**

## Demo
Both agents (one is chat client, and second is an orchestration over multiple chat clients) work.
https://github.com/user-attachments/assets/9e0c82f4-c1f0-4a3f-b0ed-597f0fc84dc1

## Contribution Checklist


dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AgentRunOptions.cs
```diff
@@ -24,4 +25,10 @@ public AgentRunOptions(AgentRunOptions options)
     {
         Throw.IfNull(options);
     }
+
+    /// <summary>
+    /// Gets a function that creates a new <see cref="AgentThread"/> instance.
+    /// Can be used if a specific AgentThread implementation is needed for the same agent run.
+    /// </summary>
+    public virtual Func<AgentThread>? GetAgentThread { get; set; }
```
stephentoub: cc: @westey-m

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Internal/A2AAgent.cs
```diff
@@ -0,0 +1,115 @@
...
+using System.Threading.Tasks;
+using A2A;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Extensions.AI.Agents.A2A.Internal;
+
+/// <summary>
+/// A2A agent that wraps an existing AIAgent and provides A2A-specific thread wrapping.
+/// </summary>
+internal sealed class A2AAgent : AIAgent
```
stephentoub: Does this supply the inverse somewhere? The ability to use an A2A agent via the AIAgent abstraction?

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Internal/A2AAgentRunOptions.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// </summary>
+    public string? TaskId { get; init; }
+
+    private A2AAgentRunOptions(string? taskId)
+    {
+        this.TaskId = taskId;
+    }
+
+    public static A2AAgentRunOptions CreateA2AMessagingOptions()
+        => new(taskId: null!);
```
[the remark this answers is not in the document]
DeagleGross: oki!

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/IA2AAgentTaskProcessor.cs
```diff
@@ -0,0 +1,39 @@
...
+    /// <param name="token"></param>
+    /// <returns></returns>
+    Task CancelTaskAsync(AgentTask task, CancellationToken token);
+
+    /// <summary>
+    /// todo
+    /// </summary>
+    /// <param name="task"></param>
+    /// <param name="token"></param>
+    /// <returns></returns>
```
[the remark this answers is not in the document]
DeagleGross: added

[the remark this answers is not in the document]
DeagleGross: added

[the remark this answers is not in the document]
DeagleGross: added

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Internal/A2AAgent.cs
```diff
@@ -0,0 +1,115 @@
...
+        if (options.TaskId is null)
+        {
+            throw new ArgumentException("TaskId must be provided in A2AAgentRunOptions.", nameof(options));
+        }
+
+        thread ??= this._innerAgent.GetNewThread();
+        thread = new A2AAgentThreadWrapper(thread, this._taskManager, options.TaskId);
+        options.GetAgentThread = () => thread;
+
+        // TODO: workaround. Some agent implementations expect a MessageStore to be set.
```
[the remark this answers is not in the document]
DeagleGross: i think my comment is fine?

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Converters/MessageConverter.cs
```diff
@@ -0,0 +1,233 @@
...
+#pragma warning restore CA1859 // Use concrete types when possible for improved performance
+    {
+        var result = part switch
+        {
+            TextPart textPart => new TextContent(textPart.Text)
+            {
+                RawRepresentation = textPart,
+                AdditionalProperties = textPart.Metadata?.ToAdditionalPropertiesDictionary()
+            },
+            FilePart or DataPart or _ => throw new NotSupportedException("Parts are not supported.")
```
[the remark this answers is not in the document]
DeagleGross: makes sense

dotnet/Directory.Packages.props
```diff
@@ -43,7 +43,7 @@
     <PackageVersion Include="OpenTelemetry.Extensions.Hosting" Version="1.12.0" />
     <!-- Microsoft.Extensions.* -->
     <PackageVersion Include="Microsoft.Bcl.HashCode" Version="6.0.0" />
-    <PackageVersion Include="Microsoft.Bcl.AsyncInterfaces" Version="9.0.7" />
+    <PackageVersion Include="Microsoft.Bcl.AsyncInterfaces" Version="10.0.0-preview.5.25277.114" />
```
crickman: Is switching to a preview version really required here?

dotnet/Directory.Packages.props
```diff
@@ -31,7 +31,7 @@
     <PackageVersion Include="Swashbuckle.AspNetCore.SwaggerUI" Version="9.0.3" />
     <!-- System.* -->
     <PackageVersion Include="System.Linq.Async" Version="6.0.3" />
-    <PackageVersion Include="System.Net.ServerSentEvents" Version="9.0.7" />
+    <PackageVersion Include="System.Net.ServerSentEvents" Version="10.0.0-preview.5.25277.114" />
```
crickman: Is switching to a preview version required here?

DeagleGross: Seems like because those are direct dependencies of A2A project:
https://www.nuget.org/packages/A2A/
image

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Internal/A2AAgent.cs
DeagleGross: overthought myself here: made it a wrapper and removed inheritance.

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Microsoft.Extensions.AI.Agents.A2A.csproj
```diff
@@ -0,0 +1,26 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>$(ProjectsCoreTargetFrameworks)</TargetFrameworks>
```
SergeyMenshykh: Should this point to ProjectsTargetFrameworks?

dotnet/Directory.Build.props
```diff
@@ -11,6 +11,7 @@
     <ImplicitUsings>disable</ImplicitUsings>
     <NoWarn>$(NoWarn);IDE0290;IDE0079;NU5128</NoWarn>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <ProjectsCoreTargetFrameworks>net9.0</ProjectsCoreTargetFrameworks>
```
SergeyMenshykh, line 14: Can't we use the `ProjectsDebugTargetFrameworks` element used for the same purpose?
CC: @westey-m

dotnet/Directory.Packages.props
SergeyMenshykh: The version of the package can be overridden in the project where it's actually needed - `dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Microsoft.Extensions.AI.Agents.A2A.csproj`, leaving the other project using the released version of the package.

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Internal/A2AAgent.cs
```diff
@@ -0,0 +1,108 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Extensions.AI.Agents.A2A.Internal;
+
+/// <summary>
+/// A2A agent that wraps an existing AIAgent and provides A2A-specific thread wrapping.
+/// </summary>
+internal sealed class A2AAgent
+{
+    // Does this supply the inverse somewhere? The ability to use an A2A agent via the AIAgent abstraction?
```
SergeyMenshykh: Please remove this comment. GitHub history keeps all changes and comments.

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Internal/A2AAgent.cs
```diff
@@ -0,0 +1,108 @@
...
+
+    private readonly ILogger _logger;
+    private readonly TaskManager _taskManager;
+
+    public AIAgent InnerAgent { get; }
+    public string? Name => this.InnerAgent.Name;
+
+    public A2AAgent(ILogger logger, AIAgent innerAgent, TaskManager taskManager)
```
SergeyMenshykh: Please move the logger to the end of the parameter list, make it optional, and change its type to ILoggerFactory to align with other agents.

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Internal/A2AAgent.cs
```diff
@@ -0,0 +1,108 @@
...
+/// </summary>
+internal sealed class A2AAgent
+{
+    // Does this supply the inverse somewhere? The ability to use an A2A agent via the AIAgent abstraction?
+
+    private readonly ILogger _logger;
+    private readonly TaskManager _taskManager;
+
+    public AIAgent InnerAgent { get; }
+    public string? Name => this.InnerAgent.Name;
```
SergeyMenshykh: The inner agent property is used to get the agent id and description by the A2AAgentCardProvider. At the same time, the provider accesses the name via the Name property. Please make it consistent.

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/A2AHostingApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,95 @@
...
+
+/// <summary>
+/// Provides extension methods for configuring A2A (Agent-to-Agent) communication in a host application builder.
+/// </summary>
+public static class A2AHostingApplicationBuilderExtensions
+{
+    /// <summary>
+    /// Attaches A2A communication via AgentTasks
+    /// </summary>
+    /// <param name="app"></param>
```
SergeyMenshykh: XML comments are empty. Same in a few other places.

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/A2AHostingApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,95 @@
...
+using Microsoft.Extensions.AI.Agents.A2A.Internal.Connectors;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Extensions.AI.Agents.A2A;
+
+/// <summary>
+/// Provides extension methods for configuring A2A (Agent-to-Agent) communication in a host application builder.
+/// </summary>
+public static class A2AHostingApplicationBuilderExtensions
```
SergeyMenshykh: We usually name the extension classes after the entities they provide extensions for. In this case, the name could be `WebApplicationExtensions`.

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/IA2AAgentTaskProcessor.cs
```diff
@@ -0,0 +1,39 @@
...
+using A2A;
+
+namespace Microsoft.Extensions.AI.Agents.A2A;
+
+/// <summary>
+/// Is a more complicated form of A2A communication where the agent can process AgentTasks.
+/// AgentTask is a representation of stateful potentially long-running conversation between user (client or another agent) and this agent.
+/// See details at <see href="https://github.com/a2aproject/A2A/blob/main/docs/topics/life-of-a-task.md">A2A docs</see>
+/// </summary>
+public interface IA2AAgentTaskProcessor : IA2AAgentCardProvider
```
SergeyMenshykh: Is the inheritance necessary here? All implementations of the `IA2AAgentTaskProcessor` interface will be broken once the `IA2AAgentCardProvider` interface changes. The same applies to IA2AMessageProcessor.

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/IA2AAgentTaskProcessor.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+    /// <summary>
+    /// Creates a new agent task, representing a stateful, potentially long-running conversation between a user and the agent.
+    /// </summary>
+    /// <param name="task">The <see cref="AgentTask"/> to be created.</param>
+    /// <param name="token">A <see cref="CancellationToken"/> to observe while waiting for the task to complete.</param>
+    /// <returns>A <see cref="Task"/> that represents the asynchronous operation.</returns>
+    Task CreateTaskAsync(AgentTask task, CancellationToken token);
```
SergeyMenshykh: All public APIs in AF name the cancellation token parameter as `cancellationToken` and provide a default value, allowing callers to skip it if they want: `CancellationToken cancellationToken = default`.

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Internal/A2AAgent.cs
```diff
@@ -0,0 +1,108 @@
...
+using System.Threading.Tasks;
+using A2A;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Extensions.AI.Agents.A2A.Internal;
+
+/// <summary>
+/// A2A agent that wraps an existing AIAgent and provides A2A-specific thread wrapping.
+/// </summary>
+internal sealed class A2AAgent
```
SergeyMenshykh: This agent might need to be renamed so that the `A2AAgent` name can be used for the other agent, which will adapt an A2A agent to an `AIAgent`. See SK's method of naming the A2A agents here: https://github.com/microsoft/semantic-kernel/tree/main/dotnet/src/Agents/A2A

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Internal/A2AAgent.cs
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    // Delegate the core methods to the inner agent
+    public Task<AgentRunResponse> RunAsync(
+        A2AAgentRunOptions options,
+        IReadOnlyCollection<ChatMessage>? messages = null,
+        AgentThread? thread = null,
+        CancellationToken cancellationToken = default)
+    {
+        if (options is not A2AAgentRunOptions a2aRunOptions)
```
SergeyMenshykh: Is this check necesssary?

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Internal/A2AAgent.cs
```diff
@@ -0,0 +1,108 @@
...
+    public A2AAgent(ILogger logger, AIAgent innerAgent, TaskManager taskManager)
+    {
+        this._logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        this._taskManager = taskManager ?? throw new ArgumentNullException(nameof(taskManager));
+
+        this.InnerAgent = innerAgent ?? throw new ArgumentNullException(nameof(innerAgent));
+    }
+
+    // Delegate the core methods to the inner agent
+    public Task<AgentRunResponse> RunAsync(
```
SergeyMenshykh: It would be good, from a readability point of view, to keep consistency with the `AIAgent` method signatures and order the parameters as: messages, thread, options, cancellationToken if possible.

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/A2AHostingApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,95 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using A2A;
+using A2A.AspNetCore;
```
SergeyMenshykh: I'm not sure we need to pull the ASP.NET Core packages into this A2A package. We might need an A2A.AspNetCore package for the web app extensions that can be used by consumers who are exposing AI agents as A2A ones.

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Internal/Connectors/A2AAgentTaskProcessor.cs
```diff
@@ -0,0 +1,63 @@
...
+    public Task UpdateTaskAsync(AgentTask task, CancellationToken token)
+    {
+        var final = task.Status.IsCompleted();
+        return this._taskManager.UpdateStatusAsync(task.Id, task.Status.State, final: final, cancellationToken: token);
+    }
+
+    public Task CancelTaskAsync(AgentTask task, CancellationToken token)
+    {
+        // cancellation is just updating a status, so that all subscribers are notified.
+        return this._taskManager.UpdateStatusAsync(task.Id, task.Status.State, final: true, cancellationToken: token);
```
SergeyMenshykh: This is already done by the task manager - https://github.com/a2aproject/a2a-dotnet/blob/36cde6443b78c718db58debe7fc0b2645f949f34/src/A2A/Server/TaskManager.cs#L112

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Internal/Connectors/A2AAgentTaskProcessor.cs
```diff
@@ -0,0 +1,63 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using A2A;
+using Microsoft.Extensions.AI.Agents.A2A.Converters;
+using Microsoft.Extensions.AI.Agents.A2A.Extensions;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Extensions.AI.Agents.A2A.Internal.Connectors;
+
+internal sealed class A2AAgentTaskProcessor : A2AAgentCardProvider, IA2AAgentTaskProcessor
```
SergeyMenshykh: What is the division of responsibilities between the processor and the A2A agent? 

To keep it simple, can we have the attach method on A2AAgent, similar to how the SK agent has it - https://github.com/microsoft/semantic-kernel/blob/6081741cc158e15bb73e63c18d59a904d6fb74f6/dotnet/src/Agents/A2A/A2AHostAgent.cs#L44 - to plug an AI agent into the A2A infrastructure? That method may have options allowing customization of handling of different events of the task manager.

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Internal/A2AAgent.cs
DeagleGross: my bad; removed

dotnet/Directory.Packages.props
DeagleGross: answered in this thread https://github.com/microsoft/agent-framework/pull/390#discussion_r2267070350

dotnet/Directory.Packages.props
DeagleGross: Sorry, I did not overcome this error before, but now I've overriden the version of dependencies in A2A package, and other projects will be built with stable 9.0.7 version of `System.Net.ServerSentEvents` and `Microsoft.Bcl.AsyncInterfaces`.

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Microsoft.Extensions.AI.Agents.A2A.csproj
DeagleGross: [A2A.AspNetCore](https://www.nuget.org/packages/A2A.AspNetCore/0.1.0-preview.2) NuGet package only targets net8.0 and net9.0, so its not possible to reuse `ProjectsTargetFrameworks` which define `netstandard2.0` and `net472` as well.

dotnet/Directory.Build.props
DeagleGross: [A2A.AspNetCore](https://www.nuget.org/packages/A2A.AspNetCore/0.1.0-preview.2) NuGet package only targets net8.0 and net9.0, so its not possible to reuse `ProjectsTargetFrameworks` which define `netstandard2.0` and `net472` as well.

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Internal/A2AAgent.cs
DeagleGross: makes sense, done!

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Internal/A2AAgent.cs
DeagleGross: removed property `Name` and referring to `_agent.InnerAgent` everywhere now so its consistent.

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/A2AHostingApplicationBuilderExtensions.cs
DeagleGross: added in this file. I will check others as well

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/A2AHostingApplicationBuilderExtensions.cs
DeagleGross: renamed

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/IA2AAgentTaskProcessor.cs
DeagleGross: removed `IA2AAgentCardProvider` completely to avoid such a problem. Now `Task<AgentCard> GetAgentCardAsync(string agentPath, CancellationToken cancellationToken = default)` lives in both `IA2AAgentTaskProcessor` and `IA2AMessageProcessor` independently

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/IA2AAgentTaskProcessor.cs
DeagleGross: added in all new interfaces i've introduced

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Internal/A2AAgent.cs
DeagleGross: a leftover from previous changes; removed

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Internal/A2AAgent.cs
DeagleGross: changed the public method to follow this convention

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Internal/Connectors/A2AAgentTaskProcessor.cs
DeagleGross: hmm, the usage is indeed different for cancel against the create\update. I've changed the `A2AAgentTaskProcessor` to call `taskManager.CancelTaskAsync` but handle the error and skip it if task is not found or already cancelled.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AgentRunOptions.cs
DeagleGross: removed this API

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/A2AHostingApplicationBuilderExtensions.cs
DeagleGross: I've introduced 2 packages now: one is `AI.Agents.A2A` and another is `AI.Agents.A2A.AspNetCore`. They are serving different purposes and there is a clear package dependency chain now

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Internal/Connectors/A2AAgentTaskProcessor.cs
DeagleGross: I've refactored as we discussed offline - there are no more processors at all

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Internal/A2AAgentWrapper.cs
```diff
@@ -0,0 +1,98 @@
...
+        this._logger = (loggerFactory ?? NullLoggerFactory.Instance).CreateLogger<A2AAgentWrapper>();
+
+        this._actorClient = actorClient;
+        this._taskManager = taskManager ?? throw new ArgumentNullException(nameof(taskManager));
+        this._innerAgent = innerAgent ?? throw new ArgumentNullException(nameof(innerAgent));
+    }
+
+    public async Task<Message> ProcessMessageAsync(MessageSendParams messageSendParams, CancellationToken cancellationToken)
+    {
+        var contextId = messageSendParams.Message.ContextId!;
```
SergeyMenshykh: The `contextId` property of the A2A message is optional, so we can't assume it will always be initialized - https://github.com/a2aproject/A2A/blob/00cf76e7bbc752842ef254f3d4136ed1b5751f6e/specification/json/a2a.json#L1661

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Internal/A2AAgentWrapper.cs
```diff
@@ -0,0 +1,98 @@
...
+
+namespace Microsoft.Extensions.AI.Agents.A2A.Internal;
+
+/// <summary>
+/// A2A agent that wraps an existing AIAgent and provides A2A-specific thread wrapping.
+/// </summary>
+internal sealed class A2AAgentWrapper
+{
+    private readonly ILogger _logger;
+    private readonly TaskManager _taskManager;
```
SergeyMenshykh: The field is not used.

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Internal/A2AAgentWrapper.cs
```diff
@@ -0,0 +1,98 @@
...
+        {
+            return Task.FromCanceled<AgentCard>(cancellationToken);
+        }
+
+        return Task.FromResult(new AgentCard()
+        {
+            Name = this._innerAgent.Name ?? string.Empty,
+            Description = this._innerAgent.Description ?? string.Empty,
+            Url = agentPath,
+            Version = this._innerAgent.Id,
```
SergeyMenshykh: That's not correct mapping.

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/AIAgentExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// <param name="actorClient">The actor client implementation to use.</param>
+    /// <param name="taskManager">Instance of <see cref="TaskManager"/> to configure for A2A messaging.</param>
+    /// <param name="loggerFactory">The logger factory to use for creating <see cref="ILogger"/> instances.</param>
+    public static void AttachA2AMessaging(
+        this AIAgent agent,
+        IActorClient actorClient,
+        TaskManager taskManager,
+        ILoggerFactory? loggerFactory = null)
+    {
+        taskManager ??= new();
```
SergeyMenshykh: This line assumes the nullability of the `taskManager` parameter, however, the method signature requires it. Should the parameter be nullable?

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Internal/A2AAgentWrapper.cs
```diff
@@ -0,0 +1,98 @@
...
+        };
+        var @params = JsonSerializer.SerializeToElement(runRequest, AgentHostingJsonUtilities.DefaultOptions.GetTypeInfo(typeof(AgentRunRequest)));
+
+        var requestHandle = await this._actorClient.SendRequestAsync(new ActorRequest(actorId, messageId, method: "Run" /* ?refer to const here? */, @params: @params), cancellationToken).ConfigureAwait(false);
+        var response = await requestHandle.GetResponseAsync(cancellationToken).ConfigureAwait(false);
+
+        return response.ToMessage();
+    }
+
+    public Task<AgentCard> GetAgentCardAsync(string agentPath, CancellationToken cancellationToken = default)
```
SergeyMenshykh: Currently, AI agents don't have all the required information to fully populate agent cards, such as version, input and output modalities, and skills. Until it's possible to obtain that data from the AI agent interface, I think this information should be provided by developers at the agent configuration point - `app.Attach*` methods. Additionally, we might consider, though not necessarily in this PR, creating some utilities that would simplify card creation from existing AI agents, such as `AIAgentExtension.CreateA2AAgentCard(version, modalities, capabilities, etc.)`.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/Microsoft.Extensions.AI.Agents.Abstractions.csproj
```diff
@@ -31,4 +31,7 @@
     <InternalsVisibleTo Include="Microsoft.Extensions.AI.Agents.Abstractions.UnitTests" />
   </ItemGroup>
 
+  <ItemGroup>
+    <InternalsVisibleTo Include="Microsoft.Extensions.AI.Agents.A2A" />
```
SergeyMenshykh: Is it necessary?

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/AIAgentExtensions.cs
```diff
@@ -0,0 +1,34 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using A2A;
+using Microsoft.Extensions.AI.Agents.A2A.Internal;
+using Microsoft.Extensions.AI.Agents.Runtime;
```
SergeyMenshykh: I don't think we need to pull runtime into the A2A package. The idea, as far as I understand, is that neither of the agents knows about runtime, and runtime does not know about agents. It is up to the orchestration to use them together. This allows AI agents to be used independently outside the orchestration if needed.

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Internal/A2AAgentWrapper.cs
DeagleGross: ah, okay, generating a new one on the fly then.

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Internal/A2AAgentWrapper.cs
DeagleGross: removed

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/AIAgentExtensions.cs
DeagleGross: it does not make sense to make parameter nullable, because taskManager is needed to be used later by the caller. I removed this lien with allocating a new instance of taskManager if passed null

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/AIAgentExtensions.cs
DeagleGross: it is essential for using `IActorClient` and its required to use the same API as current communication mechanisms are used in agentFramework

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Internal/A2AAgentWrapper.cs
DeagleGross: responded in other comment about GetAgentCardAsync: https://github.com/microsoft/agent-framework/pull/390#discussion_r2278688920

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Internal/A2AAgentWrapper.cs
DeagleGross: I agree. Lets introduce and expose API in other PR. I think it will be good if we can provide the default implementation for now, and users can override it if they want to even today. But registered endpoint will not throw, but rather respond with at least some of the fields filled in.

What do you think?

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/Microsoft.Extensions.AI.Agents.Abstractions.csproj
DeagleGross: removed

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Internal/A2AAgentWrapper.cs
DeagleGross: as discussed offline: reworked to give user opportunity to fill in the agent card query.

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/AIAgentExtensions.cs
SergeyMenshykh: These "Attach*" extension methods should not be coupled with the runtime to allow no-runtime scenarios. In case the agent runtime needs to be used to talk to an AIAgent exposed as an A2A agent, it can be easily done using an agent proxy in the hosting app:
```csharp
// Expose the "pirate" agent via A2A messaging
app.AttachA2AMessaging(
    agent: app.Services.GetRequiredKeyedService<AIAgent>("pirate"),
    path: "/a2a/pirate"
);

// Expose the "pirate" agent via A2A messaging and communicate to the agent via agent runtime
app.AttachA2AMessaging(
    agent: new AgentProxy(
        name: "pirate",
        client: app.Services.GetRequiredService<IActorClient>()
    ),
    path: "/a2a/pirate"
);
```

DeagleGross: Today `AIAgent` does not allow handling multiple conversations simultaneosly and that's how the `IActorClient` was introduced with the concept of `ActorId` encapsulating `actorType` (the agent type) and `actorKey` (the context id or conversation id).

If you follow on how sample app handles the communication today, it is done via `AgentProxy` and `IActorClient` - see this PR for intro of those types: https://github.com/microsoft/agent-framework/pull/296.

I dont understand the problem of coupling the "runtime" (even though I am not including a runtime directly, but only the abstractions as `IActorClient` which seem to exist under `.Runtime` namespace, but not actually using any specific runtime without a replacement mechanism possible).

Do you mind elaborating here? Sorry, maybe I am missing some basics :)

DeagleGross: I also looked into the implementation of `AttachXXX/A2AAgentWrapper` again, and I dont see what kind of implementation I can propose for the method without `IActorClient`. We can only do `Agent.RunAsync()` per single message and that seems to be off the intended flow per A2A.

SergeyMenshykh: The `A2AAgentWrapper.ProcessMessageAsync` method will call the `AIAgent.Run` method for each received message, possibly performing the A2A->AF->A2A message translation, similar to how the [SK A2AHostAgent](https://github.com/microsoft/semantic-kernel/blob/main/dotnet/src/Agents/A2A/A2AHostAgent.cs) does it. This will decouple `A2AAgentWrapper` from runtime primitives, so there won't be a need to pull unnecessary runtime abstraction packages for consumers that don't need runtime and want to expose an `AIAgent` as an A2A agent.
```csharp
app.AttachA2AMessaging(
    agent: app.Services.GetRequiredKeyedService<AIAgent>("pirate"),
    path: "/a2a/pirate"
);
```

For scenarios where the runtime agent needs to be exposed as an A2A agent, consumers can use `AgentProxy` to translate all calls to `AgentProxy.RunAsync` into runtime messages.
```csharp
app.AttachA2AMessaging(
    agent: new AgentProxy(
        name: "pirate",
        client: app.Services.GetRequiredService<IActorClient>()
    ),
    path: "/a2a/pirate"
);
```

> We can only do Agent.RunAsync() per single message and that seems to be off the intended flow per A2A.

I would like to understand this better. Does it mean that RunAsync cannot handle more than one message at once? and what exactly is off here?

dotnet/samples/AgentWebChat/AgentWebChat.Web/A2AActorClient.cs
```diff
@@ -0,0 +1,127 @@
...
+using System.Text.Json;
+using A2A;
+using Microsoft.Extensions.AI;
+using Microsoft.Extensions.AI.Agents.Hosting;
+using Microsoft.Extensions.AI.Agents.Hosting.A2A.Converters;
+using Microsoft.Extensions.AI.Agents.Runtime;
+
+namespace AgentWebChat.Web;
+
+public class A2AActorClient : IActorClient
```
SergeyMenshykh: internal?

dotnet/samples/AgentWebChat/AgentWebChat.AgentHost/Program.cs
```diff
@@ -92,6 +93,16 @@ Once the user has deduced what type (knight or knave) both Alice and Bob are, te
...
 app.MapActors();
 
+// attach a2a with simple message communication
+app.AttachA2AMessaging(agentName: "pirate", path: "/a2a/pirate");
+app.AttachA2AMessaging(agentName: "knights-and-knaves", path: "/a2a/knights-and-knaves", agentCard: new()
+{
+    Name = "Knights and Knaves",
+    Description = "An agent that helps you solve the knights and knaves puzzle.",
+    Version = "1.0",
+    Url = "http://localhost:5390/a2a/knights-and-knaves"
```
SergeyMenshykh: Is setting up an explicit url necessary? A2A SDK assign it as well:
image

dotnet/samples/AgentWebChat/AgentWebChat.AgentHost/Program.cs
```diff
@@ -92,6 +93,16 @@ Once the user has deduced what type (knight or knave) both Alice and Bob are, te
 
 app.MapActors();
 
+// attach a2a with simple message communication
+app.AttachA2AMessaging(agentName: "pirate", path: "/a2a/pirate");
+app.AttachA2AMessaging(agentName: "knights-and-knaves", path: "/a2a/knights-and-knaves", agentCard: new()
```
SergeyMenshykh: The A2A spec specifies a few ways to discover agents, as shown here: https://a2a-protocol.org/dev/topics/agent-discovery/. We need to make sure our solution is flexible enough to support all of them in an incremental way, without causing breaking changes.  
  
Right now, the AttachA2AMessaging implementation uses the "Well-Known URI" option, which assumes there's only one A2A agent per host that can be discovered via the `.well-known/agent.json` path. Can you check with the A2A explorer if the pirate agent is discoverable? I think only "knights-and-knave" is discoverable because calling AttachA2AMessaging again will overwrite the ".well-known/agent.json" path for the pirate agent.

dotnet/tests/CosmosDB.IntegrationTests/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB.Tests.AppHost/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB.Tests.AppHost.csproj
```diff
@@ -10,6 +10,7 @@
     <UserSecretsId>80048040-aaf1-4f44-9970-8aef39651edf</UserSecretsId>
     <IsAspireHost>true</IsAspireHost>
     <GenerateProgramFile>false</GenerateProgramFile>
+    <NoWarn>$(NoWarn);NU1008</NoWarn>
```
SergeyMenshykh: This change does not seem related to this PR.

dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatClientAgent.cs
```diff
@@ -333,7 +333,6 @@ public override AgentThread GetNewThread()
         CancellationToken cancellationToken)
     {
         ChatOptions? chatOptions = this.CreateConfiguredChatOptions(runOptions);
-
```
SergeyMenshykh: This change does not seem related to this PR.

dotnet/samples/AgentWebChat/AgentWebChat.Web/A2AActorClient.cs
DeagleGross: exists only in sample app; but sure, made it internal.

dotnet/samples/AgentWebChat/AgentWebChat.AgentHost/Program.cs
DeagleGross: good point - I've changed how we return the `AgentCard` and if user did not set `Url` explicitly, we will assign it for them.

dotnet/tests/CosmosDB.IntegrationTests/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB.Tests.AppHost/Microsoft.Extensions.AI.Agents.Runtime.Storage.CosmosDB.Tests.AppHost.csproj
DeagleGross: forgot to delete this csproj completely - i've recretaed it under different name (see `CosmosDB.Testing.AppHost`)

dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatClientAgent.cs
DeagleGross: reverted

dotnet/samples/AgentWebChat/AgentWebChat.AgentHost/Program.cs
DeagleGross: It fails today:
```
Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware[1]
      An unhandled exception has occurred while executing the request.
      Microsoft.AspNetCore.Routing.Matching.AmbiguousMatchException: The request matched multiple endpoints. Matches: 
      
      HTTP: GET /.well-known/agent.json
      HTTP: GET /.well-known/agent.json
         at Microsoft.AspNetCore.Routing.Matching.DefaultEndpointSelector.ReportAmbiguity(Span`1 candidateState)
```

I will double-check tomorrow

DeagleGross: Discussed offline: filed an #476 and this will be fixed in future version of A2A SDK + incorporated here.
Also renamed to `AttachA2A` extension methods

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/AIAgentExtensions.cs
DeagleGross: discussed offline with @SergeyMenshykh  and @westey-m. We decided to rename package to `AI.Agents.Hosting.A2A` to clearly state that runtime reference is used here. `AI.Agents.A2A` package will be probably created later for purposes of client/direct `AIAgent` implementation and will not reference runtime in any kind.

dotnet/Directory.Build.props
SergeyMenshykh: CC: @westey-m

dotnet/samples/AgentWebChat/AgentWebChat.Web/A2AActorClient.cs
```diff
@@ -0,0 +1,127 @@
...
+        {
+            throw new NotImplementedException();
+        }
+
+        public override async IAsyncEnumerable<ActorRequestUpdate> WatchUpdatesAsync([EnumeratorCancellation] CancellationToken cancellationToken)
+        {
+            var agentRunRequestData = this._request.Params.Deserialize(AgentHostingJsonUtilities.DefaultOptions.GetTypeInfo(typeof(AgentRunRequest))) as AgentRunRequest;
+            var messageTexts = agentRunRequestData!.Messages!.SelectMany(x => x.Contents.OfType<TextContent>()).Select(x => x.Text);
+            var parts = messageTexts.Select(text => new TextPart { Text = text });
+            var messageSendParams = new MessageSendParams
```
SergeyMenshykh, line 100: This conversion functionality can be moved to the MessageConverter.

dotnet/src/Microsoft.Extensions.AI.Agents.Hosting.A2A/Converters/MessageConverter.cs
```diff
@@ -0,0 +1,262 @@
...
+            {
+                result.Add(chatMessage);
+            }
+        }
+
+        return result;
+    }
+
+    /// <summary>
+    /// Converts A2A MessageSendParams to a collection of Microsoft.Extensions.AI ChatMessage objects.
```
SergeyMenshykh: The comment says that the method converts MessageSendParams while it accepts list of messages.

dotnet/samples/AgentWebChat/AgentWebChat.Web/A2AActorClient.cs
DeagleGross: there is no such a method in `MessageConverter` and it should not be used for clients here. I expected this class to be helping with conversions only for underlying types on the server side. Also this is a sample, so I think we should just keep it explicit here

dotnet/src/Microsoft.Extensions.AI.Agents.Hosting.A2A/Converters/MessageConverter.cs
DeagleGross: fixed comment

SergeyMenshykh (approved):

dotnet/src/Microsoft.Extensions.AI.Agents.Hosting.A2A/AIAgentExtensions.cs
```diff
@@ -0,0 +1,74 @@
...
+    /// <param name="loggerFactory">The logger factory to use for creating <see cref="ILogger"/> instances.</param>
+    /// <returns>The configured <see cref="TaskManager"/>.</returns>
+    public static TaskManager AttachA2A(
+        this AIAgent agent,
+        IActorClient actorClient,
+        TaskManager? taskManager = null,
+        ILoggerFactory? loggerFactory = null)
+    {
+        ArgumentNullException.ThrowIfNull(agent, nameof(agent));
+        ArgumentNullException.ThrowIfNull(actorClient, nameof(actorClient));
```
westey-m, line 32: We also have the helpers from the Microsoft.Extensions repo, which means this works as well:
```suggestion
        Throw.IfNull(agent);
        Throw.IfNull(actorClient);
```

dotnet/src/Microsoft.Extensions.AI.Agents.Hosting.A2A/Internal/A2AAgentWrapper.cs
```diff
@@ -0,0 +1,70 @@
...
+/// </summary>
+internal sealed class A2AAgentWrapper
+{
+    private readonly ILogger _logger;
+    private readonly AIAgent _innerAgent;
+    private readonly IActorClient _actorClient;
+
+    public A2AAgentWrapper(
+        IActorClient actorClient,
+        AIAgent innerAgent,
```
westey-m, line 26: This wrapper seems more about sending messages to an actor than to an agent.  That actor doesn't necessarily have to be an AIAgent.  The only usage of AIAgent here is to set the ActorType.Name.

Would it make sense to decouple this wrapper from AIAgent, and rather just take the ActorType name as a parameter instead of AIAgent?  This class then becomes an `A2AActorWrapper` instead of an `A2AAgentWrapper`.  If someone happens to register an actor that contains an AIAgent, great, but they can also use this with any Actor.

westey-m: Actually, I see that it is internal right now.  This makes my feedback less important.  It just strikes me like a useful piece of glue to have for folks who want to expose any actor over the A2A protocol.  E.g. if we want to expose agents from other frameworks (taking my SK/AG/AF hat off for a moment 😄), they would just have different actor types.

westey-m (approved):

dotnet/src/Microsoft.Extensions.AI.Agents.Hosting.A2A/Internal/A2AAgentWrapper.cs
DeagleGross: Thanks for input wesley, I will do changes in the follow-up PRs

dotnet/src/Microsoft.Extensions.AI.Agents.Hosting.A2A/AIAgentExtensions.cs
DeagleGross: will do in follow-up PR

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/Internal/A2AAgent.cs
DeagleGross: outdated comment

Issue #357, opened 2025-08-06
Conditional edge group


TaoChenOSU:
1. We need to support conditional branching based on workflow states: https://github.com/microsoft/agent-framework/pull/271#discussion_r2240776688
2. We need to support conditional edge group: https://github.com/microsoft/agent-framework/pull/271#discussion_r2241001060

ekzhu:
This is completed via #393

Issue #371, opened 2025-08-07
Support functions as executors


TaoChenOSU:
We should support functions as executors. Users don't have to create executors instead they can simply pass functions to the workflow builder and use edges to connect these functions.

Issue #319, opened 2025-08-05
Typed workflow


TaoChenOSU:
A workflow should be typed. This will be beneficial to users because they no longer need to know the internals of the workflow (i.e. the start executor) to be able to trigger a workflow.

A workflow should have the same type as the input type of the start executor.

TaoChenOSU:
Currently, if a workflow receives a message of a type that it cannot handle, an exception will be raised.

It's not possible to type the workflow given that the types of messages a workflow can accept are unknown.

PR #393, opened 2025-08-11
Python: Workflow Edge Groups

TaoChenOSU:
### Motivation and Context

Addresses: [#357](https://github.com/microsoft/agent-framework/issues/357)
Related comment: https://github.com/microsoft/agent-framework/pull/271#discussion_r2241001060

### Description

1. Introducing Edge groups
2. Adding the following edge groups:
    - SingleEdgeGroup
    - FanOutEdgeGroup
    - FanInEdgeGroup
    - SwitchCaseEdgeGroup
3. Adding tests and samples

### Contribution Checklist


python/packages/workflow/agent_framework_workflow/_edge.py
```diff
@@ -98,57 +88,353 @@ async def send_message(self, message: Message, shared_state: SharedState, ctx: R
...
+
+class EdgeGroup(ABC):
+    """Represents a group of edges that share some common properties and can be triggered together."""
+
+    def __init__(self) -> None:
+        """Initialize the edge group."""
+        self._id = f"{self.__class__.__name__}/{uuid.uuid4()}"
+
+    @abstractmethod
+    async def send_message(self, message: Message, shared_state: SharedState, ctx: RunnerContext) -> bool:
```
ekzhu, line 101: It's a bit confusing to return a boolean when there can be >2 end states. Perhaps use a Enum? E.g., "SENT", "REJECTED_BY_CONDIITON", "FAILED_TO_SEND"

python/packages/workflow/agent_framework_workflow/_edge.py
```diff
@@ -98,57 +88,353 @@ async def send_message(self, message: Message, shared_state: SharedState, ctx: R
...
+        """
+        self._edge = Edge(source=source, target=target, condition=condition)
+
+    @override
+    async def send_message(self, message: Message, shared_state: SharedState, ctx: RunnerContext) -> bool:
+        """Send a message through the single edge."""
+        if message.target_id and message.target_id != self._edge.target_id:
+            return False
+
+        if self._edge.can_handle(message.data):
```
ekzhu, line 164: Consider raising a runtime error otherwise for sanity check, even though the graph could have been validated. 

What do you think? It is conflict with my other comment about returning an Enum for different cases, but I want us to consider both possibilities.

python/packages/workflow/agent_framework_workflow/_edge.py
```diff
@@ -98,57 +88,353 @@ async def send_message(self, message: Message, shared_state: SharedState, ctx: R
...
+        elif condition_2:
+            edge_2
+        else:
+            edge_4
+    """
+
+    def __init__(
+        self,
+        source: Executor,
+        targets: Sequence[Executor],
```
ekzhu: How about use `*tuple[Executor, Callable[[Any], bool]]`. It is usually easier to collocate rather than correlate for more easy-to-understand code.

lokitoth:
Something to consider here is that keeping the low-level set of edge types the minimal needed to implement all of the conceptual edges we want to support makes the toolkit more portable acros platforms. To that end, `SourceEdgeGroup` and `ConditionalEdgeGroup` become specializations of `PartitionerEdgeGroup` (if you allow the partitioner to be `None`, and interpret that as `SourceEdgeGroup`, and a partition function which always returns only a single index is equivalent to a conditional edge group).

At the same time, having a richer set of types could make tooling easier, as it would allow tools to better understand the intent of an edge-group (conditional vs. partitioner). (Maybe we should support some kind of metadata through which higher-level toolkits could include information for tooling/visualization?)

python/packages/workflow/agent_framework_workflow/_edge.py
```diff
@@ -98,57 +88,353 @@ async def send_message(self, message: Message, shared_state: SharedState, ctx: R
...
+            # If the edge cannot handle the data, return False
+            return False
+
+        if self._is_ready_to_send():
+            # If all edges in the group have data, send the buffered messages to the target executor
+            messages_to_send = [msg for edge in self._edges for msg in self._buffer[edge.source_id]]
+            self._buffer.clear()
+            # Only trigger one edge to send the messages to avoid duplicate sends
+            await self._edges[0].send_message(
+                Message([msg.data for msg in messages_to_send], self._edges[0].source_id),
```
alliscode: Setting the source Id to one of the multiple sources seems a bit weird. I think it's ok for now  but we should think about if there is a better way to represent this.

python/packages/workflow/agent_framework_workflow/_edge.py
```diff
@@ -98,57 +88,353 @@ async def send_message(self, message: Message, shared_state: SharedState, ctx: R
...
+        Args:
+            message (Message): The message to send.
+            shared_state (SharedState): The shared state to use for holding data.
+            ctx (RunnerContext): The context for the runner.
+
+        Returns:
+            bool: True if the message was sent successfully, False otherwise. If a message can be delivered
+                  but rejected due to a condition, it will still return True.
+        """
+        ...
```
ekzhu: Would be good to have public metadata attribute fields on sources and targets. This allows visualization tool to work with public attributes instead of private ones.

TaoChenOSU:
> Something to consider here is that keeping the low-level set of edge types the minimal needed to implement all of the conceptual edges we want to support makes the toolkit more portable acros platforms. To that end, `SourceEdgeGroup` and `ConditionalEdgeGroup` become specializations of `PartitionerEdgeGroup` (if you allow the partitioner to be `None`, and interpret that as `SourceEdgeGroup`, and a partition function which always returns only a single index is equivalent to a conditional edge group).
> 
> At the same time, having a richer set of types could make tooling easier, as it would allow tools to better understand the intent of an edge-group (conditional vs. partitioner). (Maybe we should support some kind of metadata through which higher-level toolkits could include information for tooling/visualization?)

Good point! 

1. I think the `SourceEdgeGroup` is indeed a specialized version of the `PartitioningEdgeGroup` with a None partitioner. Currently, it's the opposite. The `PartitioningEdgeGroup` is a specialized version of the `SourceEdgeGroup`. But I don't think the `ConditionalEdgeGroup` can be a special case of partitioning unless we have a way to restrict the partitioner function to always return a single index, which would require a different signture.
2. Are you proposing some metadata for the type of the group?
For example,
```python
@property
def type(self) -> str:
    return self.__class__.__name__
```

python/packages/workflow/agent_framework_workflow/_edge.py
TaoChenOSU: Good catch!

python/packages/workflow/agent_framework_workflow/_edge.py
TaoChenOSU: The number conditions must be one less than the number of targets because the last target is the default. If we use a tuple to represent a pair, then the last pair must have None as the condition.

It will become:
```python
WorkflowBuilder().add_conditional _edge_group(
    source,
    [
        (target_1, condition_1),
        (target_2, condition_2),
        (target_3, None)
    ]
)
```
We then need to prevent users from putting Nones in all other pairs.

Another option I can think of is to let users provide a default target. For example:
```python
WorkflowBuilder().add_conditional _edge_group(
    source,
    [
        (target_1, condition_1),
        (target_2, condition_2),
    ],
    default=target_3,
)
```

lokitoth:
The reason I think `ConditionalEdgeGroup` is a "special case" is that you can implement it on top of `PartitionerEdgeGroup`. So when someone calls `.add_conditional_fan_out_edges`, rather than create a new type, create a partitioner function that returns the index of the first condition that passes (or the index of the "else" edge, if any).

> Are you proposing some metadata for the type of the group?

No, I do not really have a concrete proposal. Just if we do what I suggest above, we want some way for tooling to be able to go back to the higher-level concept. Like the `[DesignerXYZAttribute]` in .NET. Probably need to be talking to the tooling teams to really understand the requirements here, if any.

python/packages/workflow/agent_framework_workflow/_workflow.py
```diff
@@ -245,8 +266,57 @@ def add_fan_out_edges(self, source: Executor, targets: Sequence[Executor]) -> "S
...
+
+        The output types of the source and the input types of the targets must be compatible.
+        Messages from the source executor will be sent to one of the target executors based on
+        the provided conditions.
+
+        Think of this as a switch statement where each target executor corresponds to a case.
+        Each condition function will be evaluated in order, and the first one that returns True
+        will determine which target executor receives the message.
+
+        The number of targets must be one greater than the number of conditions. The last target
```
lokitoth: [Minor] Consider hinting the restriction explicitly by tupling together the target and the predicate: `cases: Sequence[Tuple[Executor, Callable[[Any], bool]]], default_: Executor | None)` rather than enforce `len(targets) := len(conditions) + 1` at runtime, as it is fairly deep in the documentation for the method otherwise.

python/packages/workflow/agent_framework_workflow/_edge.py
lokitoth: The other thing you can do here, now that I think of it is to model the "switch"ness explicitly:

```python
@dataclass
class Case:
   condition: Callable[[Any],bool]
   target: Executor|str

class Default:
   target: Executor|str
```

Then:

```python
add_conditional_edge_group([
   Case(IsCaseA, ExecutorA),
   Case(IsCaseB, ExecutorB),
   Default(ExecutorC)
])
```

or

```python
add_conditional_edge_group([
   Case(IsCaseA, ExecutorA),
   Case(IsCaseB, ExecutorB)],
   Default(ExecutorC)
)
```

The second one is a bit less satisfying in cutting the sequence, but it would allow us to force only a single `Default`

python/packages/workflow/agent_framework_workflow/_workflow.py
lokitoth: Just noticed there is already conversation here; resolving in favour of the existing discussion

python/packages/workflow/agent_framework_workflow/_edge.py
TaoChenOSU: I think the question is if we should do anything when it's rejected by a condition. We can log a message but I don't think we should surface it.

python/packages/workflow/agent_framework_workflow/_edge.py
TaoChenOSU: We shouldn't raise exceptions in this case, because an executor can emit message of different types and be connected to other executors that handle different types. If we rais exceptions in this case, there will be potentially many exceptions.

What I am doing now is if a message cannot be delivered after try all targets, we will log a warning message. Notice that this excludes cases when a message is rejected by a condition.

python/packages/workflow/agent_framework_workflow/_workflow.py
TaoChenOSU: Yeah, I think this may be a better solution.

python/packages/workflow/agent_framework_workflow/_edge.py
ekzhu: I like this. I think a single list looks a lot better and more "switch-like".

python/packages/workflow/agent_framework_workflow/_edge.py
ekzhu: Ok

python/packages/workflow/agent_framework_workflow/_edge.py
ekzhu: Ok

eavanvalkenburg: yeah, even with the docstring I'm clueless as to what the returned bool represents, why not just raise if something fails and return None otherwise?

python/packages/workflow/agent_framework_workflow/_edge.py
eavanvalkenburg, line 1: I'm mostly wondering why we need all these different types of classes, makes me wonder if we can't just do something like this:
```python

class EdgeGroup:
    def __init__(self, sources: Executor | list[Executor], targets: Executor | list[Executor], conditions, partition_func)
```

python/packages/workflow/agent_framework_workflow/_workflow.py
```diff
@@ -245,8 +266,57 @@ def add_fan_out_edges(self, source: Executor, targets: Sequence[Executor]) -> "S
             source: The source executor of the edges.
             targets: A list of target executors for the edges.
         """
-        for target in targets:
-            self._edges.append(Edge(source, target))
+        self._edge_groups.append(SourceEdgeGroup(source, targets))
+
+        return self
+
+    def add_conditional_fan_out_edges(
```
eavanvalkenburg: can't this also be combined and just trigger based on the presence of conditions or partition_func, and just have a single `add_fan_out_edges` method?

python/packages/workflow/agent_framework_workflow/_edge.py
```diff
@@ -98,57 +88,368 @@ async def send_message(self, message: Message, shared_state: SharedState, ctx: R
...
-                data_list = [msg.data for msg in message_list]
-                source_ids = [msg.source_id for msg in message_list]
-                await self.target.execute(data_list, WorkflowContext(self.target.id, source_ids, shared_state, ctx))
-
-    def _should_route(self, data: Any) -> bool:
-        """Determine if message should be routed through this edge."""
-        if self._condition is None:
+
+
+class EdgeGroup(ABC):
```
eavanvalkenburg: this looks mostly internal, so do we even need a ABC as the basis, or just take the simplest one and have the rest override the appropriate methods (if we can't just collapse them into one in general)

python/samples/getting_started/workflow/step_00_foundation_patterns.py
```diff
@@ -0,0 +1,287 @@
...
+- Single connection with condition
+- Fan-out and fan-in connections
+- Conditional fan-out connections
+- Partitioning fan-out connections
+
+The samples here use numbers and simple arithmetic operations to demonstrate the patterns.
+"""
+
+
+class AddOneExecutor(Executor):
```
eavanvalkenburg, line 21: could we also do this with a decorator on the function?

python/samples/getting_started/workflow/step_00_foundation_patterns.py
```diff
@@ -0,0 +1,287 @@
...
+
+    await workflow.run(1)
+
+
+async def single_edge_with_condition():
+    """A sample to demonstrate a single directed connection with a condition.
+
+    Three executors are connected: AddOneExecutor -> AddOneExecutor, AggregateResultExecutor.
+    The AddOneExecutor will loop back to itself until the number reaches 10, then it will start
+    sending the result to AggregateResultExecutor when the number is greater than 8.
```
eavanvalkenburg: maybe add a note that it will stop when the loop stops, I had to read it twice to understand the reason it ends.

python/samples/getting_started/workflow/step_00_foundation_patterns.py
```diff
@@ -0,0 +1,287 @@
...
+        Adding one to the number: 10 Result: 11
+        Aggregating results: 10
+        Aggregating results: 11
+    """
+    add_one_executor_a = AddOneExecutor()
+    aggregate_result_executor = AggregateResultExecutor()
+
+    workflow = (
+        WorkflowBuilder()
+        .add_edge(add_one_executor_a, add_one_executor_a, condition=lambda x: x < 11)
```
eavanvalkenburg, line 125: could we add this?
```suggestion
        .add_loop(add_one_executor_a, condition=lambda x: x < 11)
```

python/samples/getting_started/workflow/step_00_foundation_patterns.py
```diff
@@ -0,0 +1,287 @@
...
+    workflow = (
+        WorkflowBuilder()
+        .set_start_executor(add_one_executor)
+        .add_conditional_fan_out_edges(
+            add_one_executor,
+            [add_one_executor, divide_by_two_executor, multiply_by_two_executor],
+            # Loop back to the add_one_executor if the number is less than 11
+            # and to the multiply_by_two_executor when the number is larger than or equal to 11 and even.
+            # Otherwise, send to the divide_by_two_executor.
+            conditions=[lambda x: x < 11, lambda x: x % 2 == 0],
```
eavanvalkenburg: this is difficult to read... to me this seems like (and so when x < 11 and x is even, then send it to both add_one and divide_by_two):
if less then 11 -> add_one
if even -> divide_by_two
if neither -> multiply_by_two

while it is:
if less then 11 -> add_one
elif even -> divide_by_two
else -> multiply_by_two

but because we supply two different lambda's, it's not immediately evident that they are combined in the actual run and reading this there is also no reason why this couldn't also send to divide_by_two at the same time.

So really, this is not a fan_out, this is just a conditional target setup.

python/samples/getting_started/workflow/step_00_foundation_patterns.py
```diff
@@ -0,0 +1,287 @@
...
+    def partition_func(number: int, total_targets: int) -> list[int]:
+        """Partition function to determine which executor to send the number to."""
+        if number < 12:
+            # Loop back to the add_one_executor if the number is less than 12
+            return [0]
+
+        if number % 2 == 0:
+            # Send it to the add_one_executor to add one more time and the
+            # divide_by_two_executor to divide the result by two.
+            return [0, 2]
```
eavanvalkenburg: this seems brittle, if I change the order of the param passed with: `[add_one_executor, multiply_by_two_executor, divide_by_two_executor],` below, I get different behavior...

python/samples/getting_started/workflow/step_00_foundation_patterns.py
```diff
@@ -0,0 +1,287 @@
...
+
+        # Otherwise, send it to the multiply_by_two_executor to multiply the result by two.
+        return [1]
+
+    workflow = (
+        WorkflowBuilder()
+        .set_start_executor(add_one_executor)
+        .add_partitioning_fan_out_edges(
```
eavanvalkenburg: I'm not really sure how this is different from a conditional, I would expect partitioned to be much more about partitioning the work across multiple nodes, and this partition_func is basically a variant of the conditional lambdas.

python/packages/workflow/agent_framework_workflow/_edge.py
```diff
@@ -98,57 +88,368 @@ async def send_message(self, message: Message, shared_state: SharedState, ctx: R
...
+
+        for edge in self._edges:
+            if edge.can_handle(message.data) and edge.should_route(message.data):
+                await edge.send_message(message, shared_state, ctx)
+                return True
+
+        return False
+
+
+class PartitioningEdgeGroup(SourceEdgeGroup):
```
eavanvalkenburg: also see my comment in the sample below, but to me partitioning is about spreading the input, across multiple targets, not about choosing which target to go that, that is conditional. 

My 2cts: make this truely a partitioning strategy, meaning it takes the input data and uses a partitioning function to decide which piece of data goes to which executor, while the conditional has two styles, one is based on multiple conditions, while the other uses a single function that decides which executor is next, and maybe the distinction is really that conditional should only send to one target (and thus is not a fan_out), while partition is a fan_out and returns a list of next executors that are then called in parallel (with the option of a list of length 1)

python/samples/getting_started/workflow/step_00_foundation_patterns.py
TaoChenOSU: We had this originally but we dropped it because a loop isn't a primitive type. It's something the user can achieve by creating two edges.

python/packages/workflow/agent_framework_workflow/_edge.py
lokitoth: In some sense we can think of the underlying operation as operating over a stream of message and selecting some subset of the available edges to route each message to. 

On top of this we build higher-level constructs, like the Switch executor (which uses a chain of predicates to select a "slate" in some sense of executors, or like the Map/Reduce pattern, or some other higher-level construct. 

The "edge assigner" variant (maybe a better name than partitioner, only have it that way because we started with thinking about Map/Reduce) is simply a lower level concept than the switch-equivalent.

(See https://github.com/microsoft/agent-framework/pull/261/commits/201349f82614e474de6194afbc9a95ecb1766239#diff-f6d249ab6b341f270f45ec28402db52b72c0526c9963d486ebf18508d0fd929fR92 for the equivalent in .NET)

python/samples/getting_started/workflow/step_00_foundation_patterns.py
TaoChenOSU: Yes, that will be a future feature: https://github.com/microsoft/agent-framework/issues/371

python/packages/workflow/agent_framework_workflow/_edge.py
TaoChenOSU: We need to support visualization and it'd be much easier if we have dedicated classes for different types of groups. 

Also with a single class to represent everything, some input parameters to the constructor must be mutually exclusive + some other potential constraints, i.e. what does it mean to allow multiple source to multiple targets. IMO, this makes the API even more confusing.

python/packages/workflow/agent_framework_workflow/_edge.py
TaoChenOSU: A message will be considered successfully sent if the target executor can handle the data type contained in the message. 

If the message is rejected because it fails a condition, it will still be considered as delivered successfully. Think of it this way, the message is delivered but it fails some user-defined condition, so no processing is triggered. 

As I explained here: https://github.com/microsoft/agent-framework/pull/393#discussion_r2268034163, we cannot raise exception if an executor cannot handler a message because an executor can be connected to many different executors that handler different types of data. If we raise because one of them cannot handle, then we will have many exception and it's not the behavior we design for.

python/packages/workflow/agent_framework_workflow/_workflow.py
TaoChenOSU: This now has been renamed to `add_conditional_edge_group`.

python/samples/getting_started/workflow/step_00_foundation_patterns.py
TaoChenOSU: Updated

eavanvalkenburg:
<a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/README.md">Python Test Coverage</a><details><summary>Python Test Coverage Report • </summary><table><tr><th>File</th><th>Stmts</th><th>Miss</th><th>Cover</th><th>Missing</th></tr><tbody><tr><td colspan="5"><b>packages/workflow/agent_framework_workflow</b></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_edge.py">_edge.py</a></td><td>168</td><td>9</td><td>94%</td><td><a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_edge.py#L74">74</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_edge.py#L86">86</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_edge.py#L99">99</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_edge.py#L118">118</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_edge.py#L123">123</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_edge.py#L128">128</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_edge.py#L133">133</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_edge.py#L138">138</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_edge.py#L417">417</a></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_executor.py">_executor.py</a></td><td>109</td><td>20</td><td>81%</td><td><a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_executor.py#L62">62</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_executor.py#L80">80</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_executor.py#L236-L240">236&ndash;240</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_executor.py#L245">245</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_executor.py#L247-L250">247&ndash;250</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_executor.py#L254-L256">254&ndash;256</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_executor.py#L258">258</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_executor.py#L262">262</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_executor.py#L264-L265">264&ndash;265</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_executor.py#L330">330</a></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_runner.py">_runner.py</a></td><td>170</td><td>28</td><td>83%</td><td><a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_runner.py#L84-L86">84&ndash;86</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_runner.py#L88">88</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_runner.py#L166-L168">166&ndash;168</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_runner.py#L189-L193">189&ndash;193</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_runner.py#L197">197</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_runner.py#L201-L202">201&ndash;202</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_runner.py#L223-L224">223&ndash;224</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_runner.py#L236-L237">236&ndash;237</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_runner.py#L248-L250">248&ndash;250</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_runner.py#L254">254</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_runner.py#L261-L263">261&ndash;263</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_runner.py#L268-L269">268&ndash;269</a></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_typing_utils.py">_typing_utils.py</a></td><td>19</td><td>9</td><td>52%</td><td><a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_typing_utils.py#L18">18</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_typing_utils.py#L30">30</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_typing_utils.py#L37-L40">37&ndash;40</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_typing_utils.py#L47-L48">47&ndash;48</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_typing_utils.py#L54">54</a></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_validation.py">_validation.py</a></td><td>211</td><td>21</td><td>90%</td><td><a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_validation.py#L197">197</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_validation.py#L315">315</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_validation.py#L318">318</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_validation.py#L376">376</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_validation.py#L380">380</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_validation.py#L467">467</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_validation.py#L469">469</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_validation.py#L472-L473">472&ndash;473</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_validation.py#L475-L477">475&ndash;477</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_validation.py#L479-L481">479&ndash;481</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_validation.py#L484-L488">484&ndash;488</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_validation.py#L493">493</a></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_workflow.py">_workflow.py</a></td><td>181</td><td>33</td><td>81%</td><td><a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_workflow.py#L52">52</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_workflow.py#L118">118</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_workflow.py#L140">140</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_workflow.py#L190">190</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_workflow.py#L198-L200">198&ndash;200</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_workflow.py#L212">212</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_workflow.py#L226">226</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_workflow.py#L344">344</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_workflow.py#L364-L365">364&ndash;365</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_workflow.py#L367-L369">367&ndash;369</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_workflow.py#L384-L386">384&ndash;386</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_workflow.py#L394-L395">394&ndash;395</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_workflow.py#L404-L410">404&ndash;410</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_workflow.py#L415">415</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_workflow.py#L417">417</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_workflow.py#L489">489</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_workflow.py#L491">491</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_workflow.py#L513">513</a>, <a href="https://github.com/microsoft/agent-framework/blob/0edcac288d01187fe72d0fb1a30c416d7dffeb37/packages/workflow/agent_framework_workflow/_workflow.py#L515">515</a></td></tr><tr><td><b>TOTAL</b></td><td><b>4441</b></td><td><b>1002</b></td><td><b>77%</b></td><td>&nbsp;</td></tr></tbody></table></details>

## Python Unit Test Overview
| 511 | 31 :zzz: | 0 :x: | 0 :fire: | 8.689s :stopwatch: |

python/packages/workflow/agent_framework_workflow/_edge.py
TaoChenOSU: I do agree that the name "partitioning" is misleading and we should change that. Here are some suggestions from AI:
- MulticastEdgeGroup - Emphasizes that the same message is sent to multiple targets
- SelectiveEdgeGroup - Indicates that it selectively chooses which executors to send to
- DistributionEdgeGroup - Shows that it distributes the message to selected executors
- ReplicationEdgeGroup - Emphasizes that the message is replicated to multiple targets
- RoutingEdgeGroup - Generic routing based on selection logic
- FilteredFanOutEdgeGroup - More descriptive, showing it's a fan-out with filtering

That being said, I think it will be beneficial to distinguish between the partitioning group and the conditional edge group while sharing the same underlying selection logic.

Or, a more natural way to express both scenarios could be a way that resembles programming languages:
```python
@dataclass
class Case:
  condition: Callable[[Any], bool]
  target: Executor
  break: bool = True

@dataclass
class Default:
  target: Executor

# Switch with breaks and a default (Exactly one executor will be triggered)
.add_conditional_edge_group(
  source,
  [
    Case(condition_1, executor_1),
    Case(condition_2, executor_2),
    Default(executor_3),
  ]
)

# Switch without breaks (0 or more executors will be triggered)
.add_conditional_edge_group(
  source,
  [
    Case(condition_1, executor_1, break=False),
    Case(condition_2, executor_2, break=False),
  ]
)
```
The above even allows more customization. For example,
```python
.add_conditional_edge_group(
  source,
  [
    Case(condition_1, executor_1, break=False),
    Case(condition_2, executor_2, break=True),
    Case(condition_3, executor_3, break=True),
  ]
)
```

python/samples/getting_started/workflow/step_00_foundation_patterns.py
```diff
@@ -0,0 +1,291 @@
...
+    multiply_by_two_executor = MultiplyByTwoExecutor()
+    divide_by_two_executor = DivideByTwoExecutor()
+    aggregate_result_executor = AggregateResultExecutor()
+
+    workflow = (
+        WorkflowBuilder()
+        .set_start_executor(add_one_executor)
+        .add_conditional_edge_group(
+            add_one_executor,
+            [
```
moonbox3: Can we use kwarg params here so it's clear to devs what the arguments to `add_conditional_edge_group()` are?

python/samples/getting_started/workflow/step_00_foundation_patterns.py
```diff
@@ -0,0 +1,291 @@
...
+        Adding one to the number: 10 Result: 11
+        Aggregating results: 10
+        Aggregating results: 11
+    """
+    add_one_executor_a = AddOneExecutor()
+    aggregate_result_executor = AggregateResultExecutor()
+
+    workflow = (
+        WorkflowBuilder()
+        .add_edge(add_one_executor_a, add_one_executor_a, condition=lambda x: x < 11)
```
moonbox3, line 125: Some thoughts looking at this sample:

- I understand that this condition is simple; however, a particular condition _could_ technically spin infinitely. Are we relying on the max super steps to eventually kick us out of an unwanted scenario?
- How do we plan to create a template for a workflow when it has a lambda like this defined? Inline lambdas aren't serializable or introspectable. 
- We also talk about promoting type safety: we don't get it with this particular sample.
- How are we advertising fan-out or first-match behavior with routing? Given the overlap on 9 and 10.

TaoChenOSU: 1. I think that's correct. I don't think we can detect every user error. If they set up the workflow to run indefinitely, then we will run it indefinitely.
2. I believe this is still an open question. I will need to see how .Net does it. I believe CSDL only support PowerFx.
3. Agreed. We should think about how to do that. This can be part of the effort to bring in typed workflow for Python: https://github.com/microsoft/agent-framework/issues/319
4. I am not sure if I understand. Could you provide more context?

python/packages/workflow/agent_framework_workflow/_edge.py
```diff
@@ -98,57 +82,338 @@ async def send_message(self, message: Message, shared_state: SharedState, ctx: R
...
+        """Get the target executor of the edge."""
+        return [self._edge.target]
+
+    @property
+    def edges(self) -> list[Edge]:
+        """Get the edges in the group."""
+        return [self._edge]
+
+
+class SourceEdgeGroup(EdgeGroup):
```
eavanvalkenburg: I think this name is a bit confusing, should be in line with the `add_...` methods in the builder.

Mostly to make sure that someone doesn't suddenly see this weird term in their log that they don't understand.
```suggestion
class FanOutEdgeGroup(EdgeGroup):
```

python/packages/workflow/agent_framework_workflow/_edge.py
```diff
@@ -98,57 +82,338 @@ async def send_message(self, message: Message, shared_state: SharedState, ctx: R
...
+    def edges(self) -> list[Edge]:
+        """Get the edges in the group."""
+        return self._edges
+
+    def _validate_selection_result(self, selection_results: list[str]) -> bool:
+        """Validate the selection results to ensure all IDs are valid target executor IDs."""
+        return all(result in self._target_ids for result in selection_results)
+
+
+class TargetEdgeGroup(EdgeGroup):
```
eavanvalkenburg:
```suggestion
class FanInEdgeGroup(EdgeGroup):
```

PR #401, opened 2025-08-12
.NET: Add Declarative Workflow Interpreter

crickman:
### Motivation and Context

Initial support for declarative foundry workflow based on MCS object model: developer experience

### Description

- Using V1 agent services client
- Using unmodified CopilotStudio object model (uses `AnswerQuestionWithAI` to invoke agent)
- Working demo @ `/dotnet/demos/DeclarativeWorkflow`
- Includes functional "Student Teacher" and "Deep Research" sample workflows
   (samples @ `/workflows`)
- Supports streaming and non-streaming agent responses

### Next Steps

- https://github.com/microsoft/agent-framework/issues/485
- https://github.com/microsoft/agent-framework/issues/486
- https://github.com/microsoft/agent-framework/issues/488
- https://github.com/microsoft/agent-framework/issues/516

### Contribution Checklist

- [X] I didn't break anyone :smile:


dotnet/src/Microsoft.Agents.Workflows.Declarative/Exceptions/InvalidSegmentException.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+
+namespace Microsoft.Agents.Workflows.Declarative;
+
+/// <summary>
+/// Represents an exception that occurs when an action is invalid or cannot be processed.
+/// </summary>
+public sealed class InvalidSegmentException : DeclarativeWorkflowException
```
adityamandaleeka: I'm curious why all these exception classes need to exist rather than just having one with an enum for the error type.

dotnet/src/Microsoft.Agents.Workflows.Declarative/Interpreter/DeclarativeActionExecutor.cs
```diff
@@ -0,0 +1,114 @@
...
+internal sealed record class ExecutionResultMessage(string ExecutorId, object? Result = null);
+
+internal abstract class DeclarativeActionExecutor<TAction>(TAction model) :
+    WorkflowActionExecutor(model)
+    where TAction : DialogAction
+{
+    public new TAction Model => (TAction)base.Model;
+}
+
+internal abstract class WorkflowActionExecutor :
```
adityamandaleeka, line 28: Should these Debug.WriteLines instead be fed to a Logger?

dotnet/src/Microsoft.Agents.Workflows.Declarative/DeclarativeWorkflowBuilder.cs
```diff
@@ -0,0 +1,93 @@
...
+    {
+        BotElement rootElement = YamlSerializer.Deserialize<BotElement>(yamlReader) ?? throw new UnknownActionException("Unable to parse workflow.");
+
+        // ISSUE #486 - Use "Workflow" element for Foundry specific.
+        if (rootElement is not AdaptiveDialog workflowElement)
+        {
+            throw new UnknownActionException($"Unsupported root element: {rootElement.GetType().Name}. Expected an {nameof(AdaptiveDialog)}.");
+        }
+
+        string rootId = WorkflowActionVisitor.RootId(workflowElement.BeginDialog?.Id.Value ?? "workflow");
```
adityamandaleeka, line 59: Is this okay? Could it lead to a conflict? Should explicit IDs be enforced at authoring time?

dotnet/src/Microsoft.Agents.Workflows.Declarative/DeclarativeWorkflowBuilder.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        DeclarativeWorkflowExecutor<TInput> rootExecutor = new(rootId, WrapWithBot(workflowElement), message => DefaultTransform(message));
```
adityamandaleeka: Should this be using the inputTransform if provided?

dotnet/src/Microsoft.Agents.Workflows.Declarative/DeclarativeWorkflowBuilder.cs
```diff
@@ -0,0 +1,93 @@
...
+    /// <param name="options">The execution context for the workflow.</param>
+    /// <param name="inputTransform">An optional function to transform the input message into a <see cref="ChatMessage"/>.</param>
+    /// <returns>The <see cref="Workflow"/> that corresponds with the YAML object model.</returns>
+    public static Workflow<TInput> Build<TInput>(
+        TextReader yamlReader,
+        DeclarativeWorkflowOptions options,
+        Func<TInput, ChatMessage>? inputTransform = null)
+        where TInput : notnull
+    {
+        BotElement rootElement = YamlSerializer.Deserialize<BotElement>(yamlReader) ?? throw new UnknownActionException("Unable to parse workflow.");
```
adityamandaleeka: UnknownActionException is kinda weird for a deserialization error right?

dotnet/src/Microsoft.Agents.Workflows.Declarative/DeclarativeWorkflowInvokeEvent.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Microsoft.Agents.Workflows.Declarative;
+
+/// <summary>
+/// Event that represents a message produced by a declarative workflow.
+/// </summary>
+public class DeclarativeWorkflowInvokeEvent(string conversationId) : DeclarativeWorkflowEvent(conversationId)
+{
+    /// <summary>
+    /// The converation ID associated with the workflow.
```
adityamandaleeka:
```suggestion
    /// The conversation ID associated with the workflow.
```

dotnet/src/Microsoft.Agents.Workflows.Declarative/WorkflowAgentProvider.cs
```diff
@@ -0,0 +1,58 @@
...
+/// project endpoint and credentials to authenticate requests.</remarks>
+/// <param name="projectEndpoint">The endpoint URL of the Foundry project. This must be a valid, non-null URI pointing to the project.</param>
+/// <param name="projectCredentials">The credentials used to authenticate with the Foundry project. This must be a valid instance of <see cref="TokenCredential"/>.</param>
+/// <param name="httpClient">An optional <see cref="HttpClient"/> instance to be used for making HTTP requests. If not provided, a default client will be used.</param>
+public sealed class FoundryAgentProvider(string projectEndpoint, TokenCredential? projectCredentials = null, HttpClient? httpClient = null) : WorkflowAgentProvider
+{
+    /// <inheritdoc/>
+    public override async Task<AIAgent> GetAgentAsync(string agentId, CancellationToken cancellationToken = default)
+    {
+        AIAgent agent = await this.CreateClient().GetAIAgentAsync(agentId, chatOptions: null, cancellationToken).ConfigureAwait(false);
```
adityamandaleeka: Should this be cached or lazy? Or is it okay for a new client to be instantiated for each call?

dotnet/src/Microsoft.Agents.Workflows.Declarative/DeclarativeWorkflowOptions.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Net.Http;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+
+namespace Microsoft.Agents.Workflows.Declarative;
+
+/// <summary>
+/// Configuration options for workflow execution.
+/// </summary>
+public sealed class DeclarativeWorkflowOptions(WorkflowAgentProvider agentProvider)
```
adityamandaleeka, line 13: null check for this?

dotnet/src/Microsoft.Agents.Workflows.Declarative/DeclarativeWorkflowBuilder.cs
crickman: Indeed.  Great eye.  Updated.

dotnet/src/Microsoft.Agents.Workflows.Declarative/WorkflowAgentProvider.cs
crickman: Should be cached; although clients are relatively cheap.  Updated.

dotnet/src/Microsoft.Agents.Workflows.Declarative/DeclarativeWorkflowOptions.cs
crickman: Yes, its public surface area.  Updated.

dotnet/src/Microsoft.Agents.Workflows.Declarative/DeclarativeWorkflowBuilder.cs
crickman: Id are enforced at authoring time, but the bot model must support the definition of elements that are in an invalid state (since they may be in the midst of being authored).  As such, there's a bunch of nullable state that is generally defined.  What's going on here is that the `BeginDialog` element might be missing.

There was some early conversation about not requiring ids...particularly at the workflow level, so I didn't want to make this a failure case and I'm averse to overriding null state with `!`.

Our usage of the workflow id is for internal management only.  We have a work item to introduce a `Workflow` root node (instead of `AdaptiveDialog`).  As part of this, it was considered that `Workflow` not require any `id` property, so this is very much in-motion.  

https://github.com/microsoft/agent-framework/issues/486

The current usage is for internal booking keeping only.

dotnet/src/Microsoft.Agents.Workflows.Declarative/Interpreter/DeclarativeActionExecutor.cs
crickman: Yes, there's some debug spew I want to keep debug...but we absolutely need a telemetry pass as a follow-up to this change.  I missed this when I was filing follow-up issues:  https://github.com/microsoft/agent-framework/issues/516

dotnet/src/Microsoft.Agents.Workflows.Declarative/DeclarativeWorkflowBuilder.cs
crickman: Updated as part of the exception sweep.  

Note: The serializer will throw for a deserialization error.  For a null result, _something_ was deserialized...but its unknown / not supported.  Tweaking the message helps.

dotnet/src/Microsoft.Agents.Workflows.Declarative/Exceptions/InvalidSegmentException.cs
crickman: Well why not have one sealed `System.Exception` class that reports an error code?  

A: `catch` handler discriminates on type.  Recenty, the `when` clause provides more flexibility...but error codes seems a retro.

Anyway, I'd meant to make a pass to reconsile these...which I've done now...into two main types that share a base type:

- `DeclarativeModelException` - Failure due how workflow is modeled
- `DeclarativeActionException` - Failure during action execution

workflows/README.md
ekzhu, line 1: I think it would be good to rename the folder "workflows" to "delcarative" for better parallel with "python" and "dotnet".

workflows/README.md
```diff
@@ -0,0 +1,27 @@
+# No-Code Workflows
```
ekzhu:
```suggestion
# Declarative Workflows
```

workflows/README.md
```diff
@@ -0,0 +1,27 @@
...
+
```
ekzhu, line 2: Let's add something here regarding the readiness of this feature. e.g., "This is feature is for internal use only at this moment."

crickman: I'm not sure that is entirely true, but we plan to follow-up with additinal documentation.  Certainly, the entire repo is preview...at a minimum.

PR #404, opened 2025-08-12
Python: WIP: Python Agent Runtime

adityamandaleeka:
This PR introduces the initial Python agent runtime implementation, providing foundational actor system infrastructure that mirrors the .NET side's patterns/layering. The focus so far has been on establishing architectural parity. There are still some features missing (e.g. streaming, telemetry) which will need to be added.

### DIFF
--- a/python/runtime/agent_runtime/runtime.py
+++ b/python/runtime/agent_runtime/runtime.py
@@ -193,15 +193,17 @@ class InProcessActorContext(ActorRuntimeContext):
         logger.debug(f"Message {message.message_id} enqueued for actor {self._actor_id}")
 
     def send_request(self, request: ActorRequestMessage) -> "InProcessResponseHandle":
-        if request.message_id in self._inbox:
+        try:
             return InProcessResponseHandle(self._inbox[request.message_id])
-        entry = RequestEntry(request)
-        self._inbox[request.message_id] = entry
-        self.enqueue_message(request)
-        return InProcessResponseHandle(entry)
+        except KeyError:
+            # Message ID not in inbox, create new entry
+            entry = RequestEntry(request)
+            self._inbox[request.message_id] = entry
+            self.enqueue_message(request)
+            return InProcessResponseHandle(entry)
 
     def complete_request(self, message_id: str, response: ActorResponseMessage) -> None:
-        if message_id in self._inbox:
+        try:
             entry = self._inbox[message_id]
             entry.set_response(response)
             if message_id not in self._completed_order:
@@ -211,10 +213,16 @@ class InProcessActorContext(ActorRuntimeContext):
                 if oldest != message_id:
                     self._inbox.pop(oldest, None)
             logger.debug(f"Request {message_id} completed for actor {self._actor_id}")
+        except KeyError:
+            # Message ID not found in inbox
+            pass
 
     def on_progress_update(self, message_id: str, sequence_number: int, data: Any) -> None:
-        if message_id in self._inbox:
+        try:
             self._inbox[message_id].add_progress_update(sequence_number, data)
+        except KeyError:
+            # Message ID not found in inbox
+            pass
 
 
 class InProcessActorRuntime:


python/runtime/agent_runtime/runtime_abstractions.py
```diff
@@ -0,0 +1,173 @@
...
+class RequestStatus(Enum):
+    """Status of a request being processed by an actor"""
+    PENDING = "pending"
+    COMPLETED = "completed" 
+    FAILED = "failed"
+    NOT_FOUND = "not_found"
+
+
+@dataclass
+class ActorMessage:
```
johanste: We don't need this as a public class. An ActorMessage is really a Union[ActorRequestMessage, ActorResponseMessage].

python/runtime/agent_runtime/runtime_abstractions.py
```diff
@@ -0,0 +1,173 @@
...
+    timestamp: datetime = field(default_factory=datetime.utcnow)
+
+
+@dataclass
+class ActorRequestMessage(ActorMessage):
+    """Request message sent to an actor"""
+    method: str = ""
+    params: Optional[Dict[str, Any]] = None
+    
+    def __post_init__(self):
```
johanste: `message_type: typing. Literal[ActorMessageType.REQUEST] = ActorMessageType.REQUEST`

Presumably we don't want users to be able to mutate the message type discriminator?

python/runtime/agent_runtime/runtime_abstractions.py
```diff
@@ -0,0 +1,173 @@
...
+        self.message_type = ActorMessageType.REQUEST
+
+
+@dataclass
+class ActorResponseMessage(ActorMessage):
```
johanste: Given that there is no real ordering between the parameters and they are all optional, we should make them keyword-only.

```suggestion
class ActorResponseMessage(ActorMessage, kwonly=True):
```

python/runtime/agent_runtime/runtime_abstractions.py
```diff
@@ -0,0 +1,173 @@
...
+    sender_id: Optional[ActorId] = None
+    status: RequestStatus = RequestStatus.PENDING
+    data: Any = None
+    
+    def __post_init__(self):
+        self.message_type = ActorMessageType.RESPONSE
+
+
+# Runtime Infrastructure Interfaces
+class IActorRuntimeContext(ABC):
```
johanste: Ideally, we should try to get rid of the class altogether, but since it is presumably not something that most users have to instantiate in 101-level scenarios, it is not at the top of priorities to remove.

Is there any reason to not call this `RuntimeContext`?

python/runtime/agent_runtime/runtime.py
```diff
@@ -0,0 +1,288 @@
...
+    def send_request(self, request: ActorRequestMessage) -> "InProcessResponseHandle":
+        if request.message_id in self._inbox:
+            return InProcessResponseHandle(self._inbox[request.message_id])
+        entry = RequestEntry(request)
+        self._inbox[request.message_id] = entry
+        self.enqueue_message(request)
+        return InProcessResponseHandle(entry)
+
+    def complete_request(self, message_id: str, response: ActorResponseMessage) -> None:
+        if message_id in self._inbox:
```
johanste: Nitpick; don't check for `in` - just wrap a `self._inbox[message_id]` in a try/except and catch the `KeyError`.

python/runtime/agent_runtime/runtime.py
```diff
@@ -0,0 +1,288 @@
...
+            task.cancel()
+        
+        # Wait for tasks to complete
+        if self._actor_tasks:
+            await asyncio.gather(*self._actor_tasks.values(), return_exceptions=True)
+        
+        # Cleanup actors
+        for actor_context in self._actors.values():
+            # Allow actors to cleanup
+            pass
```
johanste, line 263: Missing cleanup?

python/runtime/agent_runtime/runtime.py
```diff
@@ -0,0 +1,288 @@
...
+
+
+class InProcessActorContext(IActorRuntimeContext):
+    """Runtime context for an in-process actor"""
+
+    def __init__(
+        self,
+        actor_id: ActorId,
+        storage: IActorStateStorage,
+        runtime: Any
```
johanste: Are we missing an actual runtime type?

python/runtime/agent_runtime/runtime.py
```diff
@@ -0,0 +1,288 @@
...
+        if actor_id in self._actors:
+            return self._actors[actor_id]
+        
+        # Check if we have a factory for this actor type
+        if actor_id.type_name not in self._actor_factories:
+            raise ValueError(f"No factory registered for actor type: {actor_id.type_name}")
+        
+        # Create the actor context
+        context = InProcessActorContext(actor_id, self._storage, self)
+        self._actors[actor_id] = context
```
johanste, line 280: There is some naming confusion here - are we creating/storing an actor or an actor context? This also applies to the method name.

python/runtime/agent_runtime/runtime_abstractions.py
```diff
@@ -0,0 +1,173 @@
...
+
+class IActor(ABC):
+    """Interface for all actors in the system (runtime infrastructure)"""
+    
+    @abstractmethod
+    async def run(self, context: IActorRuntimeContext) -> None:
+        """Main actor execution loop"""
+        pass
+    
+    async def dispose(self) -> None:
```
johanste:
```suggestion
    async def close(self) -> None:
```

python/runtime/agent_runtime/agent_actor.py
```diff
@@ -0,0 +1,326 @@
...
+import asyncio
+import json
+import logging
+import os
+import sys
+from typing import Optional, List
+
+# Runtime infrastructure types
+from .runtime_abstractions import (
+    IActor, IActorRuntimeContext, ActorMessage, ActorRequestMessage, ActorResponseMessage,
```
moonbox3: Is it possible to stick to more Pythonic naming conventions, which means we don't name things with the `I` prefix, as is done in dotnet? `IActor` -> `Actor`, `IActorRuntimeContext` -> `ActorRuntimeContext, etc.

python/runtime/agent_runtime/__init__.py
```diff
@@ -0,0 +1,17 @@
+# Agent Runtime - Python Implementation
+
+from .agent_proxy import AgentProxy, AgentProxyThread
+from .http_actor_client import HttpActorClient
+from .runtime import InProcessActorRuntime, InProcessActorClient
+from .runtime_abstractions import ActorId, IActorClient, RequestStatus
+
+__all__ = [
```
moonbox3, line 10: I think the `runtime` package should exist in the `agent_framework/python/packages` directory? It would be something we want devs to install during setup, like `agent_framework[runtime]` correct? @eavanvalkenburg thoughts?

python/runtime/agent_runtime/agent_actor.py
```diff
@@ -0,0 +1,326 @@
...
+
+class AgentActor(IActor):
+    """Runtime actor that wraps framework AI agents"""
+    
+    THREAD_STATE_KEY = "agent_thread"
+    
+    def __init__(self, agent: AIAgent):
+        """Initialize with framework agent"""
+        self._agent = agent
+        self._thread: Optional[AgentThread] = None
```
moonbox3: We support Python 3.10+, so don't need `Optional`:

```suggestion
        self._thread: AgentThread | None = None
```

python/runtime/agent_runtime/agent_actor.py
```diff
@@ -0,0 +1,326 @@
...
+    
+    async def run(self, context: IActorRuntimeContext) -> None:
+        """Main actor execution loop"""
+        agent_name = getattr(self._agent, 'name', None) or getattr(self._agent, 'id', 'unknown')
```
moonbox3: An `AIAgent` will always have an `id` and a `display_name`. Should we use that instead? Can simplify to either:

```python
agent_name = self._agent.display_name
```

If we want the user friendly name for logging

python/runtime/agent_runtime/agent_actor.py
```diff
@@ -0,0 +1,326 @@
...
+                logger.error(f"Failed to restore thread state: {e}")
+                self._thread = self._agent.get_new_thread()
+        else:
+            self._thread = self._agent.get_new_thread()
+            logger.debug("Created new thread")
+    
+    async def _handle_agent_request(self, request: ActorRequestMessage, context: IActorRuntimeContext):
+        """Handle agent run requests using framework types directly"""
+        
+        if request.method != "run":
```
moonbox3, line 193: Can we have request method enums instead of direct string comparison?

python/runtime/agent_runtime/agent_actor.py
```diff
@@ -0,0 +1,326 @@
...
+            framework_messages = self._parse_framework_messages(request)
+
+            # Ensure a thread exists (may be None if _restore_thread_state not yet called in some test paths)
+            if self._thread is None:
+                self._thread = self._agent.get_new_thread()
+
+            # Don't manually append incoming messages; allow agent implementation to manage thread.
+
+            # Call framework agent directly
+            response = await self._agent.run(framework_messages, thread=self._thread)
```
moonbox3, line 218: Will we support other agent-specific `**kwargs`?

moonbox3, line 218: What happens if an exception is raised in the agent `run` method?

Also, will we have support for `run_streaming(...)`?

python/runtime/agent_runtime/agent_actor.py
```diff
@@ -0,0 +1,326 @@
...
+                # Add timestamp if available
+                if hasattr(framework_msg, 'timestamp'):
+                    msg_data["timestamp"] = framework_msg.timestamp
+                
+                messages_data.append(msg_data)
+            
+            # Build response data
+            response_data = {
+                "messages": messages_data,
+                "status": "completed"
```
moonbox3: Do we have a runtime status enum we can re-use through the runtime code?

python/runtime/agent_runtime/agent_actor.py
```diff
@@ -0,0 +1,326 @@
...
+            role = framework_msg.role
+            # Handle enum vs string
+            return role.value if hasattr(role, 'value') else str(role)
+        except Exception:
+            return "assistant"
+    
+    def _extract_content(self, framework_msg: ChatMessage) -> str:
+        """Extract text content from framework message"""
+        try:
+            # Framework messages use 'text' attribute primarily
```
moonbox3: There could be file or image content though?

python/runtime/agent_runtime/http_actor_client.py
```diff
@@ -0,0 +1,241 @@
...
+        self._response: Optional[ActorResponseMessage] = None
+
+    async def get_response(self) -> ActorResponseMessage:
+        """Get the final response from the HTTP actor."""
+        if self._response is not None:
+            return self._response
+        
+        # Poll for the response (simplified implementation)
+        # In a real implementation, this would use WebSockets or Server-Sent Events
+        max_retries = 30  # 30 seconds timeout
```
moonbox3, line 37: Configurable?

python/runtime/agent_runtime/http_actor_client.py
```diff
@@ -0,0 +1,241 @@
...
+        
+        # If it's a successful response, we can simulate streaming
+        if response.status == RequestStatus.COMPLETED and response.data:
+            # Simulate streaming by breaking the response into chunks
+            if isinstance(response.data, dict) and "messages" in response.data:
+                for msg_data in response.data["messages"]:
+                    content = msg_data.get("text", "")
+                    if content:
+                        # Simulate streaming chunks
+                        words = content.split()
```
moonbox3, line 83: Can you please help me understand why we're trying to do pseudo-streaming?

python/runtime/agent_runtime/http_actor_client.py
```diff
@@ -0,0 +1,241 @@
...
+            message_id = str(uuid.uuid4())
+            
+        # Build the request payload matching the HTTP API format
+        payload = {
+            "agent_name": actor_id.type_name,
+            "conversation_id": actor_id.instance_id,
+            "messages": params.get("messages", []) if params else []
+        }
+        
+        async with httpx.AsyncClient(base_url=self._base_url, timeout=30.0) as client:
```
moonbox3, line 140: Will we allow custom `AsyncClient`s to be used? Or a configurable timeout?

python/runtime/agent_runtime/http_api.py
```diff
@@ -0,0 +1,160 @@
...
+from .runtime import InProcessActorRuntime, InProcessActorClient
+from .agent_actor import AgentActor, MockAIAgent, EchoAgent
+from .runtime_abstractions import ActorId
+
+
+logger = logging.getLogger(__name__)
+
+
+# Request/Response models
+class ChatMessageModel(BaseModel):
```
moonbox3: AF uses `AFBaseModel` as the base class for all Pydantic models in AF. Should we use that here? Also, do these need to be Pydantic models? Or just data classes?

python/runtime/agent_runtime/runtime.py
```diff
@@ -0,0 +1,288 @@
...
+
+    def on_progress_update(self, message_id: str, sequence_number: int, data: Any) -> None:
+        if message_id in self._inbox:
+            self._inbox[message_id].add_progress_update(sequence_number, data)
+
+
+class InProcessActorRuntime:
+    """In-process actor runtime"""
+    
+    def __init__(self, storage: Optional[IActorStateStorage] = None):
```
moonbox3:
```suggestion
    def __init__(self, storage: IActorStateStorage | None = None):
```

python/runtime/agent_runtime/runtime.py
```diff
@@ -0,0 +1,288 @@
...
+        self._storage = storage or InMemoryStateStorage()
+        self._actors: Dict[ActorId, InProcessActorContext] = {}
```
moonbox3:
```suggestion
        self._actors: dict[ActorId, InProcessActorContext] = {}
```
And for all other refs

python/runtime/agent_runtime/runtime.py
```diff
@@ -0,0 +1,288 @@
+"""In-process actor runtime implementation"""
+
+import asyncio
+import logging
+import uuid
+from concurrent.futures import Future
+from dataclasses import dataclass
+from typing import Dict, Optional, Callable, Any, AsyncIterator, List
```
moonbox3: Py 3.10+ doesn't require typing imports for `dict`, `list` and we can remove `Optional`

python/runtime/agent_runtime/runtime.py
```diff
@@ -0,0 +1,288 @@
...
+@dataclass
+class RequestEntry:
+    """Tracks a request, its response, and progress updates."""
+    request: ActorRequestMessage
+    _response_future: Future
+    _progress_queue: asyncio.Queue
+
+    def __init__(self, request: ActorRequestMessage):
+        self.request = request
+        self._response_future = Future()  # concurrent.futures.Future
```
moonbox3, line 103: We could eliminate the `Future` if we model everything as a single stream with an explicit terminal message. For example, we put progress updates on the `Queue`. Then when the final result arrives, enqueue a “final” message and then a sentinel to close the stream. Consumers of this use `async for` to iterate the stream. No separate Future needed. Thoughts?

python/runtime/agent_runtime/runtime.py
```diff
@@ -0,0 +1,288 @@
...
+    def __init__(self, storage: Optional[IActorStateStorage] = None):
+        self._storage = storage or InMemoryStateStorage()
+        self._actors: Dict[ActorId, InProcessActorContext] = {}
+        self._actor_tasks: Dict[ActorId, asyncio.Task] = {}
+        self._actor_factories: Dict[str, Callable[[], IActor]] = {}
+        self._running = False
+    
+    def register_actor_type(self, type_name: str, factory: Callable[[IActorRuntimeContext], IActor]) -> None:
+        """Register a factory for creating actors of a given type"""
+        self._actor_factories[type_name] = factory
```
moonbox3, line 240: Do we allow for overwrite or should we throw if the `type_name` already exists?

python/runtime/agent_runtime/runtime_abstractions.py
moonbox3: I feel like most of these runtime abstractions can be folded into the appropriate files, without the need for a specific `runtime_abstractions` file.

python/runtime/agent_runtime/__init__.py
eavanvalkenburg: yes, this should indeed become a separate package, where we reimport in the main, similar to `workflow` and others, and please @adityamandaleeka  copy the setup (pyproject, readme, license, etc) from one of the existing projects to ensure things like type checking etc, work.

python/runtime/agent_runtime/agent_actor.py
eavanvalkenburg: yeah, this def needs to change!

python/runtime/agent_runtime/agent_actor.py
```diff
@@ -0,0 +1,326 @@
...
+from typing import Optional, List
+
+# Runtime infrastructure types
+from .runtime_abstractions import (
+    IActor, IActorRuntimeContext, ActorMessage, ActorRequestMessage, ActorResponseMessage,
+    RequestStatus, ActorMessageType
+)
+
+# Framework agent types
+sys.path.append(os.path.join(os.path.dirname(__file__), '../../packages/main'))
```
eavanvalkenburg: this won't work when installed somewhere, when you take the setup of another project, you then have `agent_framework` as a dependency so you can just import then

python/runtime/agent_runtime/runtime_abstractions.py
```diff
@@ -0,0 +1,173 @@
+"""Runtime infrastructure abstractions (actor system only)"""
```
eavanvalkenburg: this needs to be the MS copyright header

python/runtime/agent_runtime/runtime.py
eavanvalkenburg: or use `.get`

python/runtime/agent_runtime/agent_actor.py
```diff
@@ -0,0 +1,326 @@
...
+        thread_data = await context.read_state(self.THREAD_STATE_KEY)
+        
+        if thread_data:
+            try:
+                # For now, create new thread - full serialization would need framework support
+                self._thread = self._agent.get_new_thread()
+                logger.debug("Restored thread state (simplified)")
+            except Exception as e:
+                logger.error(f"Failed to restore thread state: {e}")
+                self._thread = self._agent.get_new_thread()
```
eavanvalkenburg, line 186: why would this succeed if the first call to `get_new_thread` fails, seems this whole try block is unnecessary

python/runtime/agent_runtime/agent_actor.py
```diff
@@ -0,0 +1,326 @@
...
+        # Process messages
+        async for message in context.watch_messages():
+            if isinstance(message, ActorRequestMessage):
+                await self._handle_agent_request(message, context)
+    
+    async def _restore_thread_state(self, context: IActorRuntimeContext):
+        """Restore the agent thread state from storage"""
+        thread_data = await context.read_state(self.THREAD_STATE_KEY)
+        
+        if thread_data:
```
eavanvalkenburg, line 179: the thread_data is not used

python/pyproject.toml
```diff
@@ -185,7 +191,7 @@ build = "python run_tasks_in_packages_if_exists.py build"
 # combined checks
 check = ["fmt", "lint", "pyright", "mypy", "test", "markdown-code-lint", "samples-code-check"]
 pre-commit-check = ["fmt", "lint", "pyright", "markdown-code-lint", "samples-code-check"]
-all-tests = "pytest --cov=agent_framework --cov=agent_framework_azure --cov=agent_framework_foundry --cov=agent_framework_workflow --cov-report=term-missing:skip-covered packages/**/tests"
```
eavanvalkenburg, line 188: don't delete this!

python/pyproject.toml
```diff
@@ -44,6 +44,12 @@ dev = [
     "redis",
     "sphinx-autobuild",
     "aiofiles>=24.1.0",
+    # Web chat sample dependencies
```
eavanvalkenburg, line 47: these are not dev dependencies

python/runtime/agent_runtime/agent_actor.py
```diff
@@ -0,0 +1,326 @@
...
+            logger.error(f"Error converting framework response: {e}")
+            return {
+                "messages": [{"role": "assistant", "text": f"Response conversion error: {e}"}],
+                "status": "failed"
+            }
+    
+    def _extract_role(self, framework_msg: ChatMessage) -> str:
+        """Extract role from framework message"""
+        try:
+            role = framework_msg.role
```
eavanvalkenburg: role is always parsed to the enum
```suggestion
            return framework_msg.role.value
```

python/runtime/agent_runtime/agent_actor.py
```diff
@@ -0,0 +1,326 @@
...
+            # Handle enum vs string
+            return role.value if hasattr(role, 'value') else str(role)
+        except Exception:
+            return "assistant"
+    
+    def _extract_content(self, framework_msg: ChatMessage) -> str:
+        """Extract text content from framework message"""
+        try:
+            # Framework messages use 'text' attribute primarily
+            if hasattr(framework_msg, 'text'):
```
eavanvalkenburg: text is always present (it is a property) and it has a non-empty string value when there are 1 or more TextContents in the contents list

python/runtime/agent_runtime/agent_actor.py
```diff
@@ -0,0 +1,326 @@
...
+                return str(framework_msg.text)
+            elif hasattr(framework_msg, 'content'):
```
eavanvalkenburg: content doesn't exist, `contents` does, and it is a list of content types, which might include textcontent (but that is already returned with .text) but also other types, as Evan mentions above

python/runtime/agent_runtime/agent_actor.py
```diff
@@ -0,0 +1,326 @@
...
+    def _extract_role(self, framework_msg: ChatMessage) -> str:
+        """Extract role from framework message"""
+        try:
+            role = framework_msg.role
+            # Handle enum vs string
+            return role.value if hasattr(role, 'value') else str(role)
+        except Exception:
+            return "assistant"
+    
+    def _extract_content(self, framework_msg: ChatMessage) -> str:
```
eavanvalkenburg: if this is only interested in string content, then it is not extracting content:

```suggestion
    def _extract_string_content(self, framework_msg: ChatMessage) -> str:
```

python/runtime/agent_runtime/agent_proxy.py
```diff
@@ -0,0 +1,316 @@
...
+        *,
+        thread: Optional[AgentThread] = None,
+        **kwargs: Any,
+    ) -> AgentRunResponse:
+        """Get a response from the agent.
+
+        Args:
+            messages: The messages to send to the agent.
+            thread: Optional thread to use for the conversation.
+            **kwargs: Additional options.
```
eavanvalkenburg, line 141: please make sure to note if and how kwargs are used

python/runtime/agent_runtime/agent_proxy.py
```diff
@@ -0,0 +1,316 @@
...
+        response = await handle.get_response()
+
+        if response.status == RequestStatus.COMPLETED:
+            # Deserialize response data
+            if response.data is None:
+                raise RuntimeError("Response data is None")
+
+            # Handle both dict and already deserialized AgentRunResponse
+            if isinstance(response.data, dict):
+                return AgentRunResponse(**response.data)
```
eavanvalkenburg, line 171:
```suggestion
                return AgentRunResponse.model_validate(response.data)
```

python/runtime/agent_runtime/agent_proxy.py
```diff
@@ -0,0 +1,316 @@
...
+            if isinstance(response.data, AgentRunResponse):
+                return response.data
+            # Try to deserialize from JSON if it's a string
+            if isinstance(response.data, str):
+                data = json.loads(response.data)
```
eavanvalkenburg, line 176: also wrapped in try and this:
```suggestion
                return AgentRunResponse.model_validate_json(response.data)
```

python/runtime/agent_runtime/agent_proxy.py
eavanvalkenburg, line 171: and wrap in try: ... except (pydantic.)ValidationError

python/runtime/agent_runtime/agent_proxy.py
```diff
@@ -0,0 +1,317 @@
...
+
+    @staticmethod
+    def _validate_id(thread_id: str) -> None:
+        """Validate that the thread ID matches the required pattern.
+
+        Thread IDs must be alphanumeric and can contain hyphens, underscores, dots, and tildes.
+        """
+        import re
+
+        pattern = r"^[a-zA-Z0-9_.\-~]+$"
```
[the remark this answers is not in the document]
adityamandaleeka: No, the hyphen is already escaped.

adityamandaleeka:
Thanks for all the feedback on this POC. I'm going to close this PR. It seems like we may be pursuing another runtime design approach which won't necessitate having the actor model expressed in Python at all.

PR #420, opened 2025-08-14
.NET: feat: Implement Checkpointing API

lokitoth:
Adds support for checkpointing active Runs and StreamingRuns

TODO:
- [x] Tests
- [x] Samples
- [x] Normalize Namespaces
- [x] One-class-per-file
- [x] XMLDoc comments for all publics

### DIFF
--- a/dotnet/src/Microsoft.Agents.Workflows/Checkpointed.cs
+++ b/dotnet/src/Microsoft.Agents.Workflows/Checkpointed.cs
@@ -41,15 +41,9 @@ public class Checkpointed<TRun>
     /// <summary>
     /// Gets the most recent checkpoint information.
     /// </summary>
-    public CheckpointInfo? LastCheckpoint => this.Checkpoints[this.Checkpoints.Count];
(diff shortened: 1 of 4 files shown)


dotnet/src/Microsoft.Agents.Workflows/CheckpointManager.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Threading.Tasks;
+using Microsoft.Agents.Workflows.Checkpointing;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Agents.Workflows;
+
+/// <summary>
+/// An in-memory implementation of <see cref="ICheckpointManager"/> that stores checkpoints in a dictionary.
+/// </summary>
+public sealed class CheckpointManager : ICheckpointManager
```
ekzhu, line 13: "CheckpointManager" in .NET vs. "CheckpointStorage" in Python -- I think we should align the names,.

lokitoth: Makes sense. I'll fix the .NET name in upcoming PR.

PR #422, opened 2025-08-14
.NET: ADR to support long-running executions

SergeyMenshykh:
### Motivation and Context
Chat Clients and AI Agents need support long-running executions to accommodate a variety of asynchronous scenarios that require processing time, such as code generation and analysis tasks, complex reasoning and research operations, image and content generation, and large document processing and summarization.

This ADR outlines the initial assumptions about how this support can be added to chat clients and AI agents. The described options will be refined and improved based on feedback and prototyping results.

### DIFF
--- a/docs/decisions/000N-support-long-running-execution.md
+++ b/docs/decisions/000N-support-long-running-execution.md
@@ -1396,6 +1396,15 @@ public class ChatOptions
 
 **Pro:** No dependency on the `System.ClientModel` package.
 
+##### 6.1.5 Continuation Token of a Custom Type
+
+The option is similar the the "6.1.3 Continuation Token of System.ClientModel.ContinuationToken Type" option but suggests using a 
+custom type for the continuation token instead of the `System.ClientModel.ContinuationToken` type.
+
+**Pros**
+- There is no dependency on the `System.ClientModel` package.   
+- There is no ambiguity between extension methods for `IChatClient` that would occur if a new extension method, which accepts a continuation token of string type as the first parameter, is added.
+
 #### 6.2 Overloads of GetResponseAsync and GetStreamingResponseAsync
 
 This option proposes introducing overloads of the `GetResponseAsync` and `GetStreamingResponseAsync` methods that will accept long-running operation parameters directly:
@@ -1685,7 +1694,7 @@ AgentRunResponse response = await agent.RunAsync("<prompt>", new AgentRunOptions
  - Option **2.2 Execution Mode per `Get{Streaming}ResponseAsync` Invocation + Model Class** is selected.
  - Option **3.2: One Method to Get Status and Result** is selected.
  - Option **4.2. As Properties Of ChatResponse** is selected.
- - Option **6.1.4 Continuation Token of String Type** is selected.
+ - Option **6.1.5 Continuation Token of a Custom Type** is selected.
 
 ### Long-Running Operations Support for AF Agents
  - Options **1.1 Run{Streaming}Async Methods for Common Operations and the Update Operation + New Method Per Uncommon Operation** is selected.


docs/decisions/000N-support-long-running-execution.md
```diff
@@ -0,0 +1,1106 @@
...
+  - **Start Long-Running Execution**: Initiates a long-running execution and returns an Id of the execution.
+  - **Get Status of Long-Running Execution**: This method is used to retrieve the status of a long-running execution.
+  - **Get Result of Long-Running Execution**: Retrieves the result of a long-running execution.
+- Uncommon operations:
+  - **Update Long-Running Execution**: This method is used to update a long-running execution, such as adding new messages or modifying existing ones.
+  - **Cancel Long-Running Execution**: This method is used to cancel a long-running execution that is still in progress.
+  - **Delete Long-Running Execution**: This method is used to delete a long-running execution, typically after it has completed or been cancelled.
+
+To support these operations by `IChatClient` implementations, the following options are available:
+- **1.1 New IAsyncChatClient Interface For All Long-Running Execution Operations**
```
eavanvalkenburg: I'm not a fan of the use of the word Async for this, even though I understand that it does mean that, it also means that someone has to `await` the call, and you can see below how ugly that becomes with the `startAsyncRunAsync` method names, long-running would be better, maybe future or background

docs/decisions/000N-support-long-running-execution.md
```diff
@@ -0,0 +1,1106 @@
...
+}
+````
+
+The streaming API may return an UpdateId identifying a particular update within a streamed response. 
+This UpdateId should be available together with RunId to callers, allowing them to resume a long-running execution identified 
+by the RunId from the last received update, identified by the UpdateId.
+TBD: Explore this further if the option is selected.
+
+### Option 4.2. As Properties Of ChatResponse
+TBD
```
eavanvalkenburg: I think `RunStatus` as part of ChatReponse makes a lot of sense, especially since it can just be `Completed` for non-long running jobs, and you can then easily write:

```python
response = await chat_client.get_response(input)
while response.status != 'completed':
    response = await chat_client.get_response(response.id)
 print(response)
```

and this will work anytime, regardless of capabilties of the service used.

docs/decisions/000N-support-long-running-execution.md
SergeyMenshykh: Right, I agree. It's a draft name to start the conversation from. It will be renamed later.

docs/decisions/000N-support-long-running-execution.md
```diff
@@ -0,0 +1,1106 @@
...
+- Common operations:
+  - **Start Long-Running Execution**: Initiates a long-running execution and returns an Id of the execution.
+  - **Get Status of Long-Running Execution**: This method is used to retrieve the status of a long-running execution.
+  - **Get Result of Long-Running Execution**: Retrieves the result of a long-running execution.
+- Uncommon operations:
+  - **Update Long-Running Execution**: This method is used to update a long-running execution, such as adding new messages or modifying existing ones.
+  - **Cancel Long-Running Execution**: This method is used to cancel a long-running execution that is still in progress.
+  - **Delete Long-Running Execution**: This method is used to delete a long-running execution, typically after it has completed or been cancelled.
+
+To support these operations by `IChatClient` implementations, the following options are available:
```
DeagleGross, line 379: Why is the discussion here going around `IChatClient` interface -> it is only one of many possible clients in the actor world. I think this doc should discuss the API built around [`IActorClient`](https://github.com/microsoft/agent-framework/blob/c0c49d31d031a795460ba0a2b27f0dd240d15717/dotnet/src/Microsoft.Extensions.AI.Agents.Runtime.Abstractions/IActorClient.cs#L11) or [`AIAgent`](https://github.com/microsoft/agent-framework/blob/c0c49d31d031a795460ba0a2b27f0dd240d15717/dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIAgent.cs#L16), because those are foundational entities in agent-framework today to operate with the agent.

Maybe I dont correlate the [dotnet/extension's Microsoft.Extensions.AI.Abstractions](https://github.com/dotnet/extensions/tree/92cce3081c5a972210c0667ef3698079f9f9324f/src/Libraries/Microsoft.Extensions.AI.Abstractions) and this agent-framework, but why dont we propose a new API there then?

SergeyMenshykh: Support for long-running execution is needed for Chat Clients and AI Agents. This ADR has a section for each. The section for agents, which includes an initial design suggestion that will be refined later, is shown below.

adityamandaleeka: I think from a layering perspective, it makes sense to keep the long-running operation support at the agent/workflow level rather than having it on IChatClient. Is there a reason this needs to be at this layer?

cc @ekzhu this is the layering thing we were discussing today.

SergeyMenshykh: The plan is to support long-running executions with both chat clients and AI agents.   
  
Supporting long-running executions on the chat client level allows the [ChatClientAgent](https://github.com/microsoft/agent-framework/blob/main/dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatClientAgent.cs) to route these executions to the underlying chat client.
  
Supporting this feature on the AI agent level allows AI agents that don't use chat clients to have this capability if the underlying API supports it.

CC: @stephentoub

stephentoub: > I think from a layering perspective, it makes sense to keep the long-running operation support at the agent/workflow level rather than having it on IChatClient. Is there a reason this needs to be at this layer?

How come? Services like responses have resumption baked in at the core protocol level, with messages getting tagged with IDs and being able to later come back and ask to resume a prior-initiated stream. Why wouldn't we want the lowest layer to be able to expose that concept?

adityamandaleeka: I guess I was wondering if this is a place where we can choose to be opinionated and say "if you care about long-running executions then you should use these specific framework concepts (e.g. workflows)".

stephentoub: > "if you care about long-running executions then you should use these specific framework concepts (e.g. workflows)".

But what about long-running constituent operations, e.g. where I call out to another service exposing a Responses surface area? That might be part of a workflow or might be separate.

SergeyMenshykh:
Closing this PR because the ADR was introduced in this PR: https://github.com/microsoft/agent-framework/pull/1501.

Issue #304, opened 2025-08-01
Threads: Investigate thread ids from a hosted agent perspective


westey-m:
We support storing thread messages in various places, including the agent services, in the thread itself via (InMemoryMesageStore) and in custom remote locations via IChatMessageStore.
Some of these have stores come with their own ids (in agent threads), some do not (in memory storage).

On AgentThread we have first party support for the id of the agent service managed thread via ConversationId.
As a custom agent service builder, you may want to expose that Id to your consumers as *the* thread id for your service, but if you are using InMemory messages storing, you don't have an equivalent id.  We cannot just update the ConversationId property to some GUID when using InMemory, since ChatClientAgent uses ConversationId to mean that there is an agent service owned thread, and will start passing this thread id to the IChatClient target service incorrectly.

We should investigate having a separate thread id property for the purpose of having a public facing id, that can be different from ConversationId, but also the same if required and there is a Conversationid.

PR #430, opened 2025-08-15
ADR: Agent Thread IDs and Message Storage

westey-m:
### Motivation and Context

This ADR explores and documents the different types of IDs required when using agent threads.

#304

### Description

The ADR explores some potential changes compared to the current AgentThread:
- Support for a public agent thread id
- Support for using this public id for 3rd party storage.

### Contribution Checklist


docs/decisions/00NN-agent-thread-storage-and-ids.md
```diff
@@ -0,0 +1,256 @@
...
+
+1. **AgentService**: The messages are stored by the agent/inference service being used by the agent.
+2. **InMemory**: The messages are stored in-memory and are transient.
+3. **3rdParty**: The messages are stored in a 3rd party storage solution, e.g. Cosmos DB.
+
+With Scenario 1 the storage of messages is built into the underlying service, and the service typically only supports storage of messages it its own storage.
+
+Where services do not have built in storage (1), the caller typically has to provide all chat history on each invocation.
+The messages making up this chat history, can be stored either in memory locally or in some 3rd party storage location.
+The service is agnostic to where the messages are stored, and any such 3rd party storage needs to be provided by the agent framework.
```
dmytrostruk: > and any such 3rd party storage needs to be provided by the agent framework.

Do you mean here that messages need to be provided by the Agent Framework to the service and not the actual 3rd party storage? Or does it mean that Agent Framework should support many 3rd party storages out-of-the-box?

docs/decisions/00NN-agent-thread-storage-and-ids.md
```diff
@@ -0,0 +1,256 @@
...
+{
+    Id = "thread-1",
+    AdditionalProperties = new Dictionary<string, string>
+    {
+        { "response-id", "r-12345" }
+    }
+}
+```
+
+#### Option 2 - Public Id has concrete property, service thread id has concrete property
```
dmytrostruk: I like this option, because it has a strongly typed `AgentThread.ServiceThreadId`, which looks like is going to be useful in lots of cases, so with a separate property we just eliminate the need to work with `AdditionalProperties` dictionary all the time. We just need to clearly document the purpose if each property.

docs/decisions/00NN-agent-thread-storage-and-ids.md
```diff
@@ -0,0 +1,256 @@
...
+This is a variation of Option 1 where the `Id` of the `AgentThread` is passed to the `MessageStore` when
+messages are added or retrieved, so that messages can be stored under this id.
+
+In this case, a 3rd party message store would not need to generate its own ids for messages, as it can use the thread id provided by the `AgentThread`,
+however it is still free to generate its own ids.
+
+```csharp
+public interface IChatMessageStore
+{
+    Task<IEnumerable<ChatMessage>> GetMessagesAsync(string threadId, CancellationToken cancellationToken);
```
dmytrostruk: Option 1.1 makes sense to me, although it feels like `threadId` parameter here should be optional, because some message stores may simply ignore that, because they don't need it. Like `InMemoryChatMessageStore` - for simple cases we don't need a `threadId`, since we just store plain messages and for new chat we just create another thread instance with new in-memory messages collection.

I would probably replace `threadId` parameter with some `Options` type here, so we can put `ThreadId` property inside, which would be optional, and it will be easier to add more parameters later without breaking changes.

docs/decisions/00NN-agent-thread-storage-and-ids.md
```diff
@@ -0,0 +1,256 @@
...
+    Task AddMessagesAsync(string threadId, IReadOnlyCollection<ChatMessage> messages, CancellationToken cancellationToken);
+}
+```
+
+#### Option 2 - Store in-memory messages the thread, with an optional IChatMessageStore for 3rd party integrations
+
+The AgentThread class has an optional MessageStore property for storing messages.
+The AgentThread also has a ChatHistory property for storing messages in-memory.
+Only one of these can be used at a time, and if an Agentic/InferenceService thread id is available,
+both the ChatHistory and MessageStore would be ignored.
```
dmytrostruk, line 367: I think this makes Option 2 a little bit complicated because if users don't know which agent service they are using behind the abstraction (e.g. in different environments they use different agent services and message storages), they will need to write the code to check both `ChatHistory` or `MessageStore` to get the messages, which is more complicated. 

The option when we just have `MessageStore` looks simpler from API usage point of view, because we can write `MessageStore.GetMessages` and in case of in-memory messages we just return it, and in case of 3rd party storage - it will perform a request to that storage and return it, so no need in checking multiple properties, writing `if/else` blocks etc.

docs/decisions/00NN-agent-thread-storage-and-ids.md
westey-m: I don't really know what I was trying to say there either, so I'll just remove the sentence. :)

docs/decisions/00NN-agent-thread-storage-and-ids.md
westey-m: > so with a separate property we just eliminate the need to work with AdditionalProperties dictionary all the time

Not necessarily all the time. Some agents may use multiple agent threads internally, in which case this property doesn't really make sense.  Although we may say that that's ok to gain the benefit of having this for the common case where a single service thread is used.

I am also adding another sub classing option.

docs/decisions/00NN-agent-thread-storage-and-ids.md
westey-m: > although it feels like threadId parameter here should be optional, because some message stores may simply ignore that

If it's optional, what would message stores that require it do, if it's not provided?

docs/decisions/00NN-agent-thread-storage-and-ids.md
westey-m: Yeah, I also don't like Option 2 that much.  I'm including it since I explored it previously, and it's good to have options, but I definitely prefer Option 1.1.

Issue #450, opened 2025-08-20
.NET: Agent Discovery in the framework


DeagleGross:
The goal of the issue is to design, document and implement the agent discovery feature in the agent framework. 

Agent-framework should have a set of APIs which would help provide a determined and strongly-typed structure representing which agents user-application supports and provides the metadata about each individual agent. It should be configurable to be either explicitly defined by user in-code or provided via separate configuration (or other potential ways).

Also in case user application is a web application, agent-framework should provide a convenient API to expose well-known/documented endpoint(s) which can be used by consumers to discover agents and their capabilities.

DeagleGross:
Prototype is decided to be postponed right inside of the agent-framework; We will build it later in a bit different form. Stay tuned

DeagleGross:
Closed as no decision done to work on this

PR #454, opened 2025-08-20
.NET: [WIP] Design Proposal: Agent Discovery

DeagleGross:
Proposing a design for agent-discovery in the agent-framework as a first-class feature.

_prototype can be viewed here: #692_

Relates #450

### DIFF
--- a/docs/decisions/0008-agent-discovery.md
+++ b/docs/decisions/0008-agent-discovery.md
@@ -118,7 +118,7 @@ Questions:
 }
 ```
 
-## Communication endpoints `/actors/v1/{actorType}/{actorKey}/{messageId}`
+## Communication endpoints `/actors/v1/{actorType}/{actorKey}/messages/{messageId}`
 
 Agent Discovery is only good if there is an established way to communicate to the agents. Without an implemented in-place endpoint for sending/retrieving messages (for actor and conversation) users can build too different schemas, and make the client-side work unbearable for multiple different apps.
 


docs/decisions/0005-agent-discovery.md
```diff
@@ -0,0 +1,199 @@
...
+
+Questions:
+- do we want to initially support non-HTTP requests? should be define more complex object in endpoints?
+```json
+{
+    "transport": "HTTP",
+    "httpMethod": "GET",
+    ... 
+}
+```
```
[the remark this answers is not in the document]
DeagleGross: showing that this a part of JSON

docs/decisions/0005-agent-discovery.md
```diff
@@ -0,0 +1,199 @@
...
+
+Questions:
+- do we want to initially support non-HTTP requests? should be define more complex object in endpoints?
+```json
+{
+    "transport": "HTTP",
+    "httpMethod": "GET",
+    ...
```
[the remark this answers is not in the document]
DeagleGross: a prt of json

docs/decisions/0005-agent-discovery.md
```diff
@@ -0,0 +1,199 @@
...
+{
+    "transport": "HTTP",
+    "httpMethod": "GET",
+    ... 
+}
+```
+
+## Minimal Data Model (AgentMetadata v1)
+
+Top-level fields:
```
ekzhu: It's useful to list out the tools/functions the agent has access to following the format used by MCP. I think going with more structured/schematized metadata is better as allows for querying.

Also, we need to include the model ID used by the agent, if available. I believe it is arguably much more accurate evidence of capability than a text description.

docs/decisions/0005-agent-discovery.md
```diff
@@ -0,0 +1,199 @@
...
+- id: stable, path-safe identifier (string)
+- name: human-readable name (string)
+- version: semantic version (string)
```
ekzhu: I think we should also be able to get the instruction used by each agent. Transparency is key right now as consumer of this agent metadata need to believe the description is accurate. 

Generally speaking I think we should cover the initialization parameters for each agent in their metadata -- name, model, tools, and instructions. I don't think we are at the point when people actually trust the annotated metadata of agents in a catalog unless there are other signals like rating and usage.

DeagleGross: I've added `instructions` and `model` to the agent metadata and refactored the prototype. Please take a look at new version

docs/decisions/0005-agent-discovery.md
DeagleGross: added model / instructions fields.

docs/decisions/0008-agent-discovery.md
```diff
@@ -0,0 +1,167 @@
...
+deciders: adityam, reubenbond, sergeymenshykh, markwallace, rbarreto, dmytrostruk, westey-m, eavanvalkenburg, stephentoub
+consulted: 
+informed: 
+---
+
+# Agent Discovery: Simplified App-Scoped Design
+
+## Summary
+
+This document proposes a minimal, app-scoped discovery API for listing agents and fetching metadata for a single agent.
```
victordibia, line 15: This is looking good ⁠ @DeagleGross 

I can see this being used across multiple surfaces, especially a local dev experience like [devui](https://github.com/microsoft/agent-framework/tree/victordibia/devui/python/packages/devui)



Right now agent framework supports two high level abstractions - **agents** and **workflows**. 

Does this discovery mechanism natively support workflows (this seems quiate important as most of the real multi-agent and real world application logic are workflows). 

Currently devui implements an "[entity discovery](https://github.com/microsoft/agent-framework/tree/victordibia/devui/python/packages/devui#key-endpoints)" endpoint and it would be good to replace that with the logic we settle on here.

DeagleGross: At the moment workflows are not covered. We can add support for them later I suppose?

docs/decisions/0006-agent-discovery.md
```diff
@@ -0,0 +1,167 @@
...
+    "truth": {
+      "type": "chat",
+      "instructions": "tell only truth",
+      "model": "gpt4.1"
+    }
+  }
+}
+```
+
+## Communication endpoints `/actors/v1/{actorType}/{actorKey}/{messageId}`
```
ReubenBond: IMO we should have something here indicating that the resource is a message resource, eg add `/messages/` before `{messageId}`. You could imagine wanting to access other things, like state, stats, description, etc, so it's best that we can differentiate between these things.

docs/decisions/0008-agent-discovery.md
```diff
@@ -0,0 +1,167 @@
...
+```json
+{
+  "agents": [
+    {
+      "name": "pirate",
+      "description": "Agent which talks like a pirate",
+      "version": "1.4.2",
+      "instructions": "Talk like a pirate",
+      "model": "gpt-4o",
+      "metadata": {
```
luisquintanilla, line 73: Consider adding tools / skills / capabilities (if any) as part of metadata

docs/decisions/0008-agent-discovery.md
```diff
@@ -0,0 +1,167 @@
...
+  "method": "run",
+  "params": [
+    { "$type": "text", "content": "hey matey!" }
+  ]
+}
+```
+
+## Versioning and Content Negotiation
+
+- Protocol versioning is encoded in the URL path as /vX beneath the reserved prefix. Breaking changes will increment X (e.g., /agents/v1 -> /agents/v2). Additive, backwards-compatible changes MAY be introduced within a major version without changing X.
```
luisquintanilla, line 143: How would the versioning of this protocol map to underlying Reponses, A2A, MCP protocols.

docs/decisions/0008-agent-discovery.md
DeagleGross: I dont think at this moment agent framework has the "tools" concept on the agent definition level. Lets add this in next version or when this feature will be placed on the agent (not a breaking change if adding a new field to the object)

docs/decisions/0008-agent-discovery.md
DeagleGross: "discovery protocol" (hard to call it a full protocol, but it will be an established API) has a separate versioning of responses/a2a/mcp.

For example in v1 of discovery we can add only A2A, and then in v2 of discovery Responses will also be added to the metadata of agent (if any). And versions of protocols (like A2A or responses) can be added with specific version to the protocol / or even many versions of protocols to a single metadata. For example:
```json
{ // agent object
   "name": "pirate",
   "a2a":
   {
      "v1": { ... },
      "v2": { ... }
   }
}
```

docs/decisions/0006-agent-discovery.md
DeagleGross: agree, added the route component: `/actors/v1/{actorType}/{actorKey}/messages/{messageId}`

DeagleGross:
Decided to proceed with Agent Discovery later; and probably in a bit different component. We will revisit this and definitely build.

Issue #295, opened 2025-07-31
Add support for thread reduction (truncation/summarization)


westey-m:
We need to allow developers to provide thread truncation or summarization capabilities.
This feature is scoped to cases where we have access to the thread contents locally and in a standardized format, e.g. when using InMemoryMessageStore.

If the thread is stored somewhere else, the options open to developers will vary depending on the store location:

1. If the thread is managed by the agent service, it is expected that the agent service manages the thread, and it is not necessarily modifiable, meaning developers do not have the option of adding reduction.
2. If the thread is stored in a remote database via a custom message store, developers will need to implement their own mechanism around this, e.g. only retrieving the most recent messages on invocation (truncation) or searching the database for older messages, combining them and updating the database with the new messages while deleting the old ones.

stephentoub:
cc: @MackinnonBuck

crickman:
Have you seen this update: https://github.com/dotnet/extensions/pull/6666 ?

westey-m:
> Have you seen this update: [dotnet/extensions#6666](https://github.com/dotnet/extensions/pull/6666) ?

Thanks @crickman, yes, my PR 462 is making use of the types added in PR 6666

PR #462, opened 2025-08-21
.NET: Add support for IChatReducer to InMemoryChatMessageStore

westey-m:
### Motivation and Context

#295

Sample to follow if we are happy with this approach.

### Description

- Update InMemoryChatMessageStore to optionally support IChatReducer
- Make InMemoryChatMessageStore public, so that users can create it with a chat reducer and provide it to the agentthread factory on ChatClientAgent.

### Contribution Checklist


crickman:
Have you seen this update: https://github.com/dotnet/extensions/pull/6666 ?

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/MEAI/IChatReducer.cs
```diff
@@ -0,0 +1,23 @@
...
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>
+/// Represents a reducer capable of shrinking the size of a list of chat messages.
+/// </summary>
+public interface IChatReducer
```
markwallace-microsoft: Should we name is `IConversationReducer` in line with the move to using that term?

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/InMemoryChatMessageStore.cs
```diff
@@ -30,17 +61,32 @@ public ChatMessage this[int index]
...
     }
 
     /// <inheritdoc />
-    public Task<IEnumerable<ChatMessage>> GetMessagesAsync(CancellationToken cancellationToken)
+    public async Task<IEnumerable<ChatMessage>> GetMessagesAsync(CancellationToken cancellationToken)
     {
-        return Task.FromResult<IEnumerable<ChatMessage>>(this._messages);
+        if (this._reducerTriggerEvent == ChatReducerTriggerEvent.BeforeMessagesRetrieval && this._chatReducer is not null)
+        {
+            this._messages = (await this._chatReducer.ReduceAsync(this._messages, cancellationToken).ConfigureAwait(false)).ToList();
```
markwallace-microsoft, line 86: Why do we need to reduce when adding and retrieving?

westey-m: You can only pick one of the two, not both.  Are you thinking it's better to not let the user pick and us pick for them?

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/MEAI/IChatReducer.cs
westey-m: Note that this is just a copy from MEAI, and it has already shipped as part of MEAI, albeit as experimental.
If we want to change the naming, best to recommend there.
There is a PR open at the moment to move this to MEAI.A, where you could add feedback:
https://github.com/dotnet/extensions/pull/6738

stephentoub:
cc: @MackinnonBuck

PR #464, opened 2025-08-21
.NET: Long-running executions POC

SergeyMenshykh:
This is a draft PR that prototypes support for long-running executions by chat clients and AI agents.

### DIFF
--- a/dotnet/src/Microsoft.Extensions.AI.Agents.OpenAI/NewOpenAIResponsesChatClient.cs
+++ b/dotnet/src/Microsoft.Extensions.AI.Agents.OpenAI/NewOpenAIResponsesChatClient.cs
@@ -74,30 +74,29 @@ internal sealed class NewOpenAIResponsesChatClient : IChatClient, ICancelableCha
     public async Task<ChatResponse> GetResponseAsync(
         IEnumerable<ChatMessage> messages, ChatOptions? options = null, CancellationToken cancellationToken = default)
     {
-        _ = Throw.IfNull(messages);
+        var inputMessages = Throw.IfNull(messages) as IReadOnlyCollection<ChatMessage> ?? [.. messages];
 
-        // Convert the inputs into what OpenAIResponseClient expects.
-        var openAIResponseItems = ToOpenAIResponseItems(messages, options);
         var openAIOptions = ToOpenAIResponseCreationOptions(options);
 
-        OpenAIResponse openAIResponse;
-
-        // The continuation token, provided by a caller, indicates that the caller is interested in
-        // the result of the ongoing operation rather than in creating a new one.
+        // The continuation token provided by a caller indicates that the caller is interested
+        // in the result of the ongoing operation rather than in creating a new one.
         if (options is NewChatOptions { ContinuationToken: { } token } && ResponsesLongRunResumptionToken.FromToken(token) is { } longRunToken)
         {
-            if (messages.Any())
+            if (inputMessages.Count > 0)
             {
                 throw new InvalidOperationException("Messages are not allowed when using a continuation token.");
             }
 
-            openAIResponse = await _responseClient.GetResponseAsync(longRunToken.ResponseId, cancellationToken).ConfigureAwait(false);
+            var response = await _responseClient.GetResponseAsync(longRunToken.ResponseId, cancellationToken).ConfigureAwait(false);
+
+            return FromOpenAIResponse(response, openAIOptions);
         }
-        else
-        {
-            // Otherwise, create a new response.
-            openAIResponse = await _responseClient.CreateResponseAsync(openAIResponseItems, openAIOptions, cancellationToken).ConfigureAwait(false);
-        }
+
+        // Convert the inputs into what OpenAIResponseClient expects.
+        var openAIResponseItems = ToOpenAIResponseItems(inputMessages, options);
+
+        // Create a new response.
+        OpenAIResponse openAIResponse = await _responseClient.CreateResponseAsync(openAIResponseItems, openAIOptions, cancellationToken).ConfigureAwait(false);
 
         // Convert the response to a ChatResponse.
         return FromOpenAIResponse(openAIResponse, openAIOptions);
@@ -255,30 +254,29 @@ internal sealed class NewOpenAIResponsesChatClient : IChatClient, ICancelableCha
     public IAsyncEnumerable<ChatResponseUpdate> GetStreamingResponseAsync(
         IEnumerable<ChatMessage> messages, ChatOptions? options = null, CancellationToken cancellationToken = default)
     {
-        _ = Throw.IfNull(messages);
+        var inputMessages = Throw.IfNull(messages) as IReadOnlyCollection<ChatMessage> ?? [.. messages];
 
-        var openAIResponseItems = ToOpenAIResponseItems(messages, options);
         var openAIOptions = ToOpenAIResponseCreationOptions(options);
 
-        IAsyncEnumerable<StreamingResponseUpdate> streamingUpdates;
-
-        // The continuation token, provided by a caller, indicates that the caller is interested in
-        // the result of the ongoing operation rather than in creating a new one.
+        // The continuation token provided by a caller indicates that the caller is interested
+        // in the result of the ongoing operation rather than in creating a new one.
         if (options is NewChatOptions { ContinuationToken: { } token } && ResponsesLongRunResumptionToken.FromToken(token) is { } longRunToken)
         {
-            if (messages.Any())
+            if (inputMessages.Count > 0)
             {
                 throw new InvalidOperationException("Messages are not allowed when using a continuation token.");
             }
 
             // If response id is provided, get the response by id.
-            streamingUpdates = _responseClient.GetResponseStreamingAsync(longRunToken.ResponseId, longRunToken.SequenceNumber, cancellationToken);
+            IAsyncEnumerable<StreamingResponseUpdate> updates = _responseClient.GetResponseStreamingAsync(longRunToken.ResponseId, longRunToken.SequenceNumber, cancellationToken);
+
+            return FromOpenAIStreamingResponseUpdatesAsync(updates, openAIOptions, cancellationToken);
         }
-        else
-        {
-            // Otherwise, create a new response.
-            streamingUpdates = _responseClient.CreateResponseStreamingAsync(openAIResponseItems, openAIOptions, cancellationToken);
-        }
+
+        var openAIResponseItems = ToOpenAIResponseItems(messages, options);
+
+        // Create a new response.
+        IAsyncEnumerable<StreamingResponseUpdate> streamingUpdates = _responseClient.CreateResponseStreamingAsync(openAIResponseItems, openAIOptions, cancellationToken);
 
         return FromOpenAIStreamingResponseUpdatesAsync(streamingUpdates, openAIOptions, cancellationToken);
     }


dotnet/src/Microsoft.Extensions.AI.Agents.OpenAI/NewOpenAIResponsesChatClient.cs
```diff
@@ -83,16 +86,31 @@ public async Task<ChatResponse> GetResponseAsync(
 
         OpenAIResponse openAIResponse;
 
-        // If conversation id is provided and background mode is enabled, retrieve status or/and result of the response running in the background.
-        if (options?.ConversationId is { } conversationId && RunInBackground(options))
+        if (options?.GetPreviousResponseId() is { } prevResponseId)
         {
-            openAIResponse = (await _responseClient.GetResponseAsync(conversationId, cancellationToken).ConfigureAwait(false)).Value;
+            // If previous response id is provided, fetch the response by id.
+            openAIResponse = (await _responseClient.GetResponseAsync(prevResponseId, cancellationToken).ConfigureAwait(false)).Value;
```
peterychang: are we overloading the `prevResponseId` parameter for the long running request ID?

SergeyMenshykh: It's a new parameter that needs to be added to ChatOptions to indicate that a caller wants to get the status/result of a response rather than create a new response. Without it, it's impossible to identify the caller's intention.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIAgent.cs
```diff
@@ -261,6 +261,30 @@ public abstract IAsyncEnumerable<AgentRunResponseUpdate> RunStreamingAsync(
...
+    }
+
+    /// <summary>
+    /// Deletes a long-running operation.
+    /// </summary>
+    /// <param name="id">The unique identifier of the long-running operation to delete.</param>
+    /// <param name="options">Optional parameters for the long-running operation deletion.</param>
+    /// <param name="cancellationToken">A cancellation token that can be used to delete the operation.</param>
+    /// <returns>The <see cref="AgentRunResponse"/> representing result of the deletion if supported; otherwise, <see langword="null"/>.</returns>
+    public virtual Task<AgentRunResponse?> DeleteRunAsync(string id, AgentRunDeleteOptions? options = null, CancellationToken cancellationToken = default)
```
stephentoub: What is the difference between "cancel" and "delete" here? If it's about the actual underlying storage, I thought we were keeping management of the underlying system out of AIAgent and leaving that for a separate management mechanism from the underlying SDK?

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AgentRunOptions.cs
```diff
@@ -24,4 +24,20 @@ public AgentRunOptions(AgentRunOptions options)
...
+    /// <summary>
+    /// Specifies the identifier of an update within a conversation to start generating chat responses after.
+    /// </summary>
+    public string? StartAfter { get; set; }
+
+    /// <summary>
+    /// Specifies identifier of either a long-running operation or an identifier
+    /// of an entity representing a long-running operation.
+    /// </summary>
+    public string? ResponseId { get; set; }
```
stephentoub: Both ResponseId and StartAfter are necessary? There's no generalized way of saying "the last thing I have is X, please continue from there"?

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/NewChatOptions.cs
```diff
@@ -0,0 +1,74 @@
...
+/// Contains new properties that will be added to <see cref="ChatOptions"/> in the future.
+/// </summary>
+/// <remarks>
+/// This class contains temporary properties to support new options
+/// that are not part of the official <see cref="ChatOptions"/> class yet.
+/// Later, these properties will be moved to the official <see cref="ChatOptions"/> class
+/// and this class will be removed. Therefore, please expect a breaking change
+/// if you are using this class directly in your code.
+/// </remarks>
+public class NewChatOptions : ChatOptions
```
stephentoub, line 17: I realize we're not actually going to ship this, but for the (hopefully rare?) case where someone does want to derive from ChatOptions to add new members, we should probably look at exposing a protected Clone method that contains the bulk of the logic and accepts the instance to write to, or a CloneFrom that accepts the instance to read from.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/NewChatResponse.cs
```diff
@@ -0,0 +1,20 @@
...
+/// Later, these properties will be moved to the official <see cref="ChatResponse"/> class, and
+/// this class will be removed. Therefore, please expect a breaking change if you are using
+/// this class directly in your code.
+/// </remarks>
+public class NewChatResponse : ChatResponse
+{
+    /// <summary>
+    /// Specifies the status of the response.
+    /// </summary>
+    public NewResponseStatus? Status { get; set; }
```
stephentoub: This makes me a little nervous... I don't want folks to suddenly need to start checking an additional status. Is this only relevant when opting-in to new behaviors where a need to check status would be obvious?

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/NewResponseStatus.cs
```diff
@@ -0,0 +1,149 @@
...
+
+    /// <summary>
+    ///  Gets the <see cref="NewResponseStatus"/> that represents an operation that is incomplete or not fully processed.
+    /// </summary>
+    public static NewResponseStatus Incomplete { get; } = new("Incomplete");
+
+    /// <summary>
+    /// Gets the <see cref="NewResponseStatus"/> that represents an operation that is currently being cancelled.
+    /// </summary>
+    public static NewResponseStatus Cancelling { get; } = new("Cancelling");
```
stephentoub: Nit: one `l` please :)

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/RunId.cs
```diff
@@ -0,0 +1,62 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>
+/// Represents the unique identifier for a long-running operation.
+/// </summary>
+public sealed partial class RunId
```
stephentoub: Why is this needed?

dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ILongRunningChatClient.cs
```diff
@@ -0,0 +1,38 @@
...
+/// Alternative names suggested by AI:
+///     ILongRunningChatClient  - Emphasizes the long-running nature of the operations
+///     IAsyncChatClient        - Highlights the asynchronous, non-immediate execution model
+///     IPersistentChatClient   - Suggests operations that persist beyond immediate execution
+///	    IResumableChatClient    - Emphasizes the ability to resume/manage operations that can be paused or resumed
+///	    IBackgroundChatClient   - Indicates chat operations that run in the background and can be managed
+///	    IJobChatClient          - Treats operations as jobs that can be managed
+///	    ITaskChatClient         - Represents operations as tasks that can be managed
+/// </remarks>
+public interface ILongRunningChatClient : IChatClient
```
stephentoub: What is the thinking about the additional options on ChatOptions vs having dedicated methods here that accept those inputs as parameters on new methods? (Not pushing for one or the other, trying to understand the rationale)

dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ILongRunningChatClient.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// <returns>The <see cref="ChatResponse"/> representing result of the cancellation if supported; otherwise, <see langword="null"/>.</returns>
+    Task<ChatResponse?> CancelRunAsync(RunId runId, CancellationToken cancellationToken = default);
+
+    /// <summary>
+    /// Deletes a long-running chat operation identified by the specified <paramref name="runId"/>.
+    /// </summary>
+    /// <param name="runId">The unique identifier of the long-running operation to delete.</param>
+    /// <param name="cancellationToken">A cancellation token that can be used to cancel the operation.</param>
+    /// <returns>The <see cref="ChatResponse"/> representing result of the cancellation if supported; otherwise, <see langword="null"/>.</returns>
+    Task<ChatResponse?> DeleteRunAsync(RunId runId, CancellationToken cancellationToken = default);
```
stephentoub: I'm unsure about this one. I'm concerned it's getting too much into state management territory. What about listing all runs? What about creating new runs without actually pushing them forward? Etc. It's a slippery slope. I'm also not sure it's a concept that supported everywhere.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIAgent.cs
SergeyMenshykh: The delete operation is only supported by the OpenAI Responses API and deletes a response with a given id. The API also supports response cancellation, which cancels an existing response that was started with `background = true`. The deletion operation definitely looks to be related to managing underlying storage, so we might consider not including it in the long-running chat client/AI agent APIs.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AgentRunOptions.cs
SergeyMenshykh: The `ResponseId` identifies a long-running operation or an entity representing a long-running operation, e.g., a Task in the A2A protocol.

The `StartAfter` identifies a point in a long-running operation stream to start retrieving results from. It is only used in streaming scenarios and is only natively supported by OpenAI Responses, while Azure AI Agents APIs require additional handling to emulate its support. A2A does not support it, leaving the details to an A2A spec implementation. 
  
There are two options we can consider if we do not want to include the `StartAfter` property:    
  
1. Do not support getting results of a long-running operation from a specific point in its stream:
   a. Always return results from the point an underlying API provides them by default, or
   b. Return results from the beginning of the stream by calling the OpenAI Responses API with a `0` value for the `startAfter` parameter, streaming all steps from the first steps of Azure AI Agent API runs (which simplifies implementation since no emulation is needed). For A2A, however, we do not have control over the starting point, so we will return whatever the A2A agent provides.
  
   - Option 1.a will not work for Azure Foundry Agent because its API does not support streaming of runs identified by id. It is only possible to start a streaming run, not to get a streaming run by id. Currently, streaming support for a run is emulated by streaming results of non-streaming runs, so it is unclear what the default starting point should be.

   - Option 1.b allows us to avoid the problem of retrieving a long-running operation result from a point identifying function arguments of a function call in the stream. This problem is described here: [Function Calling](https://github.com/microsoft/agent-framework/blob/56ca3a98bd0a6bc1679a42103044ca2b926b32cc/docs/decisions/000N-support-long-running-execution.md#function-calling). However, getting results from the beginning of the stream may lead to function re-invocation, unless there is a mechanism to prevent this, if functions are called by the LLM, this could lead to unexpected results, such as the API throwing an exception when attempting to mutate(by sending function results back to API/LLM) a completed run/entity, or it could allow mutation, leading to different results than if the functions were not re-invoked.
 
1. Use the `ResponseId` property to identify both the operation/entity id and the point in the stream to continue from, by placing their values into the property value.
   We will need to agree on the property value format - whether it is a colon-separated string like `"res_dfer34354:2"` (where `res_dfer34354` is the response id and `2` is the start after point), or a serialized tuple of these values.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/NewChatResponse.cs
SergeyMenshykh: In the POC, the property is always assigned regardless of whether the new behavior is opted-in or not, and if the underlying API provides a value for it. It's very easy not to assign it if the behavior is not opted-in. I'll change this.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/RunId.cs
SergeyMenshykh: It's no longer needed. It was an attempt to model a run id from the response id or/and conversation id, or task id, but I modeled it differently and deleted this class.

dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ILongRunningChatClient.cs
SergeyMenshykh: ~Is the comment related to the new interface or to the method for deleting runs?~

We may not add this interface. If either operation is needed, it can be performed using the underlying SDK API. The interface or a different version of it might be added later if we see a clear need for it and believe it makes sense to include.

stephentoub: > Is the comment related to the new interface or to the method for deleting runs?

Deletion

> The interface or a different version of it might be added later if we see a clear need for it and believe it makes sense to include.

👍

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/NewChatResponse.cs
SergeyMenshykh: Fixed - the Status prop is assigned only if the new behavior is opted in, having a null value by default.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/NewResponseStatus.cs
SergeyMenshykh: I've done it for the Canceled status but missed it for this one. :) Fixed.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/NewChatOptions.cs
SergeyMenshykh: Yep, I thought about it as well when I was copying the content of the ChatOptions.Clone method to the new class. What do you think about a variant of the clone method that accepts a target type, like this one?https://github.com/microsoft/semantic-kernel/blob/5272beb956b415a3fad0445ab0531ec22dc59f23/dotnet/src/Connectors/Connectors.OpenAI/Settings/OpenAIPromptExecutionSettings.cs#L518C47-L518C76

stephentoub: How would that be implemented and used virtually?

SergeyMenshykh:
```csharp
// ----------------- usage -----------------
ChatOptions options = new DerivedChatOptions()
{
    AwaitLongRunCompletion = true,
    ResponseId = "responseId",
    StartAfter = "startAfter",
};

var clone = options.Clone();

// ----------------- definitions -----------------
public class ChatOptions
{
    public string? ConversationId { get; set; }
    public string? Instructions { get; set; }
    public float? Temperature { get; set; }

    public virtual ChatOptions Clone()
    {
        return this.Clone<ChatOptions>();
    }

    protected T Clone<T>() where T : ChatOptions, new()
    {
        return new()
        {
            ConversationId = ConversationId,
            Instructions = Instructions,
            Temperature = Temperature,
        };
    }
}

public class DerivedChatOptions : ChatOptions
{
    public bool? AwaitLongRunCompletion { get; set; }
    public string? StartAfter { get; set; }
    public string? ResponseId { get; set; }

    public override ChatOptions Clone()
    { 
        var clone = Clone<DerivedChatOptions>();

        clone.AwaitLongRunCompletion = AwaitLongRunCompletion;
        clone.StartAfter = StartAfter;
        clone.ResponseId = ResponseId;

        return clone;
    }
}
```

stephentoub: Thanks, but now imagine I want to add `MyDerivedChatOptions : DerivedChatOptions`. How do I implement its Clone override?

SergeyMenshykh: Got it. I've added an item to my to-do list to implement it later when I move the new/updated model to MEAI.

dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ILongRunningChatClient.cs
SergeyMenshykh: The main con of having new dedicated methods on this interface, regardless of whether they are for initiating/retrieving/canceling/deleting of long-running operations, is the fact that they can`t be decorated by existing decorators - function calling/logging/etc. - since they are part of the new interface the decorators don't/can't implement them. As a result, this may lead to a parallel set of decorators for "long-running" chat clients.

As an option, if we want to avoid adding additional properties to ChatOptions, we could pass them through an additional properties dictionary that would be initialized from IChatClient extension methods with those options as parameters in their signatures.

dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ILongRunningChatClient.cs
SergeyMenshykh: The delete method is removed.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIAgent.cs
SergeyMenshykh: As agreed, the delete methos is removed.

dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ILongRunningChatClient.cs
SergeyMenshykh: Agreed to explore possibility of using continuation token - https://github.com/microsoft/agent-framework/blob/91ae3ccfc86798023fc89ae6c80dafdb97f6f29d/docs/decisions/000N-support-long-running-execution.md#613-continuation-token

PR #475, opened 2025-08-22
.NET: Enhance console samples while preserving copyability

westey-m:
### Motivation and Context

We want our samples to be easy to use, while still keeping the ability to copy them without modifications.

### Description

- Override Environment.GetEnvironmentVariable to ask for a value if missing from environment variables, when running the sample in the demos folder.
- Reconfigure directory props and csproj files so that the csproj can be copied somewhere else and the code will still compile.

Making the samples truly standalone will also require us to replace project references with nuget packages for our own nuget packages, but there is no solution for this here yet.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Shared/Demos/SampleEnvironment.cs
+++ b/dotnet/src/Shared/Demos/SampleEnvironment.cs
@@ -12,7 +12,7 @@ internal static class SampleEnvironment
 {
     public static string? GetEnvironmentVariable(string key)
     {
-        // Allows for a demo mode, where all values are hidden for security reasons.
+        // Allows for opting into showing all setting values in the console output, so that it is easy to troubleshoot sample setup issues.
         var showAllSampleValues = SystemEnvironment.GetEnvironmentVariable("AF_SHOW_ALL_DEMO_SETTING_VALUES");
         var shouldShowValue = showAllSampleValues?.ToUpperInvariant() == "Y";
 


dotnet/demos/MinimalConsole/Program.cs
```diff
@@ -8,14 +8,17 @@
 using Microsoft.Extensions.AI.Agents;
 using OpenAI;
 
+var azureOpenAIEndpoint = Environment.GetEnvironmentVariable("AZURE_OPENAI_ENDPOINT");
```
rogerbarreto: nit.
We might just put the `!` at the variable set.

```suggestion
var azureOpenAIEndpoint = Environment.GetEnvironmentVariable("AZURE_OPENAI_ENDPOINT")!;
```

dotnet/demos/MinimalConsole/README.md
```diff
@@ -0,0 +1,30 @@
...
+```
+
+**Note**: This demo uses Azure CLI credentials for authentication. Make sure you're logged in with `az login` and have access to the Azure OpenAI resource.
+
+## Running the Demo
+
+```powershell
+cd dotnet/demos/MinimalConsole
+dotnet build
+dotnet run --no-build
```
rogerbarreto, line 29: We might just have dotnet run with build in a single command.
```suggestion
dotnet run
```

dotnet/demos/Directory.Build.props
rogerbarreto, line 1: We might just get rid of this file for samples and demos, it's not copy/paste friendly for the samples and may hide behaviors that we may not want,

dotnet/src/Shared/Demos/SampleEnvironment.cs
rogerbarreto, line 1: We might just get rid of this file for copy/paste convenience? I might have missed something but I didn't see in the samples this being used as well..

dotnet/demos/MinimalConsole/Program.cs
westey-m: I've updated to use a default for model and throw for endpoint if not provided.

dotnet/src/Shared/Demos/SampleEnvironment.cs
```diff
@@ -0,0 +1,139 @@
...
+using System;
+using System.Collections;
+using SystemEnvironment = System.Environment;
+
+namespace SampleHelpers;
+
+internal static class SampleEnvironment
+{
+    public static string? GetEnvironmentVariable(string key)
+    {
```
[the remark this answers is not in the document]
westey-m: Added a comment.

dotnet/src/Shared/Demos/SampleEnvironment.cs
```diff
@@ -0,0 +1,139 @@
...
+            Console.ForegroundColor = ConsoleColor.Yellow;
+            Console.Write("EnvironmentVariables");
+            Console.ForegroundColor = ConsoleColor.Green;
+            Console.Write(", Key='");
+            Console.ForegroundColor = ConsoleColor.Yellow;
+            Console.Write(key);
+            Console.ForegroundColor = ConsoleColor.Green;
+            Console.Write("', Value='");
+            Console.ForegroundColor = ConsoleColor.Yellow;
+            Console.Write(hideSampleValues?.ToUpperInvariant() == "Y" || key.ToUpperInvariant().EndsWith("KEY", StringComparison.InvariantCulture) ? "*****" : value);
```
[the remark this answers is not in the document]
westey-m: Moved to a separate variable.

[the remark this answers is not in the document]
westey-m: Moved to a separate variable.

dotnet/demos/MinimalConsole/README.md
westey-m: dotnet run is super slow, so I think it's better to show folks how to build and run separately, so that folks who aren't as familiar with the command line doesn't think it's their only option.

dotnet/src/Shared/Demos/SampleEnvironment.cs
```diff
@@ -0,0 +1,140 @@
...
+            var color = Console.ForegroundColor;
+            Console.ForegroundColor = ConsoleColor.Green;
+            Console.Write("Using setting: Source=");
+            Console.ForegroundColor = ConsoleColor.Yellow;
+            Console.Write("EnvironmentVariables");
+            Console.ForegroundColor = ConsoleColor.Green;
+            Console.Write(", Key='");
+            Console.ForegroundColor = ConsoleColor.Yellow;
+            Console.Write(key);
+            Console.ForegroundColor = ConsoleColor.Green;
```
markwallace-microsoft, line 53: By default at we going to show all environment variables in the console?

If yes, then we shouldn't do this (even with the hideSampleValues safeguard). This can cause someone to be running a demo and then they accidentally disclose some sensitive information. I don't think people need to see there setting on every run.

dotnet/demos/Directory.Build.props
westey-m: The goal of this sample is actually to make the csproj files copy/paste friendly.  This file only contains settings that are needed to make this work in our repo, or enhance the functionality when running from here.

dotnet/src/Shared/Demos/SampleEnvironment.cs
westey-m: This file overrides the default System.Environment experience, but just for our demo folder.
The fact that you are not seeing it used is intentional, since it is designed to preserve copy/paste convenience.

dotnet/src/Shared/Demos/SampleEnvironment.cs
westey-m: Demoing to other people is not something that a person would do without having run this before, so I don't think that accidentally disclosing sensitive information in this way is likely.
Happy to change it to opt-in though if you are convinced that this is a real problem.

PR #481, opened 2025-08-25
.NET: fix runtime's AgentActor thread manipulations

### LINKED ISSUE #465: .NET: Ensure `AgentActor` correctly continues with restoring the thread context
While going through code I spotted 
https://github.com/microsoft/agent-framework/blob/65836ab125c9e6d80b244abbed71090f31638447/dotnet/src/Microsoft.Extensions.AI.Agents.Hosting/AgentActor.cs#L41

and I am not sure if it's an intended usage, or the result of deserialization has to be assigned to `this._agentThread`. Goal is to ensure `AgentActor` works correctly and make sure there are tests asserting correct behavior

DeagleGross:
1) fixed deserialization of the thread out of the persistent state data
2) used the in-place serialization method to standartize 

Fixes #465

### DIFF
--- a/dotnet/Directory.Build.props
+++ b/dotnet/Directory.Build.props
@@ -11,6 +11,7 @@
     <ImplicitUsings>disable</ImplicitUsings>
     <NoWarn>$(NoWarn);IDE0290;IDE0079;NU5128</NoWarn>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <ProjectsCoreTargetFrameworks>net9.0</ProjectsCoreTargetFrameworks>
     <ProjectsTargetFrameworks>net9.0;net8.0;netstandard2.0;net472</ProjectsTargetFrameworks>
     <ProjectsDebugTargetFrameworks>net9.0;net472</ProjectsDebugTargetFrameworks>
     <IsAotCompatible Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net7.0'))">true</IsAotCompatible>
--- a/dotnet/Directory.Packages.props
+++ b/dotnet/Directory.Packages.props
@@ -70,6 +70,8 @@
     <PackageVersion Include="Microsoft.SemanticKernel.Connectors.InMemory" Version="1.61.0-preview" />
     <!-- Agent SDKs -->
     <PackageVersion Include="Microsoft.Agents.CopilotStudio.Client" Version="1.1.151" />
+    <PackageVersion Include="A2A" Version="0.1.0-preview.2" />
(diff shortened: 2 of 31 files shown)


dotnet/src/Microsoft.Extensions.AI.Agents.Hosting/AgentActor.cs
```diff
@@ -120,10 +122,9 @@ private async Task HandleAgentRequestAsync(ActorRequestMessage message, Cancella
                 Log.AgentStreamingUpdate(logger, requestId, i);
             }
 
-            var serializedRunResponse = JsonSerializer.SerializeToElement(
-                updates.ToAgentRunResponse(),
-                AIJsonUtilities.DefaultOptions.GetTypeInfo(typeof(AgentRunResponse)));
-            var updatedThread = JsonSerializer.SerializeToElement(this._thread, AgentHostingJsonUtilities.DefaultOptions.GetTypeInfo(typeof(AgentThread)));
+            var serializedRunResponse = JsonSerializer.SerializeToElement(updates.ToAgentRunResponse(), AIJsonUtilities.DefaultOptions.GetTypeInfo(typeof(AgentRunResponse)));
+            var updatedThread = await this._thread.SerializeAsync(AgentHostingJsonUtilities.DefaultOptions, cancellationToken).ConfigureAwait(false);
```
westey-m, line 126: You probably don't need to pass `AgentHostingJsonUtilities.DefaultOptions`.  Serialize uses AgentAbstractionsJsonUtilities by default if nothing is passed, which contains all the required built in config/types for serializing/deserializing threads.

If a dev is using a custom IChatMessageStore though, they need to be able to pass a custom JsonSerializerOptions, so that they can configure the types for their custom store.  That's probably a different issue to what you are addressing here, so feel free to get this merged and consider that separately.  Give me a shout and we can chat more about this directly. (also, we may also want to revise how we configure serialization for custom types, since it feels a bit like we are breaking out of the abstraction if we have to pass the config with specific types here.)

```suggestion
            var updatedThread = await this._thread.SerializeAsync(cancellationToken: cancellationToken).ConfigureAwait(false);
```

DeagleGross: I think `AgentHostingJsonUtilities.DefaultOptions` are different from `AgentAbstractionsJsonUtilities.DefaultOptions` because they _could_ encapsulate more of the underlying settings including the one used in `AgentAbstractionsJsonUtilities.DefaultOptions`.

Lets keep it as it is today and we can work on that later. I dont see a trivial way of setting custom serialization options for user today, and we should have a separate PR configuring that

PR #507, opened 2025-08-27
.NET: Add more console based getting started samples

westey-m:
### Motivation and Context

#482 

### Description

- Replacing samples from getting started project with console based getting started samples
- Removing the replaced samples (2 are still left that are more provider specific so will be moved there)

### Contribution Checklist

### DIFF
--- a/dotnet/samples/GettingStartedSteps/Step08_ChatClientAgent_Telemetry/Program.cs
+++ b/dotnet/samples/GettingStartedSteps/Step08_ChatClientAgent_Telemetry/Program.cs
@@ -20,6 +20,7 @@ const string JokerInstructions = "You are good at telling jokes.";
 AppContext.SetSwitch("Microsoft.Extensions.AI.Agents.EnableTelemetry", true);
 
 // Create TracerProvider with console exporter
+// This will output the telemetry data to the console.
 string sourceName = Guid.NewGuid().ToString();
 using var tracerProvider = Sdk.CreateTracerProviderBuilder()
     .AddSource(sourceName)
@@ -42,5 +43,3 @@ await foreach (var update in agent.RunStreamingAsync("Tell me a joke about a pir
 {
     Console.WriteLine(update);
 }
-
-// The telemetry data will also be output to the console by the OpenTelemetry console exporter.
--- a/dotnet/src/Shared/Demos/SampleEnvironment.cs
+++ b/dotnet/src/Shared/Demos/SampleEnvironment.cs
@@ -35,7 +35,7 @@ internal static class SampleEnvironment
             Console.ForegroundColor = ConsoleColor.Yellow;
             Console.Write(key);
             Console.ForegroundColor = ConsoleColor.Green;
-            Console.Write("'> ");
+            Console.Write("'. Just press enter to accept the default. > ");
             Console.ForegroundColor = color;
             value = Console.ReadLine();
             value = string.IsNullOrWhiteSpace(value) ? null : value.Trim();


dotnet/samples/GettingStartedSteps/Step03_ChatClientAgent_UsingFunctionTools/Program.cs
```diff
@@ -0,0 +1,116 @@
...
+using System.Threading.Tasks;
+using Azure.AI.OpenAI;
+using Azure.Identity;
+using Microsoft.Extensions.AI;
+using Microsoft.Extensions.AI.Agents;
+using OpenAI;
+using SampleApp;
+
+var azureOpenAIEndpoint = Environment.GetEnvironmentVariable("AZURE_OPENAI_ENDPOINT") ?? throw new InvalidOperationException("AZURE_OPENAI_ENDPOINT is not set.");
+var azureOpenAIDeploymentName = Environment.GetEnvironmentVariable("AZURE_OPENAI_DEPLOYMENT_NAME") ?? "gpt-4o-mini";
```
markwallace-microsoft: We need to make this much shorter as it's much to complex for the second step in our getting started. I should be able to see the full sample on a single screen i.e. approx. 30 lines.

dotnet/samples/GettingStartedSteps/Step03_ChatClientAgent_UsingFunctionTools/Program.cs
```diff
@@ -0,0 +1,116 @@
...
+
+// Define the menu tools to be used by the agent.
+var menuTools = new MenuTools();
```
markwallace-microsoft: We just need a single function which returns a hard-coded string e.g. GetWeather for this sample

dotnet/samples/GettingStartedSteps/Step03_ChatClientAgent_UsingFunctionTools/Program.cs
```diff
@@ -0,0 +1,116 @@
...
+        AIFunctionFactory.Create(menuTools.GetItemPrice)
+    ]);
+
+// Create the chat client and agent.
+AIAgent agent = new AzureOpenAIClient(
+    new Uri(azureOpenAIEndpoint),
+    new AzureCliCredential())
+     .GetChatClient(azureOpenAIDeploymentName)
+     .CreateAIAgent(agentOptions);
+
```
markwallace-microsoft, line 27: Can we collapse all of this down to somethign very simple

```csharp
var response = await agent.RunAsync("What is the weather like today in Dublin?");
Console.WriteLine($"Agent: {response}");

await foreach (var update in agent.RunStreamingAsync("What is the weather like today in Dublin?"))
{
     Console.Write(update);
}
Console.WriteLine();
```

dotnet/samples/GettingStartedSteps/Step03_ChatClientAgent_UsingFunctionTools/Program.cs
```diff
@@ -0,0 +1,116 @@
...
+// Non-streaming agent interaction with function tools.
+Console.WriteLine("\n--- Run with function tools ---\n");
+AgentThread thread = agent.GetNewThread();
```
markwallace-microsoft: No ned to introduce threads here. Ideally we introduce one new concept per step and we should add threads when we are covering conversation state.

dotnet/samples/GettingStartedSteps/Step03_ChatClientAgent_UsingFunctionTools/Program.cs
```diff
@@ -0,0 +1,116 @@
...
+await RunAgentAsync("Hello");
+await RunAgentAsync("What is the special soup and its price?");
+await RunAgentAsync("What is the special drink and its price?");
+await RunAgentAsync("Thank you");
+
+async Task RunAgentAsync(string input)
```
markwallace-microsoft: Let's avoid doing this it's adding complexity that is uneeded

dotnet/samples/GettingStartedSteps/Step03_ChatClientAgent_UsingFunctionTools/Program.cs
```diff
@@ -0,0 +1,116 @@
...
+    }
+    Console.WriteLine();
+}
+
+namespace SampleApp
+{
+    /// <summary>
+    /// MenuTools class as used in the agent's function tools.
+    /// </summary>
+    internal sealed class MenuTools
```
markwallace-microsoft: We don't need to use this pattern, we can create simple inline functions rather than a class.

dotnet/samples/GettingStartedSteps/Step04_ChatClientAgent_UsingFunctionToolsWithApprovals/Program.cs
```diff
@@ -0,0 +1,172 @@
...
+        // Pass the user input responses back to the agent for further processing.
+        updates = await agent.RunStreamingAsync(nextIterationMessages, thread).ToListAsync();
+
+        userInputRequests = updates.SelectMany(x => x.UserInputRequests).ToList();
+    }
+
+    Console.WriteLine($"\nAgent: {updates.ToAgentRunResponse()}");
+}
+
+namespace SampleApp
```
markwallace-microsoft: There seems to be a lot of duplication in th samples. Also I don't think we need functions this complicated for this sample.

dotnet/samples/GettingStartedSteps/Step03_ChatClientAgent_UsingFunctionTools/Program.cs
```diff
@@ -0,0 +1,116 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using Azure.AI.OpenAI;
+using Azure.Identity;
+using Microsoft.Extensions.AI;
+using Microsoft.Extensions.AI.Agents;
+using OpenAI;
+using SampleApp;
+
+var azureOpenAIEndpoint = Environment.GetEnvironmentVariable("AZURE_OPENAI_ENDPOINT") ?? throw new InvalidOperationException("AZURE_OPENAI_ENDPOINT is not set.");
```
markwallace-microsoft: Can we just use `endpoint` and `deploymentName`?

dotnet/samples/GettingStartedSteps/Step03_ChatClientAgent_UsingFunctionTools/Program.cs
```diff
@@ -0,0 +1,116 @@
...
+AIAgent agent = new AzureOpenAIClient(
+    new Uri(azureOpenAIEndpoint),
+    new AzureCliCredential())
+     .GetChatClient(azureOpenAIDeploymentName)
+     .CreateAIAgent(agentOptions);
+
+// Non-streaming agent interaction with function tools.
+Console.WriteLine("\n--- Run with function tools ---\n");
+AgentThread thread = agent.GetNewThread();
+await RunAgentAsync("Hello");
```
markwallace-microsoft: Using this pattern adds a level of indirection, we want the AF API to be front and center in all of the samples.

dotnet/samples/GettingStartedSteps/Step03_ChatClientAgent_UsingFunctionTools/Program.cs
rogerbarreto: Maybe just one call is enough or multiple `Console.Write(await agent.RunAsync("...", thread))`

dotnet/samples/GettingStartedSteps/Step04_ChatClientAgent_UsingFunctionToolsWithApprovals/Program.cs
rogerbarreto: +1, we might remove all the streaming block at once.

dotnet/samples/GettingStartedSteps/Step04_ChatClientAgent_UsingFunctionToolsWithApprovals/Program.cs
```diff
@@ -0,0 +1,172 @@
...
+await RunAgentAsync("What is the special soup and its price?");
+await RunAgentAsync("What is the special drink?");
+
+async Task RunAgentAsync(string input)
+{
+    Console.WriteLine($"\nUser: {input}");
+    var response = await agent.RunAsync(input, thread);
+
+    // Loop until all user input requests are handled.
+    var userInputRequests = response.UserInputRequests.ToList();
```
rogerbarreto: We can make this very compact if shown streaming side by side.

```suggestion
    var response = await agent.RunAsync(input, thread);
    // Streaming
    // var updates = await agent.RunStreamingAsync(input, thread).ToListAsync();

    // Loop until all user input requests are handled.
    var userInputRequests = response.UserInputRequests.ToList();
    // Streaming
    // var userInputRequests = updates.SelectMany(x => x.UserInputRequests).ToList();
```

dotnet/samples/GettingStartedSteps/Step04_ChatClientAgent_UsingFunctionToolsWithApprovals/Step04_ChatClientAgent_UsingFunctionToolsWithApprovals.csproj
rogerbarreto, line 1: Suggest adding `GettingStartedSteps` prefix to the project name, when listing the projects and building gives more clarity what this is.

This can become problematic when mixed with other getting started steps, and Migration samples (I'm also introducing to the repo, with a `prefix`).

dotnet/samples/GettingStartedSteps/Step03_ChatClientAgent_UsingFunctionTools/Program.cs
westey-m: Makes sense.  Renamed.

dotnet/samples/GettingStartedSteps/Step03_ChatClientAgent_UsingFunctionTools/Program.cs
westey-m: Makes sense.  I've removed the menuplugin and it's very short now.

dotnet/samples/GettingStartedSteps/Step03_ChatClientAgent_UsingFunctionTools/Program.cs
westey-m: Makes sense.  I've removed the menuplugin and it's very short now.

dotnet/samples/GettingStartedSteps/Step03_ChatClientAgent_UsingFunctionTools/Program.cs
westey-m: Sure, I've switched to a single invocation, so much shorter now.

dotnet/samples/GettingStartedSteps/Step03_ChatClientAgent_UsingFunctionTools/Program.cs
westey-m: Agreed, since we just have a single invocation, we can skip threads here.

dotnet/samples/GettingStartedSteps/Step03_ChatClientAgent_UsingFunctionTools/Program.cs
westey-m: Agreed, removed that.

dotnet/samples/GettingStartedSteps/Step03_ChatClientAgent_UsingFunctionTools/Program.cs
westey-m: Agreed, updated to just one call without the method.

dotnet/samples/GettingStartedSteps/Step03_ChatClientAgent_UsingFunctionTools/Program.cs
westey-m: Agreed, removed.

dotnet/samples/GettingStartedSteps/Step04_ChatClientAgent_UsingFunctionToolsWithApprovals/Program.cs
westey-m: Makes sense, updated.

dotnet/samples/GettingStartedSteps/Step04_ChatClientAgent_UsingFunctionToolsWithApprovals/Program.cs
westey-m: I've used Roger's recommendation for this one and replaced the streaming section with just a couple of comments around what the streaming logic would be.

dotnet/samples/GettingStartedSteps/Step04_ChatClientAgent_UsingFunctionToolsWithApprovals/Step04_ChatClientAgent_UsingFunctionToolsWithApprovals.csproj
westey-m: I'm not sure about adding that as well.  These are already in a folder both in the repo and in the solution with this name, and these names are already too long IMO.

rogerbarreto: We might reconsider the name as those won't be side-by-side with the providers, as those are now end-to-end

dotnet/samples/GettingStartedSteps/Step01_ChatClientAgent_Running/Program.cs
```diff
@@ -8,16 +8,16 @@
...
+var deploymentName = Environment.GetEnvironmentVariable("AZURE_OPENAI_DEPLOYMENT_NAME") ?? "gpt-4o-mini";
 
 const string JokerName = "Joker";
 const string JokerInstructions = "You are good at telling jokes.";
 
 AIAgent agent = new AzureOpenAIClient(
-    new Uri(azureOpenAIEndpoint),
+    new Uri(endpoint),
     new AzureCliCredential())
-     .GetChatClient(azureOpenAIDeploymentName)
```
markwallace-microsoft, line 20: Below we have lines like `Console.WriteLine("--- Run the agent ---\n");`, do we need these?

dotnet/samples/GettingStartedSteps/Step03_ChatClientAgent_UsingFunctionTools/Program.cs
```diff
@@ -0,0 +1,39 @@
...
+
+// Non-streaming agent interaction with function tools.
+Console.WriteLine("\n--- Run with function tools ---\n");
+Console.WriteLine(await agent.RunAsync("What is the weather like in Amsterdam?"));
+
+// Streaming agent interaction with function tools.
+Console.WriteLine("\n--- Run with function tools and streaming ---\n");
+await foreach (var update in agent.RunStreamingAsync("What is the weather like in Amsterdam?"))
+{
+    Console.Write(update);
```
markwallace-microsoft: In step 01 we use `WriteLine` here. The problem with using `Write` is that it will run so fast the user won't see streaming. Using `WriteLine` makes the streaming nature very clear.

westey-m: Sure, I can update that.

dotnet/samples/GettingStartedSteps/Step01_ChatClientAgent_Running/Program.cs
westey-m: We probably don't since we typically have very little output and the context is clear, especially when using writeline for streaming.

dotnet/demos/MinimalConsole/Program.cs
```diff
@@ -8,17 +8,17 @@
 using Microsoft.Extensions.AI.Agents;
 using OpenAI;
 
-var azureOpenAIEndpoint = Environment.GetEnvironmentVariable("AZURE_OPENAI_ENDPOINT") ?? throw new InvalidOperationException("AZURE_OPENAI_ENDPOINT is not set.");
-var azureOpenAIDeploymentName = Environment.GetEnvironmentVariable("AZURE_OPENAI_DEPLOYMENT_NAME") ?? "gpt-4o-mini";
+var endpoint = Environment.GetEnvironmentVariable("AZURE_OPENAI_ENDPOINT") ?? throw new InvalidOperationException("AZURE_OPENAI_ENDPOINT is not set.");
+var deploymentName = Environment.GetEnvironmentVariable("AZURE_OPENAI_DEPLOYMENT_NAME") ?? "gpt-4o-mini";
```
rogerbarreto, line 12: Current helper implementation don't allow optionality, in my experience, the most straightforward way to deal with `Optional` environment variables without changing the original signature was adding the `System`. 

```suggestion
var deploymentName = System.Environment.GetEnvironmentVariable("AZURE_OPENAI_DEPLOYMENT_NAME") ?? "gpt-4o-mini";
```

dotnet/samples/GettingStartedSteps/Step01_ChatClientAgent_Running/Program.cs
```diff
@@ -8,25 +8,23 @@
...
 // Invoke the agent and output the text result.
-Console.WriteLine("--- Run the agent ---\n");
 Console.WriteLine(await agent.RunAsync("Tell me a joke about a pirate."));
 
 // Invoke the agent with streaming support.
-Console.WriteLine("\n--- Run the agent with streaming ---\n");
 await foreach (var update in agent.RunStreamingAsync("Tell me a joke about a pirate."))
 {
-    Console.Write(update);
+    Console.WriteLine(update);
```
rogerbarreto, line 29: Streaming should be `Write` for demonstration (closest to the a real use-case scenario)

```suggestion
    Console.Write(update);
```

dotnet/samples/GettingStartedSteps/Step04_ChatClientAgent_UsingFunctionToolsWithApprovals/Program.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.ComponentModel;
+using System.Linq;
+using Azure.AI.OpenAI;
+using Azure.Identity;
+using Microsoft.Extensions.AI;
+using Microsoft.Extensions.AI.Agents;
+using OpenAI;
+
+var endpoint = Environment.GetEnvironmentVariable("AZURE_OPENAI_ENDPOINT") ?? throw new InvalidOperationException("AZURE_OPENAI_ENDPOINT is not set.");
+var deploymentName = Environment.GetEnvironmentVariable("AZURE_OPENAI_DEPLOYMENT_NAME") ?? "gpt-4o-mini";
```
rogerbarreto, line 18: To allow optionality. (Apply for others)

```suggestion
var deploymentName = System.Environment.GetEnvironmentVariable("AZURE_OPENAI_DEPLOYMENT_NAME") ?? "gpt-4o-mini";
```

dotnet/samples/GettingStartedSteps/Step01_ChatClientAgent_Running/Program.cs
westey-m: @markwallace-microsoft recommended that we use WriteLine so that it's clear to users that we are doing streaming.  Otherwise it prints out so fast it looks just like non-stremang.

dotnet/samples/GettingStartedSteps/Step06_ChatClientAgent_PersistedConversations/Step06_ChatClientAgent_PersistedConversations.csproj
```diff
@@ -0,0 +1,23 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net9.0</TargetFramework>
+    <LangVersion>12</LangVersion>
```
rogerbarreto, line 6: Do we need to enforce the langversion ?
```suggestion
```

dotnet/samples/GettingStartedSteps/Step08_ChatClientAgent_Telemetry/Program.cs
```diff
@@ -0,0 +1,46 @@
...
+// Invoke the agent and output the text result.
+Console.WriteLine(await agent.RunAsync("Tell me a joke about a pirate."));
+
+// Invoke the agent with streaming support.
+await foreach (var update in agent.RunStreamingAsync("Tell me a joke about a pirate."))
+{
+    Console.WriteLine(update);
+}
+
+// The telemetry data will also be output to the console by the OpenTelemetry console exporter.
```
rogerbarreto: I suggest moving this comment to line 26 above `ConsoleExporter()` extension

dotnet/demos/MinimalConsole/Program.cs
westey-m: The user could just enter nothing, in which case it would use the default.  I can update the helper message to tell the user that.

dotnet/samples/GettingStartedSteps/Step09_ChatClientAgent_DependencyInjection/Program.cs
```diff
@@ -0,0 +1,99 @@
...
+await host.RunAsync(appShutdownCancellationToken).ConfigureAwait(false);
+
+namespace SampleApp
+{
+    /// <summary>
+    /// A sample service that uses an AI agent to respond to user input.
+    /// </summary>
+    internal sealed class SampleService(AIAgent agent, [FromKeyedServices("AppShutdown")] CancellationTokenSource appShutdownCancellationTokenSource) : IHostedService
+    {
+        private AgentThread? _thread;
```
rogerbarreto, line 60: Start is never being called, in theory this will be always `null`, or I missed something?

```suggestion
```

dotnet/samples/GettingStartedSteps/Step09_ChatClientAgent_DependencyInjection/Program.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        public async Task StartAsync(CancellationToken cancellationToken)
+        {
+            // Create a thread that will be used for the entirety of the service lifetime so that the user can ask follow up questions.
+            this._thread = agent.GetNewThread();
+            _ = this.RunAsync(cancellationToken);
+        }
+
```
rogerbarreto, line 68: AFAIK, this is not being used and can be removed

```suggestion
```

dotnet/samples/GettingStartedSteps/Step06_ChatClientAgent_PersistedConversations/Step06_ChatClientAgent_PersistedConversations.csproj
westey-m: This setting is inherited from the directory.build.props, but I don't think we should rely on that.  If users want to replicate this sample, they should be able to copy the settings from the csproj file and it should just work.

dotnet/samples/GettingStartedSteps/Step04_ChatClientAgent_UsingFunctionToolsWithApprovals/Program.cs
westey-m: I've updated the helper to indicate that users can pick the default by just pressing enter.

dotnet/samples/GettingStartedSteps/Step08_ChatClientAgent_Telemetry/Program.cs
westey-m: Done.

dotnet/samples/GettingStartedSteps/Step09_ChatClientAgent_DependencyInjection/Program.cs
westey-m: AgentThread has to be nullable, since it is not set in the constructor, but only in StartAsync.

dotnet/samples/GettingStartedSteps/Step09_ChatClientAgent_DependencyInjection/Program.cs
westey-m: The runnable host invokes this, and it is what kicks off the RunAsync method.

dotnet/demos/MinimalConsole/Program.cs
rogerbarreto: Even so, the user will never know what is or not optional from the console window. Ideally optional interface we show when a parameter is optional and what is the default.

dotnet/samples/GettingStartedSteps/Step01_ChatClientAgent_Running/Program.cs
rogerbarreto: I would change the prompt instead for something more verbose, as I also understand the overall idea is also how we recommend users to use the API's.

westey-m: Yeah, happy to look into this more.  I'll get his merged now, so that we have something for the bugbash.

PR #512, opened 2025-08-27
Python: autogen migration guide

peterychang:
https://github.com/microsoft/agent-framework/issues/381

Guide to migrate model clients and chat client agents from `Autogen` 0.4+ to `Agent Framework`


python/autogen_migration.md
```diff
@@ -0,0 +1,110 @@
...
+    "config": {
+        "model": "gpt-4o",
+        "api_key": "sk-xxx" # os.environ["...']
+    }
+}
+
+model_client = ChatCompletionClient.load_component(config)
+```
+
+Agent Framework currently requires its components to be configured individually with a configuration dictionary, though in the future we will be implementing Copilot Studio Declarative Language (CPSDL) as our official declarative specification.
```
victordibia: I am not too confident at how things will shake out with CPSDL.
I'll rather leave this as saying there may be changes to the declarative specification in the future.

victordibia:
```suggestion
Agent Framework currently requires its components to be configured individually with a configuration dictionary, though there may be changes to this in the future for improved developer experience.
```

python/autogen_migration.md
```diff
@@ -0,0 +1,110 @@
...
+client = AzureChatClient(
+    deployment_name="gpt-4o",
+    endpoint="https://<your-endpoint>.openai.azure.com/",
+    api_key="sk-xxx"
+)
+```
+
+## Chat Agent
+
+In Autogen, a specialized agent implementation was created for use as an assistant agent. With Agent Framework, a chat agent can be created from the model client itself.
```
victordibia:
```suggestion
Autogen  provided a convenience [AssistantAgent](https://microsoft.github.io/autogen/dev/user-guide/agentchat-user-guide/tutorial/agents.html#assistant-agent) class with a set of defined functionality. With Agent Framework, a similar chat agent can be created from the model client itself.
```

python/autogen_migration.md
```diff
@@ -0,0 +1,110 @@
+# Migration from Autogen to Agent Framework
```
TaoChenOSU: Do we want to put this file in the user-documentation directory: https://github.com/microsoft/agent-framework/tree/main/user-documentation?

python/autogen_migration.md
```diff
@@ -0,0 +1,110 @@
...
+```python
+from agent_framework.openai import OpenAIChatClient
+
+config = {
+    "ai_model_id": "gpt-4o",
+    "api_key": "sk-xxx" # os.environ["..."]
+}
+
+chat_client = OpenAIChatClient.from_dict(config)
+```
```
dmytrostruk: A couple of points:
1. This block sounds like the developer experience in Agent Framework is more complicated than it is in AutoGen (`In Autogen, a self-describing configuration was used to create its various components, while Agent Framework requires its components to be configured invididually`), which may raise some questions like "_Why do I need to migrate if current experience is better?_". Maybe we can simply say that in AutoGen `ChatCompletionClient` component is required to configure various clients like `OpenAIChatCompletionClient`, while in Agent Framework the specific component like `OpenAIChatClient` is required for configuration.
2. In this proposed version of migration guide, we start with more complex initialization example using `OpenAIChatClient.from_dict(config)` syntax, which is followed by easier syntax when we use the constructor `OpenAIChatClient(...)` in the next section. One way of potential improvement is to put `### Use model client directly` section with easier syntax at the beginning and continue with more complex syntax like `OpenAIChatClient.from_dict(config)` for specific scenarios, so the document feels like a progression. Let me know if that makes sense.

python/autogen_migration.md
```diff
@@ -0,0 +1,110 @@
...
+agent = OpenAIChatClient(
+    ai_model_id="gpt-4o",
+    api_key="sk-xxx",
+).create_agent(
+    name="assistant",
+    instructions="You are a helpful assistant.",
+)
+```
+
+The functions `run` and `run_streaming` are analogous to autogen's `run` and `run_stream` calls, with both being asynchronous and returning an async iterable for the latter.
```
dmytrostruk: It should be helpful if we provide the code sample for both Agent Framework and AutoGen, so users can compare it side-by-side and understand the difference. Even if there is no difference, it will be still helpful to understand that the experience is very similar if not the same. This also applies to `### Use model client directly` section.

python/autogen_migration.md
```diff
@@ -0,0 +1,110 @@
+# Migration from Autogen to Agent Framework
+
+This is a migration guide for users of [Autogen](https://microsoft.github.io/autogen/stable/) to `Agent Framework`.
```
ekzhu: Let's use the GitHub repo link for longevity.

python/autogen_migration.md
```diff
@@ -0,0 +1,110 @@
...
+    ai_model_id="gpt-4o",
+    api_key="sk-xxx",
+)
+
+agent = ChatClientAgent(
+    client=client,
+    instructions="You are a helpful assistant.",
+    name="assistant",
+)
+```
```
ekzhu: We need to explain the differences in the message types. Suggest we list out the message types in AG, and message types in AF, and pointing to the differences.

python/autogen_migration.md
```diff
@@ -0,0 +1,110 @@
...
+
+This is a migration guide for users of [Autogen](https://microsoft.github.io/autogen/stable/) to `Agent Framework`.
+
+
+## What is Agent Framework?
+Agent Framework is an asynchronous, event-driven Agentic framework library which merges the best ideas and features from both `Autogen` and `Semantic Kernel`. The high level structure of Agent Framework is similar to that of Autogen, including the components used to build complex workflows.
+
+This guide will focus on translating the various components from `Autogen 0.4+` to `Agent Framework`
+
+## Package structure
```
ekzhu: I think it's important to note that AG has two layers of abstractions: AgentChat and Core. 
- The main package and vendor extensions maps to single agent in AG's AgentChat. 
- The workflow package maps to multi-agent in AgentChat and the Core package. 
- Multi-agent in AG maps to the high-level orchestration patterns in AF workflows
- Core in AG maps to the low-level dataflow API in AF workflow.

python/autogen_migration.md
```diff
@@ -0,0 +1,110 @@
...
+
+Autogen is broken down into several component packages, including `autogen_agentchat`, `autogen_core`, `autogen_ext`, etc. Objects are also nested into logical subpackages.
+
+The Agent Framework package structure opts toward a flatter structure. The packages are broken into the core package, and model-specific packages; including `agent-framework`, `agent-framework-azure`, `agent-framework-openai`, etc. However, each package's contents can be accessed through the `agent_framework` namespace as well. For example, `OpenAIChatClient` can be accessed with either `agent-framework-openai.OpenAIChatClient` or `agent_framework.openai.OpenAIChatClient`. Most objects, outside of specific components like errors or telemetry functionality, are contained within the top level package.
```
eavanvalkenburg:
```suggestion
The Agent Framework package structure opts toward a flatter structure. The packages are broken into the core package, and several sub-packages; such as `azure`, `openai`, `workflow`, etc. However, each package's contents can and should be accessed through the `agent_framework` namespace and installed using the provided extra to ensure all dependencies are present and have the right versions. However this does mean that for example, `AzureChatClient` can be accessed with either `agent_framework_azure.AzureChatClient` or `agent_framework.azure.AzureChatClient`. Most objects, outside of specific components like exceptions or telemetry functionality, are contained within the top level package.
```

python/autogen_migration.md
```diff
@@ -0,0 +1,110 @@
...
+    "ai_model_id": "gpt-4o",
+    "api_key": "sk-xxx" # os.environ["..."]
+}
+
+chat_client = OpenAIChatClient.from_dict(config)
+```
+
+### Use model client directly
+
+The model clients themselves can be configured nearly identically to Autogen, with the exception of the occasional parameter name change
```
eavanvalkenburg: let's also note that we use pydantic settings to read the most important config settings from the environment/a .env file, so that in many cases, just calling `OpenAIChatClient()` is enough.

python/autogen_migration.md
peterychang: In autogen, you can generate the object generically through the `ComponentLoader`. So you could get the `OpenAIChatClient`, or even the entire workflow by calling `ComponentLoader.load_component(...)`

peterychang:
> A general comments: we need to make a clear case in the migration guide about the advantages of AF over AG. You can see AG's 0.2 to 0.4 migration guide as an example.
> 
> Let's have a the list of advantages and improvements as the first section of the migration guide to motivate users.
> 
> * Support for Responses API and hosted tools
> * Unified interface across ChatCompletion, Responses, and Assistant APIs
> * Better support for open protocols: MCP, A2A.
> 
> For workflows, we can list out a number of new things like
> 
> * state management: checkpointing, time travel
> * human in the loop with pausing and resuming of execution
> * nested workflows
> * tracing
> * realtime visualization
> 
> I can update later if I thought of more. I think we need this list to get started first.

I was planning on leaving the workflow out of this PR, since I'm not familiar with the work thats being done in that space

Issue #522, opened 2025-08-28
Python: Provide guidance on packages for private preview


markwallace-microsoft:
Either we publish packages in the repo or we provide guidance for developers to create their own packages.

PR #523, opened 2025-08-28
Python: package installation from git

eavanvalkenburg:
We can add a separate section on installing from github releases once we decide to use that approach, this works.

Closes: #522

### DIFF
--- a/user-documentation-python/getting-started/package_installation.md
+++ b/user-documentation-python/getting-started/package_installation.md
@@ -18,7 +18,7 @@ Globally, these are the steps:
 	- Workflow: `python/packages/workflow`
 
 Why use the repo URL?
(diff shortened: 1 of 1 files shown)


user-documentation-python/getting-started/package_installation.md
westey-m, line 1: Can you add this to the https://github.com/microsoft/agent-framework/tree/fc72ecc889be7d6a9f7e3b47a965bfaaecab57ef/docs/docs-templates folder.  You can add a pivot for csharp that just says coming soon.

See #527 for more info.

eavanvalkenburg: This is temporary for private preview, after that it is regular installs from PyPI so not needed to have such a doc in this way (regular readme will do then)

PR #517, opened 2025-08-28
.NET: Fix AIAgentHostExecutor streaming message aggregation strategy

TaoChenOSU:
### Motivation and Context

Streaming messages from the AIAgentHostExecutor is not being sent correctly.

### Description

This pull request refactors the message streaming logic in `AIAgentHostExecutor.cs` to improve handling of streaming agent responses. The main changes focus on simplifying message assembly, ensuring correct message publication, and cleaning up state between streamed messages.

### Contribution Checklist

- [x] The code builds clean without any errors or warnings
- [x] The PR follows the [Contribution Guidelines](https://github.com/microsoft/agent-framework/blob/main/CONTRIBUTING.md)
- [x] All unit tests pass, and I have added new tests where possible
- [x] I didn't break anyone :smile:

PR #530, opened 2025-08-28
.NET: fix: ChatMessage Aggregation in AIAgentHostExecutor

lokitoth:
### Motivation and Context

The previous change to the aggregation in AIAgentHostExecutor (#515) has a bug wherein we only include the first update from a ChatMessage in the outgoing "collected" message.

### Description

The fix is to ensure we put the collected updates into the message before sending it out. Builds on the fix in #517.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Microsoft.Agents.Workflows/Specialized/AIAgentHostExecutor.cs
+++ b/dotnet/src/Microsoft.Agents.Workflows/Specialized/AIAgentHostExecutor.cs
@@ -94,7 +94,7 @@ internal class AIAgentHostExecutor : Executor
         bool emitEvents = token.EmitEvents.HasValue ? token.EmitEvents.Value : this._emitEvents;
         IAsyncEnumerable<AgentRunResponseUpdate> agentStream = this._agent.RunStreamingAsync(this._pendingMessages, this.EnsureThread(context));
 
-        List<AgentRunResponseUpdate> updates = new();
+        List<AIContent> updates = new();
         ChatMessage? currentStreamingMessage = null;
 
         await foreach (AgentRunResponseUpdate update in agentStream.ConfigureAwait(false))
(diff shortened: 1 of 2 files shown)


dotnet/src/Microsoft.Agents.Workflows/Specialized/AIAgentHostExecutor.cs
```diff
@@ -132,6 +132,9 @@ async ValueTask PublishCurrentMessageAsync()
         {
             if (currentStreamingMessage != null)
             {
+                currentStreamingMessage.Contents = updates;
```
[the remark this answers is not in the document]
lokitoth: This is incorrect. We collect all the AIContent in `updates`, including the first one we already shoved into the ChatMessage we are building. If we use `AddRange` we will duplicate the first one.

dotnet/src/Microsoft.Agents.Workflows/Specialized/AIAgentHostExecutor.cs
```diff
@@ -132,6 +132,9 @@ async ValueTask PublishCurrentMessageAsync()
         {
```
crickman, line 132: I'm not sure this local function is helping with readability.  Consider:

```c#
            if (currentStreamingMessage == null || currentStreamingMessage.MessageId != update.MessageId)
            {
                if (currentStreamingMessage != null)
                {
                    await context.SendMessageAsync(currentStreamingMessage).ConfigureAwait(false);
                }
                updates = [];
                currentStreamingMessage = new(update.Role ?? ChatRole.Assistant, update.Contents)
                {
                    AuthorName = update.AuthorName,
                    Contents = updates, // The object reference isn't changing as the list is populated
                    CreatedAt = update.CreatedAt,
                    MessageId = update.MessageId,
                    RawRepresentation = update.RawRepresentation,
                    AdditionalProperties = update.AdditionalProperties
                };
            }
```

lokitoth: It also needs to happen after exiting the loop, and to ensure we don't get into a place where we have something that needs to happen added to one and not the other, I used the local variant; with that said, happy to simplify this next time I touch it.

PR #555, opened 2025-08-29
.NET: Further fixes for compatibility with Microsoft.Extensions repo.

westey-m:
### Motivation and Context

#451 
#452

### Description

- Fixing some static analysis issues that show up in Microsoft.Extensions repo.
- Avoid using reflection based json serialization/deserializing in tests, since it is not allows in Microsoft.Extensions tests.
- Removing references to code we are not yet ready to move.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatClientAgentExtensions.cs
+++ b/dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatClientAgentExtensions.cs
@@ -33,7 +33,7 @@ public static class ChatClientAgentExtensions
         Throw.IfNull(agent);
         Throw.IfNull(messages);
 
-        return agent.RunAsync(messages, thread, new ChatClientAgentRunOptions(agentRunOptions, chatOptions), cancellationToken);
+        return agent.RunAsync(messages, thread, new ChatClientAgentRunOptions(chatOptions), cancellationToken);
     }
 
     /// <summary>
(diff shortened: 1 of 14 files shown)


dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatClientAgentExtensions.cs
```diff
@@ -33,7 +33,7 @@ public static Task<AgentRunResponse> RunAsync(
         Throw.IfNull(agent);
         Throw.IfNull(messages);
 
-        return agent.RunAsync(messages, thread, new ChatClientAgentRunOptions(agentRunOptions, chatOptions), cancellationToken);
```
rogerbarreto, line 36: This change is not related to the PR proposal.

Before changing this behavior we should consider merging them or put some clarity into how handle `agentRunOptions is ChatClientAgentRunOptions` + `chatOptions` 

OR 

Test if the provided `agentRunOptions` is already a `ChatClientAgentRunOptions` and ignore `ChatOptions` instead.

westey-m: This change is due to an update in ChatClientAgentRunOptions, where I'm removing the internal overload that takes an unused parameter.  ChatClientAgentRunOptions doesn't use AgentRunOptions for anything, since AgentRunOptions has no properties.

PR #568, opened 2025-08-31
.NET: Multiple fixes in Microsoft.Agents.Orchestration to make handoffs work end-to-end

### LINKED ISSUE #566: .NET: Handoff orchestrations can't be resumed using RunAsync
I'm trying to create a handoff sample based off [this one](https://github.com/microsoft/agent-framework/blob/main/dotnet/samples/GettingStarted/Orchestration/HandoffOrchestration_Intro.cs) in the official samples. However, instead of using `InteractiveCallback`, my app makes multiple calls to `RunAsync`. I noticed that I get a runtime exception when trying to run the handoff orchestration a second time.

### Scenario
```csharp
HandoffOrchestration orchestrationAgent = // ...
IActorRuntimeContext myRuntimeContext = // ...

// Turn 1
string message1 = "I am a customer that needs help with my orders.";
await using OrchestratingAgentResponse orchestratingResponse =
    await orchestrationAgent.RunAsync(
        messages: [new ChatMessage(ChatRole.User, message1)],
        options: null,
        runtime: myRuntimeContext);

// Response is something like "I can help with that! Can you give me more info?"
response = await orchestratingResponse.Task;

// ...

// turn 2 (fails with an error)
string message2 = "I'd like to track the status of my order";
await using OrchestratingAgentResponse orchestratingResponse =
    await orchestrationAgent.RunAsync(
        messages: [new ChatMessage(ChatRole.User, message2)],
        options: null,
        runtime: myRuntimeContext);
```

### Result

The second invocation results in the following exception:

```
Exception: System.InvalidOperationException: The next agent '' is not defined in the orchestration.
    at Microsoft.Shared.Diagnostics.Throw.InvalidOperationException(String message)
    at Microsoft.Agents.Orchestration.HandoffOrchestration.ResumeCoreAsync(JsonElement checkpointState, OrchestratingAgentContext context, CancellationToken cancellationToken)
    at Microsoft.Agents.Orchestration.OrchestratingAgent.RunAsync(IReadOnlyCollection`1 messages, AgentRunOptions options, IActorRuntimeContext runtime, CancellationToken cancellationToken)
```

### Analysis

Looking at the `HandoffOrchestration` code, it seems that a handoff with *any* state automatically goes into the `ResumeCoreAsync` method. This code path basically assumes that a next agent (`state.NextAgent`) contains a non-null value. However, it looks like `state.NextAgent` is set to null and persisted when a run doesn't result in a handoff. Any subsequent run would then fail with this error.

### Expectation

My expectation is that my scenario should work fine. If there was no handoff and the initial agent provided the response, then I expect that subsequent calls to `RunAsync` will target the initial agent.

It's less clear to me what to expect if a different agent ends the handoff chain, whether the next run should go to the initial agent or whether it should go to the previous agent.

### Workarounds

I haven't yet found a workaround for this.

### LINKED ISSUE #567: .NET: Message parameter ignored when running orchestrating agent second time
### Scenario

The scenario I have in mind is a multi-turn chat session with a stateful agent orchestration. The requirement is that I can rehydrate the stateful agent orchestration in another process and continue where I left off. The primary use case for something like this is human interaction.

The main thing I need working for the above is the ability to call `HandoffOrchestration.RunAsync` multiple times.

### Expected

1. Call RunAsync with a message, get back a response from the agent
2. Call RunAsync with a follow-up message, get back a new response from the agent based on the follow-up message

### Actual

The second call to RunAsync ignores the follow-up message and instead only considers messages in the previously saved checkpoint. As a result, I get the wrong response from the agent (it only responds to the original message).

It seems that all orchestrating agent types have the same issue, not just `HandoffOrchestration`.

cgillum:
### Motivation and Context

I'm working on an end-to-end sample that uses handoffs. The sample runs in Azure Functions behind an HTTP API and leverages a custom runtime for persistent checkpointing. However, I ran into several issues that needed to be fixed in order for a basic handoff sample to work.

This PR includes fixes for all the issues encountered, including a minor breaking change to the base orchestration abstraction.

Below is a subset of the issues fixed by this PR:

Fixes https://github.com/microsoft/agent-framework/issues/566
Fixes https://github.com/microsoft/agent-framework/issues/567

## Description

The following are the main changes needed to get an end-to-end handoff orchestration working. The description is a bit long, but the actual changes are relatively small.

### Enable Multi-RunAsync Interactions (BREAKING)

When calling the `OrchestratingAgent.RunAsync` method and resuming from a checkpoint, it called `ResumeCoreAsync` but didn't pass the new messages to it. The `ResumeCoreAsync` method signature only accepted the checkpoint state and context, not the new messages.

I made the following changes to address this:

- Updated the `RunAsync` method to pass the new messages to `ResumeCoreAsync`
- Changed the `ResumeCoreAsync` method signature to accept the new messages parameter
- Updated all orchestration implementations to handle the new messages
- Updated `HandoffOrchestration` to default to the initial agent when no "NextAgent" value was saved into the checkpoint

Although breaking, this felt like an obvious change to make based on normal developer expectations.

### Fix tool calling in HandoffOrchestration

There were two fundamental problems with tool calling in `HandoffOrchestration`.

1. `ChatMessage` objects containing JsonElement types as a result of a tool call weren't serializeable because `OrchestrationJsonContext` didn't have sufficient information to serialize them.

    The error message was the following:
  
    ```
    NotSupportedException: JsonTypeInfo metadata for type 'System.Text.Json.JsonElement' was not provided by TypeInfoResolver of type 'Microsoft.Agents.Orchestration.OrchestrationJsonContext'. If using source generation, ensure that all root types passed to the serializer have been annotated with 'JsonSerializableAttribute', along with any types that might be serialized polymorphically.
    ```

    This was fixed by adding `[JsonSerializable(typeof(JsonElement))]` to `OrchestrationJsonContext`.

1. The `RemoveHandoffFunctionCalls` method is designed to clean up handoff-related function calls, but it was also removing non-handoff-related function calls, causing Bad Request errors when invoking OpenAI endpoints.

    The error message was (in my example):

    ```
    HTTP 400 (invalid_request_error: )
    Parameter: messages.[9].role
    An assistant message with 'tool_calls' must be followed by tool messages responding to each 'tool_call_id'. The following tool_call_ids did not have response messages: call_GNSRHcHS1J5jlG0H2j7ppL5h
   ```

    The fix was to remove only handoff-related function calls.

With these fixes, I was able to get an end-to-end handoff orchestration working in an Azure Functions app using a simple custom runtime implementation.

## Contribution Checklist

- [ x The PR follows the [Contribution Guidelines](https://github.com/microsoft/agent-framework/blob/main/CONTRIBUTING.md)
    - NOTE: The unit tests seem to be missing sufficient infrastructure to actually test orchestrations in a comprehensive way (tool calling, checkpointing, etc.). That seemed out of scope for this PR. However, the existing tests still passing.
- [x] I didn't break anyone :smile: (fingers crossed)


dotnet/src/Microsoft.Agents.Orchestration/Handoffs/HandoffOrchestration.cs
```diff
@@ -52,26 +52,29 @@ protected override Task<AgentRunResponse> RunCoreAsync(IReadOnlyCollection<ChatM
...
             }
         }
 
-        if (nextAgent is null)
-        {
-            Throw.InvalidOperationException($"The next agent '{state.NextAgent}' is not defined in the orchestration.");
-        }
+        // Append the new messages to the checkpoint state
+        List<ChatMessage> allMessages = [.. state.AllMessages, .. newMessages];
+        int originalMessageCount = state.OriginalMessageCount + newMessages.Count;
```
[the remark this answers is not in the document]
cgillum: OriginalMessageCount seems like it's not being used, or at least that its purpose is not well defined. It seems the original author may have intended to use it for one purpose but didn't fully implement that purpose. For now, I'm going to remove its usage here since its current usage appears to be buggy (we're trimming the wrong number of messages from the response after the first RunAsync call).

dotnet/src/Microsoft.Agents.Orchestration/Handoffs/HandoffOrchestration.cs
```diff
@@ -215,7 +216,7 @@ public List<AITool> CreateHandoffFunctions(bool needsEndTask)
 
             static void Terminate()
             {
-                if (FunctionInvokingChatClient.CurrentContext is not { } ctx)
+                if (NewFunctionInvokingChatClient.CurrentContext is not { } ctx)
```
stephentoub, line 219: @westey-m / @rogerbarreto, when can we delete this "NewFICC"?

stephentoub (approved):

stephentoub:
Thanks!

dotnet/src/Microsoft.Agents.Orchestration/Handoffs/HandoffOrchestration.cs
westey-m: As soon as the MEAI one is available in a nuget we can remove it from this repo.  Also, it's only really required right now where we are showing approvals samples.

TaoChenOSU (approved):

PR #578, opened 2025-09-01
.NET: Organize the .Net samples

markwallace-microsoft:
### Motivation and Context

- Remove some obsolete samples
- Rename and organise the remaining samples to make then easier to navigate

### Description

image

### Contribution Checklist


SergeyMenshykh:
LGTM, just a few minor suggestions:  
- Remove the `Step0*` prefix from the samples in the `Agents` folder.  
- Rename the `Agents` folder to `Concepts`.  
- The samples in the `AgentProviders` folder use the term `AIAgent`, while those in the `Agents` folder use `Agent`. Is this intentional? Can we use a one term? Or should we consider a different structure?
  - Providers
    - A2A
    - AzureFoundry
    - AzureOpenAIChatCompletion
    - AzureOpenAIResponses
    - ... 
  - Concepts
    - Running
    - MultiturnConversation
    - FunctionTools
    - FunctionToolsWithApprovals
    - ...

rogerbarreto:
I like the structure @SergeyMenshykh suggested.

But I still think for a getting started structure having stepping guide is good for newcomers.

My suggestion assuming we stick with `Steps` is.

The `steps` projects today start with Step01 as the prefix, this is inconsistent with the other naming and also mixes the IDE list of projects when running and building. 

I suggest moving the `StepNN` to the second name, leaving the `Folder` name as the `prefix` of the project similar how was done in the SKMigration, this ensure the step ordering keeps intact and from the project listing we can easily understand what is the project we are running.

i.e:
`Agents/Step01_Agents_*` -> `Agents/Agents_Step01_*`
`AgentProviders/AIAgent_With_*` -> `AgentProviders/AgentProviders_With_*`

markwallace-microsoft:
We need to consider the experience for someone who is seeing the repo for the first time.

I think there are problems with the above suggestions

- `GettingStarted\Concepts` - What do the concepts relate to? Is it Agents or Workflows or Agent Orchestrations. We need to add additional getting started samples for Workflows and Agent Orchestrations later so the naming for the agents getting started samples must be clear.
- `GettingStarted\Providers` - What is a provider? For someone new to the repository this won't be obvious. Using `AgentProviders` eliminates this source of confusion.
- `GettingStarted\Steps` - What do the steps refer to? Where will the getting started steps for workflows and orchestrations go? Step is a term used in workflows so this may be confusing. The projects have step in the name so there is no need to duplicate this.

PR #612, opened 2025-09-04
.NET: AI agent as an MCP tool

SergeyMenshykh:
This PR adds a sample demonstrating how an AIAgent can be exposed as an MCP tool from an MCP server.

### DIFF
--- a/docs/decisions/0005-python-naming-conventions.md
+++ b/docs/decisions/0005-python-naming-conventions.md
@@ -1,7 +1,7 @@
 ---
-status: proposed
+status: accepted
 contact: eavanvalkenburg
-date: 2025-09-03
+date: 2025-09-04
 deciders: markwallace-microsoft, dmytrostruk, peterychang, ekzhu, sphenry
 consulted: taochenosu, alliscode, moonbox3, johanste
 ---
@@ -49,6 +49,9 @@ The table below represents the majority of the naming changes discussed in issue
 | ChatFinishReason | FinishReason | accepted | More concise while still clear in context. | None |
 | AIContent | BaseContent | accepted | More accurate as it serves as the base class for all content types. | Content, too generic. |
 | AIContents | Contents | accepted | This is the annotated typing object that is the union of all concrete content types, so plural makes sense and since this is used as a type hint, the generic nature of the name is acceptable. | None |
+| AIAnnotations | Annotations | accepted | In sync with contents | None |
+| AIAnnotation | BaseAnnotation | accepted | In sync with contents | None |
+| *Mcp* & *Http* | *MCP* & *HTTP* | accepted | Acronyms should be uppercased in class names, according to PEP 8. | None |
 | `agent.run_streaming` | `agent.run_stream` | accepted | Shorter and more closely aligns with AutoGen and Semantic Kernel names for the same methods. | None |
 | `workflow.run_streaming` | `workflow.run_stream` | accepted | In sync with `agent.run_stream` and shorter and more closely aligns with AutoGen and Semantic Kernel names for the same methods. | None |
 | AgentRunResponse & AgentRunResponseUpdate | AgentResponse & AgentResponseUpdate | rejected | Rejected, because it is the response to a run invocation and AgentResponse is too generic. | None |
--- a/docs/docs-templates/user-guide/agent-types.md
+++ b/docs/docs-templates/user-guide/agent-types.md
@@ -1,5 +1,6 @@
 # Microsoft Agent Framework Agent Types
 
+::: zone pivot="programming-language-csharp"
 The Microsoft Agent Framework provides support for several types of agents to accommodate different use cases and requirements.
 
 All agents are derived from a common base class, `AIAgent`, which provides a consistent interface for all agent types. This allows for building common, agent agnostic, higher level functionality such as multi-agent orchestrations.
@@ -20,7 +21,6 @@ These agents support a wide range of functionality out of the box:
 
 To create one of these agents, simply construct a `ChatClientAgent` using the ChatClient implementation of your choice.
 
-::: zone pivot="programming-language-csharp"
 
 ```csharp
 using Microsoft.Extensions.AI;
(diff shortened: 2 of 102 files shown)


dotnet/samples/GettingStarted/Agents/Agent_Step10_AsMcpTool/AIAgentExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// <param name="agent">The agent to expose as an MCP tool.</param>
+    /// <param name="title">A human-readable title for the tool that can be displayed to users. If not provided, the tool name or agent name will be used.</param>
+    /// <param name="name">The tool name to use. If not provided, the agent's name will be used.</param>
+    /// <param name="description">The tool description to use. If not provided, the agent's description will be used.</param>
+    /// <returns>The <see cref="McpServerTool"/> that wraps the agent.</returns>
+    public static McpServerTool AsMcpTool(this AIAgent agent, string? title = null, string? name = null, string? description = null)
+    {
+        async Task<string> RunAgentAsync(
+            [Description("Available information that will guide in performing this operation.")] string query,
+            [Description("Optional agent thread identifier.")] string? threadId = null,
```
stephentoub: This parameter is strange to me. The arguments for these are going to be supplied by the LLM. How would it construct a thread ID to use?

SergeyMenshykh: Right, the idea was that the function would be part of a scenario where one of the previous functions would return a conversation/thread id, and LLM would use it when calling this function. I haven't heard about this scenario, though, so I dropped the parameter for now.

dotnet/samples/GettingStarted/Agents/Agent_Step10_AsMcpTool/AIAgentExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+{
+    /// <summary>
+    /// Exposes an <see cref="AIAgent"/> as a <see cref="McpServerTool"/> that can be registered with an MCP server.
+    /// </summary>
+    /// <param name="agent">The agent to expose as an MCP tool.</param>
+    /// <param name="title">A human-readable title for the tool that can be displayed to users. If not provided, the tool name or agent name will be used.</param>
+    /// <param name="name">The tool name to use. If not provided, the agent's name will be used.</param>
+    /// <param name="description">The tool description to use. If not provided, the agent's description will be used.</param>
+    /// <returns>The <see cref="McpServerTool"/> that wraps the agent.</returns>
+    public static McpServerTool AsMcpTool(this AIAgent agent, string? title = null, string? name = null, string? description = null)
```
rogerbarreto, line 24: The name of the tool will be `run_async` ? `run` ? curious here.

If thats the case might be better to provide what will be the toolName in the `AsMcpTool` extension to make this more customizable 

Or a specific extension that accepts the `McpServerToolCreationOptions` directly.

```suggestion
    public static McpServerTool AsMcpTool(this AIAgent agent, McpServerToolCreateOptions options)
        => McpServerTool.Create(RunAgentAsync, options);
    
    public static McpServerTool AsMcpTool(this AIAgent agent, string? toolName, string? title = null, string? name = null, string? description = null)
```

rogerbarreto, line 24: Is this also limited to the non-streaming tool, is it possible to also expose a streaming tool?

dotnet/samples/GettingStarted/Agents/Agent_Step10_AsMcpTool/AIAgentExtensions.cs
markwallace-microsoft: The `threadId` would most likely be provided by the application code. In SK a developer could write a auto function invocation filter and use that to inject additional properties. Given this is just a sample this is fine but I expect developers will want to know how they would provide a thread id if needed.

dotnet/samples/GettingStarted/Agents/Agent_Step10_AsMcpTool/AIAgentExtensions.cs
SergeyMenshykh: The tool name is determined by this heuristic: `Name = name ?? agent.Name`. By default, if no name override is provided, it will fall back to the agent's name. In this example, the tool name will be `Joker`.   
  
Considering that this is a sample, I would keep it simple for now by keeping this implementation and add an additional extension method later, if/when we decide to release this extension.

SergeyMenshykh: Not sure about the whole call chain - MEAI Chat Client -> MCP Server -> MCP Tool supports streaming tools. Maybe chat clients in manual mode can stream function call results if called via GetResponseStreaming, but I'm not sure about auto mode and support from the MCP server side. Anyway, let's keep it simple and revisit this sample if/when it's needed or possible.

stephentoub: MCP currently has no concept of streaming tools. It's something being looked at for a future spec, though the value is a little niche, as the whole operation is effectively stuck on the results of the tool propagating back to the LLM as part of a subsequent request, where the input won't be streaming. Effectively a streaming MCP tool doesn't have a material impact on the end consumer's perceived latency.

dotnet/samples/GettingStarted/Agents/Agent_Step10_AsMcpTool/AIAgentExtensions.cs
stephentoub: There are lots of ways additional information could be propagated into the MCP tool, e.g. via DI. I don't think relying on the LLM to know about or provide a thread ID is desirable, though. Fine to be proven wrong, but I don't think we should show it in samples right now; thanks for removing it.

PR #619, opened 2025-09-05
.NET: Address RunAsync Extensions with ChatOptions

### LINKED ISSUE #521: .NET: Passing ChatClientAgent specific run options
Today we have multiple approaches to passing ChatClientAgent specific run options:

1. `RunAsync`, which takes `AgentRunOptions? options` as input, can be passed a `ChatClientAgentRunOptions`
2. We have a `RunAsync` overload in `ChatClientAgentExtensions` which takes `AgentRunOptions? agentRunOptions` and `ChatOptions? chatOptions` as input.
    1. Since `ChatClientAgentRunOptions` is a subclass of `AgentRunOptions`, it can in theory be passed to this overload.  `ChatClientAgentRunOptions` contains a ChatOptions inside of it, so it's unclear what would happen if someone passed both `ChatClientAgentRunOptions` and `ChatOptions`.

Since the 2nd method has some ambiguity, I propose that we remove it and only keep the original RunAsync method.

rogerbarreto:
### Motivation and Context

- Move `ChatCLientAgentRunOptions` to be internal
- Promote `ChatOptions.Tools` to be part of `AgentRunOptions.Tools`
- Promote `ChatOptions.Instructions` to be part of `AgentRunOptions.Instructions`
- Update `ChatClientAgent` extensions to be `AIAgentExtensions` instead, accepting both `AgentRunOptions` and `ChatOptions` in the parameters
    When both are provided, `Instructions` and `Tools` are merged in the internalized `ChatClientAgentRunOptions`.
- Resolves #521 as the merge is handled internally and the there are no more way on having dual purpose `ChatClient` options in the extensions. 

During this change was identified several external points in our samples and Orchestration that were relying in the `ChatCLientAgentRunOptions` type while using the `AIAgent` abstraction directly, pointing to a clear need to bring this to a lower level without disruption or added dependencies.

The proposed changes simplifies the overall experience for the caller when working with agents.

### DIFF
--- a/dotnet/tests/Microsoft.Extensions.AI.Agents.UnitTests/ChatCompletion/ChatClientAgentTests.cs
+++ b/dotnet/tests/Microsoft.Extensions.AI.Agents.UnitTests/ChatCompletion/ChatClientAgentTests.cs
@@ -1524,6 +1524,373 @@ public class ChatClientAgentTests
 
     #endregion
 
+    #region AgentRunOptions Conditional Logic Tests
+
+    /// <summary>
+    /// Verify that RunAsync with AgentRunOptions containing Instructions transfers instructions to requestChatOptions.
+    /// </summary>
+    [Fact]
+    public async Task RunAsyncWithAgentRunOptionsInstructionsTransfersInstructionsToRequestChatOptionsAsync()
+    {
+        // Arrange
+        const string TestInstructions = "Test instructions from AgentRunOptions";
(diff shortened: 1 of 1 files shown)


rogerbarreto:
I'm closing this as this actually is trying to simplify by not exposing the specialization brings even more ambiguity between `AgentRunOptions.Tools` and `ChatOptions.Tools`.

PR #636, opened 2025-09-08
.NET: Fix handoff resumption with no next agent specified

stephentoub:
Fixes https://github.com/microsoft/agent-framework/issues/566

(We don't currently have any tests that exercise orchestrations with a runtime. We need to fix that, but separately.)


stephentoub:
Just noticed this is already being fixed in https://github.com/microsoft/agent-framework/pull/568

PR #639, opened 2025-09-08
.NET: Add Ollama and custom agent samples, plus ONNX improvements

westey-m:
Add Ollama and custom agent samples, plus ONNX improvements

### Contribution Checklist

### DIFF
--- a/dotnet/samples/GettingStarted/AgentProviders/Agent_With_Ollama/README.md
+++ b/dotnet/samples/GettingStarted/AgentProviders/Agent_With_Ollama/README.md
@@ -23,12 +23,12 @@ After the container has started, launch a Terminal window for the docker contain
 From this terminal download the required models, e.g. here we are downloading the phi3 model.
 
 ```text
-ollama pull phi3
+ollama pull gpt-oss
 ```
 
 Set the following environment variables:
 
 ```powershell
 $env:OLLAMA_ENDPOINT="http://localhost:11434"
-$env:OLLAMA_MODEL_NAME="phi3"
+$env:OLLAMA_MODEL_NAME="gpt-oss"
 ```


dotnet/samples/GettingStarted/AgentProviders/Agent_With_CustomImplementation/README.md
```diff
@@ -0,0 +1,3 @@
+# Prerequisites
+
+This sample has no prerequisites.
```
rogerbarreto, line 3: nit.

```suggestion
```

dotnet/samples/GettingStarted/AgentProviders/Agent_With_Ollama/Program.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+// This sample shows how to create and use a simple AI agent with Ollama as the backend.
+
+using System;
+using Microsoft.Extensions.AI.Agents;
+using OllamaSharp;
+
+var endpoint = Environment.GetEnvironmentVariable("OLLAMA_ENDPOINT") ?? throw new InvalidOperationException("OLLAMA_ENDPOINT is not set.");
```
rogerbarreto, line 9: Running Ollama is pretty default for localhost:11434, we can make it optional.

```suggestion
var endpoint = System.Environment.GetEnvironmentVariable("OLLAMA_ENDPOINT") ?? "http://localhost:11434";
```

dotnet/samples/GettingStarted/AgentProviders/Agent_With_Ollama/README.md
```diff
@@ -0,0 +1,34 @@
...
+```powershell
+docker run -d --gpus=all -v "c:\temp\ollama:/root/.ollama" -p 11434:11434 --name ollama ollama/ollama
+```
+
+After the container has started, launch a Terminal window for the docker container, e.g. if using docker desktop, choose Open in Terminal from actions.
+
+From this terminal download the required models, e.g. here we are downloading the phi3 model.
+
+```text
+ollama pull phi3
```
rogerbarreto: Ideally lets suggest a model that supports function calling.

OpenAI Open Source seems the best use case for now.

```suggestion
ollama pull gpt-oss
```

dotnet/samples/GettingStarted/AgentProviders/Agent_With_CustomImplementation/Program.cs
```diff
@@ -0,0 +1,96 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.AI;
+using Microsoft.Extensions.AI.Agents;
+using SampleApp;
+
+AIAgent agent = new UpperCaseParrotAgent();
+
+// Invoke the agent and output the text result.
+Console.WriteLine(await agent.RunAsync("Tell me a joke about a pirate."));
```
rogerbarreto, line 20: Suggest adding the streaming example, as we do have the implementation for it.

dotnet/samples/GettingStarted/AgentProviders/Agent_With_CustomImplementation/README.md
westey-m: I think it's good to have the file, calling out the lack of prerequisites, so that it doesn't look like the missing file is an oversight.

dotnet/samples/GettingStarted/AgentProviders/Agent_With_CustomImplementation/Program.cs
westey-m: Thanks, makes sense.

dotnet/samples/GettingStarted/AgentProviders/Agent_With_Ollama/README.md
westey-m: I actually wanted to use something other than just gpt everywhere.  Tried phi4-mini which does have function calling support, but it wasn't working for me :(
So ended up using gpt-oss after all, but it would be nice to use something else here if we can find a model that works.

dotnet/samples/GettingStarted/AgentProviders/Agent_With_Ollama/Program.cs
westey-m: Trying to keep things fairly explicit, so users know what is happening, so keeping it as is for now.

PR #662, opened 2025-09-09
.NET: Using Protocol Methods to append Agent Framework Headers

rogerbarreto:
### Motivation and Context

Validate the approach on using the Protocol option to append `User-Agent` and `Version` headers.

This method requires some reflection as is (because the types are internal), the good news is that the implementation for allowing custom headers is majorily related to the the `AzureAI` package and this won't be an issue.

image


stephentoub:
This is all because it's not possible to add to the PersistentAgentsClient.Administration.Pipeline once it's been created and handed off? Can we work with the SDK folks to make that possible instead, or some other more official way to augment it without needing reflection hacks? They're almost guaranteed to break.

rogerbarreto:
> This is all because it's not possible to add to the PersistentAgentsClient.Administration.Pipeline once it's been created and handed off?

That's right, my assumption is that this is by design.

> Can we work with the SDK folks to make that possible instead 

Yes, I will check further why the `Pipeline` design is such as it cannot be modified. 

> Or some other more official way to augment it without needing reflection hacks? They're almost guaranteed to break.

As far I understand using the protocol method is the official way, I'm using reflection here because the protocol implementation from the nuget package is basically all internal, but in our case once this implementation moves to the `Azure.AI.Agents.Persistent` package, the code won't need reflection and this would be a much more cleaner impl.

PR #678, opened 2025-09-09
Python: Add actor abstractions for Python runtime

adityamandaleeka:
This change brings in a set of abstractions for the python version of the actor runtime, including request/response message types and the runtime context interface for runtime operations (state management, message handling).


markwallace-microsoft:
<a href="https://github.com/microsoft/agent-framework/blob/01c3960167b27543e4a76f871287a737c7b43bd3/README.md">Python Test Coverage</a><details><summary>Python Test Coverage Report • </summary><table><tr><th>File</th><th>Stmts</th><th>Miss</th><th>Cover</th><th>Missing</th></tr><tbody><tr><td><b>TOTAL</b></td><td><b>7753</b></td><td><b>1254</b></td><td><b>83%</b></td><td>&nbsp;</td></tr></tbody></table><i>report-only-changed-files is enabled. No files were changed during this commit :)</i></details>

## Python Unit Test Overview
| 892 | 94 :zzz: | 0 :x: | 0 :fire: | 24.842s :stopwatch: |

python/packages/runtime/agent_framework_runtime/agent_actor.py
```diff
@@ -18,9 +26,233 @@ def __str__(self) -> str:
...
+                await self._handle_agent_request(message, context)
+
+    async def _restore_thread_state(self, context: ActorRuntimeContext):
+        """Restore the agent thread state from storage."""
+        thread_data = await context.read_state(self.THREAD_STATE_KEY)
+
+        if thread_data:
+            try:
+                if hasattr(self._agent, "deserialize_thread"):
+                    self._thread = await self._agent.deserialize_thread(thread_data)  # type: ignore[attr-defined]
```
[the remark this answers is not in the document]
adityamandaleeka: Changed to try/catch approach.

python/packages/runtime/agent_framework_runtime/agent_actor.py
```diff
@@ -18,9 +26,233 @@ def __str__(self) -> str:
...
+class ActorRequestMessage(_ActorMessage):
+    """Request message sent to an actor."""
+
+    method: str = ""
+    params: dict[str, Any] | None = None
+    sender_id: ActorId | None = None
+
+    def __post_init__(self):
+        """Initialize the message type after dataclass creation."""
+        self.message_type = ActorMessageType.REQUEST
```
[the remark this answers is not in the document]
adityamandaleeka: I wonder if it's more pythonic to get rid of this entirely and use the type itself, e.g.

```python
  if isinstance(message, ActorRequestMessage):
      # handle request
  elif isinstance(message, ActorResponseMessage):
      # handle response
```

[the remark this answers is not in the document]
johanste: You can use `if message.kind == 'request'`. But, yeah, trying to encode the type/discriminator as a read/write value in the base actually causes typing problems here... - e.g. there is nothing stopping me from doing

```python
message = ActorRequestMessage(...)
message.message_type = ActorMessageType.RESPONSE
```

(and be completely within the constraints set up by typing)

I'd say we can remove the base class and type the discriminator (`message_type`) with the literal value. For instances where it can be either a request or response message, we can type it as a union of the two.

python/packages/runtime/agent_framework_runtime/agent_actor.py
```diff
@@ -18,9 +26,233 @@ def __str__(self) -> str:
...
+    """Runtime context provided to actors."""
+
+    @property
+    @abstractmethod
+    def actor_id(self) -> ActorId:
+        """Get the actor's unique identifier."""
+        pass
+
+    @abstractmethod
+    def watch_messages(self) -> AsyncIterator[ActorMessage]:
```
johanste, line 94:
```suggestion
    def iter_messages(self) -> AsyncIterator[ActorMessage]:
```
?

python/packages/runtime/agent_framework_runtime/agent_actor.py
```diff
@@ -18,9 +26,233 @@ def __str__(self) -> str:
...
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    async def read_state(self, key: str) -> Any | None:
+        """Read state value by key."""
+        pass
+
+    @abstractmethod
+    async def write_state(self, key: str, value: Any) -> bool:
```
johanste, line 116: `raise` an exception rather than returning a `bool`.

python/packages/runtime/agent_framework_runtime/agent_actor.py
```diff
@@ -18,9 +26,233 @@ def __str__(self) -> str:
...
+        """Write state value by key."""
+        pass
+
+    @abstractmethod
+    def complete_request(self, message_id: str, response: ActorResponseMessage) -> None:
```
johanste, line 121: Consider using a `done` suffix (mimicking standard `Queue.task_done()` naming). Not a strong opinion on this...
 
```suggestion
    def request_done(self, message_id: str, response: ActorResponseMessage) -> None:
```

python/packages/runtime/agent_framework_runtime/agent_actor.py
```diff
@@ -18,9 +26,233 @@ def __str__(self) -> str:
...
+        """Complete a request with a response."""
+        pass
+
+    def on_progress_update(
```
johanste, line 125: This name looks optimized for implementors of the class - but the notifier of the update is external, right? If so, `update_progress` seem to be a better name for the caller.

python/packages/runtime/agent_framework_runtime/agent_actor.py
```diff
@@ -18,9 +26,233 @@ def __str__(self) -> str:
...
+class Actor(ABC):
+    """Interface for all actors in the system (runtime infrastructure)."""
+
+    @abstractmethod
+    async def run(self, context: ActorRuntimeContext) -> None:
+        """Main actor execution loop."""
+        pass
+
+
+class AgentRunRequest(BaseModel):
```
johanste, line 146: I must admit that I have concerns about potentially gratuitous use of pydantic. It will have breaking changes, and the runtime is not in a good position to lock down which version to use even at the major version level.

I would consider making this a data class.

python/packages/runtime/agent_framework_runtime/agent_actor.py
```diff
@@ -18,9 +26,233 @@ def __str__(self) -> str:
...
+                await self._handle_agent_request(message, context)
+
+    async def _restore_thread_state(self, context: ActorRuntimeContext):
+        """Restore the agent thread state from storage."""
+        thread_data = await context.read_state(self.THREAD_STATE_KEY)
+
+        if thread_data:
+            try:
+                # Try to deserialize - if the method doesn't exist, we'll get AttributeError
+                self._thread = await self._agent.deserialize_thread(thread_data)  # type: ignore[attr-defined]
```
johanste, line 187: We should not rely on AttributeError here. It can leak out from an incorrectly implemented agent and subsequently swallowed by the fallback...

Can we merge the deserialize and creation of an existing thread into a single `agent.get_thread(previous)` where an implementation is free to ignore the previously saved data?

python/packages/runtime/agent_framework_runtime/agent_actor.py
```diff
@@ -18,9 +26,233 @@ def __str__(self) -> str:
...
+        try:
+            logger.info(f"Processing agent request: {request.message_id}")
+
+            if not request.params:
+                raise ValueError("Request params are required")
+
+            run_request = AgentRunRequest.model_validate(request.params)
+
+            # Ensure a thread exists (may be None if _restore_thread_state not yet called in some test paths)
+            if self._thread is None:
```
johanste, line 215: I'd prefer to set the _thread instance in the test code rather than possibly masking a logic bug here.

adityamandaleeka:
Based on a discussion today, I'm going to withdraw this for now. It seems like we may be pursuing another runtime design approach which won't necessitate having the actor model expressed in Python at all.

PR #684, opened 2025-09-10
Python: Introduce as_tool() for BaseAgent

ekzhu:
### Motivation and Context

To support agent-as-tool hierarchical orchestration pattern ("connected agent"). 

### Description

Usage: 

```python
import asyncio

from agent_framework import ChatAgent


async def main() -> None:
    """Main example showing ChatAgents as tools for translation."""
    
    # Create specialized translation agents
    spanish_agent = ChatAgent(
        chat_client=OpenAIChatClient(),
        name="SpanishTranslator",
        description="Translates text to Spanish",
        instructions="You are a Spanish translator. Translate the given text to Spanish and respond with only the translation."
    )
    
    japanese_agent = ChatAgent(
        chat_client=OpenAIChatClient(), 
        name="JapaneseTranslator",
        description="Translates text to Japanese",
        instructions="You are a Japanese translator. Translate the given text to Japanese and respond with only the translation."
    )
    
    # Create main agent that uses translation agents as tools
    translator_coordinator = ChatAgent(
        chat_client=OpenAIChatClient(),
        name="TranslatorCoordinator",
        instructions="""You are a translation coordinator. You can translate text to Spanish or Japanese using your translation tools.
        
        When users ask for translations:
        - Use the SpanishTranslator tool for Spanish translations
        - Use the JapaneseTranslator tool for Japanese translations
        - Provide clear responses showing the original and translated text""",
        tools=[
            spanish_agent.as_tool(description="Translate text to Spanish"),
            japanese_agent.as_tool(description="Translate text to Japanese")
        ]
    )
    
    # Example conversations
    print("=== ChatAgent as Tool Translation Example ===\n")
    
    # Example 1: Spanish translation
    user_input1 = "Please translate 'Hello, how are you?' to Spanish"
    print(f"User: {user_input1}")
    result1 = await translator_coordinator.run(user_input1)
    print(f"Coordinator: {result1}\n")

if __name__ == "__main__":
    asyncio.run(main())

```

### Contribution Checklist

### DIFF
--- a/python/packages/main/agent_framework/_agents.py
+++ b/python/packages/main/agent_framework/_agents.py
@@ -210,7 +210,7 @@ class BaseAgent(AFBaseModel):
 
         # Create dynamic input model with the specified argument name
         field_info = Field(..., description=argument_description)
-        input_model = create_model(f"{name or self.name or "agent"}_task", **{arg_name: (str, field_info)})  # type: ignore[call-overload]
+        input_model = create_model(f"{name or self.name or 'agent'}_task", **{arg_name: (str, field_info)})  # type: ignore[call-overload]
 
         # Check if callback is async once, outside the wrapper
         is_async_callback = stream_callback is not None and inspect.iscoroutinefunction(stream_callback)


markwallace-microsoft:
<a href="https://github.com/microsoft/agent-framework/blob/084bfbef993591ad614abf8b033e58d2df0d0239/README.md">Python Test Coverage</a><details><summary>Python Test Coverage Report • </summary><table><tr><th>File</th><th>Stmts</th><th>Miss</th><th>Cover</th><th>Missing</th></tr><tbody><tr><td colspan="5"><b>packages/main/agent_framework</b></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/084bfbef993591ad614abf8b033e58d2df0d0239/packages/main/agent_framework/_agents.py">_agents.py</a></td><td>210</td><td>26</td><td>87%</td><td><a href="https://github.com/microsoft/agent-framework/blob/084bfbef993591ad614abf8b033e58d2df0d0239/packages/main/agent_framework/_agents.py#L173-L175">173&ndash;175</a>, <a href="https://github.com/microsoft/agent-framework/blob/084bfbef993591ad614abf8b033e58d2df0d0239/packages/main/agent_framework/_agents.py#L203">203</a>, <a href="https://github.com/microsoft/agent-framework/blob/084bfbef993591ad614abf8b033e58d2df0d0239/packages/main/agent_framework/_agents.py#L246">246</a>, <a href="https://github.com/microsoft/agent-framework/blob/084bfbef993591ad614abf8b033e58d2df0d0239/packages/main/agent_framework/_agents.py#L344">344</a>, <a href="https://github.com/microsoft/agent-framework/blob/084bfbef993591ad614abf8b033e58d2df0d0239/packages/main/agent_framework/_agents.py#L402-L404">402&ndash;404</a>, <a href="https://github.com/microsoft/agent-framework/blob/084bfbef993591ad614abf8b033e58d2df0d0239/packages/main/agent_framework/_agents.py#L406-L409">406&ndash;409</a>, <a href="https://github.com/microsoft/agent-framework/blob/084bfbef993591ad614abf8b033e58d2df0d0239/packages/main/agent_framework/_agents.py#L416">416</a>, <a href="https://github.com/microsoft/agent-framework/blob/084bfbef993591ad614abf8b033e58d2df0d0239/packages/main/agent_framework/_agents.py#L426">426</a>, <a href="https://github.com/microsoft/agent-framework/blob/084bfbef993591ad614abf8b033e58d2df0d0239/packages/main/agent_framework/_agents.py#L499-L500">499&ndash;500</a>, <a href="https://github.com/microsoft/agent-framework/blob/084bfbef993591ad614abf8b033e58d2df0d0239/packages/main/agent_framework/_agents.py#L502">502</a>, <a href="https://github.com/microsoft/agent-framework/blob/084bfbef993591ad614abf8b033e58d2df0d0239/packages/main/agent_framework/_agents.py#L505">505</a>, <a href="https://github.com/microsoft/agent-framework/blob/084bfbef993591ad614abf8b033e58d2df0d0239/packages/main/agent_framework/_agents.py#L628-L629">628&ndash;629</a>, <a href="https://github.com/microsoft/agent-framework/blob/084bfbef993591ad614abf8b033e58d2df0d0239/packages/main/agent_framework/_agents.py#L631">631</a>, <a href="https://github.com/microsoft/agent-framework/blob/084bfbef993591ad614abf8b033e58d2df0d0239/packages/main/agent_framework/_agents.py#L634">634</a>, <a href="https://github.com/microsoft/agent-framework/blob/084bfbef993591ad614abf8b033e58d2df0d0239/packages/main/agent_framework/_agents.py#L693">693</a>, <a href="https://github.com/microsoft/agent-framework/blob/084bfbef993591ad614abf8b033e58d2df0d0239/packages/main/agent_framework/_agents.py#L725">725</a>, <a href="https://github.com/microsoft/agent-framework/blob/084bfbef993591ad614abf8b033e58d2df0d0239/packages/main/agent_framework/_agents.py#L770">770</a></td></tr><tr><td><b>TOTAL</b></td><td><b>8544</b></td><td><b>1385</b></td><td><b>83%</b></td><td>&nbsp;</td></tr></tbody></table></details>

## Python Unit Test Overview
| 991 | 94 :zzz: | 0 :x: | 0 :fire: | 26.657s :stopwatch: |

python/packages/main/agent_framework/_protocols.py
```diff
@@ -0,0 +1,123 @@
...
+
+    Methods:
+        parameters: The parameters accepted by the tool, in a json schema format.
+    """
+
+    name: str
+    """The name of the tool."""
+    description: str
+    """A description of the tool, suitable for use in describing the purpose to a model."""
+    additional_properties: Union[dict[str, Any], None]
```
moonbox3: In the other file, it was `additional_properties: dict[str, Any] | None` (no `Union` type)

python/packages/main/agent_framework/_protocols.py
```diff
@@ -0,0 +1,123 @@
...
+        """Returns the name of the agent."""
+        ...
+
+    @property
+    def display_name(self) -> str:
+        """Returns the display name of the agent."""
+        ...
+
+    @property
+    def description(self) -> Union[str, None]:
```
moonbox3: Same here: was previously `str | None` in the other file.

python/packages/main/agent_framework/_tools.py
```diff
@@ -5,33 +5,30 @@
...
-if TYPE_CHECKING:
-    from ._types import Contents
-
 tracer: trace.Tracer = trace.get_tracer("agent_framework")
 meter: metrics.Meter = metrics.get_meter_provider().get_meter("agent_framework")
 logger = get_logger()
 
 __all__ = [
     "AIFunction",
+    "AgentTool",
```
markwallace-microsoft: Why do we need a dedicated tool type?

Can we not just convert an Agent into an `AIFunction`?

python/packages/main/agent_framework/_agents.py
```diff
@@ -166,6 +82,32 @@ async def deserialize_thread(self, serialized_thread: Any, **kwargs: Any) -> Age
...
+            An AgentTool instance wrapping this agent.
+        """
+        if not isinstance(self, AgentProtocol):
+            raise TypeError("Only agents implementing AgentProtocol can be converted to tools.")
+
+        return AgentTool(
+            agent=self,
+            description=description,
+            stream=stream,
+            stream_callback=stream_callback,
```
moonbox3: Can we treat `stream=True` as "use agent.run_stream()" regardless of callback presence? If no callback, just aggregate and return final text.

It feels a little clunky to potentially have a callback per `AgentTool`?

python/packages/main/agent_framework/_protocols.py
eavanvalkenburg, line 1: why do we need to split this out? I prefer to have the protocols living next to the implementations for internal, and the public API doesn't see this...

python/packages/main/agent_framework/_protocols.py
eavanvalkenburg: indeed, Union type is not needed, `|` will do

python/packages/main/agent_framework/_tools.py
```diff
@@ -252,6 +224,101 @@ def __init__(
         super().__init__(**args, **kwargs)
 
 
+class AgentTool(BaseTool):
```
eavanvalkenburg: this should at least be a subclass of AIFunction, because otherwise function calling will not be invoked automatically without a change there as well

python/packages/main/agent_framework/_agents.py
```diff
@@ -166,6 +82,32 @@ async def deserialize_thread(self, serialized_thread: Any, **kwargs: Any) -> Age
...
+
+        Args:
+            description: Optional description for the tool. If not provided, uses the agent's description.
+            stream: Whether to use streaming when the tool is invoked.
+            stream_callback: Optional callback function to handle streaming updates when stream=True.
+
+        Returns:
+            An AgentTool instance wrapping this agent.
+        """
+        if not isinstance(self, AgentProtocol):
```
eavanvalkenburg, line 202: since self is a BaseAgent there is no need to check this, and the AgentTool constructor also checks it

python/packages/main/agent_framework/_protocols.py
eavanvalkenburg: and if we are doing this, then ChatClientProtocol should also be here.

python/packages/main/tests/main/test_tools.py
```diff
@@ -507,3 +507,211 @@ def test_hosted_code_interpreter_tool_with_unknown_input():
     """Test HostedCodeInterpreterTool with single unknown input."""
     with pytest.raises(ValueError, match="Unsupported input type"):
         HostedCodeInterpreterTool(inputs={"hosted_file": "file-single"})
+
+
+# region AgentTool tests
+
+
+class MockAgent:
```
eavanvalkenburg: Let's do:
```suggestion
class MockAgent(BaseAgent):
```
This way the code is smaller and we actually also test the as_tool function in the BaseAgent.

python/packages/main/agent_framework/_agents.py
ekzhu: BaseAgent actually doesn't implements AgentProtocol. So this is to make the type checker happy.

python/packages/main/agent_framework/_tools.py
ekzhu: Ah! This is why I couldn't see to trigger the function call.

python/packages/main/agent_framework/_protocols.py
ekzhu: I am not sure how to avoid cyclical import if we don't separate the files:

- _agents.py imports from _tools.py to get tool definitions in `as_tool`
- _tools.py imports from _agents.py to get agent definitions in `AgentTool`.

python/packages/main/agent_framework/_agents.py
ekzhu: The scenario for callback is in many cases you want to display the underlying agent's intermediate responses. 

I think we can just remove `stream` and if `stream_callback` is not set.

python/packages/main/agent_framework/_protocols.py
ekzhu: Yes. There was some forward reference issue in a different place in this file because it imports from `_types.py` which also imports from this file. So I had to use `"ChatMessage"` instead of `ChatMessage`. It turns out you cannot use `|` notation with quoted type hints that are forward references.

python/packages/main/agent_framework/_tools.py
ekzhu: Good point.

python/packages/main/agent_framework/_protocols.py
moonbox3: Should be able to - just include the other type in the quote, too: `some_var: "SomeForwardRef | str | None" = None` as an example.

python/packages/main/agent_framework/_protocols.py
eavanvalkenburg: that's already the case, just have to use TYPE_CHECKING for type hints, and then import within a function, this is already happening now and works fine

python/packages/main/agent_framework/_tools.py
eavanvalkenburg: yeah, BaseTool is for all tool types, AIFunction is basically a invokable BaseTool, so this should be that as well

python/packages/main/agent_framework/_agents.py
eavanvalkenburg: yeah, I understand now, thanks

ekzhu:
@moonbox3 @markwallace-microsoft @eavanvalkenburg I have updated this PR to use AIFunction and now it is much simpler implementation. Please review again.

python/packages/main/agent_framework/_agents.py
```diff
@@ -170,6 +170,61 @@ async def deserialize_thread(self, serialized_thread: Any, **kwargs: Any) -> Age
...
+
+            if stream_callback is not None:
+                # Use streaming mode - accumulate updates and create final response
+                response_updates: list[AgentRunResponseUpdate] = []
+                async for update in self.run_stream(input_text):
+                    response_updates.append(update)
+                    stream_callback(update)
+
+                # Create final text from accumulated updates
+                return "".join(update.text for update in response_updates)
```
moonbox3: We have an aggregator for streaming, shall we use it?

```suggestion
                final_response = AgentRunResponse.from_agent_run_response_updates(response_updates)
                return final_response.text
```

python/packages/main/agent_framework/_agents.py
```diff
@@ -170,6 +170,61 @@ async def deserialize_thread(self, serialized_thread: Any, **kwargs: Any) -> Age
...
         return thread
 
+    def as_tool(
+        self,
+        name: str | None = None,
+        description: str | None = None,
+        arg_name: str = "input",
+        arg_description: str | None = None,
+        stream_callback: Callable[[AgentRunResponseUpdate], None] | None = None,
+    ) -> AIFunction[Any, str]:
```
moonbox3:
```suggestion
    ) -> AIFunction[BaseModel, str]
```

python/packages/main/agent_framework/_agents.py
```diff
@@ -170,6 +170,61 @@ async def deserialize_thread(self, serialized_thread: Any, **kwargs: Any) -> Age
...
+            """Wrapper function that calls the agent."""
+            # Extract the input from kwargs using the specified arg_name
+            input_text = kwargs.get(arg_name, "")
+
+            if stream_callback is not None:
+                # Use streaming mode - accumulate updates and create final response
+                response_updates: list[AgentRunResponseUpdate] = []
+                async for update in self.run_stream(input_text):
+                    response_updates.append(update)
+                    stream_callback(update)
```
moonbox3, line 234: To allow for the use of async callbacks, should this be:

```python
response_updates: list[AgentRunResponseUpdate] = []
async for update in self.run_stream(input_text):
    response_updates.append(update)
    callback_result = stream_callback(update)
    if inspect.isawaitable(callback_result):
        await callback_result
```

python/packages/main/agent_framework/_agents.py
```diff
@@ -170,6 +170,61 @@ async def deserialize_thread(self, serialized_thread: Any, **kwargs: Any) -> Age
         await deserialize_thread_state(thread, serialized_thread, **kwargs)
         return thread
 
+    def as_tool(
+        self,
+        name: str | None = None,
+        description: str | None = None,
+        arg_name: str = "input",
+        arg_description: str | None = None,
+        stream_callback: Callable[[AgentRunResponseUpdate], None] | None = None,
```
moonbox3: To allow for async streaming callbacks?

```suggestion
        stream_callback: Callable[[AgentRunResponseUpdate], Awaitable[Any] | Any] | None = None,
```

python/packages/main/agent_framework/_agents.py
```diff
@@ -170,6 +170,61 @@ async def deserialize_thread(self, serialized_thread: Any, **kwargs: Any) -> Age
...
+
+        tool_name = name or self.name
+        if tool_name is None:
+            raise ValueError("Agent tool name cannot be None. Either provide a name parameter or set the agent's name.")
+        tool_description = description or self.description or ""
+        argument_description = arg_description or f"Input for {self.display_name}"
+
+        # Create dynamic input model with the specified argument name
+        field_info = Field(..., description=argument_description)
+        input_model = create_model(f"{self.display_name}_Input", **{arg_name: (str, field_info)})  # type: ignore[call-overload]
```
moonbox3: Any reason we don't want to have this match what agents actually accept? 

```python
input_model = create_model(..., **{arg_name: (ChatMessage | str | list[ChatMessage] | list[str], field_info)})
```

ekzhu: Model tends to do badly (well, at least not great) when the input argument type is a union of complex nested types with multiple unions within it... So this is to keep it simple for the model.

python/packages/main/agent_framework/_agents.py
```diff
@@ -170,6 +171,69 @@ async def deserialize_thread(self, serialized_thread: Any, **kwargs: Any) -> Age
...
+            stream_callback: Optional callback for streaming responses. If provided, uses run_stream.
+
+        Returns:
+            An AIFunction that can be used as a tool by other agents.
+        """
+        # Verify that self implements AgentProtocol
+        if not isinstance(self, AgentProtocol):
+            raise TypeError(f"Agent {self.__class__.__name__} must implement AgentProtocol to be used as a tool")
+
+        tool_name = name or self.name
```
eavanvalkenburg, line 205: why not `self.display_name` as fallback?

python/packages/main/agent_framework/_agents.py
```diff
@@ -170,6 +171,69 @@ async def deserialize_thread(self, serialized_thread: Any, **kwargs: Any) -> Age
...
+        if tool_name is None:
+            raise ValueError("Agent tool name cannot be None. Either provide a name parameter or set the agent's name.")
+        tool_description = description or self.description or ""
+        argument_description = arg_description or f"Input for {self.display_name}"
+
+        # Create dynamic input model with the specified argument name
+        field_info = Field(..., description=argument_description)
+        input_model = create_model(f"{self.display_name}_Input", **{arg_name: (str, field_info)})  # type: ignore[call-overload]
```
eavanvalkenburg:
```suggestion
        input_model = create_model(f"{tool_name}_input", **{arg_name: (str, field_info)})  # type: ignore[call-overload]
```

python/packages/main/agent_framework/_agents.py
```diff
@@ -170,6 +171,69 @@ async def deserialize_thread(self, serialized_thread: Any, **kwargs: Any) -> Age
...
+        """
+        # Verify that self implements AgentProtocol
+        if not isinstance(self, AgentProtocol):
+            raise TypeError(f"Agent {self.__class__.__name__} must implement AgentProtocol to be used as a tool")
+
+        tool_name = name or self.name
+        if tool_name is None:
+            raise ValueError("Agent tool name cannot be None. Either provide a name parameter or set the agent's name.")
+        tool_description = description or self.description or ""
+        argument_description = arg_description or f"Input for {self.display_name}"
```
eavanvalkenburg:
```suggestion
        argument_description = arg_description or f"Input for {tool_name}"
```

python/packages/main/agent_framework/_agents.py
eavanvalkenburg: Given this is a Agent, shouldn't the parameter be `task` or `query` instead of `input`? That will probably work better for function calling? (i see it's configurable, just wondering what the best default is)

python/packages/main/agent_framework/_agents.py
```diff
@@ -170,6 +171,69 @@ async def deserialize_thread(self, serialized_thread: Any, **kwargs: Any) -> Age
...
+
+        # Check if callback is async once, outside the wrapper
+        is_async_callback = stream_callback is not None and inspect.iscoroutinefunction(stream_callback)
+
+        async def agent_wrapper(**kwargs: Any) -> str:
+            """Wrapper function that calls the agent."""
+            # Extract the input from kwargs using the specified arg_name
+            input_text = kwargs.get(arg_name, "")
+
+            if stream_callback is not None:
```
eavanvalkenburg: let's inverse this logic:

```suggestion
            if stream_callback is None: return await self.run(input_text)
```
and then deindent the rest

python/packages/main/agent_framework/_agents.py
```diff
@@ -170,6 +171,69 @@ async def deserialize_thread(self, serialized_thread: Any, **kwargs: Any) -> Age
         await deserialize_thread_state(thread, serialized_thread, **kwargs)
         return thread
 
+    def as_tool(
+        self,
+        name: str | None = None,
```
eavanvalkenburg, line 180: let's make this keyword args only, because the ordering is not particularly clear and there are a number
```suggestion
        *,
        name: str | None = None,
```

python/packages/main/agent_framework/_agents.py
ekzhu: Sure, task seems like a better choice.

python/packages/main/agent_framework/_agents.py
ekzhu: Because the fallback is ID, which is likely not a meaningful name to use.

eavanvalkenburg: will the LLM care? Since it becomes the name of the tool itself, not sure if it matters what the tool name is (obviously more descriptive is better) but a guid will prob work? could we test?

python/packages/main/agent_framework/_agents.py
```diff
@@ -173,6 +174,70 @@ async def deserialize_thread(self, serialized_thread: Any, **kwargs: Any) -> Age
...
+
+        tool_name = name or self.name
+        if tool_name is None:
+            raise ValueError("Agent tool name cannot be None. Either provide a name parameter or set the agent's name.")
+        tool_description = description or self.description or ""
+        argument_description = arg_description or f"Task for {tool_name}"
+
+        # Create dynamic input model with the specified argument name
+        field_info = Field(..., description=argument_description)
+        input_model = create_model(f"{self.display_name}_Task", **{arg_name: (str, field_info)})  # type: ignore[call-overload]
```
eavanvalkenburg: I would say for this, using ID would be worse:

```suggestion
        input_model = create_model(f"{name or self.name or "agent"}_task", **{arg_name: (str, field_info)})  # type: ignore[call-overload]
```

ekzhu: Good spotting. Agree.

ekzhu:
> will the LLM care? Since it becomes the name of the tool itself, not sure if it matters what the tool name is (obviously more descriptive is better) but a guid will prob work? could we test?

We should definitely test, that's what the benchmarks in the lab packages are for.

python/packages/main/agent_framework/_agents.py
ekzhu: I think we should test. Though it will be relaxing the constraint, so I think we can do it later.

PR #686, opened 2025-09-10
.NET: [BREAKING] Change implementation to use AsFunctionTool extension method to create a function from an agent

markwallace-microsoft:
Change implementation to use `AsFunctionTool` extension method to create a function from an agent to align with Python

### Contribution Checklist


dotnet/src/Microsoft.Extensions.AI.Agents/AgentExtensions.cs
```diff
@@ -24,4 +27,27 @@ public static OpenTelemetryAgent WithOpenTelemetry(this AIAgent agent, ILoggerFa
...
+        async Task<string> RunAgentAsync(string query, CancellationToken cancellationToken)
+        {
+            var response = await agent.RunAsync(query, cancellationToken: cancellationToken).ConfigureAwait(false);
+            return response.Text;
+        }
+
+        return AIFunctionFactory.Create(RunAgentAsync, options ?? new()
+        {
+            Name = agent.Name,
+            Description = agent.Description,
```
stephentoub: If any options are passed in, even if Name and Description aren't set, this won't use the agent's. Should it be more like:
```csharp
options ??= new();
options.Name ??= agent.Name;
options.Description ??= agent.Description;
return AIFunctionFactory.Create(RunAgentAsync, options);
```
?

dotnet/src/Microsoft.Extensions.AI.Agents/AgentExtensions.cs
```diff
@@ -24,4 +27,27 @@ public static OpenTelemetryAgent WithOpenTelemetry(this AIAgent agent, ILoggerFa
...
+    /// Creates a <see cref="AIFunction"/> that will invoke the provided Agent.
+    /// </summary>
+    /// <param name="agent">The <see cref="AIAgent" /> to be represented via the created <see cref="AIFunction"/>.</param>
+    /// <param name="options">Metadata to use to override defaults inferred from <paramref name="agent"/>.</param>
+    /// <returns>The created <see cref="AIFunction"/> for invoking the <see cref="AIAgent"/>.</returns>
+    public static AIFunction AsAIFunction(this AIAgent agent, AIFunctionFactoryOptions? options = null)
+    {
+        Throw.IfNull(agent);
+
+        async Task<string> RunAgentAsync(string query, CancellationToken cancellationToken)
```
stephentoub: Probably want to add `[Description(...)]` attributes to the method and the query parameter.

dotnet/samples/GettingStarted/Agents/README.md
```diff
@@ -37,6 +37,7 @@ Before you begin, ensure you have the following prerequisites:
 |[Dependency injection with a simple agent](./Agent_Step09_DependencyInjection/)|This sample demonstrates how to add and resolve an agent with a dependency injection container|
 |[Exposing a simple agent as MCP tool](./Agent_Step10_AsMcpTool/)|This sample demonstrates how to expose an agent as an MCP tool|
 |[Using images with a simple agent](./Agent_Step11_UsingImages/)|This sample demonstrates how to use image multi-modality with an AI agent|
+|[Exposing a simple agent a function tool](./Agent_Step12_AsFunctionTool/)|This sample demonstrates how to expose an agent as a function tool|
```
rogerbarreto, line 40:
```suggestion
|[Exposing a simple agent as a function tool](./Agent_Step12_AsFunctionTool/)|This sample demonstrates how to expose an agent as a function tool|
```

dotnet/src/Microsoft.Extensions.AI.Agents/AgentExtensions.cs
```diff
@@ -24,4 +28,30 @@ public static OpenTelemetryAgent WithOpenTelemetry(this AIAgent agent, ILoggerFa
...
         };
     }
+
+    /// <summary>
+    /// Creates a <see cref="AIFunction"/> that will invoke the provided Agent.
+    /// </summary>
+    /// <param name="agent">The <see cref="AIAgent" /> to be represented via the created <see cref="AIFunction"/>.</param>
+    /// <param name="options">Metadata to use to override defaults inferred from <paramref name="agent"/>.</param>
+    /// <returns>The created <see cref="AIFunction"/> for invoking the <see cref="AIAgent"/>.</returns>
+    public static AIFunction AsAIFunction(this AIAgent agent, AIFunctionFactoryOptions? options = null)
```
westey-m: Do we need to support threads that span multiple calls via an option?
And if so, should we have an AIFunction specialization that exposes that thread publicly, so that it can be retrieved and stored?

E.g.
```csharp
var tool = myAgent.AsAIFunction();
var toolThread = ((AIAgentFunctionTool)tool).AgentThread;
// store thread, reload it, etc.
var recreatedTool = myAgent.AsAIFunction(reloadedThread);
```

dotnet/samples/GettingStarted/Agents/Agent_Step12_AsFunctionTool/Program.cs
```diff
@@ -0,0 +1,34 @@
...
+[Description("Get the weather for a given location.")]
+static string GetWeather([Description("The location to get the weather for.")] string location)
+    => $"The weather in {location} is cloudy with a high of 15°C.";
+
+// Create the chat client and agent, and provide the function tool to the agent.
+AIAgent weatherAgent = new AzureOpenAIClient(
+    new Uri(endpoint),
+    new AzureCliCredential())
+     .GetChatClient(deploymentName)
+     .CreateAIAgent(instructions: "You answer questions about the weather.", name: "WeatherAgent", tools: [AIFunctionFactory.Create(GetWeather)]);
```
westey-m: Consider adding a description here, since that is used as the tool description, and it's good to demonstrate that.

dotnet/src/Microsoft.Extensions.AI.Agents/AgentExtensions.cs
```diff
@@ -24,4 +28,30 @@ public static OpenTelemetryAgent WithOpenTelemetry(this AIAgent agent, ILoggerFa
...
+            [Description("Available information that will guide the agent.")] string query,
+            CancellationToken cancellationToken)
+        {
+            var response = await agent.RunAsync(query, cancellationToken: cancellationToken).ConfigureAwait(false);
+            return response.Text;
+        }
+
+        options ??= new();
+        options.Name ??= agent.Name;
+        options.Description ??= agent.Description;
```
rogerbarreto: If name is not provided the method name will be `Run`. I think using the `Invoke` may work best if names are aligned while exposed to LLM or update the description with `Run` naming... 

```suggestion
        [Description("Invoke an agent to retrieve some information.")]
        async Task<string> InvokeAgentAsync(
            [Description("Input prompt to invoke to the agent.")] string inputPrompt,
            CancellationToken cancellationToken)
        {
            var response = await agent.RunAsync(query, cancellationToken: cancellationToken).ConfigureAwait(false);
            return response.Text;
        }

        options ??= new();
        options.Name ??= agent.Name;
        options.Description ??= agent.Description;
```

dotnet/src/Microsoft.Extensions.AI.Agents/AgentExtensions.cs
rogerbarreto: Are we going to have a streaming option as well?

rogerbarreto:
IMHO this feature may raises more questions than answers, I would suggest this as a demonstration only and not a feature of the `Microsoft.Extensions.AI.Agents` package.

dotnet/src/Microsoft.Extensions.AI.Agents/AgentExtensions.cs
markwallace-microsoft: No, do you think we need one?

rogerbarreto: Let's leave it out for now, thinking a bit deeper not sure how this really applies into a function calling scenario, as all the streaming of a function call needs to be processed before sent back to the LLM.

dotnet/src/Microsoft.Extensions.AI.Agents/AgentExtensions.cs
```diff
@@ -24,4 +28,31 @@ public static OpenTelemetryAgent WithOpenTelemetry(this AIAgent agent, ILoggerFa
...
+    /// <param name="options">Metadata to use to override defaults inferred from <paramref name="agent"/>.</param>
+    /// <param name="thread">The <see cref="AgentThread"/> to use for the function.</param>
+    /// <returns>The created <see cref="AIFunction"/> for invoking the <see cref="AIAgent"/>.</returns>
+    public static AIFunction AsAIFunction(this AIAgent agent, AIFunctionFactoryOptions? options = null, AgentThread? thread = null)
+    {
+        Throw.IfNull(agent);
+
+        [Description("Invoke an agent to retrieve some information.")]
+        async Task<string> InvokeAgentAsync(
+            [Description("Input prompt to invoke to the agent.")] string inputPrompt,
```
rogerbarreto: nit / typo
```suggestion
            [Description("Input prompt to invoke the agent.")] string inputPrompt,
```

PR #690, opened 2025-09-10
.NET: Add missing OpenAI `GetAIAgent` extensions

rogerbarreto:
### Motivation and Context

Adding missing extensions for creating `AIAgents` from already existing hosted {Azure}OpenAI Assistant agents, similar how it's currently implemented in the `Microsoft.Extensions.AI.AzureAI` package.

### DIFF
--- a/dotnet/src/Microsoft.Extensions.AI.Agents.OpenAI/Extensions/AssistantClientResultExtensions.cs
+++ b/dotnet/src/Microsoft.Extensions.AI.Agents.OpenAI/Extensions/AssistantClientResultExtensions.cs
@@ -3,14 +3,13 @@
 using System.ClientModel;
 using Microsoft.Extensions.AI;
 using Microsoft.Extensions.AI.Agents;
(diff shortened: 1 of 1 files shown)


dotnet/src/Microsoft.Extensions.AI.Agents.OpenAI/Extensions/AssistantClientResultExtensions.cs
```diff
@@ -0,0 +1,63 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.ClientModel;
+using Microsoft.Extensions.AI;
+using Microsoft.Extensions.AI.Agents;
+using OpenAI.Assistants;
+
+namespace OpenAI;
```
SergeyMenshykh: I wonder if the namespace should be `OpenAI.Assistants` instead. It seems the `Assistant` class lives there. I'm not sure about the `ClientResult` though.

PR #691, opened 2025-09-10
.NET: Add AIContextProvider support

westey-m:
### Motivation and Context

#664 
#665

### Description

- Adding the AIContextProvider and AIContext base types
- Adding an AggregateAIContextProvider
- Adding support to AgentThread for AIContextProvider.
- Updating ChatClientAgent to support AIContextProvider.
- Adding a very simple memory sample

### Contribution Checklist


dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIContext.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// </remarks>
+    public IList<ChatMessage>? Messages { get; set; }
+
+    /// <summary>
+    /// Gets or sets a list of functions/tools to make available to the AI model for the current invocation.
+    /// </summary>
+    /// <remarks>
+    /// These functions/tools will be transitory and only apply to the current invocation.
+    /// </remarks>
+    public IList<AIFunction>? AIFunctions { get; set; }
```
stephentoub: Shouldn't this be `IList<AITool>? Tools { get; set; }`?

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIContext.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// <remarks>
+    /// These messages will permanently be added to the chat history.
+    /// </remarks>
+    public IList<ChatMessage>? Messages { get; set; }
+
+    /// <summary>
+    /// Gets or sets a list of functions/tools to make available to the AI model for the current invocation.
+    /// </summary>
+    /// <remarks>
+    /// These functions/tools will be transitory and only apply to the current invocation.
```
stephentoub:
```suggestion
    /// These functions/tools will be transient and only apply to the current invocation.
```

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIContextProvider.cs
```diff
@@ -0,0 +1,68 @@
...
+    /// Called just before messages are added to the chat by any participant.
+    /// </summary>
+    /// <remarks>
+    /// Inheritors can use this method to update their context based on the new message.
+    /// </remarks>
+    /// <param name="newMessages">The new messages.</param>
+    /// <param name="agentThreadId">The ID of the <see cref="AgentThread"/>.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>A task that completes when the context has been updated.</returns>
+    public virtual Task MessagesAddingAsync(IReadOnlyCollection<ChatMessage> newMessages, string? agentThreadId, CancellationToken cancellationToken = default)
```
stephentoub: IEnumerable rather than IReadOnlyCollection?

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIContextProvider.cs
```diff
@@ -0,0 +1,68 @@
...
+    /// <summary>
+    /// Called just before the Model/Agent/etc. is invoked
+    /// Implementers can load any additional context required at this time,
+    /// and they should return any context that should be passed to the Model/Agent/etc.
+    /// </summary>
+    /// <param name="newMessages">The most recent messages that the Model/Agent/etc. is being invoked with.</param>
+    /// <param name="agentThreadId">The ID of the <see cref="AgentThread"/>.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>A task that completes when the context has been rendered and returned.</returns>
+    public abstract Task<AIContext> ModelInvokingAsync(IEnumerable<ChatMessage> newMessages, string? agentThreadId, CancellationToken cancellationToken = default);
```
stephentoub: "Model" doesn't seem like the right terminology here. There's no guarantee there's a "model" involved, or that only a model is involved.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIContextProvider.cs
```diff
@@ -0,0 +1,68 @@
...
+    /// <param name="jsonSerializerOptions">The JSON serialization options to use.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>A <see cref="JsonElement"/> representation of the object's state.</returns>
+    protected internal virtual Task<JsonElement> SerializeAsync(JsonSerializerOptions? jsonSerializerOptions = null, CancellationToken cancellationToken = default)
+    {
+        return Task.FromResult(default(JsonElement));
+    }
+
+    /// <summary>
+    /// Deserializes the state contained in the provided <see cref="JsonElement"/> into the properties on this object.
```
stephentoub, line 56: Is the expectation that this instance is fresh? Or if it already contains state, how is this supposed to behave?

dotnet/src/Microsoft.Extensions.AI.Agents/AggregateAIContextProvider.cs
```diff
@@ -0,0 +1,123 @@
...
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.AI.Agents;
+
+/// <summary>
+/// An <see cref="AIContextProvider"/> that aggregates other <see cref="AIContextProvider"/> instances and delegates events to them.
+/// </summary>
+public class AggregateAIContextProvider : AIContextProvider, IList<AIContextProvider>
```
stephentoub: Consider sealing this, at least for now

dotnet/src/Microsoft.Extensions.AI.Agents/AggregateAIContextProvider.cs
```diff
@@ -0,0 +1,123 @@
...
+
+    /// <inheritdoc />
+    public override async Task MessagesAddingAsync(IReadOnlyCollection<ChatMessage> newMessages, string? agentThreadId, CancellationToken cancellationToken = default)
+    {
+        if (this._providers.Count == 0)
+        {
+            return;
+        }
+
+        await Task.WhenAll(this._providers.Select(x => x.MessagesAddingAsync(newMessages, agentThreadId, cancellationToken))).ConfigureAwait(false);
```
stephentoub: This is introducing concurrency where there wasn't, such that the various providers may end up running in parallel, which would be a problem if that wasn't expectated and they were accessing the same shared state.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIContextProvider.cs
stephentoub: Should this be `ValueTask<AIContent?>`? Presumably it'll be common for an AIContextProvider to not augment a given request, in which case it can return null rather than a dummy instance.

dotnet/src/Microsoft.Extensions.AI.Agents/AggregateAIContextProvider.cs
stephentoub: This parallelism also means that AggregateAIContextProvider doesn't provide a pipeline, where the output from one provider feeds as part of the input into the next. Is that the intent?

dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatClientAgent.cs
```diff
@@ -349,6 +353,33 @@ public override AgentThread GetNewThread()
...
+            var aiContext = await thread.AIContextProvider.ModelInvokingAsync(inputMessages, thread.Id, cancellationToken).ConfigureAwait(false);
+            if (aiContext.Messages is { Count: > 0 })
+            {
+                threadMessages.AddRange(aiContext.Messages);
+            }
+
+            if (aiContext.AIFunctions is { Count: > 0 })
+            {
+                chatOptions ??= new();
+                chatOptions.Tools ??= [];
```
stephentoub, line 430: This should be cloning the options rather than mutating the one provided? Or is the one provided always a fresh copy?

The concern is otherwise that if the same options is passed in each time, this becomes both a leak and a functional bug, with the additional tools being added each time.

Same below for instructions, where the instructions would keep growing and growing.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIContext.cs
westey-m: Oops, yes. Fixed.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIContext.cs
westey-m: Thanks, fixed.

dotnet/src/Microsoft.Extensions.AI.Agents/AggregateAIContextProvider.cs
westey-m: Done, thanks.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIContext.cs
```diff
@@ -0,0 +1,43 @@
...
+{
+    /// <summary>
+    /// Gets or sets any instructions to pass to the AI model in addition to any other prompts
+    /// that it may already have (in the case of an agent), or chat history that may
+    /// already exist.
+    /// </summary>
+    /// <remarks>
+    /// These instructions will be transient and only apply to the current invocation.
+    /// </remarks>
+    public string? Instructions { get; set; }
```
dmytrostruk, line 26: I replaced this with collection of `AIContent` in Python, which works pretty well, so provider can work not only with strings but with other content types.

westey-m: This maps to the ChatOptions.Instructions property, which is a string.
I also have the separate messages property.  Does that match with what you are doing with AIContent?

stephentoub: Couldn't that be done with system messages in the messages list? Which services support arbitrary transient per-request multimodal instructions?

dmytrostruk: > I also have the separate messages property. Does that match with what you are doing with AIContent?

Yes, although I decided to use `AIContent` instead of `ChatMessage` in case if agent doesn't depend on `ChatClient`. So, it has nothing to do with chat messages.

dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatClientAgent.cs
westey-m: The CreateConfiguredChatOptions method call earlier in this method always clones the provided per request options if any were provided.  Agreed cloning makes sense.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIContext.cs
westey-m: > in case if agent doesn't depend on ChatClient

AIAgent currently depends on ChatMessage itself in its public interface, even when it doesn't use ChatClient.
I think the transport types should be considered broader than just for ChatClient.
ChatMessage is also important, because it indicates whether it's a system or user message.  We need this to ensure that some content can be safely added as user messages, without risk of prompt injection.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIContextProvider.cs
westey-m: Makes sense, fixed.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIContext.cs
dmytrostruk: > AIAgent currently depends on ChatMessage itself in its public interface, even when it doesn't use ChatClient.

This is actually what bothers me a little bit, but we can discuss it separately, since it's out of scope of this PR.

dotnet/src/Microsoft.Extensions.AI.Agents/AggregateAIContextProvider.cs
westey-m: The intent here was to keep latency low, since if you use a bunch of these ai context providers latency could suffer, but sure, I can remove it.  I suppose if anyone wants concurrency it's easy to build their own concurrent aggregate ai context provider.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIContextProvider.cs
westey-m: Converted to ValueTask for all the methods.

dotnet/src/Microsoft.Extensions.AI.Agents/AggregateAIContextProvider.cs
westey-m: I'm not sure that AggregateAIContextProvider should be the one introducing pipeline behavior.  It would mean changing the signature of ModelIInvoking to take AIContext as input.  Which means that each AIContextProvider needs to really be aggregating AIContext.  It should really just be for aggregating output from independent AIContextProviders.
Instead I think that if someone wants a pipeline, they can easily add sub AIContextProviders to their own AIContextProvider, which serves as input into theirs, and only those AIContextProviders then need to aggregate AIContext and they can do so in a way that makes sense for their scenario.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIContextProvider.cs
westey-m: I've removed Model from the name, so that it's just InvokingAsync.  I think it works even without it.

dotnet/src/Microsoft.Extensions.AI.Agents/AggregateAIContextProvider.cs
stephentoub: If you think it's important, I'd be fine with the parallelism being opt-in via a property or the like. That also means, though, it can be added later if there's demand.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIContextProvider.cs
westey-m: Yes, the expectation is that this instance is fresh.
This is the same approach as with IChatMessageStore instances, where the instance has to be created by the consumer via a factory, but the state is only deserialized into it when a thread is deserialized.
That said, we could probably revise that approach, and change the factory methods to also provide an optional JsonElement so that factories can use it to deserialize on construction.

dotnet/src/Microsoft.Extensions.AI.Agents/AggregateAIContextProvider.cs
westey-m: Agreed, I think we can easily add support later.  For now I'll leave as is.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIContextProvider.cs
westey-m: I've added a work item to reconsider this approach.  See #717

dotnet/samples/GettingStarted/Agents/Agent_Step12_Memory/Program.cs
```diff
@@ -0,0 +1,149 @@
...
+
+// It is also possible to add the memory component to an individual thread only instead of
+// via the factory above. This allows you to have different user info objects for different threads.
+// thread.AIContextProvider = new SampleApp.UserInfoMemory(chatClient.AsIChatClient(), userInfo);
+
+Console.WriteLine(">> Thread with blank memory\n");
+
+// Invoke the agent and output the text result.
+Console.WriteLine(await agent.RunAsync("Hello, what is the square root of 9?", thread));
+Console.WriteLine(await agent.RunAsync("My name is Ruaidhri", thread));
```
markwallace-microsoft: Missing fada "Ruaidhrí"

dotnet/samples/GettingStarted/Agents/Agent_Step12_Memory/Program.cs
```diff
@@ -0,0 +1,149 @@
...
+Console.WriteLine(await agent.RunAsync("I am 20 years old", thread));
+
+// We can serialize the thread. The serialized state will include the state of the memory component.
+var threadElement = await thread.SerializeAsync();
+
+Console.WriteLine("\n>> Deserialized Thread with previously created memories\n");
+
+// Later we can deserialize the thread and continue the conversation with the previous memory component state.
+var deserializedThread = await agent.DeserializeThreadAsync(threadElement);
+Console.WriteLine(await agent.RunAsync("Hello, what is the square root of 9?", deserializedThread));
```
markwallace-microsoft: Asking "What is my name and age?" would be a more compelling demo

dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatClientAgent.cs
```diff
@@ -349,6 +353,33 @@ public override AgentThread GetNewThread()
...
+                foreach (AITool tool in aiContext.Tools)
+                {
+                    chatOptions.Tools.Add(tool);
+                }
+            }
+
+            if (aiContext.Instructions is not null)
+            {
+                chatOptions ??= new();
+                chatOptions.Instructions = string.IsNullOrWhiteSpace(chatOptions.Instructions) ? aiContext.Instructions : $"{chatOptions.Instructions}\n{aiContext.Instructions}";
```
markwallace-microsoft, line 440: Should we allow developers to have more control over how the instructions are combined?

dotnet/src/Microsoft.Extensions.AI.Agents/AggregateAIContextProvider.cs
```diff
@@ -0,0 +1,189 @@
...
+using System.Text.Json;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.AI.Agents;
+
+/// <summary>
+/// An <see cref="AIContextProvider"/> that aggregates other <see cref="AIContextProvider"/> instances and delegates events to them.
+/// </summary>
+public sealed class AggregateAIContextProvider : AIContextProvider, IList<AIContextProvider>
```
markwallace-microsoft: Would `AIContextProviderAggregator` be a better name?

dotnet/src/Microsoft.Extensions.AI.Agents/AggregateAIContextProvider.cs
```diff
@@ -0,0 +1,189 @@
...
+        {
+            return new AIContext();
+        }
+
+        // Invoke all the sub providers in serial and
+        // aggregate all the sub-contexts into a single context.
+        List<AITool>? allTools = null;
+        List<ChatMessage>? allMessages = null;
+        StringBuilder? allInstructions = null;
+        foreach (var provider in this._providers)
```
markwallace-microsoft: Is the order of providers something we need to think about?

westey-m: It is really important in that for serialization or deserialization if the order is not maintained we end up with the wrong state in the wrong provider.
I'm not sure if there is a great alternative though, since nothing else is unique.  E.g. you could have two of the same types of providers with different data sources, so provider type is not unique.  There is no unique id per provider instance, and adding one would land on users, which isn't great.
On the one hand I'm questioning whether I should include the AggregateAIContextProvider at all.  If users wanted more than one, we could just recommend this pattern to them.

dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatClientAgent.cs
westey-m: I think we can easily add an option for that if it is requested.  I don't think we need it right now.

dotnet/samples/GettingStarted/Agents/Agent_Step12_Memory/Program.cs
westey-m: Sounds good, updated.

dotnet/samples/GettingStarted/Agents/Agent_Step12_Memory/Program.cs
westey-m: Thanks!

dotnet/src/Microsoft.Extensions.AI.Agents/AggregateAIContextProvider.cs
westey-m: Let's decide first if we want to keep it in for now.

dotnet/src/Microsoft.Extensions.AI.Agents/AggregateAIContextProvider.cs
markwallace-microsoft: I think the order of the context could be significant, the prompt changes if the order changes and the LLM may behave in a different way.

I tend to agree we should leave this out for now until we get more information about what developers want to do and how much control they need. We could include it as a sample.

dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatClientAgent.cs
markwallace-microsoft: Sounds good

dotnet/src/Microsoft.Extensions.AI.Agents/AggregateAIContextProvider.cs
westey-m: Thanks, removed.

dotnet/samples/GettingStarted/Agents/Agent_Step12_Memory/Program.cs
```diff
@@ -0,0 +1,152 @@
...
+// factory reads the user id from the current context and scopes the memory component
+// and its storage to that user id.
+AIAgent agent = chatClient.CreateAIAgent(new ChatClientAgentOptions()
+{
+    Instructions = "You are a friendly assistant. Always address the user by their name.",
+    AIContextProviderFactory = () => new SampleApp.UserInfoMemory(chatClient.AsIChatClient())
+});
+
+// Create a new thread for the conversation.
+AgentThread thread = agent.GetNewThread();
```
stephentoub: Did we ever consider calling GetNewThread just CreateThread (or CreateSession/CreateContext if you end up renaming AgentThread accordingly)? I have a vague memory that it started there but don't remember why we would have moved away from it.

westey-m: We were avoiding the term CreateThread, because if you are using e.g. a FoundryAgent it might imply that we are creating a thread on the server side, but we aren't doing that here.
If we move away from the thread/conversation nomenclature to something like Context/Session, I think we should be able to switch to Create, since it wouldn't be potentially misleading anymore.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIContextProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// <summary>
+    /// Called just before messages are added to the chat by any participant.
+    /// </summary>
+    /// <remarks>
+    /// Inheritors can use this method to update their context based on the new message.
+    /// </remarks>
+    /// <param name="newMessages">The new messages.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>A task that completes when the context has been updated.</returns>
+    public virtual ValueTask MessagesAddingAsync(IEnumerable<ChatMessage> newMessages, CancellationToken cancellationToken = default)
```
stephentoub: This is essentially the same method that's on AgentThread itself, right? A method for updating the thread or provider with the new messages from the conversation? Should they be named the same thing if they're serving the same purpose?

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIContextProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+
+    /// <summary>
+    /// Called just before the Model/Agent/etc. is invoked
+    /// Implementers can load any additional context required at this time,
+    /// and they should return any context that should be passed to the Model/Agent/etc.
+    /// </summary>
+    /// <param name="newMessages">The most recent messages that the Model/Agent/etc. is being invoked with.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>A task that completes when the context has been rendered and returned.</returns>
+    public abstract ValueTask<AIContext> InvokingAsync(IEnumerable<ChatMessage> newMessages, CancellationToken cancellationToken = default);
```
stephentoub: What about state that's not part of the ChatMessages, e.g. what if I pass context as part of ChatOptions.Instructions? Or are we purposefully avoiding that because we're trying to avoid transient information?

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIContextProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    /// <summary>
+    /// Deserializes the state contained in the provided <see cref="JsonElement"/> into the properties on this object.
+    /// </summary>
+    /// <param name="serializedState">A <see cref="JsonElement"/> representing the state of the object.</param>
+    /// <param name="jsonSerializerOptions">Optional settings for customizing the JSON deserialization process.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>A <see cref="ValueTask"/> that completes when the state has been deserialized.</returns>
+    public virtual ValueTask DeserializeAsync(JsonElement serializedState, JsonSerializerOptions? jsonSerializerOptions = null, CancellationToken cancellationToken = default)
```
stephentoub, line 62: You moved away from an instance deserialize method with AgentThread. Are you aiming to do the same here?

westey-m: Yes, that's the plan.  This PR is a bit overlapping with the other one though, so I figured I'll get this merged with deserialize as is, and then I'll refactor after as part of the other PR.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIContextProvider.cs
westey-m: ChatClientAgent merges the Instructions from AIContext with any instructions from ChatOptions.Instructions (both from the agent constructor and agent run).  These are all additive.

stephentoub: > ChatClientAgent merges the Instructions from AIContext with any instructions from ChatOptions.Instructions (both from the agent constructor and agent run). These are all additive.

Yes, but the AIContextProvider never sees those instructions and thus can't cull any information from them for later use, or have I misunderstood?

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIContextProvider.cs
westey-m: Yes, good point.  Do you have a preference for naming?  E.g. here it is using the adding/added invoking/invoked naming that is typically recommended for events, but it reads really badly IMO.
I have to admit I like OnNewMessages more.  Similarly OnInvoking just sounds better to me.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIContextProvider.cs
westey-m: Correct, culling is not supported here.  It's not supported for messages either.  All messages returned are additive.
Are you thinking we should revisit that?

stephentoub: Sorry, I didn't mean culling as in removing, I meant extracting information to be used later. That's part of the point of these APIs, to remember information that can later be used when producing the result of InvokeAsync, right? I'm wondering whether there's any other things these APIs would need to learn from, beyond just the messages, like ChatOptions.

westey-m: I'm not sure.  Instructions should typically contain instructions that the developer supplied via code, and can therefore supply again easily if they want to.  User supplied information is certainly more valuable since the user typically does not want to repeat themselves.  Replies from the LLM are also valuable, since it allows us to store that and so provide more consistent replies.
There may of course be use cases I haven't considered though.

The interesting question is probably, how to make this extensible if we wanted to add more.  MessagesAddingAsync is very much about new messages being added, if they need to be added.  This is deliberately separate from InvokingAsync since messages shouldn't be added to the chat history or memory if the request failed.  See pseudocode below.
While it is certainly possible to add new methods, it would be very strange to have a method specifically for instructions.  MessagesAdding means that these message are currently being added to the chat history.  I'm not sure what a method for instructions would mean.

We could potentially introduce an AgentRunContext parameter, which is separate from the messages and contains additional information about the current run and/or agent.  To keep it agent agnostic though, it might not contain very much.  We could have AgentRunOptions though which could be cast to ChatClientAgentRunOptions to get ChatOptions.  That said, I was hoping to keep AIContextProviders fairly agnostic to their usage, so that they could be used with more than just agents.  Right now, they only depend on ChatMessage and AITool.

I'm also thinking we could change MessagesAddingAsync to something like `InvokedAsync(succeeded, inputMessages, resultMessages, agentRunContext)` instead.  This means that the AIContextProvider can decide itself what to do on success or failure, and that this method matches a run 1 to 1, similar to InvokingAsync.  Therefore passing in more context also feels more natural.

Here is some pseudocode of what the flow is right now:
```
RunAsync(inputMessages, inputOptions)
{
    var additionalContext = InvokingAsync(inputMessages);
    var result = InvokeLLM(
        inputMessagees + additionalContext.Messages,
        this.instructions + inputOptions.instructions + additionalContext.instructions);
    if (result.succeeded)
    {
        MessagesAddingAsync(inputMessages);
        MessagesAddingAsync(result.Messages);
    }
}
```

westey-m: We could do something like this (method naming aside):

```csharp
public abstract class AIContextProvider
{
    public abstract ValueTask<AIContext> InvokingAsync(InvokingContext invokingContext, CancellationToken cancellationToken = default);
    public abstract ValueTask InvokedAsync(InvokedContext invokedContext, CancellationToken cancellationToken = default);

    public class InvokingContext
    {
        public IEnumerable<ChatMessage> InputMessages { get; set; }
        // This could be ChatOptions or ChatAgentRunOptions or something else, depending
        // on where the AIContextProvider is used. We can also add it later.
        public object RawInvokeOptions { get; set; }
    }

    public class InvokedContext
    {
        public bool InvokeSucceeded { get; set; }
        public IEnumerable<ChatMessage> InputMessages { get; set; }
        public IEnumerable<ChatMessage> ResultMessages { get; set; }
        public object RawInvokeOptions { get; set; }
    }
}
```

westey-m: Updated to this approach.

dotnet/samples/GettingStarted/Agents/Agent_Step13_Memory/Agent_Step13_Memory.csproj
```diff
@@ -0,0 +1,23 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net9.0</TargetFramework>
+    <LangVersion>12</LangVersion>
```
stephentoub:
```suggestion
```
https://github.com/microsoft/agent-framework/pull/760

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIContextProvider.cs
```diff
@@ -0,0 +1,98 @@
...
+{
+    /// <summary>
+    /// Called just before the Model/Agent/etc. is invoked
+    /// Implementers can load any additional context required at this time,
+    /// and they should return any context that should be passed to the Model/Agent/etc.
+    /// </summary>
+    /// <param name="context">Contains the event context.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>A task that completes when the context has been rendered and returned.</returns>
+    public abstract ValueTask<AIContext> InvokingAsync(InvokingContext context, CancellationToken cancellationToken = default);
```
stephentoub, line 29: abstract rather than virtual? Is the thinking that an AIContextProvider is only useful if it provides an AIContext?

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIContextProvider.cs
```diff
@@ -0,0 +1,98 @@
...
+
+    /// <summary>
+    /// Contains the event context provided to <see cref="AIContextProvider.InvokingAsync(InvokingContext, CancellationToken)"/>.
+    /// </summary>
+    public class InvokingContext
+    {
+        /// <summary>
+        /// Gets the messages that will be sent to the Model/Agent/etc. for this invocation.
+        /// </summary>
+        public required IEnumerable<ChatMessage> RequestMessages { get; init; }
```
stephentoub: I would prefer if we just had a ctor that took the request messages, instead of this being required/init. Same for InvokedContext below.

dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatClientAgent.cs
```diff
@@ -114,27 +114,35 @@ public override async Task<AgentRunResponse> RunAsync(
 
         this._logger.LogAgentChatClientInvokingAgent(nameof(RunAsync), this.Id, agentName, this._chatClientType);
 
-        ChatResponse chatResponse = await this.ChatClient.GetResponseAsync(threadMessages, chatOptions, cancellationToken).ConfigureAwait(false);
+        // Call the chatclient and notify the AIContextProvider of any failures.
```
stephentoub:
```suggestion
        // Call the IChatClient and notify the AIContextProvider of any failures.
```

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIContextProvider.cs
westey-m: @stephentoub, what's your thoughts on the other properties?  We don't want to make them settable, so they either need to be init or private and then all on the constructor.  If we go for private set and all on the constructor, adding more properties means expanding the constructor (or adding an overload), which of course risks having the constructor parameter order be simply guided by when the properties are added, rather than a more thought out order.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIContextProvider.cs
westey-m: Yes, If it doesn't provide an AIContext, then it has no effect on the agent at all, so pretty much a no-op.  Of course, nothing stops someone from using an AIContextProvider to just listen, but that really isn't the goal of them.

dotnet/samples/GettingStarted/Agents/Agent_Step13_Memory/Agent_Step13_Memory.csproj
westey-m: Updated to match the rest, but we were actually keeping all required settings to make a sample work in each sample project file so that when copying a sample project developers wouldn't be faced with missing settings.

dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatClientAgent.cs
westey-m: Updated.

dotnet/samples/GettingStarted/Agents/Agent_Step13_Memory/Program.cs
```diff
@@ -0,0 +1,151 @@
...
+using Azure.Identity;
+using Microsoft.Extensions.AI;
+using Microsoft.Extensions.AI.Agents;
+using OpenAI;
+using SampleApp;
+
+var endpoint = Environment.GetEnvironmentVariable("AZURE_OPENAI_ENDPOINT") ?? throw new InvalidOperationException("AZURE_OPENAI_ENDPOINT is not set.");
+var deploymentName = Environment.GetEnvironmentVariable("AZURE_OPENAI_DEPLOYMENT_NAME") ?? "gpt-4o-mini";
+
+OpenAI.Chat.ChatClient chatClient = new AzureOpenAIClient(
```
markwallace-microsoft: Emm, code smell, why is fully qualified class name required here?

dotnet/samples/GettingStarted/Agents/Agent_Step13_Memory/Program.cs
```diff
@@ -0,0 +1,151 @@
...
+        }
+
+        public override ValueTask<AIContext> InvokingAsync(InvokingContext context, CancellationToken cancellationToken = default)
+        {
+            StringBuilder instructions = new();
+
+            // Add the user name and age to the instructions if we have them, otherwise ask for them.
+            instructions.AppendLine(
+                this.UserInfo.UserName == null ?
+                "Ask the user for their name and politely decline to answer any questions until they provide it." :
```
markwallace-microsoft, line 121: This seems slightly confusing, should it be "If we don't already know the user's name, ..."

dotnet/samples/GettingStarted/Agents/Agent_Step13_Memory/Program.cs
westey-m: It's not required, removed.

dotnet/samples/GettingStarted/Agents/Agent_Step13_Memory/Program.cs
westey-m: Sure, updated the comment.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AIContextProvider.cs
westey-m: I'll get this merged with just RequestMessages on the constructor for now, since this is PR blocking me making progress in other areas, but happy to update the other properties as well if needed in a follow up PR.

dotnet/samples/GettingStarted/Agents/Agent_Step13_Memory/Agent_Step13_Memory.csproj
stephentoub: Lang Version 13 is implied by a net9.0 tfm.

stephentoub: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-versioning#defaults

PR #692, opened 2025-09-10
.NET: [WIP] Agent Discovery Prototype

DeagleGross:
An initial implementation of #454 to show the prototype.

### DIFF
--- a/dotnet/src/Microsoft.Extensions.AI.Agents.Hosting.Discovery/WebApplicationExtensions.cs
+++ b/dotnet/src/Microsoft.Extensions.AI.Agents.Hosting.Discovery/WebApplicationExtensions.cs
@@ -26,7 +26,7 @@ public static class WebApplicationExtensions
     /// <summary>
     /// Enables the agent discovery endpoint at the specified path.
     /// </summary>
-    public static void EnableDiscovery(this IEndpointRouteBuilder endpoints, [StringSyntax("Route")] string? agentsPath = default, [StringSyntax("Route")] string? actorsPath = default)
+    public static void MapDiscovery(this IEndpointRouteBuilder endpoints, [StringSyntax("Route")] string? agentsPath = default, [StringSyntax("Route")] string? actorsPath = default)
     {
         agentsPath ??= AgentsBasePath;
         var agentRouteGroup = endpoints.MapGroup(agentsPath);
@@ -66,9 +66,9 @@ public static class WebApplicationExtensions
     {
         var httpActorProcessor = endpoints.ServiceProvider.GetRequiredService<HttpActorProcessor>();
 
-        // GET /actors/v1/{actorType}/{actorKey}/{messageId}
+        // GET /actors/v1/{actorType}/{actorKey}/messages/{messageId}
         endpoints.MapGet(
-            "/{actorType}/{actorKey}/{messageId}", (
+            "/{actorType}/{actorKey}/messages/{messageId}", (
             string actorType,
             string actorKey,
             string messageId,
@@ -80,7 +80,7 @@ public static class WebApplicationExtensions
             )
             .WithName("GetActorResponse");
 
-        // POST /actors/v1/{actorType}/{actorKey}/{messageId}
+        // POST /actors/v1/{actorType}/{actorKey}/messages/{messageId}
         endpoints.MapPost(
             "/{actorType}/{actorKey}/{messageId}", (
             string actorType,
@@ -95,7 +95,7 @@ public static class WebApplicationExtensions
             )
             .WithName("SendActorRequest");
 
-        // POST /actors/v1/{actorType}/{actorKey}/{messageId}:cancel
+        // POST /actors/v1/{actorType}/{actorKey}/messages/{messageId}:cancel
         endpoints.MapPost(
             "/{actorType}/{actorKey}/{messageId}:cancel", (
             string actorType,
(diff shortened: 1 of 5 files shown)


dotnet/src/Microsoft.Extensions.AI.Agents.Hosting.Discovery/Model/AgentMetadata.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.Collections.Generic;
+using System.Text.Json.Serialization;
+using Microsoft.Extensions.AI.Agents.Runtime;
+
+namespace Microsoft.Extensions.AI.Agents.Hosting.Discovery.Model;
+
+/// <summary>
+/// Representation of metadata for the <see cref="AIAgent"/>
+/// </summary>
+public sealed class AgentMetadata
```
luisquintanilla, line 12: How do developers know which tools / skills / capabilities are available in an agent?

dotnet/src/Microsoft.Extensions.AI.Agents.Hosting.Discovery/WebApplicationExtensions.cs
```diff
@@ -0,0 +1,109 @@
...
+                {
+                    return Results.NotFound();
+                }
+
+                return Results.Ok(agentMetadata);
+            })
+            .WithName("GetAgent");
+    }
+
+    private static void MapHttpActorProcessing(this IEndpointRouteBuilder endpoints)
```
luisquintanilla, line 65: Should Actors be exposed? My understanding is what end-users and developers interact with are the agents, not the actors.

dotnet/src/Microsoft.Extensions.AI.Agents.Hosting/HostApplicationBuilderAgentExtensions.cs
```diff
@@ -40,11 +36,13 @@ public static IHostApplicationBuilder AddAIAgent(this IHostApplicationBuilder bu
     /// <param name="chatClient">The chat client which the agent will use for inference.</param>
     /// <returns>The configured host application builder.</returns>
     /// <exception cref="ArgumentNullException">Thrown when <paramref name="builder"/>, <paramref name="name"/>, or <paramref name="instructions"/> is null.</exception>
-    public static IHostApplicationBuilder AddAIAgent(this IHostApplicationBuilder builder, string name, string instructions, IChatClient chatClient)
+    public static IAgentHostingBuilder AddAIAgent(this IHostApplicationBuilder builder, string name, string instructions, IChatClient chatClient)
```
luisquintanilla, line 39: Consider adding generic `AddAIAgent<T>` overloads backed by [DelegatingAIAgent](https://github.com/microsoft/agent-framework/blob/daee44d5835b8b01a0c7180d5f792ba56d56fafa/dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/DelegatingAIAgent.cs#L19).

For example, given 

```csharp
public class EditorAgent(AIAgent innie) : DelegatingAIAgent(innie);
```

A user can register it as:

```csharp
builder.AddAIAgent<EditorAgent>(
        name: $"_{nameof(EditorAgent)}", 
        description: "You help with editing tasks",
        instructions: "Edit tasks"
        );
```

dotnet/src/Microsoft.Extensions.AI.Agents.Hosting.Discovery/Model/AgentMetadata.cs
DeagleGross: We do not have a nice way to map tools on AIAgent definition yet; users can use `metadata` field for now. Later when we support it we can add a dedicated field for that (no breaking change expected)

dotnet/src/Microsoft.Extensions.AI.Agents.Hosting/HostApplicationBuilderAgentExtensions.cs
DeagleGross: I dont see how I can construct an instance of this `AIAgent`, since `AIAgent` does not provide any required instantiation contract (like constructor e.g.).

```
public static IAgentHostingBuilder AddDelegatingAIAgent<TAgent>(this IHostApplicationBuilder builder, string name)
    where TAgent : DelegatingAIAgent
    => builder.AddAIAgent(name, (sp, name) => new DelegatingAIAgent(???))
```

Today's workaround is to use this API:
```
AddAIAgent(this IHostApplicationBuilder builder, string name, Func<IServiceProvider, string, AIAgent> createAgentDelegate);

// usage
builder.AddAIAgent("pirate", (sp, name) => new DelegatingAgent(sp.GetRequiredService<AIAgent>(name)));
```
does it suffice? Let me know because it feels like I am missing something

dotnet/src/Microsoft.Extensions.AI.Agents.Hosting.Discovery/WebApplicationExtensions.cs
DeagleGross: I think it is ok, anyway the API does not show "actorId" or "actorType" anywhere except the swagger UI for example. From the users perspective its relatively simple - you need a) agent name; b) conversation id; c) message id to interact via this API.

Also, `ActorType` and `ActorId` are public types, so if user is referncing `Microsoft.Extensions.AI.Agents.Runtime.Abstractions` (which they do in case of discovery), they are anyway using the actor-model.

However, I like your proposition to "hide" this from user as much as possible. If you have an idea on how to name the endpoint better - let me know!

DeagleGross:
See https://github.com/microsoft/agent-framework/pull/454#issuecomment-3355395484

Issue #717, opened 2025-09-11
.NET: Reconsider AgentThread deserialization: Can it be done on construction?


westey-m:
We currently have a separate Deserialize method for AgentThread, which allows state to be deserialized into it after the thread id constructed.
Previously there were blockers to deserializing on construction, but I believe these may have been removed due to other changes.   We should try to prototype deserialization on construction again now.  For MemoryStore and AIContextProviders, we can pass optional JsonElements to their factories, allowing the factories to create these objects with the correct state in place.

PR #720, opened 2025-09-11
.NET: POC: Moving agent thread specifics to sub classes and moving deserialization to constructors.

westey-m:
### Motivation and Context

Currently AgentThread has a ConversationId which can be confusing and isn't consistently used.
We also have thread deserialization via a specific method on thread, which means you have to create the object first, and then deserialize.  This means you can also deserialize into an existing object, overwriting state, which can be confusing and error prone.

This PR is a POC to prototype two alternatives to these problems.  One is having subclasses for threads for different agent types and the other is to deserialize via constructors.

> IMPORTANT: This PR has not yet updated any tests, just samples.  Can update tests once I've gathered feedback and if we want to go ahead with these changes.

#717
#304

### Contribution Checklist


dotnet/src/Microsoft.Extensions.AI.Agents.A2A/A2AAgent.cs
```diff
@@ -51,15 +52,29 @@ public A2AAgent(A2AClient a2aClient, string? id = null, string? name = null, str
...
 
+        thread ??= this.GetNewThread();
+        if (thread is not ClientProxyAgentThread typedThread)
+        {
+            throw new InvalidOperationException("The provided thread is not compatible with the agent. Only threads created by the agent can be used.");
+        }
+
         // Linking the message to the existing conversation, if any.
-        a2aMessage.ContextId = thread?.ConversationId;
+        a2aMessage.ContextId = typedThread?.ServiceThreadid;
```
dmytrostruk: Small typo:
```suggestion
        a2aMessage.ContextId = typedThread?.ServiceThreadId;
```

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/ClientProxyAgentThread.cs
```diff
@@ -0,0 +1,72 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Extensions.AI.Agents;
+
+/// <summary>
+/// A thread type that may be used by any agent that proxies a remote agent where the remote agent has its own thread management.
+/// </summary>
+public class ClientProxyAgentThread : AgentThread
```
dmytrostruk: Should it be just `ProxyAgentThread` or `DelegatedAgentThread`? Not sure if `Client` part is useful here, unless I'm missing something.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/ClientProxyAgentThread.cs
```diff
@@ -0,0 +1,72 @@
...
+    /// Initializes a new instance of the <see cref="ClientProxyAgentThread"/> class from serialized state.
+    /// </summary>
+    /// <param name="serializedThreadState">A <see cref="JsonElement"/> representing the serialized state of the thread.</param>
+    public ClientProxyAgentThread(JsonElement serializedThreadState)
+    {
+        var state = JsonSerializer.Deserialize(
+            serializedThreadState,
+            AgentAbstractionsJsonUtilities.DefaultOptions.GetTypeInfo(typeof(ThreadState))) as ThreadState;
+
+        if (state?.ServiceThreadid is string threadId && !string.IsNullOrWhiteSpace(threadId))
```
dmytrostruk: If deserialization above is completed, should we still check here if `ServiceThreadId` is `string`?

dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatClientAgentThread.cs
```diff
@@ -0,0 +1,207 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Extensions.AI.Agents;
+
+/// <summary>
+/// Thread for chat client based agents.
+/// </summary>
+public class ChatClientAgentThread : ClientProxyAgentThread
```
dmytrostruk: Do we expect more classes to inherit from this or we can make it `sealed`? It may become complex, if more inherited classes will be added to the thread hierarchy.

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/A2AAgent.cs
westey-m: Oops, didn't see that at all.  Thanks, fixed.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/ClientProxyAgentThread.cs
westey-m: I think ProxyAgentThread also works.  Renamed.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/ClientProxyAgentThread.cs
westey-m: The check here for string is really only a null check on both state and ServiceThreadId.  I've removed the isnullorwhitespace check though.  It seems wrong not to deserialize exactly what was serialized.

dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatClientAgentThread.cs
westey-m: Sealed make sense.  Done.

westey-m:
Closing POC PR, since all changes are in separate PRs.

PR #727, opened 2025-09-12
Python: update to packaging section of python package design doc

eavanvalkenburg:
### Motivation and Context

Now that we are getting closer to public preview we need to finalize and make consistent the scope of subpackages, this adds a section on that topic with some considerings and a `TBD` decision.

### Contribution Checklist

### DIFF
--- a/docs/design/python-package-setup.md
+++ b/docs/design/python-package-setup.md
@@ -57,25 +57,18 @@ Overall the following structure is proposed:
         * context_providers
         * logging
         * workflows (includes multi-agent orchestration)
-        * guardrails / filters
+        * middleware
+        * telemetry (user_agent)
     * advanced components, will be exposed from `agent_framework.<component>`:
-        * vector_data (tbd, vector stores and other MEVD pieces)
+        * vector_data (tbd, vector stores and other MEVD-like pieces)
         * text_search (tbd)
         * exceptions
         * evaluations (tbd)
         * utils (optional)
-        * telemetry (could also be observability or monitoring)
-    * connectors; subpackages*
-        * Subpackages are any additional functionality that is useful for a user, to reduce friction they will imported in a similar way as advanced components, however the code for them will be in a separate package, so that they can be installed separately, they must expose all public items, in their main `__init__.py` file, so that they can be imported from the main package without additional import levels.
-        In the main package a corresponding folder will be created, with a `__init__.py` file that lazy imports the public items from the subpackage, so that they can be exposed from the main package.
-        * Some examples are:
-            * azure (non LLM integrations)
-            * will be exposed through i.e. `agent_framework.azure`
-            * anything other then a connector that we want to expose as a separate package, for instance:
-                * mem0 (memory management)
-                * would be exposed through i.e. `agent_framework.mem0`
-        * A package name cannot overlap with each other or with components in the main package, so `guardrails` cannot be used as a package name, since it is already used in the main package.
-            * A package can implement additional guardrails functionality, but that would become something like `from agent_framework.azure import ContentSafetyGuardrail`, which would contain a guardrail implementation using Azure Content Safety, but it is not the same as the `agent_framework.guardrails` component.
+        * observability
+    * vendor folders with connectors and integrations, will be exposed from `agent_framework.<vendor>`:
+        * Code can be both in folder or in subpackage with lazy import.
+        * See subpackage scope below for more detail
 * tests
 * samples
 * extensions
@@ -86,12 +79,20 @@ All the init's in the subpackages will use lazy loading so avoid importing the e
 Internal imports will be done using relative imports, so that the package can be used as a namespace package.
 
 ### File structure
-The resulting file structure will be as follows:
+The resulting file structure will be as follows (not all things currently implemented, just an example):
 
 ```plaintext
 packages/
     main/
         agent_framework/
+            azure/
+                __init__.py
+                _chat_client.py
+                ...
+            microsoft/
+                __init__.py
+                _copilot_studio.py
+                ...
             openai/
                 __init__.py
                 _chat_client.py
@@ -103,17 +104,16 @@ packages/
             _tools.py
             _models.py
             _logging.py
-            context_providers.py
-            guardrails.py
+            _middleware.py
+            _telemetry.py
+            observability.py
             exceptions.py
-            evaluations.py
             utils.py
-            telemetry.py
-            templates.py
-            text_search.py
-            vector_data.py
-            workflows.py
             py.typed
+        _workflow/
+            __init__.py
+            _workflow.py
+            ...etc...
         tests/
             unit/
                 test_types.py
@@ -122,28 +122,27 @@ packages/
         pyproject.toml
         README.md
         ...
-    mem0/
-        agent_framework/
-            mem0/
-                __init__.py
-                _mem0.py
-                ...
+    azure-ai-agents/
+        agent_framework-azure-ai-agents/
+            __init__.py
+            _chat_client.py
+            ...
+        tests/
+            test_azure_ai_agents.py
+        samples/ (optional)
+            ...
+        pyproject.toml
+        README.md
+        ...
     redis/
         ...
-    google/
-        agent_framework/
-            google/
-                __init__.py
-                _chat.py
-                _bigquery.py
-                ...
+    mem0/
+        agent_framework-mem0/
+            __init__.py
+            _provider.py
+            ...
         tests/
-            unit/
-                test_google_client.py
-                test_google_tools.py
-                ...
-            integration/
-                test_google_integration.py
+            test_mem0_provider.py
         samples/ (optional)
             ...
         pyproject.toml


docs/design/python-package-setup.md
```diff
@@ -53,19 +53,19 @@ Overall the following structure is proposed:
...
         * (single) agents (includes threads)
         * tools (includes MCP and OpenAPI)
-        * models/types (name tbd, will include the equivalent of MEAI for dotnet; content types and client abstractions)
+        * types
+        * context_providers
         * logging
+        * workflows (includes multi-agent orchestration)
     * advanced components, will be exposed from `agent_framework.<component>`:
-        * context_providers (tbd)
         * guardrails / filters
```
markwallace-microsoft: Should this be core functionality?

docs/design/python-package-setup.md
```diff
@@ -53,19 +53,19 @@ Overall the following structure is proposed:
...
         * vector_data (vector stores and other MEVD pieces)
         * text_search
         * exceptions
```
markwallace-microsoft, line 65: Should this be core functionality?

docs/design/python-package-setup.md
```diff
@@ -53,19 +53,19 @@ Overall the following structure is proposed:
...
         * evaluations
```
markwallace-microsoft: Are we doing anything with evaluations?

docs/design/python-package-setup.md
```diff
@@ -53,19 +53,19 @@ Overall the following structure is proposed:
...
         * tools (includes MCP and OpenAPI)
-        * models/types (name tbd, will include the equivalent of MEAI for dotnet; content types and client abstractions)
+        * types
+        * context_providers
         * logging
+        * workflows (includes multi-agent orchestration)
     * advanced components, will be exposed from `agent_framework.<component>`:
-        * context_providers (tbd)
         * guardrails / filters
         * vector_data (vector stores and other MEVD pieces)
```
markwallace-microsoft: Can we leave these in SK and reuse from there? That is what we plan to do on the .Net side.

docs/design/python-package-setup.md
```diff
@@ -53,19 +53,19 @@ Overall the following structure is proposed:
...
         * text_search
```
markwallace-microsoft: Can we leave these in SK and reuse from there? That is what we plan to do on the .Net side.

docs/design/python-package-setup.md
```diff
@@ -163,79 +163,68 @@ We might add a template subpackage as well, to make it easy to setup, this could
...
+3. Group connectors by vendor and maturity, so that you can graduate something from the i.e. the `google-preview` package to the `google` package when it becomes GA.
+    * Pros:
+        - fewer packages to manage, publish and maintain
+        - easier for users to find and install the right package.
+        - users that work primarily with one platform have a single package to install.
+        - clear what the status is based on extra name
+    * Cons:
+        - moving something from one to the other might be a breaking change
+        - still larger packages with more dependencies
+    It could be mitigated that the `google-preview` package is still imported from `agent_framework.google`, so that the import path does not change, when something graduates, but it is still a clear choice for users to make. And we could then have three extras on that package, `google`, `google-preview` and `google-all` to make it easy to install the right package or just all.
```
markwallace-microsoft: I think this will be confusing as things will arbitrarily move out of preview.

docs/design/python-package-setup.md
```diff
@@ -163,79 +163,68 @@ We might add a template subpackage as well, to make it easy to setup, this could
...
+1. One subpackage per vendor, so a `google` package that contains all Google related connectors, such as `GoogleChatClient`, `BigQueryCollection`, etc.
+    * Pros:
+        - fewer packages to manage, publish and maintain
+        - easier for users to find and install the right package.
+        - users that work primarily with one platform have a single package to install.
+    * Cons:
+        - larger packages with more dependencies
+        - larger installation sizes
+        - more difficult to version, since some parts may be GA, while other are in preview.
+2. One subpackage per connector, so a `google_chat` package, a `bigquery` package, etc.
```
markwallace-microsoft: Should it be `google_bigquery`, we may want to account for different vendors having their own connectors for the same service e.g. Google and Azure both having Redis.

docs/design/python-package-setup.md
```diff
@@ -163,79 +163,68 @@ We might add a template subpackage as well, to make it easy to setup, this could
 
 In the [`DEV_SETUP.md`](../../python/DEV_SETUP.md) we will add instructions for how to deal with the path depth issues, especially on Windows, where the maximum path length can be a problem.
 
+### Subpackage scope
+The scope of subpackages is a balance between the number of subpackages we end up with, the size and complexity of each subpackage and its dependencies, and the user experience of installing and using the package.
+
+The two extremes are:
+1. One subpackage per vendor, so a `google` package that contains all Google related connectors, such as `GoogleChatClient`, `BigQueryCollection`, etc.
```
markwallace-microsoft: We should use Azure as the reference sample.

We should talk about plan's for the vector data stuff.

docs/design/python-package-setup.md
```diff
@@ -163,79 +163,68 @@ We might add a template subpackage as well, to make it easy to setup, this could
...
+    It could be mitigated that the `google-preview` package is still imported from `agent_framework.google`, so that the import path does not change, when something graduates, but it is still a clear choice for users to make. And we could then have three extras on that package, `google`, `google-preview` and `google-all` to make it easy to install the right package or just all.
+4. Group connectors by vendor and type, so that you have a `google-chat` package, a `google-data` package, etc.
+    * Pros:
+        - smaller packages with fewer dependencies
+        - smaller installation sizes
+    * Cons:
+        - more packages to manage, register, publish and maintain
+        - more extras, means more difficult for users to find and install the right package.
+        - still keeps the lifecycle more difficult, since some parts may be GA, while other are in preview.
+
```
markwallace-microsoft: Can we combine granular packages with meta packages that provide typical combinations e.g. all of Azure or just Azure.AI or Azure.Data

docs/design/python-package-setup.md
eavanvalkenburg: I agree, moved it

docs/design/python-package-setup.md
eavanvalkenburg: it is common in python packages to make exceptions available from something like a `expections` namespace, so this is the right spot

docs/design/python-package-setup.md
eavanvalkenburg: not yet, so currently this is not in the package, but I think we might have something like this, added `(tbd)`

docs/design/python-package-setup.md
eavanvalkenburg: added TBD, depending on decisions we make later on

docs/design/python-package-setup.md
eavanvalkenburg: added tbd

docs/design/python-package-setup.md
eavanvalkenburg: The way I think about is, that in this scenario: during early dev, you just add what you might need through `<platform>-all` while when you move towards prod, you only import and use `<platform>` dependencies, so that you are using released things only.

docs/design/python-package-setup.md
eavanvalkenburg: yeah, these are examples, but yes makes sense to make them specific.

docs/design/python-package-setup.md
eavanvalkenburg: I didn't only want to use Azure and as the section below highlights our own platform is somewhat complex in this regard.

docs/design/python-package-setup.md
eavanvalkenburg: so we can do extras that install multiple subpackages, i.e. `google` extras in main, install `google_gemini`, `google_bigquery`, etc. but that might make the dev experience easier, but the install is as large or larger as when it is a single package anyway but we keep the overhead. So it's a bit making our lives worse, while not gaining much.

eavanvalkenburg: added this as an option

docs/design/python-package-setup.md
```diff
@@ -53,19 +53,19 @@ Overall the following structure is proposed:
...
-        * vector_data (vector stores and other MEVD pieces)
-        * text_search
+    * advanced components, will be exposed from `agent_framework.<component>`:
+        * vector_data (tbd, vector stores and other MEVD pieces)
+        * text_search (tbd)
         * exceptions
-        * evaluations
+        * evaluations (tbd)
         * utils (optional)
         * telemetry (could also be observability or monitoring)
```
TaoChenOSU: `observability` is a more common name

docs/design/python-package-setup.md
```diff
@@ -163,79 +163,83 @@ We might add a template subpackage as well, to make it easy to setup, this could
...
+1. One subpackage per vendor, so a `google` package that contains all Google related connectors, such as `GoogleChatClient`, `BigQueryCollection`, etc.
+    * Pros:
+        - fewer packages to manage, publish and maintain
+        - easier for users to find and install the right package.
+        - users that work primarily with one platform have a single package to install.
+    * Cons:
+        - larger packages with more dependencies
+        - larger installation sizes
+        - more difficult to version, since some parts may be GA, while other are in preview.
+2. One subpackage per connector, so a i.e. `google_chat` package, a i.e. `qoogle_bigquery` package, etc.
```
TaoChenOSU:
```suggestion
2. One subpackage per connector, so a i.e. `google_chat` package, a i.e. `google_bigquery` package, etc.
```

docs/design/python-package-setup.md
```diff
@@ -163,79 +163,83 @@ We might add a template subpackage as well, to make it easy to setup, this could
...
+        - easier for users to find and install the right package.
+    * Cons:
+        - larger overhead in maintaining the `__init__.py` files that do the lazy loading and error handling.
+        - larger overhead in package management, since we have to ensure that the main package.
+
+Decision:
+TBD
+
+### Microsoft vs Azure packages
+Another consideration is for Microsoft, since we have a lot of Azure services, but also other Microsoft services, such as Microsoft Copilot Studio, and potentially other services in the future, and maybe Foundry also will be marketed separate from Azure at some point. We could also have both a `microsoft` and an `azure` package, where the `microsoft` package contains all Microsoft services, excluding Azure, while the `azure` package only contains Azure services.
```
TaoChenOSU: If we follow the second option above, would this still be an issue?

docs/design/python-package-setup.md
```diff
@@ -163,79 +163,83 @@ We might add a template subpackage as well, to make it easy to setup, this could
...
+    It could be mitigated that the `google-preview` package is still imported from `agent_framework.google`, so that the import path does not change, when something graduates, but it is still a clear choice for users to make. And we could then have three extras on that package, `google`, `google-preview` and `google-all` to make it easy to install the right package or just all.
+4. Group connectors by vendor and type, so that you have a `google-chat` package, a `google-data` package, etc.
+    * Pros:
+        - smaller packages with fewer dependencies
+        - smaller installation sizes
+    * Cons:
+        - more packages to manage, register, publish and maintain
+        - more extras, means more difficult for users to find and install the right package.
+        - still keeps the lifecycle more difficult, since some parts may be GA, while other are in preview.
+5. Add `meta`-extras, that combine different subpackages as one extra, so we could have a `google` extra that includes `google-chat`, `google-bigquery`, etc.
```
dmytrostruk: I like the combination of Option 4 and 5:
1. For developers that want an entire vendor package, they can install metapackage by vendor.
2. For developers who are interested in specific functionality - they can install specific package like `google-chat`.

docs/design/python-package-setup.md
eavanvalkenburg: not too much, I've added a clarification

docs/design/python-package-setup.md
```diff
@@ -163,79 +163,83 @@ We might add a template subpackage as well, to make it easy to setup, this could
...
+    * Pros:
+        - fewer packages to manage, publish and maintain
+        - easier for users to find and install the right package.
+        - users that work primarily with one platform have a single package to install.
+        - clear what the status is based on extra name
+    * Cons:
+        - moving something from one to the other might be a breaking change
+        - still larger packages with more dependencies
+    It could be mitigated that the `google-preview` package is still imported from `agent_framework.google`, so that the import path does not change, when something graduates, but it is still a clear choice for users to make. And we could then have three extras on that package, `google`, `google-preview` and `google-all` to make it easy to install the right package or just all.
+4. Group connectors by vendor and type, so that you have a `google-chat` package, a `google-data` package, etc.
```
moonbox3: I think operating at this granularity would introduce too much of a maintenance headache. Think of all of the sub-packages we'd have to manage, deal with versioning, publishing, release notes, etc.

Also, from a dev's perspective, if they want to use components from various sub-packages, we've just introduced more complexity around knowing which ones to install, what the imports are, etc.

docs/design/python-package-setup.md
```diff
@@ -163,79 +163,83 @@ We might add a template subpackage as well, to make it easy to setup, this could
 
 In the [`DEV_SETUP.md`](../../python/DEV_SETUP.md) we will add instructions for how to deal with the path depth issues, especially on Windows, where the maximum path length can be a problem.
 
+### Subpackage scope
+The scope of subpackages is a balance between the number of subpackages we end up with, the size and complexity of each subpackage and its dependencies, and the user experience of installing and using the package.
+
+The two extremes are:
+1. One subpackage per vendor, so a `google` package that contains all Google related connectors, such as `GoogleChatClient`, `BigQueryCollection`, etc.
```
moonbox3: I would prefer a sub-package per vendor.

docs/design/python-package-setup.md
```diff
@@ -163,79 +163,83 @@ We might add a template subpackage as well, to make it easy to setup, this could
...
+        - easier for users to find and install the right package.
+    * Cons:
+        - larger overhead in maintaining the `__init__.py` files that do the lazy loading and error handling.
+        - larger overhead in package management, since we have to ensure that the main package.
+
+Decision:
+TBD
+
+### Microsoft vs Azure packages
+Another consideration is for Microsoft, since we have a lot of Azure services, but also other Microsoft services, such as Microsoft Copilot Studio, and potentially other services in the future, and maybe Foundry also will be marketed separate from Azure at some point. We could also have both a `microsoft` and an `azure` package, where the `microsoft` package contains all Microsoft services, excluding Azure, while the `azure` package only contains Azure services. Only applicable for the variants where we group by vendor, including with meta packages.
```
moonbox3: One piece of feedback I have heard from SK users is that they expect that Microsoft-related components are installed with the base package. When we first introduced the Foundry agent, they didn't like having to do `pip install semantic-kernel[azure]`. We later moved the AzureAIAgent dependencies to install with the base package. Just a thought.

eavanvalkenburg: good point, raised this for the team as well

docs/design/python-package-setup.md
```diff
@@ -163,79 +163,83 @@ We might add a template subpackage as well, to make it easy to setup, this could
 
 In the [`DEV_SETUP.md`](../../python/DEV_SETUP.md) we will add instructions for how to deal with the path depth issues, especially on Windows, where the maximum path length can be a problem.
 
+### Subpackage scope
+The scope of subpackages is a balance between the number of subpackages we end up with, the size and complexity of each subpackage and its dependencies, and the user experience of installing and using the package.
```
ekzhu: I think for subpackages, the main benefits they bring:
1. Different release schedule from the main package -- integration that has a lot of moving pieces benefit from faster release schedule, especially if the integration target has constant breaking changes.
2. Potential separation of ownership and development setup -- it could potentially be moved outside of the microsoft/agent-framework repo to be maintained by a different ownership structure.

We can look at each potential sub-package on a case-by-case basis: for a candidate, if we are not foreseeing these benefits apply then extra is better and more Pythonic.

Issue #638, opened 2025-09-08
.NET: Revisit AIAgent.RunAsync accepting IEnumerable rather than IReadOnlyCollection


stephentoub:
M.E.AI.Abstractions uses IEnumerable. The argument made for AIAgent using IROC instead was that it'll necessarily need to make multiple iterations, but a) that doesn't seem to hold true, b) it can always make a copy if necessary, c) it can reduce such copies by type testing, and d) enumerable makes consumption easier, I think we should revisit this.

cc: @westey-m

westey-m:
Yes, I think this makes sense.  I can create a PR to update.

westey-m:
@stephentoub See attached PR.  Looks like there are still a good few places where we need to cast to avoid multiple executions, but we only typically need an IReadonlyCollection, which should match most input that isn't only an IEnumerable.

PR #729, opened 2025-09-12
.NET: BREAKING Update AIAgent.Run to take IEnumerable instead of IReadonlyCollection

westey-m:
### Motivation and Context

#638

### Description

- Updating Run methods to take IEnumerable instead of IReadonlyCollection.
- Updated code to avoid multiple executions of IEnumerable

### Contribution Checklist

### DIFF
--- a/dotnet/src/Microsoft.Agents.Orchestration/OrchestratingAgent.cs
+++ b/dotnet/src/Microsoft.Agents.Orchestration/OrchestratingAgent.cs
@@ -148,7 +148,7 @@ public abstract partial class OrchestratingAgent : AIAgent
     /// <param name="messages">The input message.</param>
     /// <param name="context">The context for this operation.</param>
     /// <param name="cancellationToken">A cancellation token that can be used to cancel the operation.</param>
-    protected abstract Task<AgentRunResponse> RunCoreAsync(IReadOnlyCollection<ChatMessage> messages, OrchestratingAgentContext context, CancellationToken cancellationToken);
+    protected abstract Task<AgentRunResponse> RunCoreAsync(IEnumerable<ChatMessage> messages, OrchestratingAgentContext context, CancellationToken cancellationToken);
 
     /// <summary>
     /// Resumes processing of the orchestration.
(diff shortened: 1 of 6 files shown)


dotnet/src/Microsoft.Agents.Orchestration/OrchestratingAgent.cs
```diff
@@ -131,8 +131,8 @@ public async ValueTask<OrchestratingAgentResponse> RunAsync(
 
         JsonElement? checkpoint = await this.ReadCheckpointAsync(context, cancellationToken).ConfigureAwait(false);
         Task<AgentRunResponse> completion = checkpoint is null ?
-            this.RunCoreAsync(messages, context, cancellationToken) :
-            this.ResumeCoreAsync(checkpoint.Value, messages, context, cancellationToken);
+            this.RunCoreAsync(readonlyCollectionMessages, context, cancellationToken) :
+            this.ResumeCoreAsync(checkpoint.Value, readonlyCollectionMessages, context, cancellationToken);
```
rogerbarreto, line 135: We should be able to also change the signature of the RunCoreAsync and REsumeCoreAsync to accept the ienumerable and do this checking / toList() within the method right unless you mutate the list from `Run` -> `Resume` which I don't think is that case, as this is a `IReadOnlyCollection`.

westey-m: Makes sense.  They are typically always used as a collection in all the implementations, but they are essentially part of our public interface for inheritors, so makes sense to also convert them.

westey-m:
> Thank you. This seems better to me, as it shifts the responsibility for materializing the lists to the agent implementations, which should be much fewer than agent consumers, both making it easier to consume (you can just pass in a LINQ query without needing extra ToArray/ToList/etc. or `[.. ]`, and any overhead of that materialization can be delayed until it's actually needed in the implementation, if it's needed. It also better aligns with the underlying M.E.AI abstractions.
> 
> Are you happy with it?

Yes, I think this makes sense.  I have always generally preferred having IEnumerable for input, since it is more flexible.  My concern was just with being efficient. Using the strategy of trying to cast to IReadonlyCollection when needed and possible, we will really only need to create a new list for actual Enumerable input, so it should all be very efficient.

PR #730, opened 2025-09-12
Python : Ollama Connector for Agent Framework

nmoeller:
### Motivation and Context

This pull request introduces a new Olama integration package for the Microsoft Agent Framework, allowing users to leverage Ollama-based chat models within the framework. The changes include the implementation of the `agent-framework-olama` package, its integration as an optional dependency, and the addition of a sample demonstrating its usage.

### Screenshot of Working Samples

image

### Screenshot of test coverage

image

### Description

- [x] Add Chat Completion
- [x] Add Tool Calling
- [x] Add Unit & Integration Tests

### Contribution Checklist


markwallace-microsoft:
<a href="https://github.com/microsoft/agent-framework/blob/883e51a56b406db0bba724bd6ae1c5a35ecd109e/README.md">Python Test Coverage</a><details><summary>Python Test Coverage Report • </summary><table><tr><th>File</th><th>Stmts</th><th>Miss</th><th>Cover</th><th>Missing</th></tr><tbody><tr><td><b>TOTAL</b></td><td><b>8821</b></td><td><b>1427</b></td><td><b>83%</b></td><td>&nbsp;</td></tr></tbody></table><i>report-only-changed-files is enabled. No files were changed during this commit :)</i></details>

## Python Unit Test Overview
| 1085 | 99 :zzz: | 0 :x: | 0 :fire: | 38.294s :stopwatch: |

python/packages/main/pyproject.toml
```diff
@@ -29,6 +29,7 @@ dependencies = [
     "typing-extensions>=4.14.0",
     "opentelemetry-api ~= 1.24",
     "opentelemetry-sdk ~= 1.24",
+    "opentelemetry-semantic-conventions-ai ~=0.4.13",
```
eavanvalkenburg: there is already a PR open for this.

python/packages/main/pyproject.toml
```diff
@@ -42,6 +43,9 @@ azure = [
 foundry = [
     "agent-framework-foundry"
 ]
+olama = [
```
eavanvalkenburg: ollama is with 2 `l`'s

python/packages/olama/agent_framework_olama/_chat_client.py
```diff
@@ -0,0 +1,114 @@
...
+class OlamaSettings(AFBaseSettings):
+    """Olama settings."""
+
+    env_prefix: ClassVar[str] = "OLLAMA_"
+
+    host: str | None = None
+    chat_model_id: str | None = None
+
+
+class OlamaChatClient(BaseChatClient):
```
eavanvalkenburg: you need to add @use_telemetry and @use_function_invocation

python/packages/olama/agent_framework_olama/_chat_client.py
```diff
@@ -0,0 +1,114 @@
...
+    def _olama_chat_message_parser(self, message: ChatMessage) -> list[Message]:
+        messages: list[Message] = []
+        for content in message.contents:
+            if isinstance(content, TextContent):
+                messages.append(Message(role=message.role.value, content=content.text))
+        return messages
+
+    def _create_chat_response_update(self, response: OllamaChatResponse) -> ChatResponseUpdate:
+        if response.message.content:
+            return ChatResponseUpdate(text=response.message.content, role=Role.ASSISTANT)
```
eavanvalkenburg: there should be function call responses here as well

python/.vscode/settings.json
```diff
@@ -35,5 +35,10 @@
             "name": "azure",
             "depth": 2
         }
-    ]
+    ],
+    "python.testing.pytestArgs": [
```
eavanvalkenburg: please move this into your own user settings

python/packages/ollama/agent_framework_ollama/_chat_client.py
```diff
@@ -0,0 +1,245 @@
...
+logger = get_logger("agent_framework.ollama")
+TOllamaChatClient = TypeVar("TOllamaChatClient", bound="OllamaChatClient")
+
+
+@use_function_invocation
+@use_telemetry
+class OllamaChatClient(BaseChatClient):
+    """Ollama Chat completion class."""
+
+    client: AsyncClient
```
eavanvalkenburg, line 56: there should be a OTEL_PROVIDER_NAME var here, similar to other connectors

python/packages/ollama/agent_framework_ollama/_chat_client.py
```diff
@@ -0,0 +1,245 @@
...
+from ollama._types import Message as OllamaMessage
+
+
+class OllamaSettings(AFBaseSettings):
+    """Ollama settings."""
+
+    env_prefix: ClassVar[str] = "OLLAMA_"
+
+    host: str | None = None
+    chat_model_id: str | None = None
```
eavanvalkenburg, line 43:
```suggestion
    chat_model_id: str
```

python/packages/ollama/agent_framework_ollama/_chat_client.py
```diff
@@ -0,0 +1,245 @@
...
+        """Initialize an Ollama Chat client.
+
+        Args:
+            host: The Ollama server host URL. Can be set via the OLLAMA_HOST env variable.
+            client: An optional Ollama Client instance. If not provided, a new instance will be created.
+            chat_model_id: The Ollama chat model ID to use. Can be set via the OLLAMA_CHAT_MODEL_ID env variable.
+            env_file_path: An optional path to a dotenv (.env) file to load environment variables from.
+            env_file_encoding: The encoding to use when reading the dotenv (.env) file. Defaults to 'utf-8'.
+        """
+        ollama_settings = OllamaSettings(
```
eavanvalkenburg: surround this with a try except to catch the Pydantic error when there is not chat_model_id, for now while it is the only type of connectors, we can just not allow a None on that field and catch

python/packages/ollama/agent_framework_ollama/_chat_client.py
```diff
@@ -0,0 +1,245 @@
...
+        )
+
+    async def _inner_get_response(
+        self,
+        *,
+        messages: MutableSequence[ChatMessage],
+        chat_options: ChatOptions,
+        **kwargs: Any,
+    ) -> ChatResponse:
+        return await ChatResponse.from_chat_response_generator(
```
eavanvalkenburg: I would prefer this to implement `self.client.chat(stream=False)` because in the past there were differences in support between streaming and non-streaming

python/packages/ollama/agent_framework_ollama/_chat_client.py
```diff
@@ -0,0 +1,245 @@
...
+        if chat_options.tool_choice == "required":
+            raise ServiceInvalidRequestError("Ollama does not support required tool choice.")
+        # Always auto: remove tool_choice since Ollama does not expose configuration to force or disable tools.
+        if "tool_choice" in options_dict:
+            del options_dict["tool_choice"]
+
+        return options_dict
+
+    def _prepare_chat_history_for_request(self, messages: MutableSequence[ChatMessage]) -> list[OllamaMessage]:
+        list_of_list = [self._agent_framework_to_ollama_messages(msg) for msg in messages]
```
eavanvalkenburg: this potentially creates more ollama messages then equivalent AF messages, is that by design? in other words, does the ollama api not allow a message with lets say 2 tool calls?

python/packages/ollama/agent_framework_ollama/_chat_client.py
nmoeller: Checking this there a conflict with the constructor of the class, the constructor has as `chat_model_id: str | None = None` , which then conflicts with this since the `OllamaSettings()` constructor, which throws a Error that these value is not allowed to be `None`.

image

Other Settings like Foundry or AzureOpenAI settings l have `None` default values even when it is required. 

If that's fine I will stick to how this is handled in the other Packages :

1. Ollama Settings `chat_model_id: str | None = None`
2. The Ollama Chat Client will catch a empty `chat_model_id` in the constructor.

```
if ollama_settings.chat_model_id is None:
    raise ServiceInitializationError(
        "Ollama chat model ID must be provided via chat_model_id or OLLAMA_CHAT_MODEL_ID environment variable."
    )
```

python/packages/ollama/agent_framework_ollama/_chat_client.py
nmoeller: This is by design, the reason here is that there is a slight difference between AF and Ollama. 

In AF each `ChatMessage` has a `list[Contents]`. After AF completes a parallel Tool Call we have a `ChatMessage` with 2 `FunctionResultContent` in the list.

In Ollama there is no option to handle both tool responses in a single message according to this Sample/Class Structure (https://github.com/ollama/ollama-python/blob/main/examples/multi-tool.py). This means we have to convert a single AF ChatMessage into 2 Ollama Messages when we have parallel Tool Calling. 

I can also see that the OpenAI implementation is similar in AF.

Hope that clarifies the question and I did my research properly.

PR #735, opened 2025-09-12
.NET: [BREAKING] Support Checkpoint Serialization

lokitoth:
### Motivation and Context

In order to support durable checkpointing, it needs to be possible to serialize the checkpoints. This is a little complicated because they could contain deeply nested user-defined types, so we must provide a way for the user to provide serialization information.

Since we expect JSON to be one of the more common formats (and because we need to support it for running in Agent Runtime), we implement abstractions for JSON-backed durable checkpoints.

This results in some breaks to the interfaces and object model, particularly around handling External requests.

### Description

* Implements serialization roundtripping for checkpoints.
* Adds support for JSON serialization
* Adds FileSystem-based checkpoint persistence

### Contribution Checklist

### DIFF
--- a/.github/workflows/dotnet-build-and-test.yml
+++ b/.github/workflows/dotnet-build-and-test.yml
@@ -55,7 +55,7 @@ jobs:
             .github
             dotnet
       - name: Setup dotnet
-        uses: actions/setup-dotnet@v4.3.1
+        uses: actions/setup-dotnet@v5.0.0
         with:
           global-json-file: ${{ github.workspace }}/dotnet/global.json
       - name: Build dotnet solutions
@@ -139,7 +139,7 @@ jobs:
 
       # Generate test reports and check coverage
       - name: Generate test reports
-        uses: danielpalme/ReportGenerator-GitHub-Action@5.4.11
+        uses: danielpalme/ReportGenerator-GitHub-Action@5.4.13
         with:
           reports: "./TestResults/Coverage/**/coverage.cobertura.xml"
           targetdir: "./TestResults/Reports"
@@ -187,13 +187,13 @@ jobs:
       - name: Fail workflow if tests failed
         id: check_tests_failed
         if: contains(join(needs.*.result, ','), 'failure')
-        uses: actions/github-script@v7
+        uses: actions/github-script@v8
         with:
           script: core.setFailed('Integration Tests Failed!')
 
       - name: Fail workflow if tests cancelled
         id: check_tests_cancelled
         if: contains(join(needs.*.result, ','), 'cancelled')
-        uses: actions/github-script@v7
+        uses: actions/github-script@v8
         with:
           script: core.setFailed('Integration Tests Cancelled!')
--- a/.github/workflows/dotnet-cosmosdb-integration-tests.yml
+++ b/.github/workflows/dotnet-cosmosdb-integration-tests.yml
@@ -47,7 +47,7 @@ jobs:
           AZURE_COSMOS_EMULATOR_PARTITION_COUNT: "20" # the more the better for stable tests
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           persist-credentials: false
           sparse-checkout: |
@@ -55,7 +55,7 @@ jobs:
             .github
             dotnet
       - name: Setup dotnet
-        uses: actions/setup-dotnet@v4.3.1
+        uses: actions/setup-dotnet@v5.0.0
         with:
           global-json-file: ${{ github.workspace }}/dotnet/global.json
       - name: Build dotnet solutions
@@ -145,7 +145,7 @@ jobs:
 
       # Generate test reports and check coverage
       - name: Generate test reports
-        uses: danielpalme/ReportGenerator-GitHub-Action@5.4.11
+        uses: danielpalme/ReportGenerator-GitHub-Action@5.4.13
         with:
           reports: "./TestResults/Coverage/**/coverage.cobertura.xml"
           targetdir: "./TestResults/Reports"
--- a/.github/workflows/label-issues.yml
+++ b/.github/workflows/label-issues.yml
@@ -13,7 +13,7 @@ jobs:
     permissions:
       issues: write
     steps:
-      - uses: actions/github-script@v7
+      - uses: actions/github-script@v8
         with:
           github-token: ${{ secrets.GH_ACTIONS_PR_WRITE }}
           script: |
--- a/.github/workflows/label-title-prefix.yml
+++ b/.github/workflows/label-title-prefix.yml
@@ -15,7 +15,7 @@ jobs:
       pull-requests: write
 
     steps:
-      - uses: actions/github-script@v7
+      - uses: actions/github-script@v8
         name: "Issue/PR: update title"
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/python-merge-tests.yml
+++ b/.github/workflows/python-merge-tests.yml
@@ -251,13 +251,13 @@ jobs:
       - name: Fail workflow if tests failed
         id: check_tests_failed
         if: contains(join(needs.*.result, ','), 'failure')
-        uses: actions/github-script@v7
+        uses: actions/github-script@v8
         with:
           script: core.setFailed('Integration Tests Failed!')
 
       - name: Fail workflow if tests cancelled
         id: check_tests_cancelled
         if: contains(join(needs.*.result, ','), 'cancelled')
-        uses: actions/github-script@v7
+        uses: actions/github-script@v8
         with:
           script: core.setFailed('Integration Tests Cancelled!')
--- a/.gitignore
+++ b/.gitignore
@@ -197,3 +197,9 @@ tmp*/
 temp*/
 .tmp/
 .temp/
+
+# GAIA data directories
+data_gaia_hub/
+**/data_gaia_hub/
+python/packages/lab/gaia/**/*.jsonl
+
--- a/dotnet/Directory.Packages.props
+++ b/dotnet/Directory.Packages.props
@@ -85,7 +85,7 @@
     <PackageVersion Include="Microsoft.ML.OnnxRuntimeGenAI" Version="0.9.1" />
     <PackageVersion Include="OllamaSharp" Version="5.3.6" />
     <!-- Identity -->
-    <PackageVersion Include="Microsoft.Identity.Client.Extensions.Msal" Version="4.74.1" />
+    <PackageVersion Include="Microsoft.Identity.Client.Extensions.Msal" Version="4.77.0" />
     <!-- Workflows -->
     <PackageVersion Include="Microsoft.Bot.ObjectModel" Version="1.2025.915-1" />
     <PackageVersion Include="Microsoft.Bot.ObjectModel.Json" Version="1.2025.915-1" />
--- a/dotnet/samples/GettingStarted/Agents/Agent_Step06_PersistedConversations/Program.cs
+++ b/dotnet/samples/GettingStarted/Agents/Agent_Step06_PersistedConversations/Program.cs
@@ -40,7 +40,7 @@ await File.WriteAllTextAsync(tempFilePath, JsonSerializer.Serialize(serializedTh
 JsonElement reloadedSerializedThread = JsonSerializer.Deserialize<JsonElement>(await File.ReadAllTextAsync(tempFilePath));
 
 // Deserialize the thread state after loading from storage.
-AgentThread resumedThread = await agent.DeserializeThreadAsync(reloadedSerializedThread);
+AgentThread resumedThread = agent.DeserializeThread(reloadedSerializedThread);
 
 // Run the agent again with the resumed thread.
 Console.WriteLine(await agent.RunAsync("Now tell the same joke in the voice of a pirate, and add some emojis to the joke.", resumedThread));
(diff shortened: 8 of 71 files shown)


dotnet/src/Microsoft.Agents.Workflows/Checkpointing/PortableValue.cs
```diff
@@ -0,0 +1,177 @@
...
+            {
+                this._deserializedValueCache = delayedDeserialization.Deserialize<TValue>();
+            }
+        }
+
+        if (this.Value is TValue typedValue)
+        {
+            return typedValue;
+        }
+
```
alliscode: Should `_deserializedValueCache ` be nulled out here?

dotnet/src/Microsoft.Agents.Workflows/Checkpointing/FileSystemJsonStore.cs
```diff
@@ -0,0 +1,171 @@
...
+
+        try
+        {
+            // read the lines of indexfile and parse them as CheckpointInfos
+            this.CheckpointIndex = new HashSet<CheckpointInfo>();
+            using StreamReader reader = new(this._indexFile, encoding: Encoding.UTF8, detectEncodingFromByteOrderMarks: false, bufferSize: -1, leaveOpen: true);
+            while (!reader.EndOfStream)
+            {
+                string? line = reader.ReadLine();
+                if (line != null)
```
stephentoub: Nit: EndOfStream is an API I wish didn't exist... this can just be:
```csharp
while (reader.ReadLine() is string line)
```
See https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-10/#i/o for me further ranting about it :)

dotnet/src/Microsoft.Agents.Workflows/Checkpointing/FileSystemJsonStore.cs
```diff
@@ -0,0 +1,171 @@
...
+/// <summary>
+/// Provides a file system-based implementation of a JSON checkpoint store that persists checkpoint data and index
+/// information to disk using JSON files.
+/// </summary>
+/// <remarks>This class manages checkpoint storage by writing JSON files to a specified directory and maintaining
+/// an index file for efficient retrieval. It is intended for scenarios where durable, process-exclusive checkpoint
+/// persistence is required. Instances of this class are not thread-safe and should not be shared across multiple
+/// threads without external synchronization. The class implements IDisposable; callers should ensure Dispose is called
+/// to release file handles and system resources when the store is no longer needed.</remarks>
+public sealed class FileSystemJsonStore : JsonCheckpointStore, IDisposable
```
stephentoub: Nit: FileSystemJsonStore => FileSystemJsonCheckpointStore?

dotnet/src/Microsoft.Agents.Workflows/Checkpointing/FileSystemJsonStore.cs
```diff
@@ -0,0 +1,171 @@
...
+
+    /// <inheritdoc/>
+    public void Dispose()
+    {
+        FileStream? indexFileLocal = Interlocked.Exchange(ref this._indexFile, null);
+        indexFileLocal?.Dispose();
+    }
+
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Maintainability", "CA1513:Use ObjectDisposedException throw helper",
+        Justification = "Throw helper does not exist in NetFx 4.7.2")]
```
stephentoub: We should probably just disable CA1510-CA1513 (or downgrade to suggestion) in the .editorconfig for the whole repo, since they'll just cause noise like this with the netstandard/netfx TFMs.

dotnet/src/Microsoft.Agents.Workflows/Checkpointing/FileSystemJsonStore.cs
```diff
@@ -0,0 +1,171 @@
...
+    private string GetFileNameForCheckpoint(string runId, CheckpointInfo key)
+        => Path.Combine(this.Directory.FullName, $"{runId}_{key.CheckpointId}.json");
+
+    private CheckpointInfo GetUnusedCheckpointInfo(string runId)
+    {
+        CheckpointInfo key;
+        do
+        {
+            key = new(runId);
+        } while (!this.CheckpointIndex.Add(key));
```
stephentoub: I don't understand this... is runId supposed to be changing on each iteration?

stephentoub: Oh, the CheckpointInfo ctor being used here is doing Guid.NewGuid... in which case in what situation would Add return false?

dotnet/src/Microsoft.Agents.Workflows/Checkpointing/FileSystemJsonStore.cs
```diff
@@ -0,0 +1,171 @@
...
+        CheckpointInfo key = this.GetUnusedCheckpointInfo(runId);
+        string fileName = this.GetFileNameForCheckpoint(runId, key);
+        try
+        {
+            using Stream checkpointStream = File.Open(fileName, FileMode.Create, FileAccess.Write, FileShare.None);
+            using Utf8JsonWriter jsonWriter = new(checkpointStream);
+            value.WriteTo(jsonWriter);
+
+            JsonSerializer.Serialize(this._indexFile!, key, this.KeyTypeInfo);
+            byte[] bytes = Encoding.UTF8.GetBytes(Environment.NewLine);
```
stephentoub: Why do we need to write out the newline?

dotnet/src/Microsoft.Agents.Workflows/Checkpointing/FileSystemJsonStore.cs
```diff
@@ -0,0 +1,171 @@
...
+        if (!this.CheckpointIndex.Contains(key) ||
+            !File.Exists(fileName))
+        {
+            throw new KeyNotFoundException($"Checkpoint '{key.CheckpointId}' not found in store at '{this.Directory.FullName}'.");
+        }
+
+        using FileStream checkpointFileStream = File.Open(fileName, FileMode.Open, FileAccess.Read, FileShare.Read);
+        JsonDocument document = await JsonDocument.ParseAsync(checkpointFileStream).ConfigureAwait(false);
+
+        return document.RootElement;
```
stephentoub:
```suggestion
        using JsonDocument document = await JsonDocument.ParseAsync(checkpointFileStream).ConfigureAwait(false);

        return document.RootElement.Clone();
```

dotnet/src/Microsoft.Agents.Workflows/Checkpointing/PortableValue.cs
```diff
@@ -0,0 +1,177 @@
...
+    /// <typeparam name="TValue">The type to which the value should be cast or deserialized.</typeparam>
+    /// <returns>The value cast or deserialized to type TValue if possible; otherwise, the default value for type TValue.</returns>
+    public TValue? As<TValue>()
+    {
+        if (this.Value is IDelayedDeserialization delayedDeserialization)
+        {
+            if (this._deserializedValueCache == null)
+            {
+                this._deserializedValueCache = delayedDeserialization.Deserialize<TValue>();
+            }
```
stephentoub: Nit:
```suggestion
            this._deserializedValueCache ??= delayedDeserialization.Deserialize<TValue>();
```

dotnet/src/Microsoft.Agents.Workflows/Checkpointing/Serialization.cs
```diff
@@ -0,0 +1,405 @@
...
+        }
+        else if (obj is not IDelayedDeserialization)
+        {
+            // Assume this has the target type of deserialization; serialize it using the explicit type
+            // and compare.
+            try
+            {
+                JsonElement otherElement = serializer.Marshal(obj, obj.GetType());
+
+                // Unfortunately, JsonElement does not direclty compare to JsonElement
```
stephentoub: There's a JsonElement.DeepEquals method... can you use that?

dotnet/src/Microsoft.Agents.Workflows/Checkpointing/Serialization.cs
```diff
@@ -0,0 +1,405 @@
...
+             : $"@{value.Id}";
+    }
+}
+
+internal sealed class ScopeKeyConverter : JsonConverterDictionarySupportBase<ScopeKey>
+{
+    protected override JsonTypeInfo<ScopeKey> TypeInfo => WorkflowsJsonUtilities.JsonContext.Default.ScopeKey;
+
+    public static readonly Regex ScopeKeyPropertyNamePattern = new(@"^(?<executorId>(?:(?:(?:\|\|)|(?:[^\|]))*))\|(?<scopeName>(?:@(?:(?:(?:\|\|)|(?:[^\|]))*))?)\|(?<key>(?:(?:(?:\|\|)|(?:[^\|]))*)?)$",
+                                                                   RegexOptions.Compiled | RegexOptions.CultureInvariant);
```
stephentoub: If you want to simplify the regex pattern a bit, you could use RegexOptions.ExplicitCapture and then replace the non-capturing groups, e.g. `(?: )` => `( )`

dotnet/src/Microsoft.Agents.Workflows/Checkpointing/Serialization.cs
```diff
@@ -0,0 +1,405 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
alliscode: Wow, there's a lot going on in this file, looks great! My main ask is to add way more comments to explain what all the converters are doing and why.

dotnet/samples/GettingStarted/Workflows/Checkpoint/CheckpointWithHumanInTheLoop/Program.cs
```diff
@@ -102,9 +102,9 @@ private static async Task Main()
 
     private static ExternalResponse HandleExternalRequest(ExternalRequest request)
     {
-        if (request.Port.Request == typeof(SignalWithNumber))
+        var signal = request.Data.As<SignalWithNumber>();
```
alliscode: I've seen `Data.As<...>` and `DataAs<...>`. Do we want to prefer one over the other in samples?

dotnet/src/Microsoft.Agents.Workflows/Checkpointing/Serialization.cs
```diff
@@ -0,0 +1,405 @@
...
+    protected override ExecutorIdentity Parse(string propertyName)
+    {
+        if (propertyName.Length == 0)
+        {
+            return ExecutorIdentity.None;
+        }
+
+        if (propertyName[0] == '@')
+        {
+            return new() { Id = propertyName.Substring(1) };
```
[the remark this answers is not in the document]
lokitoth: Not available in 4.7.2 without adding extra dependencies.

dotnet/src/Microsoft.Agents.Workflows/Checkpointing/PortableValue.cs
lokitoth: No, the other comment was correct.

dotnet/src/Microsoft.Agents.Workflows/Checkpointing/FileSystemJsonStore.cs
lokitoth: I really like that construction.

dotnet/src/Microsoft.Agents.Workflows/Checkpointing/FileSystemJsonStore.cs
lokitoth: Guids technically can collide. I've been struck by the cosmic ray once, so I just add the check reflexively now.

Is .NewGuid() collision-free, or does it use the RNG method?

dotnet/src/Microsoft.Agents.Workflows/Checkpointing/Serialization.cs
lokitoth: Seems to work

dotnet/src/Microsoft.Agents.Workflows/Checkpointing/Serialization.cs
lokitoth: I forgot that exists! Thanks!

dotnet/samples/GettingStarted/Workflows/Checkpoint/CheckpointWithHumanInTheLoop/Program.cs
lokitoth: The only reason to prefer one over the other is to avoid having to make people think about the details of PortableValue. I'll normalize it on DataAs.

dotnet/src/Microsoft.Agents.Workflows/Checkpointing/Serialization.cs
lokitoth: I broke apart the file - had intended to do that before PR. Also added comments.

dotnet/src/Microsoft.Agents.Workflows/Checkpointing/FileSystemJsonStore.cs
lokitoth: Yes.

dotnet/src/Microsoft.Agents.Workflows/Checkpointing/FileSystemJsonStore.cs
lokitoth: Good catch

dotnet/src/Microsoft.Agents.Workflows/Checkpointing/FileSystemJsonStore.cs
lokitoth: I could not figure out how to make the JSON Serializer to end with a NewLine, and this seemed to be the easiest way to get a JSONL formatting.

Funny thing though: Investigating this again made me realize that the unit test was not actually testing Checkpoint serialization.

PR #738, opened 2025-09-14
.NET: [BREAKING] Enable more analyzers and various code tweaks/cleanup

stephentoub:


dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/InMemoryChatMessageStore.cs
```diff
@@ -76,18 +77,18 @@ public InMemoryChatMessageStore(IChatReducer? chatReducer, JsonElement serialize
...
 
-        this._messages = new();
+        this._messages = [];
     }
 
     /// <summary>
     /// Gets the chat reducer used to process or reduce chat messages. If null, no reduction logic will be applied.
     /// </summary>
-    public IChatReducer? ChatReducer => this._chatReducer;
+    public IChatReducer? ChatReducer { get; }
```
westey-m, line 78: Shouldn't we be setting the ChatReducer and ReducerTrigerEvent properties in the constructor, now that they are not pointing at fields anymore?

stephentoub: Yes, bad merge, thanks

PR #779, opened 2025-09-16
.NET: Update AgentThread, MessageStores and Context Providers to deserialize via constructor.

westey-m:
### Motivation and Context

Enabling these classes to be deserialized only via the constructor reduces the chance of misuse and confusion.  E.g. a developer wouldn't be able to overwrite the state of an existing object anymore.

#717

### Description

- Remove the deserialize method from AgentThread, MessageStore and ContextProviders.
- Add a new deserializing constructor.
- Since all deserialization is via constructors, async is not possible anymore, so make agent deserialize thread method sync.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/InMemoryChatMessageStore.cs
+++ b/dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/InMemoryChatMessageStore.cs
@@ -17,13 +17,14 @@ public sealed class InMemoryChatMessageStore : IList<ChatMessage>, IChatMessageS
 {
     private readonly IChatReducer? _chatReducer;
     private readonly ChatReducerTriggerEvent _reducerTriggerEvent;
-    private List<ChatMessage> _messages = new();
+    private List<ChatMessage> _messages;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="InMemoryChatMessageStore"/> class.
     /// </summary>
     public InMemoryChatMessageStore()
     {
+        this._messages = new();
     }
 
     /// <summary>
@@ -61,6 +62,7 @@ public sealed class InMemoryChatMessageStore : IList<ChatMessage>, IChatMessageS
 
         if (serializedStoreState.ValueKind != JsonValueKind.Object)
         {
+            this._messages = new();
             return;
         }
 
@@ -70,8 +72,11 @@ public sealed class InMemoryChatMessageStore : IList<ChatMessage>, IChatMessageS
 
         if (state?.Messages is { Count: > 0 } messages)
         {
-            this._messages.AddRange(messages);
+            this._messages = messages;
+            return;
         }
+
+        this._messages = new();
     }
 
     /// <summary>
@@ -174,7 +179,7 @@ public sealed class InMemoryChatMessageStore : IList<ChatMessage>, IChatMessageS
 
     internal sealed class StoreState
     {
-        public IList<ChatMessage> Messages { get; set; } = new List<ChatMessage>();
+        public List<ChatMessage> Messages { get; set; } = new List<ChatMessage>();
     }
 
     /// <summary>


dotnet/samples/GettingStarted/Agents/Agent_Step07_3rdPartyThreadStorage/Program.cs
```diff
@@ -38,12 +38,12 @@
      {
          Name = JokerName,
          Instructions = JokerInstructions,
-         ChatMessageStoreFactory = () =>
+         ChatMessageStoreFactory = (jsonElement, jso) =>
```
rogerbarreto, line 41: nit, possible typo?

```suggestion
         ChatMessageStoreFactory = (jsonElement, json) =>
```

westey-m: That is the jsonSerializerOptions, i.e. jso

dotnet/src/Microsoft.Agents.Workflows/WorkflowMessageStore.cs
```diff
@@ -14,6 +14,33 @@ internal class WorkflowMessageStore : IChatMessageStore
...
+        }
+
+        object? maybeState =
+            JsonSerializer.Deserialize(
+                serializedStoreState,
+                AgentAbstractionsJsonUtilities.DefaultOptions.GetTypeInfo(typeof(StoreState)));
+
+        if (maybeState is not StoreState state)
+        {
+            throw new JsonException("Invalid state format for WorkflowMessageStore.");
```
stephentoub: Is JsonException the right exception type here?

dotnet/src/Microsoft.Agents.Workflows/WorkflowMessageStore.cs
```diff
@@ -14,6 +14,33 @@ internal class WorkflowMessageStore : IChatMessageStore
...
+        }
+
+        this._chatMessages.Clear();
```
stephentoub: Why is Clear needed?

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AgentThread.cs
```diff
@@ -26,6 +26,46 @@ public AgentThread()
...
     }
 
+    /// <summary>
+    /// Initializes a new instance of the <see cref="AgentThread"/> class from serialized state.
+    /// </summary>
+    /// <param name="serializedThreadState">A <see cref="JsonElement"/> representing the serialized state of the thread.</param>
+    /// <param name="jsonSerializerOptions">Optional settings for customizing the JSON deserialization process.</param>
+    /// <param name="chatMessageStoreFactory">An optional factory function to create a custom <see cref="IChatMessageStore"/>.</param>
+    /// <param name="aiContextProviderFactory">An optional factory function to create a custom <see cref="AIContextProvider"/>.</param>
+    public AgentThread(
```
stephentoub, line 36: You're still pursuing the derivation approach, right? So these changes are just intermediaries?

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/InMemoryChatMessageStore.cs
```diff
@@ -26,15 +26,52 @@ public InMemoryChatMessageStore()
...
+        if (serializedStoreState.ValueKind != JsonValueKind.Object)
+        {
+            return;
+        }
+
+        var state = JsonSerializer.Deserialize(
+            serializedStoreState,
+            AgentAbstractionsJsonUtilities.DefaultOptions.GetTypeInfo(typeof(StoreState))) as StoreState;
+
+        if (state?.Messages is { Count: > 0 } messages)
```
stephentoub, line 73: What type is Messages here? If it's `List<T>`, could we do just `_messages = messages` and remove the `new()` on the field?

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AgentThread.cs
westey-m: Correct, this is intermediate, since adding it all to one PR gets a bit big.

dotnet/src/Microsoft.Agents.Workflows/WorkflowMessageStore.cs
westey-m: Oops, removed.  I moved this from deserialize and didn't look at it in enough detail.

dotnet/src/Microsoft.Agents.Workflows/WorkflowMessageStore.cs
westey-m: Updated the logic in here to throw an ArgumentException if the provided JsonElement is not an object, in which case the result from deserialize should never be null and it should be anything other than StoreState either, or JsonSerializer should throw.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/InMemoryChatMessageStore.cs
westey-m: Makes sense.  Updated to reduce unnecessary list creations.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AgentThread.cs
```diff
@@ -26,6 +26,46 @@ public AgentThread()
...
+        JsonElement serializedThreadState,
+        JsonSerializerOptions? jsonSerializerOptions = null,
+        Func<JsonElement, JsonSerializerOptions?, IChatMessageStore>? chatMessageStoreFactory = null,
+        Func<JsonElement, JsonSerializerOptions?, AIContextProvider>? aiContextProviderFactory = null)
+    {
+        if (serializedThreadState.ValueKind != JsonValueKind.Object)
+        {
+            throw new ArgumentException("The serialized thread state must be a JSON object.", nameof(serializedThreadState));
+        }
```
rogerbarreto, line 45: This seems to be a partial validation, also having an object do not garantee the serialized state is correct, we might just remove this check or check what comes after the `var state` if `null` or let it throw.

westey-m: JsonSerializer will also throw with a json exception if the json doesn't match.  One thing that isn't great is that it typically throws a pretty vague exception if you have an undefined jsonelement, so this helps avoid that.  If we don't have this check, the deserialization also succeeds for null objects, which means we need another check after anyway, so I'm not sure we save much here.

dotnet/src/Microsoft.Agents.Workflows/WorkflowMessageStore.cs
```diff
@@ -14,6 +15,30 @@ internal class WorkflowMessageStore : IChatMessageStore
...
+    }
+
+    public WorkflowMessageStore(JsonElement serializedStoreState, JsonSerializerOptions? jsonSerializerOptions = null)
+    {
+        if (serializedStoreState.ValueKind != JsonValueKind.Object)
+        {
+            throw new ArgumentException("The provided JsonElement must be a json object", nameof(serializedStoreState));
+        }
+
+        StoreState? state =
```
stephentoub, line 29: Should we catch any JsonExceptions from Deserialize and wrap them in a thrown ArgumentException?

Issue #782, opened 2025-09-16
.NET: Route messages are not handled polymorphically


stephentoub:
Registration appears to require the _exact_ message Type, as it's just looked up in a dictionary by that type. I expect that if I have a route handler for `IEnumerable<T>`, then anything that implements `IEnumerable<T>` should satisfy it, or if I have a handler for `List<T>`, `List<T>` or anything deriving from it should satisfy it, but that is not happening.

Repro:
```csharp
    [Fact]
    public async Task ReproAsync()
    {
        var workflow = new WorkflowBuilder(new StartExecutor()).Build<IEnumerable<string>>();
        StreamingRun run = await InProcessExecution.StreamAsync<IEnumerable<string>>(workflow, ["Hello, World!"]);
        await run.RunToCompletionAsync();
    }

    private sealed class StartExecutor : Executor
    {
        protected override RouteBuilder ConfigureRoutes(RouteBuilder routeBuilder) =>
            routeBuilder.AddHandler<IEnumerable<string>>(async (msg, ctx) =>
            {
                Debugger.Break();
                Console.WriteLine("In StartExecutor");
                await ctx.SendMessageAsync(msg);
            });
    }
```

Result:
```
System.NotSupportedException : No handler found for message type <>z__ReadOnlySingleElementList`1 in executor StartExecutor.
```

stephentoub:
Seems like there are a few issues here.
1. MessageRouter.RouteMessageAsync is using message.GetType() rather than using the declared type of the message.
2. The MessageRouter._typedHandlers dictionary is keyed on the single exact type. If that exact type is `IEnumerable<T>` and the message is of type `List<T>`, it won't find it in the dictionary, even though logically it should be handleable. This is called out in a comment in MessageRouter.CanHandle: "// For now we only support routing to handlers registered on the exact type (no base type delegation)". For me that's very unexpected.

markwallace-microsoft:
Looks like this will impact API surface so tagging as a v1.0

lokitoth:
Tracked by #3116

PR #783, opened 2025-09-16
.NET: Fix polymorphic routing of messages in MessageRouter

stephentoub:
Fixes https://github.com/microsoft/agent-framework/issues/782

Just learning the codebase, so I'm not 100% sure this is the right fix.

There are two main changes made:
1. Executor.ExecuteAsync now passes the declared type into RouteMessageAsync, which it uses to do the lookup instead of GetType
2. MessageRouter now maintains a dictionary of dynamic type tests; for any message that comes in whose type isn't an exact match, it then consults this lazily-computed dictionary of actual types seen. If a type is missing, it does the expensive per-route check, and then caches the answer for subsequent use.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -197,3 +197,9 @@ tmp*/
 temp*/
 .tmp/
 .temp/
+
+# GAIA data directories
+data_gaia_hub/
+**/data_gaia_hub/
+python/packages/lab/gaia/**/*.jsonl
+
--- a/dotnet/Directory.Packages.props
+++ b/dotnet/Directory.Packages.props
@@ -85,7 +85,7 @@
     <PackageVersion Include="Microsoft.ML.OnnxRuntimeGenAI" Version="0.9.1" />
     <PackageVersion Include="OllamaSharp" Version="5.3.6" />
     <!-- Identity -->
-    <PackageVersion Include="Microsoft.Identity.Client.Extensions.Msal" Version="4.74.1" />
+    <PackageVersion Include="Microsoft.Identity.Client.Extensions.Msal" Version="4.77.0" />
     <!-- Workflows -->
     <PackageVersion Include="Microsoft.Bot.ObjectModel" Version="1.2025.915-1" />
     <PackageVersion Include="Microsoft.Bot.ObjectModel.Json" Version="1.2025.915-1" />
--- a/dotnet/samples/GettingStarted/Agents/Agent_Step06_PersistedConversations/Program.cs
+++ b/dotnet/samples/GettingStarted/Agents/Agent_Step06_PersistedConversations/Program.cs
@@ -40,7 +40,7 @@ await File.WriteAllTextAsync(tempFilePath, JsonSerializer.Serialize(serializedTh
 JsonElement reloadedSerializedThread = JsonSerializer.Deserialize<JsonElement>(await File.ReadAllTextAsync(tempFilePath));
 
 // Deserialize the thread state after loading from storage.
-AgentThread resumedThread = await agent.DeserializeThreadAsync(reloadedSerializedThread);
+AgentThread resumedThread = agent.DeserializeThread(reloadedSerializedThread);
 
 // Run the agent again with the resumed thread.
 Console.WriteLine(await agent.RunAsync("Now tell the same joke in the voice of a pirate, and add some emojis to the joke.", resumedThread));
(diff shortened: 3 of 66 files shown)


dotnet/src/Microsoft.Agents.Workflows/Executor.cs
```diff
@@ -69,7 +69,7 @@ internal MessageRouter Router
     {
         await context.AddEventAsync(new ExecutorInvokedEvent(this.Id, message)).ConfigureAwait(false);
 
-        CallResult? result = await this.Router.RouteMessageAsync(message, context, requireRoute: true)
+        CallResult? result = await this.Router.RouteMessageAsync(message, messageType, context, requireRoute: true)
```
lokitoth, line 72: I think this will conflict with the Checkpoint Serialization change: Once we have to support the ability to Json Serialize (without compile-time type info) messages, the `messageType` argument changes to be `TypeId`, which only captures the assembly name and type name (since Type cannot be serialized).

Is there some way to precompute the full type information and cache it somehow so we can use TypeId in this design?

lokitoth: In particular, see how we capture the types as `TypeId` objects when constructing the `MessageRouter`: https://github.com/microsoft/agent-framework/pull/735/files#diff-554210025ed5587fa7a3881ebfbb8a103c3b07521f9dcdb59edc247594447098

stephentoub: > Is there some way to precompute the full type information and cache it somehow so we can use TypeId in this design?

If arbitrary message types are allowed, such that an arbitrary type could arrive at any time, I can't think of a way. There's no way to start at a Type `t` and explore every possible other Type in the system that could inherit from `t`; even if we were to do the incredibly expensive operation of walking through every type in every loaded assembly, additional assemblies could be loaded subsequently. The only types for which we could be guaranteed nothing might derive from them is those that are sealed, but a key case I think we need to handle is interfaces, which are obviously not.

At what point do we lose the Type information? Is it possible to arrange things in such a way that the moment a message is constructed that might possible be passed to a given Executor, that Executor is provided with its Type?

As noted, I'm most concerned about interfaces, like `IEnumerable<T>` and `IList<T>`. The AIAgent-related types are, for example, currently set up as for handling `List<ChatMessage>`, but AIAgent/IChatClient actually work in terms of `IEnumerable<ChatMessage>`, and there's nothing that guarantees a `List<T>` will be used in their inputs/outputs.

stephentoub: Another expensive enterprise would be translating a message's Type into a list of TypeIds for all of its base types and interface implementations, and flowing that around with the message. That that list could be consulted when doing the comparison, as a rudimentary IsAssignableFrom. That likely has various negative ramifications as well... but might be feasible.

lokitoth: We lose the type information when the message goes across a serialization boundary. Then on the incoming side we have the type when we can match the type exactly, though if it is a supertype, there will be truncation, unless it is explicitly configured for polymorphism.

Serialization can happen when we checkpoint, so think of it as effectively potentially happening to every message.

In some sense you can think of it as the problem of having a set of types you know your methods take, and are charged with trying to process an arbitrary JSON blob that happens to have the C# Assembly and TypeName of the object from which it was serialized.

---

One thing I can see helping here is in response to the Threading/Concurrency model issue: We are planning to make a change that will introduce a new default mode: Different Executor instances will run concurrently, but any one of them will run non-concurrently. (With appropriate fixes).

But the key change involved there will be about chasing the edges immediately when the message is sent, rather than on the step after to group the messages by receiving Executor. Once we do that, we can memoize the binding information (e.g. which actual handler is responsible, in some serializable way) - this will allow us to restore the type information on the way back.

The issue here, though, would be that in the case of interfaces we would have no way to materialize a concrete type to provide to the serialization stack to be able to produce the deserialized value.

(Fundamentally the issue is that we cannot go from TypeId to Type without a precomputed lookup table, so we have no way of persisting the real type information of the object across the roundtrip, so the best we can do is anchor based on the real type information that is available in MessageRouter).

lokitoth: Another thing we could do if enforce the user predeclaring which messages a given executor can send, but it would only be checkable at runtime. What it would do is give us a list of concrete types that would be expected when coming out of a given executor, which would give us a way to build a TypeId => Type table in a consistent manner.

```csharp
abstract ISet<Type> OutputTypes { get; }
```

Regardless of what we do, though, I do not see how we can avoid having to have the concrete type available to us outside of having an instance of a message of that type on at least one side of the pipe through which the message travels.

PR #785, opened 2025-09-16
Python: Add VS Code visualization setup instructions for Azure AI Foundry

huimiu:
### Contribution Checklist


crickman (approved):

python/samples/getting_started/workflow/README.md
```diff
@@ -98,6 +98,67 @@ Once comfortable with these, explore the rest of the samples below.
...
+          tracer_provider.add_span_processor(BatchSpanProcessor(exporter))
+          set_tracer_provider(tracer_provider)
+          print(f"Tracing enabled with endpoint: {otlp_endpoint}")
+      except Exception as e:
+          print(f"Failed to setup tracing: {e}")
+   ```
+
+3. **Set up environment variables**:
+   - **Required**: Set `AGENT_FRAMEWORK_WORKFLOW_ENABLE_OTEL=true` to enable OpenTelemetry tracing
+   - **Optional**: Set `FOUNDRY_OTLP_PORT` (defaults to 4317 if not specified) to customize the OTLP endpoint port
```
TaoChenOSU: I am assuming that this env var is reserved for the foundry VS Code extension, and users have to read this env var in their application code to set up the exporter. 

Would it be better if AF also consume this env var and set up the exporter on behalf of the user?

For example, 
1. By default, visualization is off, and it can be turned on via an API call (i.e. `enable_local_viz()`), which will add an `OTLPSpanExporter` to the global trace provider if there is one or create a new provider if there isn't one already.
2. We should also allow users to send telemetry data to other destinations while local viz is enabled. This can be achieved by allowing them add exporters to the global trace provider via an API (i.e. `add_span_exporter()`) if there is one.
3. It's also possible for users to create a trace provider, but they need to make sure it's done before any `enable_local_viz` and `add_span_exporter` (including `add_log_exporter`, `add_metric_exporter`) calls.

TaoChenOSU: @eavanvalkenburg What do you think?

eavanvalkenburg: so, in the tracing interface in VS code it shows two endpoints, gRPC and HTTP so would be good to specify which one to use, and it does have the full address rather then just the port and a copy button, so why not tell folks to copy that address and put it in OLTP_ENDPOINT env var?

huimiu: For multi-agent workflow visualization feature, we are using HTTP endpoint.

That's a great suggestion. I'll update the code to reuse OLTP_ENDPOINT env var.

eavanvalkenburg:
I have a incoming piece of work that will make setup even simpler...

Issue #772, opened 2025-09-16
Move to the new multi-agent orchestration patterns


markwallace-microsoft:
The sample complexity should not increase when we move to the new pattern

- [ ] Refactor the Microsoft.Agents.Orchestration package to use workflows
- [ ] Refactor the samples as needed

markwallace-microsoft:
@alliscode creating this for tracking purposes

stephentoub:
https://github.com/microsoft/agent-framework/pull/792 adds new APIs for creating sequential, concurrent, and handoff workflows of agents. Just group chat remains from the original orchestrations.

PR #792, opened 2025-09-17
.NET: Add AgentWorkflowBuilder with Sequential, Concurrent, Handoffs

stephentoub:
Adds a new AgentWorkflowBuilder type with static BuildSequential and BuildConcurrent accelerators. The intent is to replace the SequentialOrchestration/ConcurrentOrchestration. Once we're happy with these, I'll follow-up with GroupChat and Handoffs.

Contributes to https://github.com/microsoft/agent-framework/issues/772

These should really be able to accept any `IEnumerable<ChatMessage>`, but that's blocked by https://github.com/microsoft/agent-framework/issues/782.


stephentoub:
@lokitoth, please take a peek when you get a moment to confirm this makes sense to you (including the catch-all change, as otherwise the catch-all handler support seemed broken, at least based on my understanding of what it was supposed to achieve).

PR #796, opened 2025-09-17
Python: ADR: extended tools and types proposal

eavanvalkenburg:
### Motivation and Context

ADR for additional content types and tool types.

### Contribution Checklist

### DIFF
--- a/docs/decisions/0007-additional-types.md
+++ b/docs/decisions/0007-additional-types.md
@@ -11,7 +11,7 @@ consulted: taochenosu, alliscode, moonbox3, johanste, stephentoub
 
 ## Context and Problem Statement
 
-There are a number of capabilities that are available for chat clients that either lack a content types to accurately capture the response, or that lack a abstracted tool to allow interchangability between chat clients. This ADR captures the proposed new types and tools, the rationale, and the alternatives considered and rejected.
+There are a number of capabilities that are available for chat clients that either lack a content types to accurately capture the response, or that lack a abstracted tool to allow interchangeability between chat clients. This ADR captures the proposed new types and tools, the rationale, and the alternatives considered and rejected.
 
 Currently in MEAI there are: HostedCodeInterpreterTool, HostedFileSearchTool, HostedMcpServerTool, HostedWebSearchTool.
 
@@ -88,7 +88,7 @@ TBD
 
 ## Bash or Shell Tools
 
-Multiple API's currently support Image Generation tools, such as OpenAI Responses [here](https://platform.openai.com/docs/api-reference/responses/create#responses-create-tools). And Anthropic Claude [here](https://docs.claude.com/en/docs/agents-and-tools/tool-use/bash-tool).
+Multiple API's currently support Bash or Shell tools, such as OpenAI Responses [here](https://platform.openai.com/docs/api-reference/responses/create#responses-create-tools). And Anthropic Claude [here](https://docs.claude.com/en/docs/agents-and-tools/tool-use/bash-tool).
 
 They work by having a local bash session which is fed with commands to execute, and the output is returned to the agent, for instance for Claude, the commands are `command` or `restart`.
 
@@ -109,7 +109,7 @@ and for OpenAI:
 
 The commands should then be taken and executed with the local shell session. In OpenAI Agents SDK this is built in, with something called the `LocalShellExecutor`.
 
-The bash and computer use tools are often used in conjuction to allow more control over a local machine.
+The bash and computer use tools are often used in conjunction to allow more control over a local machine.
 
 ### Considered Options
 
@@ -157,7 +157,7 @@ Several of the specialized tools also return specialized response types that are
 
 ## Code interpreter response type
 
-The Code Interpreter tool returns a response that often comprised of code, other inputs, stdout logs and files (usually images, for instance plots).
+The Code Interpreter tool returns a response that is often comprised of code, other inputs, stdout logs and files (usually images, for instance plots).
 
 The interesting thing about the code interpreter is that it can play the role both of a tool that get's executed to provide the right context or as a way to generate the response. In the first case you could consider the inputs and outputs as Annotation or a form of Reasoning, while in the latter it makes more sense to consider it a Content.
 
--- a/docs/decisions/0007-additional-types.md
+++ b/docs/decisions/0007-additional-types.md
@@ -40,8 +40,17 @@ Multiple API's currently support Image Generation tools, such as OpenAI Response
 
 ### Decision Outcome
 
-TBD
+1. HostedImageGenerationTool - include a HostedImageGenerationTool that can be used like this:
 
+```python
+tools = [
+    get_weather,
+    HostedCodeInterpreterTool(),
+    HostedImageGenerationTool()
+]
+```
+Without additional parameters for now, Mistral AI does not take any additional parameters, while OpenAI has a bunch, which can be put into additional_properties for now.
+We can add uniform parameters later, if needed.
 
 ## Computer Use Tools
 
@@ -83,7 +92,7 @@ Support for this would entail providing a out of the box local setup for the com
 
 ### Decision Outcome
 
-TBD
+2. Nothing, allow the dict to be passed and for the moment we will not add samples as the receiving side of a computer use setup is quite complex, and depends on the local system, software and other factors. We will monitor usage and if this becomes more popular we can add a ComputerUseTool later.
 
 
 ## Bash or Shell Tools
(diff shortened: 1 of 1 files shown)


docs/decisions/0007-additional-types.md
```diff
@@ -0,0 +1,182 @@
...
+
+# Types
+
+Several of the specialized tools also return specialized response types that are not currently modeled in MEAI. Adding these types would allow for better type checking and intellisense.
+
+## Code interpreter response type
+
+The Code Interpreter tool returns a response that is often comprised of code, other inputs, stdout logs and files (usually images, for instance plots).
+
+The interesting thing about the code interpreter is that it can play the role both of a tool that get's executed to provide the right context or as a way to generate the response. In the first case you could consider the inputs and outputs as Annotation or a form of Reasoning, while in the latter it makes more sense to consider it a Content.
```
stephentoub, line 186: For M.E.AI, we almost merged this:
https://github.com/dotnet/extensions/commit/7390afd908c612e68c64b36ff4c86aeb170716f8#diff-db78b16564489aff8d17d23deed5bb1d16e99dd9c6602d6546016c2ff4156b6b
and only didn't because we needed to merge the rest of the PR and didn't have time to fully validate it. But I expect it's close to what we'd want.

docs/decisions/0007-additional-types.md
```diff
@@ -0,0 +1,182 @@
...
+```python
+tools = [
+    get_weather,
+    HostedCodeInterpreterTool(),
+    HostedTool(type="image_generation"),
+    HostedTool(type="computer_use_preview", environment="browser", display_width=1024, display_height=768),
+    HostedTool(type="local_shell")
+]
+```
+
```
stephentoub, line 161: Something we recently merged into M.E.AI.OpenAI is the ability to treat any ResponseTool (the OpenAI library's base representation for a Responses tool) as an AITool, so you can write:
```csharp
ResponseTool arbitraryResponseTool = ...;
ChatOptions options = new()
{
    Tools = new List<AITool>() { arbitraryResponseTool.AsAITool() },
}
```
or
```csharp
ResponseTool arbitraryResponseTool = ...;
ChatOptions options = new();
(options.Tools ??= []).Add(arbitraryResponseTool);
```
These generate a simple AITool-derived type that just wraps the arbitrary ResponseTool. Then when the OpenAI IChatClient implementation processes ChatOptions.Tools, it looks for that wrapper and just pulls out the underlying instance which it can then use.

This makes it trivial to break glass and use any ResponseTool, without needing it to be in the abstraction. Of course, this only works if you know you're using an IChatClient for that target service. But then that's mostly the case as well when just specifying arbitrary JSON and hoping that the service you're talking to represents it in that way.

We've also considered adding a generic HostedTool that's just configured with a JsonElement for the JSON to render, or a dictionary to serialize as JSON, or some such thing. That's akin to your last option.

docs/decisions/0007-additional-types.md
```diff
@@ -0,0 +1,182 @@
...
+}
+```
+
+The model then returns a call to make and it might need screenshots to be taken and sent.
+
+Support for this would entail providing a out of the box local setup for the computer to use, or good samples showing how this can be setup with playwright or docker for OpenAI and a local sandbox for Claude.
+
+### Considered Options
+
+1. ComputerUseTool
```
stephentoub, line 86: Creating an abstraction over computer use looks fairly complicated (not just the tool itself, but all possible requests for actions and handling of the responses to those requests). I think we should only embark on that if we're really sure it's impactful to a meaningful number of folks.

docs/decisions/0007-additional-types.md
```diff
@@ -0,0 +1,182 @@
...
+
+### Considered Options
+
+1. BashTool
+    - Pro: easier to use then a dict, can be extended later, and the additional_properties can be used to allow additional parameters, if those become needed.
+    - Con: only available in limited number of API's currently and expectation would likely be to also include a local setup for it, which will be difficult to unify across platforms and providers. Would also require a additional Content Type
+    - Alternative names: ShellTool, LocalShellTool, LocalBashTool
+2. Nothing, allow the dict to be passed and show samples of how this works.
+    - Pro: No extra effort needed.
+    - Con: less discoverable, no type checking.
```
stephentoub, line 131: It's not just less discoverable / no type checking, it's not portable. Even if you were ok working at the JSON level, you need to send different JSON based on who you're talking to.

docs/decisions/0007-additional-types.md
eavanvalkenburg: ah missed that one, I'll update to sync with that

docs/decisions/0007-additional-types.md
eavanvalkenburg: good to see, for python this is a bit less needed since we can just use a dict, but it might be nice

docs/decisions/0007-additional-types.md
eavanvalkenburg: yeah, I mention that as a con and I also consider that a pretty major Con

docs/decisions/0007-additional-types.md
```diff
@@ -0,0 +1,216 @@
...
+
+- Consistent naming with current types and tools.
+- Interoperability between chat clients.
+- Ease of implementation and adoption.
+
+# Tools
+
+## Image generation tools
+
+Multiple API's currently support Image Generation tools, such as OpenAI Responses [here](https://platform.openai.com/docs/api-reference/responses/create#responses-create-tools). And [Mistral](https://docs.mistral.ai/agents/connectors/image_generation/), For OpenAI and Mistral this can be enabled by adding: `{"type": "image_generation"}` to the tools list. You then get image data back in the response, inputs can be both text and images. Google so far, has only created specific models that can be prompted to return a generated image, no tool specified.
```
ericstj, line 30: Google's gemini models have ResponseModalities which is similar to image generation tool.  Like OpenAI not all models have image generation capability.  They also have dedicated image models (ImageGen) like OpenAI, but they don't support connecting those to their other LLMs via tools.

docs/decisions/0007-additional-types.md
```diff
@@ -0,0 +1,216 @@
...
+
+### Considered Options
+
+1. HostedImageGenerationTool
+    - Pro: easier to use then a dict, can be extended later, and the additional_properties can be used to allow additional parameters, if those become needed.
+    - Con: only available in limited number of API's currently.
+1. Nothing, allow the dict to be passed and show samples of how this works.
```
ericstj, line 37: What is the `dict` - I'm not familiar with what you mean here.

docs/decisions/0007-additional-types.md
```diff
@@ -0,0 +1,216 @@
...
+
+## Image generation tools
+
+Multiple API's currently support Image Generation tools, such as OpenAI Responses [here](https://platform.openai.com/docs/api-reference/responses/create#responses-create-tools). And [Mistral](https://docs.mistral.ai/agents/connectors/image_generation/), For OpenAI and Mistral this can be enabled by adding: `{"type": "image_generation"}` to the tools list. You then get image data back in the response, inputs can be both text and images. Google so far, has only created specific models that can be prompted to return a generated image, no tool specified.
+
+### Considered Options
+
+1. HostedImageGenerationTool
+    - Pro: easier to use then a dict, can be extended later, and the additional_properties can be used to allow additional parameters, if those become needed.
+    - Con: only available in limited number of API's currently.
```
ericstj, line 36: We're working on modeling this as middle-ware, so that even APIs that don't support it can add it.  I think even Gemini can be modeled as a tool, if it's transformed to ResponseModalities at the client layer.

docs/decisions/0007-additional-types.md
```diff
@@ -0,0 +1,216 @@
...
+
+```python
+tools = [
+    get_weather,
+    HostedCodeInterpreterTool(),
+    HostedImageGenerationTool()
+]
+```
+Without additional parameters for now, Mistral AI does not take any additional parameters, while OpenAI has a bunch, which can be put into additional_properties for now.
+We can add uniform parameters later, if needed.
```
ericstj, line 53: Knowing how to interpret image content is important, do you intend to specify how the content is provided?

docs/decisions/0007-additional-types.md
eavanvalkenburg: just a python dictionary, this is currently already handled by the code as a fallback/breakglass scenario, since all apis use a dictionary to send the tools to the service, we parse all built-in tools to a dict, and then if you pass in a dict, we just pass that through

docs/decisions/0007-additional-types.md
eavanvalkenburg: So would you pass in `HostedImageGenerationTool` to your tools and then for some clients, a middleware removes that and instead sets responseModalities as a option?

docs/decisions/0007-additional-types.md
eavanvalkenburg: we already handle returned content in appropriate ways, so if a image comes back, we parse that to a DataContent/UriContent, each connector that supports that should handle it appropriately

PR #798, opened 2025-09-17
.NET: [BREAKING] Subclass AgentThread so that different agents have their own threads with their own typed settings.

westey-m:
### Motivation and Context

#780

### Description

- Create an AgentThread subclass for each agent.
- Introduce some base abstract agent thread classes for common thread behaviors, e.g. InMemory and ServiceThreadId
- Update all usage instances

### Contribution Checklist


dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatClientAgent.cs
```diff
@@ -238,26 +238,50 @@ public override async IAsyncEnumerable<AgentRunResponseUpdate> RunStreamingAsync
...
     /// <inheritdoc/>
+#if NET5_0_OR_GREATER
+    public override ChatClientAgentThread GetNewThread()
+        => new()
+#else
     public override AgentThread GetNewThread()
-    {
-        var thread = new AgentThread
+        => new ChatClientAgentThread
+#endif
```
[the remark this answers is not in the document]
westey-m: The purpose here is covariant return types.

dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatClientAgent.cs
```diff
@@ -238,26 +238,50 @@ public override async IAsyncEnumerable<AgentRunResponseUpdate> RunStreamingAsync
...
-    }
 
     /// <inheritdoc/>
-    public override AgentThread DeserializeThread(JsonElement serializedThread, JsonSerializerOptions? jsonSerializerOptions = null, CancellationToken cancellationToken = default)
-        => new(serializedThread, jsonSerializerOptions, this._agentOptions?.ChatMessageStoreFactory, this._agentOptions?.AIContextProviderFactory);
+#if NET5_0_OR_GREATER
+    public override ChatClientAgentThread DeserializeThread(JsonElement serializedThread, JsonSerializerOptions? jsonSerializerOptions = null)
+#else
+    public override AgentThread DeserializeThread(JsonElement serializedThread, JsonSerializerOptions? jsonSerializerOptions = null)
+#endif
```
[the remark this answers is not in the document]
westey-m: The purpose here is covariant return types.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/InMemoryAgentThread.cs
```diff
@@ -0,0 +1,100 @@
...
+
+        this._messageStore =
+            messageStoreFactory?.Invoke(state?.StoreState ?? default, jsonSerializerOptions) ??
+            new InMemoryChatMessageStore(state?.StoreState ?? default, jsonSerializerOptions);
+    }
+
+    /// <summary>
+    /// Gets or sets the <see cref="InMemoryChatMessageStore"/> used by this thread.
+    /// </summary>
+    public InMemoryChatMessageStore? MessageStore => this._messageStore;
```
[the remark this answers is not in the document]
westey-m: Fixed.

dotnet/samples/GettingStarted/AgentOrchestration/AgentSample.cs
```diff
@@ -155,7 +155,7 @@ private static async Task AzureAIAgentsPersistentAgentCleanUpAsync(AIAgent agent
         // If a thread is provided, delete it as well.
         if (thread is not null)
         {
-            await persistentAgentsClient.Threads.DeleteThreadAsync(thread.ConversationId, cancellationToken);
+            await persistentAgentsClient.Threads.DeleteThreadAsync(((ChatClientAgentThread)thread).ConversationId, cancellationToken);
```
stephentoub: This downcasting seems undesirable. Is this because of lack of covariant return types on netfx?

dotnet/samples/GettingStarted/AgentProviders/Agent_With_CustomImplementation/Program.cs
```diff
@@ -96,5 +103,17 @@ private static IEnumerable<ChatMessage> CloneAndToUpperCase(IEnumerable<ChatMess
...
+
+        /// <summary>
+        /// A thread type for our custom agent that only supports in memory storage of messages.
+        /// </summary>
+        internal sealed class CustomAgentThread : InMemoryAgentThread
+        {
+            public CustomAgentThread()
+                : base() { }
+
+            public CustomAgentThread(JsonElement serializedThreadState, JsonSerializerOptions? jsonSerializerOptions = null)
```
stephentoub: For sample purposes, we don't necessarily want to encourage these ctors being public, right? We could make both of them internal?

dotnet/samples/GettingStarted/Agents/Agent_Step07_3rdPartyThreadStorage/Program.cs
```diff
@@ -38,12 +38,12 @@
      {
          Name = JokerName,
          Instructions = JokerInstructions,
-         ChatMessageStoreFactory = (jsonElement, jso) =>
+         ChatMessageStoreFactory = (ctx) =>
```
stephentoub: Nit: the parens aren't needed (here and in other similar cases below)

dotnet/samples/GettingStarted/Agents/Agent_Step13_Memory/Program.cs
```diff
@@ -63,22 +63,22 @@
...
-thread = agent.GetNewThread();
-
 // It is also possible to add the memory component to an individual thread only instead of all
 // threads via the factory above.
 // In this case we will also use the same user info object, so this thread will share the same
 // memories as the previous thread.
-thread.AIContextProvider = new UserInfoMemory(chatClient.AsIChatClient(), userInfo);
+// For this scenario, we have to know the underlying thread type used by the agent
+// to be able to provide the AIContextProvider.
+thread = new ChatClientAgentThread(new UserInfoMemory(chatClient.AsIChatClient(), userInfo));
```
stephentoub: Why is code external to an agent creating a thread? I thought that was something we actively wanted to discourage. If you have to know the underlying thread type in order to use a context provider, isn't that an enormous problem?

dotnet/src/Microsoft.Agents.Orchestration/OrchestratingAgent.cs
```diff
@@ -63,6 +63,24 @@ protected OrchestratingAgent(IReadOnlyList<AIAgent> agents, string? name = null)
     /// </summary>
     public Func<AgentRunResponseUpdate, ValueTask>? StreamingResponseCallback { get; set; }
 
+    /// <inheritdoc/>
+#if NET5_0_OR_GREATER
+    public override OrchestratingAgentThread GetNewThread()
```
stephentoub: I'm not sure how I feel about signatures differing based on TFM. I'm not 100% sure there are no problems that could result from this. I will ask around.

dotnet/src/Microsoft.Agents.Orchestration/OrchestratingAgent.cs
```diff
@@ -63,6 +63,24 @@ protected OrchestratingAgent(IReadOnlyList<AIAgent> agents, string? name = null)
     /// </summary>
     public Func<AgentRunResponseUpdate, ValueTask>? StreamingResponseCallback { get; set; }
 
+    /// <inheritdoc/>
+#if NET5_0_OR_GREATER
```
stephentoub: You can replace any such compilation constants with just `NET`. We don't build for core versions older than net8.0 anymore, as they're all out of support.

dotnet/src/Microsoft.Agents.Orchestration/OrchestratingAgentThread.cs
```diff
@@ -0,0 +1,18 @@
...
+
+using System.Text.Json;
+using Microsoft.Extensions.AI.Agents;
+
+namespace Microsoft.Agents.Orchestration;
+
+/// <summary>
+/// The thread implementation used by <see cref="OrchestratingAgent"/>.
+/// </summary>
+public class OrchestratingAgentThread : InMemoryAgentThread
```
stephentoub: Why does this need to be public?

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/A2AAgent.cs
```diff
@@ -51,15 +52,39 @@ public A2AAgent(A2AClient a2aClient, string? id = null, string? name = null, str
...
+    public override A2AAgentThread GetNewThread()
+        => new();
+#else
+    public override sealed AgentThread GetNewThread()
+        => new A2AAgentThread();
+#endif
+
+    /// <inheritdoc/>
+#if NET5_0_OR_GREATER
+    public override A2AAgentThread DeserializeThread(JsonElement serializedThread, JsonSerializerOptions? jsonSerializerOptions = null)
```
stephentoub: Even if it's possible to utilize covariant returns, it feels undesirable to do so. What do we gain by exposing the concrete type and locking ourselves into that? Shouldn't it be an implementation detail?

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/InMemoryAgentThread.cs
```diff
@@ -0,0 +1,98 @@
...
+        Func<JsonElement, JsonSerializerOptions?, InMemoryChatMessageStore>? messageStoreFactory = null)
+    {
+        if (serializedThreadState.ValueKind != JsonValueKind.Object)
+        {
+            throw new ArgumentException("The serialized thread state must be a JSON object.", nameof(serializedThreadState));
+        }
+
+        var state = JsonSerializer.Deserialize(
+            serializedThreadState,
+            AgentAbstractionsJsonUtilities.DefaultOptions.GetTypeInfo(typeof(InMemoryAgentThreadState))) as InMemoryAgentThreadState;
```
stephentoub, line 58: If we do this in multiple places, we might want a helper for it, at least internal if not public at some point

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/InMemoryAgentThread.cs
```diff
@@ -0,0 +1,98 @@
...
+        return JsonSerializer.SerializeToElement(state, AgentAbstractionsJsonUtilities.DefaultOptions.GetTypeInfo(typeof(InMemoryAgentThreadState)));
+    }
+
+    /// <inheritdoc />
+    protected internal override Task MessagesReceivedAsync(IEnumerable<ChatMessage> newMessages, CancellationToken cancellationToken = default)
+        => this.MessageStore.AddMessagesAsync(newMessages, cancellationToken);
+
+    internal sealed class InMemoryAgentThreadState
+    {
+        [JsonPropertyName("storeState")]
```
stephentoub: Is this necessary? Won't the defaults used by the context result in this anyway?

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/InMemoryChatMessageStore.cs
```diff
@@ -172,6 +173,7 @@ IEnumerator IEnumerable.GetEnumerator()
 
     internal sealed class StoreState
     {
+        [JsonPropertyName("messages")]
```
stephentoub: same question

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/A2AAgent.cs
westey-m: Sure, I can remove.  Thought it would be nice to not have to cast if you want to get at the agent specific bits, but it's certainly not required, and it does introduce extra complexity.

dotnet/src/Microsoft.Agents.Orchestration/OrchestratingAgent.cs
westey-m: I'll remove the covariance, so this is also removed then.

dotnet/src/Microsoft.Agents.Orchestration/OrchestratingAgent.cs
westey-m: I'll remove the covariance as recommend elsewhere.

dotnet/samples/GettingStarted/Agents/Agent_Step07_3rdPartyThreadStorage/Program.cs
westey-m: Fixed.

dotnet/samples/GettingStarted/AgentProviders/Agent_With_CustomImplementation/Program.cs
westey-m: Sure, I had them public since the whole class was internal, but happy to make them internal too to emphasize further.

dotnet/samples/GettingStarted/AgentOrchestration/AgentSample.cs
westey-m: This specific class isn't being used anymore, but in the case of the other samples where we were deleting threads, the agents were typically created using a helper which returns an AIAgent and not a typed agent.  Therefore the thread returned was also an AgentThread instead of a typed thread.
Since I have removed the covariance, we always have to cast now regardless.

dotnet/src/Microsoft.Agents.Orchestration/OrchestratingAgent.cs
stephentoub: FWIW, I checked with @ericstj, and we _think_ it's actually safe to do it, from a compilation/runtime/packaging perspective. That's separate from whether we want to.

dotnet/samples/GettingStarted/AgentOrchestration/AgentSample.cs
stephentoub: > we always have to cast now regardless

This feels really icky to me. Let's chat about it. We're taking something that's supposed to be an implementation detail and making it something we need to guarantee forever more, because we're saying devs need to downcast to use it.

westey-m: Yes, we should discuss.  A couple more thoughts in the meantime.
The main place where this shows up in our samples is where we delete threads.  And this only for agents that have server side threads that are deletable, so really just Foundry and Assistants.
I'm not sure whether we really need to delete threads and I'd probably vote for just removing this from our samples.
While it's nice to clean up after ourselves in sample code, it's probably super edge case in reality.  I expect that it's pretty hard to figure out when a user is finished with a conversation, since they usually just abandon it rather than taking an action to complete.

stephentoub: > I'm not sure whether we really need to delete threads and I'd probably vote for just removing this from our samples.

Agreed

dotnet/samples/GettingStarted/Agents/Agent_Step13_Memory/Program.cs
westey-m: I think we should chat more about this is one.
I agree this isn't great either, although I do think there are cases where we may want to allow users to provide existing context to a thread that isn't coming from a deserialized state.  E.g. someone has a thread id that they want to bootstrap the thread with.  They don't have a serialized thread, but they have the id.
Another possible option is to have more thread creation methods on the concrete agent types.  Ultimately any of this is breaking out of the abstraction.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/InMemoryAgentThread.cs
westey-m: Good point, I missed that the default is camel case.  I've removed these.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/InMemoryChatMessageStore.cs
westey-m: Thanks removed

dotnet/src/Microsoft.Agents.Orchestration/OrchestratingAgentThread.cs
westey-m: I don't think there's a good reason for this one, so made it internal.

dotnet/src/Microsoft.Agents.Orchestration/OrchestratingAgent.cs
ericstj: I found a way to break this.  If I derive from this type in a library that only targets `netstandard2.0` and then run in a framework that will get the newer implementation it results in a runtime exception:
```
Exception: System.TypeLoadException: Return type in method 'nslib.CopilotStudioAgentThread2.GetNewThread()' on type 'nslib.CopilotStudioAgentThread2' from assembly 'nslib, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' is not compatible with base type method 'mtLib.CopilotStudioAgentThread.GetNewThread()'.
```

This would go away if the derived type library also multi-targeted.

Just calling the base method from a netstandard library did not cause a problem.  The type system allows the reference to bind to the method with the more derived type for calling.
https://github.com/ericstj/sample-code/tree/covariantReturn

~~This might be a type-system bug, but its presence in existing runtimes means you may not want to do this on an unsealed type.~~

This is not a type system bug.  Techically the library is exposing a contract which cannot be enforced.  It's a violation for a type to derive from GetNewThread() and return anything that's less derived than it's base, however when a netstandard library overrides this method, that will always be the case.  If you could somehow force your derived types to multi-target like you do then this would be safe, or seal the type.  If it's unsealed though, I think this a pitfall.

dotnet/samples/GettingStarted/Agents/Agent_Step13_Memory/Program.cs
westey-m: As discussed, removed the public constructors in favor of more factory methods on the concrete agent implementations.

dotnet/samples/GettingStarted/AgentOrchestration/AgentSample.cs
westey-m: Thanks, I've removed DeleteThread from the core samples.  I left it in the migration samples, since they are for 1 to 1 migration and what the migration agent is instructed with, so keeping it in is probably valuable not to end up with half migrated code.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/InMemoryAgentThread.cs
westey-m: At the moment I am doing this in a few places, but they are all in different projects except for 2 in Agent.Abstractions.  We'd therefore have to make any helper public to get much value, or use a shared internal file.
It would have to be parameterized for the expected ValueKind, threadstate and TypeInfo, so the main saving would be just having a consistent argument exception message.  Am I missing something in terms of value?  Happy to revisit this when we have more benefit from a common utility for this.

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/A2AAgentThread.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Text.Json;
+
+namespace Microsoft.Extensions.AI.Agents.A2A;
+
+/// <summary>
+/// Thread for A2A based agents.
+/// </summary>
+public class A2AAgentThread : ServiceIdAgentThread
```
SergeyMenshykh: Do we want to allow inheritance from the thread?

dotnet/samples/GettingStarted/Agents/Agent_Step13_Memory/Program.cs
```diff
@@ -63,26 +63,23 @@
...
-// It is also possible to add the memory component to an individual thread only instead of all
-// threads via the factory above.
-// In this case we will also use the same user info object, so this thread will share the same
-// memories as the previous thread.
-thread.AIContextProvider = new UserInfoMemory(chatClient.AsIChatClient(), userInfo);
+// It is also possible to set the memories in a memory component on an individual thread.
+// This is useful if we want to start a new thread, but have it share the same memories as a previous thread.
+// For this scenario, we have to know the underlying agent thread type and the memory component type.
+var newThread = agent.GetNewThread();
+((newThread as ChatClientAgentThread)!.AIContextProvider as UserInfoMemory)!.UserInfo = userInfo;
```
westey-m, line 78: Added a separate work item #837 to add GetService support for AIContextProviders in order to simplify this to just
`newThread.GetService<UserInfoMemory>()!.UserInfo`

dotnet/samples/SemanticKernelMigration/AzureAIFoundry/Step01_Basics/Program.cs
```diff
@@ -71,6 +71,6 @@ async Task AFAgentAsync()
     }
 
     // Clean up
-    await azureAgentClient.Threads.DeleteThreadAsync(thread.ConversationId);
+    await azureAgentClient.Threads.DeleteThreadAsync(((ChatClientAgentThread)thread).ConversationId);
```
westey-m: While having removed delete from the basic samples, I'm keeping it here since these are showing 1 to 1 migration guides and it is also what the migration agent is used as input.

dotnet/samples/GettingStarted/AgentOrchestration/AgentSample.cs
rogerbarreto: For the thread deletion, should instead we check if the thread is XYZ to delete it? I think would be less icky.

```csharp
if (thread is ChatClientAgentThread chatThread)
{
     await persistentAgentsClient.Threads.DeleteThreadAsync(chatThread.ConversationId, cancellationToken);
}
```

dotnet/samples/GettingStarted/Agents/Agent_Step13_Memory/Program.cs
rogerbarreto: No sure I follow.

`newThread.GetService()` would be in the thread, not in the AIContext right?

But I think having a GetService() in the thread could also go downstream in the AIContextProvider in the thread to look for it's GetService... so ultimately you could have.

`newThread.GetService<UserInfo>()`

This will search for the specific type across the thread service, if non is found it would go thru the GetService of the AIContextProvider (UserInfoMemory impl) that has a GetService implementation that knows about the `UserInfo`, right?

dotnet/samples/SemanticKernelMigration/AzureAIFoundry/Step01_Basics/Program.cs
rogerbarreto: I would suggest casting when you use the `GetNewThread` instead of here.

If that's not the case and you want to show usage of the abstract `AgentThread`, ideally as a good practice and avoid breaking code we should check if the thread is the expected type before direct casting.

```csharp
if (thread is ChatClientAgentThread chatThread)
{
     await persistentAgentsClient.Threads.DeleteThreadAsync(chatThread.ConversationId, cancellationToken);
}
```

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AgentThread.cs
```diff
@@ -14,167 +13,23 @@ namespace Microsoft.Extensions.AI.Agents;
...
-    /// <item>This thread object is new and a server managed thread has not yet been created in the agent service.</item>
-    /// </list>
-    /// </para>
-    /// <para>
-    /// The id may also change over time where the id is pointing at a
-    /// agent service managed thread, and the default behavior of a service is
-    /// to fork the thread with each iteration.
-    /// </para>
-    /// </remarks>
-    public string? ConversationId
```
rogerbarreto, line 88: I personally don't like the idea of removing this property from the abstraction, all the samples need to be casted to the specific thread this is not a good experience IMHO.

Possible alternatives:

1. keep it as a `virtual` property and make it nullable.

2. Have a abstract `Id` property without `Conversation`, when the thread has `Conversation` context, this property also directs to the `ConversationId` in the specialization.

    `override string? Id => ConversationId`

dotnet/samples/GettingStarted/Agents/Agent_Step13_Memory/Program.cs
westey-m: Yes, that's the idea.  Especially for AIContextProviders this all could go quite deep, since AIContextProviders could have a stack with aggregators.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AgentThread.cs
westey-m: > all the samples need to be casted to the specific thread

It's actually very few samples that need casting.  It's only for one agent type that uses one of two underlying services, one of which is already being deprecated.  And that's if you actually want to delete a thread, which I think is an edge case TBH.  If someone is deleting a thread they are also outside of the abstraction, since we don't support this functionality, and they already need to know what low level apis they are using.

Adding it to the base AgentThread is really exposing an implementation detail of some agents to users, but it is not applicable to all agents, so I really don't think it belongs on the base class.
It's also easy to move it up later if we change our minds, but harder to remove it from AgentThread, so I'd prefer going with what we have for now.

dotnet/src/Microsoft.Extensions.AI.Agents.A2A/A2AAgentThread.cs
westey-m: Good point. Sealed this and a few othres.

dotnet/samples/SemanticKernelMigration/AzureAIFoundry/Step01_Basics/Program.cs
westey-m: Sure, updated.

dotnet/samples/GettingStarted/AgentOrchestration/AgentSample.cs
westey-m: Sure, updated.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AgentThread.cs
```diff
@@ -14,167 +13,23 @@ namespace Microsoft.Extensions.AI.Agents;
 /// Base abstraction for all agent threads.
 /// A thread represents a specific conversation with an agent.
 /// </summary>
-public class AgentThread
+public abstract class AgentThread
```
markwallace-microsoft, line 16: Are we renaming this class?

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/InMemoryAgentThread.cs
```diff
@@ -0,0 +1,96 @@
...
+/// <summary>
+/// A base class for agent threads that operate entirely in memory without external storage.
+/// </summary>
+public abstract class InMemoryAgentThread : AgentThread
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="InMemoryAgentThread"/> class.
+    /// </summary>
+    /// <param name="messageStore">An optional <see cref="InMemoryChatMessageStore"/> to use for storing chat messages. If null, a new instance will be created.</param>
+    protected InMemoryAgentThread(InMemoryChatMessageStore? messageStore = null)
```
markwallace-microsoft, line 20: Can we make this more general purpose and take an `IChatMessageStore`?

So we end up distinguishing between `AgentThread`'s where the developer provides the chat message storage or ones where the associated service provides this.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/ServiceIdAgentThread.cs
```diff
@@ -0,0 +1,83 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Extensions.AI.Agents;
+
+/// <summary>
+/// A base class for agent threads that always store conversation state in the service, and only keep an ID reference in the <see cref="AgentThread"/>.
+/// </summary>
+public abstract class ServiceIdAgentThread : AgentThread
```
markwallace-microsoft, line 14: Not sure of the name. What happens if we want to associate other properties with this type of `AgentThread`. Would something like `ServiceManagedAgentThread` be a more accurate name?

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/InMemoryAgentThread.cs
```diff
@@ -0,0 +1,96 @@
...
+    /// <param name="serializedThreadState">A <see cref="JsonElement"/> representing the serialized state of the thread.</param>
+    /// <param name="jsonSerializerOptions">Optional settings for customizing the JSON deserialization process.</param>
+    /// <param name="messageStoreFactory">A factory function to create the <see cref="InMemoryChatMessageStore"/> from its serialized state.</param>
+    /// <exception cref="ArgumentException">The <paramref name="serializedThreadState"/> is not a JSON object.</exception>
+    /// <exception cref="JsonException">The <paramref name="serializedThreadState"/> is invalid or cannot be deserialized to the expected type.</exception>
+    protected InMemoryAgentThread(
+        JsonElement serializedThreadState,
+        JsonSerializerOptions? jsonSerializerOptions = null,
+        Func<JsonElement, JsonSerializerOptions?, InMemoryChatMessageStore>? messageStoreFactory = null)
+    {
```
markwallace-microsoft, line 50: Required the AgentThread knows how to create the memory store feels awkward

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AgentThreadMetadata.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+namespace Microsoft.Extensions.AI.Agents;
+
+/// <summary>Provides metadata about an <see cref="AgentThread"/>.</summary>
+public class AgentThreadMetadata
```
markwallace-microsoft, line 6: Should this be extensible i.e. have `AdditionalProperties`

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/InMemoryAgentThread.cs
```diff
@@ -0,0 +1,96 @@
...
+using System.Text.Json;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.AI.Agents;
+
+/// <summary>
+/// A base class for agent threads that operate entirely in memory without external storage.
+/// </summary>
+public abstract class InMemoryAgentThread : AgentThread
```
markwallace-microsoft, line 14: Does anything extend this yet?

dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatClientAgentThread.cs
```diff
@@ -0,0 +1,224 @@
...
+    /// </summary>
+    /// <param name="serializedThreadState">A <see cref="JsonElement"/> representing the serialized state of the thread.</param>
+    /// <param name="jsonSerializerOptions">Optional settings for customizing the JSON deserialization process.</param>
+    /// <param name="chatMessageStoreFactory">An optional factory function to create a custom <see cref="IChatMessageStore"/>.</param>
+    /// <param name="aiContextProviderFactory">An optional factory function to create a custom <see cref="AIContextProvider"/>.</param>
+    internal ChatClientAgentThread(
+        JsonElement serializedThreadState,
+        JsonSerializerOptions? jsonSerializerOptions = null,
+        Func<JsonElement, JsonSerializerOptions?, IChatMessageStore>? chatMessageStoreFactory = null,
+        Func<JsonElement, JsonSerializerOptions?, AIContextProvider>? aiContextProviderFactory = null)
```
markwallace-microsoft, line 39: Why not just pass in the context provider and memory store?

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/InMemoryAgentThread.cs
westey-m: Yes, the OrchestratingAgentThread uses this, as well as the sample for a custom agent.  Plus 5 unit tests with custom agents.
I expect this to be a common choice for custom agents to inherit from.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AgentThreadMetadata.cs
westey-m: Do you have anything to put in there right now?  We can always add it when we need it.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/InMemoryAgentThread.cs
westey-m: The alternative we had previously was to create the memory store elsewhere and then deserialize into an existing memory store, which IMO is worse, as it allows developers to accidentally overwrite state if used incorrectly.
Shout if you have a 3rd suggestion.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/ServiceIdAgentThread.cs
westey-m: I don't think we'd want to associate other properties with this `AgentThread`.  It is a base abstract class with a very specific purpose, so anyone who wants to have more properties, would have a differently named class that inherits from this.  Or if they want something else, they simply wouldn't inherit from this.

I'm definitely open to alternative names regardless, but I'd be a bit wary of using the prefix `ServiceManaged`.
 If we rename AgentThread to something more general purpose, it would sound like the whole thing is service managed and it's important that we don't create the impression that `AgentThread == chat history`, since chat history is typically a subset of AgentThread.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/AgentThread.cs
westey-m: Yes, we should just agree on an alternative name, and definitely not in this PR.  It's big enough already.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/InMemoryAgentThread.cs
westey-m: We could, but that would be a different class to this.  I think an InMemoryAgentThread is a common enough case to have a dedicated class for regardless.  We already have 7 subclasses of this one alone in the AF code base.  We currently only have one that supports a custom IChatMessageStore, in the form of ChatClientAgentThread.
Having only one based on `IChatMessageStore` would require more complex derived classes since they would all have to pass in the chosen messages store type and a factory for the message store.

dotnet/src/Microsoft.Extensions.AI.Agents/ChatCompletion/ChatClientAgentThread.cs
westey-m: Because then you would need a public deserialize method on those classes so that they can be deserialized into, which means developers can accidentally overwrite the state that is in one of these already if used incorrectly.
Remember that we had that before and it was really not a great pattern.

dotnet/src/Microsoft.Extensions.AI.Agents.Abstractions/ServiceIdAgentThread.cs
markwallace-microsoft: OK let's revisit these names once we made a decision on the new name for `AgentThread`

PR #881, opened 2025-09-23
.NET: Update Workflow Input/Output Redesign

lokitoth:
### Motivation and Context

Simplify using Workflows with multiple types of input and support multiple workflow output sources and types. This also removes .BuildWithOutput<> in favour of being able to declare any executor as outputting.

### Description

* Make id mandatory on Executor construction
* Enable running untyped Workflows and remove async-over-sync code
* Remove .BuildWithOutput and support yielding output from any Executor

### Contribution Checklist


dotnet/src/Microsoft.Agents.Workflows/Executor.cs
```diff
@@ -134,7 +158,7 @@ internal MessageRouter Router
     /// <summary>
     /// A set of <see cref="Type"/>s, representing the messages this executor can produce as output.
     /// </summary>
-    public virtual ISet<Type> OutputTypes { get; } = new HashSet<Type>([typeof(object)]);
+    public ISet<Type> OutputTypes { get; } = new HashSet<Type>([typeof(object)]);
```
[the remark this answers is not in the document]
lokitoth: This is pending the next of the three IO/Threading PRs, when we start properly supporting polymorphic message routing. For now the OutputTypes are hardcoded to be equivalent to "Any".

dotnet/samples/GettingStarted/Workflows/Agents/WorkflowAsAnAgent/WorkflowHelper.cs
```diff
@@ -29,7 +29,8 @@ internal static Workflow<List<ChatMessage>> GetWorkflow(IChatClient chatClient)
         return new WorkflowBuilder(startExecutor)
             .AddFanOutEdge(startExecutor, targets: [frenchAgent, englishAgent])
             .AddFanInEdge(aggregationExecutor, sources: [frenchAgent, englishAgent])
-            .Build<List<ChatMessage>>();
+            .WithOutputFrom(aggregationExecutor)
+            .BuildAsync<List<ChatMessage>>();
```
stephentoub, line 33: In what situation do I use `BuildAsync` vs `Build`? Does the former do something that the latter doesn't?

dotnet/src/Microsoft.Agents.Workflows/AgentWorkflowBuilder.cs
```diff
@@ -144,17 +146,20 @@ public static HandoffsWorkflowBuilder StartHandoffWith(AIAgent initialAgent)
...
     /// <summary>Executor that forwards all relevant messages.</summary>
-    private sealed class ForwardingExecutor : Executor
+    private sealed class ChatForwardingExecutor(string id) : Executor(id)
     {
         protected override RouteBuilder ConfigureRoutes(RouteBuilder routeBuilder) =>
-            routeBuilder.AddHandler<object>((message, context) => context.SendMessageAsync(message));
+            routeBuilder.AddHandler<string>((message, context) => context.SendMessageAsync(new ChatMessage(ChatRole.User, message)))
+                        .AddHandler<ChatMessage>((message, context) => context.SendMessageAsync(message))
+                        .AddHandler<List<ChatMessage>>((messages, context) => context.SendMessageAsync(messages))
+                        .AddHandler<TurnToken>((token, context) => context.SendMessageAsync(token));
```
stephentoub: For my edification, why is forwarding everything bad?

dotnet/src/Microsoft.Agents.Workflows/Execution/MessageRouter.cs
```diff
@@ -58,11 +72,19 @@ public bool CanHandle(TypeId candidateType)
...
+            if (this._typedHandlers.TryGetValue(message.GetType(), out MessageHandlerF? handler))
             {
                 result = await handler(message, context).ConfigureAwait(false);
             }
+            else if (this.HasCatchAll)
+            {
+                if (portableValue == null)
+                {
+                    portableValue = new PortableValue(message);
+                }
```
stephentoub: Nit:
```suggestion
                portableValue ??= new(message);
```

dotnet/src/Microsoft.Agents.Workflows/RouteBuilder.cs
```diff
@@ -178,5 +203,113 @@ async ValueTask<CallResult> WrappedHandlerAsync(object msg, IWorkflowContext ctx
         }
     }
 
-    internal MessageRouter Build() => new(this._typedHandlers);
+    private RouteBuilder AddCatchAll(CatchAllF handler, bool overwrite = false)
+    {
+        if (!overwrite && this._catchAll != null)
+        {
+            throw new ArgumentException("A catch-all is already registered (overwrite = false).");
```
stephentoub: Nit: should probably be InvalidOperationException, since this is more about the state of the builder than it is about a particular argument being bad.

dotnet/src/Microsoft.Agents.Workflows/StreamingAggregators.cs
```diff
@@ -32,34 +21,30 @@ public static class StreamingAggregators
...
 
-        TResult? Aggregate(TInput input, TResult? runningResult)
+        TResult? Aggregate(TResult? runningResult, TInput input)
         {
-            if (!hasRun)
+            if (runningResult == null)
             {
-                local = conversion(input);
-                hasRun = true;
+                runningResult = conversion(input);
```
stephentoub: Nit: `runningResult ??= conversion(input);`

dotnet/src/Microsoft.Agents.Workflows/WorkflowOutputEvent.cs
```diff
@@ -0,0 +1,51 @@
...
+    public bool IsType(Type type) => this.Data == null
+                                   ? false
+                                   : type.IsAssignableFrom(this.Data.GetType());
+
+    /// <summary>
+    /// Attempts to retrieve the underlying data as the specified type.
+    /// </summary>
+    /// <typeparam name="T">The type to which to cast.</typeparam>
+    /// <returns>The value of Data as to the target type.</returns>
+    public T? As<T>() => this.Data is T value ? value : default;
```
stephentoub, line 43: Why is this better than the consumer just using casts themselves?

And is there a situation where `Is<T>()` would return true but `As<T>` would return null, or vice versa?

dotnet/samples/GettingStarted/Workflows/Checkpoint/CheckpointAndRehydrate/WorkflowHelper.cs
```diff
@@ -26,7 +26,8 @@ internal static Workflow<NumberSignal> GetWorkflow()
         return new WorkflowBuilder(guessNumberExecutor)
             .AddEdge(guessNumberExecutor, judgeExecutor)
             .AddEdge(judgeExecutor, guessNumberExecutor)
-            .Build<NumberSignal>();
+            .WithOutputFrom(judgeExecutor)
+            .Build();
```
alliscode: Should this still be `Build<NumberSignal>()` or is there a specific reason this one should not have typed inputs?

dotnet/samples/GettingStarted/Workflows/Agents/WorkflowAsAnAgent/WorkflowHelper.cs
lokitoth: Sounds like this is a potential point of confusion - you are not the first person to ask it. All it does is Type-Check the input type of the first executor, and allow you some confidence that sending a message of that type works.

dotnet/src/Microsoft.Agents.Workflows/AgentWorkflowBuilder.cs
lokitoth: The issue here is if you are running on the flip side of a serialized checkpoint transition (checkpoint => serialize => rehydrate) you will not have real type information available and will truncate when you try to deserialize `object`.

That's why this change forces a catch-all handler to take a `PortableValue`.

The other benefit of explicitly supporting this specific set of input types is that we can use it to define a "concept" over Executors (the "Chat/AIAgent Protocol") and better formalize the `AsAgent()` hosting case.

dotnet/src/Microsoft.Agents.Workflows/Execution/MessageRouter.cs
lokitoth: One day I will remember that this operator exists. In the meanwhile, thanks for the reminder!

dotnet/src/Microsoft.Agents.Workflows/RouteBuilder.cs
lokitoth: Makes sense

dotnet/src/Microsoft.Agents.Workflows/WorkflowOutputEvent.cs
lokitoth: No, there is no such case (and we could conceivably add an `Is<T>(out T? value)` overload. What this does is abstracts out the machinery of serialization so the user (in 90% of the cases) does not have to think about it except when defining the types and setting up checkpointing.

dotnet/samples/GettingStarted/Workflows/Checkpoint/CheckpointAndRehydrate/WorkflowHelper.cs
lokitoth: Avoiding instantiating the default executor and forcing .BuildAsync<>(). We can also go back to the `RunSync(this Task)` approach.

Happy to unify the samples around .BuildAsync() for single-typed workflows like we discussed.

dotnet/src/Microsoft.Agents.Workflows/WorkflowOutputEvent.cs
stephentoub: Thanks.

I'm just not clear why it's better for the dev to write:
```csharp
ev.Is<Foo>()
```
instead of:
```csharp
ev.Data is Foo
```
or
```csharp
ev.As<Foo>()
```
instead of:
```csharp
ev.Data as Foo
```

What are we gaining by introducing another way to perform the casts and type checking?

dotnet/samples/GettingStarted/Workflows/Agents/WorkflowAsAnAgent/WorkflowHelper.cs
stephentoub: > All it does is Type-Check the input type of the first executor, and allow you some confidence that sending a message of that type works

Is it that the generic type parameter provides those benefits, and those benefits in turn need to use async functionality, hence why the method is then renamed to BuildAsync?

dotnet/samples/GettingStarted/Workflows/Checkpoint/CheckpointAndRehydrate/WorkflowHelper.cs
stephentoub: I'm much happier with an XxAsync than with doing sync-over-async. :)

dotnet/samples/GettingStarted/Workflows/Agents/WorkflowAsAnAgent/WorkflowHelper.cs
lokitoth: Yes, precisely. This lets us get rid of the async-in-sync pattern we have here right now (https://github.com/microsoft/agent-framework/pull/881/files/1e94cd810002beb2c84f8c5b0fe7fe4efbdd5a4d#diff-9af621d9f691e410756a5bf3e05083c0e94f666b0bb28f5a7d45cfb8aed23127L290-L294)

dotnet/src/Microsoft.Agents.Workflows/WorkflowOutputEvent.cs
lokitoth: The issue is that because the as/is operators are not overridable, we cannot transparently manage values that are backed by `IDelayedDeserialization`. See the implementation of .AsType(): https://github.com/microsoft/agent-framework/blob/main/dotnet/src/Microsoft.Agents.Workflows/PortableValue.cs#L153-L165

dotnet/samples/GettingStarted/Workflows/Checkpoint/CheckpointAndRehydrate/WorkflowHelper.cs
lokitoth: It does seem like a better way of doing it

dotnet/src/Microsoft.Agents.Workflows/WorkflowOutputEvent.cs
stephentoub: Ok, but how does that relate WorkflowOutputEvent? Are you just trying to establish a pattern across a bunch of types?

lokitoth: I am kind of front-running our eventual need to support Distributed execution, which would introduce a potential serialization/deserialization across the yield / consume boundary (not from checkpoint, but from collecting the event back from a remote executor to be yielded out of the execution environment).

If we just rely on as/is, we run the risk of breaking existing code if we end up needing to introduce the indirection. I am happy to remove these (though in a separate PR to avoid blocking this one as a couple of downstream things are waiting on it)

PR #882, opened 2025-09-23
.Net: Visualizing dotnet workflows

alliscode:
### Description

Adding the ability to visualize Workflows in dotnet.

```csharp
Workflow workflow = BuildWorkflow();

// Generate Mermaid format
string mermaid = workflow.ToMermaidString();

// Generate DOT format
string dotString = workflow.ToDotString();
```

#### DOT

graphviz_render

#### Mermaid

mermaid_render

### Contribution Checklist

### DIFF
--- a/dotnet/agent-framework-dotnet.slnx
+++ b/dotnet/agent-framework-dotnet.slnx
@@ -50,6 +50,7 @@
     <Project Path="samples/GettingStarted/Agents/Agent_Step12_AsFunctionTool/Agent_Step12_AsFunctionTool.csproj" />
     <Project Path="samples/GettingStarted/Agents/Agent_Step13_Memory/Agent_Step13_Memory.csproj" />
     <Project Path="samples/GettingStarted/Agents/Agent_Step14_Middleware/Agent_Step14_Middleware.csproj" />
+    <Project Path="samples/GettingStarted/Agents/Agent_Step15_Plugins/Agent_Step15_Plugins.csproj" />
   </Folder>
   <Folder Name="/Samples/GettingStarted/AgentWithOpenAI/">
     <File Path="samples/GettingStarted/AgentWithOpenAI/README.md" />
--- a/dotnet/nuget/nuget-package.props
+++ b/dotnet/nuget/nuget-package.props
@@ -2,8 +2,8 @@
   <PropertyGroup>
     <!-- Central version prefix - applies to all nuget packages. -->
     <VersionPrefix>1.0.0</VersionPrefix>
-    <PackageVersion Condition="'$(VersionSuffix)' != ''">$(VersionPrefix)-$(VersionSuffix).250929</PackageVersion>
-    <PackageVersion Condition="'$(VersionSuffix)' == ''">$(VersionPrefix)</PackageVersion>
+    <PackageVersion Condition="'$(VersionSuffix)' != ''">$(VersionPrefix)-$(VersionSuffix).250930</PackageVersion>
+    <PackageVersion Condition="'$(VersionSuffix)' == ''">$(VersionPrefix)-preview.250930</PackageVersion>
 
     <Configurations>Debug;Release;Publish</Configurations>
     <IsPackable>true</IsPackable>
--- a/dotnet/samples/GettingStarted/Agents/Agent_Step15_Plugins/Agent_Step15_Plugins.csproj
+++ b/dotnet/samples/GettingStarted/Agents/Agent_Step15_Plugins/Agent_Step15_Plugins.csproj
@@ -3,10 +3,11 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>net9.0</TargetFramework>
-    
+
     <Nullable>enable</Nullable>
     <ImplicitUsings>disable</ImplicitUsings>
-	<NoWarn>VSTHRD200;CA1707</NoWarn>
+    <NoWarn>VSTHRD200;CA1707;CA1812</NoWarn>
+    <RootNamespace>Agent_Step15_Plugins</RootNamespace>
   </PropertyGroup>
 
   <ItemGroup>
@@ -17,7 +18,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\..\..\src\Microsoft.Agents.AI.OpenAI\Microsoft.Agents.AI.OpenAI.csproj" />
-	  <ProjectReference Include="..\..\..\..\src\Microsoft.Agents.AI\Microsoft.Agents.AI.csproj" />
+    <ProjectReference Include="..\..\..\..\src\Microsoft.Agents.AI\Microsoft.Agents.AI.csproj" />
   </ItemGroup>
-
+  
 </Project>
--- a/dotnet/samples/GettingStarted/Agents/Agent_Step15_Plugins/Program.cs
+++ b/dotnet/samples/GettingStarted/Agents/Agent_Step15_Plugins/Program.cs
@@ -1,0 +1,135 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+// This sample shows how to use plugins with an AI agent. Plugin classes can
+// depend on other services that need to be injected. In this sample, the
+// AgentPlugin class uses the WeatherProvider and CurrentTimeProvider classes
+// to get weather and current time information. Both services are registered
+// in the service collection and injected into the plugin.
+// Plugin classes may have many methods, but only some are intended to be used
+// as AI functions. The AsAITools method of the plugin class shows how to specify
+// which methods should be exposed to the AI agent.
+
+using System;
+using System.Collections.Generic;
+using Azure.AI.OpenAI;
+using Azure.Identity;
+using Microsoft.Agents.AI;
+using Microsoft.Extensions.AI;
+using Microsoft.Extensions.DependencyInjection;
+using OpenAI;
+
+var endpoint = Environment.GetEnvironmentVariable("AZURE_OPENAI_ENDPOINT") ?? throw new InvalidOperationException("AZURE_OPENAI_ENDPOINT is not set.");
+var deploymentName = Environment.GetEnvironmentVariable("AZURE_OPENAI_DEPLOYMENT_NAME") ?? "gpt-4o-mini";
+
+// Create a service collection to hold the agent plugin and its dependencies.
+ServiceCollection services = new();
+services.AddSingleton<WeatherProvider>();
+services.AddSingleton<CurrentTimeProvider>();
+services.AddSingleton<AgentPlugin>(); // The plugin depends on WeatherProvider and CurrentTimeProvider registered above.
+
+IServiceProvider serviceProvider = services.BuildServiceProvider();
+
+const string AgentName = "Assistant";
+const string AgentInstructions = "You are a helpful assistant that helps people find information.";
+
+AIAgent agent = new AzureOpenAIClient(
+    new Uri(endpoint),
+    new AzureCliCredential())
+     .GetChatClient(deploymentName)
+     .CreateAIAgent(
+        instructions: AgentInstructions,
+        name: AgentName,
+        tools: [.. serviceProvider.GetRequiredService<AgentPlugin>().AsAITools()],
+        services: serviceProvider); // Pass the service provider to the agent so it will be available to plugin functions to resolve dependencies.
+
+Console.WriteLine(await agent.RunAsync("Tell me current time and weather in Seattle."));
+
+/// <summary>
+/// The agent plugin that provides weather and current time information.
+/// </summary>
+/// <param name="weatherProvider">The weather provider to get weather information.</param>
+internal sealed class AgentPlugin(WeatherProvider weatherProvider)
+{
+    /// <summary>
+    /// Gets the weather information for the specified location.
+    /// </summary>
+    /// <remarks>
+    /// This method demonstrates how to use the dependency that was injected into the plugin class.
+    /// </remarks>
+    /// <param name="location">The location to get the weather for.</param>
+    /// <returns>The weather information for the specified location.</returns>
+    public string GetWeather(string location)
+    {
+        return weatherProvider.GetWeather(location);
+    }
+
+    /// <summary>
+    /// Gets the current date and time for the specified location.
+    /// </summary>
+    /// <remarks>
+    /// This method demonstrates how to resolve a dependency using the service provider passed to the method.
+    /// </remarks>
+    /// <param name="sp">The service provider to resolve the <see cref="CurrentTimeProvider"/>.</param>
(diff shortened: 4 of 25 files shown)


dotnet/src/Microsoft.Agents.Workflows/Visualization/WorkflowViz.cs
```diff
@@ -0,0 +1,275 @@
...
+#else
+        var hash = SHA256.HashData(Encoding.UTF8.GetBytes(input));
+        return Convert.ToHexString(hash).Substring(0, 8).ToUpperInvariant();
+#endif
+    }
+
+    private bool TryGetNestedWorkflow(ExecutorRegistration registration, out Workflow workflow)
+    {
+        // TODO: Implement proper nested workflow detection once nested workflow support is in. (https://github.com/microsoft/agent-framework/issues/880)
+        workflow = null!;
```
lokitoth: Is the plan to hold off on checking this in until SubWorkflow is supported?

dotnet/src/Microsoft.Agents.Workflows/Visualization/WorkflowViz.cs
```diff
@@ -0,0 +1,275 @@
...
+        return string.Join("\n", lines);
+    }
+
+    /// <summary>
+    /// Export the workflow visualization to a file.
+    /// </summary>
+    /// <param name="format">The output format (svg, png, pdf, dot).</param>
+    /// <param name="filename">Optional filename to save the output. If null, creates a temporary file.</param>
+    /// <returns>The path to the saved file.</returns>
+    public async Task<string> ExportAsync(string format = "svg", string? filename = null)
```
lokitoth: Why do we create an Enum for formats for the `DotRenderer`, but use a `string` here?

dotnet/src/Microsoft.Agents.Workflows/Visualization/DotRenderer.cs
```diff
@@ -0,0 +1,175 @@
...
+
+namespace Microsoft.Agents.Workflows;
+
+/// <summary>
+/// Provides methods for rendering Graphviz DOT source code into various output formats.
+/// </summary>
+/// <remarks>This class serves as a wrapper for the Graphviz "dot" executable, enabling the conversion of DOT
+/// source code into formats such as SVG, PNG, and PDF. The "dot" executable must be installed and accessible via the
+/// system PATH. For more information about Graphviz, visit <see href="https://graphviz.org"/>.</remarks>
+public static class DotRenderer
```
lokitoth: [Minor] Do we want to keep this public?

dotnet/src/Microsoft.Agents.Workflows/Visualization/DotRenderer.cs
```diff
@@ -0,0 +1,175 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Agents.Workflows;
+
+/// <summary>
+/// Provides methods for rendering Graphviz DOT source code into various output formats.
+/// </summary>
+/// <remarks>This class serves as a wrapper for the Graphviz "dot" executable, enabling the conversion of DOT
+/// source code into formats such as SVG, PNG, and PDF. The "dot" executable must be installed and accessible via the
```
lokitoth: If we do keep this class public (or if not, we can create a static singleton GraphVizOptions to configure), we should consider letting the user specify a location for the executable (without changing PATH)

dotnet/samples/GettingStarted/Workflows/Visualization/Program.cs
```diff
@@ -0,0 +1,250 @@
...
+            {
+                Console.WriteLine("PNG export requires Graphviz installation.");
+            }
+        }
+        catch (Exception ex)
+        {
+            Console.WriteLine($"Visualization export failed: {ex.Message}");
+        }
+
+        // Execute the workflow
```
TaoChenOSU: nit:
We probably don't need to execute the workflow given that the emphasis of this sample is visualization.

dotnet/samples/GettingStarted/Workflows/Concurrent/MapReduce/MapReduce.csproj
TaoChenOSU, line 1: The convention for the name of the project file is the name of the folder that contains it, so this should be `Visualization.csproj`.

dotnet/src/Microsoft.Agents.Workflows/Visualization/DotRenderer.cs
```diff
@@ -0,0 +1,175 @@
...
+        using var proc = new Process { StartInfo = psi };
+
+        try
+        {
+            if (!proc.Start())
+            {
+                throw new InvalidOperationException("Failed to start Graphviz 'dot' process.");
+            }
+        }
+        catch (System.ComponentModel.Win32Exception ex) when (ex.NativeErrorCode == 2) // file not found
```
stephentoub: Is this the right error code regardless of operating system?

dotnet/samples/GettingStarted/Workflows/Visualization/Program.cs
```diff
@@ -0,0 +1,250 @@
...
+        );
+
+        // Create workflow orchestration executors
+        var dispatcher = new DispatchToExpertsExecutor(["Researcher", "Marketer", "Legal"]);
+        var aggregator = new AggregateInsightsExecutor(["Researcher", "Marketer", "Legal"]);
+
+        // Build the concurrent workflow with fan-out/fan-in pattern
+        var workflow = new WorkflowBuilder(dispatcher)
+            .AddFanOutEdge(dispatcher, targets: [researcher, marketer, legal])
+            .AddFanInEdge(aggregator, sources: [researcher, marketer, legal])
```
stephentoub: Not specific to this PR, but I find the ordering of these parameters confusing. I'd expect it to always go left to right in terms of the data flow, so it'd be:
```csharp
.AddFanInEdge(sources: [researcher, marketer, legal], aggregator)
```
That means sources could no longer be params, but with collection expressions (which are used here), params only saves the caller two characters, the `[` and `]`. I suggest swapping the ordering.

dotnet/samples/GettingStarted/Workflows/Visualization/Program.cs
```diff
@@ -0,0 +1,250 @@
...
+            .Build<string>();
+
+        // Generate and display workflow visualization
+        Console.WriteLine("=== WORKFLOW VISUALIZATION ===");
+        var viz = new WorkflowViz(workflow);
```
stephentoub: Nit: it'd be better to not use "Viz" and instead use Visualizer or something spelled out more.

dotnet/samples/GettingStarted/Workflows/Visualization/Program.cs
```diff
@@ -0,0 +1,250 @@
...
+    /// <param name="context">Workflow context for emitting completion events</param>
+    public async ValueTask HandleAsync(ChatMessage message, IWorkflowContext context)
+    {
+        var expertName = message.AuthorName ?? "Unknown";
+        this._responsesByExpert[expertName] = message.Text ?? "";
+
+        Console.WriteLine($"Received response from {expertName}");
+
+        // Check if we have responses from all expected experts
+        if (this._responsesByExpert.Count == this._expertIds.Length)
```
stephentoub: What happens if the workflow is executed multiple times?

dotnet/src/Microsoft.Agents.Workflows/Visualization/DotRenderer.cs
```diff
@@ -0,0 +1,175 @@
...
+    /// <param name="timeoutMs">The maximum time, in milliseconds, to wait for the rendering process to complete. Defaults to 30,000 ms.</param>
+    /// <param name="cancellationToken">A token to monitor for cancellation requests. The operation will be canceled if the token is triggered.</param>
+    /// <returns>A byte array containing the rendered graph in the specified format.</returns>
+    /// <exception cref="ArgumentException">Thrown if <paramref name="dotSource"/> is null, empty, or consists only of whitespace.</exception>
+    /// <exception cref="ArgumentOutOfRangeException">Thrown if <paramref name="format"/> is not a valid <see cref="OutputFormat"/> value.</exception>
+    /// <exception cref="InvalidOperationException">Thrown if the "dot" process fails to start, the "dot" executable is not found on the system PATH, or the process
+    /// exits with a non-zero exit code.</exception>
+    /// <exception cref="TimeoutException">Thrown if the rendering process does not complete within the specified <paramref name="timeoutMs"/>.</exception>
+    public static async Task<byte[]> RenderAsync(
```
stephentoub: I find it a bit strange that this component would have a logical dependency on a separately installed and unrelated executable. Are there any security concerns? Seems like this is no less of a dependency than if this component were actually depending on the other via a package reference; does component governance come into play?

dotnet/src/Microsoft.Agents.AI.Workflows/Visualization/WorkflowVisualizer.cs
```diff
@@ -0,0 +1,275 @@
...
+    /// Initializes a new instance of the <see cref="WorkflowViz"/> class.
+    /// </summary>
+    /// <param name="workflow">The workflow to visualize.</param>
+    public WorkflowViz(Workflow workflow)
+    {
+        this._workflow = workflow ?? throw new ArgumentNullException(nameof(workflow));
+    }
+
+    /// <summary>
+    /// Export the workflow as a DOT format digraph string.
```
stephentoub, line 19: VS/VS Code have built-in mermaid renderers... could that be an option as well?

dotnet/src/Microsoft.Agents.Workflows/Visualization/WorkflowViz.cs
```diff
@@ -0,0 +1,275 @@
...
+    {
+        if (string.IsNullOrEmpty(format))
+        {
+            throw new ArgumentException("Format cannot be null or empty.", nameof(format));
+        }
+
+        var supportedFormats = new[] { "SVG", "PNG", "PDF", "DOT" };
+        var upperFormat = format.ToUpperInvariant();
+        if (!supportedFormats.Contains(upperFormat))
```
stephentoub: This would be a bit better as:
```csharp
private static readonly HashSet<string> s_supportedFormats = new(StringComparison.OrdinalIgnoreCase) { "svg", "png", "pdf", "dot" };
```
and then this here can just be:
```csharp
if (!s_supportedFormats.Contains(format))
```

dotnet/src/Microsoft.Agents.Workflows/Visualization/WorkflowViz.cs
```diff
@@ -0,0 +1,275 @@
...
+
+        if (format.Equals("dot", StringComparison.OrdinalIgnoreCase))
+        {
+            var dotFilename = filename ?? Path.GetTempFileName() + ".dot";
+            File.WriteAllText(dotFilename, dotContent, Encoding.UTF8);
+            return dotFilename;
+        }
+
+        // Use DotRenderer wrapper to render the graph. NOTE: GraphViz must be installed and available on the system.
+        var outputFilename = filename ?? Path.GetTempFileName() + $".{format}";
```
stephentoub:
```suggestion
        var outputFilename = filename ?? $"{Path.GetTempFileName()}.{format}";
```

dotnet/src/Microsoft.Agents.Workflows/Visualization/WorkflowViz.cs
```diff
@@ -0,0 +1,275 @@
...
+    private string ComputeFanInDigest(string target, List<string> sources)
+    {
+        var sortedSources = sources.OrderBy(x => x, StringComparer.Ordinal).ToList();
+        var input = target + "|" + string.Join("|", sortedSources);
+        return this.ComputeShortHash(input);
+    }
+
+    private string ComputeShortHash(string input)
+    {
+#if NET472
```
stephentoub:
```suggestion
#if !NET
```

dotnet/src/Microsoft.Agents.AI.Workflows/Visualization/WorkflowVisualizer.cs
```diff
@@ -0,0 +1,275 @@
...
+
+    private string ComputeShortHash(string input)
+    {
+#if NET472
+        using var sha256 = SHA256.Create();
+        var hash = sha256.ComputeHash(Encoding.UTF8.GetBytes(input));
+        return BitConverter.ToString(hash).Replace("-", "").Substring(0, 8).ToUpperInvariant();
+#else
+        var hash = SHA256.HashData(Encoding.UTF8.GetBytes(input));
+        return Convert.ToHexString(hash).Substring(0, 8).ToUpperInvariant();
```
stephentoub: Convert.ToHexString already produces upper case... you don't need the ToUpperInvariant

dotnet/samples/GettingStarted/Workflows/Visualization/Program.cs
lokitoth: This make sense. Made issue #926 for it.

dotnet/src/Microsoft.Agents.Workflows/Visualization/DotRenderer.cs
alliscode: Removed

dotnet/samples/GettingStarted/Workflows/Visualization/Program.cs
```diff
@@ -0,0 +1,421 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text.Json;
+using System.Threading.Tasks;
+using Microsoft.Agents.AI.Workflows;
+using Microsoft.Agents.AI.Workflows.Reflection;
+
+namespace WorkflowMapReduceSample;
```
alliscode: Fixing issues with file names..

dotnet/samples/GettingStarted/Workflows/Visualization/Program.cs
```diff
@@ -0,0 +1,42 @@
...
+internal static class Program
+{
+    /// <summary>
+    /// Entry point that generates and displays workflow visualizations in Mermaid and DOT formats.
+    /// </summary>
+    /// <param name="args">Command line arguments (not used).</param>
+    private static void Main(string[] args)
+    {
+        // Step 1: Build the workflow you want to visualize
+        Workflow workflow = WorkflowMapReduceSample.Program.BuildWorkflow();
```
TaoChenOSU, line 25: nit: I am not sure if we should do this because this will defeat the purpose of splitting the samples into individual projects.

alliscode: Will update

Issue #872, opened 2025-09-23
.NET: Update samples and documentation to use the term Observability rather than Telemetry


markwallace-microsoft:
Aligning terminology to match what is used in Foundry (see: https://learn.microsoft.com/en-us/azure/ai-foundry/how-to/develop/trace-agents-sdk)

PR #889, opened 2025-09-24
.NET: Use observability term

SergeyMenshykh:
Context: https://github.com/microsoft/agent-framework/issues/872


dotnet/agent-framework-dotnet.slnx
```diff
@@ -61,7 +61,7 @@
     <Project Path="samples/GettingStarted/ModelContextProtocol/Agent_MCP_Server/Agent_MCP_Server.csproj" />
     <Project Path="samples/GettingStarted/ModelContextProtocol/Agent_MCP_Server_Auth/Agent_MCP_Server_Auth.csproj" />
   </Folder>
-  <Folder Name="/Samples/GettingStarted/Telemetry/">
+  <Folder Name="/Samples/GettingStarted/Observability/">
     <Project Path="samples/GettingStarted/AgentOpenTelemetry/AgentOpenTelemetry.csproj" />
```
westey-m, line 65: Should we rename this project as well?

SergeyMenshykh: OpenTelemetry is one way to enable observability in AF. AF may add support for other observability platforms, so it would make sense to have a sample project for each alternative.

PR #914, opened 2025-09-25
.NET: Adding a few small sample improvements

westey-m:
Adding a few small sample improvements

### Contribution Checklist


dotnet/samples/GettingStarted/Agents/Agent_Step10_AsMcpTool/AIAgentExtensions.cs
```diff
@@ -23,16 +20,7 @@ internal static class AIAgentExtensions
...
-            [Description("Available information that will guide in performing this operation.")] string query,
-            CancellationToken cancellationToken = default)
-        {
-            AgentRunResponse response = await agent.RunAsync(query, cancellationToken: cancellationToken);
-
-            return response.ToString();
-        }
-
-        return McpServerTool.Create(RunAgentAsync, new McpServerToolCreateOptions()
+        return McpServerTool.Create(agent.AsAIFunction(), new McpServerToolCreateOptions()
```
stephentoub: I don't think the `?? agent.Name` and `?? agent.Description` below are necessary. AsAIFunction should be defaulting to those, right?

westey-m: Makes sense.  I've removed it for Name and Description.  Still need it for Title since title is not set from anything on AIFunction by default.

stephentoub: > Still need it for Title since title is not set from anything on AIFunction by default.

But if the caller of this method hasn't supplied a title, presumably it should just do what McpServerTool.Create does when no title is supplied, rather than forcibly duplicating the agent name at this layer?

westey-m: Yeah, I want to to get input from @SergeyMenshykh, and see if we really need to default title to name if title is not provided explicitly.  It doesn't seem that valuable considering that name will already be set.

stephentoub: From an MCP perspective, title is optional, and clients use name when title isn't provided. Title is intended to be the human-consumable description of the tool. As such, I don't think this should explicitly fall back to setting title to name.

westey-m: Thanks for the info.  I've removed this extension method completely, since I don't think it adds that much value.

PR #915, opened 2025-09-25
.NET: rename A2A extension methods

DeagleGross:
AspNetCore follows up the convention where extension methods describing the newly added set of endpoints are called `Map...`. I've mistakenly called them `AttachA2A` for A2A integration, so I am renaming them here.

Also discussed with Reuben to simplify A2A-wrapper usage to `AgentProxy`, which I also included in the PR. Tested locally and communication works for A2A


dotnet/src/Microsoft.Extensions.AI.Agents.Hosting.A2A/Microsoft.Extensions.AI.Agents.Hosting.A2A.csproj
```diff
@@ -8,7 +8,11 @@
     <VersionSuffix>alpha</VersionSuffix>
   </PropertyGroup>
 
-  <ItemGroup>
+  <PropertyGroup>
+    <InjectSharedThrow>true</InjectSharedThrow>
+  </PropertyGroup>
+
+    <ItemGroup>
```
adityamandaleeka: nit: indentation

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/Internal/A2AAgentWrapper.cs
```diff
@@ -16,51 +16,26 @@ namespace Microsoft.Extensions.AI.Agents.Hosting.A2A.Internal;
...
     public A2AAgentWrapper(
         IActorClient actorClient,
         AIAgent innerAgent,
         ILoggerFactory? loggerFactory = null)
     {
-        this._actorClient = actorClient;
-        this._innerAgent = innerAgent ?? throw new ArgumentNullException(nameof(innerAgent));
+        Throw.IfNullOrEmpty(innerAgent.Name);
+
+        this._agentProxy = new AgentProxy(innerAgent.Name, actorClient);
```
lokitoth, line 28: Does this mean this can only work if the agents are running on top of Actor Runtime?

DeagleGross: Yep, that's why the package includes `Hosting` in the name - it clearly states the runtime is the "dependency" here. Non-runtime package also exists in the repo and serves a bit different purpose (its an implementation of the `AIAgent` instead of the wrap): https://github.com/microsoft/agent-framework/tree/main/dotnet/src/Microsoft.Agents.AI.A2A

dotnet/src/Microsoft.Extensions.AI.Agents.Hosting.A2A/Microsoft.Extensions.AI.Agents.Hosting.A2A.csproj
DeagleGross:
```suggestion
  <ItemGroup>
```

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/Internal/A2AAgentWrapper.cs
ReubenBond: I have a branch where I've been refactoring this so it does not need the actor runtime. I'll rebase it on this commit once we get it merged

Issue #784, opened 2025-09-16
.NET: What are the threading guarantees around Workflows and Executors?


stephentoub:
1. Is a Workflow intended to be used concurrently, i.e. multiple StreamingRuns on the same instance all in flight at the same time?
2. Is an Executor intended to support concurrent invocation of its CanHandle/Handle methods?

Presumably the answer to (1) is "no", or else an implementation like this:
https://github.com/microsoft/agent-framework/blob/b2d201987428975c321b9cd223e5bdb77589434e/dotnet/src/Microsoft.Agents.Workflows/Specialized/AIAgentHostExecutor.cs#L12-L17
is broken, because that single Executor could end up being used for two unrelated parallel executions.

And if (2) is "yes", then samples like:
https://github.com/microsoft/agent-framework/blob/b2d201987428975c321b9cd223e5bdb77589434e/dotnet/samples/GettingStarted/Workflows/Concurrent/Program.cs#L106-L116
seem broken, as multiple messages could arrive concurrently and lead to non-thread-safe `List<T>.Add` calls. Or is there serialization of work being performed somewhere to ensure that a given executor only ever handles one message at a time?

lokitoth:
We had a good design discussion around here and came up with three possible models:

1. No Concurrency

This was the original model we were implementing, which would see the Task.WhenAll() in the message delivery stage replaced with a loop, awaiting in between. It is easy to switch back to, and fairly understandable, but from a throughput standpoint would be less desirable.

Running a single Workflow instance concurrently will no longer be allowed in this case.

2. Concurrent Executors, Single-threaded-Execution

Unlike right now, where we create a Task-per-Message, we would collate the tasks on a per-Executor basis, then have a single task per Executor. This would provide more throughput than (1), while still giving simple 

Running a single Workflow instance concurrently will no longer be allowed in this case.

3. Concurrent everything

The current model, updated to fix the internal conception of when things can run simultaneously, and ensure appropriate mutual exclusion. We believe this would be more difficult to work with particularly for those just getting started building workflows, and at minimum would require significant attention to be paid to this in the documentation.

Running a single Workflow instance concurrently will be allowed in this case. The runtime will be updated to ensure that internal data structures and communication patterns are appropriately safe when running in this mode before we enable it.

---

The current suggested plan:

We go with (2) by default and potentially open up a switch to let users change to the others (3, and maybe 1, for debug purposes) if they promise, they know what they are doing, e.g.:

```csharp
[...]

PR #925, opened 2025-09-25
.NET: Solidify Concurrency Model for Inproc Workflows

lokitoth:
### Motivation and Context

Our initial concurrency model for Workflows was single-threaded async execution (similar to Python). This meant that messages were delivered one at a time to individual executors at a time. It seemed desirable to allow concurrent execution, so we switched to delivering messages concurrently. This causes threading and timing issues, as well as leading to message deliveries being out of order.

The desired fix is detailed in #784

### Description

These changes introduce the new default concurrency model: All executors processing messages run in parallel, but one message at a time. To avoid issues with executors provided as instances -- and thus shared between runs on the same workflow instance, rather than created every run if a factory method is provided -- we disallow running a single workflow instance concurrently. 

After a run is done and disposed, if the Workflow has no shared executor instances or all of the are resettable, it may be reused in another run.

Adding support for fully single-threaded or concurrent execution is not in scope of this change.

* Disallow execution of multiple workflows at once
* Define notion of executor Reset() to allow reuse of workflows with shared executor instances
* Switch to delivering all messages to a single executor sequentially, with executors running in parallel

Technically this is a breaking change because previously it was possible to run the same workflow instance concurrently (but with undefined behaviour). Now this is no longer allowed. 

### Contribution Checklist

### DIFF
--- a/dotnet/src/Microsoft.Agents.Workflows/Execution/DeliveryMapping.cs
+++ b/dotnet/src/Microsoft.Agents.Workflows/Execution/DeliveryMapping.cs
@@ -1,0 +1,35 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Agents.Workflows.Execution;
+
+internal sealed class DeliveryMapping
+{
+    private readonly IEnumerable<MessageEnvelope> _envelopes;
+    private readonly IEnumerable<Executor> _targets;
+
+    public DeliveryMapping(IEnumerable<MessageEnvelope> envelopes, IEnumerable<Executor> targets)
+    {
+        this._envelopes = Throw.IfNull(envelopes);
+        this._targets = Throw.IfNull(targets);
+    }
+
+    public DeliveryMapping(MessageEnvelope envelope, Executor target) : this([envelope], [target]) { }
+    public DeliveryMapping(MessageEnvelope envelope, IEnumerable<Executor> targets) : this([envelope], targets) { }
+    public DeliveryMapping(IEnumerable<MessageEnvelope> envelopes, Executor target) : this(envelopes, [target]) { }
+
+    public IEnumerable<MessageDelivery> Deliveries => from target in this._targets
+                                                      from envelope in this._envelopes
+                                                      select new MessageDelivery(envelope, target);
+
+    public void MapInto(StepContext nextStep)
+    {
+        foreach (Executor target in this._targets)
+        {
+            nextStep.MessagesFor(target.Id).AddRange(this._envelopes.Select(envelope => envelope));
+        }
+    }
+}
--- a/dotnet/src/Microsoft.Agents.Workflows/Execution/EdgeMap.cs
+++ b/dotnet/src/Microsoft.Agents.Workflows/Execution/EdgeMap.cs
@@ -10,11 +10,22 @@ namespace Microsoft.Agents.Workflows.Execution;
 
 internal sealed class EdgeMap
 {
-    private readonly Dictionary<EdgeId, object> _edgeRunners = [];
-    private readonly Dictionary<EdgeId, FanInEdgeState> _fanInState = [];
+    private readonly Dictionary<EdgeId, EdgeRunner> _edgeRunners = [];
+    private readonly Dictionary<EdgeId, IStatefulEdgeRunner> _statefulRunners = new();
     private readonly Dictionary<string, InputEdgeRunner> _portEdgeRunners;
+
     private readonly InputEdgeRunner _inputRunner;
     private readonly IStepTracer? _stepTracer;
+
+    public EdgeMap(IRunnerContext runContext,
+                   Workflow workflow,
+                   IStepTracer? stepTracer)
+        : this(runContext,
+               workflow.Edges,
+               workflow.Ports.Values,
+               workflow.StartExecutorId,
+               stepTracer)
+    { }
 
     public EdgeMap(IRunnerContext runContext,
                    Dictionary<string, HashSet<Edge>> workflowEdges,
@@ -24,7 +35,7 @@ internal sealed class EdgeMap
     {
         foreach (Edge edge in workflowEdges.Values.SelectMany(e => e))
         {
-            object edgeRunner = edge.Kind switch
+            EdgeRunner edgeRunner = edge.Kind switch
             {
                 EdgeKind.Direct => new DirectEdgeRunner(runContext, edge.DirectEdgeData!),
                 EdgeKind.FanOut => new FanOutEdgeRunner(runContext, edge.FanOutEdgeData!),
@@ -32,12 +43,12 @@ internal sealed class EdgeMap
                 _ => throw new NotSupportedException($"Unsupported edge type: {edge.Kind}")
             };
 
-            if (edgeRunner is FanInEdgeRunner fanInRunner)
-            {
-                this._fanInState[edge.Data.Id] = fanInRunner.CreateState();
-            }
-
             this._edgeRunners[edge.Data.Id] = edgeRunner;
+
+            if (edgeRunner is IStatefulEdgeRunner statefulRunner)
+            {
+                this._statefulRunners[edge.Data.Id] = statefulRunner;
+            }
         }
 
         this._portEdgeRunners = workflowPorts.ToDictionary(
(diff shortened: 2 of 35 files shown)


dotnet/src/Microsoft.Agents.Workflows/Execution/DeliveryMapping.cs
```diff
@@ -0,0 +1,35 @@
...
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Agents.Workflows.Execution;
+
+internal sealed class DeliveryMapping
+{
+    private readonly IEnumerable<MessageEnvelope> _envelopes = [];
```
crickman: Is default initialization necessary? (I'm surprised its not a warning @stephentoub)

dotnet/src/Microsoft.Agents.Workflows/Run.cs
```diff
@@ -38,7 +39,7 @@ public enum RunStatus
 /// Represents a workflow run that tracks execution status and emitted workflow events, supporting resumption
 /// with responses to <see cref="RequestInfoEvent"/>.
 /// </summary>
-public class Run
+public sealed class Run : IAsyncDisposable
```
crickman: What about all the samples that aren't disposing the run?

dotnet/src/Microsoft.Agents.Workflows/Execution/DeliveryMapping.cs
stephentoub: Default would be null. This is an empty enumerable.

dotnet/src/Microsoft.Agents.Workflows/Run.cs
lokitoth: If you only intend to use a single Workflow instance once, you do not need to dispose the Run. I'm of two minds about this - I get wanting to show cleanliness, but I am also not a fan of forcing the user to wait X amount of time during shutdown to do superfluous cleanup.

dotnet/src/Microsoft.Agents.Workflows/Execution/DeliveryMapping.cs
lokitoth: Nope, unneeded. Good catch.

stephentoub:
> if the Workflow has no shared executor instances

I'm not understanding this comment. Does "shared executor" mean it's being used in other workflows? Wouldn't that then suggest it's been written to accomodate concurrent usage?

dotnet/src/Microsoft.Agents.Workflows/Execution/DeliveryMapping.cs
crickman: > Default would be null. This is an empty enumerable.

Right, but its assigned in constructor.  Isn't there an analysis rule to catch this? ;)

dotnet/src/Microsoft.Agents.Workflows/Run.cs
crickman: Won't this result in posible warnings or errors (depending on analysis rules) that an disposable object isn't disposed (for everyone using workflows)?

Is there something that actually needs disposal that is owned by a run?

lokitoth: It could, unfortunately. 

The `.Dispose()` mechanism there is to release the ownership of Workflow that is taken by the internal run machinery (specifically `InProcRunnerContext`) so it could (if resettable) be run inside of another run.

Workflow ownership is there to ensure that we cannot have two runs operating on the same workflow instance at once.

stephentoub: > If you only intend to use a single Workflow instance once, you do not need to dispose the Run

What does disposal do? I'm curious why a single invocation wouldn't benefit from disposal but multiple invocations would.

lokitoth:
> > if the Workflow has no shared executor instances
> 
> I'm not understanding this comment. Does "shared executor" mean it's being used in other workflows? Wouldn't that then suggest it's been written to accomodate concurrent usage?

I do not really have a good term for this: If you build a workflow with Executor instances rather than factories, and run the workflow in two concurrent runs, those instances will be shared between runs. If you run them sequentially, they will be shared between the sequential runs (which means that any leftover state needs to be reset or they are not in their initial state when starting a new run).

Need to think of a better way of phrasing this.

dotnet/src/Microsoft.Agents.Workflows/Run.cs
crickman: Sounds like the disposable pattern is being overloaded for something that does not require disposal to accomplish.  Also, how does this "ensure" anything if "you do not need to dispose the Run".

I'd respectfully suggest revisiting the data-flow/lifecycle of what's going on here.

I'd be curious if you had an alternative in mind already (it may be a better approach).

lokitoth: It ensures it because it prohibits the user from using a single Workflow instance in two Runs at once unless you Dispose the run that took ownership of it. (And prevents a "non-reset" Workspace from being used in another run sequentially unless cleaned up from the first).

We could conceivably create a `ReleaseAndResetWorkflowAsync` call on `[Streaming]Run`, but it would have the semantics of `.Dispose()` in that that run would no longer be usable after calling it.

stephentoub: Aren't you never allowed to use the same workflow instance in multiple concurrent runs? I'm not understanding then why anything requires explicit disposal. I'd have expected there to be a flag that flips from not-in-use to in-use when it starts and from in-use to not-in-use when it ends, not when a developer says they're done with it. I'm obviously missing something, though.

crickman: Feel free to ping me if you want to brainstorm options.

dotnet/src/Microsoft.Agents.Workflows/Execution/EdgeMap.cs
```diff
@@ -49,103 +60,52 @@ public EdgeMap(IRunnerContext runContext,
...
-        return edgeResults;
+        return edgeRunner.ChaseEdgeAsync(sourceId, message, this._stepTracer);
     }
 
-    // TODO: Should we promote Input to a true "FlowEdge" type?
-    public async ValueTask<IEnumerable<object?>> InvokeInputAsync(MessageEnvelope envelope)
+    public ValueTask<DeliveryMapping?> PrepareDeliveryForInputAsync(MessageEnvelope message)
     {
-        return [await this._inputRunner.ChaseAsync(envelope, this._stepTracer).ConfigureAwait(false)];
+        return this._inputRunner.ChaseEdgeAsync(string.Empty, message, this._stepTracer);
```
TaoChenOSU: nit: it probably doesn't have any real effect, but a real ID is less confusing than `string.Empty`.

dotnet/src/Microsoft.Agents.Workflows/Execution/EdgeMap.cs
```diff
@@ -49,103 +60,52 @@ public EdgeMap(IRunnerContext runContext,
...
-    public async ValueTask<IEnumerable<object?>> InvokeResponseAsync(ExternalResponse response)
+    public ValueTask<DeliveryMapping?> PrepareDeliveryForResponseAsync(ExternalResponse response)
     {
         if (!this._portEdgeRunners.TryGetValue(response.PortInfo.PortId, out InputEdgeRunner? portRunner))
         {
             throw new InvalidOperationException($"Port {response.PortInfo.PortId} not found in the edge map.");
         }
 
-        return [await portRunner.ChaseAsync(new MessageEnvelope(response), this._stepTracer).ConfigureAwait(false)];
+        return portRunner.ChaseEdgeAsync(string.Empty, new MessageEnvelope(response, ExecutorIdentity.None), this._stepTracer);
```
TaoChenOSU: Same here

dotnet/src/Microsoft.Agents.Workflows/Execution/MessageDelivery.cs
```diff
@@ -0,0 +1,28 @@
...
+        : this(envelope, target.Id)
+    {
+        this.TargetCache = Throw.IfNull(target);
+    }
+
+    public string TargetId { get; }
+    public MessageEnvelope Envelope { get; }
+
+    [JsonIgnore]
+    public Executor? TargetCache { get; set; }
```
TaoChenOSU: This probably doesn't need to be public.

dotnet/src/Microsoft.Agents.Workflows/Run.cs
lokitoth: > Aren't you never allowed to use the same workflow instance in multiple concurrent runs

Concurrent runs no, but sequential runs yes, with Disposal.

Think about it this way:

When the user starts a run, that run takes ownership of the Workflow. If the user disposes that run, it releases that Workflow instance and gives it a chance to Reset itself (if all `Executors` support it), allowing the user to use it in a new run.

dotnet/src/Microsoft.Agents.Workflows/Execution/MessageDelivery.cs
lokitoth: Yes, though the entire class is internal

dotnet/src/Microsoft.Agents.Workflows/Execution/EdgeMap.cs
lokitoth: Probably best to replace it with `ExecutorIdentity` and force the underlying edges to deal with `ExecutorIdentity.None` correctly. Otherwise, we might get collisions between user-provided id and whatever we pick here.

dotnet/src/Microsoft.Agents.Workflows/Run.cs
stephentoub: > When the user starts a run, that run takes ownership of the Workflow. If the user disposes that run, it releases that Workflow instance and gives it a chance to Reset itself (if all Executors support it), allowing the user to use it in a new run.

Sure, but if the user doesn't dispose of that run, why can't after the run finishes it just be used again, causing it to reset automatically if that's supported or throw if it's not? I'm not understanding the importance of the user explicitly disposing. A workflow can track whether it's in use or not, knowing that when it completes it's no longer in use. No?

At a much simpler level, a bunch of types do this. When you call ReadAsync on a DeflateStream, it flips a bit saying the stream is in use, and when the async operation finishes, it flips back. If you try to call ReadAsync while another operation is in flight, it sees the flipped bit and throws. After ReadAsync completes, the user doesn't have to explicitly dispose of something to reset it.

lokitoth: There is unfortunately no well-defined "Finished" state. The issue with doing it automatically on Idle is that there is no way to know whether the run is finished (and the user will not send more input messages) or whether the run is just idle for now.

We cannot release the Workflow to work with another run because we would lose any live state for the existing run. 

We cannot rely on Checkpointing, as if the user is not using Checkpointing some of the Executors might not properly support it, not expecting to be needing to do so.

stephentoub: > There is unfortunately no well-defined "Finished" state.

I assumed that the workflow completed event meant that it was finished. Are you saying it doesn't mean that, or we can't rely on that because not all workflows publish such an event, or we can't trust that because the dev might code it wrong, or something else?

lokitoth: We do not have an event that signal the Workflow is "converged" in the sense that it cannot receive more messages - that's part of the reason we renamed the WorkflowCompletedEvent (rather split it into RequestHaltAsync() and YieldOutputAsync()), since it implied semantics which did not exist.

If we want to have a "converged" state we can re-examine the design

stephentoub: What is it today that triggers the end of the enumerable from WatchStreamAsync? That's not some kind of converged state?

lokitoth: Right now, this gets triggered when there are no more messages to process. But the user can always send more messages, or this might be a case where the workflow is expecting an ExternalResponse.

"Converged" would mean the user cannot send more messages, and the live state inside of the workflow is no longer needed, because all outputs that can be generated have already been generated and run.

stephentoub: > the user cannot send more messages

How does a workflow get into a state where a user can't submit more messages to it? Does an executor do something to explicitly say "I don't accept any more messages"?

> the live state inside of the workflow is no longer needed

So the distinction between a new run and more messages for an existing run is just whether the state of the executors are reset to a starting state?

The dispose part doesn't feel good to me, unless we're going to say runs should always be disposed (and our samples do it). Otherwise, if we're going to say you only need to dispose if you're going to reuse the workflow, then I would just assume have running the workflow implicitly dispose the previous run, and if it's not in a good state to safely be disposed (namely the same state it's in when the watch enumerator ends), then run would throw.

With that, it seems like the 90% case "just works" and the dev doesn't need to be concerned with extra ceremony. Only in the case where they try to run and it's currently got messages or work in flight, or the workflow had non reusable executors, does ruin then fail.

lokitoth: > How does a workflow get into a state where a user can't submit more messages to it? Does an executor do something to explicitly say "I don't accept any more messages"?

There is no currently designed way to do this. This is why we cannot determine that a Run is "done".

> So the distinction between a new run and more messages for an existing run is just whether the state of the executors are reset to a starting state?

Not just executors - the full state of the RunContext and EdgeRunner states too. If we force Checkpointing we do not have this problem, but it comes at a pretty big cost if the user does not intend to use the Checkpointing feature.

> I would just assume have running the workflow implicitly dispose the previous run

It seems somewhat unintuitive for a reachable object to "become dead" out from under the user when interacting with a different object.

I feel like I am a little confused where the original confusion comes from? `.Dispose()` is the standard way in .NET to let objects know you are done with them.

stephentoub: > I feel like I am a little confused where the original confusion comes from? .Dispose() is the standard way in .NET to let objects know you are done with them.

It's not really a way to say you're done with them, it's a way to aggressively dispose resources they're using. Sometimes it's abused as a way to signal stop, but because of its resource-based nature, the culture around it is that if something is disposable you must dispose of it. Any number of analyzers try to warn you when you don't, and because the involved control flow is often tricky, they often have an annoyingly large number of false positives. Which means that every time something is disposable, it increases a burden on developers. If something does contain resources, then yeah, it should absolutely be disposable. But if it doesn't contain resources that must be cleaned up, and if there are other ways to achieve the same goal that don't involve implementing I{Async}Disposable, that's often better. This is coming from experience; more times than I'd like to count I've made something disposable and then regretted it. (First and foremost on that list is Task itself... that one is so bad, we updated all the analyzer rules we could find to special-case it.)

Which is why you see me searching for alternatives here, since there aren't actually resources that need to be disposed of, as evidenced by the comments that you don't need to dispose of it if you're not going to reuse the workflow. It's really just acting as a signal to say "I'm explicitly saying I'm done so now it's ok to reuse". And that's not actually necessary, since you could just reuse and invalidate the existing object automatically. Which means it's a question of which is better/worse: forcing a developer to always call Dispose, regardless of whether they're going to reuse the workflow, with all of the associated noise around that, or assuming that a developer knows what they're doing when they try to reuse a workflow for which the previously handed out run has quiesced.

lokitoth: Okay, that makes sense. I appreciate the explanation. For some reason my understanding of the semantics of IDisposable was as "Release Ownership / Right-to-Use". 

To be clear though: It is releasing a resource that it owns: The Workflow instance, so I am not sure it doesn't fit into the semantic meaning of Dispose.

I really do not like the idea of objects - without anything done on them or with them - going from working to throwing `InvalidOperationException` (currently ObjectDisposed). Can rename this API to something else (e.g. `ReleaseWorkflowAsync`) - but that would have the consequences of users all having to do try/finally to clean up properly.

That is what is good about IDisposable is that it does the right thing out of the box when used with `using`. I really wish there was a generic "run cleanup when exiting block" construct in C# that did not have this semantics.

>  or assuming that a developer knows what they're doing when they try to reuse a workflow for which the previously handed out run has quiesced.

That works, but at that point it seems like we are putting up a barrier for the developer whose solution would still look like an "explicit release": Since a run can be Idle but the user may still be expecting to send messages to it, there will need to be coordination between when the run is actually done, and when a new run can be started.

For developers who are not deeply aware of how it works, it becomes a footgun - in their individual tests, when they are not actually stomping over not-"finished" runs, it works fine. But when it starts running at-scale they get odd "object no longer usable" behaviours.

I get the desire to avoid explicit release, but I think it is necessary for getting users into the right usage patterns earlier (due to earlier failure and a guiding error message).

stephentoub: If you want it to be disposable, then we should always dispose it, in all of our samples. We shouldn't try to caveat it.

If we don't want to force devs to always do that, then we shouldn't make it IDisposable. It's not releasing a resource in the resource management sense; it's just setting a flag on it saying that subsequent use should progress rather than throw. If you don't dispose and you drop all references to it, there aren't any leaks that will occur.

stephentoub: > Since a run can be Idle but the user may still be expecting to send messages to it, there will need to be coordination between when the run is actually done, and when a new run can be started.

That is the case, regardless. A developer needs to be coordinating around when the workflow can be reused; the only question here is if they explicitly have to call a dispose method or not. Even with disposal, they still need said coordination.

lokitoth: I think I get the source of my confusion: When you say "resource", you are meaning something only along the lines of unmanaged memory - explicitly not automatically cleaned up, excluding the category of "can clean up early" but will still be automatically cleaned up. 

Going to go with `.EndRunAsync()` for now. Since I still intend to add (opt-in only) way of letting users who know what they are doing actually run workflows concurrently, I think this gets us to a good solution re: `Dispose()` while also keeping new users in a reasonably understandable box and not overloading them with boilerplate stuff they have to do due to `IDispose` semantics.

lokitoth: Going to mark this resolved for now and follow up offline re: `EndRunAsync()`

PR #947, opened 2025-09-26
.NET: Enable access to hosted AIAgents via OpenAI Responses

DeagleGross:
### Description

Current PR provides the API to register custom endpoint to access the `AIAgent` via OpenAI Responses protocol:
```csharp
app.MapOpenAIResponses(agentName: ..., path: ...);
```

It supports both streaming the response [as per spec](https://platform.openai.com/docs/api-reference/responses-streaming) and just getting the response [in a single entity](https://platform.openai.com/docs/api-reference/responses/create?lang=curl).

It does not support [conversations](https://platform.openai.com/docs/api-reference/conversations/create?lang=curl) yet, because AF at the moment does not have a unified way to perform management operations with the conversation via the `conversationId` (as required by the OpenAI Responses protocol). In other words - at the moment of writing we are designing the API to restore/check existence of the thread using `threadId`. It will be added as a follow-up.

### Implementation details

To simplify future maintanence and ease the development, I tried my best to not introduce extra types to represent protocol input/output models, but to rely on official [OpenAI SDK](https://github.com/openai/openai-dotnet), which are also supported by MEAI.

However, most of the types in OpenAI SDK are: 1) internal; 2) have internal constructors; 3) allow instantiation ONLY via deserialization from the JSON, and define there own logic for serialization (can't be used directly in the model with the intent of performing `JsonSerializer.Serialize()` on them). We hope this will change later to ease manipulation with the objects, and we will react by removing extra JsonConverters / types / etc. But today I made my code rely on some of the types and the serialziation implementation details, and even accessing some internal fields.

### Validation

Here is the example of streaming events output (the most complicated scenario here):
> event: response.created
data: {"response":{"metadata":{},"temperature":null,"top_p":null,"user":null,"tools":[],"id":"chatcmpl-CM98LCdU6ZDn9HoDg5cUuEBfmPCfF","object":"response","status":"completed","created_at":1759394361,"error":null,"incomplete_details":null,"output":[{"type":"message","role":"assistant","content":[{"type":"output_text","text":"","annotations":[]}]}],"parallel_tool_calls":false},"type":"response.created","sequence_number":1}

> event: response.output_item.added
data: {"output_index":1,"item":{"type":"message","role":"assistant","content":[{"type":"output_text","text":"","annotations":[]}]},"type":"response.output_item.added","sequence_number":2}

> ... <many response.output_item.added here> ...

> event: response.output_item.added
data: {"output_index":88,"item":null,"type":"response.output_item.added","sequence_number":89}

> event: response.output_item.done
data: {"output_index":89,"item":{"type":"message","role":"assistant","content":[{"type":"output_text","text":"","annotations":[]}]},"type":"response.output_item.done","sequence_number":90}

> event: response.completed
data: {"response":{"metadata":{},"temperature":null,"top_p":null,"user":null,"tools":[],"id":"chatcmpl-CM98LCdU6ZDn9HoDg5cUuEBfmPCfF","object":"response","status":"completed","created_at":1759394361,"error":null,"incomplete_details":null,"output":[{"type":"message","role":"assistant","content":[{"type":"output_text","text":"","annotations":[]}]}],"parallel_tool_calls":false},"type":"response.completed","sequence_number":91}


dotnet/samples/AgentWebChat/AgentWebChat.Web/OpenAIResponsesActorClient.cs
```diff
@@ -0,0 +1,195 @@
...
+    {
+        // this is a "root" of the OpenAI Responses surface for the agent
+        var relativeUri = "/" + request.ActorId.Type + "/v1/";
+
+        OpenAIClientOptions options = new()
+        {
+            Endpoint = new Uri(this._baseUri + relativeUri)
+        };
+
+        OpenAIResponseClient client = new(
```
stephentoub: I haven't yet fully comprehended everything you're doing here, but if there's already support for the Microsoft.Extensions.AI object model elsewhere in the project, rather than using the OpenAIResponseClient directly, you can use an IChatClient for it you get from calling AsIChatClient on the OpenAIResponseClient.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI.Responses/Internal/AIAgentResponsesProcessor.cs
```diff
@@ -0,0 +1,129 @@
...
+        var options = new OpenAIResponsesRunOptions();
+        var chatMessages = createResponse.Input.ToChatMessages();
+
+        if (createResponse.Stream)
+        {
+            return new OpenAIStreamingResponsesResult(this._agentProxy, chatMessages, agentThread, options);
+        }
+
+        var agentResponse = await this._agentProxy.RunAsync(chatMessages, agentThread, options, cancellationToken).ConfigureAwait(false);
+        var openAIResponse = agentResponse.ToOpenAIResponse(this.AgentType, agentThread, options);
```
stephentoub: There are helpers for this and related conversions that are part of Microsoft.Extensions.AI.OpenAI. Can you use those rather than writing additional ones?
https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.AI.OpenAI/MicrosoftExtensionsAIResponsesExtensions.cs
Or if there's something necessary missing there, we could look at adding it so that we don't need duplication here.

dotnet/src/Microsoft.Extensions.AI.Agents.Hosting.OpenAI.Responses/Internal/AIAgentResponsesProcessor.cs
```diff
@@ -0,0 +1,129 @@
...
+
+                StreamingResponse GiveNextStreamingResponseChunk(StreamingResponseType type) => new()
+                {
+                    Type = type,
+                    SequenceNumber = sequenceNumber++,
+                    Response = responseHandle.ToOpenAIResponse(responseUpdate, thread, options)
+                };
+            }
+
+            async ValueTask SendDataAsync<T>(T data, JsonTypeInfo typeInfo)
```
stephentoub: Can you use the SseFormatter that's part of System.Net.ServerSentEvents instead of open-coding this?

DeagleGross: Sure, did not use System.Net.ServerSentEvents before ever. Let me know if usage is correct please!

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI.Responses/Internal/AIAgentResponsesProcessor.cs
```diff
@@ -0,0 +1,128 @@
...
+
+            var streamingResponseEventTypeInfo = OpenAIResponsesJsonUtilities.DefaultOptions.GetTypeInfo(typeof(StreamingResponse));
+
+            return SseFormatter.WriteAsync(
+                source: this.GetStreamingResponsesAsync(cancellationToken),
+                destination: response.Body,
+                itemFormatter: (sseItem, bufferWriter) =>
+                {
+                    var json = JsonSerializer.SerializeToUtf8Bytes(sseItem.Data, streamingResponseEventTypeInfo);
+                    bufferWriter.Write(json);
```
stephentoub: @eiriktsarpalis, is the best option here still to wrap a new Utf8JsonWriter around the IBufferWriter? I thought we were exposing a JsonSerializer.Serialize method that took an IBufferWriter, but it looks like we didn't yet?

eiriktsarpalis: We have `PipeWriter` overloads, but not `IBufferWriter` overloads. This is tracked by https://github.com/dotnet/runtime/issues/112943

dotnet/samples/AgentWebChat/AgentWebChat.Web/OpenAIResponsesActorClient.cs
DeagleGross: Thanks for pointing that out; My main goal here (in a sample) is to use some "known and trusted" SDK (such as OpenAI or MEAI.OpenAI) to verify that whatever I return from the server is correct via the spec (and SDK).

I've refactored using `IChatClient`, but I figured out I've supported the response a bit differently that SDK expects:
1) I made a trivial implementation which returns `response.created`, `response.in_progress` (while work is done) and `response.completed`
2) `MEAI.OpenAI.OpenAIResponsesChatClient` never processes such an event (see https://github.com/dotnet/extensions/blob/6786d6d900ca18fb869baca49ea9d34972ff9f22/src/Libraries/Microsoft.Extensions.AI.OpenAI/OpenAIResponsesChatClient.cs#L257).

I am going to add a PR to MEAI to support [more of the types per spec](https://platform.openai.com/docs/api-reference/responses-streaming/response/in_progress?lang=curl); but that also raises the question whether the server implementation I've done makes sense, because probably I should send [response.content_part.added](https://platform.openai.com/docs/api-reference/responses-streaming/response/content_part/added) for example.

stephentoub: > because probably I should send [response.content_part.added](https://platform.openai.com/docs/api-reference/responses-streaming/response/content_part/added) for example.

Yes, it should conform to the full sequence of events.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI.Responses/Internal/AIAgentResponsesProcessor.cs
DeagleGross: For most of the types I managed to reuse existing MEAI mapping extensions. I think we will need to revisit both MEAI and this package implementations once OpenAI SDK "opens" more of the types and makes it more easy to work with them (at least allows to construct them normally, but not via the serialization mechanisms only)

dotnet/samples/AgentWebChat/AgentWebChat.Web/OpenAIResponsesActorClient.cs
DeagleGross: Made my code rely on MEAI extensions more and do most of the mapping/conversions via that package.

About sequence of events: I've implemented something simple as we discussed:
1) response.created
2) [multiple] response.output_item.added
3) response.output_item.done
4) response.created

We can revisit it later, but such a sequence gives clear identification of when the response streaming starts and ends, and when the all of the content is being pushed.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Utils/ResponseItemJsonConverter.cs
```diff
@@ -0,0 +1,33 @@
...
+        var item = ResponseItem.CreateUserMessageItem(""); // no other way to instantiate it.
+        var jsonModel = item as IJsonModel<ResponseItem>;
+        Debug.Assert(jsonModel is not null, "ResponseItem should implement IJsonModel<ResponseItem>");
+        return jsonModel.Create(ref reader, ModelReaderWriterOptions.Json);
+    }
+
+    public override void Write(Utf8JsonWriter writer, ResponseItem? value, JsonSerializerOptions options)
+    {
+        if (value is null)
+        {
```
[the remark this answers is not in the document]
DeagleGross: yep, nice find.

dotnet/samples/AgentWebChat/AgentWebChat.Web/OpenAIResponsesAgentClient.cs
```diff
@@ -0,0 +1,63 @@
...
+    {
+        // this is a "root" of the OpenAI Responses surface for the agent
+        var relativeUri = "/" + agentName + "/v1/";
+
+        OpenAIClientOptions options = new()
+        {
+            Endpoint = new Uri(this._baseUri + relativeUri)
+        };
+
+#pragma warning disable OPENAI001 // Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.
```
stephentoub: This should no longer be necessary. It's now suppressed here:
https://github.com/microsoft/agent-framework/blob/348a07a0b1c84a5d9c2f98341ab50bbca80d983d/dotnet/samples/.editorconfig#L15

dotnet/samples/AgentWebChat/AgentWebChat.Web/OpenAIResponsesActorClient.cs
```diff
@@ -0,0 +1,63 @@
...
+using Microsoft.Agents.AI;
+using Microsoft.Extensions.AI;
+using OpenAI;
+using OpenAI.Responses;
+
+namespace AgentWebChat.Web;
+
+[SuppressMessage("Style", "IDE0051:Remove unused private members", Justification = "for debug / sample purposes")]
+[SuppressMessage("Style", "IDE0059:Unnecessary assignment of a value", Justification = "debug")]
+internal sealed class OpenAIResponsesActorClient : IAgentClient
```
stephentoub: Consider using something other than responses for the agent, so that it's clear this isn't just a raw proxy.

dotnet/samples/AgentWebChat/AgentWebChat.Web/OpenAIResponsesAgentClient.cs
```diff
@@ -0,0 +1,63 @@
...
+        {
+            ConversationId = threadId
+        };
+
+        await foreach (var update in chatClient.GetStreamingResponseAsync(messages, chatOptions, cancellationToken: cancellationToken))
+        {
+            // Depending on what type is returned by server MEAI will convert it differently, and since OpenAI SDK
+            // has really "closed" most of types definitions
+            // (i.e. https://github.com/openai/openai-dotnet/blob/84286cd916adc7b03e50cc8031b587714af016ed/src/Generated/Models/Responses/InternalResponsesAssistantMessage.cs#L12)
+            // it may appear that sample frontend will not be able to interpret some of the output.
```
stephentoub: What kind of output do we care about that it won't be able to interpret?

I would delete this comment.

dotnet/samples/AgentWebChat/AgentWebChat.Web/OpenAIResponsesActorClient.cs
stephentoub: Why does the sample need an `IAgentClient` interface? We can't just register the AIAgent directly?

dotnet/src/Microsoft.Agents.AI.Abstractions/Extensions/AgentRunResponseExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+/// Provides extension methods for <see cref="AgentRunResponse"/>.
+/// </summary>
+public static class AgentRunResponseExtensions
+{
+    /// <summary>
+    /// Converts an <see cref="AgentRunResponse"/> instance to a <see cref="ChatResponse"/>.
+    /// </summary>
+    /// <param name="response">The <see cref="AgentRunResponse"/> to convert. Cannot be null.</param>
+    /// <returns>A <see cref="ChatResponse"/> populated with values from <paramref name="response"/>.</returns>
+    public static ChatResponse AsChatResponse(this AgentRunResponse response) => new(response.Messages)
```
stephentoub: https://github.com/microsoft/agent-framework/blob/08ea625de00a4406a5ea0a7586f66381370637d6/dotnet/src/Microsoft.Agents.AI.Abstractions/AgentRunResponseExtensions.cs#L31

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Internal/AIAgentResponsesProcessor.cs
```diff
@@ -0,0 +1,187 @@
...
+            {
+                // complete the whole streaming with the full response model
+                var responseCompleted = new StreamingCompletedResponse(sequenceNumber++)
+                {
+                    Response = lastOpenAIResponse
+                };
+                yield return GiveNextSseItem(responseCompleted);
+            }
+
+            static SseItem<StreamingResponseEventBase> GiveNextSseItem<T>(T item) where T : StreamingResponseEventBase
```
ReubenBond: IMO we should inline this for clarity (`new SseItem<StreamingResponseEventBase>(item, item.Type)` instead of calling this helper method)

dotnet/samples/AgentWebChat/AgentWebChat.Web/OpenAIResponsesActorClient.cs
ReubenBond: s/Actor/Agent

DeagleGross: renamed to `OpenAIResponsesAgentClient`

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Internal/AIAgentResponsesProcessor.cs
DeagleGross: I dont like that I need to manually write the `Type` everytime for that, so i created this method. without it I would need to manually write something like 
`new SseItem<StreamingResponseEventBase>(item, StreamingOutputItemAddedResponse.Type)`
`new SseItem<StreamingResponseEventBase>(item, StreamingOutputItemDoneResponse.Type)`

I like current approach a bit more because of readability. What do you think?

dotnet/samples/AgentWebChat/AgentWebChat.Web/OpenAIResponsesAgentClient.cs
```diff
@@ -0,0 +1,63 @@
...
+        IList<ChatMessage> messages,
+        string? threadId = null,
+        [EnumeratorCancellation] CancellationToken cancellationToken = default)
+    {
+        // this is a "root" of the OpenAI Responses surface for the agent
+        var relativeUri = "/" + agentName + "/v1/";
+
+        OpenAIClientOptions options = new()
+        {
+            Endpoint = new Uri(this._baseUri + relativeUri)
```
stephentoub: Nit: it'd be better if _baseUri were an actual Uri, and then this could use the Uri ctor that takes a base Uri and a relative string as two separate arguments

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Utils/AgentRunResponseUpdateExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+/// </summary>
+public static class AgentRunResponseUpdateExtensions
+{
+    /// <summary>
+    /// Converts an <see cref="AgentRunResponseUpdate"/> instance to a <see cref="ChatResponse"/>.
+    /// </summary>
+    /// <param name="response">The <see cref="AgentRunResponse"/> to convert. Cannot be null.</param>
+    /// <param name="role">The role of agent run response contents. By default is <see cref="ChatRole.Assistant"/>.</param>
+    /// <returns>A <see cref="ChatResponse"/> populated with values from <paramref name="response"/>.</returns>
+    public static ChatResponse AsChatResponse(this AgentRunResponseUpdate response, ChatRole? role = null) => new()
```
stephentoub, line 15: I'm not convinced this is a general-purpose API we want to expose.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/EndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,91 @@
...
+using Microsoft.Extensions.Logging;
+using OpenAI.Responses;
+
+namespace Microsoft.Agents.AI.Hosting.OpenAI;
+
+/// <summary>
+/// Provides extension methods for mapping OpenAI Responses capabilities to an <see cref="AIAgent"/>.
+/// </summary>
+[SuppressMessage("Trimming", "IL2026:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code", Justification = "<Pending>")]
+[SuppressMessage("AOT", "IL3050:Calling members annotated with 'RequiresDynamicCodeAttribute' may break functionality when AOT compiling.", Justification = "<Pending>")]
```
stephentoub: Are these just temporary but will be addressed before merging? We should only be suppressing Native AOT / linking warnings if there are other reasons that they're actually safe. In general, we should be doing the work to ensure that all of these libraries compile successfully with zero NAOT / trimmer warnings.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/EndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,91 @@
...
+    /// <param name="responsesPath"></param>
+    /// <param name="conversationsPath"></param>
+    public static void MapOpenAIResponses(
+        this IEndpointRouteBuilder endpoints,
+        string agentName,
+        [StringSyntax("Route")] string? responsesPath = null,
+        [StringSyntax("Route")] string? conversationsPath = null)
+    {
+        ArgumentNullException.ThrowIfNull(endpoints);
+        ArgumentNullException.ThrowIfNull(agentName, nameof(agentName));
```
stephentoub:
```suggestion
        ArgumentNullException.ThrowIfNull(agentName);
```

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/EndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,91 @@
...
+
+        var loggerFactory = endpoints.ServiceProvider.GetService<ILoggerFactory>();
+        var agent = endpoints.ServiceProvider.GetRequiredKeyedService<AIAgent>(agentName);
+        ArgumentNullException.ThrowIfNull(agent.Name, nameof(agent.Name));
```
stephentoub: agent here isn't an argument, so ArgumentNullException isn't the right kind of exception.

Why does this need to throw if Name is null? Can you use DisplayName, or if you need a unique identifier, Id?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Internal/AIAgentResponsesProcessor.cs
```diff
@@ -0,0 +1,188 @@
...
+            response.Headers.Connection = "keep-alive";
+            response.Headers.ContentEncoding = "identity";
+            httpContext.Features.GetRequiredFeature<IHttpResponseBodyFeature>().DisableBuffering();
+
+            return SseFormatter.WriteAsync(
+                source: this.GetStreamingResponsesAsync(cancellationToken),
+                destination: response.Body,
+                itemFormatter: (sseItem, bufferWriter) =>
+                {
+                    var json = JsonSerializer.SerializeToUtf8Bytes(sseItem.Data, sseItem.Data.GetType(), OpenAIResponsesJsonUtilities.DefaultOptions);
```
stephentoub: Is this why you have the SuppressMessage warnings?

To avoid that,
a) Make sure the OpenAIResponsesJsonUtilities context has source generated info for every relevant type, and then
b) change this to:
```suggestion
                    var json = JsonSerializer.SerializeToUtf8Bytes(sseItem.Data, OpenAIResponsesJsonUtilities.DefaultOptions.GetTypeInfo(sseItem.Data.GetType()));
```

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Internal/AIAgentResponsesProcessor.cs
```diff
@@ -0,0 +1,188 @@
...
+
+            ResponseItem? lastResponseItem = null;
+            OpenAIResponse? lastOpenAIResponse = null;
+
+            await foreach (var update in agent.RunStreamingAsync(chatMessages, thread: agentThread, cancellationToken: cancellationToken).ConfigureAwait(false))
+            {
+                if (string.IsNullOrEmpty(update.ResponseId))
+                {
+                    continue;
+                }
```
stephentoub: This looks problematic. There's no guarantee ResponseId will be populated. Is it required to have a response ID? Assuming yes, I could imagine a few options:
1. Make it a requirement that ResponseId always be filled in by agents, but that's hard to enforce, especially since it's already not a hard and fast requirement of IChatClient.
2. Manufacture a response ID that's used if the first update has a null response ID.
3. Buffer all updates that have a null response ID until either the end is reached or the first update happens where there is a response ID, at which point that ID can be used for all buffered items and all of them released.
4. Leave it as is, and document that this only works well when a response ID is set.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Internal/AIAgentResponsesProcessor.cs
stephentoub: The call site could use target-typed new as well, no? e.g.
```csharp
yield return new(responseCompleted, responseCompleted.Type);
```

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Model/ResponseStreamEvent.cs
```diff
@@ -0,0 +1,163 @@
...
+using OpenAI.Responses;
+
+namespace Microsoft.Agents.AI.Hosting.OpenAI.Responses.Model;
+
+/// <summary>
+/// Abstract base class for all streaming response events in the OpenAI Responses API.
+/// Provides common properties shared across all streaming event types.
+/// </summary>
+#pragma warning disable CA1012 // Abstract types should not have public constructors
+public abstract class StreamingResponseEventBase
```
stephentoub: Why do any of these model types need to be public?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Utils/OpenAIResponseJsonConverter.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Diagnostics;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using OpenAI.Responses;
+
+namespace Microsoft.Agents.AI.Hosting.OpenAI.Responses.Utils;
+
+internal sealed class OpenAIResponseJsonConverter : JsonConverter<OpenAIResponse>
+{
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Performance", "CA1859:Use concrete types when possible for improved performance", Justification = "relying specifically on IJsonModel for json conversion")]
```
stephentoub: What was the analyzer complaining about here?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Utils/OpenAIResponsesJsonUtilities.cs
```diff
@@ -0,0 +1,45 @@
...
+    [UnconditionalSuppressMessage("ReflectionAnalysis", "IL3050:RequiresDynamicCode", Justification = "Converter is guarded by IsReflectionEnabledByDefault check.")]
+    [UnconditionalSuppressMessage("Trimming", "IL2026:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access", Justification = "Converter is guarded by IsReflectionEnabledByDefault check.")]
+    private static JsonSerializerOptions CreateDefaultOptions()
+    {
+        JsonSerializerOptions options = new();
+
+        options.Converters.Add(new ResponseItemJsonConverter());
+        options.Converters.Add(new OpenAIResponseJsonConverter());
+
+        return options;
```
stephentoub: This should use the JSON source generator.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Utils/ResponseCreationOptionsExtensions.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        // --- Input (internal IList<ResponseItem> Input { get; set; }) ---
+        const string inputPropName = "Input";
+        var inputProp = typeof(ResponseCreationOptions).GetProperty(inputPropName, BindingFlags.Instance | BindingFlags.NonPublic)
+            ?? throw new MissingMemberException(typeof(ResponseCreationOptions).FullName!, inputPropName);
+        var inputGetter = inputProp.GetGetMethod(nonPublic: true)
+            ?? throw new MissingMethodException($"{inputPropName} getter not found.");
+
+        var inputCall = Expression.Call(param, inputGetter);
+        _getInput = Expression.Lambda<Func<ResponseCreationOptions, IList<ResponseItem>>>(inputCall, param).Compile();
```
stephentoub: Why does this need System.Linq.Expressions? Can you instead achieve it with just CreateDelegate?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Internal/AIAgentResponsesProcessor.cs
DeagleGross: ye, missed that, sorry. removed method and using `yield return new`

dotnet/samples/AgentWebChat/AgentWebChat.Web/OpenAIResponsesActorClient.cs
DeagleGross: That's a common interface for different clients (aka protocols) for sample app:
```
	// Select the appropriate client based on protocol
	IAgentClient agentClient = selectedProtocol switch
	{
		Protocol.OpenAIResponses => OpenAIResponsesAgentClient,
		Protocol.A2A or _ => A2AActorClient
	};
```

dotnet/samples/AgentWebChat/AgentWebChat.Web/OpenAIResponsesAgentClient.cs
DeagleGross: I think I formulated my thought not clearly here. It's basically because of OpenAI SDK implementation, MEAI and this sample cant really read the response correctly
image

I've unpacked `update.RawRepresentation.Response.OutputItems[0]` and its of type `OpenAI.Responses.InternalResponsesAssistantMessage` which [is internal](https://github.com/openai/openai-dotnet/blob/ef7ffbae2398a48a3be11874410ccc754ba79fb5/src/Generated/Models/Responses/InternalResponsesAssistantMessage.cs#L12), and that's why MEAI cant read that correctly.

That's the reason for `update.Contents` being empty:
image

My comment was just trying to state this situation, and sample potentially being not able to correctly parse the OpenAI Response.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Utils/AgentRunResponseUpdateExtensions.cs
DeagleGross: Agree, its too specific for OpenAI Responses streaming, so moved as internal type to Hosting.AI.OpenAI package

dotnet/src/Microsoft.Agents.AI.Abstractions/Extensions/AgentRunResponseExtensions.cs
DeagleGross: removed my extension, reusing this one

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/EndpointRouteBuilderExtensions.cs
DeagleGross: Needed for the endpoint name -> I've changed to use Name or Id if not present. Not throwing exception anymore

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/EndpointRouteBuilderExtensions.cs
DeagleGross: Yes, wanted to get the first feedback on other parts. I am not sure how to make aspnetcore endpoints AOT friendly though, because even `routeGroup.MapGet("/test", () => "test");` such a simple endpoint definition reports IL2026 and IL3050.

I will check how to resolve those correctly

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Utils/ResponseCreationOptionsExtensions.cs
DeagleGross: Removed Linq.Expressions usage here

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Model/ResponseStreamEvent.cs
DeagleGross: changed to be internal visibility

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Utils/OpenAIResponseJsonConverter.cs
DeagleGross: My bad; used a wrong initialization method. It was complaining about doing a cast to a specific type; but fixed it by creating an instance of `OpenAIResponse` properly

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/EndpointRouteBuilderExtensions.cs
stephentoub: >> Why does this need to throw if Name is null? Can you use DisplayName, or if you need a unique identifier, Id?

> I've changed to use Name or Id if not present

That's what DisplayName is, fwiw:
https://github.com/microsoft/agent-framework/blob/c3b995fb6b9294934539660e4ee49d7926868ec7/dotnet/src/Microsoft.Agents.AI.Abstractions/AIAgent.cs#L64

dotnet/samples/AgentWebChat/AgentWebChat.Web/OpenAIResponsesAgentClient.cs
stephentoub: I think the comment just muddies the waters and should be removed.

One should be able to use any inner AIAgent and the consuming code should be the same; there shouldn't be a need to fish out / use internal raw representations because there can't be an assumption that any particular kind of inner agent will be used. This is getting a bit at why I'm questioning the IAgentClient implementation... it suggests you need to do custom things for different kinds of AIAgent implementations, but you shouldn't need to... you should be able to just register any AIAgent and have it "just work".

Am I still misunderstanding?

DeagleGross: From the usability perspective - yes, AIAgent should "just work". I wanted to leave some warning to let code readers see that sample could potentially not read the message correctly.

But I agree, that it just adds confusion - removed the comment here.

dotnet/samples/AgentWebChat/AgentWebChat.Web/OpenAIResponsesActorClient.cs
ReubenBond: If we have AIAgent wrappers which use OpenAI Responses / A2A, etc, then we should be able to use those as clients, either directly or via some general-purpose wrapper if the Web app needs it.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Utils/OpenAIResponsesJsonUtilities.cs
DeagleGross: added JsonContext with configured serialization of event types.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Internal/AIAgentResponsesProcessor.cs
DeagleGross: Changed to suggested code + configured `OpenAIResponsesJsonUtilities.DefaultOptions` to support streaming types

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/EndpointRouteBuilderExtensions.cs
DeagleGross: Solved the AOT support via
``` <InterceptorsNamespaces>$(InterceptorsNamespaces);Microsoft.AspNetCore.Http.Generated</InterceptorsNamespaces>
    <EnableRequestDelegateGenerator>true</EnableRequestDelegateGenerator>
```

Following https://learn.microsoft.com/en-us/aspnet/core/fundamentals/aot/request-delegate-generator/rdg?view=aspnetcore-9.0

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Internal/AIAgentResponsesProcessor.cs
DeagleGross: I think it is not required, I just wanted to avoid an empty results streamed like:

>event: response.created
data: {"response":{"metadata":{},"temperature":null,"top_p":null,"user":null,"tools":[],"id":"","object":"response","status":"completed","created_at":0,"error":null,"incomplete_details":null,"output":[],"parallel_tool_calls":false},"type":"response.created","sequence_number":1}

For some reason performing `agent.RunStreamingAsync` may return the first update which does not have responseId, messageId or any contents.

I've changed to check that at least some useful info is in the update, and if none, skip it.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Internal/AIAgentResponsesProcessor.cs
```diff
@@ -116,7 +116,9 @@ private async IAsyncEnumerable<SseItem<StreamingResponseEventBase>> GetStreaming
 
             await foreach (var update in agent.RunStreamingAsync(chatMessages, thread: agentThread, cancellationToken: cancellationToken).ConfigureAwait(false))
             {
-                if (string.IsNullOrEmpty(update.ResponseId))
+                if (string.IsNullOrEmpty(update.ResponseId)
+                    || string.IsNullOrEmpty(update.MessageId)
+                    || update.Contents is null || update.Contents.Count == 0)
```
stephentoub: Doesn't this make it worse? Presumably you'd want most of these to be && rather than ||?

DeagleGross: Agree, just realized I wrote it incorrectly after pushing

dotnet/samples/AgentWebChat/AgentWebChat.Web/OpenAIResponsesActorClient.cs
DeagleGross: discussed offline: this is an expected proxy on the frontend part of the sample app.

stephentoub: Yup, I was misunderstanding where this fit in the architecture

.github/workflows/dotnet-format.yml
```diff
@@ -91,5 +91,5 @@ jobs:
         run: |
           for csproj in ${{ steps.find-csproj.outputs.csproj_files }}; do
             echo "Running dotnet format on $csproj"
-            docker run --rm -v $(pwd):/app -w /app mcr.microsoft.com/dotnet/sdk:${{ matrix.dotnet }} /bin/sh -c "dotnet format $csproj --verify-no-changes --verbosity diagnostic"
+            docker run --rm -v $(pwd):/app -w /app mcr.microsoft.com/dotnet/sdk:${{ matrix.dotnet }} /bin/sh -c "dotnet format $csproj --verify-no-changes --verbosity diagnostic --exclude-diagnostics IL2026 IL3050"
```
stephentoub, line 95: Is there an issue tracking the dotnet format issue somewhere? It'd be good to include a comment on this tying it back to that issue, so that this can be reverted when the issue is addressed.

dotnet/samples/AgentWebChat/AgentWebChat.Web/OpenAIResponsesAgentClient.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.Runtime.CompilerServices;
+using A2A;
+using Microsoft.Agents.AI;
+using Microsoft.Extensions.AI;
+using OpenAI;
+using OpenAI.Responses;
+
+namespace AgentWebChat.Web;
+
+internal sealed class OpenAIResponsesAgentClient : IAgentClient
```
stephentoub, line 16: Nit: To alleviate some of the confusion I previously had when reviewing this, consider adding a comment to this type indicating that it's the frontend client to exercise the agent serving Responses responses.

dotnet/samples/AgentWebChat/AgentWebChat.Web/OpenAIResponsesAgentClient.cs
```diff
@@ -0,0 +1,51 @@
...
+        string? threadId = null,
+        [EnumeratorCancellation] CancellationToken cancellationToken = default)
+    {
+        // this is a "root" of the OpenAI Responses surface for the agent
+        var relativeUri = "/" + agentName + "/v1/";
+
+        OpenAIClientOptions options = new()
+        {
+            Endpoint = new Uri(this._baseUri, relativeUri)
+        };
```
stephentoub:
```suggestion
        OpenAIClientOptions options = new()
        {
            Endpoint = new Uri(this._baseUri, $"/{agentName}/v1/")
        };
```

dotnet/samples/AgentWebChat/AgentWebChat.Web/OpenAIResponsesAgentClient.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        OpenAIResponseClient openAiClient = new(model: "myModel!", credential: new ApiKeyCredential("dummy-key"), options: options);
+        var chatClient = openAiClient.AsIChatClient();
```
stephentoub:
```suggestion
        chatClient = new OpenAIResponseClient(model: "myModel!", credential: new ApiKeyCredential("dummy-key"), options: options).AsIChatClient();
```

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/EndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,88 @@
...
+/// </summary>
+public static class EndpointRouteBuilderExtensions
+{
+    /// <summary>
+    /// Maps OpenAI Responses endpoints to the specified <see cref="IEndpointRouteBuilder"/> for the given <see cref="AIAgent"/>.
+    /// </summary>
+    /// <param name="endpoints"></param>
+    /// <param name="agentName"></param>
+    /// <param name="responsesPath"></param>
+    /// <param name="conversationsPath"></param>
```
stephentoub: Missing comments

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/EndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,88 @@
...
+    public static void MapOpenAIResponses(
+        this IEndpointRouteBuilder endpoints,
+        string agentName,
```
stephentoub, line 31: Given that this is used in a route, what constraints are there on the value?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/EndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        responsesPath ??= $"/{agentName}/v1/responses";
+        var responsesRouteGroup = endpoints.MapGroup(responsesPath);
+        MapResponses(responsesRouteGroup, agent, loggerFactory);
+
+        // Will be included once we obtain the API to operate with thread (conversation).
+
+        // conversationsPath ??= $"/{agentName}/v1/conversations";
+        // var conversationsRouteGroup = endpoints.MapGroup(conversationsPath);
+        // MapConversations(conversationsRouteGroup, agent, loggerFactory);
```
stephentoub, line 52: Is this because you're relying on the types defined in OpenAI? We don't want to just define our own DTOs, at least for now, if not for the medium to long term?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Internal/AIAgentResponsesProcessor.cs
```diff
@@ -0,0 +1,184 @@
...
+namespace Microsoft.Agents.AI.Hosting.OpenAI.Responses.Internal;
+
+/// <summary>
+/// OpenAI Responses processor associated with a specific <see cref="AIAgent"/>.
+/// </summary>
+internal sealed class AIAgentResponsesProcessor
+{
+#pragma warning disable IDE0052 // Remove unread private members
+    private readonly ILogger _logger;
+#pragma warning restore IDE0052 // Remove unread private members
```
stephentoub: Why do we need it if it's unused?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Internal/AIAgentResponsesProcessor.cs
```diff
@@ -0,0 +1,184 @@
...
+    public AIAgentResponsesProcessor(AIAgent agent, ILoggerFactory? loggerFactory = null)
+    {
+        this._logger = (loggerFactory ?? NullLoggerFactory.Instance).CreateLogger<AIAgentResponsesProcessor>();
+        this._agent = agent ?? throw new ArgumentNullException(nameof(agent));
+    }
+
+    public async Task<IResult> CreateModelResponseAsync(ResponseCreationOptions responseCreationOptions, CancellationToken cancellationToken)
+    {
+        var options = new OpenAIResponsesRunOptions();
+        AgentThread? agentThread = null!; // not supported to resolve from conversationId
```
stephentoub:
```suggestion
        AgentThread? agentThread = null; // not supported to resolve from conversationId
```

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Internal/AIAgentResponsesProcessor.cs
```diff
@@ -0,0 +1,184 @@
...
+            return new OpenAIStreamingResponsesResult(this._agent, chatMessages);
+        }
+
+        var agentResponse = await this._agent.RunAsync(chatMessages, agentThread, options, cancellationToken).ConfigureAwait(false);
+        return new OpenAIResponseResult(agentResponse);
+    }
+
+    private sealed class OpenAIResponseResult(AgentRunResponse agentResponse) : IResult
+    {
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Reliability", "CA2007:Consider calling ConfigureAwait on the awaited task", Justification = "Otherwise reports on await using var writer.")]
```
stephentoub: Do we even need/want to use using there?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Internal/AIAgentResponsesProcessor.cs
```diff
@@ -0,0 +1,184 @@
...
+            AgentThread? agentThread = null!;
+
+            ResponseItem? lastResponseItem = null;
+            OpenAIResponse? lastOpenAIResponse = null;
+
+            await foreach (var update in agent.RunStreamingAsync(chatMessages, thread: agentThread, cancellationToken: cancellationToken).ConfigureAwait(false))
+            {
+                if (string.IsNullOrEmpty(update.ResponseId)
+                    && string.IsNullOrEmpty(update.MessageId)
+                    && (update.Contents is null || update.Contents.Count == 0))
```
stephentoub: Nit:
```suggestion
                    && (update.Contents is not { Count: > 0 })
```

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Internal/AIAgentResponsesProcessor.cs
```diff
@@ -0,0 +1,184 @@
...
+                    lastOpenAIResponse = update.AsChatResponse().AsOpenAIResponse();
+
+                    var responseCreated = new StreamingCreatedResponse(sequenceNumber++)
+                    {
+                        Response = lastOpenAIResponse
+                    };
+                    yield return new(responseCreated, responseCreated.Type);
+                }
+
+                if (update.Contents is null || update.Contents.Count == 0)
```
stephentoub: Nit:
```suggestion
                if (update.Contents is not { Count: > 0 })
```

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Internal/AIAgentResponsesProcessor.cs
```diff
@@ -0,0 +1,184 @@
...
+
+                // to help convert the AIContent into OpenAI ResponseItem we pack it into the known "chatMessage"
+                // and use existing convertion extension method
+                var chatMessage = new ChatMessage(ChatRole.Assistant, update.Contents)
+                {
+                    MessageId = update.MessageId,
+                    CreatedAt = update.CreatedAt,
+                    RawRepresentation = update.RawRepresentation
+                };
+                var openAIResponseItem = MicrosoftExtensionsAIResponsesExtensions.AsOpenAIResponseItems([chatMessage]).FirstOrDefault();
```
stephentoub: A single ChatMessage can map to any number of ResponseItems. Rather than using FirstOrDefault, presumably this should be turned into a loop.

.github/workflows/dotnet-format.yml
DeagleGross: reported one to dotnet/sdk for tracking visibility: https://github.com/dotnet/sdk/issues/51136

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/EndpointRouteBuilderExtensions.cs
DeagleGross: added

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Internal/AIAgentResponsesProcessor.cs
DeagleGross: removed

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Internal/AIAgentResponsesProcessor.cs
DeagleGross: removed `await using` completely.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Internal/AIAgentResponsesProcessor.cs
DeagleGross: presented that as a loop

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/EndpointRouteBuilderExtensions.cs
DeagleGross: We dont have an API to manipulate with the conversation as a standalone entity in AF for the moment. For example if request is being sent with `conversationId: "test123"`, then we dont have a nice method like `agent.LoadThread(conversationId: threadId)` or smth similar.

We have discussed that with @ReubenBond , and we understand the necessity of adding one. Once such API will be designed and implemented, we can start using that from both A2A and OpenAI Responses Hosting packages.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/EndpointRouteBuilderExtensions.cs
DeagleGross: This is only used in route by default, and if it does not work (for any reason), then users can redefine the value here as they want it. Should we even place any constraints here for now? I would merge like that, and if given any feedback we can change later?

stephentoub: > and if it does not work (for any reason)

What is the failure mode if it doesn't work?

I'm just wondering about the end-to-end. When agents are added into DI, is the name inferred from AIAgent.Name? Because this then implies that there are constraints on what AIAgent.Name can be, that it must be valid in a query string, right?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Internal/AIAgentResponsesProcessor.cs
```diff
@@ -0,0 +1,179 @@
...
+                var chatMessage = new ChatMessage(ChatRole.Assistant, update.Contents)
+                {
+                    MessageId = update.MessageId,
+                    CreatedAt = update.CreatedAt,
+                    RawRepresentation = update.RawRepresentation
+                };
+
+                foreach (var openAIResponseItem in MicrosoftExtensionsAIResponsesExtensions.AsOpenAIResponseItems([chatMessage]))
+                {
+                    lastResponseItem ??= openAIResponseItem;
```
adityamandaleeka: Is this intentionally a `??=`? i.e. we only set it for the first one?

dotnet/Directory.Packages.props
```diff
@@ -36,7 +36,7 @@
     <PackageVersion Include="Swashbuckle.AspNetCore.SwaggerUI" Version="9.0.4" />
     <!-- System.* -->
     <PackageVersion Include="System.Linq.Async" Version="6.0.3" />
-    <PackageVersion Include="System.Net.ServerSentEvents" Version="9.0.9" />
+    <PackageVersion Include="System.Net.ServerSentEvents" Version="10.0.0-rc.1.25451.107" />
```
SergeyMenshykh: Is this update required? AF uses VersionOverride to override the common package version for projects that need a different version of a NuGet package.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Internal/AIAgentResponsesProcessor.cs
DeagleGross: no, its not. Fixed it by selecting last always, thanks for spotting

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Microsoft.Agents.AI.Hosting.OpenAI.csproj
```diff
@@ -0,0 +1,36 @@
...
+  </PropertyGroup>
+
+  <Import Project="$(RepoRoot)/dotnet/nuget/nuget-package.props" />
+  <PropertyGroup>
+    <InjectSharedThrow>true</InjectSharedThrow>
+  </PropertyGroup>
+  
+
+    <ItemGroup>
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
```
SergeyMenshykh, line 20: This project references AspNet.Core but does not have the corresponding suffix in its name as, for example, Microsoft.Agents.AI.Hosting.A2A.AspNetCore has. Is the plan not to use the prefix anymore for new projects?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Internal/AIAgentResponsesProcessor.cs
```diff
@@ -0,0 +1,179 @@
...
+using System.Threading.Tasks;
+using Microsoft.Agents.AI;
+using Microsoft.Agents.AI.Hosting.OpenAI.Responses.Model;
+using Microsoft.Agents.AI.Hosting.OpenAI.Responses.Utils;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Features;
+using Microsoft.Extensions.AI;
+using OpenAI.Responses;
+
+namespace Microsoft.Agents.AI.Hosting.OpenAI.Responses.Internal;
```
SergeyMenshykh: Is the `Internal` section in the namespace necessary taking into account that the class is internal as well?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/EndpointRouteBuilderExtensions.cs
DeagleGross: "name" which is passed as a string to the `AddAIAgent` extensions is the value which will be in `AIAgent.Name` property later. 

In order to make sure `agent.Name` could be correctly used, added a validation check that `Uri.EscapeDataString` is the same as `agentName`. Otherwise will throw `ArgumentException` clearly stating that `agent.Name` is not compatible for usage in the uri.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/AIAgentResponsesProcessor.cs
```diff
@@ -0,0 +1,184 @@
...
+                    lastOpenAIResponse = update.AsChatResponse().AsOpenAIResponse();
+
+                    var responseCreated = new StreamingCreatedResponse(sequenceNumber++)
+                    {
+                        Response = lastOpenAIResponse
+                    };
+                    yield return new(responseCreated, responseCreated.Type);
+                }
+
+                if (update.Contents is not { Count: > 0 })
```
SergeyMenshykh, line 129: Do we need to exclude updates with no content? Shouldn't we just return what the agent returns? Otherwise, there'll be a discrepancy between what the agent returns when used directly and via the hosting responses API.

dotnet/Directory.Packages.props
DeagleGross: required; placed it as override for OpenAI package only

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Internal/AIAgentResponsesProcessor.cs
DeagleGross: Removed internal section completely, moved into `OpenAI.Responses`

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/AIAgentResponsesProcessor.cs
DeagleGross: Responses API streaming output has a clear set of types that implementer should return (https://platform.openai.com/docs/api-reference/responses-streaming/response/created?lang=curl).

The problem here is that if there is no content, then we cant return `response.output_item.added` here for example, because there is no item to be added.

We are not pursuing a goal of identical return of what `AIAgent` returns, because we have to represent the response in OpenAI Responses format.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Microsoft.Agents.AI.Hosting.OpenAI.csproj
DeagleGross: I think going forward `Hosting` should be used only for the hosting in aspnetcore scenarios; A2A is kinda special due to also having an underlying packages `A2A` and `A2A.AspNetCore` which make such a distinction.

PR #959, opened 2025-09-27
.NET: Workflow observability

TaoChenOSU:
### Motivation and Context

.Net Workflow also needs observability for the AI Toolkit VS Code extension.

### Description

1. Add tracing to Workflows in .Net.
2. Add a new sample.
image

### Contribution Checklist

### DIFF
--- a/dotnet/samples/GettingStarted/Workflows/Observability/AspireDashboard/Program.cs
+++ b/dotnet/samples/GettingStarted/Workflows/Observability/AspireDashboard/Program.cs
@@ -55,9 +55,9 @@ public static class Program
         ReverseTextExecutor reverse = new();
 
         // Build the workflow by connecting executors sequentially
-        WorkflowBuilder builder = new(uppercase);
-        builder.AddEdge(uppercase, reverse).WithOutputFrom(reverse);
-        var workflow = builder.Build();
+        var workflow = new WorkflowBuilder(uppercase)
+            .AddEdge(uppercase, reverse)
+            .Build();
 
         // Execute the workflow with input data
         Run run = await InProcessExecution.RunAsync(workflow, "Hello, World!");
@@ -97,9 +97,5 @@ internal sealed class ReverseTextExecutor() : ReflectingExecutor<ReverseTextExec
     /// <param name="message">The input text to reverse</param>
     /// <param name="context">Workflow context for accessing workflow services and adding events</param>
     /// <returns>The input text reversed</returns>
-    public async ValueTask<string> HandleAsync(string message, IWorkflowContext context)
-    {
-        // Because we do not suppress it, the returned result will be yielded as an output from this executor.
-        return string.Concat(message.Reverse());
-    }
+    public async ValueTask<string> HandleAsync(string message, IWorkflowContext context) => new string(message.Reverse().ToArray());
 }
--- a/dotnet/src/Microsoft.Agents.AI.Workflows/InProc/InProcessRunnerContext.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Workflows/InProc/InProcessRunnerContext.cs
@@ -199,6 +199,7 @@ internal sealed class InProcessRunnerContext : IRunnerContext
         Dictionary<string, string>? traceContext) : IWorkflowContext
     {
         public ValueTask AddEventAsync(WorkflowEvent workflowEvent) => RunnerContext.AddEventAsync(workflowEvent);
+
         public ValueTask SendMessageAsync(object message, string? targetId = null)
         {
             return RunnerContext.SendMessageAsync(ExecutorId, message, targetId);
--- a/dotnet/src/Microsoft.Agents.AI.Workflows/StreamingRun.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Workflows/StreamingRun.cs
@@ -117,13 +117,13 @@ public sealed class StreamingRun
                 {
                     await this._stepRunner.RunSuperStepAsync(cancellationToken).ConfigureAwait(false);
                 }
-                catch (Exception ex)
+                catch (Exception ex) when (activity is not null)
                 {
-                    activity?.AddEvent(new ActivityEvent(EventNames.WorkflowError, tags: new() {
+                    activity.AddEvent(new ActivityEvent(EventNames.WorkflowError, tags: new() {
                         { Tags.ErrorType, ex.GetType().FullName },
                         { Tags.BuildErrorMessage, ex.Message },
                     }));
-                    activity?.CaptureException(ex);
+                    activity.CaptureException(ex);
                     throw;
                 }
 
--- a/dotnet/src/Microsoft.Agents.AI.Workflows/WorkflowBuilder.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Workflows/WorkflowBuilder.cs
@@ -345,13 +345,13 @@ public class WorkflowBuilder
         {
             this.Validate();
         }
-        catch (Exception ex)
+        catch (Exception ex) when (activity is not null)
         {
-            activity?.AddEvent(new ActivityEvent(EventNames.BuildError, tags: new() {
+            activity.AddEvent(new ActivityEvent(EventNames.BuildError, tags: new() {
                 { Tags.BuildErrorMessage, ex.Message },
                 { Tags.BuildErrorType, ex.GetType().FullName }
             }));
-            activity?.CaptureException(ex);
+            activity.CaptureException(ex);
             throw;
         }
 


dotnet/src/Microsoft.Agents.Workflows/Execution/FanInEdgeRunner.cs
```diff
@@ -16,10 +17,24 @@ internal sealed class FanInEdgeRunner(IRunnerContext runContext, FanInEdgeData e
...
     protected internal override async ValueTask<DeliveryMapping?> ChaseEdgeAsync(MessageEnvelope envelope, IStepTracer? stepTracer)
     {
-        Debug.Assert(!envelope.IsExternal, "FanIn edges should never be chased from external input");
+        using var activity = s_activitySource.StartActivity(ActivityNames.EdgeGroupProcess);
+        activity?
+            .SetTag(Tags.EdgeGroupType, nameof(FanInEdgeRunner))
+            .SetTag(Tags.MessageTargetId, this.EdgeData.SinkId);
+
+        try
+        {
```
stephentoub: What is the purpose of this try/catch? The catch block is basically never going to be hit, as in tense the try will be empty and in debug the asset will fail fast.

dotnet/src/Microsoft.Agents.Workflows/Execution/FanInEdgeRunner.cs
```diff
@@ -28,14 +43,29 @@ internal sealed class FanInEdgeRunner(IRunnerContext runContext, FanInEdgeData e
...
 
-        // TODO: Filter messages based on accepted input types?
-        Executor target = await this.RunContext.EnsureExecutorAsync(this.EdgeData.SinkId, stepTracer)
-                                               .ConfigureAwait(false);
+        try
+        {
+            // TODO: Filter messages based on accepted input types?
+            Executor target = await this.RunContext.EnsureExecutorAsync(this.EdgeData.SinkId, stepTracer)
+                                                   .ConfigureAwait(false);
+            var finalReleasedMessages = releasedMessages.Where(envelope => target.CanHandle(envelope.MessageType));
```
stephentoub: If the final Released Messages here is a lazy enumerator, this is going to end up doing multiple additional iterations through it.

dotnet/samples/GettingStarted/Workflows/Observability/AspireDashboard/Program.cs
```diff
@@ -0,0 +1,104 @@
...
+        // Configure OpenTelemetry for Aspire dashboard
+        var otlpEndpoint = Environment.GetEnvironmentVariable("OTLP_ENDPOINT") ?? "http://localhost:4317";
+
+        var resourceBuilder = ResourceBuilder
+            .CreateDefault()
+            .AddService("WorkflowSample");
+
+        using var traceProvider = Sdk.CreateTracerProviderBuilder()
+            .SetResourceBuilder(resourceBuilder)
+            .AddSource("Microsoft.Agents.Workflows*")
```
TaoChenOSU: Activities/spans won't be created if there are no active listeners. That means we don't need a switch to turn on the tracing feature. However, we do use a switch in `Microsoft.Extensions.AI`: https://github.com/microsoft/agent-framework/blob/main/dotnet/samples/GettingStarted/AgentOpenTelemetry/Program.cs#L24. 

Do we want to keep things consistent?

stephentoub:
I'm having trouble envisioning what the resulting traces/spans are going to look like, especially with the graph based nature. Are these spans going to end up being arbitrarily deeply nested, eg if there's a loop in the graph entering a new nested span for each iteration? Do you have some screenshots you could share?

stephentoub:
cc: @JamesNK

TaoChenOSU:
> I'm having trouble envisioning what the resulting traces/spans are going to look like, especially with the graph based nature. Are these spans going to end up being arbitrarily deeply nested, eg if there's a loop in the graph entering a new nested span for each iteration? Do you have some screenshots you could share?

Just uploaded one to the PR summary from running the new sample introduced in this PR. In short, the executor spans won't be deeply nested. They all share the same parent span which is the `workflow.run` span.

dotnet/samples/GettingStarted/Workflows/Observability/AspireDashboard/Program.cs
TaoChenOSU: This is a reminder for myself. Reviewers are welcome to provide their opinions.

JamesNK:
@TaoChenOSU Aspire 9.5 has just been released with improved GenAI telemetry support. It follows a lot of what the VS extension and AI foundry website do when it comes to displaying AI.

https://learn.microsoft.com/en-us/dotnet/aspire/fundamentals/dashboard/explore#genai-telemetry-visualization

It would be great if you could try out telemetry in Aspire dashboard alongside VS extension tooling. Bugs, suggestions, and feedback such as comparisons to other tools welcome at https://github.com/dotnet/aspire

dotnet/samples/GettingStarted/Workflows/Observability/AspireDashboard/Program.cs
stephentoub: I don't see why we'd need an enable switch. We don't have such a thing for Microsoft.Extensions.AI.OpenTelemetryChatClient and friends.

dotnet/samples/GettingStarted/Workflows/Observability/AspireDashboard/Program.cs
```diff
@@ -0,0 +1,104 @@
...
+{
+    /// <summary>
+    /// Processes the input message by converting it to uppercase.
+    /// </summary>
+    /// <param name="message">The input text to convert</param>
+    /// <param name="context">Workflow context for accessing workflow services and adding events</param>
+    /// <returns>The input text converted to uppercase</returns>
+    public async ValueTask<string> HandleAsync(string message, IWorkflowContext context) =>
+        message.ToUpperInvariant(); // The return value will be sent as a message along an edge to subsequent executors
+}
```
stephentoub, line 87: We seem to be making things more complicated by relying on ReflectingExecutor in our samples.

With reflection:
```csharp
internal sealed class UppercaseExecutor() : 
    ReflectingExecutor<UppercaseExecutor>("UppercaseExecutor"),
    IMessageHandler<string, string>
{
    public async ValueTask<string> HandleAsync(string message, IWorkflowContext context) =>
        message.ToUpperInvariant()
}
```

Without reflection:
```csharp
internal sealed class UppercaseExecutor() : 
    Executor("UppercaseExecutor")
{
        routeBuilder.AddHandler((string message, _) => message.ToUpperInvariant());
}
```

I would continue to prefer us just deleting ReflectingExecutor, IMessageHandler, and friends, and having just this single way of doing things, fewer concepts to learn, more consistency with a single approach that accomodates all cases, etc. If folks then want a separate HandleAsync method instead of a lambda, they can do that.

dotnet/src/Microsoft.Agents.Workflows/Execution/DirectEdgeRunner.cs
```diff
@@ -1,36 +1,53 @@
...
         }
-
-        Executor target = await this.FindRouterAsync(stepTracer).ConfigureAwait(false);
-        if (target.CanHandle(envelope.MessageType))
+        catch
         {
-            return new DeliveryMapping(envelope, target);
+            activity?.SetEdgeRunnerDeliveryStatus(EdgeRunnerDeliveryStatus.Exception);
+            throw;
         }
```
stephentoub: This can be:
```csharp
catch (Exception) when activity is not null
{
    activity.SetEdgeRunnerDeliveryStatus(EdgeRunnerDeliveryStatus.Exception);
    throw;
}
```
so that we don't even consider the catch block if activity is null.

dotnet/src/Microsoft.Agents.Workflows/Executor.cs
```diff
@@ -88,6 +92,12 @@ internal MessageRouter Router
     /// <exception cref="TargetInvocationException">An exception is generated while handling the message.</exception>
     public async ValueTask<object?> ExecuteAsync(object message, TypeId messageType, IWorkflowContext context)
     {
+        using var activity = s_activitySource.StartActivity(ActivityNames.ExecutorProcess, ActivityKind.Internal);
+        activity?.SetTag(Tags.ExecutorId, this.Id)
+            .SetTag(Tags.ExecutorType, this.GetType().FullName)
+            .SetTag(Tags.MessageType, messageType.TypeName);
+        activity?.CreateSourceLinks(context.TraceContext);
```
stephentoub: Nit: seems like this could be chained as well

dotnet/src/Microsoft.Agents.AI.Workflows/IWorkflowContext.cs
```diff
@@ -92,4 +92,9 @@ public interface IWorkflowContext
     /// <param name="scopeName">An optional name that specifies the scope to clear. If null, the default scope is used.</param>
     /// <returns>A ValueTask that represents the asynchronous clear operation.</returns>
     ValueTask QueueClearScopeAsync(string? scopeName = null);
+
+    /// <summary>
+    /// The trace context associated with the current message about to be processed by the executor, if any.
+    /// </summary>
+    IReadOnlyDictionary<string, string>? TraceContext { get; }
```
stephentoub, line 99: What does "trace context" mean?

dotnet/src/Microsoft.Agents.Workflows/InProc/InProcessRunnerContext.cs
```diff
@@ -132,10 +136,25 @@ public ValueTask AddEventAsync(WorkflowEvent workflowEvent)
...
 
+    private static readonly string s_namespace = typeof(IWorkflowContext).Namespace!;
+    private static readonly ActivitySource s_activitySource = new(s_namespace);
+
     public async ValueTask SendMessageAsync(string sourceId, object message, string? targetId = null)
     {
+        using Activity? activity = s_activitySource.StartActivity(ActivityNames.MessageSend, ActivityKind.Producer);
+        // Create a carrier for trace context propagation
+        var traceContext = activity is null ? null : new Dictionary<string, string>();
+        if (activity != null && traceContext != null)
```
stephentoub: Seems like the activity != null check is duplicative

dotnet/src/Microsoft.Agents.AI.Workflows/IWorkflowContext.cs
TaoChenOSU: The `TraceContext` dictionary is used for context propagation. The dictionary contains the trace ID and span ID of the current span. We need to manually propagate the context with the message envelope to create the causal relationship (by creating links) between the `message.send` spans and the `executor.process` spans because they are not nested.

dotnet/samples/GettingStarted/Workflows/Observability/AspireDashboard/Program.cs
TaoChenOSU: I think this would require bigger a change that will touch all the samples, and it's out of scope for this PR. 

We should probably consider the pattern change in another more targeted PR.

@lokitoth Do we have a GitHub issue to track the discussion?

dotnet/samples/GettingStarted/Workflows/Observability/AspireDashboard/Program.cs
TaoChenOSU: Interesting, that's not what the sample shows. Let me double check with Roger on the MEAI side.

dotnet/samples/GettingStarted/Workflows/Observability/AspireDashboard/Program.cs
crickman: The "without" example Stephen is sharing is just syntactic sugar to the explicit pattern.  I converted all the declarative executors w/o any issue.  (Agree it can be a follow-up PR.)

dotnet/samples/GettingStarted/Workflows/Observability/AspireDashboard/Program.cs
stephentoub: > that's not what the sample shows

What sample? There's a single occurrence of "enabletelemetry" in dotnet/extensions, and it's a parameter name in a test:
https://github.com/search?q=repo%3Adotnet%2Fextensions+EnableTelemetry+language%3AC%23&type=code&l=C%23

rogerbarreto: There's a typo/remainder in the samples, possibly from reusing the SK example, we don't have switching. Lets remove it.

TaoChenOSU: Ok, remove in this PR.

dotnet/src/Microsoft.Agents.AI.Workflows/Execution/FanInEdgeRunner.cs
```diff
@@ -28,14 +35,30 @@ internal sealed class FanInEdgeRunner(IRunnerContext runContext, FanInEdgeData e
...
+            activity?.SetEdgeRunnerDeliveryStatus(EdgeRunnerDeliveryStatus.Buffered);
             return null;
         }
 
-        // TODO: Filter messages based on accepted input types?
-        Executor target = await this.RunContext.EnsureExecutorAsync(this.EdgeData.SinkId, stepTracer)
-                                               .ConfigureAwait(false);
+        try
+        {
+            // TODO: Filter messages based on accepted input types?
```
alliscode, line 44: Do we need a work item to track this?

dotnet/samples/GettingStarted/Workflows/Observability/AspireDashboard/Program.cs
```diff
@@ -0,0 +1,104 @@
...
+            .AddSource(SourceName)
+            .AddOtlpExporter(options => options.Endpoint = new Uri(otlpEndpoint))
+            .Build();
+
+        // Create the executors
+        UppercaseExecutor uppercase = new();
+        ReverseTextExecutor reverse = new();
+
+        using var activity = s_activitySource.StartActivity("main");
+        Console.WriteLine($"Operation/Trace ID: {Activity.Current?.TraceId}");
```
alliscode: This part could use a comment as well.

dotnet/samples/GettingStarted/Workflows/Observability/AspireDashboard/Program.cs
```diff
@@ -0,0 +1,105 @@
...
+        // Configure OpenTelemetry for Aspire dashboard
+        var otlpEndpoint = Environment.GetEnvironmentVariable("OTLP_ENDPOINT") ?? "http://localhost:4317";
+
+        var resourceBuilder = ResourceBuilder
+            .CreateDefault()
+            .AddService("WorkflowSample");
+
+        using var traceProvider = Sdk.CreateTracerProviderBuilder()
+            .SetResourceBuilder(resourceBuilder)
+            .AddSource("Microsoft.Agents.AI.Workflows*")
```
stephentoub, line 44: Should we expand this to Microsoft.Agents.AI*?

dotnet/samples/GettingStarted/Workflows/Observability/AspireDashboard/Program.cs
```diff
@@ -0,0 +1,105 @@
...
+        Console.WriteLine($"Operation/Trace ID: {Activity.Current?.TraceId}");
+
+        // Create the executors
+        UppercaseExecutor uppercase = new();
+        ReverseTextExecutor reverse = new();
+
+        // Build the workflow by connecting executors sequentially
+        WorkflowBuilder builder = new(uppercase);
+        builder.AddEdge(uppercase, reverse).WithOutputFrom(reverse);
+        var workflow = builder.Build();
```
stephentoub: Nit: you could make this more fluent like:
```csharp
var workflow = new WorkflowBuilder(uppercase)
    .AddEdge(uppercase, reverse)
    .WithOutputFrom(reverse)
    .Build();
```

dotnet/samples/GettingStarted/Workflows/Observability/AspireDashboard/Program.cs
```diff
@@ -0,0 +1,105 @@
...
+
+        // Execute the workflow with input data
+        Run run = await InProcessExecution.RunAsync(workflow, "Hello, World!");
+        foreach (WorkflowEvent evt in run.NewEvents)
+        {
+            if (evt is ExecutorCompletedEvent executorComplete)
+            {
+                Console.WriteLine($"{executorComplete.ExecutorId}: {executorComplete.Data}");
+            }
+        }
```
stephentoub, line 70: Is this taking advantage of the WithOutputFrom? I'm wondering if there's a simpler "just print the output" this should be doing, for sample's sake.

dotnet/samples/GettingStarted/Workflows/Observability/AspireDashboard/Program.cs
```diff
@@ -0,0 +1,105 @@
...
+{
+    /// <summary>
+    /// Processes the input message by reversing the text.
+    /// </summary>
+    /// <param name="message">The input text to reverse</param>
+    /// <param name="context">Workflow context for accessing workflow services and adding events</param>
+    /// <returns>The input text reversed</returns>
+    public async ValueTask<string> HandleAsync(string message, IWorkflowContext context)
+    {
+        // Because we do not suppress it, the returned result will be yielded as an output from this executor.
```
stephentoub: "Because we do not suppress it"... aren't you explicitly asking for it, with WithOutputFrom?

dotnet/src/Microsoft.Agents.AI.Workflows/InProc/InProcessRunnerContext.cs
```diff
@@ -173,10 +192,17 @@ public bool CompleteRequest(string requestId)
...
-    private sealed class BoundContext(InProcessRunnerContext RunnerContext, string ExecutorId, OutputFilter outputFilter) : IWorkflowContext
+    private sealed class BoundContext(
+        InProcessRunnerContext RunnerContext,
+        string ExecutorId,
+        OutputFilter outputFilter,
+        Dictionary<string, string>? traceContext) : IWorkflowContext
     {
         public ValueTask AddEventAsync(WorkflowEvent workflowEvent) => RunnerContext.AddEventAsync(workflowEvent);
-        public ValueTask SendMessageAsync(object message, string? targetId = null) => RunnerContext.SendMessageAsync(ExecutorId, message, targetId);
+        public ValueTask SendMessageAsync(object message, string? targetId = null)
```
stephentoub, line 203: Nit:
```suggestion

        public ValueTask SendMessageAsync(object message, string? targetId = null)
```

dotnet/src/Microsoft.Agents.AI.Workflows/Microsoft.Agents.AI.Workflows.csproj
```diff
@@ -29,6 +29,10 @@
     <InternalsVisibleTo Include="Microsoft.Agents.AI.Workflows.UnitTests" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="OpenTelemetry.Api" />
```
stephentoub, line 33: Why does the workflows project need to reference OpenTelemetry.Api? Is that necessary and desirable?

dotnet/src/Microsoft.Agents.AI.Workflows/StreamingRun.cs
```diff
@@ -95,12 +100,33 @@ internal async IAsyncEnumerable<WorkflowEvent> WatchStreamAsync(
...
+                // is set to our activity for the duration of this loop iteration.
+                Activity.Current = activity;
+
                 // Drain SuperSteps while there are steps to run
-                await this._stepRunner.RunSuperStepAsync(cancellationToken).ConfigureAwait(false);
+                try
+                {
+                    await this._stepRunner.RunSuperStepAsync(cancellationToken).ConfigureAwait(false);
+                }
+                catch (Exception ex)
```
stephentoub: Another case where you can do `when activity is not null` and then remove the `?`s from `activity?` in the catch block

dotnet/src/Microsoft.Agents.AI.Workflows/WorkflowBuilder.cs
```diff
@@ -331,22 +337,71 @@ private void Validate()
...
     {
-        this.Validate();
+        activity?.AddEvent(new ActivityEvent(EventNames.BuildStarted));
 
-        return new Workflow(this._startExecutorId)
+        try
+        {
+            this.Validate();
+        }
+        catch (Exception ex)
```
stephentoub: Ditto... maybe just search to see if there are others

dotnet/src/Microsoft.Agents.AI.Workflows/WorkflowBuilder.cs
```diff
@@ -331,22 +337,71 @@ private void Validate()
...
+    /// </summary>
+    /// <exception cref="InvalidOperationException">Thrown if there are unbound executors in the workflow definition,
+    /// or if the start executor is not bound.</exception>
+    public Workflow Build()
+    {
+        using Activity? activity = s_activitySource.StartActivity(ActivityNames.WorkflowBuild);
+
+        var workflow = this.BuildInternal(activity);
+
+        activity?.AddEvent(new ActivityEvent(EventNames.BuildCompleted));
```
stephentoub, line 402: There are no details about the workflow we we want to include?

dotnet/samples/GettingStarted/Workflows/Observability/AspireDashboard/Program.cs
TaoChenOSU: There is no need since this project doesn't depend on agents.

dotnet/samples/GettingStarted/Workflows/Observability/AspireDashboard/Program.cs
TaoChenOSU: Yeah, remove `WithOutputFrom` for the simplicity of this sample.

dotnet/src/Microsoft.Agents.AI.Workflows/WorkflowBuilder.cs
TaoChenOSU: The Workflow definition is included in the Build Span as an attribute.

dotnet/src/Microsoft.Agents.AI.Workflows/WorkflowBuilder.cs
TaoChenOSU: This is the last one :) Thank you for pointing it all out

dotnet/src/Microsoft.Agents.AI.Workflows/Microsoft.Agents.AI.Workflows.csproj
TaoChenOSU: We need this because we need the OTel API to inject and extract trace context.

PR #967, opened 2025-09-28
.NET: Update OpenTelemetryAgent to latest spec

stephentoub:
It was stale. Rather than try to keep it up-to-date manually, I've changed it to piggy back on OpenTelemetryChatClient, so that it inherits everything OpenTelemetryChatClient does and then augments it just with agent-specific tags.

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI/OpenTelemetryAgent.cs
+++ b/dotnet/src/Microsoft.Agents.AI/OpenTelemetryAgent.cs
@@ -24,11 +24,7 @@ public sealed class OpenTelemetryAgent : DelegatingAIAgent, IDisposable
     // convention, then, and keep it up-to-date as the convention evolves, for now this implementation just delegates
     // to OpenTelemetryChatClient for the actual telemetry work. For RunAsync and RunStreamingAsync, it delegates to the
     // inner agent not directly but rather via OpenTelemetryChatClient, which wraps a ForwardingChatClient that in turn
-    // calls back into the inner agent. For state that we can't pass through directly (like AgentRunOptions), we pass that
-    // around the OpenTelemetryChatClient via an AsyncLocal.
-
-    /// <summary><see cref="AsyncLocal{ForwardedState}"/> that's used to pass input from this agent into the inner agent.</summary>
-    private static readonly AsyncLocal<ForwardedState?> s_forwardedState = new();
+    // calls back into the inner agent.
 
     /// <summary>The <see cref="OpenTelemetryChatClient"/> providing the bulk of the telemetry.</summary>
     private readonly OpenTelemetryChatClient _otelClient;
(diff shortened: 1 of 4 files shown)


dotnet/src/Microsoft.Agents.AI/AgentExtensions.cs
```diff
@@ -18,15 +17,20 @@ public static class AgentExtensions
...
-        new(agent, loggerFactory?.CreateLogger(typeof(OpenTelemetryAgent)), sourceName)
+    public static OpenTelemetryAgent WithOpenTelemetry(this AIAgent agent, string? sourceName = null, bool? enableSensitiveData = null)
+    {
+        OpenTelemetryAgent otel = new(agent, sourceName);
+
+        if (enableSensitiveData is not null)
         {
-            EnableSensitiveData = enableSensitiveData ?? false
-        };
+            otel.EnableSensitiveData = enableSensitiveData.Value;
```
JamesNK, line 29:
```suggestion
            // Only override the OpenTelemetryChatClient.EnableSensitiveData default (which can be enabled
            // via an environment variable) if there is an explicit value specified.
            otel.EnableSensitiveData = enableSensitiveData.Value;
```

dotnet/src/Microsoft.Agents.AI/OpenTelemetryAgent.cs
```diff
@@ -3,532 +3,199 @@
...
+    // metrics and the chat span. But the chat span is almost identical to the invoke_agent span, just with invoke_agent
+    // have a different value for the operation name and a few additional tags. To avoid needing to reimplement the
+    // convention, then, and keep it up-to-date as the convention evolves, for now this implementation just delegates
+    // to OpenTelemetryChatClient for the actual telemetry work. For RunAsync and RunStreamingAsync, it delegates to the
+    // inner agent not directly but rather via OpenTelemetryChatClient, which wraps a ForwardingChatClient that in turn
+    // calls back into the inner agent. For state that we can't pass through directly (like AgentRunOptions), we pass that
+    // around the OpenTelemetryChatClient via an AsyncLocal.
+
+    /// <summary><see cref="AsyncLocal{ForwardedState}"/> that's used to pass input from this agent into the inner agent.</summary>
+    private static readonly AsyncLocal<ForwardedState?> s_forwardedState = new();
```
rogerbarreto: Instead of async local + forward state what are the downsides compared in using a 

`private class ForwardChatOptions : ChatOptions` for that purpose?

We could just provide this state in extra fields in the `Formard` class and capture within the `ForwardedChatClient`. 

Thoughts?

stephentoub: Do you mean creating a new ForwardChatClient and OpenTelemetryChatClient per request?

stephentoub: Oh oh, you said ForwardChatOptions, not ForwardChatClient. We probably could.

stephentoub: Hmm. While that avoids the async local, it means that all properties from the ChatClientAgentRunOptions.ChatOptions must be copied one-by-one over to the new ForwardChatOptions instance... which means the moment M.E.AI.ChatOptions gets another property, OpenTelemetryAgent will need to be updated to propagate it, or else that information won't be logged.

We can look into copy constructors or other ways of simplifying this on ChatOptions, but this doesn't seem like the right tradeoff right now.

rogerbarreto: > We can look into copy constructors or other ways of simplifying this on ChatOptions, but this doesn't seem like the right tradeoff right now.

This seems quite strict, I don't fully agree as enough strong argument to not use the ChatOptions specializations.

I understand that new properties in the `ChatOptions` will not happen oftenly, and from those to be OTEL possibility is quite lower as well...

stephentoub: I updated the code to use a derived ChatOptions. It is not maintainable and I think it's strictly worse than the previous version, but I changed it anyway in order to make forward progress. Please re-review.

PR #970, opened 2025-09-29
.NET: Update the package version and suffix in preparation for public preview

markwallace-microsoft:
### Contribution Checklist

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
+++ b/dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
@@ -35,7 +35,8 @@ public sealed class ChatClientAgent : AIAgent
     /// <param name="description">Optional description for the agent.</param>
     /// <param name="tools">Optional list of tools that the agent can use during invocation.</param>
     /// <param name="loggerFactory">Optional logger factory to use for logging.</param>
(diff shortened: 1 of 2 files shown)


dotnet/nuget/nuget-package.props
```diff
@@ -1,8 +1,8 @@
 <Project>
   <PropertyGroup>
     <!-- Central version prefix - applies to all nuget packages. -->
-    <VersionPrefix>0.0.1</VersionPrefix>
-    <PackageVersion Condition="'$(VersionSuffix)' != ''">$(VersionPrefix)-$(VersionSuffix)</PackageVersion>
+    <VersionPrefix>1.0.0</VersionPrefix>
+    <PackageVersion Condition="'$(VersionSuffix)' != ''">$(VersionPrefix)-$(VersionSuffix).250929</PackageVersion>
```
rogerbarreto, line 5: Are we planning in starting with a 1.0.0-preview straight away?

This may not allow us to use minor/patch versions until we have the 1.0 GA, enforcing `YYMMDD` suffix after preview.

1.0.0-preview.250929
1.0.0-preview.251001
1.0.0-preview.251010
1.0.0-preview.251020

If we plan using semantic minor / patches after public preview we should stay at v0.*

markwallace-microsoft: Yes, the ask is that we use 1.0.0-preview versioning. Provided more details via teams.

PR #981, opened 2025-09-29
.NET: Fix visual output of basic Agents (Console.Write vs Console.WriteLine)

rwjdk:
### Motivation and Context

1. Why is this change required? : **The samples are more correct**
2. What problem does it solve? : **That the visual output of streaming responses are shown incorrectly to the user**
3. What scenario does it contribute to: **Basic Agent Samples**
4. If it fixes an open issue, please link to the issue here.: **None**

### Description

Simple changes of Console.WriteLine to Console.Write (+ a few incorrect tabs vs spaces that did not follow Visual Studio Default)

### Contribution Checklist

### DIFF
--- a/dotnet/samples/GettingStarted/Agents/Agent_Step01_Running/Program.cs
+++ b/dotnet/samples/GettingStarted/Agents/Agent_Step01_Running/Program.cs
@@ -15,10 +15,10 @@ const string JokerName = "Joker";
 const string JokerInstructions = "You are good at telling jokes.";
 
 AIAgent agent = new AzureOpenAIClient(
-    new Uri(endpoint),
-    new AzureCliCredential())
-     .GetChatClient(deploymentName)
-     .CreateAIAgent(JokerInstructions, JokerName);
+        new Uri(endpoint),
+        new AzureCliCredential())
+    .GetChatClient(deploymentName)
+    .CreateAIAgent(JokerInstructions, JokerName);
 
 // Invoke the agent and output the text result.
 Console.WriteLine(await agent.RunAsync("Tell me a joke about a pirate."));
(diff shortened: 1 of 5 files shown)


dotnet/samples/GettingStarted/Agents/Agent_Step01_Running/Program.cs
```diff
@@ -15,16 +15,16 @@
...
+    .CreateAIAgent(JokerInstructions, JokerName);
 
 // Invoke the agent and output the text result.
 Console.WriteLine(await agent.RunAsync("Tell me a joke about a pirate."));
 
 // Invoke the agent with streaming support.
 await foreach (var update in agent.RunStreamingAsync("Tell me a joke about a pirate."))
 {
-    Console.WriteLine(update);
+    Console.Write(update);
```
markwallace-microsoft, line 29: We are using `Console.WriteLine` deliberately so you can see the streaming nature of the response i.e. you can see each update that the model returns on a seperate line.

markwallace-microsoft:
No plans to merge this

rwjdk:
@markwallace-microsoft OK. Should there perhaps be a comment then that explains it is intentional (can't imagine I will be the last one to send you such a PR, and a comment will prevent that)

dotnet/samples/GettingStarted/Agents/Agent_Step01_Running/Program.cs
stephentoub: > you can see each update that the model returns on a separate line.

Why is it important that each update be on a separate line? Typically in these demos you just see the words get appended to the end of the current line.

PR #986, opened 2025-09-29
Python: use run_tasks_in_packages_if_exists script for all-tests

ekzhu:
### Motivation and Context

So we can properly skip tests in lab and devui in the main CI build

### Contribution Checklist


python/pyproject.toml
```diff
@@ -201,21 +202,20 @@ pytest --import-mode=importlib
...
 
 [tool.poe.tasks.all-tests]
-cmd = """
-pytest --import-mode=importlib
---ignore-glob=packages/lab/**
---ignore-glob=packages/devui/**
--n logical --dist loadfile --dist worksteal
-packages/**/tests
-"""
+cmd = "python run_tasks_in_packages_if_exists.py test --exclude packages/lab --exclude packages/devui"
```
alliscode, line 247: Could we make sure we have a work item telling us to turn these tests back on once devui is stable?

alliscode (approved):

python/pyproject.toml
ekzhu: https://github.com/microsoft/agent-framework/issues/988 Yes

python/pyproject.toml
```diff
@@ -209,13 +209,7 @@ packages/**/tests
...
```
eavanvalkenburg, line 247: This is much slower because it will have to setup and breakdown for each package

eavanvalkenburg (changes requested):

PR #991, opened 2025-09-30
.NET Workflows - Fix "Human In Loop" sample and update "AutoSend" behavior

crickman:
### Motivation and Context

1. Fix human in the loop sample
2. Modify "AutoSend" behavior to copy message to workflow thread
3. Specify converstation is in chat options

### Description

1. Need to provide runId when restarting workflow
2. When agent response is not on the workflow conversation, autosend raises workflow events and copies message to workflow conversation.
3. No need to directly handle thread instance

### Contribution Checklist


dotnet/samples/GettingStarted/Workflows/Declarative/Program.cs
```diff
@@ -88,7 +88,7 @@ private async Task ExecuteAsync()
                 // Restore the latest checkpoint.
                 Debug.WriteLine($"RESTORE #{this.LastCheckpoint.CheckpointId}");
                 Notify("\nWORKFLOW: Restore");
-                run = await InProcessExecution.ResumeStreamAsync(workflow, this.LastCheckpoint, checkpointManager);
+                run = await InProcessExecution.ResumeStreamAsync(workflow, this.LastCheckpoint, checkpointManager, run.Run.RunId);
```
RaviPidaparthi, line 91: So, it wont work if you dont keep runId the same across multiple executions?

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Extensions/AgentProviderExtensions.cs
```diff
@@ -20,35 +18,61 @@ public static async IAsyncEnumerable<AgentRunResponseUpdate> InvokeAgentAsync(
...
+        if (autoSend)
+        {
+            await context.AddEventAsync(new AgentRunResponseEvent(executorId, response)).ConfigureAwait(false);
         }
 
+        if (autoSend && !isWorkflowConversation && conversationId is not null)
+        {
+            foreach (ChatMessage message in response.Messages)
+            {
+                await agentProvider.CreateMessageAsync(conversationId, message, cancellationToken).ConfigureAwait(false);
```
RaviPidaparthi, line 78: It's not this simple. We only want to copy items of type assistant messages. 
In V2, response items will have a lot of types like messages, tool calls, reasoning etcc. . The desire is only to copy assistant messages.

dotnet/samples/GettingStarted/Workflows/Declarative/Program.cs
crickman: Apparently.  I remember Jacob mentioned something about this.  We can follow-up with him also.

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Extensions/AgentProviderExtensions.cs
crickman: Updated.

alliscode (approved):

TaoChenOSU (approved):

PR #1006, opened 2025-09-30
.NET: Remove automatic messageId assignment to fix integration test

westey-m:
### Motivation and Context

Assigning a message id to updates that do not have them in ChatClientAgent, resulted in empty messages being produced when combining updates into a single response.

### Description

- Remove automatic messageId assignment to fix integration test

### Contribution Checklist


stephentoub:
Sounds wet also remove the setting of I'd in non-streaming, for consistency?
https://github.com/westey-m/agent-framework/blob/a364d25eda3197bd369ade83d45c269fc70643cf/dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs#L148

What about the created at time? That won't affect current coalescing logic, but maybe we shouldn't be manufacturing that either?

westey-m:
> Sounds wet also remove the setting of I'd in non-streaming, for consistency? https://github.com/westey-m/agent-framework/blob/a364d25eda3197bd369ade83d45c269fc70643cf/dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs#L148
> 
> What about the created at time? That won't affect current coalescing logic, but maybe we shouldn't be manufacturing that either?

Good point.  Removed those.  I won't merge this until we have agreement with the workflow folks.  I've got a discussion about the implications on the workflow side ongoing.

PR #1008, opened 2025-09-30
.NET: fix sample app startup bugs

DeagleGross:
Fixing:
1) accidentally introduced recursion while #915 
2) the AIAgent registration via `AsAgentAsync()`
`AgentWorkflowBuilder.BuildConcurrent([knight, knave, narrator]).AsAgentAsync(name: key).AsTask().GetAwaiter().GetResult();`
otherwise `agent.Name` property will not be populated

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.Hosting.A2A.AspNetCore/WebApplicationExtensions.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Hosting.A2A.AspNetCore/WebApplicationExtensions.cs
@@ -63,7 +63,7 @@ public static class WebApplicationExtensions
         // note: current SDK version registers multiple `.well-known/agent.json` handlers here.
         // it makes app return HTTP 500, but will be fixed once new A2A SDK is released.
         // see https://github.com/microsoft/agent-framework/issues/476 for details
-        app.MapA2A(taskManager, path);
+        A2ARouteBuilderExtensions.MapA2A(app, taskManager, path);
 
         app.MapHttpA2A(taskManager, path);
     }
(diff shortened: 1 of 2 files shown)


dotnet/src/Microsoft.Agents.AI.Hosting.A2A.AspNetCore/WebApplicationExtensions.cs
```diff
@@ -63,7 +63,7 @@ public static void MapA2A(this WebApplication app, TaskManager taskManager, stri
         // note: current SDK version registers multiple `.well-known/agent.json` handlers here.
         // it makes app return HTTP 500, but will be fixed once new A2A SDK is released.
         // see https://github.com/microsoft/agent-framework/issues/476 for details
-        app.MapA2A(taskManager, path);
+        A2ARouteBuilderExtensions.MapA2A(app, taskManager, path);
```
stephentoub, line 66: What was wrong here?

stephentoub: Oh, I see, it was calling itself

DeagleGross: `app.MapA2A()` calls into itself (Microsoft.Agents.AI.Hosting.A2A.AspNetCore.WebApplicationExtensions.MapA2A()).

`A2ARouteBuilderExtensions` is a type defined in separate package `A2A.AspNetCore`, and we want to call that one. I can write `globall::` to be more explicit

Issue #923, opened 2025-09-25
.NET: Equivalent functionality to kernel.Plugins.AddFromType


markwallace-microsoft:
The transition from Semantic Kernel to the Agent Framework was relatively straightforward using the following syntax:

``` csharp
AIFunctionFactory.Create(
    (string prompt) => { // do something }, 
    nameOfTool,
    someDescription
);
```

However, I could not find an equivalent, streamlined way to create a tool that is registered in our service collection. With Semantic Kernel, it was easier by simply adding the tool into plugins and passing the service provider.  
With Sematic kernel:

```csharp
kernel.Plugins.AddFromType<TransferCallAgentPlugin>(nameof(TransferCallAgentPlugin), _serviceProvider);
```

For now, this is the approach I used:  
- Retrieve the instance of the service (`_transferPlugin`) using dependency injection in the class where the chat agent is created.  
- Pass it to the `Create` method as shown below:  

```csharp
System.Reflection.MethodInfo transferMethod = _transferPlugin.GetType().GetMethod(nameof(TransferCallAgentPlugin.TransferCallToAgentAsync));
var transferTool = AIFunctionFactory.Create(
    transferMethod,
    _transferPlugin,
    nameof(TransferCallAgentPlugin.TransferCallToAgentAsync)
);
```

SergeyMenshykh:
Related to #726

PR #1009, opened 2025-09-30
.NET: AI Agent with plugin

SergeyMenshykh:
This PR adds a sample demonstrating the way AIFunctions can be bundled into a plugin, exposed from the plugin, and provided to an AI agent, and how the plugin and its functions dependencies can be provided and resolved.

Context: https://github.com/microsoft/agent-framework/issues/923

### DIFF
--- a/dotnet/samples/GettingStarted/Agents/Agent_Step14_Middleware/Program.cs
+++ b/dotnet/samples/GettingStarted/Agents/Agent_Step14_Middleware/Program.cs
@@ -20,8 +20,8 @@ using Microsoft.Extensions.AI;
 #pragma warning disable MEAI001 // Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.
 
 // Get Azure AI Foundry configuration from environment variables
-var endpoint = Environment.GetEnvironmentVariable("AZUREOPENAI_ENDPOINT") ?? throw new InvalidOperationException("AZUREOPENAI_ENDPOINT is not set.");
-var deploymentName = System.Environment.GetEnvironmentVariable("AZUREOPENAI_DEPLOYMENT_NAME") ?? "gpt-4o";
+var endpoint = Environment.GetEnvironmentVariable("AZURE_OPENAI_ENDPOINT") ?? throw new InvalidOperationException("AZURE_OPENAI_ENDPOINT is not set.");
+var deploymentName = System.Environment.GetEnvironmentVariable("AZURE_OPENAI_DEPLOYMENT_NAME") ?? "gpt-4o";
 
 // Get a client to create/retrieve server side agents with
 var azureOpenAIClient = new AzureOpenAIClient(new Uri(endpoint), new AzureCliCredential())
--- a/dotnet/samples/GettingStarted/Agents/Agent_Step14_Middleware/README.md
+++ b/dotnet/samples/GettingStarted/Agents/Agent_Step14_Middleware/README.md
@@ -24,8 +24,8 @@ Attempting to use function middleware on agents that do not wrap a ChatClientAge
 ## Prerequisites
 
 1. Environment variables:
-   - `AZUREOPENAI_ENDPOINT`: Your Azure OpenAI endpoint
-   - `AZUREOPENAI_DEPLOYMENT_NAME`: Chat deployment name (optional; defaults to `gpt-4o`)
+   - `AZURE_OPENAI_ENDPOINT`: Your Azure OpenAI endpoint
+   - `AZURE_OPENAI_DEPLOYMENT_NAME`: Chat deployment name (optional; defaults to `gpt-4o`)
 2. Sign in with Azure CLI (PowerShell):
    ```powershell
    az login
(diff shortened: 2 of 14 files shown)


dotnet/samples/GettingStarted/Agents/Agent_Step15_Plugins/Program.cs
```diff
@@ -0,0 +1,137 @@
...
+const string AgentInstructions = "You are a helpful assistant that helps people find information.";
+
+AIAgent agent = new AzureOpenAIClient(
+    new Uri(endpoint),
+    new AzureCliCredential())
+     .GetChatClient(deploymentName)
+     .CreateAIAgent(
+        instructions: AgentInstructions,
+        name: AgentName,
+        tools: serviceProvider.GetRequiredService<AgentPlugin>().AsAITools(),
```
stephentoub: If you use `[.. serviceProvider.GetRequiredService<AgentPlugin>().AsAITools()]`, then it can demonstrate how multiple providers can be used, as they can all be listed in the collection expression. The AsAITools can also return an enumerable rather than list, such that you could use yield return if you wanted to in the implementation.

dotnet/samples/GettingStarted/Agents/Agent_Step15_Plugins/Agent_Step15_Plugins.csproj
```diff
@@ -0,0 +1,25 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net9.0</TargetFramework>
+    <LangVersion>12</LangVersion>
```
stephentoub: You don't need a LangVersion here; the TargetFramework implies a LangVersion. This is also strangely downgrading the language version, as the default for net9.0 is 13, not 12.

dotnet/samples/GettingStarted/Agents/Agent_Step15_Plugins/Agent_Step15_Plugins.csproj
```diff
@@ -0,0 +1,25 @@
...
+
+    <Nullable>enable</Nullable>
+    <ImplicitUsings>disable</ImplicitUsings>
```
stephentoub, line 8: Why?

SergeyMenshykh: To align with the rest of the samples, where I believe implicit using is disabled, to explicitly show the used namespace and potentially simplify the copy-paste scenarios.

dotnet/src/Microsoft.Agents.AI.OpenAI/Extensions/OpenAIChatClientExtensions.cs
```diff
@@ -39,7 +40,8 @@ public static AIAgent CreateAIAgent(
         string? description = null,
         IList<AITool>? tools = null,
         Func<IChatClient, IChatClient>? clientFactory = null,
-        ILoggerFactory? loggerFactory = null) =>
+        ILoggerFactory? loggerFactory = null,
+        IServiceProvider? functionInvocationServices = null) =>
```
rogerbarreto: For simplicity I would explain its usage in the `xmlDocs`.

This can be left open as also a potential parameter to use for an `AIAgent` constructor if we see fit.

```suggestion
        IServiceProvider? serviceProvider = null) =>
```

dotnet/samples/GettingStarted/Agents/Agent_Step15_Plugins/Agent_Step15_Plugins.csproj
stephentoub: > To align with the rest of the samples, where I believe implicit using is disabled, to explicitly show the used namespace and potentially simplify the copy-paste scenarios.

We can revisit them all-up. If someone is using net9.0, then implicit usings are enabled by default, which means copy/paste will "just work" (as long as we're not separately augmenting the list of implicitly-imported usings, which we're not). And allowing the default means slightly shorter samples, which is nice. Plus, if someone is using VS and copy/pastes and they don't have implicit usings, VS will insert the usings automatically unless the dev also explicitly disabled that feature. The list of default imported usings is short, really core namespaces like System, System.Linq, System.Collections.Generic, and System.Threading.Tasks... having those in each sample doesn't really educate the dev, and certainly not about AF.

SergeyMenshykh: Captured this in the issue https://github.com/microsoft/agent-framework/issues/1015 to address separately.

dotnet/src/Microsoft.Agents.AI.OpenAI/Extensions/OpenAIChatClientExtensions.cs
SergeyMenshykh: Renamed it to service to be aligned with AIAgentBuilder.

PR #1010, opened 2025-09-30
.NET: Update Migration Samples for SK OpenAIResponsesAgent Step 1 & 2

### LINKED ISSUE #1002: .NET: Semantic Kernel Migration OpenAIResponse samples failed
[Step_1 & Step_2] Microsoft.SemanticKernel.KernelException: 'OpenAIResponseAgent currently only supports agent threads of type ChatHistoryAgentThread.'

rogerbarreto:
### Motivation and Context

- Resolves #1002

Fix the usage of threads for SK sample, where those should be captured after the call and not created before.

There's was identified a bug in SK Agent were when the agent stored enabled is false the generated thread from a previous execution can't be reused in subsequent calls.

- https://github.com/microsoft/semantic-kernel/issues/13201


SergeyMenshykh:
Same issue is reproducible for **AzureOpenAIResponses_Step01_Basics** and **AzureOpenAIResponses_Step02_ReasoningModel** samples.

PR #1030, opened 2025-09-30
.Net: Bugfix FunctionInvocation Middleware not working when no ChatOptions provided

rogerbarreto:
### Motivation and Context

A bug was preventing function invocation filter to trigger when no options were provided as the carrier for upstream agents with a `ChatClientFactory`.

Unit Test added also covering the scenario.


dotnet/src/Microsoft.Agents.AI/FunctionInvocationDelegatingAgent.cs
```diff
@@ -30,6 +30,8 @@ public override IAsyncEnumerable<AgentRunResponseUpdate> RunStreamingAsync(IEnum
     // Decorate options to add the middleware function
     private AgentRunOptions? AgentRunOptionsWithFunctionMiddleware(AgentRunOptions? options)
     {
+        options ??= new ChatClientAgentRunOptions();
```
stephentoub: Can you help me understand what the issue was and why this fixes it?

What happened if options wasn't null but was instead just `new AgentRunOptions()`?

rogerbarreto: That's a good point, using `AgentRUnOptions` directly or other specializations would also skip the middleware filtering.

As we discussed there are some options.

1. Additional Properties added to the `AgentRunOptions` and use a reserved keyword to set the `ChatClientFactory` and use in higher level abstractions

2. Create a FunctionInvocationAgentOptionsWrapper, taking the original runOptions and exposing it as a `AIOptions InnerAgentOptions` property

3. Provide an external interface that needs to be implemented by the `AgentRunOptions` specialization to allow usages with FI middleware filtering, otherwise attempting to use a `AgentRunOptions` specialization that doesn't implement an interface in a FIMiddleware enabled agent will throw an exception.

    ```csharp
    public interface IFunctionInvocableEnabledAgentOptions 
    {
        public Func<IChatClient, IChatClient>? ChatClientFactory { get; set; }
    }
    ```
    
✅ 4. For now, if the `AgentRunOptions` is an specialization we will throw. 

Tracking next steps here:

- https://github.com/microsoft/agent-framework/issues/1032

PR #1038, opened 2025-10-01
Python: Changed to package true

adhazel:
### Motivation and Context

I am using the python library and poetry would not load the library after recent changes.

### Description

1. Added a folder under python called agent_framework with an __init__.py file that installs all core and optional packages of the agent framework

### Contribution Checklist

### DIFF
--- a/python/agent_framework/__init__.py
+++ b/python/agent_framework/__init__.py
@@ -1,0 +1,19 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""Microsoft Agent Framework for Python.
+
+This is the main agent-framework package that provides a convenient way to install
+all the core and optional packages of the Microsoft Agent Framework.
+"""
+
+import contextlib
+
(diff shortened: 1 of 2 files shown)


python/agent_framework/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+all the core and optional packages of the Microsoft Agent Framework.
+"""
+
+import contextlib
+
+__version__ = "1.0.0-b251001"
+
+# Re-export commonly used components from core package
+with contextlib.suppress(ImportError):
+    from agent_framework_core import *  # noqa: F403
```
moonbox3, line 16: This change pulls in `agent_framework_core` every time someone imports `agent_framework`, so the metapackage stops acting as a lightweight bundle and breaks environments without the optional extras.

eavanvalkenburg:
thanks for pointing this out @adhazel we have a slightly different solution for the same, that @moonbox3 created which is now merging, so I'm closing this!

PR #1043, opened 2025-10-01
Python: Add metapackage metadata stub to restore flit builds

moonbox3:
### Motivation and Context

- Supply agent_framework_meta/__init__.py with a typed version loader so the metapackage exposes a module for flit without touching runtime imports.
- Update [tool.flit.module] in pyproject.toml to reference that helper, fixing get_requires_for_build_wheel.
- Validation: rebuilt the wheelhouse using `uv build --wheel --out-dir dist-local …` to verify the metapackage now installs cleanly alongside the published subpackages.

### DIFF
--- a/python/agent_framework_meta/__init__.py
+++ b/python/agent_framework_meta/__init__.py
@@ -1,0 +1,31 @@
+# Copyright (c) Microsoft. All rights reserved.
+
+from importlib import metadata as _metadata
+from pathlib import Path as _Path
+from typing import Any, cast
+
+try:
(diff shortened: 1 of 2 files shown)


python/agent_framework_meta/__init__.py
```diff
@@ -0,0 +1,31 @@
+# Copyright (c) Microsoft. All rights reserved.
+
```
ekzhu, line 2: We can write some thing here to tell ourselves why we need this package.

PR #1049, opened 2025-10-01
Python: Update README with links to video content. Include initial code samples as a quickstart

moonbox3:
### Motivation and Context

Update README with links to video content. Include initial code samples as a quickstart


README.md
```diff
@@ -38,6 +62,66 @@ Welcome to Microsoft's comprehensive multi-language framework for building, orch
...
 
+## Quickstart
+
+### Basic Agent - Python
+
+Create a simple Azure Responses Agent that writes a haiku about the Microsoft Agent Framework
+
+```python
+import asyncio
+from agent_framework.azure import AzureOpenAIResponsesClient
```
eavanvalkenburg, line 73: missing AzureCliCredential import

README.md
```diff
@@ -38,6 +62,66 @@ Welcome to Microsoft's comprehensive multi-language framework for building, orch
...
+using System;
+using System.Threading.Tasks;
+using Azure.AI.OpenAI;
+using Azure.Identity;
+using Microsoft.Agents.AI;
+
+class Program
+{
+    static async Task Main()
+    {
```
markwallace-microsoft: Can you remove the class and function, we are using top-level statements in our samples, see: https://github.com/microsoft/agent-framework/blob/main/dotnet/samples/GettingStarted/Agents/Agent_Step01_Running/Program.cs

README.md
```diff
@@ -38,6 +62,66 @@ Welcome to Microsoft's comprehensive multi-language framework for building, orch
...
+            .GetOpenAIResponseClient(deploymentName)
+            .CreateAIAgent(name: "HaikuBot", instructions: "You are an upbeat assistant that writes beautifully.");
+
+        var thread = agent.GetNewThread();
+        var agentOptions = new ChatClientAgentRunOptions(new() { MaxOutputTokens = 8000 });
+
+        var result = await agent.RunAsync(userInput, thread, agentOptions);
+        Console.WriteLine(result);
+
+        Console.WriteLine("---");
```
markwallace-microsoft: No need to include this

README.md
```diff
@@ -38,6 +62,66 @@ Welcome to Microsoft's comprehensive multi-language framework for building, orch
...
+        await foreach (var update in agent.RunStreamingAsync(userInput, thread, agentOptions))
+        {
+            Console.Write(update);
```
markwallace-microsoft: Use `Console.WriteLine` so developers can see the streaming effect

README.md
rogerbarreto, line 1: The python and .net samples are different, .net shows streaming and non-streaming in the same bucket, ideally we should have the same sample for both.

moonbox3: Will make .NET non-streaming.

README.md
rogerbarreto: Does python have the same api for streaming and non-streaming, if not, I would just remove the streaming or add the streaming to the python to be compatible.

The instructions in the agent and examples differ as well, ideally we should use the same.

Python doesn't use max tokens = 8000 or options..

README.md
moonbox3: Addressed thank you

README.md
moonbox3: Addressed thank you

README.md
moonbox3: Addressed thank you

README.md
moonbox3: Addressed, please have a look. Thanks.

README.md
```diff
@@ -38,6 +62,66 @@ Welcome to Microsoft's comprehensive multi-language framework for building, orch
...
+print(result)
+
+asyncio.run(main()) 
+```
+
+### Basic Agent - .NET
+
+```c#
+using System;
+using System.Threading.Tasks;
```
rogerbarreto: We might be just ok without those 2 (implicit usings)

```suggestion
```

README.md
moonbox3: I removed the streaming path from .NET, as well as the options.

README.md
```diff
@@ -38,6 +62,48 @@ Welcome to Microsoft's comprehensive multi-language framework for building, orch
...
+
+```c#
+using System;
+using Azure.AI.OpenAI;
+using Azure.Identity;
+using Microsoft.Agents.AI;
+
+var endpoint = Environment.GetEnvironmentVariable("AZURE_OPENAI_ENDPOINT")
+    ?? throw new InvalidOperationException("AZURE_OPENAI_ENDPOINT is not set.");
+var deploymentName = Environment.GetEnvironmentVariable("AZURE_OPENAI_DEPLOYMENT_NAME") ?? "gpt-4o";
```
rogerbarreto: No throwing logic or model mentions needs to be in the sample we can simplify further

```suggestion
var endpoint = Environment.GetEnvironmentVariable("AZURE_OPENAI_ENDPOINT")!;
var deploymentName = Environment.GetEnvironmentVariable("AZURE_OPENAI_DEPLOYMENT_NAME")!;
```

PR #1053, opened 2025-10-01
.NET: Align environment variable names

SergeyMenshykh:


dotnet/samples/SemanticKernelMigration/OpenAIAssistants/Step04_CodeInterpreter/Program.cs
```diff
@@ -13,7 +13,7 @@
 #pragma warning disable CS8321 // Local function is declared but never used
 
 var apiKey = Environment.GetEnvironmentVariable("OPENAI_API_KEY") ?? throw new InvalidOperationException("OPENAI_API_KEY is not set.");
-var modelId = System.Environment.GetEnvironmentVariable("OPENAI_MODELID") ?? "gpt-4o";
+var modelId = System.Environment.GetEnvironmentVariable("OPENAI_MODEL") ?? "gpt-4o";
```
rogerbarreto: The name of the variable is ModelId, and the `argument` name as well, should we reconsider keeping the `OPENAI_MODELID` instead? Thoughts?

```suggestion
var modelId = System.Environment.GetEnvironmentVariable("OPENAI_MODELID") ?? "gpt-4o";
```

SergeyMenshykh: The variables need to be renamed as well. They eventually passed to downstream methods like `GetChatClient` that have `model` parameter.

PR #613, opened 2025-09-04
.NET: Support SubWorkflow as Executor

lokitoth:
* Makes input-type annotation optional
* Add WorkflowHostExecutor and Workflow-type ExecutorIsh

TODO:
- [x] Route Subworkflow Outputs through SendMessage
- [ ] Validate external requests work right
- [ ] Validate checkpointing works right
- [ ] ~~Multi-output Refactoring~~

PR #439, opened 2025-08-18
.NET: feat: Decouple running SuperSteps from delivering WorkflowEvent

lokitoth:
The goal of WorkflowEvent is to allow asynchronous bubbling up of events from the workflow. Right now, we run supersteps and deliver events in lockstep: SuperStep = tick, DeliverEvents = tock. This change separates the running of steps from event delivery.

Remaining Todos:
- [ ] Enable optional Lock-step mode
- [ ] Remove need to have Task.Delay in sample tests

Issue #410, opened 2025-08-13
.NET: Ensure correct propagation of Cancellation matching reasonable expectations


lokitoth:
Right now it is not always clear which cancellation token should cancel which operation (especially for Async operations that result in a StreamingRun object).

This task is about ensuring:

1. All async APIs have an appropriate mechanism to be cancelled through `CancellationToken`s
2. `CancellationToken`s are propagated to asynchronous user extension points (e.g. executor message handlers)

PR #1066, opened 2025-10-01
.NET: Add support for Subworkflows and many threading fixes

lokitoth:
### Motivation and Context

Running workflows as an executor node in another workflow is very helpful for building componentized, composable systems. We add this by exposing a way for a parent executor to drive SuperSteps inside of a child executor.

This (and some of the work around supporting off-thread execution) lead to a number of concurrency fixes which have been pulled together into a train on top of the Subworkflow change (which exposes more chances for some of the run conditions being fixed)

### Description

Changes:
* Add support for Workflow-as-Executor (replaces #613)
* Some plumbing for OffThread execution (progress towards #439)
* Many concurrency fixes

BREAKING: 
* Renames InputPort => RequestPort to better match Python nomenclature and to separate from top-level workflow input.
* Changes executor id to no longer be an optional parameter (matching Python)
* Run.RunStatus becomes an async call: GetRunStatusAsync()
* StreamingRun.HasUnprocessedRequests removed, use GetRunStatusAsync() instead

### Contribution Checklist


dotnet/src/Microsoft.Agents.AI.Workflows/AsyncBarrier.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Microsoft.Agents.AI.Workflows;
+
+internal sealed class AsyncBarrier
+{
+    private readonly TaskCompletionSource<object> _completionSource = new();
+
+    public async ValueTask JoinAsync(CancellationToken cancellation = default)
+    {
+        using CancellationTokenRegistration registration = cancellation.Register(() => this.ReleaseBarrier());
```
stephentoub: This will end up completing the completion source as successful... will that then mean that others waiting on the barrier may continue running thinking all conditions are satisfied? Should this instead do TrySetCanceled on _completionSource?

dotnet/src/Microsoft.Agents.AI.Workflows/AsyncCoordinator.cs
```diff
@@ -0,0 +1,26 @@
...
+
+internal sealed class AsyncCoordinator
+{
+    private AsyncBarrier? _coordinationBarrier;
+
+    public async ValueTask WaitForCoordinationAsync(CancellationToken cancellation = default)
+    {
+        AsyncBarrier newBarrier = new();
+        AsyncBarrier? actualBarrier = Interlocked.CompareExchange(ref this._coordinationBarrier, newBarrier, null);
+        actualBarrier ??= newBarrier;
```
stephentoub: This is going to create a new AsyncBarrier each time, even if there already is one.
```suggestion
        AsyncBarrier actualBarrier =
            this._coordinationBarrier ??
            Interlocked.CompareExchange(ref this._coordinationBarrier, new(), null) ??
            this._coordinationBarrier;
```
That said, I see in MarkCoordinationPoint below that it's nulling out _coordinationBarrier, and then here it's falling back to a new instance if it was nulled out, and it's not reinitializing coordination barrier in that case. Won't this then end up joining forever, or at least until cancellation is requested?

dotnet/src/Microsoft.Agents.AI.Workflows/Execution/LockstepRunEventStream.cs
```diff
@@ -0,0 +1,117 @@
...
+                         { Tags.ErrorType, ex.GetType().FullName },
+                         { Tags.BuildErrorMessage, ex.Message },
+                    }));
+                    activity.CaptureException(ex);
+                    throw;
+                }
+
+                if (cancellation.IsCancellationRequested)
+                {
+                    yield break; // Exit if cancellation is requested
```
stephentoub, line 71: When we're explicitly checking, we're just breaking out of the loop, but presumably when the token is passed to other methods, like RunSuperStepAsync, those may throw OperationCanceledException? Do we need to exhibit similar behaviors in both cases?

dotnet/src/Microsoft.Agents.AI.Workflows/Execution/OffThreadRunEventStream.cs
```diff
@@ -0,0 +1,193 @@
...
+                        }
+
+                        // We hit a halt signal for a previous epoch, so we ignore it.
+                    }
+                    else
+                    {
+                        yield return @event;
+                    }
+                }
+                else if (!cancellation.IsCancellationRequested)
```
stephentoub: There's a lot of polling of the token happening. Is it all necessary?

dotnet/src/Microsoft.Agents.AI.Workflows/Execution/OffThreadRunEventStream.cs
```diff
@@ -0,0 +1,193 @@
...
+        }
+        finally
+        {
+            this.StepRunner.OutgoingEvents.EventRaised -= InspectAndForwardWorkflowEventAsync;
+        }
+
+        [SuppressMessage("Usage", "VSTHRD100:Avoid async void methods", Justification = "Internal event handler - nowhere to return exceptions to.")]
+        async ValueTask InspectAndForwardWorkflowEventAsync(object? sender, WorkflowEvent e)
+        {
+            int epoch = Volatile.Read(ref this._streamEpoch);
```
stephentoub: What is this for?

dotnet/src/Microsoft.Agents.AI.Workflows/Execution/OffThreadRunEventStream.cs
```diff
@@ -0,0 +1,193 @@
...
+        int epoch = Math.Max(Volatile.Read(ref this._streamEpoch), 0);
+        this._eventSink.Enqueue(new OffThreadHaltSignal(epoch));
+    }
+
+    private async Task RunLoopAsync(CancellationToken cancellation = default)
+    {
+        this.StepRunner.OutgoingEvents.EventRaised += InspectAndForwardWorkflowEventAsync;
+        bool hadRequestHaltEvent = false;
+
+        int epoch = Volatile.Read(ref this._streamEpoch);
```
stephentoub: Is this used anywhere?

dotnet/src/Microsoft.Agents.AI.Workflows/InProc/InProcStepTracer.cs
```diff
@@ -72,19 +73,19 @@ public override string ToString()
     {
         StringBuilder sb = new();
 
-        if (this.Instantiated.Count != 0)
+        if (this.Instantiated.Keys.Count != 0)
```
stephentoub: Keys is unexpectedly expensive. Better to access it once and store it into a local.

dotnet/src/Microsoft.Agents.AI.Workflows/AsyncBarrier.cs
lokitoth: Maybe as an option when constructing (though probably set to cancel by default) - right now it is intentional that the "wait for coordination" be released cleanly and let the threads react to the new status (cancelled) through their own CancellationTokens.

dotnet/src/Microsoft.Agents.AI.Workflows/Execution/LockstepRunEventStream.cs
lokitoth: We probably should at least catch and normalize the cases, since it is a bit stochastic how different parts of the code would react to cancellation.

dotnet/src/Microsoft.Agents.AI.Workflows/Execution/OffThreadRunEventStream.cs
lokitoth: Bad rebase. It used to be async void, then I finally got to a good place to change the event it binds to.

dotnet/src/Microsoft.Agents.AI.Workflows/Execution/OffThreadRunEventStream.cs
lokitoth: No, this entire class needs to be rewritten; it was dead code in the change, but pulling it since we have other places where this code is archived.

dotnet/src/Microsoft.Agents.AI.Workflows/AsyncCoordinator.cs
lokitoth: I've changed it so it only actually instantiates the TaskCompletionSource now when it gets first awaited. Otherwise, I'd expect it to be a fairly cheap object: int and a null pointer, and it gets bashed over by actualBarrier if one exists.

dotnet/src/Microsoft.Agents.AI.Workflows/AsyncBarrier.cs
lokitoth: I've fixed the logic here; now we properly create a unique "cancellationSource" for the individual waiter that can be used to selectively cancel only that wait, without propagating it to the rest.

dotnet/src/Microsoft.Agents.AI.Workflows/Execution/LockstepRunEventStream.cs
lokitoth: Further update: There is another bit of work around Cancellation propagation and doing the right thing (and probably updating the docs to make it clear) that I am going to tackle after this (and returning OffThread execution back): #410

I think it makes sense to punt dealing with OperationCancelledException in the main loop to that. Once we trap it there, make sure we are in a comprehensible state we can throw it (or maybe not, since there is nowhere for it to go when running OffThread: It would need to turn into a WorkflowEvent, but there's no guarantee the user will continue pumping it, etc.)

dotnet/src/Microsoft.Agents.AI.Workflows/Execution/OffThreadRunEventStream.cs
lokitoth: I think it's meant to be used to break out of the inner/outer loops, but this code is gone for now, to be fixed and restored in a later PR.

dotnet/src/Microsoft.Agents.AI.Workflows/InProc/InProcStepTracer.cs
lokitoth: Will keep that in mind in the future, thanks for pointing it out

crickman (approved):

dotnet/src/Microsoft.Agents.AI.Workflows/Checkpointing/WorkflowInfo.cs
```diff
@@ -94,9 +94,4 @@ public bool IsMatch(Workflow workflow)
 
     public bool IsMatch<TInput>(Workflow<TInput> workflow) =>
         this.IsMatch(workflow as Workflow) && this.InputType?.IsMatch<TInput>() == true;
-
-    //public bool IsMatch<TInput, TResult>(WorkflowWithOutput<TInput, TResult> workflow)
-    //    => this.IsMatch(workflow as Workflow)
-    //       && this.OutputType?.IsMatch(typeof(TResult)) is true
-    //       && this.OutputCollectorId is not null && this.OutputCollectorId == workflow.OutputCollectorId;
```
alliscode, line 101: nit: we can delete this

lokitoth: I believe it is already getting deleted?

PR #1075, opened 2025-10-01
Update README hello world example

### LINKED ISSUE #1073: Python: README: Python Quickstart should be clear enough users can run without error
The python snippet in the Quickstart section of the readme fails to run because it doesn't have a deployment name or an endpoint specified anywhere.  There are a few ways that this can be approached (env vars, direct setting in the instantiation) and the right way to do it should be advertised for people getting started

ekzhu:
Resolves #1073

### Contribution Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,8 +28,9 @@ Welcome to Microsoft's comprehensive multi-language framework for building, orch
 Python
 
 ```bash
-# This will install all sub-packages, see `python/packages` for individual packages
-pip install agent-framework
+pip install agent-framework --prerelease=allow
+# This will install all sub-packages, see `python/packages` for individual packages.
+# It may take a minute on first install on Windows.
 ```
 
 .NET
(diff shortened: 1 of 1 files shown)


victordibia (approved):

README.md
```diff
@@ -87,20 +88,27 @@ dotnet add package Microsoft.Agents.AI
...
+
 async def main():
-  # Initialize a chat agent with basic instructions
-  agent = AzureOpenAIResponsesClient(credential=AzureCliCredential()).create_agent(
-    name="HaikuBot",
-    instructions="You are an upbeat assistant that writes beautifully.",
-  )
+    # Initialize a chat agent with Azure OpenAI Responses
+    agent = AzureOpenAIResponsesClient(
+        endpoint="https://your-custom-endpoint.openai.azure.com/",
```
moonbox3, line 101: These are automatically picked up from env vars or a .env file, which is why they aren’t explicitly included.

PR #1135, opened 2025-10-02
Python: docs: added examples to OTEL with Opik

vincentkoc:
### Motivation and Context

Opik (https://github.com/comet-ml/opik) is one of the leading open-source telemetry, evaluation and optimization platform. A large % of our open-source users are `Semantic Kernel` and `Autogen` users and they have been asking for `agent-framework` support from day 1. Work is based on our reference implementations (https://www.comet.com/docs/opik/integrations/microsoft-agent-framework).

### Description

Added a working example of Opik via OTEL in docs under `python/samples/getting_started/observability/`

### Contribution Checklist


eavanvalkenburg:
@vincentkoc sorry it took a while to get back on this, we have just updated the observability setup and it is much more generic now, and would work for this as well, (setting a OTEL_EXPORTER_OTLP_ENDPOINT and HEADERS will be used) when you can `configure...` so this sample doesn't add much value, I would propose adding a snippet on your docs, then we can link to that, we did the same with langfuse now. See the new readme in the observability sample folder. I am going to close this PR, let me know if we can do something like that link for you as well!

vincentkoc:
> @vincentkoc sorry it took a while to get back on this, we have just updated the observability setup and it is much more generic now, and would work for this as well, (setting a OTEL_EXPORTER_OTLP_ENDPOINT and HEADERS will be used) when you can `configure...` so this sample doesn't add much value, I would propose adding a snippet on your docs, then we can link to that, we did the same with langfuse now. See the new readme in the observability sample folder. I am going to close this PR, let me know if we can do something like that link for you as well!

@eavanvalkenburg i finnaly got around to it and updated with a new PR https://github.com/microsoft/agent-framework/pull/3940

PR #1156, opened 2025-10-03
.NET: fix: Make CheckpointInfo constructor public

lokitoth:
### Motivation and Context

In order to be able to restore a `StreamingRun/Run` from a checkpoint, it is necessary to pass in a `CheckpointInfo` object. Without a public constructor it can only be used if a `CheckpointInfo` object is generated by the checkpointing system and kept around until the checkpoint needs to be restored, or if JSON serialization is used to produce the object.

### Description

Make the constructor public.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.Workflows/CheckpointInfo.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Workflows/CheckpointInfo.cs
@@ -29,8 +29,13 @@ public sealed class CheckpointInfo : IEquatable<CheckpointInfo>
     /// hexadecimal format  and sets the timestamp to the current UTC time.</remarks>
     internal CheckpointInfo(string runId) : this(runId, Guid.NewGuid().ToString("N")) { }
 
+    /// <summary>
+    /// Initializes a new instance of the CheckpointInfo class with the specified run and checkpoint identifiers.
+    /// </summary>
+    /// <param name="runId">The unique identifier for the run. Cannot be null or empty.</param>
+    /// <param name="checkpointId">The unique identifier for the checkpoint. Cannot be null or empty.</param>
     [JsonConstructor]
-    internal CheckpointInfo(string runId, string checkpointId)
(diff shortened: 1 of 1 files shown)


stephentoub (approved):

dotnet/src/Microsoft.Agents.AI.Workflows/CheckpointInfo.cs
```diff
@@ -29,8 +29,13 @@ public sealed class CheckpointInfo : IEquatable<CheckpointInfo>
...
     internal CheckpointInfo(string runId) : this(runId, Guid.NewGuid().ToString("N")) { }
 
+    /// <summary>
+    /// Initializes a new instance of the CheckpointInfo class with the specified run and checkpoint identifiers.
+    /// </summary>
+    /// <param name="runId">The unique identifier for the run. Cannot be null or empty.</param>
+    /// <param name="checkpointId">The unique identifier for the checkpoint. Cannot be null or empty.</param>
     [JsonConstructor]
-    internal CheckpointInfo(string runId, string checkpointId)
+    public CheckpointInfo(string runId, string checkpointId)
```
stephentoub, line 38: Do we need tests for this now that it's public?

crickman: I'd assert we tests even if its internal.  We just need tests, period.

lokitoth: Broadly we need to increase the test coverage of Workflows, though this specific area I am less worried about, since it is only used as a key in a dictionary (essentially a CheckpointId)

stephentoub: > I'd assert we tests even if its internal. We just need tests, period.

IVT is my arch nemesis.

Issue #637, opened 2025-09-08
.NET: Avoid duplicating coalescing logic in ToAgentRunResponse


stephentoub:
ToAgentRunResponse has a copy of the AIContent coalescing logic from ToChatResponse. We should find a way to avoid that, to reduce the amount of custom code but also because that coalescing logic is evolving and the copy is becoming stale. Worst case, we should be able to create temporary ChatResponseUpdate instances containing the same AIContents, coalesce, and then copy back, but we could also look at the best way to expose that coalescing logic publicly from M.E.AI.Abstractions and then use that directly.

lokitoth:
There is also another "copy" in Workflow's codebase: https://github.com/microsoft/agent-framework/blob/main/dotnet/src/Microsoft.Agents.Workflows/MessageMerger.cs

PR #1169, opened 2025-10-03
.NET: Removing duplicating update coalescing logic in AgentRunResponseExtensions

stephentoub:
Fixes https://github.com/microsoft/agent-framework/issues/637

@westey-m, how do you feel about this approach?

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.Abstractions/AgentRunResponseExtensions.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Abstractions/AgentRunResponseExtensions.cs
@@ -2,7 +2,7 @@
 
 using System;
 using System.Collections.Generic;
-using System.Text;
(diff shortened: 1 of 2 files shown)


westey-m:
> Fixes #637
> 
> @westey-m, how do you feel about this approach?

@stephentoub, I think this makes sense.  While we have to convert twice, these conversations are relatively cheap and I think it's better than duplicating most of the code.

stephentoub:
> While we have to convert twice, these conversations are relatively cheap and I think it's better than duplicating most of the code.

Also, if the underlying agent is a ChatClientAgent, AsChatResponseUpdate will be a nop, just returning the ChatResponseUpdate instance that's already in the AgentRunResponseUpdate.RawRepresentation.

westey-m (approved):

PR #1172, opened 2025-10-03
.NET: Improve structured output for chat client agent

SergeyMenshykh:
This PR adds several overloads of the `RunAsync<T>` methods to `ChatClientAgent`, improving the development experience when using the agent to produce structured output from **non-streaming** APIs.

#### Code before
```csharp
// Create the agent options, specifying the response format to use a JSON schema based on the PersonInfo class.
ChatClientAgentOptions agentOptions = new(name: "HelpfulAssistant", instructions: "You are a helpful assistant.")
{
    ChatOptions = new()
    {
        ResponseFormat = ChatResponseFormat.ForJsonSchema<PersonInfo>()
    }
};

// Create the agent using Azure OpenAI.
    new Uri(endpoint),
    new AzureCliCredential())
        .GetChatClient(deploymentName)
        .CreateAIAgent(agentOptions);

// Invoke the agent with some unstructured input, to extract the structured information from.
var response = await agent.RunAsync("Please provide information about John Smith, who is a 35-year-old software engineer.");

// Deserialize the response into the PersonInfo class.
var personInfo = response.Deserialize<PersonInfo>(JsonSerializerOptions.Web);

Console.WriteLine("Assistant Output:");
Console.WriteLine($"Name: {personInfo.Name}");
Console.WriteLine($"Age: {personInfo.Age}");
Console.WriteLine($"Occupation: {personInfo.Occupation}");
```

#### Code after
```csharp
// Create the ChatClientAgent with the specified name and instructions.
ChatCilentAgent agent = new AzureOpenAIClient(
    new Uri(endpoint),
    new AzureCliCredential())
        .GetChatClient(deploymentName)
        .CreateAIAgent(agentOptions);

 // Set PersonInfo as the type parameter of RunAsync method to specify the expected structured output from the agent and invoke the agent with some unstructured input.
 AgentRunResponse<PersonInfo> response = await agent.RunAsync<PersonInfo>("Please provide information about John Smith, who is a 35-year-old software engineer.");

 // Access the structured output via the Result property of the agent response.
 Console.WriteLine("Assistant Output:");
 Console.WriteLine($"Name: {response.Result.Name}");
 Console.WriteLine($"Age: {response.Result.Age}");
 Console.WriteLine($"Occupation: {response.Result.Occupation}");
```

### DIFF
--- a/dotnet/samples/GettingStarted/Agents/Agent_Step05_StructuredOutput/Program.cs
+++ b/dotnet/samples/GettingStarted/Agents/Agent_Step05_StructuredOutput/Program.cs
@@ -34,7 +34,7 @@ Console.WriteLine($"Age: {response.Result.Age}");
 Console.WriteLine($"Occupation: {response.Result.Occupation}");
 
 // Create the ChatClientAgent with the specified name, instructions, and expected structured output the agent should produce.
-agent = chatClient.CreateAIAgent(new ChatClientAgentOptions(name: "HelpfulAssistant", instructions: "You are a helpful assistant.")
+ChatClientAgent agentWithPersonInfo = chatClient.CreateAIAgent(new ChatClientAgentOptions(name: "HelpfulAssistant", instructions: "You are a helpful assistant.")
 {
     ChatOptions = new()
     {
@@ -43,7 +43,7 @@ agent = chatClient.CreateAIAgent(new ChatClientAgentOptions(name: "HelpfulAssist
 });
 
 // Invoke the agent with some unstructured input while streaming, to extract the structured information from.
-var updates = agent.RunStreamingAsync("Please provide information about John Smith, who is a 35-year-old software engineer.");
+var updates = agentWithPersonInfo.RunStreamingAsync("Please provide information about John Smith, who is a 35-year-old software engineer.");
 
 // Assemble all the parts of the streamed output, since we can only deserialize once we have the full json,
 // then deserialize the response into the PersonInfo class.


dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgentStructuralOutput.cs
```diff
@@ -0,0 +1,197 @@
...
+            chatResponseMessage.AuthorName ??= agentName;
+        }
+
+        // Only notify the thread of new messages if the chatResponse was successful to avoid inconsistent message state in the thread.
+        await NotifyThreadOfNewMessagesAsync(safeThread, inputMessages.Concat(chatResponse.Messages), cancellationToken).ConfigureAwait(false);
+
+        // Notify the AIContextProvider of all new messages.
+        await NotifyAIContextProviderOfSuccessAsync(safeThread, inputMessages, chatResponse.Messages, cancellationToken).ConfigureAwait(false);
+
+        return new AgentRunResponse<T>(chatResponse) { AgentId = this.Id };
```
stephentoub: Do you need to duplicate all the code from RunAsync? Seems like the only things that differ are the GetResponseAsync call and the construction of the response at the end, in which case you could share most of it using a delegate or interface and two implementations or something like that?

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgentStructuredOutput.cs
```diff
@@ -0,0 +1,197 @@
...
+    /// <returns>A task that represents the asynchronous operation. The task result contains an <see cref="AgentRunResponse"/> with the agent's output.</returns>
+    /// <remarks>
+    /// This overload is useful when the agent has sufficient context from previous messages in the thread
+    /// or from its initial configuration to generate a meaningful response without additional input.
+    /// </remarks>
+    public Task<AgentRunResponse<T>> RunAsync<T>(
+        AgentThread? thread = null,
+        JsonSerializerOptions? serializerOptions = null,
+        AgentRunOptions? options = null,
+        bool? useJsonSchemaResponseFormat = null,
```
stephentoub, line 41: I don't know how useful useJsonSchemaResponseFormat: false is. I guess we can keep exposing it, as it's just mirroring what the underlying API provides, but I don't see folks using false and it's not actually using structured output so its results are questionable.

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgentRunResponse{T}.cs
SergeyMenshykh, line 1: The `AgentRunResponse<T>` class can potentially be used by other agents that may support structural output; hence, it should be declared in the `Microsoft.Agents.AI.Abstractions` project rather than the `Microsoft.Agents.AI` one.  

The reason it's declared in the `Microsoft.Agents.AI` project is that the class depends on the `ChatResponse<T>` class from `Microsoft.Extension.AI` package, and moving the class to `Microsoft.Agents.AI.Abstractions` project would require adding a dependency on `Microsoft.Extension.AI` package to the project, which does not seem like the right thing to do.

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgentStructuredOutput.cs
SergeyMenshykh: The only case when developers will set it to false, as far as I understand, is when they need to get SO out of a non-SO-capable AI model, which may not be reliable and may not work. Setting it to false with an SO-capable model does not make sense.

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgentStructuralOutput.cs
SergeyMenshykh: Thanks for pointing this out. There's no need for duplication. It's a draft PR to check the overall approach. I'll polish it, add tests, and remove duplication in the next commit.

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgentRunResponse{T}.cs
SergeyMenshykh: Fixed by introducing the ChatClientAgentRunResponse<T> class and moving the AgentRunResponse<T> to the abstraction project.

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgentRunResponse{T}.cs
```diff
@@ -0,0 +1,46 @@
...
+
+/// <summary>
+/// Represents the response of the specified type <typeparamref name="T"/> to an <see cref="ChatClientAgent"/> run request.
+/// </summary>
+/// <typeparam name="T">The type of value expected from the chat response.</typeparam>
+/// <remarks>
+/// Language models are not guaranteed to honor the requested schema. If the model's output is not
+/// parsable as the expected type, you can access the underlying JSON response on the <see cref="AgentRunResponse.Text"/> property.
+/// </remarks>
+public class ChatClientAgentRunResponse<T> : AgentRunResponse<T>
```
westey-m: Can this be sealed?

dotnet/src/Microsoft.Agents.AI.Abstractions/AgentRunResponse{T}.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// <summary>Initializes a new instance of the <see cref="AgentRunResponse{T}"/> class.</summary>
+    protected AgentRunResponse()
+    {
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="AgentRunResponse{T}"/> class from an existing <see cref="AgentRunResponse"/>.
+    /// </summary>
+    /// <param name="response">The <see cref="ChatResponse"/> from which to populate this <see cref="AgentRunResponse{T}"/>.</param>
+    protected AgentRunResponse(ChatResponse response) : base(response)
```
markwallace-microsoft, line 22: Do we need this?

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgentRunResponse{T}.cs
SergeyMenshykh: Sealed!

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgentStructuredOutput.cs
```diff
@@ -0,0 +1,162 @@
...
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>A task that represents the asynchronous operation. The task result contains an <see cref="AgentRunResponse"/> with the agent's output.</returns>
+    /// <remarks>
+    /// This overload is useful when the agent has sufficient context from previous messages in the thread
+    /// or from its initial configuration to generate a meaningful response without additional input.
+    /// </remarks>
+    public Task<ChatClientAgentRunResponse<T>> RunAsync<T>(
+        AgentThread? thread = null,
+        JsonSerializerOptions? serializerOptions = null,
+        AgentRunOptions? options = null,
```
westey-m, line 40: Since we are at the chat client level here, and AgentRunOptions have no properties, I wonder if we should just take a ChatClientAgentRunOptions here.  Same with the other overloads.

```suggestion
        ChatClientAgentRunOptions? options = null,
```

westey-m (approved):

dotnet/src/Microsoft.Agents.AI.Abstractions/AgentRunResponse{T}.cs
SergeyMenshykh: Without this constructor we won't be able to reuse `AgentRunResponse` initialization from `ChatResponse`:
Image

that is very easy to do in the `ChatClientAgentRunResponse<T>` constructor:
Image

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgentStructuredOutput.cs
SergeyMenshykh: Discussed offline and agreed to keep it as is because there are no strong points in favor of each approach.

dotnet/src/Microsoft.Agents.AI.Abstractions/AgentRunResponse{T}.cs
```diff
@@ -0,0 +1,30 @@
...
+/// <typeparam name="T">The type of value expected from the agent.</typeparam>
+public abstract class AgentRunResponse<T> : AgentRunResponse
+{
+    /// <summary>Initializes a new instance of the <see cref="AgentRunResponse{T}"/> class.</summary>
+    protected AgentRunResponse()
+    {
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="AgentRunResponse{T}"/> class from an existing <see cref="AgentRunResponse"/>.
```
DeagleGross: comment was copied most likely, should be
```suggestion
/// Initializes a new instance of the <see cref="AgentRunResponse{T}"/> class from an existing <see cref="ChatResponse"/>.
```

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
```diff
@@ -369,6 +336,66 @@ public override AgentThread DeserializeThread(JsonElement serializedThread, Json
 
     #region Private
 
+    private async Task<TAgentRunResponse> RunCoreAsync<TAgentRunResponse, TChatClientResponse>(
+        Func<IChatClient, List<ChatMessage>, ChatOptions?, CancellationToken, Task<TChatClientResponse>> chatClientRunFunc,
```
DeagleGross, line 340: is anyone also using this? Its a private member and used in a single place, so why complication with `Func<>` is being used here?

dotnet/src/Microsoft.Agents.AI.Abstractions/AgentRunResponse{T}.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// </summary>
+    /// <param name="response">The <see cref="ChatResponse"/> from which to populate this <see cref="AgentRunResponse{T}"/>.</param>
+    protected AgentRunResponse(ChatResponse response) : base(response)
+    {
+    }
+
+    /// <summary>
+    /// Gets the result value of the agent response as an instance of <typeparamref name="T"/>.
+    /// </summary>
+    public abstract T Result { get; }
```
DeagleGross, line 29: can there be a case where Result is accessible via async invocation? Should this be `ValueTask<T>` just in case?

dotnet/tests/Microsoft.Agents.AI.UnitTests/ChatClient/ChatClientAgentTests.cs
```diff
@@ -3,14 +3,16 @@
...
+using System.Text.Json.Serialization;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.AI;
 using Moq;
 
 namespace Microsoft.Agents.AI.UnitTests;
 
-public class ChatClientAgentTests
+public partial class ChatClientAgentTests
```
DeagleGross, line 15: why partial?

SergeyMenshykh: To allow source generation, otherwise compiler says: _"ChatClientAgentTests.cs(1954,34,1954,46): error SYSLIB1032: Derived 'JsonSerializerContext' type 'JsonContext2' specifies JSON-serializable types. The type and all containing types must be made partial to kick off source generation."_

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
SergeyMenshykh: It's used in the same class but in different files: ChatClientAgentю.cs and ChatClientAgentStructuredOutput.cs

dotnet/src/Microsoft.Agents.AI.Abstractions/AgentRunResponse{T}.cs
SergeyMenshykh: The AgentRunResponse is a model class/DTO/POCO object, which means that when you have its instance, it already has everything needed initialized by an agent and does not require any extra calls to the agent or any external service to get additional information. The result property operates on a collection of messages that have already been added by the agent to the `Messages` collection and that don't require any async invocation to process them to form a result.

dotnet/samples/GettingStarted/Agents/Agent_Step05_StructuredOutput/Program.cs
```diff
@@ -1,58 +1,68 @@
...
+    Console.WriteLine($"Name: {response.Result.Name}");
+    Console.WriteLine($"Age: {response.Result.Age}");
+    Console.WriteLine($"Occupation: {response.Result.Occupation}");
+}
 
-Console.WriteLine("Assistant Output:");
-Console.WriteLine($"Name: {personInfo.Name}");
-Console.WriteLine($"Age: {personInfo.Age}");
-Console.WriteLine($"Occupation: {personInfo.Occupation}");
+async Task GetStructuredOutputWithRunStreamingAsync(ChatClient chatClient)
```
markwallace-microsoft: I don't think we need these functions, for me they make the sample harder to read. Can we in line the agent creation and create one called `agent` and one called something like `agentWithPersonInfo`

SergeyMenshykh: Agree, refactored!

markwallace-microsoft (approved):

PR #1198, opened 2025-10-05
Python: Adds NVIDIA NIM example

thegovind:
Adds an example demonstrating how to use NVIDIA NIM models with the Agent Framework through Azure AI Foundry.

The example shows how to configure the OpenAI Chat Client to use NVIDIA NIM models deployed on Azure AI Foundry, including both streaming and non-streaming responses with tool calling capabilities.


python/samples/getting_started/agents/nvidia/nvidia_nim_with_openai_chat_client.py
```diff
@@ -0,0 +1,139 @@
...
+   - Once deployed, note your endpoint URL and API key
+   - The endpoint URL should include '/v1' (e.g., 'https://<endpoint>.<region>.inference.ml.azure.com/v1')
+
+## Environment Variables
+
+Set the following environment variables before running this example:
+
+- AZURE_AI_ENDPOINT: Your Azure AI Foundry endpoint URL (e.g., 'https://<endpoint>.<region>.inference.ml.azure.com/v1')
+- AZURE_AI_API_KEY: Your Azure AI Foundry API key
+- NVIDIA_NIM_MODEL: The NVIDIA NIM model to use (e.g., 'meta/llama-3.1-8b-instruct')
```
eavanvalkenburg: if you use `OPENAI_CHAT_MODEL_ID` for this, and `OPENAI_BASE_URL` and `OPENAI_API_KEY` then you can just call the OpenAIChatClient without any differences. this is also mixing the Azure env variables with a OpenAIChatClient which is not ideal, either switch these names or use `AzureOpenAIChatClient`

thegovind: Thanks for the feedback! I updated the variable names to align with the guidance in the official [[NVIDIA NIM on Azure AI Foundry blog post](https://developer.nvidia.com/blog/accelerated-ai-inference-with-nvidia-nim-on-azure-ai-foundry/)](https://developer.nvidia.com/blog/accelerated-ai-inference-with-nvidia-nim-on-azure-ai-foundry/), which uses `OPENAI_BASE_URL`, `OPENAI_API_KEY`, and `OPENAI_CHAT_MODEL_ID` as the standard pattern for NVIDIA NIM deployments on Azure.

The `nvidia` folder structure is intentional - customers transitioning from Semantic Kernel will have a clearer migration path to this framework with dedicated NVIDIA examples. We also have additional NIM integrations and examples coming soon that will live in this folder, so it made sense to establish the structure now.

Happy to discuss further if you have concerns about this approach! Thanks

thegovind:
@ekzhu @markwallace-microsoft @eavanvalkenburg - Please hold off on merging this PR. I have a few updates to make (working with @azeltov) and will push the changes within the next couple of hours

thegovind:
Simplified the abstractions and example. Thanks to @azeltov. We will be adding more to this `nvidia` subfolder with nemotron family of open models as they become available on AI Foundry. Let's hold off on this PR until the 10/16 due to these new nemotron updates.

thegovind:
We are good to merge now. Tested and confirmed with @azeltov

python/samples/getting_started/agents/nvidia/nvidia_nim_chat_client.py
```diff
@@ -0,0 +1,79 @@
...
+    Role,
+    FunctionCallContent,
+    FunctionResultContent,
+    FunctionApprovalRequestContent,
+    FunctionApprovalResponseContent,
+    prepare_function_call_results,
+)
+
+
+class NVIDIANIMChatClient(OpenAIChatClient):
```
moonbox3, line 20: Directed to @eavanvalkenburg: we have other clients based on the OpenAIChatClient, why wouldn't we want to bring this NVIDIANIMChatClient into its own package, instead of it being purely in the samples?

eavanvalkenburg: Good question @moonbox3 I've asked Shawn for his opinion

eavanvalkenburg:
Closing because this is stale, feel free to reopen, I wouldn't mind having a native NVidia NIM client, but it would require some work with the new structure (raw clients etc.)

PR #1207, opened 2025-10-05
.NET: Align skill naming in A2A sample

AndrewNikolin:
### Motivation and Context

Fixing what looks like a simple copy-paste naming issue spotted while exploring this sample


stephentoub (approved):
Thanks

Issue #1228, opened 2025-10-06
.NET: Ignore ChatMessageStore with service thread storage instead of failing


westey-m:
Currently when providing a ChatMessageStore and using a service with server side storage, we fail when trying to set the conversation id, since we are already setting the chat messages store when constructing the thread.

We should delay setting the store, so that it is only set when we know what the service supports.

PR #1229, opened 2025-10-06
.NET: Fix bug where ChatClientAgent throws when providing a ChatMessageStore with a service that requries service storage

westey-m:
### Motivation and Context

#1228

### Description

- Fix bug where ChatClientAgent throws when providing a ChatMessageStore with a service that requires service storage

### Contribution Checklist

### DIFF
--- a/dotnet/tests/Microsoft.Agents.AI.UnitTests/ChatClient/ChatClientAgentTests.cs
+++ b/dotnet/tests/Microsoft.Agents.AI.UnitTests/ChatClient/ChatClientAgentTests.cs
@@ -1786,7 +1786,7 @@ public class ChatClientAgentTests
     /// Verify that RunStreamingAsync uses the ChatMessageStore factory when the chat client returns no conversation id.
     /// </summary>
     [Fact]
-    public async Task RunStreamingAsynccUsesChatMessageStoreWhenNoConversationIdReturnedByChatClientAsync()
+    public async Task RunStreamingAsyncUsesChatMessageStoreWhenNoConversationIdReturnedByChatClientAsync()
     {
         // Arrange
         Mock<IChatClient> mockService = new();


dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
```diff
@@ -322,7 +322,6 @@ public override async IAsyncEnumerable<AgentRunResponseUpdate> RunStreamingAsync
     public override AgentThread GetNewThread()
         => new ChatClientAgentThread
         {
-            MessageStore = this._agentOptions?.ChatMessageStoreFactory?.Invoke(new() { SerializedState = default, JsonSerializerOptions = null }),
```
stephentoub, line 325: Just deleting this is sufficient because we already lazy-initialize it to the exact same thing later?

westey-m: Yes, see UpdateThreadWithTypeAndConversationId below.  It gets set when we get back a response and we know whether the service supports storage or not.
This line should never have been added as well.

crwells:
Is it possible to create a ChatClientAgentThread that has an initialized MessageStore (in memory)?  I have a routing agent, and I want him to have some chat history for context -- but I don't want its response in my real agent thread, because I'm going to have other agents answer the user query.  Ideally I would have the agent create a new thread and I would copy the messages over.

I can't set the MessageStore myself because it's all internal. I could use the deserialization constructor, but I'm going to be embarrassed if anyone sees the way I end up coding that.

I would even settle for a way to derive from ChatClientAgentThread that causes the MessageStore to be initialized.  Is reflecting _messageStore and setting it my only option? Or am I missing something basic?

Issue #1218, opened 2025-10-06
Ollama example with OpenAIChatClient doesn't work


jmiddleton:
I just tried to run the ollama example `ollama_with_openai_chat_client.py` and it seems it needs the OPENAI API KEY. The error I'm getting is:

```
=== Ollama with OpenAI Chat Client Agent Example ===
=== Non-streaming Response Example ===
Traceback (most recent call last):
  File "/Users/jmiddleton/Documents/git/microsoft-agent-framework-sample01/ollama_with_openai_chat_client.py", line 80, in <module>
    asyncio.run(main())
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/asyncio/runners.py", line 194, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/asyncio/base_events.py", line 687, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/Users/jmiddleton/Documents/git/microsoft-agent-framework-sample01/ollama_with_openai_chat_client.py", line 75, in main
    await non_streaming_example()
  File "/Users/jmiddleton/Documents/git/microsoft-agent-framework-sample01/ollama_with_openai_chat_client.py", line 35, in non_streaming_example
    agent = OpenAIChatClient(
            ^^^^^^^^^^^^^^^^^
  File "/Users/jmiddleton/Documents/git/microsoft-agent-framework-sample01/.venv/lib/python3.12/site-packages/agent_framework/openai/_chat_client.py", line 511, in __init__
    raise ServiceInitializationError(
agent_framework.exceptions.ServiceInitializationError: OpenAI API key is required. Set via 'api_key' parameter or 'OPENAI_API_KEY' environment variable.
```

I believe the exception is from the constructor of `OpenAIChatClient` class.

My pyproject.toml is:

```
[project]
name = "microsoft-agent-framework-sample01"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.12"
dependencies = [
    "agent-framework>=1.0.0b251001"
]
```

marswangyang:
[...]

PR #1240, opened 2025-10-06
samples: add api_key for Ollama sample usage

MarsWangyang:
### Motivation and Context

1. Why is this change required? 
- The current sample for Ollama usage is not working because lacking of `api_key` containing in the `OpenAIChatClient()`, which breaks the program.
2. What problem does it solve? 
- It provides a correct quickstart instruction for new users.
3. What scenario does it contribute to?
- When new user want to have a quick sample for Ollama usage case
4. If it fixes an open issue, please link to the issue here.
#1218 

### Description

Ollama provides compatible API endpoints with OpenAI. But the initialization of `OpenAIChatClient()` is mandatory to provide `api_key`, or the assertion won't pass and throw out an exception like the following:
```bash
agent_framework.exceptions.ServiceInitializationError: OpenAI API key is required. Set via 'api_key' parameter or 'OPENAI_API_KEY' environment variable.
```
The solution here is to provide **random** `api_key` for the class, which also follows the official sample on Ollama documentation ([link](https://ollama.com/blog/openai-compatibility)).

### Contribution Checklist

### DIFF
--- a/python/samples/getting_started/agents/ollama/README.md
+++ b/python/samples/getting_started/agents/ollama/README.md
@@ -21,6 +21,9 @@ The examples use environment variables for configuration:
 ### Environment Variables
 
 Set the following environment variables before running the examples:
+- `OLLAMA_API_KEY`: Random string for api key
+  - Example: `export OLLAMA_API_KEY="ollama"`
+  - It is required but unused.
 
 - `OLLAMA_ENDPOINT`: The base URL for your Ollama server
   - Example: `export OLLAMA_ENDPOINT="http://localhost:11434/v1/"`
--- a/python/samples/getting_started/agents/ollama/ollama_with_openai_chat_client.py
+++ b/python/samples/getting_started/agents/ollama/ollama_with_openai_chat_client.py
@@ -33,6 +33,7 @@ async def non_streaming_example() -> None:
     print("=== Non-streaming Response Example ===")
 
     agent = OpenAIChatClient(
+        api_key=os.getenv("OLLAMA_API_KEY"),
         base_url=os.getenv("OLLAMA_ENDPOINT"),
         model_id=os.getenv("OLLAMA_MODEL"),
     ).create_agent(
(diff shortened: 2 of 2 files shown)


python/samples/getting_started/agents/ollama/ollama_with_openai_chat_client.py
```diff
@@ -33,6 +33,7 @@ async def non_streaming_example() -> None:
     print("=== Non-streaming Response Example ===")
 
     agent = OpenAIChatClient(
+        api_key=os.getenv("OLLAMA_API_KEY"),
```
dmytrostruk: nit: I'm not sure it's worth to define a separate environment variable for the value which should be always random. Maybe we can just put the following in the code, I think it should be enough:
```suggestion
        api_key="ollama", # Just a placeholder, Ollama doesn't require API key
```

dmytrostruk (approved):

python/samples/getting_started/agents/ollama/ollama_with_openai_chat_client.py
marswangyang: Agree. That might be much clearer for users with the comment. Thanks

python/samples/getting_started/agents/ollama/ollama_with_openai_chat_client.py
```diff
@@ -52,6 +53,7 @@ async def streaming_example() -> None:
     print("=== Streaming Response Example ===")
 
     agent = OpenAIChatClient(
+        api_key=os.getenv("OLLAMA_API_KEY"),
```
giles17: You can change this too like you did for the non-streaming example
```suggestion
        api_key="ollama" # Just a placeholder, Ollama doesn't require API key,
```

python/samples/getting_started/agents/ollama/README.md
```diff
@@ -21,6 +21,9 @@ The examples use environment variables for configuration:
 ### Environment Variables
 
 Set the following environment variables before running the examples:
+- `OLLAMA_API_KEY`: Random string for api key
+  - Example: `export OLLAMA_API_KEY="ollama"`
+  - It is required but unused.
```
giles17: You can remove this

Issue #1232, opened 2025-10-06
.NET: OpenTelemetry event 'workflow.build' is missing 'workflow.definition' content


weng5e:
The OpenTelemetry event 'workflow.build' has a 'workflow.definition' property that can emit the workflow executors. The value is current missing. The repro sample is 'dotnet\samples\GettingStarted\Workflows\Observability\AspireDashboard'. When configured to emit data into Application Insights, this can be easy exposed.

<img width="3365" height="1723" alt="Image" src="https://github.com/user-attachments/assets/d5ededfe-e0d5-4db2-ae4c-f0264038fe68" />

weng5e:
repro code is here:  https://github.com/microsoft/agent-framework/pull/1231

PR #1261, opened 2025-10-07
.NET: Add missing workflow properties to the workflow.definition tag

SergeyMenshykh:
Closes: https://github.com/microsoft/agent-framework/issues/1232

image


crickman:
Would it make sense to introduce any tests around this?

dotnet/src/Microsoft.Agents.AI.Workflows/EdgeData.cs
```diff
@@ -1,12 +1,17 @@
...
 
 namespace Microsoft.Agents.AI.Workflows;
 
 /// <summary>
 /// A base class for edge data, providing access to the <see cref="EdgeConnection"/> representation of the edge.
 /// </summary>
+[JsonPolymorphic(TypeDiscriminatorPropertyName = "$type")]
+[JsonDerivedType(typeof(FanInEdgeData), typeDiscriminator: "fanIn")]
+[JsonDerivedType(typeof(FanOutEdgeData), typeDiscriminator: "fanOut")]
+[JsonDerivedType(typeof(DirectEdgeData), typeDiscriminator: "direct")]
```
crickman: Is this something that will create divergence with Python (a la ChatHistory serialization)

SergeyMenshykh: Can the edge and edge data end up in the chat history?

SergeyMenshykh: In any case, if it ends up in chat history, then all required edge data information will be preserved, so it will be possible to deserialize it back to the right type and not throw an exception which most likely will be thrown without this code. If we target chat history compatibility between .NET and Python, then we need a similar change on the Python side.

lokitoth: Is there a reason not to use the `WorkflowInfo` series of classes? They provide much of the same data, and are designed to be a faithful and round-trippable representation of the Workflow. They are also tested for round-trip serialization correctness and are used to validate the correctness of a candidate Workflow during checkpoint restore.

SergeyMenshykh: Good idea, I've updated this PR to use workflow info classes.
CC: @TaoChenOSU

TaoChenOSU (approved):

TaoChenOSU (approved):

Issue #1057, opened 2025-10-01
.NET feedback: Structured output feels more cumbersome in AF + Have issues with Enums in result object


rwjdk:
(Based on how it is done in the sample: Agent_Step05_StructuredOutput)

After playing around with structured Output in AF, it feels like a step down from MEAI and SK

1. MEAI feels like it has the best solution, as you can do something like this:

```cs
string question = "What is the capital of Denmark and what is its population?";
ChatResponse<City> structuredOutputResponse = await chatClient.GetResponseAsync<City>(question);
City result = structuredOutputResponse.Result;

//(AF do not have a similar agent.RunAsync<City>(question)) ☹️
```

2. SK was a bit more cumbersome than MEAI with the need to set the typeof(<object>) and manual deserialization

```cs
ChatCompletionAgent agent = new()
{
    Kernel = kernel,
    Instructions = "You are a City Expert",
    Arguments = new KernelArguments(new AzureOpenAIPromptExecutionSettings
    {
        ResponseFormat = typeof(City)
    })
};

...

await foreach (AgentResponseItem<ChatMessageContent> response in agent.InvokeAsync(message))
{
    string json = response.Message.Content!;
    City? city = JsonSerializer.Deserialize<City>(json, new JsonSerializerOptions
    {
        Converters = { new JsonStringEnumConverter() }
    });
}
```

3. AF, however, feels even more cumbersome than SK
```cs
ChatOptions = new()
{
    ResponseFormat = ChatResponseFormat.ForJsonSchema<City>()
}
```

...And if the Structured Output object has Enums, you need the following JSON Serializer options for it to work in AF

```cs
JsonSerializerOptions jsonSerializerOptions = new()
{
    TypeInfoResolver = new DefaultJsonTypeInfoResolver(),
    PropertyNameCaseInsensitive = true,
    Converters = { new JsonStringEnumConverter() }
};
```

(It even seems to generate the incorrect schema if Enums are inside nested objects in the Structured output object, giving back non-existing enum values as the schema does not generate the enum valid values)

---

It would be cool if AF could have the MEAI approach...

---

For context, here is the City and Underlying Enum

```cs
public class City
{
    public required string NameOfCapital { get; set; }
    public required string Population { get; set; }
    public required Size SizeOfCity { get; set; }
}

public enum Size
{
    RandomValue1, //values AI can't guess randomly
    RandomValue2 //values AI can't guess randomly
}
```

SergeyMenshykh:
[...]

PR #1267, opened 2025-10-07
.NET: Support enum serialization as strings

SergeyMenshykh:
This PR enables the serialization of enums to strings for the chat client agent to address the issue: https://github.com/microsoft/agent-framework/issues/1057. 

It also updates the chat client agent JSON utilities to enable relaxed JSON escaping, making them consistent with AgentAbstractionsJsonUtilities and MEAI.AIJsonUtilities.


stephentoub:
This is fine... but it seems like https://github.com/microsoft/agent-framework/blob/1d27b5767242127bcfad59d82b5fe6139f9ab933/dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgentStructuredOutput.cs#L149 should just be referencing AgentAbstractionJsonUtilities.DefaultOptions rather than (the internal) AgentJsonUtilities.DefaultOptions, and that would also fix this, yes?

stephentoub (approved):

SergeyMenshykh:
> This is fine... but it seems like
> 
> https://github.com/microsoft/agent-framework/blob/1d27b5767242127bcfad59d82b5fe6139f9ab933/dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgentStructuredOutput.cs#L149
> 
> should just be referencing AgentAbstractionJsonUtilities.DefaultOptions rather than (the internal) AgentJsonUtilities.DefaultOptions, and that would also fix this, yes?

Yep, that would fix the problem.

Issue #1266, opened 2025-10-07
.NET: Sequential orchestration passes entire context


Ryan-Palmer:
I have two agents.
1. Vision capable, extracts content from screenshots and returns a structured 'Document Chunks' model.
2. Reasoning capable, finds required info in the doc chunks and returns as a structured domain model.

I only want the structured output of 1 fed into 2.

In Semantic Kernel I used the Sequential Orchestration to plug one into the next, which worked great.

Using the Agent Framework's version of Sequential Orchestration this fails, as 1 is feeding the entire history into 2, including the binary image content, but 2 isn't a vision capable model.

From the description in the AF docs, it sounds like Handoff should pass the full history, but Sequential should pass the direct output only?

Ryan-Palmer:
I think this is the line (`includeInputInOutput`) https://github.com/microsoft/agent-framework/blob/714e7b50d4bd740048c8c3c23044450853d81353/dotnet/src/Microsoft.Agents.AI.Workflows/AgentWorkflowBuilder.cs#L38

Ryan-Palmer:
A simple single-edge workflow works as expected.

Examples (in F#)

Not working:
```
let sequentialWorkflow = 
    [ docExtractionAgent; researchAgent ]
    |> AgentWorkflowBuilder.BuildSequential
```

Working:
```
let sequentialWorkflow =
    WorkflowBuilder(docExtractionAgent)
        .AddEdge(docExtractionAgent, researchAgent)
        .Build();
```

stephentoub:
> From the description in the AF docs, it sounds like Handoff should pass the full history, but Sequential should pass the direct output only?

Can you point to those docs?

There are valid scenarios for both. I think it's reasonable for it to be configurable. The current choice is simply because if you pass everything, and agent can choose to ignore anything it wants, but if you only pass part of it, an agent has no chance of recovering what wasn't provided.

Ryan-Palmer:
> Can you point to those docs?

Sure, here you go:

Sequential:

>In sequential orchestration, agents are organized in a pipeline where each agent processes the task in turn, passing output to the next agent in the sequence.

>Key Concepts
Sequential Processing: Each agent processes the output of the previous agent in order

https://learn.microsoft.com/en-us/agent-framework/user-guide/workflows/orchestrations/sequential

Handoff:

>Handoff orchestration allows agents to transfer control to one another based on the context or user request. Each agent can "handoff" the conversation to another agent

>Key Concepts
[...]

PR #1276, opened 2025-10-07
.NET: Add AgentWorkflowBuilder.CreateSequentialBuilder

stephentoub:
Enables each agent's output to be customized via a filter, to accomodate the request that an agent be able to forward all input+output, just output, or something else. I debated just having it be a bool, but opted to instead go the slightly more complicated but more flexible route of a delegate. We could go back to the bool, or have both, or have neither and just have this be when you fall off the cliff to using WorkflowBuilder directly. Once we improve the built-in AIAgentHost to accomodate the affordances these agent workflow's AgentRunExecutor has, maybe we could get rid of at least sequential workflow, since that one is fairly straightforward with WorkflowBuilder. I'm still not in love with AgentWorkflowBuilder, but I don't have better thoughts at the moment on how to expose these patterns in a simple way. Ideally there'd be a really clean grow-up story that let's you incrementally bring in more workflow stuff.

Closes https://github.com/microsoft/agent-framework/issues/1266


dotnet/src/Microsoft.Agents.AI.Workflows/AgentWorkflowBuilder.cs
```diff
@@ -18,55 +18,125 @@ namespace Microsoft.Agents.AI.Workflows;
...
+        /// An optional callback that will be invoked to determine the output messages to forward to the next step in the workflow.
+        /// The function accepts the list of messages provided as input to this step and the list of messages produced by the agent,
+        /// and returns the list of messages to forward. If <see langword="null"/>, the default behavior is to forward the concatenation
+        /// of all input and output messages.
+        /// </param>
+        /// <returns>The current <see cref="SequentialWorkflowBuilder"/> instance.</returns>
+        /// <exception cref="ArgumentNullException"><paramref name="agent"/> is <see langword="null"/>.</exception>
+        public SequentialWorkflowBuilder Append(
+            AIAgent agent,
+            Func<IList<ChatMessage>, IList<ChatMessage>, IList<ChatMessage>>? outputFilter = null)
```
lokitoth, line 85: I think this shows a gap in being able to easily create two routing graphs: One for `ChatMessage`s and one for `TurnToken`s. Maybe worth thinking about a higher-level `WorkflowBuilder` that specializes in that?

We should consider adding a set of "well known" `OutputFilters` (in this case `ForwardAll`, `ForwardNew`). I suspect most people will also want "forward the original input messages, plus the just the previous agent's output" some portion of the time too.

dotnet/src/Microsoft.Agents.AI.Workflows/AgentWorkflowBuilder.cs
```diff
@@ -18,55 +18,125 @@ namespace Microsoft.Agents.AI.Workflows;
...
     /// </summary>
     /// <param name="agents">The sequence of agents to compose into a sequential workflow.</param>
     /// <returns>The built workflow composed of the supplied <paramref name="agents"/>, in the order in which they were yielded from the source.</returns>
+    /// <remarks>
+    /// Each agent in the sequence receives the full conversation history as input, with each agent outputting to the next its input and output messages.
+    /// If more customized behavior is desired, such as to only forward the output messages of an agent to the next, use <see cref="CreateSequentialBuilder"/>.
+    /// Each agent is added to the builder individually with <see cref="SequentialWorkflowBuilder.Append"/>, which accepts an optional output filter function
+    /// for controlling exactly what messages are propagated.
+    /// </remarks>
     public static Workflow BuildSequential(params IEnumerable<AIAgent> agents)
```
lokitoth, line 40: Thinking of your overall comment about `AgentWorkflowBuilder`, and the way the one-liners work right now, maybe it makes sense for some of the simpler ones to be classes that remove the need to drive the `WorkflowBuilder` at all (more akin to the original Orchestrations implementation), but also support the more parametrized forms or fluent forms for those that one to dive a layer deeper?

Then we can have the Builder-forms operate more like `.AddSwitch()` and be able to be built into a workflow, obviating the need to host the orchestration as a subworkflow.

dotnet/src/Microsoft.Agents.AI.Workflows/AgentWorkflowBuilder.cs
stephentoub: > Maybe worth thinking about a higher-level WorkflowBuilder that specializes in that?

We could certainly think about it... not sure exactly what it would look like.

> "forward the original input messages, plus the just the previous agent's output"

What's in this PR doesn't easily accomodate that. An agent just receives the previous agent's forwarded messages and doesn't have a way to distinguish which were output from the agent, which were input to the agent, and within the input where they came from. You might be able to achieve it with knowledge of your workflow, using AuthorName to distinguish, but that's not something we could do automatically. If this is a case we want to handle, it'd require a more complicated solution.

dotnet/src/Microsoft.Agents.AI.Workflows/AgentWorkflowBuilder.cs
stephentoub: We should work through what that would look like.

PR #1298, opened 2025-10-08
.NET: Change MAAI and MAAI.A unit tests to use source generation for .net core serialization

westey-m:
### Motivation and Context

By default, reflection is used, which means that misconfiguration of source generation may not be caught by tests.
The .net framework version of the tests still use reflection, so both paths are covered.

### Description

- Change JsonSerializerIsReflectionEnabledByDefault to false for test projects.
- Fix unit tests to use JsonSerializerContext where needed
- Disable reflection only tests for .net core

### Contribution Checklist


dotnet/tests/Microsoft.Agents.AI.Abstractions.UnitTests/ServiceIdAgentThreadTests.cs
```diff
@@ -36,7 +36,8 @@ public void Constructor_WithServiceThreadId_SetsProperty()
     public void Constructor_WithSerializedId_SetsProperty()
     {
         // Arrange
-        var json = JsonSerializer.SerializeToElement(new { ServiceThreadId = "service-id-456" });
+        var serviceThreadWrapper = new ServiceIdAgentThread.ServiceIdAgentThreadState { ServiceThreadId = "service-id-456" };
```
stephentoub, line 39: I'm unclear what's being tested here. Previously this test wasn't doing anything with internal state, but now it's needing to use an internal type from the library, and that type needs to be added to the test's JsonSerializerContext?

westey-m: It's testing that the base `ServiceIdAgentThread` can deserialize a `ServiceIdAgentThreadState` object correctly.
On line 43, we are not passing in any `JsonSerializerOptions`, so the `ServiceIdAgentThread` is using it's own built-in `JsonSerializerOptions` to deserialize, and it should just work out of the box.

PR #1304, opened 2025-10-08
.NET: Add support for getting and creating Assistant and Foundry agents with ChatClientAgentOptions

westey-m:
### Motivation and Context

#1226 

We don't have extensions to consistently get and create Foundry Agents and Assistants using ChatClientAgentOptions, which means that adding chat message store factories or ai context provider factories is not always possible.

### Description

- Add missing get and create overloads with ChatClientAgentOptions.

### Contribution Checklist


markwallace-microsoft (approved):

dotnet/src/Microsoft.Agents.AI.AzureAI/PersistentAgentsClientExtensions.cs
```diff
@@ -234,4 +375,134 @@ public static ChatClientAgent CreateAIAgent(
...
+        {
+            foreach (AITool tool in options.ChatOptions.Tools)
+            {
+                switch (tool)
+                {
+                    case HostedCodeInterpreterTool:
+                        var codeInterpreterToolDefinition = new CodeInterpreterToolDefinition();
+                        persistentAgentTools.Add(codeInterpreterToolDefinition);
+                        break;
+                }
```
rogerbarreto: We may have other tools to add as a support in here.

See: 
https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/ai/Azure.AI.Agents.Persistent/src/Custom/PersistentAgentsChatClient.cs#L334-L374

dotnet/src/Microsoft.Agents.AI.OpenAI/Extensions/OpenAIAssistantClientExtensions.cs
```diff
@@ -143,6 +143,155 @@ public static async Task<ChatClientAgent> GetAIAgentAsync(
...
+    public static ChatClientAgent GetAIAgent(
+        this AssistantClient assistantClient,
+        ClientResult<Assistant> assistantClientResult,
+        ChatClientAgentOptions options,
+        Func<IChatClient, IChatClient>? clientFactory = null)
+    {
+        if (assistantClientResult is null)
+        {
+            throw new ArgumentNullException(nameof(assistantClientResult));
+        }
```
rogerbarreto, line 164:
```suggestion
        if (assistantClient is null)
        {
            throw new ArgumentNullException(nameof(assistantClient));
        }

        if (assistantClientResult is null)
        {
            throw new ArgumentNullException(nameof(assistantClientResult));
        }
```

rogerbarreto (approved):

dotnet/src/Microsoft.Agents.AI.OpenAI/Extensions/OpenAIAssistantClientExtensions.cs
westey-m: This check is already done on line 189.  No need to repeat it here.

dotnet/src/Microsoft.Agents.AI.AzureAI/PersistentAgentsClientExtensions.cs
```diff
@@ -234,4 +375,134 @@ public static ChatClientAgent CreateAIAgent(
...
+        if (options.ChatOptions?.Tools is not null)
+        {
+            foreach (AITool tool in options.ChatOptions.Tools)
+            {
+                switch (tool)
+                {
+                    case HostedCodeInterpreterTool:
+                        var codeInterpreterToolDefinition = new CodeInterpreterToolDefinition();
+                        persistentAgentTools.Add(codeInterpreterToolDefinition);
+                        break;
```
markwallace-microsoft: What about the other tools types e.g. file search, web search and mcp. 

We also need to handle creating the tool resources. I have some of this code written for declarative agent support so I'm OK with merging this and then I can pull out the pieces that can be reused here as a seperate PR.

dotnet/src/Microsoft.Agents.AI.AzureAI/PersistentAgentsClientExtensions.cs
westey-m: Thanks, added more tools, plus integration tests for those.

dotnet/src/Microsoft.Agents.AI.AzureAI/PersistentAgentsClientExtensions.cs
westey-m: I've added the other tools support already with integration tests.

dotnet/tests/AzureAIAgentsPersistent.IntegrationTests/AzureAIAgentsPersistentCreateTests.cs
```diff
@@ -0,0 +1,263 @@
...
+                toolResources: new ToolResources() { CodeInterpreter = toolResource }),
+            "CreateWithFoundryOptionsSync" => this._persistentAgentsClient.CreateAIAgent(
+                s_config.DeploymentName,
+                instructions: AgentInstructions,
+                tools: [new CodeInterpreterToolDefinition()],
+                toolResources: new ToolResources() { CodeInterpreter = toolResource }),
+            _ => throw new InvalidOperationException($"Unknown create mechanism: {createMechanism}")
+        };
+
+        try
```
rogerbarreto, line 204: Suggest adding all this block in the `try`.

Or move the File.Delete outside of the `finally`.

Same for other tests involving file creation/clean up

westey-m: Not sure I get what you are suggesting or why it is better.

rogerbarreto (approved):

dotnet/tests/AzureAIAgentsPersistent.IntegrationTests/AzureAIAgentsPersistentCreateTests.cs
rogerbarreto: Normally we add resource clean up in finally for resources created in the `try`, thats all.

There are some very specific scenarios where the file can succeed being created but fail when writing the text.

Scenarios that cause this

- Disk full after file creation but before data write.
- IOException (e.g., sudden network drop if path is on a share).
- Encoding failure converting the string to bytes.
- Anti-virus or OS lock interrupting mid-write.

westey-m: But here I don't expect the create to fail, rather I expect the run assert to fail.  So the cleanup is to release resources that were created already before the failure.  Also, there wouldn't be much point in cleaning up resources that were never created since creating them failed.

SergeyMenshykh (approved):

PR #1333, opened 2025-10-09
.NET: Bump Bedrock package version

stephentoub:


markwallace-microsoft:
We have dependabot running weekly so it also made this update

PR #1365, opened 2025-10-09
Python: Added function approval example with streaming

dmytrostruk:
### Motivation and Context

Updated existing example with function approval to showcase streaming scenario.

### Contribution Checklist


ekzhu:
I think it would be good to also show how to checkpoint the state of the agent in between producing approval request and handling response.

PR #1367, opened 2025-10-09
.NET: fix: Make State Persistence APIs work better with PortableValue

### LINKED ISSUE #1361: .NET: PortableValue type-cast cache does not do the right then when first cast is to object
It treats the cast as successful and after that fails to cast to more concrete types, e.g. string, even when the cast should work.

lokitoth:
### Motivation and Context

With the changes to make our type system work a better with transparent serialization round-tripping during checkpointing, we replaced the `object`-based catch-all in `MessageRouter` and `RouteBuilder` with one taking `PortableValue`, since `object` as a target-type can result in truncation or leaking `JsonElement` objects when interacting with JSON serialization.

We did not make a similar change to state persistence, which means that it is unclear how to get a "supertype" of everything to operate on when a specific type is not known. Users could get into trouble writing arbitrary types and trying to read back `object`. They could use `PortableValue` but this caues

It also exposed a pair of bugs:
* in cases like the abstraction leak detailed above, the deserialization process caches the bad leaked value and makes it impossible to cleanly deserialize as the correct value afterwards. 
* in cases where the user asks for the wrong type backed by JSON via `IDelayedDeserialization` (and would leak if requested to convert to `object`), `PortableValue` throws out of `Is<T>`/`As<T>`.

### Description

* QOL improvement for developer: Now loading a state as `PortableValue` will correctly wrap non-PV objects in it, but also avoid double-wrapping if it was stored as a PV in the first place.
* Fix the two bugs in PortableValue

Closes #1361

### Contribution Checklist

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.Workflows/Execution/StateManager.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Workflows/Execution/StateManager.cs
@@ -102,7 +102,8 @@ internal sealed class StateManager
         if (typeof(T) == typeof(object))
         {
             // Reading as object will break across serialize/deserialize boundaries, e.g. checkpointing, distributed runtime, etc.
-            throw new NotSupportedException("Reading state as 'object' is not supported. Use 'PortableValue' instead for variants.");
+            // Disabled pending upstream updates for this change; see https://github.com/microsoft/agent-framework/issues/1369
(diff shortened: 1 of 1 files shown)


lokitoth:
PR failed to merge because IT fails due to breaking change on `.ReadStateAsync<object>()`. Will remove the check and re-add later.

Issue #1344, opened 2025-10-09
Bug: AIAgent names are not allowed to be longer than 20 characters, when used in a Handsoff-Orchestration-Workflow.


nx-s:
In .NET:

When an AIAgent is build by a ChatClient and used afterwards in an Handoff-workflow, the `name` of the agent cannot be longer than 20 characters.
Otherwise an exception is thrown: `HTTP 400 (invalid_request_error: string_above_max_length)`

stacktrace:
```
Invalid 'messages[2].tool_calls[0].function.name': string too long. Expected a string with maximum length 64, but got a string with length 66 instead.
   at OpenAI.ClientPipelineExtensions.ProcessMessageAsync(ClientPipeline pipeline, PipelineMessage message, RequestOptions options)
   at OpenAI.Chat.ChatClient.CompleteChatAsync(BinaryContent content, RequestOptions options)
   at OpenAI.Chat.ChatClient.<>c__DisplayClass19_0.<<CompleteChatStreamingAsync>b__0>d.MoveNext()
```

Assumed Reason:
When the agent is invoked in an orchestration workflow, then the agent name is renamed as follows:

1) all `-` (dashes) are replaced by `_` (underscores)
2) the agent name is prefixed with the workflow type, i.e, `handoff_to_`
3) the agent name is suffixed with `_agent_<some_GUID>` (e.g. agent_52f607cf61294d3ea48a69778f1d2f48)

Example:
```
{ TurnToken = Microsoft.Agents.AI.Workflows.TurnToken, InvokedHandoff = handoff_to_my_very_special_agent_52f607cf61294d3ea48a69778f1d2f48, Messages = System.Collections.Generic.List`1[Microsoft.Extensions.AI.ChatMessage] })
```

The length of the `InvokedHandoff`-name seems to match the too-long length stated in the error message.

PR #1370, opened 2025-10-09
.NET: Fix handoff function naming

stephentoub:
We don't need the agent's name or a guid in the handoff name... we can just use simple numbering. There's a possibility that someone built an agent with a built-in function tool named "handoff_to_x"; if that turns out to be an issue, we could add back some longer bit of randomness.

Fixes https://github.com/microsoft/agent-framework/issues/1344

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/HandoffAgentExecutor.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/HandoffAgentExecutor.cs
@@ -47,7 +47,7 @@ internal sealed class HandoffAgentExecutor(
                 foreach (HandoffTarget handoff in handoffs)
                 {
                     index++;
-                    var handoffFunc = AIFunctionFactory.CreateDeclaration($"{FunctionPrefix}{index}", handoff.Reason, s_handoffSchema);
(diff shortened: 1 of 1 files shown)


ReubenBond (approved):

dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/HandoffAgentExecutor.cs
```diff
@@ -43,9 +43,11 @@ public void Initialize(
                     },
                 };
 
+                int index = 0;
                 foreach (HandoffTarget handoff in handoffs)
                 {
-                    var handoffFunc = AIFunctionFactory.CreateDeclaration($"{HandoffsWorkflowBuilder.FunctionPrefix}{handoff.Target.GetDescriptiveId()}", handoff.Reason, s_handoffSchema);
+                    index++;
+                    var handoffFunc = AIFunctionFactory.CreateDeclaration($"{FunctionPrefix}{index}", handoff.Reason, s_handoffSchema);
```
stephentoub: Bad merge
```suggestion
                    var handoffFunc = AIFunctionFactory.CreateDeclaration($"{HandoffsWorkflowBuilder.FunctionPrefix}{index}", handoff.Reason, s_handoffSchema);
```

crickman (approved):

PR #1382, opened 2025-10-10
.NET: Duplicate System Instruction when creating a ChatClientAgentOptions instance with the constructor

htilssu:
#### 1. Why is this change required?
Creating a `ChatClientAgentOptions` instance with the constructor that accepts only the `instruction` argument caused the instruction text to be duplicated internally when the agent options were later merged into chat options.

#### 2. What problem does it solve?
This change removes redundant instruction concatenation logic to ensure that `Instructions` are applied only once.  
It prevents recursive or repeated values being passed into `ChatOptions` during agent initialization.

#### 3. What scenario does it contribute to?
Improves consistency and correctness when initializing `ChatClientAgent` instances with predefined instructions, especially in scenarios where low-level options are already configured through `CreateConfiguredChatOptions()`.

### Motivation and Context

When `ChatClientAgent.RunAsync()` is called with `requestOptions == null`, 
the logic at  
https://github.com/microsoft/agent-framework/blob/7e1fd67e7660ea2d2a5d45f7076dae690f4948d1/dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs#L609-L613  
manually concatenates `this.Instructions` with `chatOptions.Instructions`.

However, `this.Instructions` is just a proxy for  
`_agentOptions.Instructions`   

https://github.com/microsoft/agent-framework/blob/7e1fd67e7660ea2d2a5d45f7076dae690f4948d1/dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs#L143-L145

and during `ChatClientAgentOptions` construction, the same instruction value is already 
propagated into its `ChatOptions` field:  
https://github.com/microsoft/agent-framework/blob/7e1fd67e7660ea2d2a5d45f7076dae690f4948d1/dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgentOptions.cs#L48-L51  

### Contribution Checklist

### DIFF
--- a/dotnet/tests/Microsoft.Agents.AI.UnitTests/ChatClient/ChatClientAgentOptionsTests.cs
+++ b/dotnet/tests/Microsoft.Agents.AI.UnitTests/ChatClient/ChatClientAgentOptionsTests.cs
@@ -61,9 +61,7 @@ public class ChatClientAgentOptionsTests
         Assert.Null(options.Name);
         Assert.Equal(Instructions, options.Instructions);
         Assert.Null(options.Description);
-        Assert.NotNull(options.ChatOptions);
-        Assert.Equal(Instructions, options.ChatOptions.Instructions);
-        Assert.Null(options.ChatOptions.Tools);
+        Assert.Null(options.ChatOptions);
     }
 
     [Fact]
@@ -107,7 +105,7 @@ public class ChatClientAgentOptionsTests
         Assert.Equal(Instructions, options.Instructions);
         Assert.Null(options.Description);
         Assert.NotNull(options.ChatOptions);
-        Assert.Equal(Instructions, options.ChatOptions.Instructions);
+        Assert.Null(options.ChatOptions.Instructions);
         Assert.Same(tools, options.ChatOptions.Tools);
     }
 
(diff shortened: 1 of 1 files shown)


htilssu:
I noticed that six tests are currently failing because they depend on the code in constructor I just removed.
Would you prefer that I update those tests to match the new implementation, or remove them if they’re no longer applicable?

westey-m:
> I noticed that six tests are currently failing because they depend on the code in constructor I just removed. Would you prefer that I update those tests to match the new implementation, or remove them if they’re no longer applicable?

@htilssu, thanks for the contribution!
Yes, please update those tests to match the new implementation.  From what I can tell they should still be applicable, just the part that checks for instructions in both the ChatClientAgentOptions and ChatOptions is wrong.

htilssu:
> @htilssu, thanks for the contribution! Yes, please update those tests to match the new implementation. From what I can tell they should still be applicable, just the part that checks for instructions in both the ChatClientAgentOptions and ChatOptions is wrong.

@westey-m Thanks for the feedback! I've updated the unit tests accordingly to align with the new implementation.

htilssu:
@TaoChenOSU  Could you please review the changes in this PR when you have a moment?

Issue #1413, opened 2025-10-12
.NET: Agent AuthorName not preserved during group chat role-switching


ltwlf:
## Summary

In group chat workflows using `AgentWorkflowBuilder`, agent responses lose their original `AuthorName` during role-switching between `ChatRole.Assistant` and `ChatRole.User`. This causes routing failures and incorrect message attribution in multi-agent conversations.

## Reproduction

**Note:** This is a timing-dependent issue that may not reproduce consistently. It appears to be a race condition in how `ChatMessage` objects are modified during role-switching.

1. Create a group chat workflow with multiple agents (5+ agents increases likelihood)
2. Implement custom `GroupChatManager` that checks `message.AuthorName` for routing
3. Run workflow with multiple sequential agent responses
4. Intermittently, `SelectNextAgentAsync` receives messages with incorrect `AuthorName` values
5. Messages that should have `AuthorName="agentA"` instead show `AuthorName="facilitator"` or other incorrect values

## Expected Behavior

Agent responses should maintain their original AuthorName throughout the workflow, regardless of role-switching operations.

## Actual Behavior

- `AuthorName` is set from agent's `Name` property during initial response
- Role-switching changes `Role` but loses `AuthorName`
- Restored messages have correct `Role` but `AuthorName` remains incorrect or null
- Causes routing failures when checking message author for turn-taking logic
- **Timing-dependent:** Appears to be a race condition where message modifications aren't fully synchronized
- **Inconsistent:** Same workflow may work correctly on some runs, fail on others

## Impact

- Custom routing managers cannot reliably identify message authors
- Multi-agent conversations fail or route incorrectly
- Conversation history shows incorrect attribution
- Affects any workflow using `AgentWorkflowBuilder.CreateGroupChatBuilderWith()`

## Additional Notes

**Debugging:** The issue is difficult to trace because:

- Streaming events (`AgentRunUpdateEvent`) show correct `AuthorName` values
- But `ChatMessage` history passed to `SelectNextAgentAsync` shows incorrect values
- Suggests the issue is in how messages are modified/restored between event emission and history building
- Adding debug logging to `ChangeAssistantToUserForOtherParticipants` and `ResetUserToAssistantForChangedRoles` helps identify when attribution is lost

Related issues discovered:

- Agent name comparison uses case-sensitive string matching (should be case-insensitive)
- `DisplayName` property used for routing instead of stable `Name` property

I'm preparing a PR with fixes for these issues.

[...]

PR #1427, opened 2025-10-13
.NET Fix: AuthorName misattribution in multi-agent streaming (null MessageId handling)

ltwlf:
In multi-agent workflows (sequential/handoff), assistant messages get the AuthorName of the next agent. Root cause: null-MessageId streaming updates are appended to the previous message during reconstruction, and last-write-wins metadata (AuthorName) overwrites the prior message.

## Where it happens
- Agent sets author on every streaming update (expected):
  - `agent-framework/dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs` → `RunStreamingAsync`
  ```csharp
  if (update is not null)
  {
      update.AuthorName ??= this.Name; // all updates, including reasoning/non-message deltas
  }
  ```
- Workflow reconstructs messages from updates:
  - `agent-framework/dotnet/src/Microsoft.Agents.AI.Workflows/AgentWorkflowBuilder.cs` → AgentRunStreamingExecutor
  ```csharp
  // before fix
  messages.AddRange(updates.ToAgentRunResponse().Messages);
  ```
  - `ToAgentRunResponse()` (Microsoft.Extensions.AI) uses MessageId for boundaries; null MessageId updates are treated as part of the previous message, polluting metadata.

## Fix (workflow layer)
Filter out updates without MessageId before reconstruction:
```csharp
var filtered = updates.Where(u => !string.IsNullOrEmpty(u.MessageId)).ToList();
if (filtered.Count > 0)
{
    messages.AddRange(filtered.ToAgentRunResponse().Messages);
}
```

Fixes issue #1413

See also https://github.com/dotnet/extensions/pull/6910

### DIFF
--- a/dotnet/tests/Microsoft.Agents.AI.Workflows.UnitTests/AgentWorkflowBuilderTests.cs
+++ b/dotnet/tests/Microsoft.Agents.AI.Workflows.UnitTests/AgentWorkflowBuilderTests.cs
@@ -26,6 +26,24 @@ public class AgentWorkflowBuilderTests
     }
 
     [Fact]
+    public async Task Workflows_FilterNullMessageIdUpdates_PreventsIncorrectAuthorAsync()
+    {
+        // Test that null MessageId updates are filtered out during message reconstruction in all workflow types
+        // This prevents reasoning/metadata tokens from being incorrectly appended to previous messages
+        var agent = new ChatClientAgent(new MockChatClientWithNullMessageIds(), name: "testAgent");
+        var workflow = AgentWorkflowBuilder.BuildSequential([agent]);
+
+        (string updateText, List<ChatMessage>? result) = await RunWorkflowAsync(workflow, [new ChatMessage(ChatRole.User, "test")]);
+
+        Assert.NotNull(result);
+        Assert.Equal(2, result.Count); // User message + 1 assistant message (2 null-MessageId updates filtered out)
+
+        Assert.Equal(ChatRole.Assistant, result[1].Role);
+        Assert.Equal("testAgent", result[1].AuthorName);
+        Assert.Equal("validContent", result[1].Text);
+    }
+
+    [Fact]
     public void BuildConcurrent_InvalidArguments_Throws()
     {
         Assert.Throws<ArgumentNullException>("agents", () => AgentWorkflowBuilder.BuildConcurrent(null!));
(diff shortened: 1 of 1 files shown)


ltwlf:
@stephentoub this is my proposed fix. I checked https://github.com/dotnet/extensions/pull/6910 and I think it should also solve the issue, because it handles message boundaries much better. This would be my favorite - but would be good to have a fix fast.
What do you think?

dotnet/src/Microsoft.Agents.AI.Workflows/AgentWorkflowBuilder.cs
```diff
@@ -877,6 +892,16 @@ private static void ResetUserToAssistantForChangedRoles(List<ChatMessage>? roleC
...
 
+    /// <summary>
+    /// Filters streaming updates to include only those with valid MessageId values.
+    /// This prevents null-MessageId updates (reasoning tokens, metadata) from being incorrectly
+    /// appended to previous messages during ToAgentRunResponse reconstruction.
+    /// </summary>
+    /// <param name="updates">The collection of streaming updates to filter.</param>
+    /// <returns>A list containing only updates with non-empty MessageId values.</returns>
+    private static List<AgentRunResponseUpdate> FilterUpdatesWithMessageId(IEnumerable<AgentRunResponseUpdate> updates) =>
+        updates.Where(u => !string.IsNullOrEmpty(u.MessageId)).ToList();
```
stephentoub, line 903: Thank you, but this is not a valid solution. The similar check in AIAgentHost needs to be removed after discussion, and one of the reasons this Executor was needed was to avoid that check.

stephentoub:
> @stephentoub this is my proposed fix. I checked [dotnet/extensions#6910](https://github.com/dotnet/extensions/pull/6910) and I think it should also solve the issue, because it handles message boundaries much better. This would be my favorite - but would be good to have a fix fast. What do you think?

Thanks, but it is not a valid fix. It will break other things. Updates do not always have a message id. The fix is to wait until the new build of M.E.AI is published, which should be tomorrow.

ltwlf:
Tomorrow is great - I wish I would have seen your PR in AI Extension earlier. 
But on the positve site, I get more and more familiar with the internals of Agent Framework :)

Issue #1137, opened 2025-10-02
.NET sample whatif it's NOT in Azure?


fboucher:
As dev I'm using AI a bit everywhere, not only Azure. Is it supose to work just the same let's say by switch the `AzureOpenAIClient` to `OpenAIClient` and specifying my Endpoint?

ex:
```csharp

// Create the agent using Azure OpenAI.
// AIAgent agent = new AzureOpenAIClient(
//     new Uri(endpoint),
//     new AzureCliCredential())
//         .GetChatClient(deploymentName)
//         .CreateAIAgent(agentOptions);

AIAgent agent = new OpenAIClient(new ApiKeyCredential(API_KEY), new OpenAIClientOptions { Endpoint = new Uri(endpoint) })
                    .GetChatClient(deploymentName)
                    .CreateAIAgent(agentOptions);

var response = await agent.RunAsync("Please provide information about John Smith, who is a 35-year-old software engineer.");
```

stephentoub:
There's no need for it to be Azure at all. AIAgent is just an abstract base class, which could be backed by anything. And ChatClientAgent, as one of the most prominent derivations, is just backed by an IChatClient, which has implementations for OpenAI, AWS, Gemini, Anthropic, ONNX Runtime GenAI, Hugging Face, and so on.

https://github.com/microsoft/agent-framework/tree/main/dotnet/samples/GettingStarted/AgentProviders has examples with various providers.

And https://github.com/microsoft/agent-framework/blob/main/dotnet/samples/GettingStarted/Workflows/_Foundational/05_MultiModelService/Program.cs provides an example of multiple providers working together in a workflow.

In your example, yes, you can use OpenAIClient, with whatever Endpoint you want.

ghost:
Hello, Is there any Dotnet samples with Gemini client?

mikeholczer:
I'm having mixed results using a IChatClient implementation from the Anthropic.SDK nuget package. It's mostly works, but when a TurnToken is sent to the agent based on it, it's causing a call to GetStreamingResponseAsync with an empty list of messages which ends up causing an error.

When I switch to using a OpenAI based IChatClient backed by GitHub Models that call to GetStreamingResponseAsync doesn't happen.

I'm going to try to do some more debugging on Monday, but curious if anyone else is seeing something similar. 

I'll plan to either post my findings or at least sample code to reproduce the issue here early next week.

stephentoub:
[...]

PR #1439, opened 2025-10-13
.NET: AIAgentHostExecutor to use ToAgentRunResponse

TaoChenOSU:
### Motivation and Context

Addressed https://github.com/microsoft/agent-framework/issues/1137

### Description
Use the `ToAgentRunResponse` extension method to assemble streaming updates into fully-formed updates. This will avoid unintended bugs caused by differences between underlying clients/services driving the agents. 

### Contribution Checklist


dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/AIAgentHostExecutor.cs
```diff
@@ -74,36 +67,9 @@ protected override async ValueTask TakeTurnAsync(List<ChatMessage> messages, IWo
...
-            {
-                currentStreamingMessage.Contents = updates;
-                updates = [];
-
-                await context.SendMessageAsync(currentStreamingMessage, cancellationToken: cancellationToken).ConfigureAwait(false);
-            }
-
-            currentStreamingMessage = null;
-        }
+        await context.SendMessageAsync(updates.ToAgentRunResponse().Messages, cancellationToken: cancellationToken).ConfigureAwait(false);
```
stephentoub: Note that the new M.E.AI build that should be out tomorrow updates the logic used by this to also factor in AuthorName for message boundary detection, which should help in the case where message ID isn't set but AuthorName is (and AuthorName is always set when using ChatClientAgent).

lokitoth:
This changes the logic of the resulting joined messages - they lose the initial splits as sent out by the agent in individual updates. We were keeping them around deliberately: I recall that fixing a bug that @crickman had with Declarative Workflows at one point (though the details elude me now).

Are we sure this behaviour change will not cause problems?

stephentoub:
>  they lose the initial splits as sent out by the agent in individual updates.

Can you elaborate on this?

lokitoth:
> Can you elaborate on this?

Looks like I had outdated info - it used to be that when the AgentRunResponses were merged using the existing method in the AgentFramework they would merge the multiple updates into a single ChatMessage. But it looks like now that is no longer the logic going on.

dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/AIAgentHostExecutor.cs
```diff
@@ -74,36 +67,9 @@ protected override async ValueTask TakeTurnAsync(List<ChatMessage> messages, IWo
             // In some sense: We should just let it be handled as a ChatMessage, though we should consider
             // providing some mechanisms to help the user complete the request, or route it out of the
             // workflow.
```
stephentoub: Re the TODO, this should be paying attention to any FunctionCallContent (without a paired FunctionResultContent) or UserInputRequestContent (without a paired UserInputResultContent) and integrating with the workflow's HITL mechanism so that the workflow can suspend until the user provides the corresponding result contents and then those should be routed back to this agent.

dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/AIAgentHostExecutor.cs
stephentoub: This has been merged into the AF repo, fwiw.

dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/AIAgentHostExecutor.cs
```diff
@@ -54,17 +54,10 @@ protected override async ValueTask TakeTurnAsync(List<ChatMessage> messages, IWo
         emitEvents ??= this._emitEvents;
         IAsyncEnumerable<AgentRunResponseUpdate> agentStream = this._agent.RunStreamingAsync(messages, this.EnsureThread(context), cancellationToken: cancellationToken);
```
stephentoub: If emitEvents is false, should this just be using RunAsync, or is there a reason it still needs to use RunStreamingAsync even if it doesn't need to output individual events for each update?

stephentoub (approved):

dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/AIAgentHostExecutor.cs
TaoChenOSU: I am working on a redesign of the request-response pattern. Once that's done in Python (#1466), we likely port it to .Net which will support request from the Agent Executor.

dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/AIAgentHostExecutor.cs
TaoChenOSU: Good catch!

alliscode (approved):

lokitoth (approved):

PR #1442, opened 2025-10-13
.NET Workflows - Support agent level function invocation for declarative workflow

crickman:
### Motivation and Context

Enable function calling for declarative workflows (when an agent within the workflow has a function tool).

### Description

- An agent that defines a function will either have that function auto-executed (if available) or else a request is made to the caller via `AgentToolRequest`.  The caller may then respond with `AgentToolResponse`.

- Integration tests add for both auto-execute and caller-requested modes.

- Sample updated

### Contribution Checklist


dotnet/samples/GettingStarted/Workflows/Declarative/ExecuteWorkflow/Program.cs
```diff
@@ -293,6 +319,23 @@ private static InputResponse HandleExternalRequest(ExternalRequest request)
         return new InputResponse(userInput);
     }
 
+    private async ValueTask<AgentToolResponse> HandleToolRequestAsync(AgentToolRequest request)
```
stephentoub, line 371: This shouldn't be necessary, right? If the AIFunctions have been provided to the workflow, it should be able to provide them to the agents and the agents should be able to do the function invocation automatically? Is this only here as a workaround for the lack of https://github.com/Azure/azure-sdk-for-net/pull/53041 ?

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/AzureAgentProvider.cs
```diff
@@ -97,8 +97,31 @@ IEnumerable<MessageInputContentBlock> GetContent()
...
+                    AllowMultipleToolCalls = this.AllowMultipleToolCalls,
+                },
+                clientFactory: null,
+                cancellationToken).ConfigureAwait(false);
+
+        FunctionInvokingChatClient? functionInvokingClient = agent.GetService<FunctionInvokingChatClient>();
+        if (functionInvokingClient is not null)
+        {
+            // Make functions available for execution.  Doesn't change what tool is available for any given agent.
+            functionInvokingClient.AdditionalTools = this.Functions is null ? null : [.. this.Functions.OfType<AITool>()];
```
stephentoub: This assumes that AdditionalTools is already null, which is a fine assumption given that you're in control of the agent creation. If that were to change, however, you'd want to add to AdditionalTools rather than overwriting it.

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Events/AgentToolRequest.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    /// <summary>
+    /// The name of the agent associated with the tool request.
+    /// </summary>
+    public string AgentName { get; }
+
+    /// <summary>
+    /// A list of tool requests.
+    /// </summary>
+    public IReadOnlyList<FunctionCallContent> FunctionCalls { get; }
```
stephentoub: There are other kinds of user input, namely anything deriving from UserInputRequestContent. Do you want to model that as part of this as well?

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Events/AgentToolRequest.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Collections.Immutable;
+using System.Text.Json.Serialization;
+using Microsoft.Extensions.AI;
+
+namespace Microsoft.Agents.AI.Workflows.Declarative.Events;
+
+/// <summary>
+/// Represents a request for user input.
+/// </summary>
+public sealed class AgentToolRequest
```
stephentoub, line 12: Should this be moved out of Declarative and just be the general representation for this in Workflows? It's not clear to me why this is a declarative-specific need.

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Events/AgentToolResponse.cs
```diff
@@ -0,0 +1,54 @@
...
+{
+    /// <summary>
+    /// The name of the agent associated with the tool response.
+    /// </summary>
+    public string AgentName { get; }
+
+    /// <summary>
+    /// A list of tool responses.
+    /// </summary>
+    public IReadOnlyList<FunctionResultContent> FunctionResults { get; }
```
stephentoub: Ditto

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Events/AgentToolResponse.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// <returns>An <see cref="AgentToolResponse"/> that can be provided to the workflow.</returns>
+    /// <exception cref="DeclarativeActionException">Not all <see cref="AgentToolRequest.FunctionCalls"/> have a corresponding <see cref="FunctionResultContent"/>.</exception>
+    public static AgentToolResponse Create(AgentToolRequest toolRequest, params IEnumerable<FunctionResultContent> functionResults)
+    {
+        HashSet<string> callIds = [.. toolRequest.FunctionCalls.Select(call => call.CallId)];
+        HashSet<string> resultIds = [.. functionResults.Select(call => call.CallId)];
+        if (!callIds.SetEquals(resultIds))
+        {
+            throw new DeclarativeActionException($"Missing results for: {string.Join(",", callIds.Except(resultIds))}");
+        }
```
stephentoub, line 50: Do we need to do this check? Or could we instead just rely on the agent exercising its own checks or fallbacks or whatever other logic it may have?

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Kit/PortableValueExtensions.cs
```diff
@@ -0,0 +1,102 @@
...
+    /// <returns>
+    /// A <see cref="object"/> instance representing the underlying value.
+    /// </returns>
+    public static object? Normalize(this PortableValue value) =>
+        Throw.IfNull(value, nameof(value)).TypeId switch
+        {
+            _ when value.IsType(out string? stringValue) => stringValue,
+            _ when value.IsSystemType(out bool? boolValue) => boolValue.Value,
+            _ when value.IsSystemType(out int? intValue) => intValue.Value,
+            _ when value.IsSystemType(out long? longValue) => longValue.Value,
```
stephentoub, line 68: There are plenty of other primitive types, e.g. uint, short, byte, etc... do we not need those? What does PortableValue store?

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Kit/PortableValueExtensions.cs
```diff
@@ -0,0 +1,102 @@
...
+    {
+        return GetValues().ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
+
+        IEnumerable<KeyValuePair<string, object?>> GetValues()
+        {
+            foreach (string key in source.Keys)
+            {
+                object? value = source[key];
+                yield return new KeyValuePair<string, object?>(key, value.NormalizePortableValue());
+            }
```
stephentoub:
```suggestion
            foreach (DictionaryEntry entry in source)
            {
                yield return new KeyValuePair<string, object?>((string)entry.Key, entry.Value.NormalizePortableValue());
            }
```

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Kit/PortableValueExtensions.cs
```diff
@@ -0,0 +1,102 @@
...
+        return GetValues().ToArray();
+
+        IEnumerable<object?> GetValues()
+        {
+            IEnumerator enumerator = source.GetEnumerator();
+            while (enumerator.MoveNext())
+            {
+                yield return enumerator.Current.NormalizePortableValue();
+            }
+        }
```
stephentoub:
```suggestion
        return source.Cast<object?>().Select(NormalizePortableValue).ToArray();
```
?

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/ObjectModel/InvokeAzureAgentExecutor.cs
```diff
@@ -16,23 +19,67 @@ namespace Microsoft.Agents.AI.Workflows.Declarative.ObjectModel;
...
-        AgentRunResponse agentResponse = await agentProvider.InvokeAgentAsync(this.Id, context, agentName, conversationId, autoSend, additionalInstructions, inputMessages, cancellationToken).ConfigureAwait(false);
+        bool isComplete = true;
 
-        await this.AssignAsync(this.AgentOutput?.Messages?.Path, agentResponse.Messages.ToTable(), context).ConfigureAwait(false);
+        AgentRunResponse agentResponse = await agentProvider.InvokeAgentAsync(this.Id, context, agentName, conversationId, autoSend, additionalInstructions, messages, cancellationToken).ConfigureAwait(false);
 
-        return default;
+        if (string.IsNullOrEmpty(agentResponse.Text))
+        {
+            HashSet<FunctionCallContent> toolCalls = [.. agentResponse.Messages.SelectMany(m => m.Contents.OfType<FunctionCallContent>())];
```
stephentoub: What if some or all of these also have paired FunctionResultContents?

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Events/AgentToolRequest.cs
crickman: As we discussed, I do...as a separate PR.

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Events/AgentToolRequest.cs
crickman: Perhaps, we can start here and then promote (as we had discussed).

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Events/AgentToolResponse.cs
crickman: Ditto

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Events/AgentToolResponse.cs
crickman: This doesn't eliminate our reliance on the agent to exercise its own checks.  It just provides an fast fail for this one case with an actionable message.  (For this case, I felt a very clear message could add value...in the hosted scenario clear failure messages are critical.)

dotnet/samples/GettingStarted/Workflows/Declarative/ExecuteWorkflow/Program.cs
crickman: Yes, but there's the case where someone the functions aren't present.  I wanted to enable local debugging for the hosted workflow scenario without adding separate sample.  This is the last thing I touched...**I'll look for opportunities to clarify the patterns here:**

1. No functions (no inboke)
2. Agent has function-tool defined, but `AITool` doesn't exist within workflow (manual invoke)
3. Agent has function-tool defined and its been included in workflow (auto invoke)

You are correct that `HandleToolRequestAsync` isn't relevant for #1 & #3, but it is for #2.

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Events/AgentToolResponse.cs
stephentoub: > It just provides an fast fail for this one case with an actionable message. (For this case, I felt a very clear message could add value...in the hosted scenario clear failure messages are critical.)

This assumes it is definitively an error. That may be a valid assumption right now, but you could imagine it not being one, e.g. if there's no result for a function call request, the service could decide to try something else.

dotnet/samples/GettingStarted/Workflows/Declarative/ExecuteWorkflow/Program.cs
stephentoub: > but it is for https://github.com/microsoft/agent-framework/pull/2.

Yes, except that (2) isn't shown here, because either (a) the tools were provided and thus they could be auto-invoked by the agent, or (b) the tools weren't provided in which case this method won't be able to handle them either.

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Kit/PortableValueExtensions.cs
crickman: Updated

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Kit/PortableValueExtensions.cs
crickman: Updated (lost curly brackets for oneliners)

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Events/AgentToolResponse.cs
crickman: I'm actively working with integrating with the agent/workflow services.  I don't believe there's anything else that will be tried.  I'll follow-up as an integration activity.  For 99.99% of the non-hosted cased, I'd expect the auto-execute case to be in force.

Feels like an ok starting point (conservative) as we gather data-points.  Also, they've specifically requested clear error messages.

This will be easy to delete if it gets in the way / isn't needed.  (I've made a note for follow-up/monitoring)

dotnet/samples/GettingStarted/Workflows/Declarative/ExecuteWorkflow/Program.cs
crickman: Correct, which is why I'm going to take the opportunity to do another pass and clarify as part of this PR.

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/AzureAgentProvider.cs
crickman: Good point, updated.

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Kit/PortableValueExtensions.cs
crickman: Good question, I had considered the same.

Generally `PortableValue` can be anything, for this usage/domain (function calling) we're constrained to JSON types:  For JSON Schema / Open API the set of support types of fairly constrained.  This may even be a super set of what's allowed: https://json-schema.org/understanding-json-schema/reference/type.

For the converstion logic related to `FormulaValue` or `DataValue` types (other similar conversation code), the set of supported types is constrained to a known domain.  `byte`, `uint`, etc...aren't present is that set.

I'm working with Jacob to refine these patterns around `PortableValue` and we may likely end up promoting to `Workflows`.  As such, I expect some iteration as we move forward down the path with discovering what i means to handle `PortableValue`.

I've made a note to follow-up on this.  In the meantime, its critical that we unblock agent-services integration for this scenario.

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/ObjectModel/InvokeAzureAgentExecutor.cs
crickman: Good point.  Also the `HashSet` isn't really used here (other than being `IEnumerable<>`

crickman: Updated - O(2N)

dotnet/samples/GettingStarted/Workflows/Declarative/ExecuteWorkflow/Program.cs
crickman: Updated pattern for additional clarity on all three cases and verified execution for each case.  

Once the agent services standup the workflow API, a separate sample for manual tool calling could make sense (either here or as part of Azure Agent SDK).  In the meantime, this is the blueprint on how to do manual function calling (for agent services and allows us to debug all three cases).

alliscode (approved):

dotnet/samples/GettingStarted/Workflows/Declarative/ExecuteWorkflow/Program.cs
```diff
@@ -142,12 +172,21 @@ private Program(string workflowFile, string? workflowInput)
 
         this.FoundryEndpoint = this.Configuration[ConfigKeyFoundryEndpoint] ?? throw new InvalidOperationException($"Undefined configuration setting: {ConfigKeyFoundryEndpoint}");
         this.FoundryClient = new PersistentAgentsClient(this.FoundryEndpoint, new AzureCliCredential());
+
+        List<AIFunction> functions =
+            [
+                // Manually define any custom functions that may be required by agents within the workflow.
+                // By default, this sample does not include any functions.
+                //AIFunctionFactory.Create(),
```
TaoChenOSU, line 180: nit: remove?

TaoChenOSU (approved):

dotnet/samples/GettingStarted/Workflows/Declarative/ExecuteWorkflow/Program.cs
crickman: This is included as direction for a user that wan't to use function tools with a declarative workflow.  (requires customization)

PR #1464, opened 2025-10-14
.NET: [BREAKING] Enable sharing of workflow instances across concurrently executing runs

lokitoth:
### Motivation and Context

Right now, a workflow can only be executed in one run a time. This stems from the capacity of the workflow to be built using direct instances of Executors, which will cause multiple runs of the workflow to potentially be sharing live objects.

To make hosting scenarios simpler, it is necessary to provide a mode where a run does not take ownership of a workflow during execution. To do this we need to enforce delay-instantiating executors using factory methods, or for the user to declare that the Executor is cross-run sharing aware, and introduce affordances to detect this case.

### Description

Completes the Execution Mode refactorings, adding "Concurrent" mode, which enables multiple simultaneous runs against the same workflow instance, and updates in-package Eexecutors to either be cross-run sharing aware, or always be attached to a workflow via a factory.

Also breaks up the AgentWorkflowBuilder.cs file into multiple class-specific files, promotes publics to top-level, and moves internals to Specialized.

Breaking Changes:
* Moves public child classes under AgentWorkflowBuilder to top level.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Interpreter/DeclarativeWorkflowContext.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Interpreter/DeclarativeWorkflowContext.cs
@@ -3,6 +3,7 @@
 using System;
 using System.Collections.Frozen;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
@@ -76,23 +77,56 @@ internal sealed class DeclarativeWorkflowContext : IWorkflowContext
         this.State.Bind();
     }
 
+    private bool IsManagedScope(string? scopeName) => scopeName is not null && VariableScopeNames.IsValidName(scopeName);
+
     /// <inheritdoc/>
     public async ValueTask<TValue?> ReadStateAsync<TValue>(string key, string? scopeName = null, CancellationToken cancellationToken = default)
     {
-        bool isManagedScope =
-            scopeName is not null && // null scope cannot be managed
-            VariableScopeNames.IsValidName(scopeName);
-
         return typeof(TValue) switch
         {
             // Not a managed scope, just pass through.  This is valid when a declarative
             // workflow has been ejected to code (where DeclarativeWorkflowContext is also utilized).
-            _ when !isManagedScope => await this.Source.ReadStateAsync<TValue>(key, scopeName, cancellationToken).ConfigureAwait(false),
+            _ when !this.IsManagedScope(scopeName) => await this.Source.ReadStateAsync<TValue>(key, scopeName, cancellationToken).ConfigureAwait(false),
             // Retrieve formula values directly from the managed state to avoid conversion.
             _ when typeof(TValue) == typeof(FormulaValue) => (TValue?)(object?)this.State.Get(key, scopeName),
             // Retrieve native types from the source context to avoid conversion.
             _ => await this.Source.ReadStateAsync<TValue>(key, scopeName, cancellationToken).ConfigureAwait(false),
         };
+    }
+
+    public async ValueTask<TValue> ReadOrInitStateAsync<TValue>(string key, Func<TValue> initialStateFactory, string? scopeName = null, CancellationToken cancellationToken = default)
+    {
+        return typeof(TValue) switch
+        {
+            // Not a managed scope, just pass through.  This is valid when a declarative
+            // workflow has been ejected to code (where DeclarativeWorkflowContext is also utilized).
+            _ when !this.IsManagedScope(scopeName) => await this.Source.ReadOrInitStateAsync(key, initialStateFactory, scopeName, cancellationToken).ConfigureAwait(false),
+            // Retrieve formula values directly from the managed state to avoid conversion.
+            _ when typeof(TValue) == typeof(FormulaValue) => await EnsureFormulaValueAsync().ConfigureAwait(false),
+            // Retrieve native types from the source context to avoid conversion.
+            _ => await this.Source.ReadOrInitStateAsync(key, initialStateFactory, scopeName, cancellationToken).ConfigureAwait(false),
+        };
+
+        async ValueTask<TValue> EnsureFormulaValueAsync()
+        {
+            Debug.Assert(typeof(TValue) == typeof(FormulaValue), "It is a bug to call this method with TValue not === FormulaValue");
+            FormulaValue? result = this.State.Get(key, scopeName);
+
+            if (result is null or BlankValue)
+            {
+                result = initialStateFactory() as FormulaValue;
+                if (result is null)
+                {
+                    throw new InvalidOperationException($"The initial state factory for key '{key}' in scope '{scopeName}' did not return a FormulaValue.");
+                }
+
+                this.State.Set(key, result, scopeName);
+                await this.Source.QueueStateUpdateAsync(key, result.AsPortable(), scopeName, cancellationToken)
+                                 .ConfigureAwait(false);
+            }
+
+            return (TValue)(object)result!; // The null analyzer is confused here, but it is impossible to hit this line with result is null
+        }
     }
 
     /// <inheritdoc/>
@@ -173,10 +207,5 @@ internal sealed class DeclarativeWorkflowContext : IWorkflowContext
 
             return this.Source.QueueStateUpdateAsync(key, formulaValue.AsPortable(), scopeName, cancellationToken);
         }
-    }
-
-    public ValueTask<T> ReadOrInitStateAsync<T>(string key, Func<T> initialStateFactory, string? scopeName = null, CancellationToken cancellationToken = default)
-    {
-        return this.Source.ReadOrInitStateAsync(key, initialStateFactory, scopeName, cancellationToken);
     }
 }


dotnet/src/Microsoft.Agents.AI.Workflows/Workflow.cs
```diff
@@ -180,19 +202,18 @@ internal void TakeOwnership(object ownerToken, bool subworkflow = false, object?
...
     {
-        if (this._ownerToken == null)
+        object? originalToken = Interlocked.CompareExchange(ref this._ownerToken, null, ownerToken);
+        if (originalToken == null)
         {
             throw new InvalidOperationException("Attempting to release ownership of a Workflow that is not owned.");
         }
 
-        if (!ReferenceEquals(this._ownerToken, this._ownerToken))
+        if (!ReferenceEquals(originalToken, ownerToken))
```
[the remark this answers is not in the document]
lokitoth: This is not correct, because the value in `this._originalToken` has been replaced with the incoming `null`.

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Interpreter/DeclarativeWorkflowContext.cs
```diff
@@ -170,4 +174,9 @@ ValueTask QueueNativeStateAsync(object rawValue)
             return this.Source.QueueStateUpdateAsync(key, formulaValue.AsPortable(), scopeName, cancellationToken);
         }
     }
+
+    public ValueTask<T> ReadOrInitStateAsync<T>(string key, Func<T> initialStateFactory, string? scopeName = null, CancellationToken cancellationToken = default)
+    {
+        return this.Source.ReadOrInitStateAsync(key, initialStateFactory, scopeName, cancellationToken);
```
crickman: I suspect this requires logic similiar to `ReadStateAsync`

dotnet/src/Microsoft.Agents.AI.Workflows/StatefulExecutor.cs
```diff
@@ -0,0 +1,186 @@
...
+using Microsoft.Agents.AI.Workflows.Reflection;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Agents.AI.Workflows;
+
+/// <summary>
+/// Provides a base class for executors that maintain and manage state across multiple message handling operations.
+/// </summary>
+/// <typeparam name="TState">The type of state associated with this Executor.</typeparam>
+public abstract class StatefulExecutor<TState> : Executor
```
alliscode, line 15: Is this to be the recommended approach when using state rather than using the context directly? From within a handler in a `StatefulExecutor` will you see both options as in:

 ```
protected override async ValueTask HandleAsync(Message msg, IWorkflowContext context, ...)
  {
      // StatefulExecutor
      await this.InvokeWithStateAsync(
          async (state, ctx, ct) => { /* ... */ return state; },
          context,
          cancellationToken);

      // Context
      await context.InvokeWithStateAsync(
          async (state, ctx, ct) => { /* ... */ return state; },
          "SomeKey",
          () => new MyState(),
          cancellationToken: cancellationToken);
  }
```

File: dotnet/src/Microsoft.Agents.AI.Workflows/InProc/InProcessRunner.cs [hunk 4]
```diff
@@ -21,21 +21,47 @@ namespace Microsoft.Agents.AI.Workflows.InProc;
 /// scenarios where workflow execution does not require executor distribution. </para></remarks>
 internal sealed class InProcessRunner : ISuperStepRunner, ICheckpointingHandle
 {
-    public InProcessRunner(Workflow workflow, ICheckpointManager? checkpointManager, string? runId = null, object? workflowOwnership = null, bool subworkflow = false, IEnumerable<Type>? knownValidInputTypes = null)
+    //
```
TaoChenOSU: Remove

dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/WorkflowHostExecutor.cs
```diff
@@ -224,6 +229,12 @@ private async ValueTask ForwardWorkflowEventAsync(object? sender, WorkflowEvent
     internal async ValueTask AttachSuperStepContextAsync(ISuperStepJoinContext joinContext)
     {
         this._joinContext = Throw.IfNull(joinContext);
+
+        // TODO: Make subworkflows work under concurrent execution
+        //if (joinContext.ConcurrentRunsEnabled)
+        //{
+        //    throw new InvalidOperationException("Not yet supported.");
+        //}
```
alliscode: Is this comment still valid?

dotnet/src/Microsoft.Agents.AI.Workflows/StatefulExecutor.cs
lokitoth: (tldr: in this case, they should be made the same, in principle there may be reasons to use both, e.g. in a static context, or when accessing shared state, with `key != this.StateKey`; the only use of `context.` should be to implement `InvokeWithStateAsync`)

It depends: In a world where the `StatefulExecutor`'s state is private, but it needs to access shared state, supporting this on context is still valuable. In a real sense it is just shorthand for invoking some set of actions inside of a try finally and reads/updates the state (as appropriate).

dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/WorkflowHostExecutor.cs
lokitoth: No, this is to be removed now that support is properly implemented (by ensuring that Workflows are always hosted in a per-Run-constructed HostExecutor). I started down the road of making a truly free-contexted runner and just pass the root workflow context to it when running, but that forces us to expose the internal executors to the external workflow and bookkeeping them becomes challenging (not to mention breaks hard in sub-sub-workflows- which is something we should add to the tests)

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Interpreter/DeclarativeWorkflowContext.cs
lokitoth: From an implementation standpoint, both `ReadStateAsync` and `ReadOrInitStateAsync` use the same modified variant of the original `ReadStateAsync` under the hood. Consuming it will require similar logic to `ReadStateAsync`, and the dynamics of objects and types are consistent between them.

crickman: Right, but the declarative layer has additional logic that I'm calling out (in `ReadStateAsync`).  Ping me and we can work through this together.  I think we just need to rationalize how the factory fits in.

crickman (changes requested):
Let's take a deeper look at `DeclarativeWorkflowContext` together.

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Interpreter/DeclarativeWorkflowContext.cs
lokitoth: Okay, makes sense; I'm going to fix the other issues and rebase, then ping you

dotnet/src/Microsoft.Agents.AI.Workflows/AIAgentExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        string id = string.IsNullOrEmpty(agent.Name) ? agent.Id : $"{agent.Name}_{agent.Id}";
+        return InvalidNameCharsRegex().Replace(id, "_");
+    }
+
+    /// <summary>
+    /// Regex that flags any character other than ASCII digits or letters or the underscore.
+    /// </summary>
+#if NET
+    [GeneratedRegex("[^0-9A-Za-z_]+")]
```
stephentoub: Pre-existing, but we might want to remove the `_` from the pattern... that will then make it so that runs of multiple invalid characters including `_` will be consolidated into a single `_`.

dotnet/src/Microsoft.Agents.AI.Workflows/ChatProtocolExecutor.cs
```diff
@@ -13,62 +12,70 @@ internal class ChatProtocolExecutorOptions
...
+            routeBuilder = routeBuilder.AddHandler<string>(
+                (message, context) => this.AddMessageAsync(new(this._stringMessageChatRole.Value, message), context));
         }
 
-        // Routing requires exact type matches. The runtime may dispatch either List<ChatMessage> or ChatMessage[].
-        return routeBuilder.AddHandler<ChatMessage>((message, _, __) => this._pendingMessages.Add(message))
-                           .AddHandler<List<ChatMessage>>((messages, _, __) => this._pendingMessages.AddRange(messages))
-                           .AddHandler<ChatMessage[]>((messages, _, __) => this._pendingMessages.AddRange(messages))
+        return routeBuilder.AddHandler<ChatMessage>(this.AddMessageAsync)
+                           .AddHandler<IEnumerable<ChatMessage>>(this.AddMessagesAsync)
```
stephentoub, line 37: This is still a nop, right? Or does the polymorphism work now?

dotnet/src/Microsoft.Agents.AI.Workflows/ChatProtocolExecutor.cs
```diff
@@ -13,62 +12,70 @@ internal class ChatProtocolExecutorOptions
...
+        async ValueTask<List<ChatMessage>?> InvokeTakeTurnAsync(List<ChatMessage>? maybePendingMessages, IWorkflowContext context, CancellationToken cancellationToken)
         {
-            List<ChatMessage> messages = messagesValue.Value.DeserializeMessages();
-            this._pendingMessages.AddRange(messages);
+            await this.TakeTurnAsync(maybePendingMessages ?? [], context, token.EmitEvents, cancellationToken)
+                      .ConfigureAwait(false);
+
+            await context.SendMessageAsync(token, cancellationToken: cancellationToken).ConfigureAwait(false);
+
+            return new([]);
```
stephentoub: Is this supposed to be `return [];`? Or even `return null;`?

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Interpreter/DeclarativeWorkflowContext.cs
lokitoth: Synced offline - updated for correct Read (and init) logic. Thanks for the help @crickman

dotnet/src/Microsoft.Agents.AI.Workflows/ChatProtocolExecutor.cs
lokitoth: Not yet, but most of the pieces are in place to get polymorphism working right. Just had higher priority issues to deal with so far

dotnet/src/Microsoft.Agents.AI.Workflows/ChatProtocolExecutor.cs
lokitoth: Yes. Vestigial from before this method went async. Good catch

TaoChenOSU (approved):

PR #1484, opened 2025-10-15
.NET: ChatClientBuilderExtensions should be in the M.E.AI namespace

markwallace-microsoft:
### Motivation and Context

`ChatClientBuilderExtensions` extends `ChatClientBuilder` and should use the same namespace.

**We should avoid using `ChatClient` in namespaces because they will conflict with the `OpenAI.Chat.ChatClient` class.**

### Contribution Checklist


rogerbarreto (commented):
Suggest using `CreateAIAgent` instead of `BuildAIAgent` for consistency.

We currently also have the `AIAgentBuilder`, in theory `BuildAIAgent` should be returning the `AIAgentBuilder` to make justice to the name.

markwallace-microsoft:
> Suggest using `CreateAIAgent` instead of `BuildAIAgent` for consistency.
> 
> We currently also have the `AIAgentBuilder`, in theory `BuildAIAgent` should be returning the `AIAgentBuilder` to make justice to the name.

The scope of this PR is just the namespace change.

westey-m (approved):

PR #1498, opened 2025-10-15
.NET: Update WorkflowHostAgent post Checkpointing implementation

### LINKED ISSUE #797: .NET: AIAgentHostExecutor maintains thread across all operations
If I include an AIAgent in a workflow, each subsequent use of that workflow ends up seeing the state from previous runs of the workflow, because the executor is maintaining a thread for all uses. This does not seem desirable? Presumably each use should end up throwing away the thread and creating a new one?  Unless workflows aren't meant to be reused at all?

lokitoth:
### Motivation and Context

WorkflowHostAgent was initially implemented before Checkpointing was available. This meant that in order to support resuming, the WorkflowHostAgent needed to keep the runs around, which broke it when stricter rules about concurrent sharing of workflows during execution were introduced.

### Description

This change updates the hosting logic to release the underlying StreamingRun when the RunStreamingAsync or RunAsync are invoked, in favour of keeping the checkpointing information in the WorkflowThread to enable resumption.

Also fixes #797

### Contribution Checklist

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.Workflows/CheckpointManager.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Workflows/CheckpointManager.cs
@@ -20,7 +20,7 @@ public sealed class CheckpointManager : ICheckpointManager
         return new CheckpointManagerImpl<TStoreObject>(marshaller, store);
     }
 
-    private CheckpointManager(ICheckpointManager impl)
+    internal CheckpointManager(ICheckpointManager impl)
     {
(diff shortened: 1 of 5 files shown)


crickman:
Is anything here testable?  (strike while the iron is hot, they say)

lokitoth:
It gets smoke tested via the existing tests but will definitely need more (like everything else in Workflows). With that said, this is a fix for a blocker, so going to avoid the special-case tests for now.

PR #1501, opened 2025-10-15
.NET: Add support for background responses

SergeyMenshykh:
This PR updates AF model to use model for background responses (long-running operations) from MEAI and updates `ChatClientAgent` to support it.

Background responses use a continuation token returned by the `RunAsync` and `RunStreamingAsync` methods in the `AgentRunResponse` and `AgentRunResponseUpdate` classes, respectively, when background responses are enabled and supported by the AI agent.

The continuation token contains all necessary details to enable polling for a background response using the non-streaming `RunAsync` method. It also allows resuming a streamed background response with the `RunStreamingAsync` method if the stream is interrupted. In both cases, the continuation token obtained from the initial agent response or the streamed update received before the interruption should be supplied as input for follow-up calls.

When a background response has completed, failed, or cannot proceed further (for example, when user input is required), the continuation token returned by either method will be null, signaling to consumers that processing is complete and there is nothing to poll or resume. The result returned by either method can then be used for further processing.

#### Non-streaming API:
````csharp
AgentRunOptions options = new()
{
    AllowBackgroundResponses = true
};

// Get initial response with continuation token
AgentRunResponse response = await agent.RunAsync("What's the biggest animal?", options: options);

// Continue to poll until the final response is received
while (response.ContinuationToken is not null)
{
    options.ContinuationToken = response.ContinuationToken;
    response = await agent.RunAsync([], options: options);
}

Console.WriteLine(response.Text);
````

#### Streaming API:
````csharp
AgentRunOptions options = new()
{
    AllowBackgroundResponses = true
};

AgentRunResponseUpdate? lastReceivedUpdate = null;

await foreach (var update in agent.RunStreamingAsync("What is the capital of France?", options: options))
{
    Console.Write(update.Text);
  
    lastReceivedUpdate = update;
  
   // Simulate an interruption
    break;
}

// Resume from interruption point captured by the continuation token
options.ContinuationToken = lastReceivedUpdate?.ContinuationToken;
await foreach (var update in agent.RunStreamingAsync([], options: options))
{
    Console.Write(update.Text);
}
````

### DIFF
--- a/dotnet/samples/GettingStarted/Agents/Agent_Step17_BackgroundResponses/Program.cs
+++ b/dotnet/samples/GettingStarted/Agents/Agent_Step17_BackgroundResponses/Program.cs
@@ -1,6 +1,6 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 
-// This sample shows how use background responses with ChatClientAgent and OpenAI Responses.
+// This sample shows how to use background responses with ChatClientAgent and OpenAI Responses.
 
 using Azure.AI.OpenAI;
 using Azure.Identity;

### CODE REVIEW

File: dotnet/src/Microsoft.Agents.AI.Abstractions/AgentRunOptions.cs [hunk 1]
```diff
@@ -33,5 +33,48 @@ public AgentRunOptions()
...
+    /// This property is used for background responses that can be activated via the <see cref="AllowBackgroundResponses"/>
+    /// property if the <see cref="AIAgent"/> implementation supports them.
+    /// Streamed background responses, such as those returned by default by "AIAgent.RunStreamingAsync",
+    /// can be resumed if interrupted. This means that a continuation token obtained from the <see cref="AgentRunResponse.ContinuationToken"/>
+    /// of an update just before the interruption occurred can be passed to this property to resume the stream from the point of interruption.
+    /// Non-streamed background responses, such as those returned by "AIAgent.RunAsync",
+    /// can be polled for completion by obtaining the token from the <see cref="AgentRunResponse.ContinuationToken"/> property
+    /// and passing it to this property on subsequent calls to "AIAgent.RunAsync".
+    /// </remarks>
+    public object? ContinuationToken { get; set; }
```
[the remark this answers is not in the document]
SergeyMenshykh: by design

dotnet/src/Microsoft.Agents.AI.Abstractions/AgentRunOptions.cs
```diff
@@ -33,5 +33,48 @@ public AgentRunOptions()
...
+    /// response updates until the operation is completed. If the streaming connection is interrupted, the
+    /// continuation token obtained from the last update that has one should be supplied to a subsequent call to the same streaming API
+    /// to resume the stream from the point of interruption and continue receiving updates until the operation is completed.
+    /// </para>
+    /// <para>
+    /// This property only takes effect if the implementation it's used with supports background responses.
+    /// If the implementation does not support background responses, this property will be ignored.
+    /// </para>
+    /// </remarks>
+    public bool? AllowBackgroundResponses { get; set; }
```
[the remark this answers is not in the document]
SergeyMenshykh: By design atm. A dedicated `ResponseContinuationToken` type will be used later.

dotnet/src/Microsoft.Agents.AI.Abstractions/AgentRunResponse.cs
```diff
@@ -159,6 +160,23 @@ public IList<ChatMessage> Messages
...
+    /// and the result of the response has not been obtained yet. If the response has completed and the result has been obtained,
+    /// the token will be <see langword="null"/>.
+    /// <para>
+    /// This property should be used in conjunction with <see cref="AgentRunOptions.ContinuationToken"/> to
+    /// continue to poll for the completion of the response. Pass this token to
+    /// <see cref="AgentRunOptions.ContinuationToken"/> on subsequent calls to "AIAgent.RunAsync"
+    /// to poll for completion.
+    /// </para>
+    /// </remarks>
+    public object? ContinuationToken { get; set; }
```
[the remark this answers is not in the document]
SergeyMenshykh: By design at the moment, a dedicated `ResponseContinuationToken` type will be used later.

dotnet/src/Microsoft.Agents.AI.Abstractions/AgentRunResponseUpdate.cs
```diff
@@ -148,6 +149,21 @@ public IList<AIContent> Contents
...
+    /// <see cref="AIAgent"/> implementations that support background responses will return
+    /// a continuation token on each update if background responses are allowed in <see cref="AgentRunOptions.AllowBackgroundResponses"/>
+    /// except of the last update, for which the token will be <see langword="null"/>.
+    /// <para>
+    /// This property should be used for stream resumption, where the continuation token of the latest received update should be
+    /// passed to <see cref="AgentRunOptions.ContinuationToken"/> on subsequent calls to "AIAgent.RunAsync"
+    /// to resume streaming from the point of interruption.
+    /// </para>
+    /// </remarks>
+    public object? ContinuationToken { get; set; }
```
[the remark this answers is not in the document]
SergeyMenshykh: By design at the moment, a dedicated `ResponseContinuationToken` type will be used later.

dotnet/src/Microsoft.Agents.AI.Abstractions/AgentRunOptions.cs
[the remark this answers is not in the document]
stephentoub: > by design

Is there a reason they can't be ResponseContinuationToken at the AgentRunXx level? They're not in M.E.AI because they're experimental in a stable library... but this library isn't stable. Is the concern still that it triggers errors by default if someone has their own JSON source generation context?

dotnet/src/Microsoft.Agents.AI.Abstractions/AgentRunResponseUpdate.cs
```diff
@@ -148,6 +149,21 @@ public IList<AIContent> Contents
...
+    /// <summary>
+    /// Gets or sets the continuation token for resuming the streamed agent response of which this update is a part.
+    /// </summary>
+    /// <remarks>
+    /// <see cref="AIAgent"/> implementations that support background responses will return
+    /// a continuation token on each update if background responses are allowed in <see cref="AgentRunOptions.AllowBackgroundResponses"/>
+    /// except of the last update, for which the token will be <see langword="null"/>.
+    /// <para>
+    /// This property should be used for stream resumption, where the continuation token of the latest received update should be
+    /// passed to <see cref="AgentRunOptions.ContinuationToken"/> on subsequent calls to "AIAgent.RunAsync"
```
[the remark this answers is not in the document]
SergeyMenshykh: The `RunAsync` method has a few overloads, and it is impossible to specify `<see cref="AIAgent.RunAsync"/>` without the compilation error: _"Ambiguous reference in cref attribute: 'AIAgent.RunAsync'. Assuming 'AIAgent.RunAsync(AgentThread?, AgentRunOptions?, CancellationToken)', but it could have also matched other overloads including 'AIAgent.RunAsync(string, AgentThread?, AgentRunOptions?, CancellationToken)'."_

dotnet/src/Microsoft.Agents.AI.Abstractions/AgentRunResponse.cs
```diff
@@ -159,6 +160,23 @@ public IList<ChatMessage> Messages
...
+    /// </summary>
+    /// <remarks>
+    /// <see cref="AIAgent"/> implementations that support background responses will return
+    /// a continuation token if background responses are allowed in <see cref="AgentRunOptions.AllowBackgroundResponses"/>
+    /// and the result of the response has not been obtained yet. If the response has completed and the result has been obtained,
+    /// the token will be <see langword="null"/>.
+    /// <para>
+    /// This property should be used in conjunction with <see cref="AgentRunOptions.ContinuationToken"/> to
+    /// continue to poll for the completion of the response. Pass this token to
+    /// <see cref="AgentRunOptions.ContinuationToken"/> on subsequent calls to "AIAgent.RunAsync"
```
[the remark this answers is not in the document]
SergeyMenshykh: The RunAsync method has a few overloads, and it is impossible to specify <see cref="AIAgent.RunAsync"/> without the compilation error: "Ambiguous reference in cref attribute: 'AIAgent.RunAsync'. Assuming 'AIAgent.RunAsync(AgentThread?, AgentRunOptions?, CancellationToken)', but it could have also matched other overloads including 'AIAgent.RunAsync(string, AgentThread?, AgentRunOptions?, CancellationToken)'."

dotnet/src/Microsoft.Agents.AI.Abstractions/AgentRunResponseUpdate.cs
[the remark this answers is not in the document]
stephentoub: You could try putting an "m:" before it.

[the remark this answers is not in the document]
SergeyMenshykh: Like this - `<see cref="AIAgent.m:RunAsync"/>`? It does not work either.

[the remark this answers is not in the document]
stephentoub: No, like `<see cref="m:AIAgent.RunAsync"/>`

dotnet/src/Microsoft.Agents.AI.Abstractions/AgentRunOptions.cs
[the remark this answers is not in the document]
SergeyMenshykh: > Is the concern still that it triggers errors by default if someone has their own JSON source generation context?

Yes, that's the main reason I used object as the type for all those props. Later, when we update MEAI and AF to use a dedicated ResponseContinuationToken type instead of object, AF consumers, at least the ones that don't pin STJ nuget, will get the latest AF with no warnings.

dotnet/src/Microsoft.Agents.AI.Abstractions/AgentRunResponseUpdate.cs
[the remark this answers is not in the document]
SergeyMenshykh: Thanks, it worked, but not as I would like it to. It generated the "CA1200 // Avoid using cref tags with a prefix" warning, which I suppressed. As a result, I got this tooltip for the ContinuationToken property:
imageIt is not rendered as a clickable link and looks a little bit better than "AIAgent.RunAsync". I'll add the "m:" prefix to all the new properties and suppress the warnings.

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
```diff
@@ -532,7 +532,19 @@ await thread.AIContextProvider.InvokedAsync(new(inputMessages) { InvokeException
...
+        return ApplyBackgroundResponsesProperties(requestChatOptions, runOptions);
+
+        static ChatOptions? ApplyBackgroundResponsesProperties(ChatOptions? chatOptions, AgentRunOptions? agentRunOptions)
+        {
+            if (agentRunOptions?.AllowBackgroundResponses is not null)
+            {
+                chatOptions ??= new ChatOptions();
+                chatOptions.AllowBackgroundResponses = agentRunOptions.AllowBackgroundResponses;
+                chatOptions.ContinuationToken = agentRunOptions.ContinuationToken;
+            }
```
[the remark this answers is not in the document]
SergeyMenshykh: Fixed

dotnet/src/Microsoft.Agents.AI.Abstractions/AgentRunResponseUpdate.cs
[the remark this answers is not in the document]
SergeyMenshykh: Fixed by using full signature of the Run{Streaming}Async methods that don't require messages.

dotnet/samples/GettingStarted/Agents/Agent_Step17_BackgroundResponses/Program.cs
```diff
@@ -0,0 +1,70 @@
...
+while (response.ContinuationToken is { } token)
+{
+    // Wait before polling again.
+    await Task.Delay(TimeSpan.FromSeconds(2));
+
+    // Continue with the token.
+    options.ContinuationToken = token;
+
+    response = await agent.RunAsync(thread, options);
+}
```
rogerbarreto, line 37: I wonder why do we need to handle two tokens?

Shouldn't the continuation token work as the `cancellationToken` as well?

```suggestion
while (response.ContinuationToken is { } token)
{
    // Wait before polling again.
    await Task.Delay(TimeSpan.FromSeconds(2));

    response = await agent.RunAsync(thread, options, token);
}
```

dotnet/src/Microsoft.Agents.AI.Abstractions/AgentRunOptions.cs
```diff
@@ -33,5 +33,50 @@ public AgentRunOptions()
...
```
rogerbarreto, line 76: Suggest making this virtual, so we can override the `ChatClientAgentRunOptions` to use the `ChatOptions.AllowBackgroundResponses` directly and don't have to manage two memory references.

```suggestion
public virtual bool? AllowBackgroundResponses { get; set; }
```

dotnet/src/Microsoft.Agents.AI.Abstractions/AgentRunResponseUpdate.cs
```diff
@@ -148,6 +149,23 @@ public IList<AIContent> Contents
     /// <summary>Gets or sets a timestamp for the response update.</summary>
     public DateTimeOffset? CreatedAt { get; set; }
 
+#pragma warning disable CA1200 // Avoid using cref tags with a prefix
```
rogerbarreto: Do you need this pragma? Suggest fixing the cref
```suggestion
```

dotnet/samples/GettingStarted/Agents/Agent_Step17_BackgroundResponses/Program.cs
SergeyMenshykh: > I wonder why do we need to handle two tokens?

Because they have different responsibilities: one allows for the continuation of a long-running operation, and the other one allows for cancellation.

> I wonder why do we need to handle two tokens?

How would a continuation token work as a cancellation token?

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
```diff
@@ -532,7 +532,20 @@ await thread.AIContextProvider.InvokedAsync(new(inputMessages) { InvokeException
...
+        return ApplyBackgroundResponsesProperties(requestChatOptions, runOptions);
+
+        static ChatOptions? ApplyBackgroundResponsesProperties(ChatOptions? chatOptions, AgentRunOptions? agentRunOptions)
+        {
+            // If any of the background response properties are set in the run options, we should apply both to the chat options.
+            if (agentRunOptions?.AllowBackgroundResponses is not null || agentRunOptions?.ContinuationToken is not null)
+            {
+                chatOptions ??= new ChatOptions();
+                chatOptions.AllowBackgroundResponses = agentRunOptions.AllowBackgroundResponses;
+                chatOptions.ContinuationToken = agentRunOptions.ContinuationToken;
```
rogerbarreto, line 544: When `chatOptions.AllowBackground` is not null but `agentRunOptions.Allowbackground` is null, this will override the `chatOptions.AllowBackground` to be `null`.

What happens when both values are set but mismatch (chatOptions vs runOptions) ? Throw `InvalidOperationException` ?

```suggestion
                chatOptions.AllowBackgroundResponses ??= agentRunOptions.AllowBackgroundResponses;
                chatOptions.ContinuationToken ??= agentRunOptions.ContinuationToken;
```

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
```diff
@@ -551,49 +564,66 @@ await thread.AIContextProvider.InvokedAsync(new(inputMessages) { InvokeException
     {
         ChatOptions? chatOptions = this.CreateConfiguredChatOptions(runOptions);
 
+        // Supplying a thread for background responses is required to prevent inconsistent experience
+        // for callers if they forget to provide the thread for initial or follow-up runs.
+        if (chatOptions?.AllowBackgroundResponses is true && thread is null)
+        {
+            throw new InvalidOperationException("A thread must be provided when continuing a background response using a continuation token.");
```
rogerbarreto: nit

```suggestion
            throw new InvalidOperationException("A thread must be provided when continuing a background response with a continuation token.");
```

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
```diff
@@ -551,49 +564,66 @@ await thread.AIContextProvider.InvokedAsync(new(inputMessages) { InvokeException
...
         }
 
-        // Add any existing messages from the thread to the messages to be sent to the chat client.
-        List<ChatMessage> threadMessages = [];
-        if (typedThread.MessageStore is not null)
+        // Supplying messages when continuing a background response is not allowed.
+        if (chatOptions?.ContinuationToken is not null && inputMessages.Any())
         {
-            threadMessages.AddRange(await typedThread.MessageStore.GetMessagesAsync(cancellationToken).ConfigureAwait(false));
+            throw new InvalidOperationException("Messages are not allowed when continuing a background response using a continuation token.");
```
rogerbarreto:
```suggestion
            throw new InvalidOperationException("Input messages are not allowed when continuing a background response using a continuation token.");
```

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
```diff
@@ -551,49 +564,66 @@ await thread.AIContextProvider.InvokedAsync(new(inputMessages) { InvokeException
...
-                chatOptions ??= new();
-                chatOptions.Instructions = string.IsNullOrWhiteSpace(chatOptions.Instructions) ? aiContext.Instructions : $"{chatOptions.Instructions}\n{aiContext.Instructions}";
+                if (aiContext.Tools is { Count: > 0 })
+                {
+                    chatOptions ??= new();
+                    chatOptions.Tools ??= [];
+                    foreach (AITool tool in aiContext.Tools)
+                    {
+                        chatOptions.Tools.Add(tool);
+                    }
```
rogerbarreto, line 614: nit Optimization.

```suggestion
                    chatOptions.Tools ??= [];
                    if (chatOptions.Tools is List<AITool> toolsList) 
                    {
                        toolsList.AddRange(aiContext.Tools)
                    }
                    else 
                    {
                        foreach (AITool tool in aiContext.Tools)
                        {
                            chatOptions.Tools.Add(tool);
                        }
                    }
```

dotnet/tests/Microsoft.Agents.AI.UnitTests/ChatClient/ChatClientAgentTests.cs
```diff
@@ -1951,6 +1951,499 @@ public void GetNewThreadUsesAIContextProviderFactoryIfProvided()
...
+
+        // Assert
+        Assert.NotNull(capturedChatOptions);
+
+        Assert.True(capturedChatOptions.AllowBackgroundResponses);
+        Assert.Same(continuationToken, capturedChatOptions.ContinuationToken);
+    }
+
+    [Fact]
+    public async Task RunStreamingAsyncPrioritizesBackgroundResponsesPropertiesFromAgentRunOptionsOverOnesFromChatOptionsAsync()
```
rogerbarreto, line 2107: I think this test should ensure that both are always the same after applying the suggetion of the `virtual Allow`.

So if you are using a `ChatClientAgentRunOptions` and set the `AllowBackground` it would point the the `ChatOptions.Allow` and vice versa.

```suggestion
    public async Task RunStreamingAsyncPrioritizesBackgroundResponsesPropertiesFromAgentRunOptionsOverOnesFromChatOptionsAsync()
```

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
SergeyMenshykh: > When chatOptions.AllowBackground is not null but agentRunOptions.Allowbackground is null, this will override the chatOptions.AllowBackground to be null.

Yep, the idea is that agent run options overrides whatever is set in chat optons.

> What happens when both values are set but mismatch (chatOptions vs runOptions) ? Throw InvalidOperationException ?

Values from agent run options override the ones supplied via chat options.

dotnet/src/Microsoft.Agents.AI.Abstractions/AgentRunResponseUpdate.cs
SergeyMenshykh: Thanks, no need in these pragmas anymore, removed.

dotnet/samples/GettingStarted/Agents/Agent_Step17_BackgroundResponses/Program.cs
rogerbarreto: Never mind, I think is better keep as is.

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
SergeyMenshykh: It's outside the scope of this PR.

dotnet/src/Microsoft.Agents.AI.Abstractions/AgentRunOptions.cs
SergeyMenshykh: Prototyped and discussed the idea in a call and decided not to proceed with it because it does not introduce the benefits it aims to and unnecessarily duplicates existing fallback logic in ChatClientAgent.

dotnet/tests/Microsoft.Agents.AI.UnitTests/ChatClient/ChatClientAgentTests.cs
SergeyMenshykh: Not relevant now that we have decided not to go with the virtual approach.

dotnet/src/Microsoft.Agents.AI.Abstractions/AgentRunOptions.cs
```diff
@@ -10,7 +10,7 @@ namespace Microsoft.Agents.AI;
 /// </summary>
 /// <remarks>
 /// <para>
-/// This class currently has no options, but may be extended in the future to include additional configuration settings.
+/// This class currently has minimal options, but may be extended in the future to include additional configuration settings.
 /// </para>
```
westey-m: I think we can just remove this entire para now.  It was just to explain why we have an empty options class.
```suggestion
```

dotnet/src/Microsoft.Agents.AI.Abstractions/AgentRunOptions.cs
```diff
@@ -33,5 +33,48 @@ public AgentRunOptions()
...
     }
+
+    /// <summary>
+    /// Gets or sets the continuation token for resuming and getting the result of the agent response identified by this token.
+    /// </summary>
+    /// <remarks>
+    /// This property is used for background responses that can be activated via the <see cref="AllowBackgroundResponses"/>
+    /// property if the <see cref="AIAgent"/> implementation supports them.
+    /// Streamed background responses, such as those returned by default by <see cref="AIAgent.RunStreamingAsync(AgentThread?, AgentRunOptions?, System.Threading.CancellationToken)"/>
+    /// can be resumed if interrupted. This means that a continuation token obtained from the <see cref="AgentRunResponse.ContinuationToken"/>
```
westey-m: Should this be mentioning AgentRunResponseUpdate instead of AgentRunResponse?

dotnet/src/Microsoft.Agents.AI.Abstractions/AgentRunOptions.cs
```diff
@@ -33,5 +33,48 @@ public AgentRunOptions()
...
+    /// of an update just before the interruption occurred can be passed to this property to resume the stream from the point of interruption.
+    /// Non-streamed background responses, such as those returned by <see cref="AIAgent.RunAsync(AgentThread?, AgentRunOptions?, System.Threading.CancellationToken)"/>,
+    /// can be polled for completion by obtaining the token from the <see cref="AgentRunResponse.ContinuationToken"/> property
+    /// and passing it to this property on subsequent calls to <see cref="AIAgent.RunAsync(AgentThread?, AgentRunOptions?, System.Threading.CancellationToken)"/>.
```
westey-m:
```suggestion
    /// and passing it via this property on subsequent calls to <see cref="AIAgent.RunAsync(AgentThread?, AgentRunOptions?, System.Threading.CancellationToken)"/>.
```

dotnet/src/Microsoft.Agents.AI.Abstractions/AgentRunResponse.cs
```diff
@@ -74,6 +74,7 @@ public AgentRunResponse(ChatResponse response)
         this.RawRepresentation = response;
         this.ResponseId = response.ResponseId;
         this.Usage = response.Usage;
+        this.ContinuationToken = response.ContinuationToken; // propagate continuation token
```
westey-m: This looks like one of those random AI generated comments.
```suggestion
        this.ContinuationToken = response.ContinuationToken;
```

dotnet/src/Microsoft.Agents.AI.Abstractions/AgentRunResponse.cs
```diff
@@ -159,6 +160,23 @@ public IList<ChatMessage> Messages
     /// </value>
     public string? ResponseId { get; set; }
 
+    /// <summary>
+    /// Gets or sets the continuation token for getting result of the background agent response.
```
westey-m:
```suggestion
    /// Gets or sets the continuation token for getting the result of a background agent response.
```

dotnet/samples/GettingStarted/Agents/Agent_Step17_BackgroundResponses/Program.cs
```diff
@@ -0,0 +1,70 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+// This sample shows how use background responses with ChatClientAgent and OpenAI Responses.
```
westey-m:
```suggestion
// This sample shows how to use background responses with ChatClientAgent and OpenAI Responses.
```

PR #1517, opened 2025-10-16
.NET [Breaking] Simplify and Refactor ChatclientAgentOptions Ctor + Instructions

### LINKED ISSUE #1453: .NET; Setting tools in ChatClientAgentOptions constructor, while also providing a new ChatOptions override the given tools/instructions
When creating an Agent using `CreateAIAgent` and you need to define AI Reasoning Effort, you need to use the overload that take in a ChatClientAgentOptions (so you can set ChatOptions' RawRepresentationFactory property). 

This however lead to the override of any tool given to the constructor of ChatClientAgentOptions, leading to an invalid state of the Agent where you provided Tools, but they are not define anymore.

IMHO: I would say it is wrong to let the constructor receive the `tools` parameter (and should be removed)

Steps to reproduce:

Paste the following code into a Console App:

```cs
//Define the basics
AzureOpenAIClient azureOpenAiClient = new(new Uri("endpoint"), new ApiKeyCredential("key"));
ChatClient chatClient = azureOpenAiClient.GetChatClient("gpt-5-nano");
IList<AITool> aiTools = [AIFunctionFactory.Create(WhatIsTheUsersName)];
string? myInstructions = "Use tools to answer all questions.";
string myQuestion = "What is my Name?";

//Setup to be able to set Reasoning Effort on a Thinking Model
Func<IChatClient, object?> rawRepresentationFactory = _ => new ChatCompletionOptions()
{
#pragma warning disable OPENAI001
    ReasoningEffortLevel = "low",
#pragma warning restore OPENAI001
};

//This setup of an Agent does not work when ChatOptions are set and Tools are given in the ChatClientAgentOptions constructor
Console.WriteLine("FAILING EXAMPLE:");
ChatClientAgent agentSetupThatFailCallingTools = chatClient.CreateAIAgent(
    options: new ChatClientAgentOptions(instructions: myInstructions, tools: aiTools)
    {
        ChatOptions = new ChatOptions
        {
            RawRepresentationFactory = rawRepresentationFactory
        }
    });

AgentRunResponse response1 = await agentSetupThatFailCallingTools.RunAsync(myQuestion);
Console.WriteLine(response1);

Console.WriteLine("-------------------------------------");

//This setup work (because the tools are given via the chat-options)
Console.WriteLine("WORKING EXAMPLE:");
ChatClientAgent agentSetupThatWork = chatClient.CreateAIAgent(
    options: new ChatClientAgentOptions(instructions: myInstructions)
    {
        ChatOptions = new ChatOptions
        {
            RawRepresentationFactory = rawRepresentationFactory,
            Tools = aiTools
        },
    });

AgentRunResponse response2 = await agentSetupThatWork.RunAsync(myQuestion);
Console.WriteLine(response2);

//Tool
static string WhatIsTheUsersName()
{
    return "John AI";
}
```

rogerbarreto:
### Motivation and Context

- Fixes #1453

### Context 

Before `Microsoft.Extensions.AI.ChatOptions` had the `Instructions` property we added this property into the `AgentOptions.Instructions`

In the past months the `chatOptions.Instructions` property was added 

And now we have `agentOptions.Instructions` serving the same purpose of `agentOptions.ChatOptions.Instructions` when configuring the `ChatClientAgent` which is not ideal. 

As a first step to simplify, the problem above this PR updates `options.Instructions` to pointing to a single reference: e.g: `agentOptions.Instructions => agentOptions.ChatOptions.Instructions`.

> [!NOTE]
> We may consider the full removal of the `options.Instructions` in favor of the `options.ChatOptions.Instructions` as a next step, happy to get some input from community.


dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
```diff
@@ -451,7 +451,13 @@ await thread.AIContextProvider.InvokedAsync(new(inputMessages) { InvokeException
         requestChatOptions.AllowMultipleToolCalls ??= this._agentOptions.ChatOptions.AllowMultipleToolCalls;
         requestChatOptions.ConversationId ??= this._agentOptions.ChatOptions.ConversationId;
         requestChatOptions.FrequencyPenalty ??= this._agentOptions.ChatOptions.FrequencyPenalty;
-        requestChatOptions.Instructions ??= this._agentOptions.ChatOptions.Instructions;
+
+        requestChatOptions.Instructions = !string.IsNullOrEmpty(requestChatOptions.Instructions) && !string.IsNullOrEmpty(this.Instructions)
```
westey-m:
```suggestion
        requestChatOptions.Instructions = !string.IsNullOrWhitespace(requestChatOptions.Instructions) && !string.IsNullOrWhitespace(this.Instructions)
```

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
```diff
@@ -451,7 +451,13 @@ await thread.AIContextProvider.InvokedAsync(new(inputMessages) { InvokeException
...
+            ? $"{this.Instructions}\n{requestChatOptions.Instructions}"
+            : (!string.IsNullOrEmpty(requestChatOptions.Instructions)
```
westey-m:
```suggestion
            : (!string.IsNullOrWhitespace(requestChatOptions.Instructions)
```

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgentOptions.cs
```diff
@@ -69,7 +62,31 @@ public ChatClientAgentOptions(string? instructions, string? name = null, string?
...
+            if (providedOptions is not null)
+            {
+                // Ensure immutable copy of provided options.
+                providedOptions = providedOptions.Clone();
+            }
+
+            if (this._chatOptions is not null && providedOptions is not null && string.IsNullOrWhiteSpace(providedOptions.Instructions))
+            {
+                // Preserve existing agent options instructions if new ChatOptions does not have instructions set.
+                providedOptions.Instructions = this._chatOptions.Instructions;
```
westey-m: I would just remove the Instructions property from here to avoid this complexity.

The problem with this instructions prop approach is that it is unclear whether the user intended to replace instructions as well with the chatoptions replacement or not.

Most places where someone has to construct a ChatClientAgentOptions they do it since they need to set some other property on ChatOptions anyway, so it doesn't add much complexity to require them to set instructions there too.

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
```diff
@@ -59,13 +59,13 @@ public ChatClientAgent(IChatClient chatClient, string? instructions = null, stri
               chatClient,
               new ChatClientAgentOptions
               {
-                  Name = name,
-                  Description = description,
-                  Instructions = instructions,
                   ChatOptions = tools is null ? null : new ChatOptions
                   {
                       Tools = tools,
-                  }
```
markwallace-microsoft, line 68: Why are the tools no longer being set?

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgentOptions.cs
```diff
@@ -59,7 +38,21 @@ public ChatClientAgentOptions(string? instructions, string? name = null, string?
...
+            if (value is null && this._chatOptions is null)
+            {
+                // No instructions to set and no existing chat options, nothing to do.
+                return;
+            }
+
+            this._chatOptions ??= new();
+            this._chatOptions.Instructions = value;
+        }
+    }
```
markwallace-microsoft: Should we just remove this now. Having two Instructions properties where one points to the other seems confusing.

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgentOptions.cs
```diff
@@ -59,7 +38,21 @@ public ChatClientAgentOptions(string? instructions, string? name = null, string?
...
     /// </summary>
-    public string? Instructions { get; set; }
+    public string? Instructions
+    {
+        get => this._chatOptions?.Instructions;
+        set
+        {
+            if (value is null && this._chatOptions is null)
+            {
+                // No instructions to set and no existing chat options, nothing to do.
```
markwallace-microsoft: If `_chatOptions` exists should the instructions be set to null?

MithrilMan:
any chance this PR can help having custom options so that we can use gemini with the openai compatibility API 
https://ai.google.dev/gemini-api/docs/openai

and use custom options like you see in the Thinking section of their guide
https://ai.google.dev/gemini-api/docs/openai#thinking

where it is even possible to disable thinking at all or even obtain thinking tokens?
```
curl "https://generativelanguage.googleapis.com/v1beta/openai/chat/completions" \
-H "Content-Type: application/json" \
-H "Authorization: Bearer GEMINI_API_KEY" \
-d '{
    "model": "gemini-2.5-flash",
      "messages": [{"role": "user", "content": "Explain to me how AI works"}],
      "extra_body": {
        "google": {
           "thinking_config": {
             "include_thoughts": true
           }
        }
      }
    }'
```

westey-m:
@MithrilMan, any options that are supported by the OpenAI SDK can be provided.  ChatOptions allows you to provide the options for the underlying service via the `RawRepresentationFactory` property, whether that be chat completion, responses, or any other service.

See here for example where it is providing reasoning options for responses via `ChatOptions.RawRepresentationFactory`:
https://github.com/microsoft/agent-framework/blob/main/dotnet/samples/GettingStarted/AgentWithOpenAI/Agent_OpenAI_Step02_Reasoning/Program.cs#L18
(Note that chat options can also be set on ChatClientAgentOptions.ChatOptions, it doesn't have to be via ChatClientBuilder.ConfigureOptions).

See e.g. for responses, this is the options object, and it does have some reasoning settings:
https://github.com/openai/openai-dotnet/blob/main/src/Custom/Responses/ResponseCreationOptions.cs#L46

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
rogerbarreto: They are, the tool line is unchanged.

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgentOptions.cs
rogerbarreto: I updated the PR, removing the `agentOptions.Instructions` this is not necessary anymore.

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgentOptions.cs
rogerbarreto: Yes, I commented in the Description of this PR considering this option, I'm updating this PR for the simplification.

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgentOptions.cs
rogerbarreto: Agreed, doing this now.

MithrilMan:
@westey-m Yeah I saw that but the problem is that google offers more control over thinking that OpenAI doesn't.
E.g. on google you can switch off thinking but OpenAI doesn't have an option for `ResponseReasoningEffortLevel` of none.

Also in the documentation I linked, they show how you can set specific thinking budget (something OpenAI doesn't) and how to receive thinking tokens to see the internal thinking process of the LLM.

I understand that you can't add explicit value on OpenAI options that the OpenAI doesn't handle, but you could allow custom values that the user can set manually

**_Of course would be better to have a specific gemini client with its option_**

westey-m:
> I understand that you can't add explicit value on OpenAI options that the OpenAI doesn't handle, but you could allow custom values that the user can set manually

@MithrilMan, definitely open an issue in the OpenAI SDK repo with your requirement: https://github.com/openai/openai-dotnet.
We are building on top of that, so even supporting the ability to pass unofficial values is something that needs to be supported in the OpenAI SDK.

MithrilMan:
@westey-m I've the feeling it would be something they don't have the incentive in doing it, since it would mean supporting potentially competitors.

Anyway does your reply means you are not gonna support gemini as first citizen, with dedicated client api classes?

dotnet/samples/GettingStarted/Agents/Agent_Step05_StructuredOutput/Program.cs
```diff
@@ -22,7 +22,7 @@
         .GetChatClient(deploymentName);
 
 // Create the ChatClientAgent with the specified name and instructions.
-ChatClientAgent agent = chatClient.CreateAIAgent(new ChatClientAgentOptions() { Name = "HelpfulAssistant", Instructions = "You are a helpful assistant." });
+ChatClientAgent agent = chatClient.CreateAIAgent(new ChatClientAgentOptions() { Name = "HelpfulAssistant", ChatOptions = new() { Instructions = "You are a helpful assistant." } });
```
westey-m: For this sample, we should be able to switch to using `CreateAIAgent(name: "HelpfulAssistant", instructions: "You are a helpful assistant.")`

westey-m (approved):

.github/upgrades/prompts/SemanticKernelToAgentFramework.md
```diff
@@ -1,1639 +1,5 @@
-# Instructions for migrating from Semantic Kernel Agents to Agent Framework in .NET projects.
+# [Instructions for migrating from Semantic Kernel Agents to Agent Framework in .NET projects.](https://github.com/microsoft/semantic-kernel/blob/main/.github/upgrades/prompts/SemanticKernelToAgentFramework.md)
 
-## Scope
+Instructions moved to Semantic Kernel repository:
```
westey-m: NIT: Referring to the fact that it used to be here, is probably unnecessary.
```suggestion
See the Semantic Kernel repository for instructions:
```

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
```diff
@@ -451,7 +451,13 @@ await thread.AIContextProvider.InvokedAsync(new(inputMessages) { InvokeException
         requestChatOptions.AllowMultipleToolCalls ??= this._agentOptions.ChatOptions.AllowMultipleToolCalls;
         requestChatOptions.ConversationId ??= this._agentOptions.ChatOptions.ConversationId;
         requestChatOptions.FrequencyPenalty ??= this._agentOptions.ChatOptions.FrequencyPenalty;
-        requestChatOptions.Instructions ??= this._agentOptions.ChatOptions.Instructions;
+
+        requestChatOptions.Instructions = !string.IsNullOrWhiteSpace(requestChatOptions.Instructions) && !string.IsNullOrWhiteSpace(this.Instructions)
+            ? $"{this.Instructions}\n{requestChatOptions.Instructions}"
+            : (!string.IsNullOrWhiteSpace(requestChatOptions.Instructions)
+            ? requestChatOptions.Instructions
+            : this.Instructions);
```
SergeyMenshykh: Please move this block of code that merges instructions down below where more complex merging is done. Nit for readability: consider using "if" for readability, as it takes a little bit of time to untangle two ternary operators.

SergeyMenshykh (approved):

SergeyMenshykh (approved):

markwallace-microsoft:
> @westey-m I've the feeling it would be something they don't have the incentive in doing it, since it would mean supporting potentially competitors.
> 
> Anyway does your reply means you are not gonna support gemini as first citizen, with dedicated client api classes?

For Gemini you can reuse the Semantic Kernel `GoogleAIGeminiChatCompletionService` which can be converted to an IChatClient and used with `ChatClientAgent`.

We are working to push the IChatClient implementations into existing client SDK's or M.E.AI.

markwallace-microsoft (approved):

dotnet/tests/Microsoft.Agents.AI.UnitTests/ChatClient/ChatClientAgentTests.cs
```diff
@@ -882,7 +970,7 @@ public void ConstructorUsesOptionalParams()
     /// Verify that ChatOptions property returns null when no params are provided that require a ChatOptions instance.
     /// </summary>
     [Fact]
-    public void ChatOptionsToolsReturnsNullWhenConstructorToolsNotProvided()
+    public void ChatOptionsReturnsNullWhenConstructorToolsNotProvided()
```
westey-m: The test name says that chat options returns null, but the test checks for not null.

westey-m (approved):

dotnet/tests/Microsoft.Agents.AI.UnitTests/ChatClient/ChatClientAgentTests.cs
```diff
@@ -970,7 +970,7 @@ public void ConstructorUsesOptionalParams()
     /// Verify that ChatOptions property returns null when no params are provided that require a ChatOptions instance.
```
westey-m:
```suggestion
    /// Verify that ChatOptions is created with instructions when instructions are provided and no tools are provided.
```

markwallace-microsoft (approved):

westey-m (approved):

markwallace-microsoft (approved):

lindexi:
The document (https://github.com/MicrosoftDocs/semantic-kernel-docs/blob/main/agent-framework/tutorials/agents/structured-output.md ) do not be updated. 

The sample code use the `ChatClientAgentOptions.Instructions`:

```csharp
using System;
using Azure.AI.OpenAI;
using Azure.Identity;
using Microsoft.Agents.AI;
using OpenAI;

    new Uri("https://<myresource>.openai.azure.com"),
    new AzureCliCredential())
        .GetChatClient("gpt-4o-mini")
        .CreateAIAgent(new ChatClientAgentOptions()
        {
            Name = "HelpfulAssistant",
            Instructions = "You are a helpful assistant.",
            ChatOptions = chatOptions
        });
```

PR #1520, opened 2025-10-16
.NET: API to manage AgentThreads in hosting scenarios

DeagleGross:
This is a prototype / design proposal for the API which allows managing `AgentThread` in relationship with `AIAgent`.

PR showcases a simple scenario where Hosting.A2A package now obtains an ability to load the thread from in-app-memory based on `contextId` (input of the user), and `agentId` (the id of a specific `AIAgent`). It then saves the agent thread content back via serialization.

Default behavior is to not have any thread store - the `NoContextAgentThreadStore` simply never saves or returns a thread. I think this is useful to not force any default behavior here, because storing all agent thread content in memory can quickly drain all of memory.

The API added: `IAgentThreadStore` which regulates how agent thread data is being stored:
```diff
+ public interface IAgentThreadStore
{
+    ValueTask SaveThreadAsync( string conversationId, string agentId, AgentThread thread, CancellationToken cancellationToken = default);
+    ValueTask<JsonElement?> GetThreadAsync( string conversationId, string agentId, CancellationToken cancellationToken = default);
}
```

and `AIHostAgent` which wraps the `AIAgent` and a specific `IAgentThreadStore` implementation.
```diff
+ public class AIHostAgent : AIAgent
{
+ public async ValueTask<AgentThread> GetOrCreateThreadAsync(string conversationId, CancellationToken cancellationToken = default)

+ public async ValueTask SaveThreadAsync(string conversationId, AgentThread thread, CancellationToken cancellationToken = default)
}
```

## Schema of flow

This is a flow which shows the implementation as in the current PR. It leverages the separate `IAgentThreadStore` which is used to . It definitely works, uses the existing `AIAgent` abstraction and uses a separate interface to save/restore thread via the conversationId.

```mermaid
sequenceDiagram
  actor User as User
  participant code as Server Endpoint Handler
  participant store as IAgentThreadStore
  participant db as IDatabase
  participant agent as AIAgent
  participant msgStore as IMessageStore
  User ->> code: POST /v1/responses (conversationId)
    activate code
  code ->> store: GetThreadAsync(agent, conversationId)
    activate store
  store -->> db: JsonElement GetState(conversationId)
    activate db
  db -->> store: 
    deactivate db  
  store -->> code: state not null ? AIAgent.DeserializeThread(state, options)
  store -->> code: state null ? AIAgent.GetNewThread()
    deactivate store
  code ->> agent: AIAgent.RunAsync(messages, thread)
    activate agent
  agent ->> msgStore: GetMessagesAsync()
    activate msgStore
  msgStore -->> agent:
    deactivate msgStore
  agent -->> code:
    deactivate agent
  code ->> store: SaveThreadAsync(agent, thread)
    activate store
  store ->> db: SaveState(thread) as JsonElement
    activate db
  db -->> store:
    deactivate db
  store -->> code: 
    deactivate store
  code ->> User: { Response as JSON }
    deactivate code
```

This is a "direct" flow, where no "extra" store is used, but the restoration of thread happens inside of the same MessageStore.

```mermaid
sequenceDiagram
  actor User as User
  participant code as Server Endpoint Handler
  participant agent as AIHostAgent
  participant someStore as ? (maybe MessageStore)
  participant msgStore as IMessageStore
  User ->> code: POST /v1/responses (conversationId)
    activate code
  code ->> agent: AIHostAgent.RestoreThread(conversationId)
    activate agent
    agent ->> someStore:
    someStore -->> agent:
  agent -->> code:
    deactivate agent
  code ->> agent: AIAgent.RunAsync(messages, thread)
    activate agent
  agent ->> msgStore: GetMessagesAsync()
    activate msgStore
  msgStore -->> agent:
    deactivate msgStore
  agent -->> code:
    deactivate agent
  code ->> User: { Response as JSON }
    deactivate code
```

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.Hosting/AIHostAgent.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Hosting/AIHostAgent.cs
@@ -14,7 +14,7 @@ namespace Microsoft.Agents.AI.Hosting;
 /// <remarks>
 /// <para>
 /// <see cref="AIHostAgent"/> wraps an existing agent implementation and adds the ability to
-/// persist and restore conversation threads using an <see cref="IAgentThreadStore"/>.
+/// persist and restore conversation threads using an <see cref="AgentThreadStore"/>.
 /// </para>
 /// <para>
 /// This wrapper enables thread persistence without requiring type-specific knowledge of the thread type,
@@ -23,7 +23,7 @@ namespace Microsoft.Agents.AI.Hosting;
 /// </remarks>
 public class AIHostAgent : DelegatingAIAgent
 {
-    private readonly IAgentThreadStore _threadStore;
+    private readonly AgentThreadStore _threadStore;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="AIHostAgent"/> class.
@@ -33,7 +33,7 @@ public class AIHostAgent : DelegatingAIAgent
     /// <exception cref="ArgumentNullException">
     /// <paramref name="innerAgent"/> or <paramref name="threadStore"/> is <see langword="null"/>.
     /// </exception>
-    public AIHostAgent(AIAgent innerAgent, IAgentThreadStore threadStore)
+    public AIHostAgent(AIAgent innerAgent, AgentThreadStore threadStore)
         : base(innerAgent)
     {
         this._threadStore = Throw.IfNull(threadStore);
--- a/dotnet/src/Microsoft.Agents.AI.Hosting/HostedAgentBuilderExtensions.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Hosting/HostedAgentBuilderExtensions.cs
@@ -18,7 +18,7 @@ public static class HostedAgentBuilderExtensions
     /// <returns>The same <paramref name="builder"/> instance, configured to use an in-memory thread store.</returns>
     public static IHostedAgentBuilder WithInMemoryThreadStore(this IHostedAgentBuilder builder)
     {
-        builder.ServiceCollection.AddKeyedSingleton<IAgentThreadStore>(builder.Name, new InMemoryAgentThreadStore());
+        builder.ServiceCollection.AddKeyedSingleton<AgentThreadStore>(builder.Name, new InMemoryAgentThreadStore());
         return builder;
     }
 
@@ -29,7 +29,7 @@ public static class HostedAgentBuilderExtensions
     /// <param name="builder">The host agent builder to configure with the thread store. Cannot be null.</param>
     /// <param name="store">The agent thread store instance to register. Cannot be null.</param>
     /// <returns>The same host agent builder instance, allowing for method chaining.</returns>
-    public static IHostedAgentBuilder WithThreadStore(this IHostedAgentBuilder builder, IAgentThreadStore store)
+    public static IHostedAgentBuilder WithThreadStore(this IHostedAgentBuilder builder, AgentThreadStore store)
     {
         builder.ServiceCollection.AddKeyedSingleton(builder.Name, store);
         return builder;
@@ -42,7 +42,7 @@ public static class HostedAgentBuilderExtensions
     /// <param name="createAgentThreadStore">A factory function that creates an agent thread store instance using the provided service provider and agent
     /// name.</param>
     /// <returns>The same host agent builder instance, enabling further configuration.</returns>
-    public static IHostedAgentBuilder WithThreadStore(this IHostedAgentBuilder builder, Func<IServiceProvider, string, IAgentThreadStore> createAgentThreadStore)
+    public static IHostedAgentBuilder WithThreadStore(this IHostedAgentBuilder builder, Func<IServiceProvider, string, AgentThreadStore> createAgentThreadStore)
     {
         builder.ServiceCollection.AddKeyedSingleton(builder.Name, (sp, key) =>
         {
@@ -52,7 +52,7 @@ public static class HostedAgentBuilderExtensions
             var store = createAgentThreadStore(sp, keyString);
             if (store is null)
             {
-                throw new InvalidOperationException($"The agent thread store factory did not return a valid {nameof(IAgentThreadStore)} instance for key '{keyString}'.");
+                throw new InvalidOperationException($"The agent thread store factory did not return a valid {nameof(AgentThreadStore)} instance for key '{keyString}'.");
             }
 
             return store;
--- a/dotnet/src/Microsoft.Agents.AI.Hosting/IAgentThreadStore.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Hosting/IAgentThreadStore.cs
@@ -13,7 +13,7 @@ namespace Microsoft.Agents.AI.Hosting;
 /// allowing conversations to be resumed across HTTP requests, application restarts,
 /// or different service instances in hosted scenarios.
 /// </remarks>
-public interface IAgentThreadStore
+public abstract class AgentThreadStore
 {
     /// <summary>
     /// Saves a serialized agent thread to persistent storage.
@@ -23,7 +23,7 @@ public interface IAgentThreadStore
     /// <param name="thread">The thread to save.</param>
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests.</param>
     /// <returns>A task that represents the asynchronous save operation.</returns>
-    ValueTask SaveThreadAsync(
+    public abstract ValueTask SaveThreadAsync(
         AIAgent agent,
         string conversationId,
         AgentThread thread,
@@ -39,7 +39,7 @@ public interface IAgentThreadStore
     /// A task that represents the asynchronous retrieval operation.
     /// The task result contains the serialized thread state, or <see langword="null"/> if not found.
     /// </returns>
-    ValueTask<AgentThread> GetThreadAsync(
+    public abstract ValueTask<AgentThread> GetThreadAsync(
         AIAgent agent,
         string conversationId,
         CancellationToken cancellationToken = default);
--- a/dotnet/src/Microsoft.Agents.AI.Hosting/Local/InMemoryAgentThreadStore.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Hosting/Local/InMemoryAgentThreadStore.cs
@@ -8,7 +8,7 @@ using System.Threading.Tasks;
 namespace Microsoft.Agents.AI.Hosting;
 
 /// <summary>
-/// Provides an in-memory implementation of <see cref="IAgentThreadStore"/> for development and testing scenarios.
+/// Provides an in-memory implementation of <see cref="AgentThreadStore"/> for development and testing scenarios.
 /// </summary>
 /// <remarks>
 /// <para>
@@ -25,12 +25,12 @@ namespace Microsoft.Agents.AI.Hosting;
 /// such as Redis, SQL Server, or Azure Cosmos DB.
 /// </para>
 /// </remarks>
-public sealed class InMemoryAgentThreadStore : IAgentThreadStore
+public sealed class InMemoryAgentThreadStore : AgentThreadStore
 {
     private readonly ConcurrentDictionary<string, JsonElement> _threads = new();
 
     /// <inheritdoc/>
-    public ValueTask SaveThreadAsync(AIAgent agent, string conversationId, AgentThread thread, CancellationToken cancellationToken = default)
+    public override ValueTask SaveThreadAsync(AIAgent agent, string conversationId, AgentThread thread, CancellationToken cancellationToken = default)
     {
         var key = GetKey(conversationId, agent.Id);
         this._threads[key] = thread.Serialize();
@@ -38,7 +38,7 @@ public sealed class InMemoryAgentThreadStore : IAgentThreadStore
     }
 
     /// <inheritdoc/>
-    public ValueTask<AgentThread> GetThreadAsync(AIAgent agent, string conversationId, CancellationToken cancellationToken = default)
+    public override ValueTask<AgentThread> GetThreadAsync(AIAgent agent, string conversationId, CancellationToken cancellationToken = default)
     {
         var key = GetKey(conversationId, agent.Id);
         JsonElement? threadContent = this._threads.TryGetValue(key, out var existingThread) ? existingThread : null;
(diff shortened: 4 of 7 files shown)


dotnet/src/Microsoft.Agents.AI.Hosting/IAgentThreadStore.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// <param name="thread">The thread to save.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests.</param>
+    /// <returns>A task that represents the asynchronous save operation.</returns>
+    /// <remarks>
+    /// The <paramref name="conversationId"/> is used as the primary key for thread lookup.
+    /// The <paramref name="agentId"/> provides additional scoping to prevent cross-agent thread access.
+    /// </remarks>
+    ValueTask SaveThreadAsync(
+        string conversationId,
+        string agentId,
```
ReubenBond: use AIAgent instead

dotnet/src/Microsoft.Agents.AI.Hosting/IAgentThreadStore.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// <param name="conversationId">The unique identifier for the conversation/thread to retrieve.</param>
+    /// <param name="agentId">The identifier of the agent that owns this thread.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests.</param>
+    /// <returns>
+    /// A task that represents the asynchronous retrieval operation.
+    /// The task result contains the serialized thread state, or <see langword="null"/> if not found.
+    /// </returns>
+    ValueTask<JsonElement?> GetThreadAsync(
+        string conversationId,
+        string agentId,
```
ReubenBond: Use `AIAgent` instead here, and return `AgentThread`?

Make `AIAgent` the first arg for both methods

dotnet/src/Microsoft.Agents.AI.Hosting/NoContextAgentThreadStore.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Agents.AI.Hosting;
+
+/// <summary>
+/// This store implementation does not have any store under the hood and operates with empty threads.
+/// It the "noop" store, and could be used if you are keeping the thread contents on the client side for example.
+/// </summary>
+public sealed class NoContextAgentThreadStore : IAgentThreadStore
```
markwallace-microsoft: The naming here isn't clear, should this be `NoopAgentThreadStore`.

dotnet/src/Microsoft.Agents.AI.Hosting/IAgentThreadStore.cs
```diff
@@ -0,0 +1,46 @@
...
+
+/// <summary>
+/// Defines the contract for storing and retrieving agent conversation threads.
+/// </summary>
+/// <remarks>
+/// Implementations of this interface enable persistent storage of conversation threads,
+/// allowing conversations to be resumed across HTTP requests, application restarts,
+/// or different service instances in hosted scenarios.
+/// </remarks>
+public interface IAgentThreadStore
```
markwallace-microsoft: @westey-m Please take a look at this.

dotnet/src/Microsoft.Agents.AI.Hosting/Local/InMemoryAgentThreadStore.cs
```diff
@@ -0,0 +1,55 @@
...
+    /// <inheritdoc/>
+    public ValueTask<AgentThread> GetThreadAsync(AIAgent agent, string conversationId, CancellationToken cancellationToken = default)
+    {
+        var key = GetKey(conversationId, agent.Id);
+        JsonElement? threadContent = this._threads.TryGetValue(key, out var existingThread)
+            ? existingThread : null!;
+
+        return threadContent switch
+        {
+            null => new ValueTask<AgentThread>(agent.GetNewThread()),
```
markwallace-microsoft, line 48: Should the new thread be added to the `_threads` dictionary?

dotnet/src/Microsoft.Agents.AI.Hosting/IAgentThreadStore.cs
westey-m: Also consider a base abstract class, since it is possible to extend with new virtual methods without breaking implementers, even in .net framework.

dotnet/src/Microsoft.Agents.AI.Hosting/HostedAgentBuilderExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    /// <summary>
+    /// Configures the host agent builder to use a custom thread store implementation for agent threads.
+    /// </summary>
+    /// <param name="builder">The host agent builder to configure.</param>
+    /// <param name="createAgentThreadStore">A factory function that creates an agent thread store instance using the provided service provider and agent
+    /// name.</param>
+    /// <returns>The same host agent builder instance, enabling further configuration.</returns>
+    public static IHostedAgentBuilder WithThreadStore(this IHostedAgentBuilder builder, Func<IServiceProvider, string, IAgentThreadStore> createAgentThreadStore)
```
westey-m: Would it make sense to support an optional `ServiceLifetime lifetime = ServiceLifetime.Singleton` parameter?
E.g. For scenarios where storage is accessed with a OBO token and therefore needs to created be per request?

dotnet/src/Microsoft.Agents.AI.Hosting/AIHostAgent.cs
```diff
@@ -0,0 +1,113 @@
...
+/// <para>
+/// <see cref="AIHostAgent"/> wraps an existing agent implementation and adds the ability to
+/// persist and restore conversation threads using an <see cref="IAgentThreadStore"/>.
+/// </para>
+/// <para>
+/// This wrapper enables thread persistence without requiring type-specific knowledge of the thread type,
+/// as all thread operations work through the base <see cref="AgentThread"/> abstraction.
+/// </para>
+/// </remarks>
+public class AIHostAgent : AIAgent
```
westey-m: Would inheriting from `DelegatingAIAgent` reduce the boilerplate in this class?

dotnet/src/Microsoft.Agents.AI.Hosting/AIHostAgent.cs
```diff
@@ -0,0 +1,113 @@
...
+
+    /// <inheritdoc/>
+    public override string? Name => this._innerAgent.Name;
+
+    /// <inheritdoc/>
+    public override string? Description => this._innerAgent.Description;
+
+    /// <inheritdoc/>
+    public override object? GetService(Type serviceType, object? serviceKey = null)
+        => base.GetService(serviceType, serviceKey) ?? this._innerAgent.GetService(serviceType, serviceKey);
```
westey-m: It's probably unlikely that this class would be decorated, but if there is the chance, consider also exposing the `IAgentThreadStore` here.

dotnet/src/Microsoft.Agents.AI.Hosting/Local/InMemoryAgentThreadStore.cs
DeagleGross: I think it's ok to be added only in `SaveThreadAsync()`; I will revisit this once using this in other places to make sure implementation is consistent. Maybe your proposition makes sense for the case where conversation is added, but then you can immediately fetch it, however caller does not get the conversationId until full response is returned, so not sure yet.

I will revisit this for sure, marking as resolved for current PR

dotnet/src/Microsoft.Agents.AI.Hosting/IAgentThreadStore.cs
DeagleGross: I would start with interface for now - we are anyway in preview, so that we can see if interface works better. Let me know if you are strongly against this

dotnet/src/Microsoft.Agents.AI.Hosting/AIHostAgent.cs
DeagleGross: yes!

dotnet/src/Microsoft.Agents.AI.Hosting/AIHostAgent.cs
DeagleGross: lets keep it non-public for now, and we can see if we need to expose it

dotnet/src/Microsoft.Agents.AI.Hosting/HostedAgentBuilderExtensions.cs
DeagleGross: Today `AIAgent` has a singleton lifetime, and we are reusing same instance of `AIHostAgent` for all endpoint handler invocations today:

```csharp
var hostAgent = new AIHostAgent(
    innerAgent: agent,
    threadStore: agentThreadStore ?? new NoopAgentThreadStore());
...

async Task<A2AResponse> OnMessageReceivedAsync(MessageSendParams messageSendParams, CancellationToken cancellationToken)
{
   ... await hostAgent.RunAsync()...
}
```

That means it will be captured once and reused accross invocations - lets keep it as is today, and I can see if we can allow creating instances with different lifetimes in a separate PR. Let me know if you are OK with that!

westey-m: Sure, I think it's OK for this PR, but probably something we need to consider in the future.

dotnet/src/Microsoft.Agents.AI.Hosting/IAgentThreadStore.cs
westey-m: I'm not necessarily against Interfaces, but make sure you make a decision about this before GA.  Once we go GA we will not be able to extend any interface.

westey-m (approved):

dotnet/src/Microsoft.Agents.AI.Hosting/IAgentThreadStore.cs
SergeyMenshykh: If, with high probability, we know that it's going to be an abstract class, let's make it one now, because most likely the next time we remember about it will be when we need to update it, but we would not be able to because it'll be a breaking change.

dotnet/src/Microsoft.Agents.AI.Hosting/Local/InMemoryAgentThreadStore.cs
```diff
@@ -0,0 +1,54 @@
...
+
+    /// <inheritdoc/>
+    public ValueTask<AgentThread> GetThreadAsync(AIAgent agent, string conversationId, CancellationToken cancellationToken = default)
+    {
+        var key = GetKey(conversationId, agent.Id);
+        JsonElement? threadContent = this._threads.TryGetValue(key, out var existingThread) ? existingThread : null;
+
+        return threadContent switch
+        {
+            null => new ValueTask<AgentThread>(agent.GetNewThread()),
```
SergeyMenshykh, line 48: I wonder if the store should be responsible for obtaining a new thread from the agent, in addition to its existing "store" responsibilities. Shouldn't the store return null to indicate to the calling code that there is no thread available, allowing the calling/consumer code to decide whether to create a new thread or not? This would provide a clear separation of responsibilities and additional flexibility: the store would be responsible for storage, while the calling code would handle the absence of a thread.

DeagleGross: It's a bit confusing at the moment, I agree. Current implementation is what we decided before via the call with Reuben, Weslie and Stephen: see https://github.com/microsoft/agent-framework/pull/1520#discussion_r2440751082.

The point is that we have a "better" API on the store interface: otherwise I would expose `JsonElement?` which is basically exposing inner implementation details. Maybe not everyone expects that type to be a representation of an `AgentThread`.

also whatever mechanism under the hood **has to return** a thread, otherwise you would not pass anything into `AIAgent.RunAsync()` and thread will not be returned to you. So we have to create a thread (that's why return type is not nullable)

dotnet/src/Microsoft.Agents.AI.Hosting/HostedAgentBuilderExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+/// Provides extension methods for configuring <see cref="AIAgent"/>.
+/// </summary>
+public static class HostedAgentBuilderExtensions
+{
+    /// <summary>
+    /// Configures the host agent builder to use an in-memory thread store for agent thread management.
+    /// </summary>
+    /// <param name="builder">The host agent builder to configure with the in-memory thread store.</param>
+    /// <returns>The same <paramref name="builder"/> instance, configured to use an in-memory thread store.</returns>
+    public static IHostedAgentBuilder WithInMemoryThreadStore(this IHostedAgentBuilder builder)
```
ReubenBond, line 19: This should be `AddInMemoryThreadStore`, IMO. We should make sure we use consistent naming conventions. In this case, there are no extension methods hanging off IHostedAgentBuilder yet, but we can follow what other APIs do.

ReubenBond: Same for the other APIs

DeagleGross: renamed. I was trying to follow what Aspire has (`WithUrls`, `WithEndpoint`), but I dont have a strong opinion.

ReubenBond: Actually, scratch that. I think `With` is good. It does not convey that it's adding services to DI, but I think it's ok

markwallace-microsoft (approved):

dotnet/src/Microsoft.Agents.AI.Hosting/HostedAgentBuilderExtensions.cs
DeagleGross: pushed `with` back

dotnet/src/Microsoft.Agents.AI.Hosting/IAgentThreadStore.cs
DeagleGross: Thanks for suggestion, i've changed to abstract class. Did not have experience building something similar so I believe you !

SergeyMenshykh (approved):

Issue #1390, opened 2025-10-10
.NET: Python: Add support for Azure Cosmos DB as a chat message store and a provider


aayush3011:
- Azure Cosmos DB is used by customers as a memory store to store chat histories, threads and memory.
- Azure Cosmos DB is used by customers as a vector store for vector search. Other search capabilities supported by Cosmos DB are full text, hybrid, fuzzy, and phrase search.

Public docs for the search capabilities supported by Azure Cosmos DB:
- [Vector Search](https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/vector-search)
- [Text Search](https://learn.microsoft.com/en-us/azure/cosmos-db/gen-ai/full-text-search?context=%2Fazure%2Fcosmos-db%2Fnosql%2Fcontext%2Fcontext)
- [Hybrid Search](https://learn.microsoft.com/en-us/azure/cosmos-db/gen-ai/hybrid-search?context=%2Fazure%2Fcosmos-db%2Fnosql%2Fcontext%2Fcontext)

We would like to add support for all this similar to the [Redis Integration](https://github.com/microsoft/agent-framework/tree/main/python/packages/redis/agent_framework_redis)

hemanthsavasere:
@aayush3011 and @markwallace-microsoft I would like to work on this issue. Thanks.

aayush3011:
@hemanthsavasere this is already in the Azure CosmosDB SDK's backlog, and we have started some initial work.

markwallace-microsoft:
Related to #1396

markwallace-microsoft:
Resolved by https://github.com/microsoft/agent-framework/pull/1838

k-miyake:
It looks like #1838  only covers the .NET implementation; I don’t see Python support for the Cosmos DB chat message store yet.

krackinDev:
It looks like the Python version was closed as a duplicate of the original #871, which was actually targeting Python. This issue however references both .NET and Python, but only the .NET implementation was completed before it was closed. It should probably be reopened until both implementations are finished, or a new work item should be created for the Python version so it doesn’t get lost.

[...]

PR #1533, opened 2025-10-16
Python: Add Cosmos DB ChatMessageStore sample example for external chat history

aman-panjwani:
### Motivation and Context

This change introduces a **Cosmos DB–backed `ChatMessageStore`** implementation that enables developers to persist agent chat history externally instead of relying on the default in-memory storage.

1. **Why is this change required?**  
   Many production scenarios require durable and queryable chat history beyond process lifetime, especially when scaling agents horizontally or resuming long-running threads.

2. **What problem does it solve?**  
   Existing examples only demonstrate in-memory or vector-based message storage. This contribution provides a clear, working reference for integrating **Azure Cosmos DB (Core SQL API)** as a persistent external store.

3. **What scenario does it contribute to?**  
   - Persist and retrieve conversation history across sessions  
   - Maintain chronological context reconstruction for resumed threads  
   - Demonstrate thread-based partitioning (`/thread_id`) for scalable storage  
   - Showcase proper async resource management for Cosmos DB and chat clients

4. **Related issue (if any):**  
   N/A – this is a new sample addition.

---

### Description

This PR adds a new sample file:  
`chat_store/third_party_chat_store_cosmosDB.py`

**Key highlights:**

- Implements a lightweight, async `CosmosDBChatMessageStore` for the Microsoft Agent Framework.  
- Uses **key-based authentication** with an existing Cosmos DB database and container.  
- Provides:
  - Thread-scoped message persistence (`partition_key=/thread_id`)
  - Chronological message retrieval (`ORDER BY ts ASC`)
  - Thread state serialization/deserialization for persistence
  - Clean shutdown via `store.aclose()` and `chat_client.aclose()`  
- Includes a minimal usage example integrating with `ChatAgent` and `AzureOpenAIChatClient`.  
---

### Contribution Checklist

- [x] The code builds cleanly without any errors or warnings  
- [x] All existing tests pass, and this sample introduces no breaking dependencies  


aman-panjwani:

python/samples/getting_started/chat_store/third_party_chat_store_cosmosDB.py
```diff
@@ -0,0 +1,237 @@
...
+        - Uses key-based authentication.
+        - Stores one conversation thread per partition (partition key = /thread_id).
+        - Appends messages and retrieves them chronologically.
+        - Supports serialization/deserialization for persistent AgentThread state.
+        - Assumes database and container already exist.
+    """
+
+    def __init__(
+        self,
+        cosmos_endpoint: str,
```
nmoeller: Maybe instead of passing a Endpoint and a Key, you could let user's pass a `CosmosClient` Object here, with this the user can control the Auth himself. This would make your example very flexible.

```
store = CosmosDBChatMessageStore(
        cosmos_client=CosmosClient(url="https://test.de", credentials=DefaultAzureCredential())
        database_name="agent-chat-conversation",
        container_name="chat_messages",
    )

```

python/samples/getting_started/chat_store/third_party_chat_store_cosmos_db.py
```diff
@@ -0,0 +1,237 @@
...
+        self.cosmos_key = cosmos_key
+        self.thread_id = thread_id or f"thread_{uuid4()}"
+        self.database_name = database_name
+        self.container_name = container_name
+
+        self._client: CosmosClient | None = None
+        self._container = None
+        self._ready = False
+
+    async def _ensure(self) -> None:
```
nmoeller, line 68: Does this method need to be `async` ?

python/samples/getting_started/chat_store/third_party_chat_store_cosmosDB.py
```diff
@@ -0,0 +1,237 @@
...
+            raw = message.model_dump()
+        elif hasattr(message, "dict"):
+            raw = message.dict()
+        else:
+            raw = vars(message)
+        return make_safe(raw)
+
+    def _from_dict(self, data: dict) -> ChatMessage:
+        """Reconstruct a ChatMessage from a stored dictionary."""
+        if hasattr(ChatMessage, "model_validate"):
```
nmoeller: Do we really need this `hasattr` checks ? maybe we could only use `ChatMessage.from_dict(data)`.

https://github.com/microsoft/agent-framework/blob/8f87328cf835f889ac87b8b4bec9e5fe48d8df92/python/packages/core/agent_framework/_types.py#L1930

python/samples/getting_started/chat_store/third_party_chat_store_cosmosDB.py
```diff
@@ -0,0 +1,237 @@
...
+                return {k: make_safe(v) for k, v in obj.items()}
+            if hasattr(obj, "dict"):
+                return make_safe(obj.dict())
+            if hasattr(obj, "model_dump"):
+                return make_safe(obj.model_dump())
+            if hasattr(obj, "__dict__"):
+                return make_safe(vars(obj))
+            return str(obj)
+
+        if hasattr(message, "model_dump"):
```
nmoeller: Is it maybe possible to use the `message.to_dict()` method from the ChatMessage Class ? 

https://github.com/microsoft/agent-framework/blob/8f87328cf835f889ac87b8b4bec9e5fe48d8df92/python/packages/core/agent_framework/_types.py#L1930

python/samples/getting_started/chat_store/third_party_chat_store_cosmos_db.py
aman-panjwani: The `_ensure()` method ensures the internal `_container `is ready before any operation (like `add_messages`, `list_messages`, etc.) runs. so it needs to be async also we are using cosmos DB functions all fully asynchronous with `.aio` so All of its methods that hit the network (like `query_items`, `upsert_item`, `get_database_client`, `get_container_client`, etc.) are async-compatible.

aman-panjwani:
Thanks, @nmoeller the idea to package this is excellent and really helpful. I’m planning to implement that. In the meantime, I’ve updated the sample with your suggested changes.

python/samples/getting_started/chat_store/third_party_chat_store_cosmos_db.py
nmoeller: Ok got it, I would expect a `await` in front of the method's `get_database_client` , `get_container_client`, otherwise I think this method maybe does not need declared as `async` . 

The other methods of your store definitely need to be async for sure . Could you test this once if the `async` is really needed ?

python/samples/getting_started/chat_store/third_party_chat_store_cosmos_db.py
```diff
@@ -0,0 +1,171 @@
...
+from azure.identity.aio import DefaultAzureCredential
+
+class CosmosDBStoreState(BaseModel):
+    """Serializable state for CosmosDB chat message store."""
+    thread_id: str
+    database_name: str = "agent_framework"
+    container_name: str = "chat_messages"
+
+
+class CosmosDBChatMessageStore:
```
nmoeller, line 41: Since we write a custom chat message store we should probably inherit from `ChatMessageStoreProtocol` (`from agent_framework import ChatMessageStoreProtocol`)

To fulfill the Protocol, we should also keep the `deserialize_state` and `serialize_state` methods, if possible.

Like in the sample : https://github.com/microsoft/agent-framework/blob/main/python/samples/getting_started/threads/custom_chat_message_store_thread.py

python/samples/getting_started/chat_store/third_party_chat_store_cosmosDB.py
```diff
@@ -0,0 +1,171 @@
...
+      AZURE_OPENAI_ENDPOINT, AZURE_OPENAI_API_KEY,
+      AZURE_OPENAI_API_VERSION, AZURE_OPENAI_CHAT_DEPLOYMENT_NAME
+  - Dependencies:
+      pip install azure-cosmos pydantic "agent-framework"
+
+Run:
+  python third_party_chat_store_cosmosDB.py
+"""
+
+import os
```
nmoeller: Let's move the imports to the top of the file like in all other samples and the docstring below the imports.

python/samples/getting_started/chat_store/third_party_chat_store_cosmosDB.py
```diff
@@ -0,0 +1,171 @@
...
+
+Requirements:
+  - Azure Cosmos DB (Core SQL API) with an existing database and container.
+  - Container partition key must be /thread_id.
+  - Environment variables:
+      COSMOS_DB_ENDPOINT,
+      AZURE_OPENAI_ENDPOINT, AZURE_OPENAI_API_KEY,
+      AZURE_OPENAI_API_VERSION, AZURE_OPENAI_CHAT_DEPLOYMENT_NAME
+  - Dependencies:
+      pip install azure-cosmos pydantic "agent-framework"
```
nmoeller: Pydantic is already part of the AF Core Package

```suggestion
      pip install azure-cosmos "agent-framework"
```

python/samples/getting_started/chat_store/third_party_chat_store_cosmosDB.py
```diff
@@ -0,0 +1,171 @@
...
+
+Demonstrates how to store and retrieve chat history using Azure Cosmos DB
+as an external message store for the Microsoft Agent Framework.
+
+Scenarios:
+  1) Persist chat messages in Cosmos DB with thread-based partitioning.
+  2) Retrieve messages in chronological order for conversation continuity.
+
+Requirements:
+  - Azure Cosmos DB (Core SQL API) with an existing database and container.
```
nmoeller: I think we are using the `Azure Cosmos DB for NoSQL` here right ?

python/samples/getting_started/chat_store/third_party_chat_store_cosmosDB.py
```diff
@@ -0,0 +1,171 @@
...
+  - Container partition key must be /thread_id.
+  - Environment variables:
+      COSMOS_DB_ENDPOINT,
+      AZURE_OPENAI_ENDPOINT, AZURE_OPENAI_API_KEY,
+      AZURE_OPENAI_API_VERSION, AZURE_OPENAI_CHAT_DEPLOYMENT_NAME
+  - Dependencies:
+      pip install azure-cosmos pydantic "agent-framework"
+
+Run:
```
nmoeller: NIT: I would remove the `Run:` part, people can decide themselves how to run this. I am for example using VSCode and uv. Also i would suggest to change the filename to not contain upper letters.

python/samples/getting_started/chat_store/third_party_chat_store_cosmos_db.py
aman-panjwani: We kept `_ensure()` as `async `because it’s part of the async call chain and ensures consistency across all async methods.
The `get_database_client()` and `get_container_client()` calls are sync factory methods, so adding `await `isn’t valid and will result in errors.
Removing `async` would break existing `await _ensure()` calls and force refactoring other functions for no gain.

python/samples/getting_started/chat_store/third_party_chat_store_cosmos_db.py
```diff
@@ -0,0 +1,187 @@
...
+"""
+
+class CosmosDBStoreState(BaseModel):
+    """Serializable state for CosmosDB chat message store."""
+    thread_id: str
+    database_name: str = "agent_framework"
+    container_name: str = "chat_messages"
+
+
+class CosmosDBChatMessageStore:
```
nmoeller, line 41:
```suggestion
class CosmosDBChatMessageStore(ChatMessageStoreProtocol):
```

Let's add the Inheritance for the Protocol with this import `from agent_framework import ChatMessageStoreProtocol`

nmoeller (approved):
One small thing from my side, which is again more editorial.

Thanks for working on this, let's see what the Core Team says to this as well 👍

python/samples/getting_started/chat_store/third_party_chat_store_cosmos_db.py
eavanvalkenburg: That's not needed

eavanvalkenburg:
Closing, the addition of a Cosmos History Provider is covered here: https://github.com/microsoft/agent-framework/issues/1390 we want it built in, not just as a sample

PR #1542, opened 2025-10-17
.NET: reformat event sequence in OpenAI Responses hosting

DeagleGross:
Reformatted event sequence in streaming responses as per spec example: https://platform.openai.com/docs/api-reference/responses/create

image

PR #1550, opened 2025-10-17
.NET: Improve fidelity of OpenAI Responses hosting

ReubenBond:
### Motivation and Context

The current Responses implementation does not match the streaming output produced by Responses.

### Description

This PR migrates from the OpenAI SDK types to generated types for Responses and aligns output with the expected output.

The Responses HTTP API surface is still not fully covered, but this PR does not change that, and we can address it in a subsequent PR. Currently, only the CreateResponse API is supported.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/AgentRunResponseExtensions.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/AgentRunResponseExtensions.cs
@@ -2,7 +2,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.IO;
 using System.Linq;
 using System.Text.Json;
 using Microsoft.Agents.AI.Hosting.OpenAI.Responses.Converters;


stephentoub:
Does this replace https://github.com/microsoft/agent-framework/pull/1542 ?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Generated/Models/ComputerActionScreenshot.cs
```diff
@@ -0,0 +1,26 @@
...
+
+#nullable disable
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Agents.AI.Hosting.OpenAI.Responses.Generated.Models
+{
+    /// <summary> A screenshot action. </summary>
+    internal partial class ComputerActionScreenshot : ComputerAction
```
stephentoub: There are a ton of types being added, presumably all autogenerated from the typespec, but many of which we likely will never use, like this one. Do we need to add them all? Would it be possible to have them consolidated into fewer files, rather than adding literally 600 files?

ReubenBond:
> Does this replace #1542 ?

Yes

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Generated/Models/ComputerActionScreenshot.cs
ReubenBond: Yes, these are generated from the typespec. We almost certainly don't need them all. On the other hand, it's difficult to know which ones we need for certain and we'd need to maintain that manually unless we maintained a fork of the typespec that these were generated from. What do you suggest?

stephentoub: Could we group them into a single file at least?

ReubenBond: I've updated the PR to replace the generated models

dotnet/Directory.Packages.props
```diff
@@ -30,6 +30,7 @@
     <PackageVersion Include="System.Text.Json" Version="9.0.10" />
     <PackageVersion Include="System.Threading.Channels" Version="9.0.10" />
     <PackageVersion Include="System.Threading.Tasks.Extensions" Version="4.6.3" />
+    <PackageVersion Include="System.ClientModel" Version="1.7.0" />
```
stephentoub: What's this being used for now?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Microsoft.Agents.AI.Hosting.OpenAI.csproj
```diff
@@ -3,30 +3,29 @@
...
+    <PackageReference Include="Microsoft.Extensions.AI.Abstractions" />
   </ItemGroup>
-
+  
   <ItemGroup>
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
     <PackageReference Include="System.Net.ServerSentEvents" VersionOverride="10.0.0-rc.2.25502.107" />
     <PackageReference Include="Microsoft.Extensions.AI.OpenAI" />
+    <PackageReference Include="System.ClientModel" />
+    <PackageReference Include="System.Linq.Async" />
```
stephentoub: If we need LINQ for IAsyncEnumerable, let's depend on System.Linq.AsyncEnumerable instead of System.Linq.Async.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Common/AsyncEnumerableExtensions.cs
```diff
@@ -0,0 +1,151 @@
...
+        var pendingKey = keySelector(pending);
+        var hasPending = true;
+
+        while (hasPending)
+        {
+            var currentKey = pendingKey;
+
+            yield return InnerAsync(cancellationToken);
+
+            [SuppressMessage("ReSharper", "AccessToDisposedClosure")]
```
stephentoub: Can we remove ReSharper attributes?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Common/AsyncEnumerableExtensions.cs
```diff
@@ -0,0 +1,151 @@
...
+    /// <returns>Chunked async enumerable by key.</returns>
+    public static async IAsyncEnumerable<IAsyncEnumerable<TSource>> ChunkByKeyAsync<TSource, TKey>(
+        this IAsyncEnumerable<TSource> source,
+        Func<TSource, TKey> keySelector,
+        IEqualityComparer<TKey>? comparer = null,
+        [EnumeratorCancellation] CancellationToken cancellationToken = default)
+    {
+        comparer ??= EqualityComparer<TKey>.Default;
+
+        var e = source.GetAsyncEnumerator(cancellationToken);
```
stephentoub: If this were a general purpose routine, we'd want to make sure this was disposed. If we know source will never be anything that needs disposal, we could get away without it, but in general it's still better to clean up.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Common/Id/DefaultIdGenerator.cs
```diff
@@ -0,0 +1,153 @@
...
+/// </summary>
+internal sealed partial class DefaultIdGenerator : IIdGenerator
+{
+    private readonly string _partitionId;
+
+#if NET9_0_OR_GREATER
+    [GeneratedRegex("^[A-Za-z0-9]*$")]
+    private static partial Regex WatermarkRegex();
+#else
+    private static readonly Regex s_watermarkRegex = new("^[A-Za-z0-9]*$", RegexOptions.Compiled);
```
stephentoub: Should this be + rather than * ? It may not matter, as it looks like we only use it on inputs already validated to be non-empty.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Common/Id/DefaultIdGenerator.cs
```diff
@@ -0,0 +1,153 @@
...
+        var entropy = "";
+        while (entropy.Length != stringLength)
+        {
+            var buffer = RandomNumberGenerator.GetBytes(stringLength);
+            var entropyBase64 = Convert.ToBase64String(buffer);
+            entropy = new string(entropyBase64.Where(char.IsLetterOrDigit).ToArray())[..stringLength];
+        }
+
+        return entropy;
+    }
```
stephentoub: Is this only used on .NET Core? If so, it can just use RandomNumberGenerator.GetString, e.g.
```csharp
private static string SecureEntropy(int stringLength) =>
    RandomNumberGenerator.GetString("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789", stringLength);
```

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Converters/AfItemResourceGenerator.cs
```diff
@@ -0,0 +1,306 @@
...
+    private bool IsChanged(AgentRunResponseUpdate? previous, AgentRunResponseUpdate? current) => previous != null && current != null && this.Changed(previous, current);
+
+    /// <summary>
+    /// Determines whether the message ID has changed between two agent run response updates.
+    /// </summary>
+    /// <param name="previous">The previous <see cref="AgentRunResponseUpdate"/> instance to compare.</param>
+    /// <param name="current">The current <see cref="AgentRunResponseUpdate"/> instance to compare.</param>
+    /// <returns><see langword="true"/> if the message ID of the current update differs from the previous update; otherwise, <see
+    /// langword="false"/>.</returns>
+    private bool Changed(AgentRunResponseUpdate previous, AgentRunResponseUpdate current)
```
stephentoub: In ToAgentRunResponse / ToChatResponse, this partitioning happens based on MessageId, AuthorName, and Role (we also avoid introducing boundaries when those transition from null to non-null). We should probably ensure the semantics stay in sync. I wonder if there's another API the core libs could expose that this could just use.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Converters/AfItemResourceGenerator.cs
```diff
@@ -0,0 +1,306 @@
...
+                {
+                    case UsageContent usageContent:
+                        if (this.NotifyOnUsageUpdate != null && usageContent.Details != null)
+                        {
+                            this.NotifyOnUsageUpdate(usageContent.Details.ToResponseUsage()!);
+                        }
+                        continue;
+                    case FunctionCallContent or FunctionResultContent or TextContent:
+                        yield return (update, content);
+                        break;
```
stephentoub: I'm not clear on the logic... why are these being special-cased?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Converters/AfItemResourceGenerator.cs
```diff
@@ -0,0 +1,306 @@
...
+            ItemId = itemId,
+            OutputIndex = groupSeq,
+            ContentIndex = 0,
+            Part = new ItemContentOutputText(string.Empty, [])
+        };
+
+        var text = new StringBuilder();
+        await foreach (var content in source.WithCancellation(cancellationToken).ConfigureAwait(false))
+        {
+            if (content is not TextContent textContent)
```
stephentoub: It looks like we're not surfacing other forms of content, like reasoning (TextReasoningContent)? We can of course add it later.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Converters/ResponseConverterExtensions.cs
```diff
@@ -0,0 +1,193 @@
...
+            Id = context.ResponseId,
+            CreatedAt = (agentRunResponse.CreatedAt ?? DateTimeOffset.UtcNow).ToUnixTimeSeconds(),
+            Model = request.Agent?.Name ?? request.Model ?? "unknown",
+            Status = ResponseStatus.Completed,
+            Agent = request.Agent?.ToAgentId(),
+            Conversation = new ConversationReference { Id = context.ConversationId },
+            Metadata = request.Metadata is IReadOnlyDictionary<string, string> metadata ? new Dictionary<string, string>(metadata) : null,
+            Output = output.ToList(),
+            Usage = agentRunResponse.Usage.ToResponseUsage(),
+            ParallelToolCalls = true,
```
stephentoub: Should this incorporate whatever parallel tool calls option can in request?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Converters/ResponseConverterExtensions.cs
```diff
@@ -0,0 +1,193 @@
...
+    /// <param name="functionCallContent">The function call content to convert.</param>
+    /// <param name="id">The ID to assign to the resource.</param>
+    /// <param name="jsonSerializerOptions">The JSON serializer options to use.</param>
+    /// <returns>A FunctionToolCallItemResource.</returns>
+    public static FunctionToolCallItemResource ToFunctionToolCallItemResource(
+        this FunctionCallContent functionCallContent,
+        string id,
+        JsonSerializerOptions jsonSerializerOptions)
+    {
+#pragma warning disable IL2026, IL3050 // JSON serialization requires dynamic access
```
stephentoub: We shouldn't be suppressing this. What's necessitating this? If it's the use of jsonSerialzerOptions, we should ensure the JSO has the right type info in it, and then L112 below can use the overload that takes a JsonTypeInfo can call jsonSerializer.Options.GetTypeInfo(typeof(...))

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Converters/ResponseConverterExtensions.cs
```diff
@@ -0,0 +1,193 @@
...
+            TotalTokens = (int)(usage.TotalTokenCount ?? 0)
+        };
+    }
+
+    /// <summary>
+    /// Converts AIContent to ItemContent.
+    /// </summary>
+    /// <param name="content">The AI content to convert.</param>
+    /// <returns>An ItemContent object, or null if the content cannot be converted.</returns>
+    public static ItemContent? ToItemContent(this AIContent content)
```
stephentoub: This is a lot less complete than ConvertToItemContent later on... why do we have two?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Invocation/AgentInvocationException.cs
```diff
@@ -0,0 +1,50 @@
...
+
+using System;
+using Microsoft.Agents.AI.Hosting.OpenAI.Responses.Models;
+
+namespace Microsoft.Agents.AI.Hosting.OpenAI.Responses.Invocation;
+
+/// <summary>
+/// Exception thrown when an agent invocation fails.
+/// </summary>
+public class AgentInvocationException : Exception
```
stephentoub: This is public because...?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Invocation/Stream/NestedResponseGenerator.cs
```diff
@@ -0,0 +1,150 @@
...
+            Id = this.ResponseId,
+            CreatedAt = this._createdAt.ToUnixTimeSeconds(),
+            Model = this.Request.Agent?.Name ?? this.Request.Model ?? "unknown",
+            Status = status,
+            Agent = this.Request.Agent?.ToAgentId(),
+            Conversation = new ConversationReference { Id = this.ConversationId },
+            Metadata = this.Request.Metadata != null ? new Dictionary<string, string>(this.Request.Metadata) : null,
+            Output = outputs?.ToList() ?? [],
+            Usage = this._latestUsage,
+            ParallelToolCalls = true,
```
stephentoub: Same question here as earlier

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Invocation/Stream/NestedResponseGenerator.cs
```diff
@@ -0,0 +1,150 @@
...
+        yield return new StreamingResponseCompleted { SequenceNumber = this.Seq.GetNext(), Response = this.CompletedResponse };
+    }
+
+    private Response ToResponse(ResponseStatus status = ResponseStatus.Completed, IEnumerable<ItemResource>? outputs = null)
+    {
+        return new Response
+        {
+            Id = this.ResponseId,
+            CreatedAt = this._createdAt.ToUnixTimeSeconds(),
+            Model = this.Request.Agent?.Name ?? this.Request.Model ?? "unknown",
```
stephentoub: "unknown" is the right default rather than just leaving it null?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/AgentRunResponseExtensions.cs
```diff
@@ -0,0 +1,193 @@
...
+            Model = request.Agent?.Name ?? request.Model ?? "unknown",
+            Status = ResponseStatus.Completed,
+            Agent = request.Agent?.ToAgentId(),
+            Conversation = new ConversationReference { Id = context.ConversationId },
+            Metadata = request.Metadata is IReadOnlyDictionary<string, string> metadata ? new Dictionary<string, string>(metadata) : null,
+            Output = output.ToList(),
+            Usage = agentRunResponse.Usage.ToResponseUsage(),
+            ParallelToolCalls = true,
+            Tools = []
+        };
```
stephentoub, line 79: Isn't more detail typically roundtripped from the request, settings like temperature and top_p and the like?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Converters/AfItemResourceGenerator.cs
ReubenBond: It's because they have their own sequences of events which they emit, i.e, streamed function calls are distinct from text and other content. I intend to simplify this code more to reduce the amount of indirection, so hopefully it becomes clearer. As it stands, this is copied from an internal repo and only slightly modified.

dotnet/Directory.Packages.props
ReubenBond: Microsoft.Agents.AI.OpenAI uses it, eg for StreamingUpdatePipelineResponse
Microsoft.Agents.AI.Hosting.OpenAI.UnitTests uses it for the places where we use OpenAIResponseClient (i.e, the official client)

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Microsoft.Agents.AI.Hosting.OpenAI.csproj
ReubenBond: Done, I ref'd v10.0 rc2

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Common/Id/DefaultIdGenerator.cs
ReubenBond: I'll change it to + here and below

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Common/Id/DefaultIdGenerator.cs
ReubenBond: Nice

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Converters/AfItemResourceGenerator.cs
ReubenBond: I will copy the logic directly and we can remove it later if we add a method to check for this form of equality, eg `AgentRunResponseUpdate.IsSameMessage(AgentRunResponseUpdate? first, AgentRunResponseUpdate? second)`

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Invocation/AgentInvocationException.cs
ReubenBond: Codebase analyzers complain otherwise. I could suppress it or throw a general-purpose exception.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Invocation/Stream/NestedResponseGenerator.cs
ReubenBond: Fixed, thanks

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/AgentRunResponseExtensions.cs
ReubenBond: You're right. I've fixed this to include more detail.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Common/Id/IIdGenerator.cs
```diff
@@ -0,0 +1,50 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+namespace Microsoft.Agents.AI.Hosting.OpenAI.Responses.Common.Id;
+
+/// <summary>
+/// Defines the interface for generating IDs.
+/// </summary>
+internal interface IIdGenerator
```
DeagleGross: should we move this under `Microsoft.Agents.AI.Hosting`? I guess it would be useful for other hosting projects like A2A e.g.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/HostApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+/// Extension methods for <see cref="IHostApplicationBuilder"/> to configure OpenAI Responses support.
+/// </summary>
+public static class HostApplicationBuilderExtensions
+{
+    /// <summary>
+    /// Adds support for exposing <see cref="AIAgent"/> instances via OpenAI Responses.
+    /// </summary>
+    /// <param name="builder">The <see cref="IHostApplicationBuilder"/> to configure.</param>
+    /// <returns>The <see cref="IHostApplicationBuilder"/> for method chaining.</returns>
+    public static IHostApplicationBuilder AddOpenAIResponses(this IHostApplicationBuilder builder)
```
DeagleGross, line 20: got a PR recently where people ask for extensions on `IServiceCollection` as well: https://github.com/microsoft/agent-framework/pull/1456

we may consider having such an extension here as well

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Microsoft.Agents.AI.Hosting.OpenAI.csproj
```diff
@@ -3,34 +3,42 @@
...
 
   <ItemGroup>
-    <FrameworkReference Include="Microsoft.AspNetCore.App" />
-    <PackageReference Include="Microsoft.VisualStudio.Threading.Analyzers">
-      <PrivateAssets>all</PrivateAssets>
-      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
-    </PackageReference>
+    <Compile Remove="Responses\Invocation\Stream\**" />
+    <EmbeddedResource Remove="Responses\Invocation\Stream\**" />
+    <None Remove="Responses\Invocation\Stream\**" />
```
DeagleGross: where is it generated? is it needed?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/AIAgentResponsesProcessor.cs
```diff
@@ -85,100 +54,20 @@ public Task ExecuteAsync(HttpContext httpContext)
             response.Headers.ContentEncoding = "identity";
             httpContext.Features.GetRequiredFeature<IHttpResponseBodyFeature>().DisableBuffering();
 
+            var chatMessages = createResponse.Input.GetInputMessages().Select(i => i.ToChatMessage()).ToList();
+            var events = agent.RunStreamingAsync(chatMessages, cancellationToken: cancellationToken)
+                .ToStreamingResponseAsync(createResponse, context, cancellationToken)
```
DeagleGross, line 51: out of curiousity, why this extension method lives in a separate class? it is used only here, so why not implement it in-place?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/AgentRunResponseUpdateExtensions.cs
```diff
@@ -0,0 +1,196 @@
...
+        this IAsyncEnumerable<AgentRunResponseUpdate> updates,
+        CreateResponse request,
+        AgentInvocationContext context,
+        [EnumeratorCancellation] CancellationToken cancellationToken = default)
+    {
+        var seq = SequenceNumberFactory.CreateDefault();
+        var createdAt = DateTimeOffset.UtcNow;
+        var latestUsage = ResponseUsage.Zero;
+        yield return new StreamingResponseCreated { SequenceNumber = seq.GetNext(), Response = CreateResponse(status: ResponseStatus.InProgress) };
+        yield return new StreamingResponseInProgress { SequenceNumber = seq.GetNext(), Response = CreateResponse(status: ResponseStatus.InProgress) };
```
DeagleGross: should we only return `created` and `inprogress` when we know we are starting return the response? What if the first item from enumerator fails, does then we will return `created` and `progress` anyway, which is probably not correct? I am not sure how to correctly work with OpenAI spec since there is no example

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Models/AgentId.cs
```diff
@@ -0,0 +1,87 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Agents.AI.Hosting.OpenAI.Responses.Models;
```
DeagleGross, line 5: All of models types were generated from typespec, right? Should we add those definitions to the repo, so that we are able to easily regenerate those?

DeagleGross (approved):

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Common/Id/IIdGenerator.cs
ReubenBond: Maybe, but I prefer we do that at the same time as using it in those other places.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Microsoft.Agents.AI.Hosting.OpenAI.csproj
ReubenBond: Ah, this looks like a mistake, and I don't know how this happened... I'll double-check

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/AIAgentResponsesProcessor.cs
ReubenBond: Just for separation of concerns, nothing major. I wanted this file to be HTTP focused and the extension method to be focused on the conversion logic

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/AgentRunResponseUpdateExtensions.cs
ReubenBond: This is following OpenAI's observed/documented behavior (eg, in the streaming tab of the platform docs for response creation)

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Models/AgentId.cs
ReubenBond: They're not directly generated from the type spec, I need to update the PR comment. Originally, they were, but that ended up adding 600 files which were mostly unused and which were not nice to work with (eg, emitting `BinaryData` for every union type, of which there are many)

DeagleGross: cool, its just up to us to decide how to go forward - for example we can explicitly document "only generate these models via typespec and here are the definitions" or we can go with do it once, and update once there are changes to the spec. I am fine with either

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/EndpointRouteBuilderExtensions.Responses.cs
```diff
@@ -1,23 +1,21 @@
...
-using Microsoft.AspNetCore.Builder;
-using Microsoft.AspNetCore.Http;
+using Microsoft.Agents.AI.Hosting.OpenAI.Responses.Models;
+using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.Extensions.DependencyInjection;
-using OpenAI.Responses;
 
-namespace Microsoft.Agents.AI.Hosting.OpenAI;
+namespace Microsoft.AspNetCore.Builder;
```
DeagleGross, line 14: should this still be `Microsoft.AspNetCore.Builder`? Why not `Microsoft.Agents.AI.Hosting.OpenAI;`

dotnet/samples/AgentWebChat/AgentWebChat.AgentHost/Program.cs
```diff
@@ -81,6 +81,7 @@ Once the user has deduced what type (knight or knave) both Alice and Bob are, te
 
 builder.AddSequentialWorkflow("science-sequential-workflow", [chemistryAgent, mathsAgent, literatureAgent]).AddAsAIAgent();
 builder.AddConcurrentWorkflow("science-concurrent-workflow", [chemistryAgent, mathsAgent, literatureAgent]).AddAsAIAgent();
+builder.AddOpenAIResponses();
```
DeagleGross, line 83: since you renamed namespace to `Microsoft.AspNetCore.Builder`, we can get rid of line 9;

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/EndpointRouteBuilderExtensions.Responses.cs
DeagleGross: discussed; it is correct now

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Converters/ResponseConverterExtensions.cs
ReubenBond: Fixed

.github/workflows/dotnet-build-and-test.yml
```diff
@@ -148,7 +156,15 @@ jobs:
...
-            dotnet test -f ${{ matrix.targetFramework }} -c ${{ matrix.configuration }} $project --no-build -v Normal --logger trx
+            # Query the project's target frameworks using MSBuild with the current configuration
+            target_frameworks=$(dotnet msbuild $project -getProperty:TargetFrameworks -p:Configuration=${{ matrix.configuration }} -nologo 2>/dev/null | tr -d '\r')
+            
+            # Check if the project supports the target framework
+            if [[ "$target_frameworks" == *"${{ matrix.targetFramework }}"* ]]; then
+              dotnet test -f ${{ matrix.targetFramework }} -c ${{ matrix.configuration }} $project --no-build -v Normal --logger trx
+            else
+              echo "Skipping $project - does not support target framework ${{ matrix.targetFramework }} (supports: $target_frameworks)"
+            fi
```
stephentoub, line 167: What problem are these changes addressing?

dotnet/Directory.Packages.props
```diff
@@ -30,6 +30,8 @@
...
+    <PackageVersion Include="System.Linq.AsyncEnumerable" Version="10.0.0-rc.2.25502.107" />
```
stephentoub: Nit: Can you add these in alphabetical order within the section?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/Utils/ChatCompletionsOptionsExtensions.cs
```diff
@@ -12,41 +12,41 @@ namespace Microsoft.Agents.AI.Hosting.OpenAI.ChatCompletions.Utils;
 [System.Diagnostics.CodeAnalysis.SuppressMessage("Design", "CA1065:Do not raise exceptions in unexpected locations", Justification = "Specifically for accessing hidden members")]
 internal static class ChatCompletionsOptionsExtensions
 {
-    private static readonly Func<ChatCompletionOptions, bool?> _getStreamNullable;
-    private static readonly Func<ChatCompletionOptions, IList<ChatMessage>> _getMessages;
+    private static readonly Func<ChatCompletionOptions, bool?> s_getStreamNullable;
+    private static readonly Func<ChatCompletionOptions, IList<ChatMessage>> s_getMessages;
```
stephentoub, line 16: Presumably we'll subsequently also switch chat completions to be on a similar plan and this will go away?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/EndpointRouteBuilderExtensions.Responses.cs
```diff
@@ -1,90 +1,94 @@
...
-        responsesPath ??= $"/{agentName}/v1/responses";
-        var responsesRouteGroup = endpoints.MapGroup(responsesPath);
-        MapResponses(responsesRouteGroup, agent);
-
-        // Will be included once we obtain the API to operate with thread (conversation).
-
-        // conversationsPath ??= $"/{agentName}/v1/conversations";
-        // var conversationsRouteGroup = endpoints.MapGroup(conversationsPath);
-        // MapConversations(conversationsRouteGroup, agent, loggerFactory);
+        responsesPath ??= $"/{agent.Name}/v1/responses";
```
stephentoub, line 45: In the client sample earlier you changed it from /agentName/v1 to just /v1:
https://github.com/microsoft/agent-framework/pull/1550/files#diff-46fc60d96b0c00a4f9ae16250d485b983307024fe59823f06a8859278ea94455R26
Are these supposed to match up?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/EndpointRouteBuilderExtensions.Responses.cs
```diff
@@ -1,90 +1,94 @@
...
-            var requestBinary = await BinaryData.FromStreamAsync(requestContext.Request.Body, cancellationToken).ConfigureAwait(false);
-
-            var responseOptions = new ResponseCreationOptions();
-            var responseOptionsJsonModel = responseOptions as IJsonModel<ResponseCreationOptions>;
-            Debug.Assert(responseOptionsJsonModel is not null);
-
-            responseOptions = responseOptionsJsonModel.Create(requestBinary, ModelReaderWriterOptions.Json);
-            if (responseOptions is null)
+            // DevUI uses the 'model' field to specify the agent name.
+            var agentName = createResponse.Agent?.Name ?? createResponse.Model;
```
stephentoub, line 77: There's an "Agent" field in CreateResponse but DevUI doesn't use it? Why? Or if model gets ignored in the presence of an agent, I'm curious why an additional agent field is necessary... why not always use model for that?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/AgentRunResponseExtensions.cs
```diff
@@ -0,0 +1,314 @@
...
+        var usage = agentRunResponse.Usage.ToResponseUsage();
+
+        // If reasoning is enabled but no reasoning tokens are reported, add placeholder reasoning tokens
+        // This ensures conformance with the API spec that reasoning tokens should be > 0 for reasoning models
+        if (request.Reasoning != null && usage.OutputTokensDetails.ReasoningTokens == 0)
+        {
+            // Add fake reasoning tokens for testing/conformance purposes
+            usage = usage with
+            {
+                OutputTokensDetails = usage.OutputTokensDetails with { ReasoningTokens = 128 }
```
stephentoub: Reasoning tokens are considered part of output tokens. Do we increase the output tokens count accordingly? Could the value be 1 instead of 128 to create less discrepancy?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/AgentRunResponseExtensions.cs
```diff
@@ -0,0 +1,314 @@
...
+            Tools = request.Tools?.Select(ProcessTool).ToList() ?? [],
+            ToolChoice = request.ToolChoice,
+            ServiceTier = request.ServiceTier ?? "default",
+            Store = request.Store ?? true,
+            PreviousResponseId = request.PreviousResponseId,
+            Reasoning = request.Reasoning,
+            Text = request.Text,
+            MaxOutputTokens = request.MaxOutputTokens,
+            Truncation = request.Truncation,
+#pragma warning disable CS0618 // Type or member is obsolete
```
stephentoub, line 69: Interesting... why did we mark these as [Obsolete]... I guess OpenAI cites them as such in their docs? Do we need to send them back at all?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/AgentRunResponseExtensions.cs
```diff
@@ -0,0 +1,314 @@
...
+                }
+                else if (property.Name != "strict")
+                {
+                    // Copy other properties except strict (we'll add it explicitly)
+                    property.WriteTo(writer);
+                }
+            }
+
+            // Add strict flag
+            writer.WriteBoolean("strict", true);
```
stephentoub: Why? strict implies much tighter constraints than many tools are able to handle. It looks like we're adding this to all tool information we roundtrip?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/AgentRunResponseExtensions.cs
```diff
@@ -0,0 +1,314 @@
...
+
+            writer.WriteEndObject();
+        }
+
+        return JsonDocument.Parse(stream.ToArray()).RootElement.Clone();
```
stephentoub:
```suggestion
        using var doc = JsonDocument.Parse(stream.ToArray());
        return doc.RootElement.Clone();
```

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/AgentRunResponseExtensions.cs
```diff
@@ -0,0 +1,314 @@
...
+                writer.WriteStringValue(prop.Name);
+            }
+            writer.WriteEndArray();
+        }
+
+        // Add additionalProperties flag
+        writer.WriteBoolean("additionalProperties", false);
+
+        writer.WriteEndObject();
+    }
```
stephentoub: There are helpers for this kind of thing in Microsoft.Extensions.AI. I would hope those could be used instead of open-coding this. @eiriktsarpalis, can you take a look?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/AgentRunResponseExtensions.cs
```diff
@@ -0,0 +1,314 @@
...
+                    break;
+                case FunctionResultContent functionResultContent:
+                    // message.Role == ChatRole.Tool
+                    yield return functionResultContent.ToFunctionToolCallOutputItemResource(
+                        idGenerator.GenerateFunctionOutputId());
+                    break;
+                default:
+                    // message.Role == ChatRole.Assistant
+                    var itemContent = ItemContentConverter.ToItemContent(content);
+                    if (itemContent != null)
```
stephentoub: Nit:
```suggestion
                    if (ItemContentConverter.ToItemContent(content) is { } itemContent)
```

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Converters/ItemContentConverter.cs
```diff
@@ -0,0 +1,163 @@
...
+                new ItemContentInputFile
+                {
+                    FileId = hostedFile.FileId
+                },
+            DataContent fileData when !fileData.HasTopLevelMediaType("image") && !fileData.HasTopLevelMediaType("audio") =>
+                new ItemContentInputFile
+                {
+                    FileData = fileData.Uri,
+                    Filename = fileData.Name
+                },
```
stephentoub, line 123: Nit: if you moved this down, you could drop the condition

.github/workflows/dotnet-build-and-test.yml
ReubenBond: The newly added test project does not support net472 and because we enumerate projects into a variable and run them one-by-one, we need to exclude that project when running the net472 job

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/Utils/ChatCompletionsOptionsExtensions.cs
ReubenBond: Yes, hopefully. cc @DeagleGross

ReubenBond: I was just fixing linter warnings here. I'm not sure how they passed the linter before. Maybe the linter rule changed, but either way they were flagged for me.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/EndpointRouteBuilderExtensions.Responses.cs
ReubenBond: I changed the client to use the other endpoint, which stuffs the agent name into the model parameter, since that's what DevUI does and it's what Foundry expects

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/EndpointRouteBuilderExtensions.Responses.cs
ReubenBond: The 'Agent' prop is from Foundry's extension. We should probably upgrade DevUI to use it. cc @victordibia

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/AgentRunResponseExtensions.cs
ReubenBond: Yes, that's right: OpenAI flagged them as obsolete, but they still send them back in the response.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/AgentRunResponseExtensions.cs
ReubenBond: This is the default that OpenAI sets, but I'll see if I can pass through the input as-is and remove the check from conformance tests. I'll consult the typespec & OAI reference docs

dotnet/Directory.Packages.props
ReubenBond: Done. That also involved rearranging some of the other lines, but now the section is ordered

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/AgentRunResponseExtensions.cs
ReubenBond: I removed this for now. We don't need to emit these properties necessarily, at least not yet, so we can just pass-thru input through to the output.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/EndpointRouteBuilderExtensions.Responses.cs
victordibia: Thanks for tagging here @ReubenBond .
I dont see an agent field in the official oai [response obj](https://platform.openai.com/docs/api-reference/responses/object) (except I have missed soemthing); is it correct tot say this is a foundry specific field? i.e. what do we expect for clients that use a model field (and want to swap out the baseurl here for e.g., other responses api base urls?) . Perhaps DevUI has a fallback design where it also checks for an agent field? 
Happy to hear thoughts

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/AgentRunResponseExtensions.cs
ReubenBond: I removed this - they were just there to make sure the output had the right shape for tests... but really we shouldn't fabricate fields just like this, so I changed the test to not assert on this.

PR #1551, opened 2025-10-17
.NET: [BREAKING] Remove generic Workflow<T>

lokitoth:
### Motivation and Context

The current Workflow APIs support Workflow<T> to describe a Workflow taking in messages of type T as input. This is inelegant for workflows supporting multiple different types of input, especially ones that require multiple types of input in concert, such as ChatProtocol workflows, as can be hosted in a `WorkflowHostAgent` via `WorkflowHostingExtensions.AsAgent()`

Due to late-instantiation of Executors, and the need to allow async instantiation, type checking a workflow is an async operation, resulting in `WorkflowBuilder.BuildAsync<T>()`, `WorkflowHostingExtensions.AsAgentAsync()`, and `Workflow.TryPromoteAsync<T>()`. These are very difficult to use in DI scenarios, which do not support async initialization, resulting in async-over-sync / `.RunSync()` patterns.

### Description

The fix is to delay the type checking to a context where async is present, e.g. workflow run time.

Remove the input type checking in favour of explicit `.DescribeProtocolAsync()` flow. Also removes `.AsAgentAsync()` as the validation happens at workflow run time. This makes it easier to use Workflows with DI without resorting to async-over-sync.

Breaking Changes:
* Remove `WorkflowBuilder.BuildAsync`
* Remove `WorkflowHostingExtensions.AsAgentAsync`
* Remove `Workflow<T>`
* Remove `Workflow.TryPromoteAsync<T>`
* Change `WorkflowHostingExtensions.AsAgent` to require a `Workflow` rather than `Workflow<T>`

The mechanism of type checking of a workflow is now handled via `Workflow.DescribeProtocolAsync()`

### Contribution Checklist

### DIFF
--- a/dotnet/samples/GettingStarted/Workflows/Checkpoint/CheckpointAndResume/WorkflowHelper.cs
+++ b/dotnet/samples/GettingStarted/Workflows/Checkpoint/CheckpointAndResume/WorkflowHelper.cs
@@ -13,7 +13,7 @@ internal static class WorkflowHelper
     /// 2. JudgeExecutor: Evaluates the guess and provides feedback.
     /// The workflow continues until the correct number is guessed.
     /// </summary>
-    internal static ValueTask<Workflow<NumberSignal>> GetWorkflowAsync()
+    internal static Workflow GetWorkflow()
     {
         // Create the executors
         GuessNumberExecutor guessNumberExecutor = new(1, 100);
@@ -24,7 +24,7 @@ internal static class WorkflowHelper
             .AddEdge(guessNumberExecutor, judgeExecutor)
             .AddEdge(judgeExecutor, guessNumberExecutor)
             .WithOutputFrom(judgeExecutor)
-            .BuildAsync<NumberSignal>();
+            .Build();
     }
 }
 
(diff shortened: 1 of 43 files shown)


stephentoub (approved):

dotnet/samples/GettingStarted/Workflows/Checkpoint/CheckpointAndResume/WorkflowHelper.cs
crickman, line 7: nit - `WorkflowProvider` - all (i've been trained to avoid usage of the terms: "Helper", "Manager", and "Utilities").  Reason: There's always a more precise term that desribes the component function.

alliscode (approved):

dotnet/samples/GettingStarted/Workflows/Checkpoint/CheckpointAndResume/WorkflowHelper.cs
lokitoth: Makes sense; I made an issue to track it (there are a few around naming of classes in Workflows, and I do not want to make this PR hang out longer)

PR #1581, opened 2025-10-20
.NET: dotnet samples suggestions

jozkee:
* Remove named arguments where not needed
* Workflows samples code suggestions.

@TaoChenOSU @westey-m

### DIFF
--- a/dotnet/samples/GettingStarted/Workflows/HumanInTheLoop/HumanInTheLoopBasic/Program.cs
+++ b/dotnet/samples/GettingStarted/Workflows/HumanInTheLoop/HumanInTheLoopBasic/Program.cs
@@ -48,23 +48,20 @@ public static class Program
 
     private static ExternalResponse HandleExternalRequest(ExternalRequest request)
     {
-        if (request.DataIs<NumberSignal>())
+        switch (request.DataAs<NumberSignal?>())
         {
-            switch (request.DataAs<NumberSignal>())
-            {
-                case NumberSignal.Init:
-                    int initialGuess = ReadIntegerFromConsole("Please provide your initial guess: ");
-                    return request.CreateResponse(initialGuess);
-                case NumberSignal.Above:
-                    int lowerGuess = ReadIntegerFromConsole("You previously guessed too large. Please provide a new guess: ");
-                    return request.CreateResponse(lowerGuess);
-                case NumberSignal.Below:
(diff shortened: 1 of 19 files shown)


dotnet/samples/GettingStarted/Workflows/HumanInTheLoop/HumanInTheLoopBasic/Program.cs
```diff
@@ -48,23 +48,20 @@ private static async Task Main()
 
     private static ExternalResponse HandleExternalRequest(ExternalRequest request)
     {
-        if (request.DataIs<NumberSignal>())
+        switch (request.DataAs<NumberSignal?>())
         {
```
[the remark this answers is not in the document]
jozkee: If its null it means the Data wasn't `NumberSignal`.

dotnet/samples/GettingStarted/Workflows/_Foundational/03_AgentsInWorkflows/Program.cs
```diff
@@ -28,7 +28,7 @@ public static class Program
     private static async Task Main()
     {
         // Set up the Azure OpenAI client
-        var endpoint = Environment.GetEnvironmentVariable("AZURE_OPENAI_ENDPOINT") ?? throw new InvalidOperationException("AZURE_OPENAI_ENDPOINT is not set.");
+        var endpoint = Environment.GetEnvironmentVariable("AZURE_OPENAI_ENDPOINT") ?? throw new Exception("AZURE_OPENAI_ENDPOINT is not set.");
```
TaoChenOSU, line 31: What is the reason to use a less concrete exception?

jozkee: The change is about not throwing InvalidOperationException since those are meant for instance method calls and convey that the object's state is invalid.

https://learn.microsoft.com/dotnet/api/system.invalidoperationexception
> The exception that is thrown when a method call is invalid for the object's current state.

TaoChenOSU: Then perhaps another exception type? For example: "ArgumentException".

jozkee:
This is stale now and is not adding significant value, closing.

PR #1600, opened 2025-10-21
.NET: Use extension methods from A2A package for converting between MEAI & A2A model classes

SergeyMenshykh:
This PR updates AF to the latest A2A packages and replaces extension methods for converting between MEAI and A2A model classes with those from the A2A package.


dotnet/tests/Microsoft.Agents.AI.A2A.UnitTests/A2AAgentTests.cs
```diff
@@ -371,7 +371,7 @@ public async Task RunAsync_WithHostedFileContent_ConvertsToFilePartAsync()
             new(ChatRole.User,
             [
                 new TextContent("Check this file:"),
-                new HostedFileContent("https://example.com/file.pdf")
+                new UriContent("https://example.com/file.pdf", "file/pdf")
```
stephentoub:
```suggestion
                new UriContent("https://example.com/file.pdf", "application/pdf")
```

PR #1612, opened 2025-10-21
.NET: Add workflow as an agent with observability sample

TaoChenOSU:
### Motivation and Context

We need a sample to show how to enable observability of workflow agents (a workflow used as an AI agent).

Also closing #1814 

### Description

Add a sample to show how to enable observability for workflow as an agent. The sample enables telemetry at multiple levels:
1. The chat client level
2. The agent level for agents inside the workflow
3. The workflow level (implicit)
4. The workflow agent level

image

This sample is known to fail after more than one turn. The issue is tracked here: https://github.com/microsoft/agent-framework/issues/1613

### Contribution Checklist

### DIFF
--- a/dotnet/samples/GettingStarted/Workflows/Observability/WorkflowAsAnAgent/WorkflowHelper.cs
+++ b/dotnet/samples/GettingStarted/Workflows/Observability/WorkflowAsAnAgent/WorkflowHelper.cs
@@ -73,7 +73,7 @@ internal static class WorkflowHelper
     /// <summary>
     /// Executor that aggregates the results from the concurrent agents.
     /// </summary>
-    private sealed class ConcurrentAggregationExecutor() : Executor<ChatMessage>("ConcurrentAggregationExecutor")
+    private sealed class ConcurrentAggregationExecutor() : Executor<List<ChatMessage>>("ConcurrentAggregationExecutor")
     {
         private readonly List<ChatMessage> _messages = [];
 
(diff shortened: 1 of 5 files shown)


ekzhu (approved):

dotnet/samples/GettingStarted/Workflows/Observability/WorkflowAsAnAgent/WorkflowHelper.cs
```diff
@@ -0,0 +1,98 @@
...
+        // Create executors
+        var startExecutor = new ConcurrentStartExecutor();
+        var aggregationExecutor = new ConcurrentAggregationExecutor();
+        AIAgent frenchAgent = GetLanguageAgent("French", chatClient, sourceName);
+        AIAgent englishAgent = GetLanguageAgent("English", chatClient, sourceName);
+
+        // Build the workflow by adding executors and connecting them
+        return new WorkflowBuilder(startExecutor)
+            .AddFanOutEdge(startExecutor, targets: [frenchAgent, englishAgent])
+            .AddFanInEdge(aggregationExecutor, sources: [frenchAgent, englishAgent])
```
lokitoth, line 28: This will not work because FanIn does not work well with ChatProtocol. Working on a PR to fix it.

lokitoth: Actually there's an easier short-term fix in this case: Update the `ConcurrentAggregationExecutor` to be `Executor<List<ChatMessage>>` rather than `Executor<ChatMessage>`.

No need to wait on the pending PR.

Issue #1454, opened 2025-10-14
.NET: Simplify Agent Registration & A2A Mapping (Library-Friendly Additive API Extensions)


jst4upiyush:
## **Problem**

Current onboarding flows (“just give me an agent with name + instructions”) often push users into `IHostApplicationBuilder` patterns.
However, library authors — the primary audience here — typically depend only on `Microsoft.Extensions.*` primitives and avoid adding hosting/web abstractions.
Many reusable frameworks extend only `IServiceCollection`, so the lack of direct overloads forces extra ceremony and discourages incremental adoption in library contexts.

## **Proposed (Implemented in Branch)**

This PR adds new **`IServiceCollection` overloads and `IEndpointRouteBuilder` helpers**, all purely additive and hosting-agnostic:

* **`IServiceCollection.AddAIAgent(name, instructions)`**
* **`IServiceCollection.AddAIAgent(name, instructions, IChatClient chatClient)`**
* **`IServiceCollection.AddAIAgent(name, instructions, object? chatClientServiceKey)`**
* **`IServiceCollection.AddAIAgent(name, instructions, string? description, object? chatClientServiceKey)`**
* Mirror overloads on **`IHostApplicationBuilder`** (for backward compatibility) that forward internally to the new service-collection methods.
* New **`IEndpointRouteBuilder`** A2A helpers: `MapA2A(agentName, path)` (optionally with `AgentCard` or `TaskManager`), enabling modular route-group usage without requiring `WebApplication`.

## **Motivation**

These changes make agent registration **library-first and dependency-light**, aligning with the `IServiceCollection` DI pattern already pervasive across .NET. They enable earlier registration (pre-host build), simplify library packaging, and allow modular A2A mapping within grouped routes — all without requiring `WebApplication`.

## **Backward Compatibility & Scope**

Fully backward-compatible: existing `IHostApplicationBuilder` extensions remain, now forwarding internally to the new `IServiceCollection` versions.
No removals, no behavior changes — this is a **non-breaking additive API** that can be safely adopted incrementally.
All builds and tests pass locally.

## **Checklist**

- [x] Builds succeed locally
- [x] All existing tests pass
- [x] No public API removals
- [x] Added XML docs for all new public methods
- [x] Follows repo coding conventions
- [x] No licensing / header changes
- [x] No breaking changes introduced

## **Request for Feedback**

1. Are the new additive overloads acceptable for library-first use cases?
2. Should we deprecate host-builder variants in a future minor?
[...]

PR #1625, opened 2025-10-22
.NET: Refactor A2A and AIAgent hosting extensions

DeagleGross:
PR includes several things:
1) gets rid of existing `A2AHostAgent` from `Microsoft.Agents.AI.A2A` package. The idea is to use `Microsoft.Agents.AI.Hosting.A2A` package instead for hosting the `AIAgent` via A2A protocol.
2) introduces more granular API for `MapA2A()`: current `A2AServer` sample uses instance of `AIAgent` to register, so I added such API:
```diff
namespace Microsoft.AspNetCore.Builder;

public static class MicrosoftAgentAIHostingA2AEndpointRouteBuilderExtensions
{
+ public static ITaskManager MapA2A(this IEndpointRouteBuilder endpoints, AIAgent agent, string path)
+ public static ITaskManager MapA2A(this IEndpointRouteBuilder endpoints, AIAgent agent, string path, AgentCard agentCard)
}
```
`MicrosoftAgentAIHostingA2AEndpointRouteBuilderExtensions` was also:
- changed to be an extension on top of `IEndpointRouteBuilder`, not `WebApplicationBuilder`
- moved under `Microsoft.AspNetCore.Builder` namespace as per convention (as done in this PR as well: https://github.com/microsoft/agent-framework/pull/1550#discussion_r2449628493)
- return type changed to `ITaskManager` which allows further setup. Sample uses it then to configure well-known agent-card endpoint like `app.MapWellKnownAgentCard(taskManager, "/");`

3) #1454 and #1456 asked for change to `IServiceCollection` for agent registration as well, so since I am refactoring here, I decided to include this change as well.

### Contribution Checklist

### DIFF
--- a/dotnet/Directory.Packages.props
+++ b/dotnet/Directory.Packages.props
@@ -38,7 +38,7 @@
     <PackageVersion Include="OpenTelemetry" Version="1.13.1" />
     <PackageVersion Include="OpenTelemetry.Api" Version="1.13.1" />
     <PackageVersion Include="OpenTelemetry.Exporter.Console" Version="1.13.1" />
-    <PackageVersion Include="OpenTelemetry.Exporter.InMemory" Version="1.12.0" />
+    <PackageVersion Include="OpenTelemetry.Exporter.InMemory" Version="1.13.1" />
     <PackageVersion Include="OpenTelemetry.Exporter.OpenTelemetryProtocol" Version="1.13.1" />
     <PackageVersion Include="OpenTelemetry.Extensions.Hosting" Version="1.12.0" />
     <PackageVersion Include="OpenTelemetry.Instrumentation.AspNetCore" Version="1.12.0" />
@@ -94,7 +94,7 @@
     <!-- Community -->
     <PackageVersion Include="System.Linq.Async" Version="6.0.3" />
     <!-- Test -->
-    <PackageVersion Include="FluentAssertions" Version="8.7.1" />
+    <PackageVersion Include="FluentAssertions" Version="8.8.0" />
     <PackageVersion Include="Microsoft.AspNetCore.Mvc.Testing" Version="9.0.10" />
     <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="18.0.0" />
     <PackageVersion Include="Moq" Version="[4.18.4]" />


dotnet/src/Microsoft.Agents.AI.Hosting/HostApplicationBuilderAgentExtensions.cs
```diff
@@ -99,44 +86,6 @@ public static IHostedAgentBuilder AddAIAgent(this IHostApplicationBuilder builde
     public static IHostedAgentBuilder AddAIAgent(this IHostApplicationBuilder builder, string name, Func<IServiceProvider, string, AIAgent> createAgentDelegate)
     {
         Throw.IfNull(builder);
```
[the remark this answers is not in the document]
DeagleGross: validated in inner method

dotnet/src/Microsoft.Agents.AI.Hosting/AgentHostingServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,145 @@
...
+using Microsoft.Extensions.AI;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Agents.AI.Hosting;
+
+/// <summary>
+/// Provides extension methods for configuring AI agents in a service collection.
+/// </summary>
+public static class AgentHostingServiceCollectionExtensions
```
crickman, line 15: Is any of this unit-testable?

dotnet/src/Microsoft.Agents.AI.Hosting.A2A.AspNetCore/EndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,93 @@
...
+public static class MicrosoftAgentAIHostingA2AEndpointRouteBuilderExtensions
+{
+    /// <summary>
+    /// Attaches A2A (Agent2Agent) communication capabilities via Message processing to the specified web application.
+    /// </summary>
+    /// <param name="endpoints">The <see cref="IEndpointRouteBuilder"/> to add the A2A endpoints to.</param>
+    /// <param name="agentName">The name of the agent to use for A2A protocol integration.</param>
+    /// <param name="path">The route group to use for A2A endpoints.</param>
+    /// <returns>Configured <see cref="ITaskManager"/> for A2A integration.</returns>
+    public static ITaskManager MapA2A(this IEndpointRouteBuilder endpoints, string agentName, string path)
```
markwallace-microsoft, line 26: If hosting multiple A2A agents at the same endpoint but on different paths allowed?

We got some feedback from the Darrel that we should not allow this but things may have changed.

dotnet/src/Microsoft.Agents.AI.Hosting/AgentHostingServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,145 @@
...
+{
+    /// <summary>
+    /// Adds an AI agent to the service collection using only a name and instructions, resolving the chat client from dependency injection.
+    /// </summary>
+    /// <param name="services">The service collection to configure.</param>
+    /// <param name="name">The name of the agent.</param>
+    /// <param name="instructions">The instructions for the agent.</param>
+    /// <returns>The same <see cref="IServiceCollection"/> instance so that additional calls can be chained.</returns>
+    /// <exception cref="ArgumentNullException">Thrown when <paramref name="services"/> or <paramref name="name"/> is <see langword="null"/>.</exception>
+    public static IHostedAgentBuilder AddAIAgent(this IServiceCollection services, string name, string? instructions)
```
markwallace-microsoft, line 25: This method always creates a `ChatClientAgent` so should it be named `AddChatClientAgent`?

We should have an overload that takes `ChatClientAgentOptions` so that you can add tools, context providers, ...

westey-m: +1,
Also compare with https://github.com/microsoft/agent-framework/blob/main/dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientExtensions.cs.
All the methods for creating agents elsewhere typically follow a common pattern where we support two signatures:
1. string? instructions = null, string? name = null, string? description = null, IList<AITool>? tools = null,
2. ChatClientAgentOptions? options

westey-m, line 25: Should these helpers take an optional `ServiceLifetime lifetime = ServiceLifetime.Singleton`?
For cases where the agent is relying on tools or anything else that requires obo tokens for downstream auth, these may need to be recreated for each request and users may want to choose transient.

DeagleGross: I have already created an issue to suport tools creation and I can include `ChatClientAgentOptions` overload as well. Let me do in separate PR please to not put different changes in one PR: https://github.com/microsoft/agent-framework/issues/1634

DeagleGross: same as here: https://github.com/microsoft/agent-framework/pull/1520#discussion_r2454683736

I would say we can try to design it, but requires more changes. I will try to do it, but lets do in a separate PR please

dotnet/src/Microsoft.Agents.AI.Hosting/AgentHostingServiceCollectionExtensions.cs
DeagleGross: added

dotnet/src/Microsoft.Agents.AI.Hosting.A2A.AspNetCore/EndpointRouteBuilderExtensions.cs
DeagleGross: We are not tracking on which paths agents were registered. I think it is a responsibility of user to do it and we should not intervene.

I think we can think about including a set of Rolsyn analyzers to point user to a possible conflict, like is done in aspnetcore:
image

dotnet/src/Microsoft.Agents.AI.Hosting.A2A.AspNetCore/EndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,93 @@
...
+
+    /// <summary>
+    /// Attaches A2A (Agent2Agent) communication capabilities via Message processing to the specified web application.
+    /// </summary>
+    /// <param name="endpoints">The <see cref="IEndpointRouteBuilder"/> to add the A2A endpoints to.</param>
+    /// <param name="agentName">The name of the agent to use for A2A protocol integration.</param>
+    /// <param name="path">The route group to use for A2A endpoints.</param>
+    /// <param name="agentCard">Agent card info to return on query.</param>
+    /// <returns>Configured <see cref="ITaskManager"/> for A2A integration.</returns>
+    public static ITaskManager MapA2A(this IEndpointRouteBuilder endpoints, string agentName, string path, AgentCard agentCard)
```
SergeyMenshykh, line 45: It would be easier for consumers to know which extension method to use if we could associate each extension method with the A2A discovery mechanism from the spec (https://github.com/a2aproject/A2A/blob/main/docs/topics/agent-discovery.md) by putting the details, maybe with the link, into the remarks section of each method.

westey-m (approved):

PR #1630, opened 2025-10-22
.NET: Add Rag AIContext Provider

westey-m:
### Motivation and Context

#1594

### Description

- Add RAG AI Context Provider with unit tests

### Contribution Checklist

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI/Data/TextSearchProvider.cs
+++ b/dotnet/src/Microsoft.Agents.AI/Data/TextSearchProvider.cs
@@ -14,50 +14,51 @@ using Microsoft.Shared.Diagnostics;
 namespace Microsoft.Agents.AI.Data;
 
 /// <summary>
-/// A Retrieval Augmented Generation (RAG) context provider that performs a search over external knowledge
+/// A text search context provider that performs a search over external knowledge
 /// and injects the formatted results into the AI invocation context, or exposes a search tool for on-demand use.
+/// This provider can be used to enable Retrieval Augmented Generation (RAG) on an agent.
 /// </summary>
 /// <remarks>
 /// <para>
-/// The provider supports two behaviors controlled via <see cref="RagProviderOptions.SearchTime"/>:
+/// The provider supports two behaviors controlled via <see cref="TextSearchProviderOptions.SearchTime"/>:
 /// <list type="bullet">
-/// <item><description><see cref="RagProviderOptions.RagBehavior.BeforeAIInvoke"/> – Automatically performs a search prior to every AI invocation and injects results as additional instructions.</description></item>
-/// <item><description><see cref="RagProviderOptions.RagBehavior.OnDemandFunctionCalling"/> – Exposes a function tool that the model may invoke to retrieve contextual information when needed.</description></item>
+/// <item><description><see cref="TextSearchProviderOptions.TextSearchBehavior.BeforeAIInvoke"/> – Automatically performs a search prior to every AI invocation and injects results as additional instructions.</description></item>
+/// <item><description><see cref="TextSearchProviderOptions.TextSearchBehavior.OnDemandFunctionCalling"/> – Exposes a function tool that the model may invoke to retrieve contextual information when needed.</description></item>
 /// </list>
 /// </para>
 /// <para>
-/// When <see cref="RagProviderOptions.RecentMessageMemoryLimit"/> is greater than zero the provider will retain the most recent
+/// When <see cref="TextSearchProviderOptions.RecentMessageMemoryLimit"/> is greater than zero the provider will retain the most recent
 /// user and assistant messages (up to the configured limit) across invocations and prepend them (in chronological order)
 /// to the current request messages when forming the search input. This can improve search relevance by providing
 /// multi-turn context to the retrieval layer without permanently altering the conversation history.
 /// </para>
 /// </remarks>
-public sealed class RagProvider : AIContextProvider
+public sealed class TextSearchProvider : AIContextProvider
 {
     private const string DefaultPluginSearchFunctionName = "Search";
     private const string DefaultPluginSearchFunctionDescription = "Allows searching for additional information to help answer the user question.";
     private const string DefaultContextPrompt = "## Additional Context\nConsider the following information from source documents when responding to the user:";
     private const string DefaultCitationsPrompt = "Include citations to the source document with document name and link if document name and link is available.";
 
-    private readonly Func<string, CancellationToken, Task<IEnumerable<RagSearchResult>>> _searchAsync;
-    private readonly ILogger<RagProvider>? _logger;
+    private readonly Func<string, CancellationToken, Task<IEnumerable<TextSearchSearchResult>>> _searchAsync;
+    private readonly ILogger<TextSearchProvider>? _logger;
     private readonly AITool[] _tools;
     private readonly Queue<string> _recentMessagesText;
-    private readonly RagProviderOptions _options;
+    private readonly TextSearchProviderOptions _options;
 
     /// <summary>
-    /// Initializes a new instance of the <see cref="RagProvider"/> class.
+    /// Initializes a new instance of the <see cref="TextSearchProvider"/> class.
     /// </summary>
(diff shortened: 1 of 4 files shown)


dotnet/src/Microsoft.Agents.AI/Rag/RagProviderOptions.cs
```diff
@@ -0,0 +1,77 @@
...
+    /// Gets or sets a value indicating when the search should be executed.
+    /// </summary>
+    /// <value><see cref="RagBehavior.BeforeAIInvoke"/> by default.</value>
+    public RagBehavior SearchTime { get; set; } = RagBehavior.BeforeAIInvoke;
+
+    /// <summary>
+    /// Gets or sets the name of the exposed search tool when operating in on-demand mode.
+    /// </summary>
+    /// <value>Defaults to "Search".</value>
+    public string? PluginFunctionName { get; set; }
```
SergeyMenshykh: Should it not be `ToolName` and `ToolDescription` for the property below?

dotnet/src/Microsoft.Agents.AI/Rag/RagProviderOptions.cs
```diff
@@ -0,0 +1,77 @@
...
+
+    /// <summary>
+    /// Gets or sets the context prompt prefixed to automatically injected results.
+    /// </summary>
+    public string? ContextPrompt { get; set; }
+
+    /// <summary>
+    /// Gets or sets the instruction appended after automatically injected results to request citations.
+    /// </summary>
+    public string? IncludeCitationsPrompt { get; set; }
```
SergeyMenshykh: There's an inconsistency in the `IncludeCitationsPrompt` and `ContextPrompt` property names. Both are used to format results, but one has the `Include` prefix while the other does not. I suggest aligning the names unless the prefix has a special meaning that I did not catch.

SergeyMenshykh (approved):

dotnet/src/Microsoft.Agents.AI/Rag/RagProviderOptions.cs
westey-m: Good point, I've updated to FunctionToolName and FunctionToolDescription.

dotnet/src/Microsoft.Agents.AI/Rag/RagProviderOptions.cs
westey-m: The prompt is to instruct the LLM to include citations, hence the name, but it can be confusing.
I've changed it to CitationsPrompt, since I think it's good enough.

dotnet/src/Microsoft.Agents.AI/Data/TextSearchProvider.cs
```diff
@@ -0,0 +1,300 @@
...
+using System.Linq;
+using System.Text;
+using System.Text.Json;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.AI;
+using Microsoft.Extensions.Logging;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Agents.AI.Data;
```
rogerbarreto, line 14: Not sure why are we hiding behind the `Data` namespace, normally we expose all the public surface under `Agents.AI`
```suggestion
namespace Microsoft.Agents.AI;
```

dotnet/tests/Microsoft.Agents.AI.UnitTests/Data/TextSearchProviderTests.cs
```diff
@@ -0,0 +1,503 @@
...
+using System.Linq;
+using System.Text.Json;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Agents.AI.Data;
+using Microsoft.Extensions.AI;
+using Microsoft.Extensions.Logging;
+using Moq;
+
+namespace Microsoft.Agents.AI.UnitTests.Data;
```
rogerbarreto, line 14: Suggest renaming the folder as well, should end with `UnitTests`
```suggestion
namespace Microsoft.Agents.AI.Data.UnitTests;
```

dotnet/src/Microsoft.Agents.AI/Data/RagProvider.cs
```diff
@@ -0,0 +1,300 @@
...
+        }
+        sb.AppendLine(this._options.CitationsPrompt ?? DefaultCitationsPrompt);
+        sb.AppendLine();
+        return sb.ToString();
+    }
+
+    /// <summary>
+    /// Represents a single retrieved knowledge result for RAG enrichment.
+    /// </summary>
+    public sealed class RagSearchResult
```
markwallace-microsoft: Should this be in it's own file?

dotnet/src/Microsoft.Agents.AI/Data/TextSearchProvider.cs
```diff
@@ -0,0 +1,300 @@
...
+        /// <summary>
+        /// Gets or sets a link/URL to the source document (optional).
+        /// </summary>
+        public string? Link { get; init; }
+
+        /// <summary>
+        /// Gets or sets the textual content of the retrieved chunk.
+        /// </summary>
+        public string Value { get; init; } = string.Empty;
+    }
```
markwallace-microsoft, line 305: Should this has a `RawRepresentation` for braking glass scenarios?

dotnet/src/Microsoft.Agents.AI/Data/RagProvider.cs
```diff
@@ -0,0 +1,300 @@
...
+/// </list>
+/// </para>
+/// <para>
+/// When <see cref="RagProviderOptions.RecentMessageMemoryLimit"/> is greater than zero the provider will retain the most recent
+/// user and assistant messages (up to the configured limit) across invocations and prepend them (in chronological order)
+/// to the current request messages when forming the search input. This can improve search relevance by providing
+/// multi-turn context to the retrieval layer without permanently altering the conversation history.
+/// </para>
+/// </remarks>
+public sealed class RagProvider : AIContextProvider
```
markwallace-microsoft: Are you planning a getting started sample showing how to use this?

westey-m: Yes, I'll add samples and docs once the core implementations are merged for this and the others.

dotnet/src/Microsoft.Agents.AI/Data/TextSearchProvider.cs
westey-m: Yes, I think this makes sense.  If users provide their own formatter, they might want to get at the raw representation they provided in the search delegate.  I'll add..

dotnet/src/Microsoft.Agents.AI/Data/RagProvider.cs
westey-m: This won't ever be used without this class, so not sure if it's worth cluttering the namespace with it.
Let's decide on this when we have decided what namespace/package this code will live in.

dotnet/src/Microsoft.Agents.AI/Data/TextSearchProvider.cs
westey-m: As mentioned in the meeting we had about this, this is data access related, and we had this in the same place for SK as well.  I think it continues to make sense not to add all data related features to the root namespace.

dotnet/tests/Microsoft.Agents.AI.UnitTests/Data/TextSearchProviderTests.cs
westey-m: As decided in the scrum, this will stay in the MAAI package.

SergeyMenshykh (approved):

verdie-g:
Is it expected that the class is called `TextSearchProvider.TextSearchSearchResult` and not `TextSearchProvider.TextSearchResult` or just `TextSearchResult` with no parent class?

westey-m:
> Is it expected that the class is called `TextSearchProvider.TextSearchSearchResult` and not `TextSearchProvider.TextSearchResult` or just `TextSearchResult` with no parent class?

@verdie-g, thanks for the callout.  Missed that after a rename.  Fixing in a separate PR #1780

Issue #1505, opened 2025-10-15
.NET: Excessive boilerplate for executors compared to Python - poor discoverability of simpler patterns


joslat:
## Summary
Creating executors in .NET requires significantly more boilerplate code compared to the Python SDK, creating a poor developer experience and high barrier to entry. While simpler patterns exist (e.g., .AsExecutor() for functions), they are poorly documented and not shown in official samples.
Python SDK (3 lines):
```
@executor(id="reverse_text_executor")
async def reverse_text(text: str, ctx: WorkflowContext[Never, str]) -> None:
    result = text[::-1]
    await ctx.yield_output(result)
```

.NET SDK - Official Pattern (10+ lines):
```
internal sealed class ReverseExecutor : ReflectingExecutor<ReverseExecutor>, IMessageHandler<string, string>
{
    public ReverseExecutor() : base("ReverseExecutor") { }

    public ValueTask<string> HandleAsync(string message, IWorkflowContext context, CancellationToken cancellationToken = default)
    {
        var result = string.Concat(message.Reverse());
        return ValueTask.FromResult(result);
    }
}
```

.NET SDK - Undocumented Simpler Pattern (3 lines):
```
Func<string, IWorkflowContext, CancellationToken, ValueTask<string>> reverseFunc = 
    async (text, ctx, ct) => string.Concat(text.Reverse());
var executor = reverseFunc.AsExecutor("ReverseExecutor");
```
Problem: The simpler pattern exists but is hidden - all documentation and samples use the verbose class-based approach.

---

## Current Issues
1. Excessive Boilerplate Required
Class-based pattern requires:
•	Creating a class (sealed for performance)
•	Inheriting from ReflectingExecutor<T> with recursive generic
•	Implementing IMessageHandler<TIn, TOut> interface
•	Writing a constructor that calls base with ID
•	Implementing HandleAsync with all parameters
•	Understanding ValueTask, CancellationToken, etc.
For a simple function, this is 10x more code than Python

2. Poor Discoverability of Alternatives
The .AsExecutor() extension method exists but:
•	❌ Not mentioned in official documentation
•	❌ Not shown in any samples
•	❌ Not discoverable through IntelliSense on functions
•	❌ No examples in getting started guide
Result: Developers assume the verbose pattern is the only way.

3. Documentation Mismatch
Python tutorial (https://learn.microsoft.com/en-us/agent-framework/tutorials/workflows/simple-sequential-workflow?pivots=programming-language-python):
[...]

Issue #1305, opened 2025-10-08
Latest release changes WorkflowBuilder API


Ryan-Palmer:
I think an unintentional change might have crept in, and if not the docs will need to be updated.

Previously, I could do e.g.

```
let workflow =
    WorkflowBuilder(agent1)
        .AddEdge(agent1, agent2)
        .WithOutputFrom(agent2)
        .Build();
```

Now I have to pass an `ExecutorIsh` like so:

```
let workflow =
    WorkflowBuilder(ExecutorIsh agent1)
        .AddEdge(ExecutorIsh agent1, ExecutorIsh agent2)
        .WithOutputFrom(ExecutorIsh agent2)
        .Build();
```

Side note, I think the name `ExecutorIsh` is really not great (specifically the 'ish' part - either it is or isn't and needs a new name).

stephentoub:
> Side note, I think the name ExecutorIsh is really not great (specifically the 'ish' part - either it is or isn't and needs a different name).

100%
https://github.com/microsoft/agent-framework/issues/758

Ryan-Palmer:
>100%

Ah I guess the implicit casting you mentioned has been removed (yay!) but the ExecutorIsh remains (boo!)

lokitoth:
There should not be any changes here. Do you have a code sample that used to work and doesn't now? - and were there other changes you made?

ExecutorIsh was the top-level input since the very beginning, and this has not changed, so I am unsure what is causing you to need to explicitly cast to it now?

With that said: ExecutorIsh is likely to undergo further changes, but in principle, being able to pass in Executors, Agents, etc. directly to the builder APIs should remain.

Ryan-Palmer:
I pasted a sample in the original issue, I don't know what more to say? I could pass agents in, then I pulled a nuget update after which I had to wrap them in Executorish.

From Stephen's linked issue, it seems there was some implicit casting going on under the surface which F# doesn't tend to play along with, so perhaps it doesn't show up in C#? It was working before I pulled the update anyway so something must have changed in the implementation.
________________________________

teknikal-wizard:
(Ryan from another account) 
Looks like F# won't play with implicit operators as I suspected:
https://stackoverflow.com/questions/10719770/is-there-anyway-to-use-c-sharp-implicit-operators-from-f

[...]

PR #1637, opened 2025-10-22
.NET: [BREAKING] Unify ExecutorIsh and ExecutorRegistration, unify/simplify APIs

### LINKED ISSUE #758: .NET: ExecutorIsh is confusingly named and hard to discover how to use
https://github.com/microsoft/agent-framework/blob/main/dotnet/src/Microsoft.Agents.Workflows/ExecutorIsh.cs

The name is hard to understand, with "Ish" as a suffix.

Also when looking at APIs that accept one, like WorkflowBuilder's constructor, it's hard to figure out how to get one. It relies on implicit cast operators to turn other things into this... and such implicit operators are also something we recommend folks avoid:
https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/operator-overloads#conversion-operators

The implicit operators also make it so that reference identity breaks, because code like:
```csharp
AIAgent agent = ...;
ExecutorIsh e1 = agent;
ExecutorIsh e2 = agent;
```
ends up allocating new objects as part of those implicit casts.

### LINKED ISSUE #1494: .NET: Provide explicit API to convert AIAgent to Executor for workflow composition
## Summary
When building workflows with AI agents, there is no explicit, documented API to convert an AIAgent (e.g., ChatClientAgent) into an Executor for use in workflow graphs. While agents can be added directly to workflows using AddFanOutEdge, there's ambiguity about whether agents can be treated as executors in all contexts, and no way to customize the executor name or description when doing so.

This request is for an explicit conversion API (e.g., .AsExecutor()) that would:
1.	Make the agent-to-executor pattern more discoverable
2.	Allow customization of executor metadata (name, description)
3.	Clarify the mental model of how agents integrate into workflows

## Related issue
https://github.com/microsoft/agent-framework/issues/1495

---

## Current Behavior
Agents can be used directly in workflows via fan-out edges, but there's no explicit conversion API:
```
// ✅ This works (implicit conversion)
ChatClientAgent physicist = new ChatClientAgent(
    chatClient,
    name: "Physicist",
    instructions: "You are an expert in physics."
);

var workflow = new WorkflowBuilder(startExecutor)
    .AddFanOutEdge(startExecutor, targets: [physicist, chemist])  // Agents used directly
    .AddFanInEdge(aggregationExecutor, sources: [physicist, chemist])
    .Build();
```

Issues with current approach:
1.	Discoverability: Not clear that agents can be used as executors
2.	Customization: Cannot override executor name/description independently from agent name
3.	Visualization: Agents show as GUIDs in workflow diagrams (see related issue https://github.com/microsoft/agent-framework/issues/1495)
4.	Mental model: Unclear if agents ARE executors or are CONVERTED to executors

---

## Expected Behavior
Provide an explicit extension method or API to convert agents to executors with customizable metadata:

```
// ✅ Explicit conversion with customization
ChatClientAgent physicist = new ChatClientAgent(
    chatClient,
    name: "Physicist",
    instructions: "You are an expert in physics."
);

// Convert agent to executor with custom metadata
ExecutorIsh physicistExecutor = physicist.AsExecutor(
    name: "PhysicistAgent_v1",           // Override display name for workflows
    description: "Physics domain expert"  // Add description for documentation
);

var workflow = new WorkflowBuilder(startExecutor)
    .AddFanOutEdge(startExecutor, targets: [physicistExecutor])
    .AddFanInEdge(aggregationExecutor, sources: [physicistExecutor])
    .Build();
```

Benefits:
1.	Self-documenting code: Makes intent explicit
2.	Customization: Allows workflow-specific naming without changing agent identity
3.	Better visualizations: Custom names appear in Mermaid/DOT diagrams
4.	Consistency: Similar to .AsAgent() pattern used for workflows

### LINKED ISSUE #1613: .NET: Cannot reuse Workflow with shared Executor instances that do not implement IResettableExecutor
- Sample: https://github.com/microsoft/agent-framework/tree/main/dotnet/samples/GettingStarted/Workflows/Agents/WorkflowAsAnAgent
- Detail: Running the sample linked above will result in exception `InvalidOperationException("Cannot reuse Workflow with shared Executor instances that do not implement IResettableExecutor.");` when the conversation exceeds one turn.

lokitoth:
### Motivation and Context

`ExecutorIsh`, a type built to work around the lack of full support for anonymous discriminated unions in .NET, leads to confusion, and makes it difficult to find how to integrate `Executor`-like types into a workflow. It also duplicates some functionality with `ExecutorRegistration`, and now that something broke in F#'s ability to feed the auto-cast types to `WorkflowBuilder` the value of the split between the two is lowered.

### Description

* Unifies `ExecutorIsh` and `ExecutorRegistration`, relying on the record type hierarchy for sum-type-like behaviour
* Normalizes extension methods to convert objects into `ExecutorRegistration`s: `.AsExecutor()` and `.RegisterExecutor()`
* Add extra explicit extension methods to correspond to implicit casts
* Adds overload to support simpler function-to-executor pattern
* Fixes #1613

Breaking:
* Removes `ExecutorIsh` type, in favour of `ExecutorRegistration`
* Obsoletes `.ConfigureFactory()` and `.ConfigureSubworkflow()` extension methods in favour of `.RegisterExecutor()` and `.AsExecutor()`, respectively

Relevant Issues:
* Closes #1613 
* #1305
* #1505
* Closes #758
* Closes #1494

### Contribution Checklist

### DIFF
--- a/dotnet/samples/GettingStarted/Workflows/_Foundational/01_ExecutorsAndEdges/Program.cs
+++ b/dotnet/samples/GettingStarted/Workflows/_Foundational/01_ExecutorsAndEdges/Program.cs
@@ -21,7 +21,7 @@ public static class Program
     {
         // Create the executors
         Func<string, string> uppercaseFunc = s => s.ToUpperInvariant();
-        var uppercase = uppercaseFunc.AsExecutor("UppercaseExecutor");
+        var uppercase = uppercaseFunc.BindAsExecutor("UppercaseExecutor");
 
         ReverseTextExecutor reverse = new();
 
--- a/dotnet/samples/GettingStarted/Workflows/_Foundational/06_SubWorkflows/Program.cs
+++ b/dotnet/samples/GettingStarted/Workflows/_Foundational/06_SubWorkflows/Program.cs
@@ -40,7 +40,7 @@ public static class Program
             .Build();
 
         // Step 2: Configure the sub-workflow as an executor for use in the parent workflow
-        ExecutorBinding subWorkflowExecutor = subWorkflow.AsExecutor("TextProcessingSubWorkflow");
+        ExecutorBinding subWorkflowExecutor = subWorkflow.BindAsExecutor("TextProcessingSubWorkflow");
 
         // Step 3: Build a main workflow that uses the sub-workflow as an executor
         Console.WriteLine("Building main workflow that uses the sub-workflow as an executor...\n");
--- a/dotnet/src/Microsoft.Agents.AI.Workflows/Checkpointing/WorkflowInfo.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Workflows/Checkpointing/WorkflowInfo.cs
@@ -11,13 +11,13 @@ internal sealed class WorkflowInfo
 {
     [JsonConstructor]
     internal WorkflowInfo(
-        Dictionary<string, ExecutorInfo> executorBindings,
+        Dictionary<string, ExecutorInfo> executors,
         Dictionary<string, List<EdgeInfo>> edges,
         HashSet<RequestPortInfo> requestPorts,
         string startExecutorId,
         HashSet<string>? outputExecutorIds)
     {
-        this.ExecutorBindings = Throw.IfNull(executorBindings);
+        this.Executors = Throw.IfNull(executors);
         this.Edges = Throw.IfNull(edges);
         this.RequestPorts = Throw.IfNull(requestPorts);
 
@@ -25,7 +25,7 @@ internal sealed class WorkflowInfo
         this.OutputExecutorIds = outputExecutorIds ?? [];
     }
 
-    public Dictionary<string, ExecutorInfo> ExecutorBindings { get; }
+    public Dictionary<string, ExecutorInfo> Executors { get; }
     public Dictionary<string, List<EdgeInfo>> Edges { get; }
     public HashSet<RequestPortInfo> RequestPorts { get; }
 
@@ -47,10 +47,10 @@ internal sealed class WorkflowInfo
         }
 
         // Validate the executors
-        if (workflow.ExecutorBindings.Count != this.ExecutorBindings.Count ||
-            this.ExecutorBindings.Keys.Any(
-            executorId => workflow.ExecutorBindings.TryGetValue(executorId, out ExecutorBinding? registration)
-                       && !this.ExecutorBindings[executorId].IsMatch(registration)))
+        if (workflow.ExecutorBindings.Count != this.Executors.Count ||
+            this.Executors.Keys.Any(
+            executorId => workflow.ExecutorBindings.TryGetValue(executorId, out ExecutorBinding? binding)
+                       && !this.Executors[executorId].IsMatch(binding)))
         {
             return false;
         }
--- a/dotnet/src/Microsoft.Agents.AI.Workflows/ExecutorBindingExtensions.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Workflows/ExecutorBindingExtensions.cs
@@ -137,7 +137,7 @@ public static class ExecutorBindingExtensions
     /// <param name="id">An id for the executor to be instantiated.</param>
     /// <param name="options">An optional parameter specifying the options.</param>
     /// <returns>An <see cref="ExecutorBinding"/> instance that resolves to the result of the factory call when messages get sent to it.</returns>
-    [Obsolete("Use RegisterExecutor() instead")]
+    [Obsolete("Use BindExecutor() instead")]
     [EditorBrowsable(EditorBrowsableState.Never)]
     public static ExecutorBinding ConfigureFactory<TExecutor, TOptions>(this Func<Config<TOptions>, string, ValueTask<TExecutor>> factoryAsync, string id, TOptions? options = null)
         where TExecutor : Executor
@@ -161,10 +161,10 @@ public static class ExecutorBindingExtensions
     /// <param name="id">A unique identifier for the sub-workflow execution. Used to distinguish this sub-workflow instance.</param>
     /// <param name="options">Optional configuration options for the sub-workflow executor. If null, default options are used.</param>
     /// <returns>An ExecutorRegistration instance representing the configured sub-workflow executor.</returns>
-    [Obsolete("Use AsExecutor() instead")]
+    [Obsolete("Use BindAsExecutor() instead")]
     [EditorBrowsable(EditorBrowsableState.Never)]
     public static ExecutorBinding ConfigureSubWorkflow(this Workflow workflow, string id, ExecutorOptions? options = null)
-        => workflow.AsExecutor(id, options);
+        => workflow.BindAsExecutor(id, options);
 
     /// <summary>
     /// Configures a sub-workflow executor for the specified workflow, using the provided identifier and options.
@@ -173,7 +173,7 @@ public static class ExecutorBindingExtensions
     /// <param name="id">A unique identifier for the sub-workflow execution. Used to distinguish this sub-workflow instance.</param>
     /// <param name="options">Optional configuration options for the sub-workflow executor. If null, default options are used.</param>
     /// <returns>An <see cref="ExecutorBinding"/> instance representing the configured sub-workflow executor.</returns>
-    public static ExecutorBinding AsExecutor(this Workflow workflow, string id, ExecutorOptions? options = null)
+    public static ExecutorBinding BindAsExecutor(this Workflow workflow, string id, ExecutorOptions? options = null)
         => new SubworkflowBinding(workflow, id, options);
 
     /// <summary>
@@ -186,7 +186,7 @@ public static class ExecutorBindingExtensions
     /// <param name="options">Configuration options for the executor. If <c>null</c>, default options will be used.</param>
     /// <param name="threadsafe">Declare that the message handler may be used simultaneously by multiple runs concurrently.</param>
     /// <returns>An <see cref="ExecutorBinding"/> instance that wraps the provided asynchronous message handler and configuration.</returns>
-    public static ExecutorBinding AsExecutor<TInput>(this Func<TInput, IWorkflowContext, CancellationToken, ValueTask> messageHandlerAsync, string id, ExecutorOptions? options = null, bool threadsafe = false)
+    public static ExecutorBinding BindAsExecutor<TInput>(this Func<TInput, IWorkflowContext, CancellationToken, ValueTask> messageHandlerAsync, string id, ExecutorOptions? options = null, bool threadsafe = false)
         => new FunctionExecutor<TInput>(id, messageHandlerAsync, options, declareCrossRunShareable: threadsafe).ToBinding(messageHandlerAsync);
 
     /// <summary>
@@ -198,10 +198,10 @@ public static class ExecutorBindingExtensions
     /// <param name="id">An optional unique identifier for the executor. If <c>null</c>, will use the function argument as an id.</param>
     /// <param name="options">Configuration options for the executor. If <c>null</c>, default options will be used.</param>
     /// <param name="threadsafe">Declare that the message handler may be used simultaneously by multiple runs concurrently.</param>
-    /// <returns>An <see cref="ExecutorRegistration"/> instance that wraps the provided asynchronous message handler and configuration.</returns>
-    public static ExecutorRegistration AsExecutor<TInput>(this Func<TInput, ValueTask> messageHandlerAsync, string id, ExecutorOptions? options = null, bool threadsafe = false)
+    /// <returns>An <see cref="ExecutorBinding"/> instance that wraps the provided asynchronous message handler and configuration.</returns>
+    public static ExecutorBinding BindAsExecutor<TInput>(this Func<TInput, ValueTask> messageHandlerAsync, string id, ExecutorOptions? options = null, bool threadsafe = false)
         => ((Func<TInput, IWorkflowContext, CancellationToken, ValueTask>)((input, _, __) => messageHandlerAsync(input)))
-                .AsExecutor(id, options, threadsafe);
+                .BindAsExecutor(id, options, threadsafe);
 
     /// <summary>
     /// Configures a function-based asynchronous message handler as an executor with the specified identifier and
@@ -212,10 +212,10 @@ public static class ExecutorBindingExtensions
     /// <param name="id">An optional unique identifier for the executor. If <c>null</c>, will use the function argument as an id.</param>
     /// <param name="options">Configuration options for the executor. If <c>null</c>, default options will be used.</param>
     /// <param name="threadsafe">Declare that the message handler may be used simultaneously by multiple runs concurrently.</param>
-    /// <returns>An <see cref="ExecutorRegistration"/> instance that wraps the provided asynchronous message handler and configuration.</returns>
-    public static ExecutorRegistration AsExecutor<TInput>(this Func<TInput, IWorkflowContext, ValueTask> messageHandlerAsync, string id, ExecutorOptions? options = null, bool threadsafe = false)
+    /// <returns>An <see cref="ExecutorBinding"/> instance that wraps the provided asynchronous message handler and configuration.</returns>
+    public static ExecutorBinding BindAsExecutor<TInput>(this Func<TInput, IWorkflowContext, ValueTask> messageHandlerAsync, string id, ExecutorOptions? options = null, bool threadsafe = false)
         => ((Func<TInput, IWorkflowContext, CancellationToken, ValueTask>)((input, ctx, __) => messageHandlerAsync(input, ctx)))
-                .AsExecutor(id, options, threadsafe);
+                .BindAsExecutor(id, options, threadsafe);
 
     /// <summary>
     /// Configures a function-based asynchronous message handler as an executor with the specified identifier and
@@ -226,10 +226,10 @@ public static class ExecutorBindingExtensions
     /// <param name="id">An optional unique identifier for the executor. If <c>null</c>, will use the function argument as an id.</param>
     /// <param name="options">Configuration options for the executor. If <c>null</c>, default options will be used.</param>
     /// <param name="threadsafe">Declare that the message handler may be used simultaneously by multiple runs concurrently.</param>
-    /// <returns>An <see cref="ExecutorRegistration"/> instance that wraps the provided asynchronous message handler and configuration.</returns>
-    public static ExecutorRegistration AsExecutor<TInput>(this Func<TInput, CancellationToken, ValueTask> messageHandlerAsync, string id, ExecutorOptions? options = null, bool threadsafe = false)
+    /// <returns>An <see cref="ExecutorBinding"/> instance that wraps the provided asynchronous message handler and configuration.</returns>
+    public static ExecutorBinding BindAsExecutor<TInput>(this Func<TInput, CancellationToken, ValueTask> messageHandlerAsync, string id, ExecutorOptions? options = null, bool threadsafe = false)
         => ((Func<TInput, IWorkflowContext, CancellationToken, ValueTask>)((input, _, ct) => messageHandlerAsync(input, ct)))
-                .AsExecutor(id, options, threadsafe);
+                .BindAsExecutor(id, options, threadsafe);
 
     /// <summary>
     /// Configures a function-based message handler as an executor with the specified identifier and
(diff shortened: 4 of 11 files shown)


crickman:
Good riddance `Executorish` (imho).

crickman:
For thew new public registration types, is there sample debt being incurred here? If so, can some issues be created to track new samples that might be desired?

crickman:
Are there any patterns here that should be channeled back into the python platform? @ekzhu?

crickman:
Do we have tests that cover the issues being closed?

lokitoth:
> For thew new public registration types, is there sample debt being incurred here?

Not really: It's the same thing as with `ExecutorIsh` before, just broken out into a type hierarchy, and merged with `ExecutorRegistration`. Not really expected to be used directly, vs. using the `.AsExecutor()`, `RegisterExecutor()` and implicits. There is an issue I am leaving opened referenced here that will be used to track to finish the documentation of the FunctionExecutor APIs and helpers. (#1505)

> Do we have tests that cover the issues being closed?

Yes, the test was updated to cover the scenario that discovered this.

dotnet/tests/Microsoft.Agents.AI.Workflows.UnitTests/RepresentationTests.cs
```diff
@@ -53,17 +52,19 @@ public async Task Test_Executorish_InfosAsync()
...
 
         Func<int, IWorkflowContext, CancellationToken, ValueTask> function = MessageHandlerAsync;
         await RunExecutorishTestAsync(function.AsExecutor("FunctionExecutor"));
 
-        if (Enum.GetValues(typeof(ExecutorIsh.Type)).Length > testsRun + 1)
-        {
-            Assert.Fail("Not all ExecutorIsh types were tested.");
-        }
+        // It would be great if there was a way to test exhaustion for record-based Sum Types
+        // TODO: Unit tests don't have to be AOT-compatible, so...
```
ReubenBond: Looks like there's a proposal for "closed hierarchies" https://github.com/dotnet/csharplang/blob/7603e3bee2ea3edb53e6a3b27efbaf24949b609f/proposals/closed-hierarchies.md

crickman:
> > For thew new public registration types, is there sample debt being incurred here?
> 
> Not really: It's the same thing as with `ExecutorIsh` before, just broken out into a type hierarchy, and merged with `ExecutorRegistration`. Not really expected to be used directly, vs. using the `.AsExecutor()`, `RegisterExecutor()` and implicits. There is an issue I am leaving opened referenced here that will be used to track to finish the documentation of the FunctionExecutor APIs and helpers. (#1505)
> 
> > Do we have tests that cover the issues being closed?
> 
> Yes, the test was updated to cover the scenario that discovered this.

if its not expected / intended to be used directly, then why `public`?

dotnet/samples/GettingStarted/Workflows/_Foundational/01_ExecutorsAndEdges/Program.cs
```diff
@@ -20,7 +20,9 @@ public static class Program
     private static async Task Main()
     {
         // Create the executors
-        UppercaseExecutor uppercase = new();
+        Func<string, string> uppercaseFunc = s => s.ToUpperInvariant();
+        var uppercase = uppercaseFunc.AsExecutor("UppercaseExecutor");
```
stephentoub: If this were instead just a static method like `Executor.Create`, it could instead be:
```csharp
var uppercase = Executor.Create<string, string>(s => s.ToUpperInvariant(), "UppercaseExecutor");
```

dotnet/samples/GettingStarted/Workflows/_Foundational/06_SubWorkflows/Program.cs
```diff
@@ -40,7 +40,7 @@ private static async Task Main()
             .Build();
 
         // Step 2: Configure the sub-workflow as an executor for use in the parent workflow
-        ExecutorIsh subWorkflowExecutor = subWorkflow.ConfigureSubWorkflow("TextProcessingSubWorkflow");
+        ExecutorRegistration subWorkflowExecutor = subWorkflow.AsExecutor("TextProcessingSubWorkflow");
```
stephentoub: Does `AsExecutor` return an "Executor" or an "ExecutorRegistration"?

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Interpreter/WorkflowModelBuilder.cs
```diff
@@ -24,7 +24,7 @@ public void Connect(IModeledAction source, IModeledAction target, Func<object?,
             condition);
     }
 
-    private static ExecutorIsh GetExecutorIsh(IModeledAction action) =>
+    private static ExecutorRegistration GetExecutorIsh(IModeledAction action) =>
```
stephentoub: "Ish"

dotnet/src/Microsoft.Agents.AI.Workflows/AgentWorkflowBuilder.cs
```diff
@@ -124,8 +124,8 @@ private static Workflow BuildConcurrentCore(
         // so that the final accumulator receives a single list of messages from each agent. Otherwise, the
         // accumulator would not be able to determine what came from what agent, as there's currently no
         // provenance tracking exposed in the workflow context passed to a handler.
-        ExecutorIsh[] agentExecutors = (from agent in agents select (ExecutorIsh)new AgentRunStreamingExecutor(agent, includeInputInOutput: false)).ToArray();
-        ExecutorIsh[] accumulators = [.. from agent in agentExecutors select (ExecutorIsh)new CollectChatMessagesExecutor($"Batcher/{agent.Id}")];
+        ExecutorRegistration[] agentExecutors = (from agent in agents select (ExecutorRegistration)new AgentRunStreamingExecutor(agent, includeInputInOutput: false)).ToArray();
```
stephentoub: "Executors"... these are now "Registrations"?

dotnet/src/Microsoft.Agents.AI.Workflows/ConfiguredExecutorRegistration.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Agents.AI.Workflows;
+
+// TODO: Unwrap the Configured object, just like for WorkflowRegistration
+internal record ConfiguredExecutorRegistration(Configured<Executor> ConfiguredExecutor, Type ExecutorType)
```
stephentoub: Is "Configured" still needed? I guess I don't understand its purpose separate from these registrations (and its name is fairly ambiguous).

dotnet/src/Microsoft.Agents.AI.Workflows/ExecutorRegistrationExtensions.cs
```diff
@@ -0,0 +1,429 @@
...
+{
+    /// <summary>
+    /// Configures an <see cref="Executor"/> instance for use in a workflow.
+    /// </summary>
+    /// <remarks>
+    /// Note that Executor Ids must be unique within a workflow.
+    /// </remarks>
+    /// <param name="executor">The executor instance.</param>
+    /// <returns>An <see cref="ExecutorRegistration"/> instance wrapping the specified <see cref="Executor"/>.</returns>
+    public static ExecutorInstanceRegistration RegisterExecutor(this Executor executor)
```
stephentoub: ExecutorInstanceRegistration instead of ExecutorRegistration?

dotnet/src/Microsoft.Agents.AI.Workflows/ExecutorRegistrationExtensions.cs
```diff
@@ -0,0 +1,429 @@
...
+    /// Note that Executor Ids must be unique within a workflow.
+    ///
+    /// Although this will generally result in a delay-instantiated <see cref="Executor"/> once messages are available
+    /// for it, it will be instantiated if a <see cref="ProtocolDescriptor"/> for the <see cref="Workflow"/> is requested,
+    /// and it is the starting executor.
+    /// </remarks>
+    /// <typeparam name="TExecutor">The type of the resulting executor</typeparam>
+    /// <param name="factoryAsync">The factory method.</param>
+    /// <returns>An <see cref="ExecutorRegistration"/> instance that resolves to the result of the factory call when messages get sent to it.</returns>
+    public static ExecutorRegistration RegisterExecutor<TExecutor>(this Func<string, string, ValueTask<TExecutor>> factoryAsync)
```
stephentoub: "Register" makes it sound like something is being stored somewhere, some existing object is being mutated to hold this new record, etc., like on CancellationToken. But this is only creating the ExecutorRegistration instance, right? It's not actually registering it with something?

dotnet/src/Microsoft.Agents.AI.Workflows/ExecutorRegistrationExtensions.cs
```diff
@@ -0,0 +1,429 @@
...
+    /// Note that Executor Ids must be unique within a workflow.
+    ///
+    /// Although this will generally result in a delay-instantiated <see cref="Executor"/> once messages are available
+    /// for it, it will be instantiated if a <see cref="ProtocolDescriptor"/> for the <see cref="Workflow"/> is requested,
+    /// and it is the starting executor.
+    /// </remarks>
+    /// <typeparam name="TExecutor">The type of the resulting executor</typeparam>
+    /// <param name="factoryAsync">The factory method.</param>
+    /// <returns>An <see cref="ExecutorRegistration"/> instance that resolves to the result of the factory call when messages get sent to it.</returns>
+    [Obsolete("Use RegisterExecutor() instead.")]
```
stephentoub: Is this to give folks time to migrate off of it? Consider also adding EditorBrowsableState.Never

dotnet/src/Microsoft.Agents.AI.Workflows/ExecutorRegistration.cs
```diff
@@ -1,67 +1,130 @@
...
+    /// RequestPort.</remarks>
+    /// <param name="port">The RequestPort instance to convert.</param>
+    public static implicit operator ExecutorRegistration(RequestPort port) => port.AsExecutor();
 
-    public async ValueTask<Executor> CreateInstanceAsync(string runId) => this.CheckId(await this.ProviderAsync(runId).ConfigureAwait(false));
+    /// <summary>
+    /// Defines an implicit conversion from an AIAgent to an ExecutorRegistrationIsh instance.
+    /// </summary>
+    /// <param name="agent"></param>
+    public static implicit operator ExecutorRegistration(AIAgent agent) => agent.AsExecutor();
```
stephentoub: You know my opinion on these, but I'll stop commenting on it.

dotnet/src/Microsoft.Agents.AI.Workflows/ExecutorRegistrationExtensions.cs
lokitoth: Yes: I intend to remove it when we get to the "RC-like" phase or before. And good call on Browsable.

dotnet/src/Microsoft.Agents.AI.Workflows/ExecutorRegistration.cs
lokitoth: I think we're going to have to agree to disagree on this point, yeah :). 

I promise to remove it is as soon as I can replace it anonymous type unions that work across languages, though, and not add more except in specifically cases like these.

dotnet/src/Microsoft.Agents.AI.Workflows/ExecutorRegistrationExtensions.cs
lokitoth: No, but I could not come up with a clearly better name for it. The thinking around "registration" was in the sense of being the "registration" one submits when trying to get registered for something, not the established "registration" after the fact. 

I need a name for an object that represents a future executor (or a placeholder for one - so factory is not really a good term, as it only deals with one part), with certain info about its underlying type that get erased by the time it is in that object, and the ability to create and "reset" it (if shared).

Maybe "ExecutorProvider"? Though it goes a fair bit beyond "Providing" an Executor, so it does not really fit either.

Aside: The `Registration` bit would be internal to the `Workflow`/`Builder`, but previous name for such a thing was `...Ish`, and that was pretty polarizing.

dotnet/src/Microsoft.Agents.AI.Workflows/ConfiguredExecutorRegistration.cs
lokitoth: A way to smuggle options objects and certain other important information and avoid having to constantly break the interface for developers, but that did not last long as a conviction, due to the arrival of RunId, which is not available when the configuration objects are created.

This is probably the next place to "fix" as we review the APIs.

dotnet/src/Microsoft.Agents.AI.Workflows/AgentWorkflowBuilder.cs
lokitoth: Until we get real anonymous union types we need to model the type explicitly. We need a way to allow passing information about future Executors without encouraging passing Executor instances to WorkflowBuilder, but also without disallowing it.

We are trying to avoid having to do `.RegisterExecutor()` on the builder object, as it feels more heavy-weight for the developer, and we already have much more ceremony than Python due to limitations stemming from AOT, among other things. 

At some point we have to make compromises somewhere due to the limits of the expressivity of the language.

dotnet/src/Microsoft.Agents.AI.Workflows/ExecutorRegistrationExtensions.cs
lokitoth: At one point I got a CodeAnalysis performance complaint about something similar where I was returning a more specific type, and I flipped all of them. Doesn't seem to be triggering now, so must have had some other reason to fire and I did not understand it right.

dotnet/samples/GettingStarted/Workflows/_Foundational/01_ExecutorsAndEdges/Program.cs
lokitoth: Is there a reason to prefer `Executor.Create<TIn, TOut>` over `Executor<TIn, TOut>.Create`?

Also, one can just do `new FunctionExecutor<TIn, TOut>(s => s.ToUpperInvariant(), "...")`, or even better, get rid of `var`, put the type on the left, and use typeless `new(...)`. Maybe we should remove the `.AsExecutor()` overloads for functions and instead put the overloads on the `FunctionExecutor` ctors?

(I was genuinely disappointed when I realized that it cannot be applied directly to a method, or a lambda in parens, so less valuable as an API than I had thought it would be. Feels like a very low-hanging fruit to be able to bind that way, especially in the extensions-everything world - but then, this is the "implicit cast" thing again, and I know your stance on that :))

dotnet/samples/GettingStarted/Workflows/_Foundational/06_SubWorkflows/Program.cs
lokitoth: `ToRegistration` would be more accurate, but my thinking was that `AsExecutor` is more obviously relevant when browsing IntelliSense. Will think more on the name today, to see if I can come up with something better.

dotnet/samples/GettingStarted/Workflows/_Foundational/01_ExecutorsAndEdges/Program.cs
stephentoub: > Is there a reason to prefer `Executor.Create<TIn, TOut>` over `Executor<TIn, TOut>.Create`?

The main benefit of the former is you get generic type inference if the compiler can discern it, e.g. with the former you can write:
```csharp
Executor.Create((string s) => s.ToUpperInvariant(), "UppercaseExecutor");
```
but not with the latter. The former also lets you avoid having the extra type with two generic type parameters if we don't already have it.

dotnet/tests/Microsoft.Agents.AI.Workflows.UnitTests/RepresentationTests.cs
lokitoth: Nice, that could help with certainty of exhaustiveness! 

Would still probably need reflection to count them, which is more what I was getting at in the TODO.

dotnet/src/Microsoft.Agents.AI.Workflows/AgentWorkflowBuilder.cs
joslat: I am with @stephentoub how about WorkflowExecutor or a similar name? thus adding parity to the naming in Python?

lokitoth:
> if its not expected / intended to be used directly, then why `public`?

I need a type to use on a public API that has the property of being a type-union of the types I actually want people to use. That's why `Ish` was, in my opinion, a better name.

dotnet/src/Microsoft.Agents.AI.Workflows/ExecutorRegistrationExtensions.cs
joslat: I believe that the process mostly is doing is assigning ownership of the workflow instance so it cannot be used while "taken" or "assigned", so I am all in with finding a suitable name that clearly shows what is happening. And IMHO it is not exactly "registration" its more an assignation, or giving ownership of this workflow executor. 
But I am all a newbie here so can be completely wrong...

dotnet/src/Microsoft.Agents.AI.Workflows/AgentWorkflowBuilder.cs
lokitoth: ~~What do people think about `WorkflowNode` or `ExecutorUnion`, or `BoundExecutor` (especially since I just realized we need to add object-ownership semantics to executor instances too if we want to be really careful)?~~

TODO: Need to add ownership semantics to Executor to avoid cross-workflow sharing. (In a separate PR)

lokitoth: Had a bit of an epiphany overnight. New object name: `ExecutorBinding`. It is more than a factory, and carries some state information with it, but it is not yet "registered" so Registration is not a good name for it - as pointed out above.

dotnet/src/Microsoft.Agents.AI.Workflows/Checkpointing/WorkflowInfo.cs
```diff
@@ -11,21 +11,21 @@ internal sealed class WorkflowInfo
 {
     [JsonConstructor]
     internal WorkflowInfo(
-        Dictionary<string, ExecutorInfo> executors,
+        Dictionary<string, ExecutorInfo> executorBindings,
```
TaoChenOSU: nit: the values are really executor info items

dotnet/src/Microsoft.Agents.AI.Workflows/Visualization/WorkflowVisualizer.cs
```diff
@@ -264,9 +264,9 @@ private static string ComputeShortHash(string input)
 #endif
     }
 
-    private static bool TryGetNestedWorkflow(ExecutorRegistration registration, [NotNullWhen(true)] out Workflow? workflow)
+    private static bool TryGetNestedWorkflow(ExecutorBinding registration, [NotNullWhen(true)] out Workflow? workflow)
```
TaoChenOSU:
```suggestion
    private static bool TryGetNestedWorkflow(ExecutorBinding binding, [NotNullWhen(true)] out Workflow? workflow)
```

dotnet/src/Microsoft.Agents.AI.Workflows/ExecutorBindingExtensions.cs
```diff
@@ -0,0 +1,434 @@
...
+            typeof(FunctionExecutor<TInput, TOutput>));
+
+    /// <summary>
+    /// Configures a sub-workflow executor for the specified workflow, using the provided identifier and options.
+    /// </summary>
+    /// <param name="workflow">The workflow instance to be executed as a sub-workflow. Cannot be null.</param>
+    /// <param name="id">A unique identifier for the sub-workflow execution. Used to distinguish this sub-workflow instance.</param>
+    /// <param name="options">Optional configuration options for the sub-workflow executor. If null, default options are used.</param>
+    /// <returns>An ExecutorRegistration instance representing the configured sub-workflow executor.</returns>
+    [Obsolete("Use AsExecutor() instead")]
```
TaoChenOSU:
```suggestion
    [Obsolete("Use BindAsExecutor() instead")]
```

mikeholczer:
I hope this is an OK place to ask this, but I'm seeing that in these changes in `AIAgentBinding` that the `ExecutorBinding` constructor is passed the `AIAgent`'s `Name` if it;'s not `null` and otherwise the `AIAgent`'s `Id` and in `ExecutorBinding.CheckId` there is a check that the `Executor`'s `Id` matches the `ExecutorBinding`'s `Id`. On the otherhand the `AIAgentHostExecutor`'s constructor always passes the `AIAgent`'s `Id` to it's base classes constructor. That creates a mismatch if an `AIAgent` has it's `Name` and `Id` set to different values. 

I don't know if that's a bug, or if it's not intended that one would set both the `Name` and `Id` of a `AIAgent`.

mikeholczer:
Looks like #1840 addressed the issue I was having.

PR #1657, opened 2025-10-23
.NET: Azure.AI.Agents Package Split + Initial Extensions

rogerbarreto:


dotnet/nuget.config
```diff
@@ -3,14 +3,14 @@
   <packageSources>
     <clear />
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
-    <add key="local" value="C:\localpackages" />
+    <add key="azure-sdk-for-net-pr" value="https://pkgs.dev.azure.com/azure-sdk/internal/_packaging/azure-sdk-for-net-pr/nuget/v3/index.json" />
```
crickman, line 6: Couldn't this be applied to the feature branch without waiting for the agent features?  If this is merging soon, then it call all come along together.

.github/workflows/dotnet-format.yml
```diff
@@ -7,7 +7,7 @@ name: dotnet-format
 on:
   workflow_dispatch:
   pull_request:
-    branches: ["main", "feature*"]
+    branches: ["main"]
```
crickman, line 10: Couldn't this be applied to the feature branch without waiting for the agent features?  If this is merging soon, then it call all come along together.

crickman (approved):

.github/workflows/dotnet-format.yml
rogerbarreto: I'm just removing this for to allow the PR to pass, not tackling this issue in the scope of this PR.

AFAIK keeping it fails, as the current pipeline doesn't have access to the resource. 

Now that our ADO pipelines are configured, just need to validate if any change needs to be made 

@dmytrostruk Is anything here I need to be aware of?

PR #1670, opened 2025-10-23
Python: Experimental: DevUI Workflow Chat UI

lavaman131:
### Motivation and Context

Please help reviewers and future users, providing the following information:
  1. Why is this change required?
 Exploration of alternative UIs to represent workflows.
  3. What problem does it solve?
Aiming to improve the UI/UX experience of users working with workflows.

### Description

This PR makes enhancements to the workflow UI in DevUI, allowing users to interact with users with a chat-like interface through another collapsible side-panel. See images of UI below.

This pull request improves the usability of the `python/check_md_code_blocks.py` script by allowing users to specify Markdown files using glob patterns, making it easier to check multiple files at once.

**Enhancements to file selection:**

* Added support for glob patterns in the command-line arguments, enabling users to specify files using wildcards (e.g., `docs/**/*.md`). (`python/check_md_code_blocks.py`) [[1]](diffhunk://#diff-fb3fa1115df52657161283c491ea8e19084c69f712586770ec8a3f791a5dda8dR7) [[2]](diffhunk://#diff-fb3fa1115df52657161283c491ea8e19084c69f712586770ec8a3f791a5dda8dL116-R131)

**UI Images**

Screenshot 2025-10-23 011344
Screenshot 2025-10-23 011408
Screenshot 2025-10-23 011436
Screenshot 2025-10-23 011500
Screenshot 2025-10-23 011516
Screenshot 2025-10-23 011531
Screenshot 2025-10-23 011639

PR #1697, opened 2025-10-25
Python: excluded devui frontend in pyproject.toml

lavaman131:
### Description

Updated devui `pyproject.toml` to exclude devui frontend following up #1670.


python/pyproject.toml
```diff
@@ -26,6 +26,7 @@ dependencies = [
     "agent-framework-a2a",
     "agent-framework-azure-ai",
     "agent-framework-copilotstudio",
+    "agent-framework-lab",
```
victordibia, line 29: Thanks @lavaman131 , what is the rationale for this addition/.

lavaman131: Not sure, this was added in @ekzhu's latest PR.

victordibia (approved):

python/pyproject.toml
victordibia: If it is already in a different PR. It should not be a diff in this PR right? 
Approving for now, thanks

moonbox3 (approved):

Issue #1710, opened 2025-10-26
.NET: ChatClientAgent having internal ChatOptions is overly restrictive


PederHP:
`ChatOptions` being internal means that IChatClient-based agents are restricted to being very static in nature. For example:

* Claude-based agents cannot change extended thinking parameters on the fly.
* OpenAI-based agents cannot change reasoning effort levels.
* Tools and ToolMode are static

Having agents that change extended thinking / reasoning effort levels based on user input or other signals is a common pattern, and used in many coding agents. Dynamic Tool strategies are also relatively common.

Making `ChatOptions` public and mutable would enable a lot of patterns. `Instructions` should also be mutable - having a dynamic system prompt / instructions is also a very common pattern.

stephentoub:
Note that you can send a different ChatOptions in as part of a ChatClientAgentRunOptions on each request, and it will be merged with the agent-level settings for that request.

PederHP:
> Note that you can send a different ChatOptions in as part of a ChatClientAgentRunOptions on each request, and it will be merged with the agent-level settings for that request.

True, and I did consider if that was enough, but it's awkward to keep a parallel datastructure and making sure it's passed whenever the agent is run. When deriving from ChatClientAgent (and keeping the overrides in the derived class state) it means having to merge with incoming ChatClientAgentRunOptions before passing them on.

Also for collections it will unionize and not replace the property, which I think is the right choice for per-request overrides, but not really for a mutable agent.

I couldn't really see any downside to making them mutable, as I think stateful agents are pretty common pattern. It's much simpler and I don't see any reason an agent should be restricted to immutability (although that is a reasonable choice for some architectures, but not all).

stephentoub:
(You can also insert ConfigureOptions as part of the IChatClient.)

Encouraging mutating the agent seems like a footgun. Serialized tests will show it all working fine, but then expected concurrent use will lead to hard to diagnose failures.

PederHP:
But that won't work for instructions will it? And again, it's very awkward for something that is trivial to do when using a custom wrapper around an IChatClient, instead of agent. I see your point about concurrency, though.

Maybe an AgentThread specific set of overrides?

stephentoub:
> But that won't work for instructions will it?

ConfigureOptions? Why not?

stephentoub:
[...]

PR #1711, opened 2025-10-26
.NET: Introduce MutableChatClientAgent

PederHP:
### Motivation and Context

Solves #1710 

It is a common pattern to have data in the `Instructions` (for caching and to achieve highest model trust level). System notifications and reminders are another example of dynamic instruction/system level data.

Mutable `ChatOptions` is needed to have agents with dynamic Tool sets and dynamic reasoning/thinking, which is also a common pattern. Another example is Claude skills and container ids.

`ChatClientAgentRunOptions` could be used as an alternative to this, but it is awkward, and for Tools it only allows adding Tools not changing the available Tools or removing some of them. 

I believe it is needed to support both per-request overrides and mutable agent properties. The alternative is to create a new agent whenever mutation is needed and migrate state to it.

### Description

If no `ChatClientAgentOptions` are provided in the constructor a default is instantiated. This is functionally equivalent to the current behavior and allows making the properties mentioned above public and mutable.

### Contribution Checklist


PederHP:

PederHP:
I'll update this. One reason it doesn't work well to just use a different agent if instructions/tools change is that sometimes one wants them to change during an agent invocation loop. This is relevant for agents that have tools which expand/collapse tool groups, or use system context for information dynamically updated by tools through a side-channel (as it is a higher level of trust than tool results and gets curated/culled differently when using providers that offer this or even doing it manually).

chetantoshniwal:
Closing as stale and since there are some alternative options.

PR #1603, opened 2025-10-21
.NET: Add DevUI package for .NET

ReubenBond:
### Motivation and Context

This adds a DevUI NuGet package for .NET and implements just enough API surface to support it. Some of what is implemented here (eg Conversations API) should be replaced by first-class implementations rather than what I've implemented in the DevUI project. HOWEVER, for this PR, I am happy with this hacked together implementation for now to promote forward progress. ~~I also fixed just enough of the Responses API implementation to make it work, but that will be replaced by #1550.~~

To try it out, run the DevUI_Step01_BasicUsage sample.

### Contribution Checklist

- [x] The code builds clean without any errors or warnings
- [x] The PR follows the [Contribution Guidelines](https://github.com/microsoft/agent-framework/blob/main/CONTRIBUTING.md)
- [x] All unit tests pass, and I have added new tests where possible
- [x] **Is this a breaking change?** If yes, add "[BREAKING]" prefix to the title of the PR.

PR #1731, opened 2025-10-27
.NET: Remove sequential\concurrent workflow extensions

DeagleGross:
Removing `AddConcurrentWorkflow` and `AddSequentialWorkflow` as per https://github.com/microsoft/agent-framework/pull/1603#discussion_r2453700479

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.Hosting.A2A.AspNetCore/EndpointRouteBuilderExtensions.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Hosting.A2A.AspNetCore/EndpointRouteBuilderExtensions.cs
@@ -22,45 +22,11 @@ public static class MicrosoftAgentAIHostingA2AEndpointRouteBuilderExtensions
     /// Attaches A2A (Agent2Agent) communication capabilities via Message processing to the specified web application.
     /// </summary>
     /// <param name="endpoints">The <see cref="IEndpointRouteBuilder"/> to add the A2A endpoints to.</param>
-    /// <param name="agentBuilder">The configuration builder for <see cref="AIAgent"/>.</param>
-    /// <param name="path">The route group to use for A2A endpoints.</param>
-    /// <returns>Configured <see cref="ITaskManager"/> for A2A integration.</returns>
-    /// <remarks>
-    /// This method can be used to access A2A agents that support the
-    /// <see href="https://github.com/a2aproject/A2A/blob/main/docs/topics/agent-discovery.md#2-curated-registries-catalog-based-discovery">Curated Registries (Catalog-Based Discovery)</see>
-    /// discovery mechanism.
-    /// </remarks>
-    public static IEndpointConventionBuilder MapA2A(this IEndpointRouteBuilder endpoints, IHostedAgentBuilder agentBuilder, string path)
-        => endpoints.MapA2A(agentBuilder, path, _ => { });
-
-    /// <summary>
-    /// Attaches A2A (Agent2Agent) communication capabilities via Message processing to the specified web application.
-    /// </summary>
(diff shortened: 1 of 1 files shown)


dotnet/samples/AgentWebChat/AgentWebChat.AgentHost/Program.cs
```diff
@@ -78,8 +78,20 @@ Once the user has deduced what type (knight or knave) both Alice and Bob are, te
     description: "An agent that helps with literature.",
     chatClientServiceKey: "chat-model");
 
-builder.AddSequentialWorkflow("science-sequential-workflow", [chemistryAgent, mathsAgent, literatureAgent]).AddAsAIAgent();
-builder.AddConcurrentWorkflow("science-concurrent-workflow", [chemistryAgent, mathsAgent, literatureAgent]).AddAsAIAgent();
+var scienceSequentialWorkflow = builder.AddWorkflow("science-sequential-workflow", (sp, key) =>
+{
+    List<IHostedAgentBuilder> usedAgents = [chemistryAgent, mathsAgent, literatureAgent];
+    var agents = usedAgents.Select(ab => sp.GetRequiredKeyedService<AIAgent>(ab.Name));
```
stephentoub, line 84: Why does this need to use GetRequiredKeyedService? It can't just use the AIAgent returned from AddAIAgent?

DeagleGross: `AddAIAgent` returns `IHostedAgentBuilder` which is not an actual `AIAgent` instance. It is a builder which has a name of registration + `IHostApplicationBuilder`. 

I took inspiration for this pattern from Aspire's "builders" API

glennc: The Aspire builders all have dependencies on DI though so you would never need to do this right? I assume the layering means we don't want to pass sp into BuildSequential to allow it to retrieve the agent so the extension methods on builder that this removes are the place we could combine both to make a convenience method that didn't require using the service provider. Why did we want to remove them?

stephentoub: > Why did we want to remove them?

Because there's an early and steep cliff associated with them. The only two methods of this ilk that were added were for Sequential and Concurrent, which are fine for demos and some limited scenarios but which represent only a very small portion of the kinds of patterns developers need to express; even with what's on AgentWorkflowBuilder, this didn't propagate handoffs or groupchat, nevermind everything you can expresson WorkflowBuilder, with custom edges and conditional edges and fan in/out and so on. I don't think we should try to propagate all of that surface area to live on these hosting builders, and should instead ensure it's easy to register any workflow.

dotnet/samples/AgentWebChat/AgentWebChat.AgentHost/Program.cs
```diff
@@ -78,8 +78,20 @@ Once the user has deduced what type (knight or knave) both Alice and Bob are, te
     description: "An agent that helps with literature.",
     chatClientServiceKey: "chat-model");
 
-builder.AddSequentialWorkflow("science-sequential-workflow", [chemistryAgent, mathsAgent, literatureAgent]).AddAsAIAgent();
-builder.AddConcurrentWorkflow("science-concurrent-workflow", [chemistryAgent, mathsAgent, literatureAgent]).AddAsAIAgent();
+var scienceSequentialWorkflow = builder.AddWorkflow("science-sequential-workflow", (sp, key) =>
+{
+    List<IHostedAgentBuilder> usedAgents = [chemistryAgent, mathsAgent, literatureAgent];
```
stephentoub, line 83: FWIW, it's still counterintuitive to me that something called "Agent" and that's returned from a method called "AddAIAgent" is actually a builder. I understand why it is, but maybe there are things we could do to ease this.

Issue #1684, opened 2025-10-24
.NET: Executor instances given to a WorkflowBuilder should not be usable in another WorkflowBuilder


lokitoth:
There is no way to track sharing / reset timing information between multiple workflows. We need to disallow use of Executor instances in multiple Workflows without proper ownership tracking.

(Diagnosing failure conditions here may require Workflows to have ids)

markwallace-microsoft:
@lokitoth tagging this as `v1.0` as it looks like a feature gap in .Net

lokitoth:
Not a feature gap per-se, but it would be nice to have this. Removing v1 since it is not critical and we can expect most people not to share executor instances between workflows.

PR #1756, opened 2025-10-28
Python: Ensure agent thread is part of checkpoint

### LINKED ISSUE #1698: Python: Public API to Save and Load ChatAgent State
#### **Summary**

It would be great to have a **public API for saving and loading the state of a `ChatAgent`**.

Currently, this can be done through `serialize` and `deserialize`, but a dedicated public API would make it much easier and more consistent to manage agent state within complex workflows.

---

#### **Use Case**

In a workflow, we might have **multiple ChatAgents** running concurrently or sequentially. If an API error occurs during execution, we can already rely on **checkpoints** to resume from the last running `AgentExecutor`.

However, what’s missing is the ability to **restore the ChatAgent’s state** itself — including the conversation context, chat history, and internal thread instance — so that it can continue seamlessly from where it left off.

A `save_state()` / `load_state()` (or similar) public API would make this workflow recovery much more robust and developer-friendly.

---

#### **Why This Is Useful**

* Enables **fine-grained recovery** inside workflows without manual serialization.
* Makes it easy to **persist conversational context** between runs or sessions.
* Simplifies handling of **partial failures** when multiple agents are involved.
* Aligns with checkpoint-based resumability in `AgentExecutor`.

---

#### **Proposed API (Example)**

```python
# Save agent state
state = chat_agent.save_state()

# Restore agent state later
chat_agent.load_state(state)
```

This would complement existing `serialize`/`deserialize` functionality while providing a clearer, officially supported interface for developers.

cc: @ekzhu , @moonbox3 , @victordibia

ekzhu:
### Motivation and Context

Resolves #1698 #1684

### Description

`AgentExecutor` currently doesn't snapshot the state of the agent thread, leading to agent "forgetting" previous conversation when restored from checkpoints.

This PR add mechanism to snapshot thread state. Added warning for `AzureAIAgent` which doesn't work with workflow checkpointing currently.

### Contribution Checklist

### DIFF
--- a/python/packages/core/agent_framework/_workflows/_agent_executor.py
+++ b/python/packages/core/agent_framework/_workflows/_agent_executor.py
@@ -86,21 +86,6 @@ class AgentExecutor(Executor):
         self._output_response = output_response
         self._cache: list[ChatMessage] = []
 
-        # Check if using AzureAIAgentClient and warn about checkpointing limitations
-        if isinstance(agent, ChatAgent):
-            client_class_name = agent.chat_client.__class__.__name__
-            client_module = agent.chat_client.__class__.__module__
-            if client_class_name == "AzureAIAgentClient" and "azure_ai" in client_module:
-                # TODO(ekzhu): update this warning when we surface the hooks for
-                # custom executor checkpointing.
-                logger.warning(
-                    "AgentExecutor is being used with AzureAIAgentClient which uses server-side threads. "
-                    "Currently, checkpointing does not capture messages from server-side threads. "
-                    "The thread state in checkpoints is not immutable and can be modified by subsequent runs. "
-                    "If you need reliable checkpointing with Azure AI agents, consider implementing a custom "
-                    "executor and managing the thread state yourself."
-                )
-
     @property
     def workflow_output_types(self) -> list[type[Any]]:
         # Override to declare AgentRunResponse as a possible output type only if enabled.
@@ -218,6 +203,24 @@ class AgentExecutor(Executor):
         Returns:
             Dict containing serialized cache and thread state
         """
+        # Check if using AzureAIAgentClient with server-side thread and warn about checkpointing limitations
+        if isinstance(self._agent, ChatAgent) and self._agent_thread.service_thread_id is not None:
+            client_class_name = self._agent.chat_client.__class__.__name__
+            client_module = self._agent.chat_client.__class__.__module__
+
+            if client_class_name == "AzureAIAgentClient" and "azure_ai" in client_module:
+                # TODO(TaoChenOSU): update this warning when we surface the hooks for
+                # custom executor checkpointing.
+                # https://github.com/microsoft/agent-framework/issues/1816
+                logger.warning(
+                    "Checkpointing an AgentExecutor with AzureAIAgentClient that uses server-side threads. "
+                    "Currently, checkpointing does not capture messages from server-side threads "
+                    "(service_thread_id: %s). The thread state in checkpoints is not immutable and can be "
+                    "modified by subsequent runs. If you need reliable checkpointing with Azure AI agents, "
+                    "consider implementing a custom executor and managing the thread state yourself.",
+                    self._agent_thread.service_thread_id,
+                )
+
         serialized_thread = await self._agent_thread.serialize()
 
         return {


markwallace-microsoft:
<a href="https://github.com/microsoft/agent-framework/blob/3f5bef8997722b0545a7c497b43be95bd270d940/README.md">Python Test Coverage</a><details><summary>Python Test Coverage Report • </summary><table><tr><th>File</th><th>Stmts</th><th>Miss</th><th>Cover</th><th>Missing</th></tr><tbody><tr><td colspan="5"><b>packages/core/agent_framework/_workflows</b></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/3f5bef8997722b0545a7c497b43be95bd270d940/packages/core/agent_framework/_workflows/_agent_executor.py">_agent_executor.py</a></td><td>106</td><td>26</td><td>75%</td><td><a href="https://github.com/microsoft/agent-framework/blob/3f5bef8997722b0545a7c497b43be95bd270d940/packages/core/agent_framework/_workflows/_agent_executor.py#L82">82</a>, <a href="https://github.com/microsoft/agent-framework/blob/3f5bef8997722b0545a7c497b43be95bd270d940/packages/core/agent_framework/_workflows/_agent_executor.py#L92-L94">92&ndash;94</a>, <a href="https://github.com/microsoft/agent-framework/blob/3f5bef8997722b0545a7c497b43be95bd270d940/packages/core/agent_framework/_workflows/_agent_executor.py#L113-L114">113&ndash;114</a>, <a href="https://github.com/microsoft/agent-framework/blob/3f5bef8997722b0545a7c497b43be95bd270d940/packages/core/agent_framework/_workflows/_agent_executor.py#L129">129</a>, <a href="https://github.com/microsoft/agent-framework/blob/3f5bef8997722b0545a7c497b43be95bd270d940/packages/core/agent_framework/_workflows/_agent_executor.py#L163-L164">163&ndash;164</a>, <a href="https://github.com/microsoft/agent-framework/blob/3f5bef8997722b0545a7c497b43be95bd270d940/packages/core/agent_framework/_workflows/_agent_executor.py#L166-L167">166&ndash;167</a>, <a href="https://github.com/microsoft/agent-framework/blob/3f5bef8997722b0545a7c497b43be95bd270d940/packages/core/agent_framework/_workflows/_agent_executor.py#L182-L183">182&ndash;183</a>, <a href="https://github.com/microsoft/agent-framework/blob/3f5bef8997722b0545a7c497b43be95bd270d940/packages/core/agent_framework/_workflows/_agent_executor.py#L208-L209">208&ndash;209</a>, <a href="https://github.com/microsoft/agent-framework/blob/3f5bef8997722b0545a7c497b43be95bd270d940/packages/core/agent_framework/_workflows/_agent_executor.py#L211">211</a>, <a href="https://github.com/microsoft/agent-framework/blob/3f5bef8997722b0545a7c497b43be95bd270d940/packages/core/agent_framework/_workflows/_agent_executor.py#L215">215</a>, <a href="https://github.com/microsoft/agent-framework/blob/3f5bef8997722b0545a7c497b43be95bd270d940/packages/core/agent_framework/_workflows/_agent_executor.py#L243-L245">243&ndash;245</a>, <a href="https://github.com/microsoft/agent-framework/blob/3f5bef8997722b0545a7c497b43be95bd270d940/packages/core/agent_framework/_workflows/_agent_executor.py#L255-L257">255&ndash;257</a>, <a href="https://github.com/microsoft/agent-framework/blob/3f5bef8997722b0545a7c497b43be95bd270d940/packages/core/agent_framework/_workflows/_agent_executor.py#L259">259</a>, <a href="https://github.com/microsoft/agent-framework/blob/3f5bef8997722b0545a7c497b43be95bd270d940/packages/core/agent_framework/_workflows/_agent_executor.py#L263-L264">263&ndash;264</a></td></tr><tr><td><b>TOTAL</b></td><td><b>11961</b></td><td><b>1852</b></td><td><b>84%</b></td><td>&nbsp;</td></tr></tbody></table></details>

## Python Unit Test Overview
| 1442 | 98 :zzz: | 0 :x: | 0 :fire: | 28.766s :stopwatch: |

python/packages/core/agent_framework/_workflows/_agent_executor.py
```diff
@@ -191,19 +191,51 @@ async def from_messages(
...
+            # Create a new local thread with message store (not server-side)
+            # This makes the checkpoint self-contained and independent of external services
+            message_store = ChatMessageStore()
+            new_thread = AgentThread(
+                message_store=message_store,
+                context_provider=self._agent_thread.context_provider,
+            )
+
+            # Copy messages from the current thread if it has a message store
+            if self._agent_thread.message_store is not None:
```
alliscode: Can it have a message store and a service_thread_id at the same time?

python/packages/core/agent_framework/_workflows/_agent_executor.py
```diff
@@ -191,19 +191,51 @@ async def from_messages(
...
         Returns:
-            Dict containing serialized cache state
+            Dict containing serialized cache and thread state
         """
+        from .._threads import ChatMessageStore
         from ._conversation_state import encode_chat_messages
 
+        serialized_thread: dict[str, Any]
+
+        # Handle server-side threads by creating a local copy with messages
```
alliscode: If we do need to snapshot server-side threads, that functionality should probably be in the thread impl. Some of them won't support it and some like Responses only need to save the last response Id.

python/packages/core/agent_framework/_workflows/_agent_executor.py
ekzhu: No... Looks like a bug. Now we are short of options for server-side thread.

eavanvalkenburg: by design, you can't have both a service_thread_id and a message_store.

eavanvalkenburg: and it also depends per service whether it is easy, or not to actually get the messages back, I don't think we have that abstracted. So I get the notion of fully localizing the checkpoint, but I think it also makes sense to just refer to the thread id instead, since there are non-technical reasons to not have all messages stored locally, like compliance, so I think it might make sense to specify that a checkpoint just serializes the thread and restarts with it, knowing that if you depend on service side threads, you need to ensure they are persisted as well, and the default behavior for all clients is to store locally, so probably for most cases just doing the thread serialize and deserialize and be done with it.

ekzhu: @eavanvalkenburg the issue is server-side thread implemented by Azure AI won't work with checkpointing: a checkpoint is considered immutable and should not be modified in subsequent runs. If the thread object only stores a reference to the server-side thread, subsequent runs will modify the server-side thread and corrupt the checkpoint.

ekzhu: Given the discussion, I think the conclusion should be that we document that the current implementation of Azure AI agent won't work with workflow checkpointing. We could fix this by special casing it however following potential compliance concern we may not want to do it. 

Server-side thread with Responses API should work fine.

python/packages/core/agent_framework/_workflows/_agent_executor.py
ekzhu: Make sense. I am removing the data copying logic.

python/packages/core/agent_framework/_workflows/_agent_executor.py
ekzhu: I updated the implementation and removed data copying. Added warning for AzureAIAgent.

python/packages/core/agent_framework/_workflows/_agent_executor.py
```diff
@@ -191,19 +192,43 @@ async def from_messages(
...
+                    "modified by subsequent runs. If you need reliable checkpointing with Azure AI agents, "
+                    "consider implementing a custom executor and managing the thread state yourself.",
+                    self._agent_thread.service_thread_id,
+                )
+
+        serialized_thread = await self._agent_thread.serialize()
 
         return {
             "cache": encode_chat_messages(self._cache),
+            "agent_thread": serialized_thread,
```
TaoChenOSU, line 228: nit: we should also make sure that use some reserved key for built-in modules

moonbox3 (approved):

python/packages/core/agent_framework/_workflows/_agent_executor.py
ekzhu: This key is local to just this executor right? So I think it shouldn't matter as other executors will be in different dictionaries

droideronline:
@ekzhu , @TaoChenOSU , @moonbox3 - I understand that after each super step, we now snapshot the full state of all executors — which includes serialized threads and therefore each agent’s conversation history.

Let’s consider a scenario where we’re running 4 agents. After the first agent completes, we take a snapshot and store the state of all agents in the checkpoint. Then the second agent begins execution. If, for any reason, the process is interrupted (for example, if the user presses Ctrl + C), we would still need to persist the current executor’s state so that when we resume, the agent can continue seamlessly with its previous conversation history intact.

Would it be possible to support this behavior as well? I think this would be an essential addition to ensure full resilience and continuity in multi-agent workflows.

Issue #1775, opened 2025-10-29
.NET: [AG-UI] Setup infrastructure and enable text streaming


javiercn:
**Infrastructure:** Implement the foundational infrastructure for AG-UI protocol support in the .NET Agent Framework, including:
- Two new packages: `Microsoft.Agents.AI.AGUI` for client-side consumption and `Microsoft.Agents.AI.Hosting.AGUI.AspNetCore` for server-side hosting
- Core event models for the AG-UI protocol with bidirectional conversion between AG-UI events and framework-native types (`AgentRunResponseUpdate`, `ChatMessage`)
- HTTP/SSE streaming support and thread management via `AGUIAgentThread`
- Server-side `MapAGUIAgent` extension method for exposing .NET agents as AG-UI endpoints
- Client-side `AGUIAgent` class for consuming any AG-UI-compliant server

**Text Streaming:** Implement Phase 1 text streaming capabilities with support for lifecycle events (`RunStarted`, `RunFinished`, `RunError`) and text message events (`TextMessageStart`, `TextMessageContent`, `TextMessageEnd`). Enable real-time streaming of conversational text between AG-UI clients and servers with proper Server-Sent Events (SSE) formatting. Include comprehensive testing (unit tests, integration tests for end-to-end scenarios) and a complete `AGUIClientServer` sample application demonstrating both client and server usage. Keep AG-UI protocol details internal while exposing clean public APIs aligned with the framework's abstractions.

PR #1776, opened 2025-10-29
.NET: AG-UI support for .NET

javiercn:
## Summary

This PR implements Phase 1 of AG-UI protocol support in the .NET Agent Framework, enabling text streaming capabilities for both consuming AG-UI-compliant servers and exposing .NET agents via the AG-UI protocol.

## Motivation and goals

The AG-UI (Agent-User Interaction) protocol is an open, lightweight, event-based protocol that standardizes how AI agents connect to user-facing applications. It addresses key challenges in agentic applications:

- **Streaming support**: Agents are long-running and stream intermediate work across multi-turn sessions
- **Event-driven architecture**: Supports nondeterministic agent behavior with real-time UI updates
- **Protocol interoperability**: Complements MCP (tool/context) and A2A (agent-to-agent) protocols in the AI ecosystem

Without AG-UI support, .NET agents cannot interoperate with the growing ecosystem of AG-UI-compatible frontends and agent frameworks (LangGraph, CrewAI, Pydantic AI, etc.).

This implementation enables:
- .NET developers to consume AG-UI servers from any framework
- .NET agents to be accessible from any AG-UI-compatible client (web, mobile, CLI)
- Standardized streaming communication patterns for agentic applications

## In scope

1. **Client-side AG-UI consumption** (`Microsoft.Agents.AI.AGUI` package)
   - `AGUIAgent` class for connecting to remote AG-UI servers
   - `AGUIAgentThread` for managing conversation threads
   - HTTP/SSE streaming support
   - Event-to-framework type conversion

2. **Server-side AG-UI hosting** (`Microsoft.Agents.AI.Hosting.AGUI.AspNetCore` package)
   - `MapAGUIAgent` extension method for ASP.NET Core
   - Server-Sent Events (SSE) response formatting using `SseFormatter`
   - Framework-to-event type conversion
   - Agent factory pattern for per-request agent instantiation

3. **Text streaming events** (Phase 1)
   - Lifecycle events: `RunStarted`, `RunFinished`, `RunError`
   - Text message events: `TextMessageStart`, `TextMessageContent`, `TextMessageEnd`
   - Bidirectional event/update conversion
   - Thread and run ID management via `ConversationId` and `ResponseId` properties

4. **Testing and samples**
   - Unit tests for the core protocol
   - Integration tests for end-to-end scenarios
   - AGUIClientServer sample demonstrating both client and server usage

## Out of scope

The following AG-UI features are intentionally deferred to future PRs:

- Tool call events (`ToolCallStart`, `ToolCallArgs`, `ToolCallEnd`, `ToolCallResult`)
- State management events (`StateSnapshot`, `StateDelta`, `MessagesSnapshot`)
- Step events (`StepStarted`, `StepFinished`)

### Server: Exposing a .NET agent via AG-UI

```csharp
var builder = WebApplication.CreateBuilder(args);
var app = builder.Build();

// Map an AG-UI endpoint at /agent
app.MapAGUIAgent("/", (messages) =>
{
    var chatClient = new AzureOpenAIClient(endpoint, credential)
        .GetChatClient(deploymentName);
    return chatClient.CreateAIAgent(name: "AGUIAssistant");
});

app.Run();
```

### Client: Consuming an AG-UI server

```csharp
using var httpClient = new HttpClient();
var agent = new AGUIAgent(
    id: "agui-client",
    description: "AG-UI Client Agent",
    httpClient: httpClient,
    endpoint: "http://localhost:5100");

var thread = agent.GetNewThread();
var messages = new List<ChatMessage>
{
    new(ChatRole.System, "You are a helpful assistant."),
    new(ChatRole.User, "What is the capital of France?")
};

// Stream responses in real-time
await foreach (var update in agent.RunStreamingAsync(messages, thread))
{
    ChatResponseUpdate chatUpdate = update.AsChatResponseUpdate();
    
    // Access thread and run IDs from first update
    if (chatUpdate.ConversationId != null)
    {
        Console.WriteLine($"Thread: {chatUpdate.ConversationId}, Run: {chatUpdate.ResponseId}");
    }
    
    foreach (var content in update.Contents)
    {
        if (content is TextContent text)
        {
            Console.Write(text.Text); // Display streaming text
        }
        else if (content is ErrorContent error)
        {
            Console.WriteLine($"Error: {error.Message}");
        }
    }
}
```

### SSE Event Stream Format

When a client sends a request:
```json
POST /agent
{
  "threadId": "thread_123",
  "runId": "run_456",
  "messages": [{"role": "user", "content": "Hello"}]
}
```

The server streams back SSE events:
```
data: {"type":"run_started","threadId":"thread_123","runId":"run_456"}

data: {"type":"text_message_start","messageId":"msg_1","role":"assistant"}

data: {"type":"text_message_content","messageId":"msg_1","delta":"Hello"}

data: {"type":"text_message_content","messageId":"msg_1","delta":" there!"}

data: {"type":"text_message_end","messageId":"msg_1"}

data: {"type":"run_finished","threadId":"thread_123","runId":"run_456"}
```

---

# Detailed design

### Package Structure

```
Microsoft.Agents.AI.AGUI/
├── AGUIAgent.cs                          // Public: Client agent implementation
├── AGUIAgentThread.cs                    // Public: Thread management
├── AGUIHttpService.cs                    // Internal: HTTP/SSE communication
└── Shared/                               // Internal: Protocol types
    ├── BaseEvent.cs
    ├── RunStartedEvent.cs
    ├── TextMessageStartEvent.cs
    ├── AGUIMessage.cs
    ├── RunAgentInput.cs
    └── [conversion extensions]

Microsoft.Agents.AI.Hosting.AGUI.AspNetCore/
├── AGUIEndpointRouteBuilderExtensions.cs // Public: MapAGUIAgent
├── AGUIServerSentEventsResult.cs         // Internal: SSE formatting
└── Shared/                               // Internal: Protocol types & converters
```

### Key Design Decisions

**1. Event Models as Internal Types**
- AG-UI event types (e.g., `RunStartedEvent`, `TextMessageContentEvent`) are internal
- Conversion happens at boundaries via extension methods
- Public API uses framework-native types (`AgentRunResponseUpdate`, `ChatMessage`)
- **Rationale**: Protects consumers from protocol changes; maintains framework abstractions

**2. No Custom Content Types**
- Run lifecycle communicated through `ChatResponseUpdate.ConversationId` and `ResponseId` properties
- Errors use standard `ErrorContent` type
- First update in a stream contains `ConversationId` and `ResponseId` indicating run started
- Last update with `FinishReason` indicates run finished
- **Rationale**: Avoids introducing non-standard content types; uses existing framework abstractions

**3. Agent Factory Pattern in MapAGUIAgent**
- Factory function creates agent per request: `(messages) => AIAgent`
- Allows request-specific agent configuration (e.g., different models per user)
- **Rationale**: Supports multi-tenancy and request customization without global state

**4. Bidirectional Conversion Architecture**
- Server: `AgentRunResponseUpdate` → AG-UI events
- Client: AG-UI events → `AgentRunResponseUpdate`
- Extension methods in `Shared/` namespace (compiled in both packages)
- **Rationale**: Symmetric conversion logic; shared code between client and server

**5. Thread Management**
- `AGUIAgentThread` stores only `ThreadId` (string)
- Thread ID communicated via `ConversationId` property on updates
- No automatic thread persistence; applications manage storage
- **Rationale**: AG-UI spec defines threads by ID only; persistence is application-specific

**6. SSE Formatting**
- Uses ASP.NET Core's `SseFormatter` for Server-Sent Events
- Standard SSE format with `data:` prefix and double newline delimiter
- **Rationale**: Leverages framework infrastructure for SSE compliance

### Event Conversion Logic

**Text Message Streaming (Server → Client):**

```csharp
// Server side: Framework updates → AG-UI events
IAsyncEnumerable<AgentRunResponseUpdate> updates = agent.RunStreamingAsync(...);
IAsyncEnumerable<BaseEvent> events = updates.AsAGUIEventStreamAsync(threadId, runId);

// Emits: RunStarted
//        TextMessageStart (on role/messageId change)
//        TextMessageContent (per text chunk)
//        TextMessageEnd (on message completion)
//        RunFinished
```

**Text Message Streaming (Client → Framework):**

```csharp
// Client side: AG-UI events → Framework updates
IAsyncEnumerable<BaseEvent> events = httpService.PostRunAsync(...);
IAsyncEnumerable<AgentRunResponseUpdate> updates = events.AsAgentRunResponseUpdatesAsync();

// Converts: RunStarted → Update with ConversationId/ResponseId
//           TextMessageStart → (tracks current message)
//           TextMessageContent → ChatResponseUpdate with text delta
//           TextMessageEnd → (closes message)
//           RunFinished → Update with FinishReason
//           RunError → Update with ErrorContent
```

### JSON Serialization

- Uses `System.Text.Json.Serialization.JsonConverter` for polymorphic events
- `BaseEventJsonConverter` dispatches based on `"type"` field
- Source-generated serialization context (`AGUIJsonSerializerContext`) for AOT compatibility
- **Rationale**: Built-in STJ polymorphism requires discriminator as first property, which AG-UI protocol doesn't guarantee

### Error Handling

- HTTP errors (4xx, 5xx) throw `HttpRequestException`
- AG-UI `RunError` events convert to `ErrorContent` in update stream
- SSE parsing errors throw `JsonException` or `InvalidOperationException`
- Connection failures propagate as `OperationCanceledException` when cancelled
- **Rationale**: Standard .NET exception patterns; errors surface naturally through async streams

## Drawbacks

1. **Custom JSON Converter**: Required custom polymorphic deserialization instead of built-in STJ support
   - *Mitigation*: Internal implementation detail; doesn't affect public API

2. **Shared Code via Preprocessor Directives**: Files in `Shared/` use `#if ASPNETCORE` to compile into both packages
   - *Mitigation*: Necessary to avoid duplicating event definitions; contained to internal types

### Alternative 1: Expose AG-UI Event Types Publicly
**Rejected**: We already have common abstractions for AI concepts. We can break AG-UI specific payloads into a separate assembly in the future and take a dependency on that if needed.

### Alternative 2: Implement All Event Types Upfront
**Rejected**: Keeping the PR manageable and focused on text streaming. Tool calls and state management can be added incrementally.

### Alternative 3: Custom AIContent Types for Lifecycle
**Rejected after review**: Initially proposed `RunStartedContent`, `RunFinishedContent`, and `RunErrorContent`. Feedback indicated these should use existing framework properties (`ConversationId`, `ResponseId`, `ErrorContent`) instead.

#Fixes https://github.com/microsoft/agent-framework/issues/1775

### DIFF
--- a/dotnet/agent-framework-dotnet.slnx
+++ b/dotnet/agent-framework-dotnet.slnx
@@ -277,6 +277,7 @@
     <Project Path="src/Microsoft.Agents.AI.A2A/Microsoft.Agents.AI.A2A.csproj" />
     <Project Path="src/Microsoft.Agents.AI.Abstractions/Microsoft.Agents.AI.Abstractions.csproj" />
     <Project Path="src/Microsoft.Agents.AI.AGUI/Microsoft.Agents.AI.AGUI.csproj" />
+    <Project Path="src/Microsoft.Agents.AI.AzureAI.Persistent/Microsoft.Agents.AI.AzureAI.Persistent.csproj" />
     <Project Path="src/Microsoft.Agents.AI.CopilotStudio/Microsoft.Agents.AI.CopilotStudio.csproj" />
     <Project Path="src/Microsoft.Agents.AI.Hosting.A2A.AspNetCore/Microsoft.Agents.AI.Hosting.A2A.AspNetCore.csproj" />
     <Project Path="src/Microsoft.Agents.AI.Hosting.A2A/Microsoft.Agents.AI.Hosting.A2A.csproj" />


dotnet/samples/AGUIClientServer/README.md
```diff
@@ -0,0 +1,190 @@
...
+> **Warning**
+> The AG-UI protocol is still under development and changing.
+> We will try to keep these samples updated as the protocol evolves.
+
+## Configuring Environment Variables
+
+Configure the required Azure OpenAI environment variables:
+
+```powershell
+$env:AZURE_OPENAI_ENDPOINT="https://ag-ui-agent-framework.openai.azure.com/"
```
javiercn: Replace this with `<<your-model-endpoint>>`

dotnet/samples/AGUIClientServer/README.md
```diff
@@ -0,0 +1,190 @@
...
+- **RunFinishedContent**: Signals successful completion of the agent run
+- **RunErrorContent**: Signals an error occurred during processing
+
+## Next Steps
+
+- Explore adding tools/functions to the agent
+- Implement conversation history management
+- Add authentication and authorization
+- Deploy to production environments
+- Integrate with web applications using the AG-UI protocol
```
javiercn: Remove this section

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIAgentThread.cs
```diff
@@ -0,0 +1,76 @@
...
+using System;
+using System.Text.Json;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Agents.AI.AGUI;
+
+/// <summary>
+/// Provides an <see cref="AgentThread"/> implementation for AG-UI agents.
+/// </summary>
+internal sealed class AGUIAgentThread : InMemoryAgentThread
```
javiercn, line 9: Remove XML doc comments from this type as it's internal, from its members too

dotnet/src/Microsoft.Agents.AI.AGUI/Shared/AgentRunResponseUpdateAGUIExtensions.cs
```diff
@@ -0,0 +1,140 @@
...
+internal static class AgentRunResponseUpdateAGUIExtensions
+{
+#if !ASPNETCORE
+    public static async IAsyncEnumerable<AgentRunResponseUpdate> AsAgentRunResponseUpdatesAsync(
+        this IAsyncEnumerable<BaseEvent> events,
+        [EnumeratorCancellation] CancellationToken cancellationToken = default)
+    {
+        string? currentMessageId = null;
+        ChatRole currentRole = default!;
+        await foreach (var evt in events.ConfigureAwait(false))
```
javiercn: Pass in the CT

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIAgent.cs
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        string runId = Guid.NewGuid().ToString();
+
+        RunAgentInput input = new()
+        {
+            ThreadId = typedThread.ThreadId,
+            RunId = runId,
+            Messages = messages.AsAGUIMessages(),
+            Context = new Dictionary<string, string>(StringComparer.Ordinal)
```
stephentoub: What is Context? It needs to be set to a new dictionary even if it's never used?

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIHttpService.cs
```diff
@@ -0,0 +1,57 @@
...
+
+    private static BaseEvent ItemParser(string type, ReadOnlySpan<byte> data)
+    {
+        var result = JsonSerializer.Deserialize(data, AGUIJsonSerializerContext.Default.BaseEvent);
+        if (result != null)
+        {
+            return result;
+        }
+
+        throw new InvalidOperationException("Failed to deserialize SSE item.");
```
stephentoub: Nit:
```suggestion
        return JsonSerializer.Deserialize(data, AGUIJsonSerializerContext.Default.BaseEvent) ??
            throw new InvalidOperationException("Failed to deserialize SSE item.");
```

dotnet/src/Microsoft.Agents.AI.AGUI/RunErrorContent.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using Microsoft.Extensions.AI;
+
+namespace Microsoft.Agents.AI.AGUI;
+
+/// <summary>
+/// Represents content indicating that an agent run encountered an error.
+/// </summary>
+public sealed class RunErrorContent : AIContent
```
stephentoub: Can this not use the built-in ErrorContent?

dotnet/src/Microsoft.Agents.AI.AGUI/RunStartedContent.cs
```diff
@@ -0,0 +1,32 @@
...
+
+    /// <summary>
+    /// Gets the ID of the conversation thread.
+    /// </summary>
+    public string ThreadId { get; }
+
+    /// <summary>
+    /// Gets the ID of the agent run.
+    /// </summary>
+    public string RunId { get; }
```
stephentoub: Does this actually need an AIContent-derived type to communicate this, or could it be modeled as ChatResponse{Update}.ConversationId and ResponseId?

dotnet/src/Microsoft.Agents.AI.AGUI/RunFinishedContent.cs
```diff
@@ -0,0 +1,39 @@
...
+
+    /// <summary>
+    /// Gets the ID of the agent run.
+    /// </summary>
+    public string RunId { get; }
+
+    /// <summary>
+    /// Gets optional result data from the run.
+    /// </summary>
+    public string? Result { get; }
```
stephentoub: This can't just be TextContent?

dotnet/src/Microsoft.Agents.AI.AGUI/Shared/AGUIChatMessageExtensions.cs
```diff
@@ -0,0 +1,72 @@
...
+
+            result.Add(chatMessage);
+        }
+        return result ?? [];
+    }
+
+    public static IEnumerable<AGUIMessage> AsAGUIMessages(
+        this IEnumerable<ChatMessage> chatMessages)
+    {
+        List<AGUIMessage>? result = null;
```
stephentoub: Why does this buffer everything up rather than implementing it as yield return?

dotnet/src/Microsoft.Agents.AI.AGUI/Shared/AGUIChatMessageExtensions.cs
```diff
@@ -0,0 +1,72 @@
...
+                Id = message.MessageId,
+                Role = message.Role.Value,
+                Content = message.Text,
+            };
+            result.Add(aguiMessage);
+        }
+        return result ?? [];
+    }
+
+    public static ChatRole MapChatRole(string role)
```
stephentoub: Could this just be `=> new(role);`?

dotnet/src/Microsoft.Agents.AI.AGUI/Shared/AGUIRoles.cs
```diff
@@ -0,0 +1,18 @@
...
+
+internal static class AGUIRoles
+{
+    public const string System = "system";
+
+    public const string User = "user";
+
+    public const string Assistant = "assistant";
+
+    public const string Developer = "developer";
```
stephentoub, line 17: What distinction does AG-UI make between "system" and "developer"?

dotnet/src/Microsoft.Agents.AI.AGUI/Shared/AgentRunResponseUpdateAGUIExtensions.cs
```diff
@@ -0,0 +1,140 @@
...
+        {
+            switch (evt)
+            {
+                case RunStartedEvent runStarted:
+                    yield return new AgentRunResponseUpdate(new ChatResponseUpdate(
```
stephentoub, line 34: Can / should we populate more of the properties on these updates? e.g. CreatedAt, Conversation/Response/MessageId?

dotnet/src/Microsoft.Agents.AI.AGUI/RunStartedContent.cs
```diff
@@ -0,0 +1,32 @@
...
+}
```
stephentoub: I'm surprised you're not running into serialization issues with these AIContent-derived types, e.g. I'd expect doing serialization on a ChatMessage containing these will fail, because they're not connected to the base type's polymorphism infrastructure. Whatever JsonSerializerOptions is used would likely need to use the AddAIContentType method to augment the resolver in that options. It's doable, but it also adds complication, and is one of the reason it'd be nice to avoid needing custom AIContent types, if possible.

dotnet/src/Microsoft.Agents.AI.AGUI/Shared/BaseEventJsonConverter.cs
```diff
@@ -0,0 +1,103 @@
...
+namespace Microsoft.Agents.AI.Hosting.AGUI.AspNetCore.Shared;
+#else
+namespace Microsoft.Agents.AI.AGUI.Shared;
+#endif
+
+/// <summary>
+/// Custom JSON converter for polymorphic deserialization of BaseEvent and its derived types.
+/// Uses the "type" property as a discriminator to determine the concrete type to deserialize.
+/// </summary>
+internal sealed class BaseEventJsonConverter : JsonConverter<BaseEvent>
```
stephentoub, line 17: It can't use the built-in polymorphism support in STJ?

dotnet/src/Microsoft.Agents.AI.AGUI/Shared/RunAgentInput.cs
```diff
@@ -0,0 +1,42 @@
...
+    [Required]
+    [JsonPropertyName("runId")]
+    public string RunId { get; set; } = string.Empty;
+
+    [JsonPropertyName("state")]
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingDefault)]
+    public JsonElement State { get; set; }
+
+    [Required]
+#pragma warning disable IL2026 // MinLengthAttribute on arrays is trim-safe
```
stephentoub: I'm missing why this is ok? This also isn't an array per-se, but an `IEnumerable<AGUIMessage>`

dotnet/src/Microsoft.Agents.AI.AGUI/Shared/RunErrorEvent.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Text.Json.Serialization;
+
+#if ASPNETCORE
+namespace Microsoft.Agents.AI.Hosting.AGUI.AspNetCore.Shared;
+#else
+namespace Microsoft.Agents.AI.AGUI.Shared;
+#endif
```
stephentoub, line 9: Why do we need to change this internal namespace based on what project it's being built into?

dotnet/src/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore/AGUIServerSentEventsResult.cs
```diff
@@ -0,0 +1,81 @@
...
+            await foreach (var item in this._events.ConfigureAwait(false))
+            {
+                await body.WriteAsync(s_data, cancellationToken).ConfigureAwait(false);
+                await JsonSerializer.SerializeAsync(
+                    body,
+                    item,
+                    AGUIJsonSerializerContext.Default.BaseEvent, cancellationToken).ConfigureAwait(false);
+                await body.WriteAsync(s_newLines, cancellationToken).ConfigureAwait(false);
+                await body.FlushAsync(cancellationToken).ConfigureAwait(false);
+            }
```
stephentoub: Could this use SseFormatter?

dotnet/src/Microsoft.Agents.AI.AGUI/RunErrorContent.cs
javiercn: Yes, good catch.

dotnet/src/Microsoft.Agents.AI.AGUI/Shared/BaseEventJsonConverter.cs
javiercn: The built-in polymorphism support requires the discriminator to be the first element in the payload, which we can't guarantee for clients that we don't own.

I started with that approach and switched to this approach because the copilotkit typescript client was not working that way.

dotnet/src/Microsoft.Agents.AI.AGUI/Shared/RunErrorEvent.cs
javiercn: This is because the test projects have IVT to different assemblies, and otherwise it creates conflicts when you import the namespace.

markwallace-microsoft:
@javiercn can you copy the information in the PR description to an ADR and add to this folder: https://github.com/microsoft/agent-framework/tree/main/docs/decisions

It will be much easier to find the information later and it will be with all of the other design decisions.

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIAgent.cs
```diff
@@ -0,0 +1,102 @@
...
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="AGUIAgent"/> class.
+    /// </summary>
+    /// <param name="id">The agent ID.</param>
+    /// <param name="description">Optional description of the agent.</param>
+    /// <param name="messages">Initial conversation messages.</param>
+    /// <param name="httpClient">The HTTP client to use for communication with the AG-UI server.</param>
+    /// <param name="endpoint">The URL for the AG-UI server.</param>
+    public AGUIAgent(string id, string description, IEnumerable<ChatMessage> messages, HttpClient httpClient, string endpoint)
```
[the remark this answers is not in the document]
javiercn: Whoops, we should attach the initial messages when creating a new thread

dotnet/src/Microsoft.Agents.AI.AGUI/Shared/AGUIChatMessageExtensions.cs
javiercn: I switched to yield

dotnet/src/Microsoft.Agents.AI.AGUI/Shared/AGUIRoles.cs
javiercn: Yep https://docs.ag-ui.com/sdk/js/core/types#role

I believe foundry agents also recognize developer and map it instructions https://github.com/Azure/azure-sdk-for-net/blob/1715553938a0d926e5ae5356eece00d58d981348/sdk/ai/Azure.AI.Agents.Persistent/src/Custom/PersistentAgentsChatClient.cs#L526-L527

dotnet/src/Microsoft.Agents.AI.AGUI/Shared/AgentRunResponseUpdateAGUIExtensions.cs
javiercn: Is it fair  to map ConversationId = ThreadId and ResponseId = RunId?

dotnet/src/Microsoft.Agents.AI.AGUI/RunFinishedContent.cs
javiercn: https://docs.ag-ui.com/sdk/js/core/events#runfinishedevent

Result can be any structured data. I think it is fair to map it to a separate message + some TextContent, but what's the best way to represent this when the data is structured? Ideally we would want some JsonContent or similar, wouldn't we? For now I'll just map it to a string

dotnet/src/Microsoft.Agents.AI.AGUI/RunStartedContent.cs
javiercn: I got rid of this in favor of an empty update

javiercn: This info is captured in ConversationId, and ResponseId

dotnet/src/Microsoft.Agents.AI.AGUI/Shared/AgentRunResponseUpdateAGUIExtensions.cs
javiercn: I've populated the response ID and conversation ID as well as the message ID for the text updates. For the RunStarted/Finished events there isn't a message ID.

dotnet/tests/Microsoft.Agents.AI.AGUI.UnitTests/AGUIAgentTests.cs
```diff
@@ -0,0 +1,337 @@
...
+            .Protected()
+            .Setup<Task<HttpResponseMessage>>(
+                "SendAsync",
+                ItExpr.IsAny<HttpRequestMessage>(),
+                ItExpr.IsAny<CancellationToken>())
+            .ReturnsAsync(new HttpResponseMessage
+            {
+                StatusCode = HttpStatusCode.OK,
+                Content = new StringContent(sseContent)
+            });
```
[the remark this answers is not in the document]
javiercn: It's fine for a test.

dotnet/src/Microsoft.Agents.AI.AGUI/Shared/AGUIChatMessageExtensions.cs
javiercn: Do we care about the string being in the "canonical" form? https://docs.ag-ui.com/sdk/js/core/types#role

dotnet/src/Microsoft.Agents.AI.AGUI/Shared/RunAgentInput.cs
javiercn: Just removed this, didn't realize copilot left it

dotnet/src/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore/AGUIServerSentEventsResult.cs
javiercn: Fixed this. Copilot was fighting me and kept undoing the change. I realized it was missing the 10.0-rc2 version of the SSE library.

javiercn (commented):
Addressed feedback:
* I've gotten rid of the custom contents and mapped the info from the events as ResponseUpdate.
* I switched to use the SseFormatter instead of the manual SSE Event serialization.
* I've gotten rid of Context/Tools/ForwardedProps for now and will bring them in a separate PR.
* I've mapped additional properties to response updates.
* I removed some buffering when dealing with messages.
* I've fixed the build dependencies and cleaned up some copilot leftover.

dotnet/src/Microsoft.Agents.AI.AGUI/RunFinishedContent.cs
stephentoub: > Ideally we would want some JsonContent or similar, wouldn't we? For now I'll just map it to a string

Mapping to a string sounds fine for now.

But you could also map it to a DataContent with a MediaType of "application/json".

dotnet/src/Microsoft.Agents.AI.AGUI/RunErrorContent.cs
javiercn: Replaced with ErrorContent

dotnet/src/Microsoft.Agents.AI.AGUI/RunStartedContent.cs
javiercn: Got rid of all the custom content types, things are mapped to response updates. I think this is fine for now, but I want to align later on in what the best ways are for us representing some of this information

javiercn:
> @javiercn can you copy the information in the PR description to an ADR and add to this folder: https://github.com/microsoft/agent-framework/tree/main/docs/decisions
> 
> It will be much easier to find the information later and it will be with all of the other design decisions.

Yep, I saw the ADRs, and was planning to do that, but wanted to check with you first. I'll include the ADRs as a last step as we work out through the details

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIAgent.cs
javiercn: I've taken this out from the current PR. Context is data that the client can pass to the agent. I'll deal with it in a separate PR so that we can better discuss how to map it.

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIAgent.cs
```diff
@@ -0,0 +1,102 @@
...
+using Microsoft.Agents.AI.AGUI.Shared;
+using Microsoft.Extensions.AI;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Agents.AI.AGUI;
+
+/// <summary>
+/// Provides an <see cref="AIAgent"/> implementation that communicates with an AG-UI compliant server.
+/// </summary>
+public sealed class AGUIAgent : AIAgent
```
javiercn, line 20: I'm going to make this implement IChatClient instead in the next PR, as it will allow me to layer tool calling on top.

dotnet/src/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore/AGUIEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,71 @@
...
+            }
+            catch (JsonException)
+            {
+                await TypedResults.BadRequest().ExecuteAsync(context).ConfigureAwait(false);
+                return;
+            }
+
+            if (input is null)
+            {
+                await TypedResults.BadRequest().ExecuteAsync(context).ConfigureAwait(false);
```
ReubenBond: I would expect this to be `return TypedResults.BadRequest();`, and similar for the others. Why perform the execution inline?

javiercn: I'm passing in a `RequestDelegate`. I can't use return `TypedResults.BadRequests()` because otherwise I get AOT warnings.

dotnet/src/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore/AGUIEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,71 @@
...
+    /// Maps an AG-UI agent endpoint.
+    /// </summary>
+    /// <param name="endpoints">The endpoint route builder.</param>
+    /// <param name="pattern">The URL pattern for the endpoint.</param>
+    /// <param name="agentFactory">Factory function to create an agent instance.</param>
+    /// <returns>An <see cref="IEndpointConventionBuilder"/> for the mapped endpoint.</returns>
+    public static IEndpointConventionBuilder MapAGUIAgent(
+        this IEndpointRouteBuilder endpoints,
+        string pattern,
+        Func<IEnumerable<ChatMessage>, AIAgent> agentFactory)
```
DeagleGross: This extension follows different pattern than existing today's OpenAI Responses and A2A ones. See https://github.com/microsoft/agent-framework/blob/50d9b13bfc66573ac1d19e7ba25b0b4d7f8247a1/dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/EndpointRouteBuilderExtensions.Responses.cs#L35.

I think we should not be exposing `Func<IEnumerable<ChatMessage>, AIAgent> agentFactory` and if people need to create custom mapping, they still can do.

dotnet/src/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore/AGUIEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,71 @@
...
+    {
+        return endpoints.MapPost(pattern, async context =>
+        {
+            var cancellationToken = context.RequestAborted;
+
+            RunAgentInput? input;
+            try
+            {
+                input = await JsonSerializer.DeserializeAsync(context.Request.Body, AGUIJsonSerializerContext.Default.RunAgentInput, cancellationToken).ConfigureAwait(false);
```
DeagleGross: ideally, that should just be a `[FromBody] RunAgentInput input`. Any blockers on rewriting handler in this way?

dotnet/src/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore/AGUIEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,71 @@
...
+                await TypedResults.BadRequest().ExecuteAsync(context).ConfigureAwait(false);
+                return;
+            }
+
+            var messages = input.Messages.AsChatMessages();
+            var contextValues = input.Context;
+            var forwardedProps = input.ForwardedProperties;
+            var agent = agentFactory(messages);
+
+            var events = agent.RunStreamingAsync(
```
DeagleGross, line 43: i assume AG-UI only supports streaming of events, that's why we use `RunStreamingAsync` only here, right?

dotnet/src/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore/AGUIEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,71 @@
...
+            if (input is null)
+            {
+                await TypedResults.BadRequest().ExecuteAsync(context).ConfigureAwait(false);
+                return;
+            }
+
+            var messages = input.Messages.AsChatMessages();
+            var contextValues = input.Context;
+            var forwardedProps = input.ForwardedProperties;
+            var agent = agentFactory(messages);
```
DeagleGross: I think this only adds entropy to the implementation; we definitely should change the method signature

dotnet/src/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore/AGUIEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,71 @@
...
+            var events = agent.RunStreamingAsync(
+                messages,
+                cancellationToken: cancellationToken)
+                .AsAGUIEventStreamAsync(
+                    input.ThreadId,
+                    input.RunId,
+                    cancellationToken);
+
+            var logger = context.RequestServices.GetRequiredService<ILogger<AGUIServerSentEventsResult>>();
+            await new AGUIServerSentEventsResult(events, logger).ExecuteAsync(context).ConfigureAwait(false);
```
DeagleGross: if `AGUIServerSentEventsResult` is disposable, we should dispose it here

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIAgentThread.cs
```diff
@@ -0,0 +1,61 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+using System.Text.Json;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Agents.AI.AGUI;
+
+internal sealed class AGUIAgentThread : InMemoryAgentThread
```
DeagleGross, line 9: why is `AGUIAgentThread` based on `InMemoryAgentThread` particularly? Can't it be base `AgentThread`? I am thinking we should not restrict type as much beforehand

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIAgent.cs
```diff
@@ -0,0 +1,102 @@
...
+    {
+        this.Id = Throw.IfNullOrWhitespace(id);
+        this.Description = description;
+        this._client = new AGUIHttpService(
+            httpClient ?? Throw.IfNull(httpClient),
+            endpoint ?? Throw.IfNullOrEmpty(endpoint));
+    }
+
+    /// <inheritdoc/>
+    public override string Id { get; }
```
DeagleGross, line 41: why do you need override for `Id` / `Description`?

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIAgent.cs
```diff
@@ -0,0 +1,102 @@
...
+        new AGUIAgentThread(serializedThread, jsonSerializerOptions);
+
+    /// <inheritdoc/>
+    public override async Task<AgentRunResponse> RunAsync(
+        IEnumerable<ChatMessage> messages,
+        AgentThread? thread = null,
+        AgentRunOptions? options = null,
+        CancellationToken cancellationToken = default)
+    {
+        return await this.RunStreamingAsync(messages, thread, null, cancellationToken)
```
DeagleGross, line 60: not sure: maybe its better to explicitly throw here, saying "go use streaming version" because that is what protocol is about?

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIAgent.cs
```diff
@@ -0,0 +1,102 @@
...
+        List<ChatResponseUpdate> updates = [];
+
+        _ = Throw.IfNull(messages);
+
+        if ((thread ?? this.GetNewThread()) is not AGUIAgentThread typedThread)
+        {
+            throw new InvalidOperationException("The provided thread is not compatible with the agent. Only threads created by the agent can be used.");
+        }
+
+        string runId = Guid.NewGuid().ToString();
```
DeagleGross: should we try to enforce some convention here for our own convenience? something like `agui_uoierjg8934hg3`

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIAgent.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        var llmMessages = typedThread.MessageStore.Concat(messages);
```
DeagleGross, line 83: If you are basing the type on top of `InMemoryAgentThread` in order to obtain some kind of message store, you can ignore it for now. I dont think there is any restriction to have a message store for AG-UI and we can solve it differently from the hosting perspective.

basically if you need to pass more data for now - you could do it by passing more `messages` into the `RunAsync` or `RunStreamingAsync`

dotnet/src/Microsoft.Agents.AI.AGUI/Shared/AGUIEventTypes.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+#if ASPNETCORE
+namespace Microsoft.Agents.AI.Hosting.AGUI.AspNetCore.Shared;
```
DeagleGross, line 4: why do we choose namespace here?

dotnet/src/Microsoft.Agents.AI.AGUI/Shared/RunErrorEvent.cs
DeagleGross: can you elaborate further please? I am very confused by a lot of #if #defs on namespaces / behavior in AI.AGUI/Shared folder

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIAgentThread.cs
javiercn: I'm getting rid of this on the next PR, AGUIAgent -> AGUIChatClient and then we just use ChatClientAgent

dotnet/src/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore/AGUIEndpointRouteBuilderExtensions.cs
javiercn: Yep, the whole point is that it "standardizes" the updates from agents into a "standard" set of events that come back as part of the response. Hence you job is to translate the original AgentRunResponseUpdate instances into the relevant AG-UI events.

dotnet/src/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore/AGUIEndpointRouteBuilderExtensions.cs
javiercn: Yep, I get trimming warnings. Since this is a library, I don't know 100% if we can SG an intercepted method. It's the same reason why I don't use `return TypedResults.BadRequest`

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIAgent.cs
javiercn: I'm getting rid of this code in `https://github.com/microsoft/agent-framework/pull/1896` as I'm replating the implementation on top of IChatClient, so all the thread management is done by ChatClientAgent.

dotnet/src/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore/AGUIEndpointRouteBuilderExtensions.cs
javiercn: I see now how this was solved by bringing in the SG into the library, I'm changing it for that.

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIAgent.cs
javiercn: I'm going to go with `run_` similar to foundry agents.

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIAgent.cs
javiercn: Hmm, this is for clients, we should let them decide. The client can talk to an AG-UI service that we don't own, and they might not want to bother with the updates

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIAgent.cs
javiercn: This is gone on the next PR

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIAgent.cs
javiercn: The ag-ui typescript and python clients both do it the same way.

dotnet/src/Microsoft.Agents.AI.AGUI/Shared/AGUIEventTypes.cs
javiercn: Both `Microsoft.Agents.AI.Hosting.AGUI.AspNetCore` and `Microsoft.Agents.AI.AGUI` share all the types within the `Shared` folder. Since we give `InternalsVisibleTo` to the tests projects, we run into issues when we reference one of the shared types, because the compiler sees two definitions for the same type coming from two separate assemblies.

The solution is to put them in a namespace specific to that project. We chose to it this way because we didn't want to create a public AGUI library at the moment, but that might change in the future. If we choose to do so, then we would have a set of public types we will reuse.

dotnet/src/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore/AGUIEndpointRouteBuilderExtensions.cs
javiercn: That's a fair point, there are some challenges with AG-UI that might not be obvious at the moment. You might receive information on a per request basis that the agent needs to be able to handle:
* The list of messages is transmitted in full on each request, so if you are storing messages server side, you'll have to tweak the message list.
* Frontend tools: That need to be merged with the existing agent tools. This right now is agent specific.
  * ChatClientAgent can do it through ChatClientOptions in ChatClientAgentRunOptions, but there is no general way to do it with AIAgent.
* Context: Same deal, as a developer you need to be able to receive it and map it however you see fit.
* Any other props that are forwarded by the client.

If you have alternative ideas on how to tackle these that don't require a factory/function to tweak things, I'll be happy to change things. My idea with using a function is that it gives you control of what to do.

dotnet/src/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore/AGUIEndpointRouteBuilderExtensions.cs
javiercn: I've changed this.

dotnet/src/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore/AGUIEndpointRouteBuilderExtensions.cs
javiercn: https://github.com/microsoft/agent-framework/pull/1776#discussion_r2491440232

dotnet/src/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore/AGUIEndpointRouteBuilderExtensions.cs
javiercn: Got rid of this as I'm now returning the request

dotnet/src/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore/AGUIEndpointRouteBuilderExtensions.cs
DeagleGross: you need these 2 lines
https://github.com/microsoft/agent-framework/blob/8b4aa1ebb533ba941d26fc75fa737f7b5f1f4dec/dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Microsoft.Agents.AI.Hosting.OpenAI.csproj#L9-L10

and everything will work well without workarounds.

javiercn (commented):
Addressed feedback.

dotnet/src/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore/AGUIEndpointRouteBuilderExtensions.cs
javiercn: This is done now

dotnet/src/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore/AGUIEndpointRouteBuilderExtensions.cs
DeagleGross: > there are some challenges with AG-UI that might not be obvious at the moment
I dont think that really is related to the signature of the endpoint handler. I am fearing that we already have a2a / openai responses / openai chatcompletions in the repo, and they follow the similar pattern, but for some reason AG-UI would have a different one.

> My idea with using a function is that it gives you control of what to do.
Your signature is definitely giving more control, but I would propose something like:
```csharp
public static IEndpointConventionBuilder MapAGUIAgent(
        this IEndpointRouteBuilder endpoints,
        AIAgent agent,
        string path);

public static IEndpointConventionBuilder MapAGUIAgent(
        this IEndpointRouteBuilder endpoints,
        IHostedAgentBuilder agentBuilder,
        string path)
{
   var agent = endpoints.ServiceProvider.GetRequiredKeyedService<AIAgent>(agentBuilder.Name);
   return MapAGUIAgent(endpoints, agent, path);
}
```

that would allow users to do the following:
```
var agentBuilder = builder.AddAIAgent("pirate", ...);
...
var app = builder.Build();
...
app.MapAGUI(agentBuilder, "/");
```

I dont understand why based on the input message we would choose a different agent honestly. If users need such feature - lets collect their needs via a separate issue and then we can implement something similar.

DeagleGross: Also consider renaming to `MapAGUI()`. Also follows a naming convention and looks nicer IMO.

javiercn: I've switched this for now.

dotnet/src/Microsoft.Agents.AI.AGUI/Shared/BaseEventJsonConverter.cs
ReubenBond: Does it work if you set `AllowOutOfOrderMetadataProperties` on the `JsonSourceGeneratorOptions`? Eg

```csharp
[JsonSourceGenerationOptions(JsonSerializerDefaults.Web,
    DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull,
    NumberHandling = JsonNumberHandling.AllowReadingFromString,
    AllowOutOfOrderMetadataProperties = true,
    WriteIndented = false)]
```

javiercn:
> https://github.com/microsoft/agent-framework/pull/1776/files#diff-805d443b87ef13f1f4e96d50c89e5dc5940af17c524a3f2412738d66c4adae33

I'll give it a go on the other PR that I have in flight as I'm touching the json serialization there.

dotnet/src/Microsoft.Agents.AI.AGUI/Shared/BaseEventJsonConverter.cs
javiercn: I'll give that a try on my other PR

Issue #1437, opened 2025-10-13
.NET: implement own OpenAI type-model and remove OpenAI SDK as a dependency


DeagleGross:
Both [OpenAI Responses and ChatCompletions](https://github.com/microsoft/agent-framework/tree/main/dotnet/src/Microsoft.Agents.AI.OpenAI) implementation in agent-framework rely on the official [OpenAI SDK NuGet package](https://github.com/openai/openai-dotnet), which has a "non-traditional" type model definition. That includes:
1) no public exposure for some of openai protocol types (e.g. `StreamingResponseOutputItemAddedUpdate`). One could not instantiate such a type.
2) some of members (like `ResponseCreationOptions.Stream`) are **not** public.

Workaround today is to rely on the reflection, but that will easily break the consumers. So we should move from using OpenAI SDK sooner rather than later.

DeagleGross:
thanks to Reuben, OpenAI Responses is now not depending on OpenAI SDK package: https://github.com/microsoft/agent-framework/pull/1550

PR #1785, opened 2025-10-29
.NET: Improve fidelity of OpenAI ChatCompletions Hosting

DeagleGross:
### Motivation and Context

1) Reuben has already done a fantastic job improving OpenAI Responses at #1550. We need to do the same for ChatCompletions, and this is what I am doing here
2) OpenAI SDK is not suitable for writing server-based (hosting) scenarios, so we need to completely go off this package. See #1437

### Contribution Checklist


dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/Converters/MessageContentPartConverter.cs
```diff
@@ -0,0 +1,43 @@
...
+        return part switch
+        {
+            // text
+            TextContentPart textPart => new TextContent(textPart.Text),
+
+            // image
+            ImageContentPart imagePart when !string.IsNullOrEmpty(imagePart.UrlOrData) =>
+                imagePart.UrlOrData.StartsWith("data:", StringComparison.OrdinalIgnoreCase)
+                    ? new DataContent(imagePart.UrlOrData, "image/*")
+                    : new UriContent(imagePart.UrlOrData, "image/*"),
```
stephentoub: Can we infer relevant media types in most situations from the file extension of the url?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/Models/ChatCompletionChoice.cs
```diff
@@ -0,0 +1,155 @@
...
+    /// <summary>
+    /// A chat completion message generated by the model.
+    /// </summary>
+    [JsonPropertyName("message")]
+    public required ChoiceMessage Message { get; init; }
+}
+
+internal sealed record ChoiceMessage
+{
+    [JsonPropertyName("role")]
```
stephentoub, line 44: What dictates whether we're adding an XML comment to a property? Some have them, some don't.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/AgentRunResponseExtensions.cs
```diff
@@ -0,0 +1,127 @@
...
+                        {
+                            Data = audioContent.Base64Data.ToString(),
+                            Id = audioContent.Name,
+                            //Transcript = ,
+                            //ExpiresAt = ,
+                        },
+                    },
+
+                    // file
+                    DataContent fileContent when !fileContent.HasTopLevelMediaType("image") && !fileContent.HasTopLevelMediaType("audio") => new()
```
stephentoub: FWIW, these conditions are nops as earlier cases already handled them

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/Models/CreateChatCompletion.cs
```diff
@@ -0,0 +1,259 @@
...
+    [JsonPropertyName("service_tier")]
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public string? ServiceTier { get; set; }
+
+    /// <summary>
+    /// Up to 4 sequences where the API will stop generating further tokens.
+    /// </summary>
+    [JsonPropertyName("stop")]
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public object? Stop { get; set; }
```
stephentoub: Is this the right type?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/AIAgentChatCompletionsProcessor.cs
```diff
@@ -1,70 +1,37 @@
...
-        if (chatCompletionOptions.GetStream())
+        if (request.Stream == true)
         {
-            return new OpenAIStreamingChatCompletionResult(this._agent, chatMessages);
+            return new StreamingResponse(agent, request);
         }
 
-        var agentResponse = await this._agent.RunAsync(chatMessages, agentThread, cancellationToken: cancellationToken).ConfigureAwait(false);
-        return new OpenAIChatCompletionResult(agentResponse);
+        var chatMessages = request.Messages.Select(i => i.ToChatMessage());
```
stephentoub: Do we want to map other aspects of the request into ChatOptions that we could pass through in ChatClientAgentRunOptions?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/AIAgentChatCompletionsProcessor.cs
```diff
@@ -79,26 +46,73 @@ public Task ExecuteAsync(HttpContext httpContext)
...
+                        Index = 0,
+                        Delta = delta
+                    };
+
+                    choiceChunks.Add(choiceChunk);
+                }
+
+                var chunk = new ChatCompletionChunk
+                {
+                    Id = Guid.NewGuid().ToString(),
```
stephentoub: Can we use the same naming scheme as OpenAI? These have some kind of standardized prefix.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/AIAgentChatCompletionsProcessor.cs
```diff
@@ -79,26 +46,73 @@ public Task ExecuteAsync(HttpContext httpContext)
...
+                    Created = 1,
```
stephentoub: ?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/Converters/MessageContentPartConverter.cs
```diff
@@ -0,0 +1,43 @@
...
+using Microsoft.Agents.AI.Hosting.OpenAI.ChatCompletions.Models;
+using Microsoft.Extensions.AI;
+
+namespace Microsoft.Agents.AI.Hosting.OpenAI.ChatCompletions.Converters;
+
+internal static class MessageContentPartConverter
+{
+    public static AIContent? ToAIContent(MessageContentPart part)
+    {
+        return part switch
```
stephentoub, line 13: Anything to be done for annotations?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/Converters/MessageContentPartConverter.cs
```diff
@@ -0,0 +1,43 @@
...
+                    : new UriContent(imagePart.UrlOrData, "image/*"),
+
+            // audio
+            AudioContentPart audioPart =>
+                new DataContent(audioPart.InputAudio.Data, audioPart.InputAudio.Format.ToUpperInvariant() switch
+                {
+                    // only MP3 or WAV are supported
+                    "MP3" => "audio/mpeg",
+                    "WAV" => "audio/wav",
+                    _ => "audio/*"
```
stephentoub, line 34: According to the API reference docs, it looks like "wav, mp3, flac, opus, or pcm16" could all be valid?

DeagleGross: wanted to be on par with [API ref](https://platform.openai.com/docs/api-reference/chat/create), but we can support more i guess - does not hurt
image

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/Models/ChatCompletionChoice.cs
DeagleGross: did not finish yet; published a PR too early. Added missing XML comments

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/EndpointRouteBuilderExtensions.ChatCompletions.cs
```diff
@@ -20,47 +18,46 @@ public static partial class MicrosoftAgentAIHostingOpenAIEndpointRouteBuilderExt
...
     /// <param name="endpoints">The <see cref="IEndpointRouteBuilder"/> to add the OpenAI ChatCompletions endpoints to.</param>
-    /// <param name="agentName">The name of the AI agent service registered in the dependency injection container. This name is used to resolve the <see cref="AIAgent"/> instance from the keyed services.</param>
+    /// <param name="agentName">The name of <see cref="AIAgent"/> to map the OpenAI ChatCompletions endpoints for.</param>
     /// <param name="path">Custom route path for the chat completions endpoint.</param>
-    public static void MapOpenAIChatCompletions(
-        this IEndpointRouteBuilder endpoints,
-        string agentName,
-        [StringSyntax("Route")] string? path = null)
+    public static IEndpointConventionBuilder MapOpenAIChatCompletions(this IEndpointRouteBuilder endpoints, string agentName, string? path = null)
     {
```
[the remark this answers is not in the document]
DeagleGross: verified in inner call

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/Converters/MessageContentPartConverter.cs
stephentoub: Ok, I was looking at this part of the same spec:
image

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/ChatCompletionsJsonContext.cs
```diff
@@ -0,0 +1,42 @@
...
+[JsonSerializable(typeof(MessageContent))]
+[JsonSerializable(typeof(MessageContentPart))]
+[JsonSerializable(typeof(IReadOnlyList<MessageContentPart>))]
+[JsonSerializable(typeof(TextContentPart))]
+[JsonSerializable(typeof(ImageContentPart))]
+[JsonSerializable(typeof(AudioContentPart))]
+[JsonSerializable(typeof(FileContentPart))]
+[JsonSerializable(typeof(FileContentPart))]
+[JsonSerializable(typeof(ChatCompletionChoice))]
+[JsonSerializable(typeof(IList<ChatCompletionChoice>))]
```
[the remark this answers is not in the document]
DeagleGross: actually we need `[JsonSerializable(typeof(IList<ChatCompletionChoiceChunk>))]`

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/AgentRunResponseExtensions.cs
DeagleGross: makes sense! Simplified

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/Models/CreateChatCompletion.cs
DeagleGross: Thought we can skip it for now, because I am not sure how we can support it in the AIAgent. Anyway, defined properly per docs

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/AIAgentChatCompletionsProcessor.cs
DeagleGross: yes! Done it as I could. Not sure if I mapped everything I could: mostly I am not understanding how I can instantiate `AITool` from the request. Is there a "DescriptionAITool" one which does not require a delegate?

stephentoub: > Is there a "DescriptionAITool" one which does not require a delegate?

AIFunctionFactory.CreateDeclaration, or derive your own from AIFunctionDeclaration.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/Converters/MessageContentPartConverter.cs
DeagleGross: added in the `AgentRunResponseExtensions.ToChoiceMessageAnnotations`

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/AIAgentChatCompletionsProcessor.cs
DeagleGross: Moved `IdGeneratorHelpers` in shared namespace (`Microsoft.Agents.AI.Hosting.OpenAI`), and using this one to generate an id like `chatcmpl-B9MBs8CjcvOU2jLn4n570S5qMJKcT`.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/AIAgentChatCompletionsProcessor.cs
DeagleGross: fixed

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/Converters/MessageContentPartConverter.cs
DeagleGross: makes sense, done.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/AIAgentChatCompletionsProcessor.cs
DeagleGross: done, I think I covered all; If you spot anyhting - let me know

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/EndpointRouteBuilderExtensions.ChatCompletions.cs
```diff
@@ -20,47 +18,46 @@ public static partial class MicrosoftAgentAIHostingOpenAIEndpointRouteBuilderExt
...
+    /// <summary>
+    /// Maps OpenAI ChatCompletions API endpoints to the specified <see cref="IEndpointRouteBuilder"/> for the given <see cref="AIAgent"/>.
+    /// </summary>
+    /// <param name="endpoints">The <see cref="IEndpointRouteBuilder"/> to add the OpenAI ChatCompletions endpoints to.</param>
+    /// <param name="agent">The <see cref="AIAgent"/> instance to map the OpenAI ChatCompletions endpoints for.</param>
+    /// <param name="path">Custom route path for the chat completions endpoint.</param>
+    public static IEndpointConventionBuilder MapOpenAIChatCompletions(
+        this IEndpointRouteBuilder endpoints,
+        AIAgent agent,
+        [StringSyntax("Route")] string? path = null)
```
ReubenBond: IMO we should avoid default parameters where possible. I rather have overloads. It makes it easier to modify things later.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/EndpointRouteBuilderExtensions.ChatCompletions.cs
```diff
@@ -20,47 +18,46 @@ public static partial class MicrosoftAgentAIHostingOpenAIEndpointRouteBuilderExt
...
     /// </summary>
     /// <param name="endpoints">The <see cref="IEndpointRouteBuilder"/> to add the OpenAI ChatCompletions endpoints to.</param>
-    /// <param name="agentName">The name of the AI agent service registered in the dependency injection container. This name is used to resolve the <see cref="AIAgent"/> instance from the keyed services.</param>
+    /// <param name="agentName">The name of <see cref="AIAgent"/> to map the OpenAI ChatCompletions endpoints for.</param>
     /// <param name="path">Custom route path for the chat completions endpoint.</param>
-    public static void MapOpenAIChatCompletions(
-        this IEndpointRouteBuilder endpoints,
-        string agentName,
-        [StringSyntax("Route")] string? path = null)
+    public static IEndpointConventionBuilder MapOpenAIChatCompletions(this IEndpointRouteBuilder endpoints, string agentName, string? path = null)
```
ReubenBond: I rather we don't have this convenience overload, since if we wanted to add an endpoint which extracted the agent name out of the request, it would naturally only accept a string for the path, and then the calls become ambiguous.

ReubenBond:
I don't see any tests. We should have extensive testing to demonstrate conformance/compatibility and to prevent regressions. Eg, it's easy to make a mistake with JSON serialization, such as having enums serialized with the wrong string convention or union/polymorphic types having the wrong format.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/EndpointRouteBuilderExtensions.ChatCompletions.cs
DeagleGross: removed `= null`.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/EndpointRouteBuilderExtensions.ChatCompletions.cs
DeagleGross: Removed this overload. Included one on top of `IHostedAgentBuilder` for convenience.

Now we have the following API:
```csharp
IEndpointConventionBuilder MapOpenAIChatCompletions(this IEndpointRouteBuilder endpoints, IHostedAgentBuilder agentBuilder)

IEndpointConventionBuilder MapOpenAIChatCompletions(this IEndpointRouteBuilder endpoints, IHostedAgentBuilder agentBuilder, string? path);

IEndpointConventionBuilder MapOpenAIChatCompletions(this IEndpointRouteBuilder endpoints, AIAgent agent);

IEndpointConventionBuilder MapOpenAIChatCompletions(this IEndpointRouteBuilder endpoints, AIAgent agent, [StringSyntax("Route")] string? path);
```

DeagleGross:
thanks for review Reuben, added tests similar to what we have in OpenAI Responses

dotnet/samples/AgentWebChat/AgentWebChat.AgentHost/Program.cs
```diff
@@ -80,6 +80,8 @@ Once the user has deduced what type (knight or knave) both Alice and Bob are, te
 
 builder.AddSequentialWorkflow("science-sequential-workflow", [chemistryAgent, mathsAgent, literatureAgent]).AddAsAIAgent();
 builder.AddConcurrentWorkflow("science-concurrent-workflow", [chemistryAgent, mathsAgent, literatureAgent]).AddAsAIAgent();
+
+builder.AddOpenAIChatCompletions();
```
westey-m, line 84: I don't think ChatCompletion is typically referred to in the plural, so suggesting `AddOpenAIChatCompletion` instead.  Same below for mapping `MapOpenAIChatCompletion`

Note that Responses typically is referred to in the plural, just not in the .net SDK.

DeagleGross: I've seen only plural honestly :D
https://platform.openai.com/docs/api-reference/chat
image

Also the endpoint itself per API reference is `POST https://api.openai.com/v1/chat/completions` which speaks of the chat completions in the plural form.

I would leave it as it is honestly, but if other people or other SDKs use the "chatCompletio**n**" then I can change for sure. What do you think?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/AIAgentChatCompletionsProcessor.cs
```diff
@@ -1,70 +1,44 @@
...
     {
-        this._agent = agent;
-    }
+        ArgumentNullException.ThrowIfNull(agent);
 
-    public async Task<IResult> CreateChatCompletionAsync(ChatCompletionOptions chatCompletionOptions, CancellationToken cancellationToken)
-    {
-        AgentThread? agentThread = null; // not supported to resolve from conversationId
+        var chatMessages = request.Messages.Select(i => i.ToChatMessage());
+        var chatClientAgentRunOptions = request.BuildOptions();
```
westey-m, line 26: This should come with a bit of a health warning, in that not all agents will necessarily be able to make use of the *ChatClient*AgentRunOptions.  With function tools it'll be even more restrictive, in that it would have to be a ChatClientAgent that has explicitly opted out of automatic function calling or set TerminateOnUnknownCalls to true on FunctionInvokingChatClient.  None of these are of course the defaults.

dotnet/samples/AgentWebChat/AgentWebChat.AgentHost/Program.cs
westey-m: Yeah, you are right, it is not as consistent as I had initially thought.  Fine to leave as is then.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ServiceCollectionExtensions.cs
```diff
@@ -2,16 +2,31 @@
...
+    /// <summary>
+    /// Adds support for exposing <see cref="AIAgent"/> instances via OpenAI ChatCompletions.
+    /// </summary>
+    /// <param name="services">The <see cref="IServiceCollection"/> to configure.</param>
+    /// <returns>The <see cref="IServiceCollection"/> for method chaining.</returns>
+    public static IServiceCollection AddOpenAIChatCompletions(this IServiceCollection services)
+    {
+        ArgumentNullException.ThrowIfNull(services);
+
+        services.Configure<JsonOptions>(options => options.SerializerOptions.TypeInfoResolverChain.Add(ChatCompletionsJsonSerializerOptions.Default.TypeInfoResolver!));
```
westey-m, line 25: Where are these options being used?  From what I can tell, I see the Processor just using the static `ChatCompletionsJsonSerializerOptions` class.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/IdGeneratorHelpers.cs
```diff
@@ -0,0 +1,98 @@
...
+    /// <param name="partitionKeyHint">An existing ID to extract the partition key from. When provided, the same partition key will be used instead of generating a new one.</param>
+    /// <returns>A new ID with format "{prefix}{delimiter}{infix}{entropy}{delimiter}{partitionKey}".</returns>
+    /// <exception cref="ArgumentException">Thrown when the watermark contains non-alphanumeric characters.</exception>
+    public static string NewId(string prefix, int stringLength = 32, int partitionKeyLength = 16, string infix = "",
+        string watermark = "", string delimiter = "_", string? partitionKey = null, string partitionKeyHint = "")
+    {
+        ArgumentOutOfRangeException.ThrowIfLessThan(stringLength, 1);
+        var entropy = GetRandomString(stringLength);
+
+        string pKey = partitionKey ?? GetPartitionIdOrDefault(partitionKeyHint) ?? GetRandomString(partitionKeyLength);
```
westey-m, line 41: Since we are using a random string generator here to generate new ids, how do we ensure that these ids are unique?  I assume we want them to be unique, so that we don't accidentally generate the same number for a new thread, and therefore provide the new thread access to the existing thread's chat history.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/AIAgentChatCompletionsProcessor.cs
westey-m: Actually, I just remembered that we are passing and can only pass AIFunctionDeclerations, and they are automatically ignored by FICC.  So it shouldn't be too bad.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/IdGeneratorHelpers.cs
stephentoub: RandomNumberGenerator is cryptographically-secure. This is using 32 characters selected from a 62 character alphabet. That's ~190 bits of entropy, which is much more than Guid's ~123. The chances of a collision are impossibly small.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/AgentRunResponseExtensions.cs
```diff
@@ -0,0 +1,208 @@
...
+    public static ChatCompletion ToChatCompletion(this AgentRunResponse agentRunResponse, CreateChatCompletion request)
+    {
+        IList<ChatCompletionChoice> choices = agentRunResponse.ToChoices();
+
+        return new ChatCompletion
+        {
+            Id = IdGeneratorHelpers.NewId(prefix: "chatcmpl", delimiter: "-", stringLength: 13),
+            Choices = choices,
+            Created = (agentRunResponse.CreatedAt ?? DateTimeOffset.UtcNow).ToUnixTimeSeconds(),
+            Model = /* request.Agent?.Name ?? */ request.Model,
```
westey-m:
```suggestion
            Model = request.Model,
```

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ServiceCollectionExtensions.cs
DeagleGross: from my understanding, in order to configure how aspnetcore will parse the underlying type, you need to configure it via `services.Configure<JsonOptions>(...)`. That's needed because we expect the input as a parameter of a minimal-api method like `app.MapPost("/", [FromBody] Input input => ...)`, and aspnetcore has to know what serialization/deserialization mechanism to use.

Some related info: https://stackoverflow.com/a/77111105/6779773 / https://learn.microsoft.com/en-us/aspnet/core/fundamentals/minimal-apis/responses?view=aspnetcore-9.0#configure-json-serialization-options

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/AgentRunResponseExtensions.cs
DeagleGross: yep, forgot to remove it

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/AIAgentChatCompletionsProcessor.cs
```diff
@@ -79,26 +53,99 @@ public Task ExecuteAsync(HttpContext httpContext)
...
+                        // image
+                        DataContent imageContent when imageContent.HasTopLevelMediaType("image") => new() { Content = imageContent.Base64Data.ToString() },
+                        UriContent urlContent when urlContent.HasTopLevelMediaType("image") => new() { Content = urlContent.Uri.ToString() },
+
+                        // audio
+                        DataContent audioContent when audioContent.HasTopLevelMediaType("audio") => new() { Content = audioContent.Base64Data.ToString() },
+
+                        // file
+                        DataContent fileContent when !fileContent.HasTopLevelMediaType("image") && !fileContent.HasTopLevelMediaType("audio")
+                            => new() { Content = fileContent.Base64Data.ToString() },
```
stephentoub:
```suggestion
                        DataContent fileContent => new() { Content = fileContent.Base64Data.ToString() },
```

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/AIAgentChatCompletionsProcessor.cs
```diff
@@ -79,26 +53,99 @@ public Task ExecuteAsync(HttpContext httpContext)
...
+                    ChatCompletionDelta? delta = content switch
+                    {
+                        TextContent textContent => new() { Content = textContent.Text },
+
+                        // image
+                        DataContent imageContent when imageContent.HasTopLevelMediaType("image") => new() { Content = imageContent.Base64Data.ToString() },
+                        UriContent urlContent when urlContent.HasTopLevelMediaType("image") => new() { Content = urlContent.Uri.ToString() },
+
+                        // audio
+                        DataContent audioContent when audioContent.HasTopLevelMediaType("audio") => new() { Content = audioContent.Base64Data.ToString() },
```
stephentoub, line 102: Does it not support audio as a uri?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/AIAgentChatCompletionsProcessor.cs
```diff
@@ -79,26 +53,99 @@ public Task ExecuteAsync(HttpContext httpContext)
...
+                        // function call
+                        FunctionCallContent functionCallContent => new()
+                        {
+                            ToolCalls = [functionCallContent.ToChoiceMessageToolCall()]
+                        },
+
+                        // function result. ChatCompletions dont provide the results of function result per API reference
+                        FunctionResultContent functionResultContent => null,
+
+                        _ => throw new InvalidOperationException($"Got unsupported content: {content.GetType()}")
```
stephentoub: We've typically just ignored unknown content, i.e.
```csharp
_ => null;
```

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/AgentRunResponseExtensions.cs
```diff
@@ -0,0 +1,208 @@
...
+                    // function call
+                    FunctionCallContent functionCallContent => new()
+                    {
+                        ToolCalls = [functionCallContent.ToChoiceMessageToolCall()]
+                    },
+
+                    // function result. ChatCompletions dont provide the results of function result per API reference
+                    FunctionResultContent functionResultContent => null,
+
+                    _ => throw new InvalidOperationException($"Got unsupported content: {content.GetType()}")
```
stephentoub: Similarly here.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/AIAgentChatCompletionsProcessor.cs
DeagleGross: Docs dont mention audio as a uri.
image

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/AIAgentChatCompletionsProcessor.cs
DeagleGross: changed

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/AIAgentChatCompletionsProcessor.cs
DeagleGross: simplified

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/Converters/ChatClientAgentRunOptionsConverter.cs
```diff
@@ -0,0 +1,118 @@
...
+        {
+            return ChatResponseFormat.Json;
+        }
+        if (responseFormat.IsJsonSchema)
+        {
+            var schema = responseFormat.JsonSchema.JsonSchema;
+            return ChatResponseFormat.ForJsonSchema(schema.Schema, schema.Name, schema.Description);
+        }
+
+        throw new ArgumentOutOfRangeException(nameof(responseFormat));
```
stephentoub, line 61: Similar question for all of these, whether we want to throw or just default to something reasonable, e.g. like Text here

stephentoub (approved):

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/Converters/ChatClientAgentRunOptionsConverter.cs
DeagleGross: i think for content types that's ok to ignore/fallback, but here it is strongly according to the spec - a union. I would leave as it is.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/AgentRunResponseExtensions.cs
```diff
@@ -0,0 +1,209 @@
...
+        }
+
+        return chatCompletionChoices;
+    }
+
+    /// <summary>
+    /// Converts UsageDetails to ResponseUsage.
+    /// </summary>
+    /// <param name="usage">The usage details to convert.</param>
+    /// <returns>A ResponseUsage object with zeros if usage is null.</returns>
```
adityamandaleeka: CompletionUsage? (and above)

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/Models/MessageContent.cs
```diff
@@ -0,0 +1,165 @@
...
+        }
+
+        // Both text
+        if (this.Text is not null && other.Text is not null)
+        {
+            return this.Text == other.Text;
+        }
+
+        // Both contents
+        if (this.Contents is not null && other.Contents is not null)
```
adityamandaleeka: Consider a Count check here?

PR #1787, opened 2025-10-29
.NET: fix: WorkflowAsAgent Sample

### LINKED ISSUE #1613: .NET: Cannot reuse Workflow with shared Executor instances that do not implement IResettableExecutor

lokitoth:
When we started introducing the different execution modes to solidify the concurrency model, we broke the WorkflowAsAgent Sample, which was not updated for either of `IResettableExecutor` or Cross-Run Shareable.

The fix is to make public and switch to using the `ChatForwardingExecutor` in place of a self-defined `ConcurrentStartExecutor`, and make the `ConcurrentAggregationExecutor` implement `IResettableExecutor`

### Description

* Fixes the Workflow in WorkflowAsAgent sample to be ChatProtocol and resettable (closes #1613)
* Also makes ChatForwardingExecutor public

### Contribution Checklist

### DIFF
--- a/dotnet/samples/GettingStarted/Workflows/Agents/WorkflowAsAnAgent/WorkflowFactory.cs
+++ b/dotnet/samples/GettingStarted/Workflows/Agents/WorkflowAsAnAgent/WorkflowFactory.cs
@@ -41,7 +41,8 @@ internal static class WorkflowFactory
     /// <summary>
     /// Executor that aggregates the results from the concurrent agents.
     /// </summary>
-    private sealed class ConcurrentAggregationExecutor() : Executor<List<ChatMessage>>("ConcurrentAggregationExecutor")
+    private sealed class ConcurrentAggregationExecutor() :
+        Executor<List<ChatMessage>>("ConcurrentAggregationExecutor"), IResettableExecutor
     {
         private readonly List<ChatMessage> _messages = [];
 


dotnet/samples/GettingStarted/Workflows/Agents/WorkflowAsAnAgent/WorkflowFactory.cs
```diff
@@ -86,5 +63,12 @@ public override async ValueTask HandleAsync(ChatMessage message, IWorkflowContex
                 await context.YieldOutputAsync(formattedMessages, cancellationToken);
             }
         }
+
+        /// <inheritdoc/>
+        public ValueTask ResetAsync()
+        {
+            this._messages.Clear();
+            return default;
+        }
```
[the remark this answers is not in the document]
lokitoth: This line already uses `this.`

dotnet/src/Microsoft.Agents.AI.Workflows/ChatForwardingExecutor.cs
```diff
@@ -4,17 +4,26 @@
 using System.Threading.Tasks;
 using Microsoft.Extensions.AI;
 
-namespace Microsoft.Agents.AI.Workflows.Specialized;
+namespace Microsoft.Agents.AI.Workflows;
 
-/// <summary>Executor that forwards all messages.</summary>
-internal sealed class ChatForwardingExecutor(string id) : Executor(id, declareCrossRunShareable: true), IResettableExecutor
+/// <summary>
+/// A ChatProtocol executor that forwards all messages it receives. Useful for splitting inputs into parallel
```
stephentoub, line 24: At the moment it's only handling/forwarding string/ChatMessage/`List<ChatMessage>`/TurnToken. Can you remind me why it doesn't work to just have it have a catch-all handler for object?

lokitoth: We cannot have a CatchAll handler for `object` because it would leak the `JsonElement` object through the abstraction when resuming from a Checkpoint if it is serialized to JSON. Unfortunately, System.Text.Json is incapable of handling `object` as a target type in an easy manner or one that I expect could work generally without weird edge cases.

We could use `PortableValue`, but we disallow passing forward a `PortableValue` because we would not be able to validate / transform it to a real type to route the message (since checkpoint deletes real type information in favour of TypeId, and using TypeId for inheritance does not always work).

We would have to give up on AOT in order to be able to load types by name to make this work properly, or build a big type table, and still likely run into issues.

lokitoth: Expanding on the above a bit: 

We could build `ChatForwardingExecutor` with a catchall on `PortableValue`, check the type, and extract/forward it only when the type matches. But that would mean that any message sent to it will cause the handler to be invoked, even if it could be known a priori that it does not handle that type of message.

I did update it to take the same set of inputs as ChatProtocolExecutor for consistency, though.

The correct long-term fix is not easy, because we need to have definite knowledge of the source runtime `Type` that can survive a serialization/deserialization boundary; AOT makes that difficult because load-type-by-name does not work in that context.

lokitoth:
Added one more fix to update for a change in how message aggregation happens in `AIAgentHostExecutor` post #1439

lokitoth:
Pending #1612

Issue #1735, opened 2025-10-27
.NET: Clarify that AIContext.Messages aren't permanently incorporated into the conversation history


kzu:
The wording on AIContext is confusing: 

https://github.com/microsoft/agent-framework/blob/f21b9d14f85169f6053a4c5f9a002c0d2377c425/dotnet/src/Microsoft.Agents.AI.Abstractions/AIContext.cs#L26-L29

Upon closer inspection, it turns out that `NotifyThreadOfNewMessagesAsync` (which is invoked to ultimately persist the conversation history, "making it permanent", only the input messages plus chat response messages are passed in, without any of the messages provided by the AIContext: 

https://github.com/microsoft/agent-framework/blob/f21b9d14f85169f6053a4c5f9a002c0d2377c425/dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs#L197-L198

https://github.com/microsoft/agent-framework/blob/f21b9d14f85169f6053a4c5f9a002c0d2377c425/dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs#L309-L310

https://github.com/microsoft/agent-framework/blob/f21b9d14f85169f6053a4c5f9a002c0d2377c425/dotnet/src/Microsoft.Agents.AI.Abstractions/AIAgent.cs#L346-L351

https://github.com/microsoft/agent-framework/blob/f21b9d14f85169f6053a4c5f9a002c0d2377c425/dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgentThread.cs#L201-L202

Suggestion is to reword that so it's clear they are never persisted, although they are always added for execution.

westey-m:
Thanks for raising this issue @kzu, this is a good callout.

I actually consider the fact that ChatClientAgent doesn't pass these messages to the thread the bug, rather than the comment being incorrect.  The comment is actually correct for service managed threads, but incorrect for locally managed threads (e.g. InMemory).
The only option that will work consistently with all the different thread types is to always update the thread.

Right now, for service managed threads, once we include the AIContextProvider messages in the request to the underlying IChatClient, the service managed thread will be updated with these additional messages.  Responses e.g. has no support for including messages in the request while also requiring that these messages shouldn't become a permanent part of the chat history.

I'll create a PR to fix.

kzu:
[...]

PR #1788, opened 2025-10-29
.NET: [BREAKING] Fix issue where AIContextProvider messages were not added to MessageStores

westey-m:
### Motivation and Context

#1735

### Description

- Fixing issue where AIContextProvider messages are not added to local MessageStores.  This meant that the behavior differed for locally and service managed threads, so this fix brings consistency.
- Fixing issue with out-of-the-box AIContextProviders that were adding messages to memory even if the request failed.

### Contribution Checklist

### DIFF
--- a/dotnet/agent-framework-dotnet.slnx
+++ b/dotnet/agent-framework-dotnet.slnx
@@ -59,6 +59,7 @@
     <Project Path="samples/GettingStarted/Agents/Agent_Step16_ChatReduction/Agent_Step16_ChatReduction.csproj" />
     <Project Path="samples/GettingStarted/Agents/Agent_Step17_BackgroundResponses/Agent_Step17_BackgroundResponses.csproj" />
     <Project Path="samples/GettingStarted/Agents/Agent_Step18_TextSearchRag/Agent_Step18_TextSearchRag.csproj" />
+    <Project Path="samples/GettingStarted/Agents/Agent_Step19_Mem0Provider/Agent_Step19_Mem0Provider.csproj" />
   </Folder>
   <Folder Name="/Samples/GettingStarted/AgentWithOpenAI/">
     <File Path="samples/GettingStarted/AgentWithOpenAI/README.md" />
--- a/dotnet/samples/GettingStarted/Agents/Agent_Step19_Mem0Provider/Agent_Step19_Mem0Provider.csproj
+++ b/dotnet/samples/GettingStarted/Agents/Agent_Step19_Mem0Provider/Agent_Step19_Mem0Provider.csproj
@@ -1,0 +1,22 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net9.0</TargetFramework>
+
+    <Nullable>enable</Nullable>
+    <ImplicitUsings>enable</ImplicitUsings>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Azure.AI.OpenAI" />
+    <PackageReference Include="Azure.Identity" />
+    <PackageReference Include="Microsoft.Extensions.AI.OpenAI" />
+  </ItemGroup>
+
+  <ItemGroup>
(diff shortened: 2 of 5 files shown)


dotnet/src/Microsoft.Agents.AI.Abstractions/AIContextProvider.cs
```diff
@@ -163,36 +162,47 @@ public class InvokedContext
         /// Initializes a new instance of the <see cref="InvokedContext"/> class with the specified request messages.
         /// </summary>
         /// <param name="requestMessages">The messages that were sent to the AI model or agent for this invocation.</param>
+        /// <param name="aiContextProviderMessages">The messages provided by the <see cref="AIContextProvider"/> for this invocation, if any.</param>
         /// <exception cref="ArgumentNullException"><paramref name="requestMessages"/> is <see langword="null"/>.</exception>
-        public InvokedContext(IEnumerable<ChatMessage> requestMessages)
+        public InvokedContext(IEnumerable<ChatMessage> requestMessages, IEnumerable<ChatMessage>? aiContextProviderMessages)
```
markwallace-microsoft, line 164: Should the new parameter default to null

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
```diff
@@ -601,7 +610,8 @@ await thread.AIContextProvider.InvokedAsync(new(inputMessages) { InvokeException
                 var aiContext = await typedThread.AIContextProvider.InvokingAsync(invokingContext, cancellationToken).ConfigureAwait(false);
                 if (aiContext.Messages is { Count: > 0 })
                 {
-                    threadMessages.AddRange(aiContext.Messages);
+                    inputMessagesForChatClient.AddRange(aiContext.Messages);
```
markwallace-microsoft, line 613: Above this is a comment "_This does not include any <see cref="AIContextProvider"/> supplied messages._" so I was expecting that aiContext.Messages would not be included. Can you explain what is happening? I think we need some better comments.

dotnet/src/Microsoft.Agents.AI.Abstractions/AIContextProvider.cs
westey-m: While this class is public, since users need to consume it in `AIContextProvider` implementations, and it's got a public constructor since users may need to construct it in their tests, I don't expect that users will ever construct this outside of a test scenario.
I therefore think it's better to force a decision on what to provide for this parameter, rather than letting it be an optional param.  In the main use case, we always provide this parameter, some times it just happens to be null since the AIContextProvider produced no messages.

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
westey-m: I've updated the xml docs for `AIContextProvider` further.  Since I think that they can be used without an Agent, the comments worded as such, but I think this adds too much confusion.  I've reworded to clarify that.
I also reworded the request and aicontetprovider messages props to be clearer that the first is what the caller provided, i.e. via the request, and the second is what the context provider provided.

Here specifically I originally renamed threadMessages to inputMessagesForChatClient to clarify that this is the messages passed to the chat client.  With the AIContextProvider, it is in the context of input into the agent, not into the chat client, that's why aiContext.Messages are being included here.

dotnet/src/Microsoft.Agents.AI.Abstractions/AIContextProvider.cs
```diff
@@ -18,25 +18,25 @@ namespace Microsoft.Agents.AI;
...
+/// <item><description>Providing additional context to agents during invocation</description></item>
 /// <item><description>Supplying additional function tools for enhanced capabilities</description></item>
 /// <item><description>Processing invocation results for state management or learning</description></item>
 /// </list>
 /// </para>
 /// <para>
-/// Context providers operate through a two-phase lifecycle: they are called before invocation via
-/// <see cref="InvokingAsync"/> to provide context, and optionally called after invocation via
+/// Context providers operate through a two-phase lifecycle: they are called at the start of invocation via
+/// <see cref="InvokingAsync"/> to provide context, and optionally called at the end of invocation via
```
markwallace-microsoft, line 28: To me before/after are clearer terms

PR #1822, opened 2025-10-30
.NET: Add SuppressAssistantName option to ChatClientAgentOptions

### LINKED ISSUE #1807: .NET: Allow disabling name on assistant messages (OpenAI-compatible backends like vLLM 0.9.2 reject assistant.name)
### Environment

.NET SDK: 8.0.18
OS: Windows 10 (22631)
Microsoft.Agents.AI: (preview; from User-Agent: MEAI/9.10.1-preview.1.25521.4)
Microsoft.Extensions.AI + OpenAI .NET SDK (latest preview matching above)
Backend: vLLM 0.9.2 (OpenAI-compatible API server)

### Summary
When using ChatClientAgent with the Microsoft Agent Framework against an OpenAI-compatible backend (vLLM 0.9.2), requests to /v1/chat/completions fail with 400 Bad Request because the framework emits a name field on assistant messages. Even if I omit the name parameter or pass null to ChatClientAgent, the outbound payload still contains "name": "UnnamedAgent". vLLM validates the schema strictly and rejects assistant.name (extra_forbidden).

### Why this matters
Some OpenAI-compatible servers (including vLLM’s FastAPI/Pydantic models) accept name only for legacy role:"function" messages and forbid extra properties on assistant/user/system. Today there is no supported way to suppress the name field from messages produced by the framework, so these backends break out of the box.

### Minimal repro

```
var chatClient = new OpenAI.Chat.ChatClient(
    model: modelId,
    credential: new ApiKeyCredential(apiKey),
    options: new OpenAIClientOptions { Endpoint = new Uri(baseUrl) }  // OpenAI-compatible endpoint (vLLM 0.9.2)
);

// Adapt and build pipeline
IChatClient iChat = new ChatClientBuilder(chatClient.AsIChatClient())
    .UseFunctionInvocation()
    .Build();

// Build agent — even if I omit or set name: null, framework still emits "UnnamedAgent"
var agent = new ChatClientAgent(
    iChat,
    instructions: "You are concise, answer in French.",
    /* name: null, */           // ← omitting or null still yields "UnnamedAgent"
    tools: new List<AITool> { /* …tool(s)… */ }
);

// Multi-turn message on same thread
var thread = agent.GetNewThread();
var resp = await agent.RunAsync("Quel temps à Paris ?", thread);
```

### Observed request payload (excerpt)

```
{
  "messages": [
    {"role": "system", "content": "Tu es un agent concis qui répond en français. Utilise tes tools quand c’est pertinent."},
    {"role": "user", "content": "c quoi la météo a paris ?"},
    {
      "role": "assistant",
      "name": "UnnamedAgent",                      // ← injected by framework, cannot disable
      "tool_calls": [
        {
          "id": "DPYF7J6FN",
          "type": "function",
          "function": { "name": "GetWeather", "arguments": "{ \"city\": \"paris\" }" }
        }
      ]
    },
    { "role": "tool", "tool_call_id": "DPYF7J6FN", "content": "\"À Paris (France), 13,2°C, Nuageux, vent 11,9 km/h.\"" },
    { "role": "assistant", "content": "…", "name": "UnnamedAgent" },   // ← again
    { "role": "user", "content": "et a Brest ?" }
  ],
  "model": "internal",
  "tools": [ /* … */ ],
  "tool_choice": "auto"
}
```

### vLLM error response

```
400 Bad Request
{
  "error": {
    "message": "2 validation errors for ChatCompletionRequest
messages.2.assistant.name
  Extra inputs are not permitted [type=extra_forbidden, input_value='UnnamedAgent', input_type=str]
messages.4.assistant.name
  Extra inputs are not permitted [type=extra_forbidden, input_value='UnnamedAgent', input_type=str]",
    "type": "BadRequestError",
    "code": 400
  }
}
```

### Expected behavior

If I omit the name argument or set it to null, the framework should not emit any name for assistant/user/system messages in the outbound payload.

Ideally, provide an opt-in knob to control emission of name per role (e.g., assistant/user/system), defaulting to current behavior for backward compatibility.

### Actual behavior

The framework injects name automatically (as "UnnamedAgent") on assistant messages, even when the agent constructor name parameter is omitted/null, causing vLLM to return 400.

### Workarounds

Wrap IChatClient and strip ChatMessage.Name on all non-function messages before calling the inner client; or

Inject an HTTP handler that rewrites the JSON body to remove messages[*].name.
Both workarounds are brittle/extra plumbing and defeat the goal of using the framework “as is”.

### Proposed improvements

Add a switch to disable name emission on assistant/user/system messages (e.g., AgentMessageOptions.SuppressAssistantName = true).
If name is omitted/null in the constructor, do not fallback to "UnnamedAgent".
Document compatibility notes for strict OpenAI-compatible servers (vLLM) that disallow assistant.name.

Thanks for considering this. This change would greatly improve interoperability with strict OpenAI-compatible backends such as vLLM.

T0mSIlver:
### Description

Fixes #1807

Adds support for suppressing the `name` field on assistant messages when the agent name is not set, enabling compatibility with strict OpenAI-compatible backends (e.g., vLLM's 0.9.2+ OpenAI-Compatible server) that reject assistant messages containing a `name` field.

### Problem

When using `ChatClientAgent` with an OpenAI-compatible backend that strictly validates the schema, requests fail with `400 Bad Request` because the framework automatically emits `"name": "UnnamedAgent"` on assistant messages even when the agent's `Name` property is `null`. Some backends only accept `name` for legacy `role:"function"` messages and forbid extra properties on `assistant`/`user`/`system` messages.

### Solution

Added a new `SuppressAssistantName` boolean property to `ChatClientAgentOptions` that controls whether the name field is included on assistant messages:

- When `SuppressAssistantName = true` and `Name = null`: Assistant messages will not include a `name` field (returns `null` for `AuthorName`)
- When `SuppressAssistantName = false` (default) and `Name = null`: Framework uses `"UnnamedAgent"` as fallback (maintains backward compatibility)
- When `Name` is explicitly set: The name is always used regardless of the `SuppressAssistantName` flag

### Changes

- Added `SuppressAssistantName` property to `ChatClientAgentOptions`
- Updated `ChatClientAgent` to respect the flag in both `RunAsync` and `RunStreamingAsync` methods
- Added `GetMessageAuthorName()` helper method to centralize the logic
- Updated `Clone()` method to include the new property
- Added unit tests

### Backward Compatibility

- The new property defaults to `false`, preserving existing behavior where `"UnnamedAgent"` is used as a fallback name.

### Testing

- Added `RunAsyncRespectsSuppressAssistantNameOptionAsync` theory test covering 3 scenarios:
  - `SuppressAssistantName=true, Name=null` → `AuthorName=null`
  - `SuppressAssistantName=false, Name=null` → `AuthorName="UnnamedAgent"` (backward compatibility)
  - `SuppressAssistantName=true, Name="MyAgent"` → `AuthorName="MyAgent"` (respects explicit name)
- Added RunStreamingAsyncRespectsSuppressAssistantNameOptionAsync theory test covering the same 3 scenarios for the streaming code path
- Added `Clone_CopiesSuppressAssistantNameProperty` test to verify cloning behavior

### Usage Example

```csharp
var agent = new ChatClientAgent(
    chatClient,
    options: new ChatClientAgentOptions
    {
        SuppressAssistantName = true  // Suppress name field when Name is null
    }
);
```

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
+++ b/dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
@@ -246,7 +246,11 @@ public sealed partial class ChatClientAgent : AIAgent
             var update = responseUpdatesEnumerator.Current;
             if (update is not null)
             {
-                update.AuthorName ??= this.Name;
+                var messageAuthorName = this.GetMessageAuthorName();
+                if (messageAuthorName is not null)
+                {
+                    update.AuthorName ??= messageAuthorName;
+                }
 
                 responseUpdates.Add(update);
                 yield return new(update) { AgentId = this.Id };
@@ -376,9 +380,13 @@ public sealed partial class ChatClientAgent : AIAgent
         this.UpdateThreadWithTypeAndConversationId(safeThread, chatResponse.ConversationId);
 
         // Ensure that the author name is set for each message in the response.
+        var messageAuthorName = this.GetMessageAuthorName();
         foreach (ChatMessage chatResponseMessage in chatResponse.Messages)
         {
-            chatResponseMessage.AuthorName ??= agentName;
+            if (messageAuthorName is not null)
+            {
+                chatResponseMessage.AuthorName ??= messageAuthorName;
+            }
         }
 
         // Only notify the thread of new messages if the chatResponse was successful to avoid inconsistent message state in the thread.
@@ -687,5 +695,22 @@ public sealed partial class ChatClientAgent : AIAgent
     }
 
     private string GetLoggingAgentName() => this.Name ?? "UnnamedAgent";
+
+    /// <summary>
+    /// Gets the author name to use for messages.
+    /// </summary>
+    /// <returns>
+    /// The agent's name, or "UnnamedAgent" if the name is not set, or <see langword="null"/> if
+    /// <see cref="ChatClientAgentOptions.SuppressAssistantName"/> is <see langword="true"/> and the agent has no name.
+    /// </returns>
+    private string? GetMessageAuthorName()
+    {
+        if (this._agentOptions?.SuppressAssistantName is true && this.Name is null)
+        {
+            return null;
+        }
+
+        return this.Name ?? "UnnamedAgent";
+    }
     #endregion
 }
--- a/dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgentOptions.cs
+++ b/dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgentOptions.cs
@@ -99,6 +99,18 @@ public class ChatClientAgentOptions
     public bool UseProvidedChatClientAsIs { get; set; }
 
     /// <summary>
+    /// Gets or sets a value indicating whether to suppress the name field on assistant messages when the agent name is not set.
+    /// </summary>
+    /// <remarks>
+    /// When set to <see langword="true"/>, assistant messages will not include a name field in the serialized payload
+    /// if the agent's <see cref="Name"/> property is <see langword="null"/>. This is useful for compatibility with
+    /// strict OpenAI-compatible backends that reject assistant messages with a name field.
+    /// When set to <see langword="false"/> (the default), the framework will use "UnnamedAgent" as a fallback name
+    /// when the agent name is not set, maintaining backward compatibility.
+    /// </remarks>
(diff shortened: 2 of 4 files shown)


T0mSIlver:

westey-m:
Thanks for the contribution.  The issue was addressed in a separate PR already.  See #1830
If you are still seeing issues, feel free to reopen the PR.

PR #1830, opened 2025-10-31
.NET: [BREAKING] Don't stamp unknown author name on output messages.

westey-m:
### Motivation and Context

We log "Unknown Agent" when an agent has no name, but we were also setting that as the author name on output messages when the Agent had no name, which is incorrect.

### Description

- Don't set "Unknown Agent" for logging on output messages.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
+++ b/dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
@@ -353,9 +353,9 @@ public sealed partial class ChatClientAgent : AIAgent
 
         chatClient = ApplyRunOptionsTransformations(options, chatClient);
 
-        var agentName = this.GetLoggingAgentName();
+        var loggingAgentName = this.GetLoggingAgentName();
 
-        this._logger.LogAgentChatClientInvokingAgent(nameof(RunAsync), this.Id, agentName, this._chatClientType);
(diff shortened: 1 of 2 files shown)


dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
```diff
@@ -378,7 +378,7 @@ private async Task<TAgentRunResponse> RunCoreAsync<TAgentRunResponse, TChatClien
         // Ensure that the author name is set for each message in the response.
         foreach (ChatMessage chatResponseMessage in chatResponse.Messages)
         {
-            chatResponseMessage.AuthorName ??= agentName;
+            chatResponseMessage.AuthorName ??= this.Name;
```
rogerbarreto, line 381: If have second thoughts on modifying / filling content that comes from server side.

Gathering agent information should come from the agent itself, or this correction should be handled by the caller if the backend don't represent it, right?

```suggestion
            chatResponseMessage.AuthorName ??= this.Name;
```

westey-m: This backfill is really for cases where there isn't a service agent.  For something like a Foundry Agent the author name would already be set by what we get back from the service, but if we are using ChatCompletion, there is no authorname, and here to be consistent it should be whatever the agent was constructed with, so I think this is the right behavior.

PR #1831, opened 2025-10-31
.NET: Background responses sample

SergeyMenshykh:
This PR adds a sample demonstrating how to use background responses with function calling and agent state persistence between polls.


dotnet/samples/GettingStarted/Agents/Agent_Step20_BackgroundResponsesWithToolsAndPersistence/Program.cs
```diff
@@ -0,0 +1,51 @@
...
+using Azure.AI.OpenAI;
+using Azure.Identity;
+using Microsoft.Agents.AI;
+using Microsoft.Extensions.AI;
+using OpenAI;
+
+var endpoint = Environment.GetEnvironmentVariable("AZURE_OPENAI_ENDPOINT") ?? throw new InvalidOperationException("AZURE_OPENAI_ENDPOINT is not set.");
+var deploymentName = Environment.GetEnvironmentVariable("AZURE_OPENAI_DEPLOYMENT_NAME") ?? "gpt-5";
+
+var stateStore = new AgentStateStore();
```
rogerbarreto: Suggest bringing the `AgentStateStore` impl, with PersistAgentStore being a LocalFunction as part of the `Program.cs`.

```suggestion
```

dotnet/samples/GettingStarted/Agents/Agent_Step20_BackgroundResponsesWithToolsAndPersistence/Program.cs
```diff
@@ -0,0 +1,51 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+// This sample demonstrates how to use background responses with ChatClientAgent and Azure OpenAI Responses for long-running operations.
+// It shows polling for completion using continuation tokens, function calling during background operations,
+// and persisting/restoring agent state between polling cycles.
+
+#pragma warning disable CA1050 // Declare types in namespaces
+
+using Azure.AI.OpenAI;
```
rogerbarreto, line 11: For new implementations, to not be in the same place we were before with `Azure.AI.OpenAI` package the lagged considerably (skipping 2.4.0) in relation to the `2.5.0` I would drop the dependency from `Azure.AI.OpenAI`.

Check how to do it using `OpenAI` directly.

https://github.com/dotnet/extensions/pull/6873/files#diff-7d09a128901ad93b237558aa85734f7d993adc594c181936f5c2efe08b5013ddR27-R38

```suggestion
```

dotnet/samples/GettingStarted/Agents/Agent_Step20_BackgroundResponsesWithToolsAndPersistence/Program.cs
SergeyMenshykh: I originally had it as local functions and a dictionary and tools in Program.cs, but it made the sample harder to read. So, I moved the state and tools to separate classes, and it looks simpler and easier to read. What are the concerns about having the store as a separate classe?

dotnet/samples/GettingStarted/Agents/Agent_Step20_BackgroundResponsesWithToolsAndPersistence/Program.cs
SergeyMenshykh: Taking into account that almost all samples depend on Azure.AI.OpenAI, I would suggest migrating them to OpenAI in one go in a separate PR.

dotnet/samples/GettingStarted/Agents/Agent_Step20_BackgroundResponsesWithToolsAndPersistence/README.md
```diff
@@ -0,0 +1,25 @@
...
+> **Note:** Background responses are currently only supported by OpenAI Responses.
+
+For more information, see the [official documentation](https://learn.microsoft.com/en-us/agent-framework/user-guide/agents/agent-background-responses?pivots=programming-language-csharp).
+
+# Prerequisites
+
+Before you begin, ensure you have the following prerequisites:
+
+- .NET 8.0 SDK or later
+- OpenAI api key
```
westey-m: Azure CLI Credentials?

dotnet/samples/GettingStarted/Agents/Agent_Step20_BackgroundResponsesWithToolsAndPersistence/Program.cs
rogerbarreto: IMHO, this counter intuitively also makes the sample harder to read (as now the developer needs to jump around other files to get the full understanding. Myself when I was reading the PR for a moment though there was a `AgentStateStore` component in the package. 

We moved away from this approach in benefit of end-to-end samples, that can potentially be used as a single file sample in .NET 10.

westey-m: Would it make sense to store the state in a file with a hardcoded name in the temp folder and also to check for it's existence on app startup.
That way, you could encourage the user to terminate the application and run it again, to pick up from where it left off, showing how this will continue to run even when the our app here isn't running.
I suppose we'd need a parameter to determine whether we want to start fresh or not though.
Feel free to disregard if this makes it too complex.

westey-m (approved):

dotnet/samples/GettingStarted/Agents/Agent_Step20_BackgroundResponsesWithToolsAndPersistence/Program.cs
SergeyMenshykh: > This counter intuitively also makes the sample harder to read (as now the developer needs to jump around other files to get the full understanding

I’m not sure I agree that it’s counterintuitive or harder to read. Modularity reduces cognitive load and has other benefits, like reusability, though it is not relevant to samples. It creates a clear, high-level scenario that is easy for developers to understand, while secondary details are moved to modules/classes. This allows developers to explore those details separately if they want to learn more.

> We moved away from this approach in benefit of end-to-end samples, that can potentially be used as a single file sample in .NET 10.

Right, agree, moving everything back the to the Program.cs.

SergeyMenshykh: Yes, that would probably make the sample a bit more complex than necessary, but it would better demonstrate the agent's continued operation from the point it left off. Let's keep it as it is and iterate on it if it's not clear.

westey-m (approved):

PR #1838, opened 2025-10-31
.NET: Add Cosmos DB implementations for ChatMessageStore and CheckpointStore.

TheovanKraay:
### Motivation and Context

This change adds Cosmos DB implementations for Agent Framework's storage abstractions to provide persistent storage options for chat messages and workflow checkpoints.

Partially addresses issue (.NET only): https://github.com/microsoft/agent-framework/issues/1390

### Description

Added Cosmos DB implementations for three core storage abstractions:

- **`CosmosChatMessageStore`** - Implements `ChatMessageStore` for chat message persistence
- **`CosmosCheckpointStore`** - Implements `JsonCheckpointStore` for workflow state persistence  

Added convenience extensions:
- **`CosmosDBChatExtensions`** - Extension methods for configuring chat agents with Cosmos DB
- **`CosmosDBWorkflowExtensions`** - Factory methods for creating checkpoint stores

Features implemented:
- Connection string and DefaultAzureCredential authentication
- Hierarchical partitioning (`/tenantId/userId/sessionId`) 
- Transactional batch operations with automatic splitting
- Configurable batch sizes and TTL settings
- State serialization/deserialization support

Added corresponding unit tests.

### Contribution Checklist

- [x] **Is this a breaking change?** No - this is a new feature addition

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
+++ b/dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
@@ -253,48 +253,37 @@ public sealed class CosmosChatMessageStore : ChatMessageStore, IDisposable
     }
 
     /// <summary>
-    /// Initializes a new instance of the <see cref="CosmosChatMessageStore"/> class from previously serialized state.
+    /// Creates a new instance of the <see cref="CosmosChatMessageStore"/> class from previously serialized state.
     /// </summary>
     /// <param name="cosmosClient">The <see cref="CosmosClient"/> instance to use for Cosmos DB operations.</param>
     /// <param name="serializedStoreState">A <see cref="JsonElement"/> representing the serialized state of the message store.</param>
     /// <param name="databaseId">The identifier of the Cosmos DB database.</param>
     /// <param name="containerId">The identifier of the Cosmos DB container.</param>
     /// <param name="jsonSerializerOptions">Optional settings for customizing the JSON deserialization process.</param>
+    /// <returns>A new instance of <see cref="CosmosChatMessageStore"/> initialized from the serialized state.</returns>
     /// <exception cref="ArgumentNullException">Thrown when <paramref name="cosmosClient"/> is null.</exception>
     /// <exception cref="ArgumentException">Thrown when the serialized state cannot be deserialized.</exception>
-    public CosmosChatMessageStore(CosmosClient cosmosClient, JsonElement serializedStoreState, string databaseId, string containerId, JsonSerializerOptions? jsonSerializerOptions = null)
+    public static CosmosChatMessageStore CreateFromSerializedState(CosmosClient cosmosClient, JsonElement serializedStoreState, string databaseId, string containerId, JsonSerializerOptions? jsonSerializerOptions = null)
     {
-        this._cosmosClient = cosmosClient ?? throw new ArgumentNullException(nameof(cosmosClient));
-        this.DatabaseId = Throw.IfNullOrWhitespace(databaseId);
-        this.ContainerId = Throw.IfNullOrWhitespace(containerId);
-        this._container = this._cosmosClient.GetContainer(databaseId, containerId);
-        this._ownsClient = false;
+        Throw.IfNull(cosmosClient);
+        Throw.IfNullOrWhitespace(databaseId);
+        Throw.IfNullOrWhitespace(containerId);
 
-        if (serializedStoreState.ValueKind is JsonValueKind.Object)
+        if (serializedStoreState.ValueKind is not JsonValueKind.Object)
         {
-            var state = JsonSerializer.Deserialize<StoreState>(serializedStoreState, jsonSerializerOptions);
-            if (state?.ConversationIdentifier is { } conversationId)
-            {
-                this.ConversationId = conversationId;
-
-                // Initialize hierarchical partitioning if available in state
-                this._tenantId = state.TenantId;
-                this._userId = state.UserId;
-                this._useHierarchicalPartitioning = state.UseHierarchicalPartitioning;
-
-                this._partitionKey = (this._useHierarchicalPartitioning && this._tenantId != null && this._userId != null)
-                    ? new PartitionKeyBuilder()
-                        .Add(this._tenantId)
-                        .Add(this._userId)
-                        .Add(conversationId)
-                        .Build()
-                    : new PartitionKey(conversationId);
-
-                return;
-            }
+            throw new ArgumentException("Invalid serialized state", nameof(serializedStoreState));
         }
 
-        throw new ArgumentException("Invalid serialized state", nameof(serializedStoreState));
+        var state = JsonSerializer.Deserialize<StoreState>(serializedStoreState, jsonSerializerOptions);
+        if (state?.ConversationIdentifier is not { } conversationId)
+        {
+            throw new ArgumentException("Invalid serialized state", nameof(serializedStoreState));
+        }
+
+        // Use the internal constructor with all parameters to ensure partition key logic is centralized
+        return state.UseHierarchicalPartitioning && state.TenantId != null && state.UserId != null
+            ? new CosmosChatMessageStore(cosmosClient, databaseId, containerId, conversationId, ownsClient: false, state.TenantId, state.UserId)
+            : new CosmosChatMessageStore(cosmosClient, databaseId, containerId, conversationId, ownsClient: false);
     }
 
     /// <inheritdoc />
@@ -320,7 +309,6 @@ public sealed class CosmosChatMessageStore : ChatMessageStore, IDisposable
         });
 
         var messages = new List<ChatMessage>();
-        int messageCount = 0;
 
         while (iterator.HasMoreResults)
         {
@@ -328,7 +316,7 @@ public sealed class CosmosChatMessageStore : ChatMessageStore, IDisposable
 
             foreach (var document in response)
             {
-                if (this.MaxMessagesToRetrieve.HasValue && messageCount >= this.MaxMessagesToRetrieve.Value)
+                if (this.MaxMessagesToRetrieve.HasValue && messages.Count >= this.MaxMessagesToRetrieve.Value)
                 {
                     break;
                 }
@@ -339,12 +327,11 @@ public sealed class CosmosChatMessageStore : ChatMessageStore, IDisposable
                     if (message != null)
                     {
                         messages.Add(message);
-                        messageCount++;
                     }
                 }
             }
 
-            if (this.MaxMessagesToRetrieve.HasValue && messageCount >= this.MaxMessagesToRetrieve.Value)
+            if (this.MaxMessagesToRetrieve.HasValue && messages.Count >= this.MaxMessagesToRetrieve.Value)
             {
                 break;
             }
@@ -412,18 +399,44 @@ public sealed class CosmosChatMessageStore : ChatMessageStore, IDisposable
     /// </summary>
     private async Task ExecuteBatchOperationAsync(List<ChatMessage> messages, long timestamp, CancellationToken cancellationToken)
     {
-        // Validate all messages will use the same partition key (required for transactional batch)
-        // In practice, this is guaranteed by conversationId, but validate defensively
-        if (messages.Count > 0 && !messages.All(m => true)) // All messages in this store share the same partition key by design
-        {
-            throw new InvalidOperationException("All messages in a transactional batch must share the same partition key (conversationId).");
-        }
-
-        var batch = this._container.CreateTransactionalBatch(this._partitionKey);
-
+        // Create all documents upfront for validation and batch operation
+        var documents = new List<CosmosMessageDocument>(messages.Count);
         foreach (var message in messages)
         {
-            var document = this.CreateMessageDocument(message, timestamp);
+            documents.Add(this.CreateMessageDocument(message, timestamp));
+        }
+
+        // Defensive check: Verify all messages share the same partition key values
+        // In hierarchical partitioning, this means same tenantId, userId, and sessionId
+        // In simple partitioning, this means same conversationId
+        if (documents.Count > 0)
+        {
+            if (this._useHierarchicalPartitioning)
+            {
+                // Verify all documents have matching hierarchical partition key components
+                var firstDoc = documents[0];
+                if (!documents.All(d => d.TenantId == firstDoc.TenantId && d.UserId == firstDoc.UserId && d.SessionId == firstDoc.SessionId))
+                {
+                    throw new InvalidOperationException("All messages in a batch must share the same partition key values (tenantId, userId, sessionId).");
+                }
+            }
+            else
+            {
+                // Verify all documents have matching conversationId
+                var firstConversationId = documents[0].ConversationId;
+                if (!documents.All(d => d.ConversationId == firstConversationId))
+                {
+                    throw new InvalidOperationException("All messages in a batch must share the same partition key value (conversationId).");
+                }
+            }
+        }
+
+        // All messages in this store share the same partition key by design
+        // Transactional batches require all items to share the same partition key
+        var batch = this._container.CreateTransactionalBatch(this._partitionKey);
+
+        foreach (var document in documents)
+        {
             batch.CreateItem(document);
         }
 
@@ -572,17 +585,16 @@ public sealed class CosmosChatMessageStore : ChatMessageStore, IDisposable
 
         var iterator = this._container.GetItemQueryIterator<string>(query, requestOptions: new QueryRequestOptions
         {
-            PartitionKey = new PartitionKey(this.ConversationId),
+            PartitionKey = this._partitionKey,
             MaxItemCount = this.MaxItemCount
         });
 
         var deletedCount = 0;
-        var partitionKey = new PartitionKey(this.ConversationId);
 
         while (iterator.HasMoreResults)
         {
             var response = await iterator.ReadNextAsync(cancellationToken).ConfigureAwait(false);
-            var batch = this._container.CreateTransactionalBatch(partitionKey);
+            var batch = this._container.CreateTransactionalBatch(this._partitionKey);
             var batchItemCount = 0;
 
             foreach (var itemId in response)
--- a/dotnet/tests/Microsoft.Agents.AI.CosmosNoSql.UnitTests/CosmosChatMessageStoreTests.cs
+++ b/dotnet/tests/Microsoft.Agents.AI.CosmosNoSql.UnitTests/CosmosChatMessageStoreTests.cs
@@ -640,7 +640,7 @@ public sealed class CosmosChatMessageStoreTests : IAsyncLifetime, IDisposable
         {
             TypeInfoResolver = new DefaultJsonTypeInfoResolver()
         };
-        using var deserializedStore = new CosmosChatMessageStore(cosmosClient, serializedState, s_testDatabaseId, HierarchicalTestContainerId, serializerOptions);
+        using var deserializedStore = CosmosChatMessageStore.CreateFromSerializedState(cosmosClient, serializedState, s_testDatabaseId, HierarchicalTestContainerId, serializerOptions);
 
         // Wait a moment for eventual consistency
         await Task.Delay(100);


dotnet/src/Microsoft.Agents.AI.Abstractions/CosmosAgentThread.cs
```diff
@@ -0,0 +1,233 @@
...
+/// in Azure Cosmos DB for durability, scalability, and cross-session availability. This approach provides
+/// reliable persistence while maintaining efficient access to conversation data.
+/// </para>
+/// <para>
+/// Cosmos threads persist conversation data across application restarts and can be shared across
+/// multiple application instances.
+/// </para>
+/// </remarks>
+[DebuggerDisplay("{DebuggerDisplay,nq}")]
+public abstract class CosmosAgentThread : AgentThread
```
westey-m: We shouldn't need a special AgentThread for CosmosDB.  Instead, everything related to cosmos ChatHistory storage should live in CosmosChatMessageStore so that it can be passed to a ChatClientAgent ChatMessageStoreFactory, similar to InMemoryChatMessageStore.

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
```diff
@@ -0,0 +1,853 @@
...
+using Azure.Identity;
+using Microsoft.Azure.Cosmos;
+using Microsoft.Extensions.AI;
+
+namespace Microsoft.Agents.AI;
+
+/// <summary>
+/// Provides a Cosmos DB implementation of the <see cref="ChatMessageStore"/> abstract class.
+/// </summary>
+public sealed class CosmosChatMessageStore : ChatMessageStore, IDisposable
```
westey-m, line 22: This should not live in Microft.Agents.AI.Abstractions since this library is a lightweight abstractions library that AIAgent implementation libraries can depend on without pulling in any unnecessary dependencies.
Instead it should really live in it's own package, e.g. `Microft.Agents.AI.Abstractions.CosmosNoSql` or `Microsoft.Azure.Cosmos.Agents` or `Microsoft.Azure.Cosmos.AgentFramework` or `Microsoft.Azure.Cosmos.Agents.AI`

dotnet/src/Microsoft.Agents.AI.Abstractions/CosmosChatMessageStore.cs
```diff
@@ -0,0 +1,853 @@
...
+    [SuppressMessage("Performance", "CA1812:Avoid uninstantiated internal classes", Justification = "Instantiated by Cosmos DB operations")]
+    private sealed class CosmosMessageDocument
+    {
+        [Newtonsoft.Json.JsonProperty("id")]
+        public string Id { get; set; } = string.Empty;
+
+        [Newtonsoft.Json.JsonProperty("conversationId")]
+        public string ConversationId { get; set; } = string.Empty;
+
+        [Newtonsoft.Json.JsonProperty(nameof(Timestamp))]
```
westey-m: Why are some of these pascal case (e.g. `nameof(Timestamp)` is `"Timestamp"`), while others are camel case (e.g. `"conversationId"`)

dotnet/src/Microsoft.Agents.AI.Abstractions/CosmosChatMessageStore.cs
```diff
@@ -0,0 +1,853 @@
...
+    private readonly PartitionKey _partitionKey;
+    private readonly bool _useHierarchicalPartitioning;
+
+    /// <summary>
+    /// Cached JSON serializer options for .NET 9.0 compatibility.
+    /// </summary>
+    private static readonly JsonSerializerOptions s_defaultJsonOptions = CreateDefaultJsonOptions();
+
+    [UnconditionalSuppressMessage("Trimming", "IL2026:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access", Justification = "JSON serialization is controlled")]
+    [UnconditionalSuppressMessage("AOT", "IL3050:Calling members annotated with 'RequiresDynamicCodeAttribute' may break functionality when AOT compiling", Justification = "JSON serialization is controlled")]
```
westey-m: Rather than suppressing these errors, we should instead be surfacing them so that callers who are using native AOT knows that we are using incompatible serialization techniques in the class.

That is of course assuming that we cannot make this Native AOT compatible.

e.g.
```csharp
    [RequiresUnreferencedCode("The CosmosChatMessageStore is currently incompatible with trimming.")]
    [RequiresDynamicCode("The CosmosChatMessageStore is currently incompatible with NativeAOT.")]
```

dotnet/src/Microsoft.Agents.AI.Abstractions/CosmosAgentThread.cs
TheovanKraay: Makes sense, will refactor

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
TheovanKraay: It seemed like I should take that approach, but InMemoryChatMessageStore made me think that implementations also live in here. In any case, will re-work.

dotnet/src/Microsoft.Agents.AI.Abstractions/CosmosChatMessageStore.cs
TheovanKraay: Will make it consistent.

dotnet/src/Microsoft.Agents.AI.Abstractions/CosmosChatMessageStore.cs
TheovanKraay: We can't make this native AOT compatible (I know, its a long standing ask for the SDK). Will surface the errors.

TheovanKraay: p.s. @westey-m for the build errors, they don't seem related to the changes, correct? Can I ignore these? Or should the PR be to a different branch?

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
westey-m: Apologies, just saw a typo in my message above, should have been `Microft.Agents.AI.CosmosNoSql` and not `Microft.Agents.AI.Abstractions.CosmosNoSql`. 😢

> InMemoryChatMessageStore made me think that implementations also live in here

Thanks for the call out.  We should reconsider having InMemoryChatMessageStore in abstractions.  The main reason why it had to be in Abstractions is now gone, and we may be able to push it down into core as well.
A few iterations ago, the base AgentThread class had a lot of the ChatClientAgentThread capabilities built in, which included defaulting to InMemoryChatMessageStore, and this has since been pushed down into the implementation package, but InMemoryChatMessageStore still lingers.

dotnet/src/Microsoft.Agents.AI.Abstractions.CosmosNoSql/CosmosChatMessageStore.cs
```diff
@@ -0,0 +1,853 @@
...
+        {
+            throw new ArgumentException("Container ID cannot be null or whitespace.", nameof(containerId));
+        }
+
+        if (string.IsNullOrWhiteSpace(conversationId))
+        {
+            throw new ArgumentException("Conversation ID cannot be null or whitespace.", nameof(conversationId));
+        }
+
+        this._cosmosClient = new CosmosClient(connectionString);
```
westey-m: We have some helpers in the code base that allows in-lining the param validation:
```suggestion
        this._cosmosClient = new CosmosClient(Throw.IfNullOrWhitespace(connectionString));
```

Looks like you are already including them via `<InjectSharedThrow>true</InjectSharedThrow>` in the project file.

dotnet/src/Microsoft.Agents.AI.Abstractions.CosmosNoSql/CosmosChatMessageStore.cs
```diff
@@ -0,0 +1,853 @@
...
+        {
+            throw new ArgumentException("Cannot be null or whitespace", nameof(conversationId));
+        }
+
+        if (!useManagedIdentity)
+        {
+            throw new ArgumentException("This constructor requires useManagedIdentity to be true. Use the connection string constructor for key-based authentication.", nameof(useManagedIdentity));
+        }
+
+        this._cosmosClient = new CosmosClient(accountEndpoint, new DefaultAzureCredential());
```
westey-m: Instead of defaulting to `DefaultAzureCredential` and requiring users to set useManagedIdentity to true, have you considered taking the credential as a parameter and removing useManagedIdentity?
It has the benefit of allowing the caller to choose their preferred credential type, and useManagedIdentity seems a bit odd since it's a bool that can only ever have one value.

dotnet/src/Microsoft.Agents.AI.Abstractions.CosmosNoSql/CosmosChatMessageStore.cs
westey-m: Switching to this should allow you to chain constructors more easily, e.g.
```csharp
public CosmosChatMessageStore(string connectionString, string databaseId, string containerId, string conversationId)
    : this(new CosmosClient(Throw.IfNullOrWhitespace(connectionString)), string databaseId, string containerId, string conversationId) {}
```
thereby reducing the amount of duplicated constructor code.  You may have to introduce an internal constructor with an ownsClient param, that all the others can call though.

dotnet/src/Microsoft.Agents.AI.Abstractions.CosmosNoSql/CosmosChatMessageStore.cs
```diff
@@ -0,0 +1,853 @@
...
+    /// Initializes a new instance of the <see cref="CosmosChatMessageStore"/> class from previously serialized state.
+    /// </summary>
+    /// <param name="serializedStoreState">A <see cref="JsonElement"/> representing the serialized state of the message store.</param>
+    /// <param name="cosmosClient">The <see cref="CosmosClient"/> instance to use for Cosmos DB operations.</param>
+    /// <param name="jsonSerializerOptions">Optional settings for customizing the JSON deserialization process.</param>
+    /// <exception cref="ArgumentNullException">Thrown when <paramref name="cosmosClient"/> is null.</exception>
+    /// <exception cref="ArgumentException">Thrown when the serialized state cannot be deserialized.</exception>
+    [UnconditionalSuppressMessage("Trimming", "IL2026:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access", Justification = "StoreState type is controlled and used for serialization")]
+    [UnconditionalSuppressMessage("AOT", "IL3050:Calling members annotated with 'RequiresDynamicCodeAttribute' may break functionality when AOT compiling", Justification = "StoreState type is controlled and used for serialization")]
```
westey-m: Do we still need this suppression now that the whole class is marked as incompatible?
```suggestion
```

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
```diff
@@ -0,0 +1,853 @@
...
+    /// <param name="cosmosClient">The <see cref="CosmosClient"/> instance to use for Cosmos DB operations.</param>
+    /// <param name="jsonSerializerOptions">Optional settings for customizing the JSON deserialization process.</param>
+    /// <exception cref="ArgumentNullException">Thrown when <paramref name="cosmosClient"/> is null.</exception>
+    /// <exception cref="ArgumentException">Thrown when the serialized state cannot be deserialized.</exception>
+    [UnconditionalSuppressMessage("Trimming", "IL2026:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access", Justification = "StoreState type is controlled and used for serialization")]
+    [UnconditionalSuppressMessage("AOT", "IL3050:Calling members annotated with 'RequiresDynamicCodeAttribute' may break functionality when AOT compiling", Justification = "StoreState type is controlled and used for serialization")]
+    public CosmosChatMessageStore(JsonElement serializedStoreState, CosmosClient cosmosClient, JsonSerializerOptions? jsonSerializerOptions = null)
```
westey-m: For other constructors that take CosmosClient it typically comes first in the param list.  Have you considered making it the same here?

```suggestion
    public CosmosChatMessageStore(CosmosClient cosmosClient, JsonElement serializedStoreState, JsonSerializerOptions? jsonSerializerOptions = null)
```

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
```diff
@@ -0,0 +1,853 @@
...
+        this._ownsClient = false;
+
+        if (serializedStoreState.ValueKind is JsonValueKind.Object)
+        {
+            var state = JsonSerializer.Deserialize<StoreState>(serializedStoreState, jsonSerializerOptions);
+            if (state?.ConversationIdentifier is { } conversationId && state.DatabaseIdentifier is { } databaseId && state.ContainerIdentifier is { } containerId)
+            {
+                this._conversationId = conversationId;
+                this._databaseId = databaseId;
+                this._containerId = containerId;
```
westey-m: I am wondering if we should store database id and container id with the serialized state.

The implication is that if the developer wanted to change the DB / container name that messages are stored in and they did some migration where all the messages were moved to the new DB / container, or they renamed either of those, any threads that are deserialized after the change would be still pointing at the old DB / container names.

On the other hand, if they wanted to use DBs and Containers for partitioning messages, they may want to store those in the serialized state, since each conversation may be in a different DB/container.

Thoughts?

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
```diff
@@ -0,0 +1,853 @@
...
+        if (this._disposed)
+        {
+            throw new ObjectDisposedException(GetType().FullName);
+        }
+#pragma warning restore CA1513
+
+        // Use type discriminator for efficient queries
+        var query = new QueryDefinition("SELECT * FROM c WHERE c.conversationId = @conversationId AND c.type = @type ORDER BY c.timestamp ASC")
+            .WithParameter("@conversationId", this._conversationId)
+            .WithParameter("@type", "ChatMessage");
```
westey-m: It would be useful to give developers a mechanism to set the maximum number of messages to fetch, so that if the conversation continues for a very long time, they have the ability to limit the number returned so as not to exceed the LLM context window.
Note that it would need to fetch the most recent x messages, even though we are returning them in ascending order.

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
```diff
@@ -0,0 +1,853 @@
...
+        }
+
+#pragma warning disable CA1513 // Use ObjectDisposedException.ThrowIf - not available on all target frameworks
+        if (this._disposed)
+        {
+            throw new ObjectDisposedException(GetType().FullName);
+        }
+#pragma warning restore CA1513
+
+        var messageList = messages.ToList();
```
westey-m: In many cases messages will already be an IReadOnlyCollection<ChatMessage>, so casting to it avoids an enumeration of the IEnumerable.
```suggestion
        var messageList = messages as IReadOnlyCollection<ChatMessage> ?? messages.ToList();
```

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
```diff
@@ -0,0 +1,853 @@
...
+    [UnconditionalSuppressMessage("Trimming", "IL2026:Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access", Justification = "ChatMessage serialization is controlled")]
+    [UnconditionalSuppressMessage("AOT", "IL3050:Calling members annotated with 'RequiresDynamicCodeAttribute' may break functionality when AOT compiling", Justification = "ChatMessage serialization is controlled")]
+    private CosmosMessageDocument CreateMessageDocument(ChatMessage message, long timestamp)
+    {
+        return new CosmosMessageDocument
+        {
+            Id = Guid.NewGuid().ToString(),
+            ConversationId = this._conversationId,
+            Timestamp = timestamp,
+            MessageId = message.MessageId ?? Guid.NewGuid().ToString(),
```
westey-m: What is the purpose of generating a Guid for message id when we don't have one?

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
```diff
@@ -0,0 +1,853 @@
...
+            Role = message.Role.Value ?? "unknown",
```
westey-m: Would null also suffice for unknown?

dotnet/src/Microsoft.Agents.AI.Abstractions/CosmosAgentThread.cs
westey-m: Thanks, updates look good.

dotnet/src/Microsoft.Agents.AI.Abstractions/CosmosChatMessageStore.cs
westey-m: These ones?
>Error: /home/runner/work/agent-framework/agent-framework/dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Extensions/AgentProviderExtensions.cs(14,43): error CS0234: The type or namespace name 'AI' does not exist in the namespace 'Microsoft.Azure' (are you missing an assembly reference?) [/home/runner/work/agent-framework/agent-framework/dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Microsoft.Agents.AI.Workflows.Declarative.csproj::TargetFramework=net9.0]

Yeah, those don't look related.  Maybe hit the update branch button below at a convenient juncture, and hopefully it should fix it, and if not I can help figure out what's wrong.

TheovanKraay:
@westey-m I tried to address all your comments, and hit merge from main but builds still failing. Do I need to rebase?

TheovanKraay:
@westey-m I believe I have addressed all your comments. 

Builds are passing now. To do this I had to: 
- push a change to `dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Extensions/AgentProviderExtensions.cs` that was in main instead of rebasing which was taking a long time
- add CI pipelines but skip Linux tests (we do the same in .NET SDK for Cosmos DB as emulator does not run well in Linux)
- add editorconfig to suppress some false positives

dotnet/src/Microsoft.Agents.AI.Workflows/Microsoft.Agents.AI.Workflows.csproj
```diff
@@ -31,6 +31,9 @@
 
   <ItemGroup>
     <PackageReference Include="OpenTelemetry.Api" />
+    <PackageReference Include="Azure.Identity" />
+    <PackageReference Include="Microsoft.Azure.Cosmos" />
+    <PackageReference Include="Newtonsoft.Json" />
```
westey-m: We shouldn't need these new references here now that the new code is in its own package.
```suggestion
```

dotnet/src/Microsoft.Agents.AI/Microsoft.Agents.AI.csproj
```diff
@@ -22,6 +22,9 @@
     <PackageReference Include="Microsoft.Extensions.VectorData.Abstractions" />
     <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" />
     <PackageReference Include="System.Diagnostics.DiagnosticSource" />
+    <PackageReference Include="Azure.Identity" />
+    <PackageReference Include="Microsoft.Azure.Cosmos" />
+    <PackageReference Include="Newtonsoft.Json" />
```
westey-m: We shouldn't need these new references here now that the new code is in its own package.
```suggestion
```

dotnet/tests/Microsoft.Agents.AI.CosmosNoSql.UnitTests/.editorconfig
```diff
@@ -0,0 +1,9 @@
+# EditorConfig overrides for Cosmos DB Unit Tests
+# Multi-targeting (net472 + net9.0) causes false positives for IDE0005 (unnecessary using directives)
+
+root = false
+
+[*.cs]
+# Suppress IDE0005 for this project - multi-targeting causes false positives
+# These using directives ARE necessary but appear unnecessary in one target framework
+dotnet_diagnostic.IDE0005.severity = none
```
westey-m, line 9: Not a big deal but you can also add conditional compilation to only include the using clause for the target framework version that needs the clause.

dotnet/tests/Microsoft.Agents.AI.UnitTests/Microsoft.Agents.AI.UnitTests.csproj
```diff
@@ -18,6 +18,9 @@
     <PackageReference Include="OpenTelemetry" />
     <PackageReference Include="OpenTelemetry.Exporter.InMemory" />
     <PackageReference Include="System.Linq.Async" />
+    <PackageReference Include="Azure.Identity" />
+    <PackageReference Include="Microsoft.Azure.Cosmos" />
+    <PackageReference Include="Newtonsoft.Json" />
```
westey-m: Should be able to revert these changes as well.
```suggestion
```

dotnet/tests/Microsoft.Agents.AI.Abstractions.CosmosNoSql.UnitTests/CosmosChatMessageStoreTests.cs
westey-m, line 1: Looks like this file is a dupe left over after the rename?  Can we remove it?

dotnet/tests/Microsoft.Agents.AI.Abstractions.CosmosNoSql.UnitTests/CosmosCheckpointStoreTests.cs
westey-m, line 1: Looks like this file is a dupe left over after the rename?  Can we remove it?

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosDBChatExtensions.cs
```diff
@@ -0,0 +1,140 @@
...
+
+        if (string.IsNullOrWhiteSpace(databaseId))
+        {
+            throw new ArgumentException("Cannot be null or whitespace", nameof(databaseId));
+        }
+
+        if (string.IsNullOrWhiteSpace(containerId))
+        {
+            throw new ArgumentException("Cannot be null or whitespace", nameof(containerId));
+        }
```
westey-m: Considering that the CosmosChatMessageStore constructor will already check these params, do we need the additional checks here too?

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
```diff
@@ -0,0 +1,666 @@
...
+            {
+                if (this.MaxMessagesToRetrieve.HasValue && messageCount >= this.MaxMessagesToRetrieve.Value)
+                {
+                    break;
+                }
+
+                if (!string.IsNullOrEmpty(document.Message))
+                {
+                    var message = JsonSerializer.Deserialize<ChatMessage>(document.Message, s_defaultJsonOptions);
+                    if (message != null)
```
kirankumarkolli, line 327: Is it expected?

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
```diff
@@ -0,0 +1,666 @@
...
+#pragma warning disable CA1513 // Use ObjectDisposedException.ThrowIf - not available on all target frameworks
+        if (this._disposed)
+        {
+            throw new ObjectDisposedException(this.GetType().FullName);
+        }
+#pragma warning restore CA1513
+
+        // Fetch most recent messages in descending order when limit is set, then reverse to ascending
+        var orderDirection = this.MaxMessagesToRetrieve.HasValue ? "DESC" : "ASC";
+        var query = new QueryDefinition($"SELECT * FROM c WHERE c.conversationId = @conversationId AND c.type = @type ORDER BY c.timestamp {orderDirection}")
```
kirankumarkolli, line 301: Is the semantics always get TOP X ORDER_BY with no continutions?

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
```diff
@@ -0,0 +1,666 @@
...
+            await this.ExecuteBatchOperationAsync(batchMessages, currentTimestamp, cancellationToken).ConfigureAwait(false);
+        }
+    }
+
+    /// <summary>
+    /// Executes a single batch operation with retry logic and enhanced error handling.
+    /// </summary>
+    private async Task ExecuteBatchOperationAsync(List<ChatMessage> messages, long timestamp, CancellationToken cancellationToken)
+    {
+        var batch = this._container.CreateTransactionalBatch(this._partitionKey);
```
kirankumarkolli, line 436: assert that PK is same across all? (BE might fail if different)

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
```diff
@@ -0,0 +1,666 @@
...
+            var firstHalf = messages.Take(midpoint).ToList();
+            var secondHalf = messages.Skip(midpoint).ToList();
+
+            await this.ExecuteBatchOperationAsync(firstHalf, timestamp, cancellationToken).ConfigureAwait(false);
+            await this.ExecuteBatchOperationAsync(secondHalf, timestamp, cancellationToken).ConfigureAwait(false);
+        }
+        catch (CosmosException ex) when (ex.StatusCode == (System.Net.HttpStatusCode)429) // TooManyRequests
+        {
+            // Handle rate limiting with exponential backoff
+            await Task.Delay(TimeSpan.FromMilliseconds(ex.RetryAfter?.TotalMilliseconds ?? 1000), cancellationToken).ConfigureAwait(false);
```
kirankumarkolli: SDK throttling retry policy do takes care of it. Is application level retry extra on-top?

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
```diff
@@ -0,0 +1,666 @@
...
+    /// </summary>
+    private async Task ExecuteBatchOperationAsync(List<ChatMessage> messages, long timestamp, CancellationToken cancellationToken)
+    {
+        var batch = this._container.CreateTransactionalBatch(this._partitionKey);
+
+        foreach (var message in messages)
+        {
+            var document = this.CreateMessageDocument(message, timestamp);
+            batch.CreateItem(document);
+        }
```
kirankumarkolli, line 441: While adding item it-self might fail with `RequestEntityTooLarge`

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
```diff
@@ -0,0 +1,666 @@
...
+                await this.AddSingleMessageAsync(messages[0], cancellationToken).ConfigureAwait(false);
+                return;
+            }
+
+            // Split the batch in half and retry
+            var midpoint = messages.Count / 2;
+            var firstHalf = messages.Take(midpoint).ToList();
+            var secondHalf = messages.Skip(midpoint).ToList();
+
+            await this.ExecuteBatchOperationAsync(firstHalf, timestamp, cancellationToken).ConfigureAwait(false);
```
kirankumarkolli, line 466: So transactionality is not a requirement but batching.

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
```diff
@@ -0,0 +1,666 @@
...
+            PartitionKey = this._partitionKey
+        });
+
+        if (iterator.HasMoreResults)
+        {
+            var response = await iterator.ReadNextAsync(cancellationToken).ConfigureAwait(false);
+            return response.FirstOrDefault();
+        }
+
+        return 0;
```
kirankumarkolli: Dead code, pattern wise right way to do it though

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
```diff
@@ -0,0 +1,666 @@
...
+        public string? TenantId { get; set; }
+        public string? UserId { get; set; }
+        public bool UseHierarchicalPartitioning { get; set; }
+    }
+
+    /// <summary>
+    /// Represents a document stored in Cosmos DB for chat messages.
+    /// </summary>
+    [SuppressMessage("Performance", "CA1812:Avoid uninstantiated internal classes", Justification = "Instantiated by Cosmos DB operations")]
+    private sealed class CosmosMessageDocument
```
Pilchie: One thing I've done in the past in Semantic Kernel, was forcing the `CosmosClient` to use System.Text.Json. Of course, it means you have to  control the settings that go to the client.

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
TheovanKraay: Do you mean is null expected?

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
TheovanKraay: Good shout, we should add continuations, will add.

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
TheovanKraay: Ok will remove

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
TheovanKraay: No transactionality is a requirement here, why do you ask? Implementation is wrong?

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
TheovanKraay: I will add this for belt and braces, but all messages should have same conversationId

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
TheovanKraay: With that said @kirankumarkolli , the [GetMessagesAsync](vscode-file://vscode-app/c:/Users/thvankra/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) doesn't support them in it's contract, so it has to be internal.... and `iterator.HasMoreResults` already uses them internally, correct?

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
TheovanKraay: Will add explicit Exception

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
TheovanKraay: Or you are asking about auto-generated ConversationId? Yes, this is intentional and (I think) follows a correct pattern. It's so users don't need to specify conversation ID up front. Users still have full control - `ConversationId` property is public with an init accessor, so auto-generated ID is immediately accessible via `messageStore.ConversationId` after construction. This pattern is consistent with `InMemoryChatMessageStore` and provides flexibility: users can either provide explicit IDs (for resuming conversations or controlled scenarios) or use auto-generation for simplicity while still being able to retrieve and use the generated ID for subsequent operations.

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
TheovanKraay: re-worked this

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
```diff
@@ -0,0 +1,676 @@
...
+    {
+#pragma warning disable CA1513 // Use ObjectDisposedException.ThrowIf - not available on all target frameworks
+        if (this._disposed)
+        {
+            throw new ObjectDisposedException(this.GetType().FullName);
+        }
+#pragma warning restore CA1513
+
+        // Batch delete for efficiency
+        var query = new QueryDefinition("SELECT VALUE c.id FROM c WHERE c.conversationId = @conversationId AND c.Type = @type")
```
kirankumarkolli, line 582: Can Delete PK be leveraged instead?

kirankumarkolli (approved):

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
kirankumarkolli: Assuming all are chat messages

westey-m (approved):

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosCheckpointStore.cs
```diff
@@ -0,0 +1,279 @@
...
+            return document.RootElement.Clone();
+        }
+        catch (CosmosException ex) when (ex.StatusCode == System.Net.HttpStatusCode.NotFound)
+        {
+            throw new InvalidOperationException($"Checkpoint with ID '{key.CheckpointId}' for run '{runId}' not found.");
+        }
+    }
+
+    /// <inheritdoc />
+    public override async ValueTask<IEnumerable<CheckpointInfo>> RetrieveIndexAsync(string runId, CheckpointInfo? withParent = null)
```
ReubenBond, line 164: Unrelated to this PR, but we should flow `CancellationToken` through all of these methods. cc @lokitoth

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosDBChatExtensions.cs
```diff
@@ -0,0 +1,95 @@
...
+    /// </summary>
+    /// <param name="options">The chat client agent options to configure.</param>
+    /// <param name="accountEndpoint">The Cosmos DB account endpoint URI.</param>
+    /// <param name="databaseId">The identifier of the Cosmos DB database.</param>
+    /// <param name="containerId">The identifier of the Cosmos DB container.</param>
+    /// <returns>The configured <see cref="ChatClientAgentOptions"/>.</returns>
+    /// <exception cref="ArgumentNullException">Thrown when <paramref name="options"/> is null.</exception>
+    /// <exception cref="ArgumentException">Thrown when any string parameter is null or whitespace.</exception>
+    [RequiresUnreferencedCode("The CosmosChatMessageStore uses JSON serialization which is incompatible with trimming.")]
+    [RequiresDynamicCode("The CosmosChatMessageStore uses JSON serialization which is incompatible with NativeAOT.")]
```
ReubenBond, line 53:
```suggestion
    [RequiresUnreferencedCode("The CosmosChatMessageStore uses JSON serialization that is incompatible with trimming.")]
    [RequiresDynamicCode("The CosmosChatMessageStore uses JSON serialization that is incompatible with NativeAOT.")]
```

Similar above... it's a very minor nit so feel free to ignore. JSON serialization in general is trimming compatible (eg System.Text.Json), but the serialization used by default in Cosmos DB (Newtonsoft.Json) is not. Can Cosmos DB be configured to use STJ instead so we can remove this suppression?

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/Microsoft.Agents.AI.CosmosNoSql.csproj
```diff
@@ -0,0 +1,41 @@
...
+    <InjectIsExternalInitOnLegacy>true</InjectIsExternalInitOnLegacy>
+    <InjectRequiredMemberOnLegacy>true</InjectRequiredMemberOnLegacy>
+    <InjectCompilerFeatureRequiredOnLegacy>true</InjectCompilerFeatureRequiredOnLegacy>
+  </PropertyGroup>
+
+  <Import Project="$(RepoRoot)/dotnet/nuget/nuget-package.props" />
+
+  <PropertyGroup>
+    <!-- NuGet Package Settings -->
+    <Title>Microsoft Agent Framework Cosmos DB NoSQL Integration</Title>
```
ReubenBond, line 23: The naming feels odd to me, but it seems like Cosmos DB is branding its document storage API as Cosmos DB NoSQL? Should we have 'DB' in the package id?

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
```diff
@@ -0,0 +1,676 @@
...
+                    break;
+                }
+
+                if (!string.IsNullOrEmpty(document.Message))
+                {
+                    var message = JsonSerializer.Deserialize<ChatMessage>(document.Message, s_defaultJsonOptions);
+                    if (message != null)
+                    {
+                        messages.Add(message);
+                        messageCount++;
```
ReubenBond: Why not `messages.Count`?

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
```diff
@@ -0,0 +1,676 @@
...
+
+    /// <summary>
+    /// Executes a single batch operation with enhanced error handling.
+    /// Cosmos SDK handles throttling (429) retries automatically.
+    /// </summary>
+    private async Task ExecuteBatchOperationAsync(List<ChatMessage> messages, long timestamp, CancellationToken cancellationToken)
+    {
+        // Validate all messages will use the same partition key (required for transactional batch)
+        // In practice, this is guaranteed by conversationId, but validate defensively
+        if (messages.Count > 0 && !messages.All(m => true)) // All messages in this store share the same partition key by design
```
ReubenBond: I don't understand the purpose of `m => true` here

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
```diff
@@ -0,0 +1,676 @@
...
+            .WithParameter("@type", "ChatMessage");
+
+        var iterator = this._container.GetItemQueryIterator<string>(query, requestOptions: new QueryRequestOptions
+        {
+            PartitionKey = new PartitionKey(this.ConversationId),
+            MaxItemCount = this.MaxItemCount
+        });
+
+        var deletedCount = 0;
+        var partitionKey = new PartitionKey(this.ConversationId);
```
ReubenBond: Should this use `_partitionKey` to account for hierarchical keys?

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
```diff
@@ -0,0 +1,676 @@
...
+    /// <param name="cosmosClient">The <see cref="CosmosClient"/> instance to use for Cosmos DB operations.</param>
+    /// <param name="serializedStoreState">A <see cref="JsonElement"/> representing the serialized state of the message store.</param>
+    /// <param name="databaseId">The identifier of the Cosmos DB database.</param>
+    /// <param name="containerId">The identifier of the Cosmos DB container.</param>
+    /// <param name="jsonSerializerOptions">Optional settings for customizing the JSON deserialization process.</param>
+    /// <exception cref="ArgumentNullException">Thrown when <paramref name="cosmosClient"/> is null.</exception>
+    /// <exception cref="ArgumentException">Thrown when the serialized state cannot be deserialized.</exception>
+    public CosmosChatMessageStore(CosmosClient cosmosClient, JsonElement serializedStoreState, string databaseId, string containerId, JsonSerializerOptions? jsonSerializerOptions = null)
+    {
+        this._cosmosClient = cosmosClient ?? throw new ArgumentNullException(nameof(cosmosClient));
```
ReubenBond: Would `Throw.IfNull` work here?

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
```diff
@@ -0,0 +1,676 @@
...
+    /// Initializes a new instance of the <see cref="CosmosChatMessageStore"/> class from previously serialized state.
+    /// </summary>
+    /// <param name="cosmosClient">The <see cref="CosmosClient"/> instance to use for Cosmos DB operations.</param>
+    /// <param name="serializedStoreState">A <see cref="JsonElement"/> representing the serialized state of the message store.</param>
+    /// <param name="databaseId">The identifier of the Cosmos DB database.</param>
+    /// <param name="containerId">The identifier of the Cosmos DB container.</param>
+    /// <param name="jsonSerializerOptions">Optional settings for customizing the JSON deserialization process.</param>
+    /// <exception cref="ArgumentNullException">Thrown when <paramref name="cosmosClient"/> is null.</exception>
+    /// <exception cref="ArgumentException">Thrown when the serialized state cannot be deserialized.</exception>
+    public CosmosChatMessageStore(CosmosClient cosmosClient, JsonElement serializedStoreState, string databaseId, string containerId, JsonSerializerOptions? jsonSerializerOptions = null)
```
ReubenBond: An alternative is to have a static `Create` method which deserializes the state and calls the ctor with the values. That ensures that the partition key building logic, etc, is centralized and there is no room for divergence.

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosDBChatExtensions.cs
TheovanKraay: Unfortunately not, we have a hard Newtonsoft.Json dependency (looking to remove in next major version, and in fact about to add docs to flag that Newtonsoft.Json dependencies need to be explicitly managed).

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/Microsoft.Agents.AI.CosmosNoSql.csproj
TheovanKraay: I am ok with the naming. The full official branding name for the database is actually "Microsoft Azure Cosmos DB" so really it should be "MicrosoftAzureCosmosDBNoSQLAPI". We usually shorten to CosmosNoSQL, or similar :-)

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
TheovanKraay: Good shout, will change.

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
TheovanKraay: Ah, this was meant to be  defensive check that all pk values match, but I never implemented (partly as in practice it will be redundant). Will review.

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
TheovanKraay: Yes, good spot!

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
TheovanKraay: yes, will add

dotnet/src/Microsoft.Agents.AI.CosmosNoSql/CosmosChatMessageStore.cs
TheovanKraay: Good shout, will implement.

PR #1846, opened 2025-10-31
.NET: Change GetAIAgentAsync to be synchronous

stephentoub:


ReubenBond (approved):

rogerbarreto (approved):
@stephentoub Why can't we have both? Async and Synchronous?

This is normally what we have for other packages.

Stopped the merge just to confirm. Before going forward.

SergeyMenshykh (approved):

stephentoub:
> @stephentoub Why can't we have both? Async and Synchronous?

Why? What would the asynchronous one do, @rogerbarreto?

rogerbarreto:
> Why? What would the asynchronous one do, @rogerbarreto?

I just notice now that there's no `await` operation within the method, thought there was a dependency there. No issues.

PR #1854, opened 2025-11-02
.NET: Python: [Feature Branch] Azure AI Responses client

dmytrostruk:
### Motivation and Context

Implemented Azure AI responses client with initial basic and thread management examples.

### Contribution Checklist

### DIFF
--- a/python/packages/azure-ai/agent_framework_azure_ai/_chat_client_v2.py
+++ b/python/packages/azure-ai/agent_framework_azure_ai/_chat_client_v2.py
@@ -247,7 +247,7 @@ class AzureAIAgentClientV2(OpenAIBaseResponsesClient):
             await self.project_client.close()
 
     def _prepare_input(self, messages: MutableSequence[ChatMessage]) -> tuple[list[ChatMessage], str | None]:
-        """Prepares input from messages and converts system/developer messages to instructions."""
+        """Prepare input from messages and convert system/developer messages to instructions."""
         result: list[ChatMessage] = []
         instructions_list: list[str] = []
         instructions: str | None = None


Zochory:
Hello a small feedback to maybe avoid confusion, in the title it is mention Azure AI Response Client, but in the PR, there is for example a "AzureAIAgentClientV2". that is added, which can be confusing due to the different type of API/SDK ;
 -Azure AI Agents SDKs,
-  Foundry Agent Service and the 
-Azure OpenAI Response Format


If we stick to the documentation , it recommends using the OpenAI() client with the responses.create() endpoint for stateful interaction and thread management.
https://learn.microsoft.com/en-us/azure/ai-foundry/openai/api-version-lifecycle?tabs=python#api-evolution

Also worth mentionning :
_Microsoft Entra ID:

 Important

Handling automatic token refresh was previously handled through use of the AzureOpenAI() client. The v1 API removes this dependency, by adding automatic token refresh support to the OpenAI() client._

https://learn.microsoft.com/en-us/azure/ai-foundry/openai/api-version-lifecycle?tabs=python#api-evolution

other ressources  : 

Azure AI Agent Client (Python) : https://learn.microsoft.com/en-us/python/api/overview/azure/ai-agents-readme?view=azure-python-preview&preserve-view=true 

Azure OpenAI Response API v1 GA - 
https://learn.microsoft.com/en-us/azure/ai-foundry/openai/latest#create-response


Azure OpenAI v1 REST API : https://learn.microsoft.com/en-us/azure/ai-foundry/openai/latest

OpenAPI v1 https://github.com/Azure/azure-rest-api-specs/blob/main/specification/ai/data-plane/OpenAI.v1/azure-v1-v1-generated.json

python/packages/azure-ai/agent_framework_azure_ai/_chat_client.py
```diff
@@ -256,15 +258,17 @@ def __init__(
...
+    ) -> None:
         """Use this method to setup tracing in your Azure AI Project.
 
         This will take the connection string from the project project_client.
         It will override any connection string that is set in the environment variables.
         It will disable any OTLP endpoint that might have been set.
         """
         try:
-            conn_string = await self.project_client.telemetry.get_application_insights_connection_string()
+            conn_string = await project_client.telemetry.get_application_insights_connection_string()
```
eavanvalkenburg: if we can't get the telemetry from the agentClient, then we should just remove this method altogether, because you would still have to manually import the project client, so the point of this method is kind of gone

python/packages/azure-ai/agent_framework_azure_ai/_chat_client.py
```diff
@@ -43,6 +43,7 @@
 from agent_framework._pydantic import AFBaseSettings
 from agent_framework.exceptions import ServiceInitializationError, ServiceResponseException
 from agent_framework.observability import use_observability
+from azure.ai.agents.aio import AgentsClient
```
eavanvalkenburg, line 45: could you add a note on why this change was needed in addition to adding the v2 version?

python/packages/azure-ai/agent_framework_azure_ai/_chat_client_v2.py
```diff
@@ -0,0 +1,365 @@
...
+if sys.version_info >= (3, 11):
+    from typing import Self  # pragma: no cover
+else:
+    from typing_extensions import Self  # pragma: no cover
+
+
+logger = get_logger("agent_framework.azure")
+
+
+class AzureAISettings(AFBaseSettings):
```
eavanvalkenburg: could we use the same one across? maybe move it to a `_shared.py`?

python/packages/azure-ai/agent_framework_azure_ai/_chat_client_v2.py
```diff
@@ -0,0 +1,365 @@
...
+                Can also be set via environment variable AZURE_AI_MODEL_DEPLOYMENT_NAME.
+            async_credential: Azure async credential to use for authentication.
+            env_file_path: Path to environment file for loading settings.
+            env_file_encoding: Encoding of the environment file.
+            kwargs: Additional keyword arguments passed to the parent class.
+
+        Examples:
+            .. code-block:: python
+
+                from agent_framework_azure_ai import AzureAIAgentClient
```
eavanvalkenburg:
```suggestion
                from agent_framework.azure import AzureAIAgentClientV2
```

python/packages/azure-ai/agent_framework_azure_ai/_chat_client_v2.py
```diff
@@ -0,0 +1,365 @@
...
+
+        store = run_options.get("store", False)
+
+        if store:
+            conversation_id = await self._get_conversation_id_or_create(run_options)
+            run_options["conversation"] = conversation_id
+
+        run_options["extra_body"] = {"agent": agent_reference}
+
+        # Remove properties that are not supported:
```
eavanvalkenburg: maybe note that they are captured in the agent setup

python/packages/core/agent_framework/openai/_responses_client.py
```diff
@@ -89,23 +89,24 @@ async def _inner_get_response(
         chat_options: ChatOptions,
         **kwargs: Any,
     ) -> ChatResponse:
-        options_dict = self._prepare_options(messages, chat_options)
+        client = await self.ensure_client()
+        run_options = await self.prepare_options(messages, chat_options)
         try:
             if not chat_options.response_format:
-                response = await self.client.responses.create(
+                response = await client.responses.create(
```
eavanvalkenburg, line 96: why not just do self.client? it should point to the same object?

python/packages/core/agent_framework/openai/_responses_client.py
```diff
@@ -70,7 +70,7 @@
 
 logger = get_logger("agent_framework.openai")
 
-__all__ = ["OpenAIResponsesClient"]
+__all__ = ["OpenAIBaseResponsesClient", "OpenAIResponsesClient"]
```
eavanvalkenburg: `OpenAIBaseResponsesClient` shouldn't be in all, people can still import this directly from this file, but when not needed we shouldn't expose a base client like this one.

python/packages/azure-ai/agent_framework_azure_ai/_chat_client_v2.py
```diff
@@ -0,0 +1,365 @@
...
+    ChatOptions,
+    TextContent,
+    get_logger,
+    use_chat_middleware,
+    use_function_invocation,
+)
+from agent_framework._pydantic import AFBaseSettings
+from agent_framework.exceptions import ServiceInitializationError
+from agent_framework.observability import use_observability
+from agent_framework.openai import OpenAIBaseResponsesClient
```
eavanvalkenburg: See my note, even though mypy may complain about this, a base class like that shouldn't be part of the public API, so just import from the actual file.
```suggestion
from agent_framework.openai._responses_client import OpenAIBaseResponsesClient
```

python/packages/azure-ai/agent_framework_azure_ai/_chat_client_v2.py
```diff
@@ -0,0 +1,365 @@
...
+    async def __aexit__(self, exc_type: type[BaseException] | None, exc_val: BaseException | None, exc_tb: Any) -> None:
+        """Async context manager exit."""
+        await self.close()
+
+    async def close(self) -> None:
+        """Close the project_client."""
+        await self._close_client_if_needed()
+
+    @classmethod
+    def from_settings(cls: type[TAzureAIAgentClient], settings: dict[str, Any]) -> TAzureAIAgentClient:
```
eavanvalkenburg: I think this method will be replaced by the declarative stuff, so let's remove it from new ones.

python/packages/azure-ai/agent_framework_azure_ai/_chat_client.py
dmytrostruk: Removed the method.

python/packages/azure-ai/agent_framework_azure_ai/_chat_client.py
dmytrostruk: As per guidance from responsible team, to keep V1 working with V2 package, `AgentsClient` needs to be used instead of `AIProjectClient` and `AgentsClient` can't be created from `AIProjectClient`. While in V2 implementation - `AIProjectClient` needs to be used.

python/packages/azure-ai/agent_framework_azure_ai/_chat_client_v2.py
dmytrostruk: Moved it to `_shared.py`.

python/packages/core/agent_framework/openai/_responses_client.py
dmytrostruk: In our `OpenAIBase` implementation `client: AsyncOpenAI` parameter in constructor is non-nullable, but I had to switch it to be nullable, because in V2 library the initialization of OpenAI client is asynchronous, so we shouldn't initialize it in constructor anymore. 

If I use just `self.client` here, there will be a warning that `self.client` may be `None`, so I have to either suppress it or use it differently. I decided to introduce `self.ensure_client()` method, which will always return non-nullable client instance, and if it's not initialized - it will try to initialize it asynchronously.

Zochory:
I mean the official terminology is v1 for the APIs Azure OpenAI but not only, also for the Foundry AI Agents in the microsoft documentation, might be kind of confusing to provide in agent-framework an azure ai client v2  imho

Also the Response API (OpenAI Response format) is for now in Azure, only for the Azure OpenAI models, having a microsoft entra id auth as example that would use from openai import OpenAI  seems simplier and fully in accordance with potential future changes in azure openai 

image

also the agent-framework-core -mem0 -azure-ai-evaluation are already  install a openai 



https://learn.microsoft.com/en-us/azure/ai-foundry/openai/latest

dmytrostruk:
> might be kind of confusing to provide in agent-framework an azure ai client v2 imho

@Zochory Sorry, I'm not sure I understand your comment, but current naming is temporary.

dmytrostruk:
Will create another PR, since I would like to change base branch and keep it separately for Python.

dmytrostruk:
I would like to change base branch to keep Python and .NET changes independently, so will create another PR.

Zochory:
> > might be kind of confusing to provide in agent-framework an azure ai client v2 imho
> 
> @Zochory Sorry, I'm not sure I understand your comment, but current naming is temporary.

If I understand correctly, the purpose of this PR is to add an OpenAI Response API client supporting both Azure OpenAI and Foundry Agent, correct?
The _chat_response_v2.py file appears to use OpenAIResponse with the OpenAI SDK. I believe similar functionality is already available in the repository (for example, in [azure_responses_client_basic.py](https://github.com/microsoft/agent-framework/blob/main/python/samples/getting_started/agents/azure_openai/azure_responses_client_basic.py) and the openai_response_client).

Additionally, this format is not used in the same way as the Azure AI Agents or the current implementation.
If I haven't missed anything, it seems that azure.identity is not utilized here—instead, it uses from azure.core.credentials_async import AsyncTokenCredential.

For further information, see:
[Azure OpenAI Responses Agent User Guide](https://learn.microsoft.com/en-us/agent-framework/user-guide/agents/agent-types/azure-openai-responses-agent?pivots=programming-language-python)

[Foundry Models Responses Agent User Guide](https://learn.microsoft.com/en-us/agent-framework/user-guide/agents/agent-types/azure-ai-foundry-models-responses-agent?pivots=programming-language-python)

the other links above

PR #1867, opened 2025-11-03
.NET: Chathistory memory provider add

westey-m:
### Motivation and Context

#1680

### Description

- Add an AIContextProvider that stores chat history in a vector store and can find related messages for context later.

### Contribution Checklist


dotnet/src/Microsoft.Agents.AI.VectorDataMemory/Microsoft.Agents.AI.VectorDataMemory.csproj
```diff
@@ -0,0 +1,38 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
markwallace-microsoft: What about just `Microsoft.Agents.AI.VectorData` as a name. We generally use that naming convention for other integrations e.g. A2A, AG-UI, ...

dotnet/src/Microsoft.Agents.AI.VectorDataMemory/ChatHistoryMemoryProvider.cs
```diff
@@ -0,0 +1,530 @@
...
+/// </para>
+/// <para>
+/// Behavior is configurable through <see cref="ChatHistoryMemoryProviderOptions"/>. When
+/// <see cref="ChatHistoryMemoryProviderOptions.SearchBehavior.OnDemandFunctionCalling"/> is selected the provider
+/// exposes a function tool that the model can invoke to retrieve relevant memories on demand instead of
+/// injecting them automatically on each invocation.
+/// </para>
+/// </remarks>
+[RequiresDynamicCode("This API is not compatible with NativeAOT. For dynamic mapping via Dictionary<string, object?>, use GetCollectionDynamic() instead.")]
+[RequiresUnreferencedCode("This API is not compatible with trimming. For dynamic mapping via Dictionary<string, object?>, use GetCollectionDynamic() instead.")]
```
stephentoub: What is a consumer of ChatHistoryMemoryProvider supposed to do with the recommendation "use GetCollectionDynamic() instead"?

stephentoub: Also, what aspect is preventing this from being NativeAOT friendly?
cc: @roji

roji: This uses MEVD's static mapping mode, i.e. the .NET type ChatHistoryItem is being mapped to the database; that's the mode that isn't currently compatible with trimming/NativeAOT (a source generator is needed - this is tracked by https://github.com/microsoft/semantic-kernel/issues/10256).

However... I'm seeing that ChatHistoryItem is purely internal, so this memory provider could simply switch to dynamic mapping instead, replacing ChatHistoryItem with `Dictionary<string, object?>`; it's less nice for the internal implementation, but it's fully NativeAOT compatible. That would be my recommendation until https://github.com/microsoft/semantic-kernel/issues/10256 is done, at which point we can switch back to full static typing.

Just noting: having MEVD itself be NativeAOT-compatible isn't enough, the lower-level SDKs also need to be. Last time I checked, of the 12 MEVD providers we have, only 2 were fully NativeAOT-compatible, warning-free.

westey-m: > What is a consumer of ChatHistoryMemoryProvider supposed to do with the recommendation "use GetCollectionDynamic() instead"?

Oops, missed that, thanks for the callout.

> of the 12 MEVD providers we have, only 2 were fully NativeAOT-compatible, warning-free

This is why I haven't bothered with trying to make this NativeAOT-compatible yet, but it's probably worth converting to dynamic anyway, so we don't have to revisit.  I'll look into it today.

dotnet/src/Microsoft.Agents.AI.VectorDataMemory/ChatHistoryMemoryProviderScope.cs
```diff
@@ -0,0 +1,53 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Agents.AI.VectorDataMemory;
+
+/// <summary>
+/// Allows scoping of chat history for the <see cref="ChatHistoryMemoryProvider"/>.
+/// </summary>
+public class ChatHistoryMemoryProviderScope
```
SergeyMenshykh: can we seal this class and ChatHistoryMemoryProviderOptions one as well?

dotnet/src/Microsoft.Agents.AI.VectorDataMemory/ChatHistoryMemoryProvider.cs
stephentoub: > I'll look into it today.

Thanks, appreciate it. In general I'd really like for us to avoid accruing more debt here. For the providers that are wrappers around another library that itself is not NAOT-friendly, well, there's not much we can do and little benefit in investing in those until the underlying library is addressed. But for the MEVD infrastructure, providers that can be made safe, and especially for libraries that are just consumers of arbitrary MEVD implementations, it'd be great if we could start clean / stay clean.

dotnet/src/Microsoft.Agents.AI.VectorDataMemory/ChatHistoryMemoryProviderScope.cs
westey-m: Yes, makes sense.  Done.

dotnet/src/Microsoft.Agents.AI.VectorDataMemory/ChatHistoryMemoryProvider.cs
westey-m: Makes sense.  I updated this one and also a text store sample that we want to graduate to a feature to also use dynamic.

dotnet/src/Microsoft.Agents.AI/Memory/ChatHistoryMemoryProvider.cs
```diff
@@ -0,0 +1,453 @@
...
+/// <see cref="InvokingAsync"/> method using semantic similarity search.
+/// </para>
+/// <para>
+/// Behavior is configurable through <see cref="ChatHistoryMemoryProviderOptions"/>. When
+/// <see cref="ChatHistoryMemoryProviderOptions.SearchBehavior.OnDemandFunctionCalling"/> is selected the provider
+/// exposes a function tool that the model can invoke to retrieve relevant memories on demand instead of
+/// injecting them automatically on each invocation.
+/// </para>
+/// </remarks>
+public sealed class ChatHistoryMemoryProvider : AIContextProvider, IDisposable
```
rogerbarreto, line 38: Can we also have samples added showing how this new components would be used?

This helps understanding in the bigger picture.

dotnet/src/Microsoft.Agents.AI/Memory/ChatHistoryMemoryProvider.cs
```diff
@@ -0,0 +1,453 @@
...
+    /// <param name="options">Optional configuration options.</param>
+    /// <param name="loggerFactory">Optional logger factory.</param>
+    /// <exception cref="ArgumentNullException">Thrown when <paramref name="vectorStore"/> is <see langword="null"/>.</exception>
+    public ChatHistoryMemoryProvider(
+        VectorStore vectorStore,
+        string collectionName,
+        int vectorDimensions,
+        ChatHistoryMemoryProviderScope storageScope,
+        ChatHistoryMemoryProviderScope? searchScope = null,
+        ChatHistoryMemoryProviderOptions? options = null,
```
rogerbarreto, line 77: Given the number of parameters, maybe we can have just `Options` and all others be part of it?

```csharp
options.VectorStore
options.CollectionName
options.Dimensions
options.ProviderState = new ChatHistoryMemoryProviderState(serializedState, jsonSerializerOptions)
// OR
options.ProviderState = new ChatHistoryMemoryProviderState(storageScope, searchScope)

```

westey-m: I'm not sure this would necessarily be much better.  VectorStore, CollectionName, Dimensions and StoreScope are all required on this constructor overload, so they would need to be required on options too.
We'd just end up swapping this:
```csharp
        AIContextProviderFactory = (ctx) => new ChatHistoryMemoryProvider(
            vectorStore,
            "chathistory",
            3072,
            storageScope: new() { ThreadId = "MyThread" },
            options: new() { MaxResults = 5 })
```
for this:
```csharp
        AIContextProviderFactory = (ctx) => new ChatHistoryMemoryProvider(new Options(
            vectorStore,
            "chathistory",
            3072,
            storageScope: new() { ThreadId = "MyThread" })
        {
            MaxResults = 5
        })
```

And for cases where the optional options are not required, I think what we have is simpler.

dotnet/src/Microsoft.Agents.AI/Memory/ChatHistoryMemoryProvider.cs
westey-m: Sure, I've added a sample, check out it.

dotnet/src/Microsoft.Agents.AI.VectorDataMemory/Microsoft.Agents.AI.VectorDataMemory.csproj
westey-m: Moved to MAAI as discussed separately.

PR #1868, opened 2025-11-03
Python: Basic I/O for Encoded Image

rads-1996:
Image Spec POC


rads-1996:
Can refer to trace id - `trace_id: a49ce82b63e0653ea72ae5a397f2f051`

Also, the path for the image might vary

rads-1996:
> We appreciate the contribution. What exact scenario are we trying to solve with these samples? Do we have a gap?

This is a POC for an image specification for semantic conventions which will ultimately be raised in upstream opentelemetry. This is not a gap but something we are working on to present in upstream OTEL. Sorry for the confusion.

Issue #1815, opened 2025-10-30
.NET: Consider adding AdditionalProperties to AgentRunOptions


kzu:
When building pipelines of behavior, I found the ChatOptions.AdditionalProperties very useful to place extended info a given cross-cutting concern provides. 

A similar pattern in AgentRunOptions would be very helpful. See: https://github.com/dotnet/extensions/blob/09472076b1c0d53822d9ce085d9df408286bb867/src/Libraries/Microsoft.Extensions.AI.Abstractions/ChatCompletion/ChatOptions.cs#L222-L223

kzu:
I've sent a PR, hopefully this will make it sooner than later?

PR #1872, opened 2025-11-03
.NET: Allow extending agent run options via additional properties

kzu:
This mirrors the M.E.AI model in ChatOptions.AdditionalProperties which is very useful when building functionality pipelines.

Fixes https://github.com/microsoft/agent-framework/issues/1815

### Motivation and Context

`ChatOptions.AdditionalProperties` is very useful to pass cross-cutting data around in a funcionality pipeline. The same would be very useful in pipelines built around agents. 

### Description

Simply clones the approach in M.E.AI.

### DIFF
--- a/dotnet/tests/Microsoft.Agents.AI.Abstractions.UnitTests/AgentRunOptionsTests.cs
+++ b/dotnet/tests/Microsoft.Agents.AI.Abstractions.UnitTests/AgentRunOptionsTests.cs
@@ -18,7 +18,12 @@ public class AgentRunOptionsTests
         var options = new AgentRunOptions
         {
             ContinuationToken = new object(),
-            AllowBackgroundResponses = true
+            AllowBackgroundResponses = true,
+            AdditionalProperties = new AdditionalPropertiesDictionary
+            {
+                ["key1"] = "value1",
+                ["key2"] = 42
+            }
         };
 
         // Act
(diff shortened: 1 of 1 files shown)


dotnet/src/Microsoft.Agents.AI.Abstractions/AgentRunOptions.cs
```diff
@@ -74,4 +76,18 @@ public AgentRunOptions(AgentRunOptions options)
...
+    /// <value>
+    /// An <see cref="AdditionalPropertiesDictionary"/> containing custom properties,
+    /// or <see langword="null"/> if no additional properties are present.
+    /// </value>
+    /// <remarks>
+    /// Additional properties provide a way to include custom metadata or provider-specific
+    /// information that doesn't fit into the standard options schema. This is useful for
+    /// preserving implementation-specific details or extending the options with custom data.
+    /// </remarks>
+    public AdditionalPropertiesDictionary? AdditionalProperties { get; set; }
```
westey-m, line 92: Thanks for the contribution @kzu. Do you mind also updating the unit tests for this file.  Shouldn't be much to do there, but these two would be good to update: `CloningConstructorCopiesProperties` & `JsonSerializationRoundtrips`

kzu: Done!

westey-m (approved):

rogerbarreto (approved):

PR #1879, opened 2025-11-03
Python: fix: Map max_tokens to max_completion_tokens in OpenAI chat client

### LINKED ISSUE #1866: .NET: Python: OpenAI Chat Client consider mapping `max_tokens` to `max_completion_tokens` instead of `max_tokens`
Currently we map `max_tokens` from ChatOptions to `max_tokens` for the Chat Completion API, newer models, including reasoning models, use `max_completion_tokens` instead of `max_tokens`, we could map to the new field instead, rather then having people do that through additional_properties.

In dotnet, the field in the sdk is already deprecated (called `_deprecatedMaxTokens`), this is not the case in python, where the field is still shown, it is unclear from docs whether older api versions would break with this change.

Jeyaramjj:
### Motivation and Context

Fixes #1866

This change is required to ensure compatibility with newer OpenAI models, particularly reasoning models (`o1-preview`, `o1-mini`, etc.) that expect `max_completion_tokens` instead of the legacy `max_tokens` parameter.

**Problem it solves:**
- Users currently have to manually use `additional_properties` to set `max_completion_tokens` when working with newer OpenAI models  
- The framework was sending `max_tokens`, which can cause API errors or unexpected behavior with reasoning models  
- Inconsistency with the .NET implementation where `max_tokens` is already deprecated  

**Scenario it contributes to:**
- Seamless usage of reasoning models without requiring users to understand OpenAI API parameter differences  
- Better developer experience when transitioning between different OpenAI model generations  
- Maintains backward compatibility while supporting newer API requirements  

---

### Description

This PR implements automatic mapping from `max_tokens` to `max_completion_tokens` in the OpenAI chat client's `_prepare_options()` method.

**Implementation details:**
- Added mapping logic in `_prepare_options()` method to convert `max_tokens` to `max_completion_tokens`  
- Only maps when `max_tokens` is present and non-null to maintain safety  
- Uses `pop()` to cleanly remove the old parameter and add the new one  
- Follows the same pattern as the assistants client (as mentioned in code comments)  
- Added comprehensive test coverage for all scenarios (present, absent, null values)  

**Key changes:**
```python
# Map max_tokens to max_completion_tokens for newer models (reasoning models, etc.)
# This follows the same pattern as the assistants client
if "max_tokens" in options_dict and options_dict["max_tokens"] is not None:
    options_dict["max_completion_tokens"] = options_dict.pop("max_tokens")

### DIFF
--- a/python/packages/core/agent_framework/openai/_chat_client.py
+++ b/python/packages/core/agent_framework/openai/_chat_client.py
@@ -187,6 +187,12 @@ class OpenAIBaseChatClient(OpenAIBase, BaseChatClient):
             and issubclass(chat_options.response_format, BaseModel)
         ):
             options_dict["response_format"] = type_to_response_format_param(chat_options.response_format)
+        
+        # Map max_tokens to max_completion_tokens for newer models (reasoning models, etc.)
+        # This follows the same pattern as the assistants client
+        if "max_tokens" in options_dict and options_dict["max_tokens"] is not None:
+            options_dict["max_completion_tokens"] = options_dict.pop("max_tokens")
+        
         if additional_properties := options_dict.pop("additional_properties", None):
             for key, value in additional_properties.items():
                 if value is not None:
--- a/python/packages/core/tests/openai/test_openai_chat_client.py
+++ b/python/packages/core/tests/openai/test_openai_chat_client.py
@@ -99,6 +99,39 @@ def test_init_base_url_from_settings_env() -> None:
         assert str(client.client.base_url) == "https://custom-openai-endpoint.com/v1/"
 
 
+def test_openai_chat_client_max_tokens_mapping(openai_unit_test_env: dict[str, str]) -> None:
+    """Test that max_tokens from ChatOptions is properly mapped to max_completion_tokens for OpenAI API."""
+    client = OpenAIChatClient()
+    
+    # Test with max_tokens set in ChatOptions
+    chat_options = ChatOptions(max_tokens=100)
+    prepared_messages = client.prepare_messages([ChatMessage(role="user", text="Hello")], chat_options)
+    request_options = client._prepare_options(prepared_messages, chat_options)  # type: ignore[reportPrivateUsage]
+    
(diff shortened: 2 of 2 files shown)


Jeyaramjj:
@microsoft-github-policy-service agree company="Microsoft"

eavanvalkenburg:
@Jeyaramjj  I appreciate the enthousiasm, but I created the issue because I first need to understand the implications of this change, especially wrt azure openai api versions, so the implementation looks good, but I would want to hold off on this for the moment.

Jeyaramjj:
@eavanvalkenburg Thank you for the feedback and I completely understand your concern about the implications, especially regarding Azure OpenAI API versions. You're absolutely right to want to thoroughly understand the impact before implementing this change.

I appreciate you taking the time to review the implementation. Since this affects API compatibility across different OpenAI service versions, it makes perfect sense to pause and investigate further.

Jeyaramjj:
I did some early research and found the following:

**Key Findings:**
1. **Framework already uses this pattern**: The assistants client already maps `max_tokens` → `max_completion_tokens`
2. **Current default API version (`2024-10-21`) supports it**: Latest Azure OpenAI API versions support the parameter
3. **Risk exists for older API versions**: Users on `2023-*` versions could experience issues

**Proposed Solution:**
Implement conditional mapping based on API version detection, with fallback handling and comprehensive testing across API versions.

**Testing Plan:**
I can help implement a test matrix covering different Azure API versions, model types, and edge cases.

Would you like me to create a safer implementation with API version detection, or should we first run compatibility tests across different Azure API versions?

eavanvalkenburg:
@Jeyaramjj thanks for the research, I'm not a fan of having a bunch of references to API versions in our codebase because that is bound to break or miss something. I think given that it's the default as of 2024 versions, we can go ahead with this change, but let's add a sample demonstrating how to use the old param (through `additional_options`/`additional_chat_options`) for older API versions. What do you think?

eavanvalkenburg:
I included this fix in a broader PR, so I'm closing this, thanks for showing me how and what the impact was.

PR #1892, opened 2025-11-04
.NET: Update Extensions for Strict Agent Definitions + Improvements

rogerbarreto:
This pull request updates the Azure AI Agents integration to align with recent SDK changes and improve the handling of agent-specific chat operations. The main changes include updating the Azure.AI.Agents package version, refactoring the `AzureAIAgentChatClient` to remove the explicit model parameter (since the model is now defined at agent creation), and improving how chat options and agent references are managed internally. Additionally, an obsolete JSON context file was removed, and asynchronous API usage was updated in the sample code.

**SDK and Dependency Updates:**

* Updated the `Azure.AI.Agents` NuGet package version in `Directory.Packages.props` to `2.0.0-alpha.20251031.4` for compatibility with the latest SDK features.
* Added a dependency on the `OpenAI` package in `Microsoft.Agents.AI.AzureAI.csproj`.

**Refactoring for Agent Model Handling:**

* Refactored the `AzureAIAgentChatClient` class to remove the explicit `model` parameter from constructors, using a "no-op" model string when necessary and retrieving the model from the agent definition at runtime.
* Improved internal chat options management to ignore per-request options (since agent options can't be set per-request) and to ensure correct agent and conversation references are set in `ResponseCreationOptions`. [[1]](diffhunk://#diff-fee97fe60e41f5c94f99328cd091c9b4610f3be28a233c261c13017f2b9e94d0L87-R138) [[2]](diffhunk://#diff-fee97fe60e41f5c94f99328cd091c9b4610f3be28a233c261c13017f2b9e94d0L63-R75)

**Sample Code Updates:**

* Updated sample code in `Agent_With_AzureAIAgent/Program.cs` to use asynchronous APIs and correct usage of returned agent version objects.

**Code Cleanup:**

* Removed the obsolete `AgentsClientJsonContext.cs` file, which is no longer needed due to SDK updates.

**Internal Improvements:**

* Added helper methods for setting agent and conversation references in chat response creation, and included comments and pragma directives for future SDK compatibility.

### DIFF
--- a/dotnet/tests/Microsoft.Agents.AI.AzureAI.UnitTests/AgentsClientExtensionsTests.cs
+++ b/dotnet/tests/Microsoft.Agents.AI.AzureAI.UnitTests/AgentsClientExtensionsTests.cs
@@ -275,7 +275,7 @@ public sealed class AgentsClientExtensionsTests
 
         // Act & Assert
         var exception = await Assert.ThrowsAsync<ArgumentNullException>(() =>
-            mockClient.Object.GetAIAgentAsync(null!));
+            mockClient.Object.GetAIAgentAsync(name: null!));
 
         Assert.Equal("name", exception.ParamName);
     }
@@ -564,7 +564,7 @@ public sealed class AgentsClientExtensionsTests
 
         // Act & Assert
         var exception = await Assert.ThrowsAsync<ArgumentNullException>(() =>
-            mockClient.Object.CreateAIAgentAsync("agent-name", null!));
+            mockClient.Object.CreateAIAgentAsync(name: "agent-name", null!));
 
         Assert.Equal("agentDefinition", exception.ParamName);
     }
@@ -784,7 +784,7 @@ public sealed class AgentsClientExtensionsTests
 
             // Workaround the bug with the AgentTool.CreateAzureAISearchTool() extension
             // Using the extension method AgentTool.CreateAzureAISearchTool() fails serialization, 
-            // TODO: Revert back once Bugfix is applied: https://github.com/Azure/azure-sdk-for-net/pull/53656
+            // TODO: Revert back once bug fix is applied: https://github.com/Azure/azure-sdk-for-net/pull/53656
             ((ResponseTool)new AzureAISearchAgentTool(new())).AsAITool()
         };
 


dotnet/src/Microsoft.Agents.AI.AzureAI/AzureAIAgentChatClient.cs
```diff
@@ -21,23 +23,31 @@ internal sealed class AzureAIAgentChatClient : DelegatingChatClient
     private readonly AgentsClient _agentsClient;
     private readonly AgentVersion _agentVersion;
 
+    /// <summary>
+    /// The usage of a no-op model is a necessary change to avoid OpenAIClients to throw exceptions when
+    /// used with Azure AI Agents as the model used is now defined at the agent creation time.
+    /// </summary>
+    private const string NoOpModel = "no-op";
```
markwallace-microsoft, line 30: Can we use an empty string here?

dotnet/src/Microsoft.Agents.AI.AzureAI/AzureAIAgentChatClient.cs
```diff
@@ -84,27 +94,46 @@ private async Task<AgentConversation> GetOrCreateConversationAsync(IEnumerable<C
...
 
-        // Clear out the conversation ID to prevent the inner client from attempting to use it as a PreviousResponseId
-        conversationChatOptions.ConversationId = null;
-        // Clear out any instructions to avoid conflicts with the agent's instructions
-        conversationChatOptions.Instructions = null;
-
         return conversationChatOptions;
     }
+
+    // While Azure.AI.Agents is not yet supporting OpenAI 2.6.0 we need to use those APIs directly.
```
markwallace-microsoft: Can you provide a more detailed comment for these changes as they are non obvious

dotnet/src/Microsoft.Agents.AI.AzureAI/AzureAIAgentChatClient.cs
rogerbarreto: I'm using a specific value because, empty will trigger error in upstream validation OpenAI SDK

[OpenAI SDK - OpenAIResponseClient](https://github.com/openai/openai-dotnet/blob/23f814bd725275bf09c21b27e2ac43791d500cfe/src/Custom/Responses/OpenAIResponseClient.cs#L86)

dotnet/samples/GettingStarted/AgentProviders/Agent_With_AzureAIAgent/Program.cs
```diff
@@ -19,17 +19,17 @@
 var agentDefinition = new PromptAgentDefinition(model: deploymentName) { Instructions = JokerInstructions };
 
 // You can create a server side agent with the Azure.AI.Agents SDK.
-var agentVersion = agentsClient.CreateAgentVersion(agentName: JokerName, definition: agentDefinition).Value;
+var agentVersion = await agentsClient.CreateAgentVersionAsync(agentName: JokerName, definition: agentDefinition);
```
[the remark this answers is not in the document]
rogerbarreto: Yeah, you are right, but `ClientResult` has a implicit operator to resolve as T´s.

dotnet/samples/GettingStarted/AgentProviders/Agent_With_AzureAIAgent/Program.cs
```diff
@@ -19,17 +19,17 @@
...
 
 // You can retrieve an already created server side agent as an AIAgent.
-AIAgent existingAgent = await agentsClient.GetAIAgentAsync(agentVersion.Name);
+AIAgent existingAgent = agentsClient.GetAIAgent(agentVersion);
```
westey-m: Can we use the Async version of GetAIAgent here, so we don't mix and match async and non-async in the sample?

dotnet/samples/GettingStarted/AgentProviders/Agent_With_AzureAIAgent/Program.cs
```diff
@@ -19,17 +19,17 @@
...
 
 // You can also create a server side persistent agent and return it as an AIAgent directly.
-var createdAgent = agentsClient.CreateAIAgent(name: JokerName, model: deploymentName, instructions: JokerInstructions);
+var createdAgent = await agentsClient.CreateAIAgentAsync(name: JokerName, model: deploymentName, instructions: JokerInstructions);
 
 // You can then invoke the agent like any other AIAgent.
 AgentThread thread = existingAgent.GetNewThread();
 Console.WriteLine(await existingAgent.RunAsync("Tell me a joke about a pirate.", thread));
 
 // Cleanup by agent name (removes both agent versions created by existingAgent + createdAgent).
```
westey-m: If we are creating a different version of the same agent on line 28, we should probably explain that in the comment on line 27.

dotnet/samples/GettingStarted/AgentProviders/Agent_With_AzureAIAgent/Program.cs
rogerbarreto: > Can we use the Async version of GetAIAgent 

You mean `Sync`, as of your suggestion?

dotnet/src/Microsoft.Agents.AI.AzureAI/AzureAIAgentChatClient.cs
westey-m: Why are we calling `GetConversationAsync` when we already have a conversation id?

dotnet/src/Microsoft.Agents.AI.AzureAI/AzureAIAgentChatClient.cs
```diff
@@ -84,27 +94,46 @@ private async Task<AgentConversation> GetOrCreateConversationAsync(IEnumerable<C
...
+    private static void SetAdditionalProperty<T>(ResponseCreationOptions responseCreationOptions, string key, BinaryData value)
+    {
+        responseCreationOptions.Patch.Set([.. "$."u8, .. Encoding.UTF8.GetBytes(key)], value);
+    }
+
+    private static void SetAgentReference(ResponseCreationOptions responseCreationOptions, AgentVersion agentVersion)
+    {
+        var agentReference = new AgentReference(agentVersion.Name) { Version = agentVersion.Version };
+
+        SetAdditionalProperty<BinaryData>(responseCreationOptions, "agent", ModelReaderWriter.Write(agentReference, new ModelReaderWriterOptions("W"), AzureAIAgentsContext.Default));
```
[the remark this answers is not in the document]
rogerbarreto: It's part of another Library.

dotnet/src/Microsoft.Agents.AI.AzureAI/AgentsClientExtensions.cs
```diff
@@ -411,453 +279,261 @@ public static ChatClientAgent CreateAIAgent(
...
-        };
-
-        AgentVersionCreationOptions? versionCreationOptions = null;
-        if (metadata is not null)
+        List<AITool>? agentTools = null;
+        if (agentDefinition is PromptAgentDefinition { Tools: { Count: > 0 } definitionTools })
         {
-            versionCreationOptions ??= new();
-            foreach (var kvp in metadata)
+            // The no tools were provided while the agent definition requires in-proc tools.
```
SergeyMenshykh: The article "The" is not needed here.

dotnet/src/Microsoft.Agents.AI.AzureAI/AzureAIAgentChatClient.cs
rogerbarreto: That's a good point, there's no public API to create a `AgentConversation`, and I was following the normal expected path, which I also agree may create an extra hop into the backend to retrieve something that might be already there. 

I will attempt to deserialize a `AgentConversation` when we have the ConversationId to avoid this potential unnecessary call.

rogerbarreto: Actually I can just use the `conversationId` straight away, don't need to use `agentConversation` type anywhere. Will refactor this.

dotnet/src/Microsoft.Agents.AI.AzureAI/AgentsClientExtensions.cs
```diff
@@ -411,453 +274,261 @@ public static ChatClientAgent CreateAIAgent(
...
-                    sb ??= new();
-
-                    if (sb.Length > 0)
-                    {
-                        _ = sb.AppendLine();
-                    }
+                Temperature = promptAgentDefinition.Temperature,
+                TopP = promptAgentDefinition.TopP,
+                Instructions = promptAgentDefinition.Instructions,
+            };
```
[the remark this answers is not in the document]
rogerbarreto: This is something that is currently debatable, we might just remove the property from the `ChatClientAgentOptions` and keep the `ChatOptions.Instructions` as the source for instructions, this is out of scope of this PR though.

dotnet/src/Microsoft.Agents.AI.AzureAI/AzureAIAgentChatClient.cs
```diff
@@ -21,23 +23,31 @@ internal sealed class AzureAIAgentChatClient : DelegatingChatClient
...
+        : this(agentsClient, Throw.IfNull(agentRecord).Versions.Latest, openAIClientOptions)
     {
     }
 
-    internal AzureAIAgentChatClient(AgentsClient agentsClient, AgentVersion agentVersion, string model, OpenAIClientOptions? openAIClientOptions = null)
-        : base(agentsClient.GetOpenAIClient(openAIClientOptions).GetOpenAIResponseClient(model).AsIChatClient())
+    internal AzureAIAgentChatClient(AgentsClient agentsClient, AgentVersion agentVersion, OpenAIClientOptions? openAIClientOptions = null)
+        : base(agentsClient
+            .GetOpenAIClient(openAIClientOptions)
+            .GetOpenAIResponseClient((agentVersion.Definition as PromptAgentDefinition)?.Model ?? NoOpModel)
```
SergeyMenshykh, line 49: What happens when the no-op model is provided to the GetOpenAIResponseClient method? Does it throw immediately, or later when the chat client is invoked, or?

dotnet/src/Microsoft.Agents.AI.AzureAI/AgentsClientExtensions.cs
```diff
@@ -127,24 +87,30 @@ public static async Task<ChatClientAgent> GetAIAgentAsync(
...
     /// <param name="openAIClientOptions">An optional <see cref="OpenAIClientOptions"/> for configuring the underlying OpenAI client.</param>
     /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests. The default is <see cref="CancellationToken.None"/>.</param>
-    /// <returns>A <see cref="ChatClientAgent"/> instance that can be used to perform operations based on the latest version of the Azure AI Agent record.</returns>
-    /// <exception cref="ArgumentNullException">Thrown when <paramref name="agentsClient"/>, <paramref name="agentRecord"/> or <paramref name="options"/> is <see langword="null"/>.</exception>
+    /// <returns>A <see cref="ChatClientAgent"/> instance that can be used to perform operations based on the latest version of the Azure AI Agent.</returns>
+    /// <remarks>When using prompt agent definitions with tools the parameter <paramref name="tools"/> needs to be provided.</remarks>
     public static ChatClientAgent GetAIAgent(
         this AgentsClient agentsClient,
         AgentRecord agentRecord,
-        ChatClientAgentOptions? options = null,
```
westey-m, line 147: Without ChatClientAgentOptions we don't have the ability to add AIContextProviders.
Maybe they don't make sense with these Agents anyway, since ad-hoc function advertising is not supported, so some functionality won't necessarily work, but it'll be good for us to agree a clear position on this.
Custom ChatMessageStores would also not be supported, but that's fine, since these agents force you into service storage anyway.

dotnet/samples/GettingStarted/AgentProviders/Agent_With_AzureAIAgent/Program.cs
```diff
@@ -18,18 +18,35 @@
...
+
+// Once you have the AIAgent, you can invoke it like any other AIAgent.
+AgentThread thread = jokerAgentLatest.GetNewThread();
+Console.WriteLine(await jokerAgentLatest.RunAsync("Tell me a joke about a pirate.", thread));
+
+// This will use the same thread to continue the conversation.
+Console.WriteLine(await jokerAgentLatest.RunAsync("Now tell me a joke about a cat and a dog using last joke as the anchor.", thread));
+
+// Cleanup by agent name removes both agent versions created (jokerAgentV1 + jokerAgentV2).
+agentsClient.DeleteAgent(jokerAgentV1.Name);
```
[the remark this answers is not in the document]
rogerbarreto: Yeah, I'm following @westey-m point trying to use all `Sync` which also expose one aspect of Agent Framework API's that don't provide `Sync`.

Ideally we should be able to have `AIAgent.Run()` and all be sync.

dotnet/src/Microsoft.Agents.AI.AzureAI/AgentsClientExtensions.cs
```diff
@@ -411,453 +279,261 @@ public static ChatClientAgent CreateAIAgent(
...
+    /// <summary>For already created agent versions, retrieve the definition and validate the tools parameter.</summary>
+    /// <exception cref="ArgumentException"><see cref="PromptAgentDefinition.Tools"/> cannot be used. The <paramref name="tools"/> parameter should be used instead.</exception>
+    /// <remarks>
+    /// Because <see cref="PromptAgentDefinition.Tools"/> doesn't support in-proc tools (only declarative/definitions),
+    /// the <paramref name="tools"/> parameter needs to be the single source of truth for tools, and must be provided when using tools.
+    /// </remarks>
+    private static void ValidateUsingToolsParameter(AgentVersion agentVersion, IList<AITool>? tools)
     {
-        foreach (var toolName in toolNames)
+        if (agentVersion.Definition is PromptAgentDefinition { Tools.Count: > 0 } && tools is { Count: 0 })
```
westey-m: Is it supported that we pass in 1 tool when getting an agent that has 5 defined?

dotnet/samples/GettingStarted/AgentProviders/Agent_With_AzureAIAgent/Program.cs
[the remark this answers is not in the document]
stephentoub: > Ideally we should be able to have AIAgent.Run() and all be sync.

What agent implementation are you talking to that doesn't use I/O?

dotnet/src/Microsoft.Agents.AI.AzureAI/AzureAIAgentChatClient.cs
rogerbarreto: No-Op model will not be used, regardless of the value, it is stripped from the `RequestBody` in the `SetAgentReference`.

The backend assumes what is the model based on what was the `agentDefinition`, this control is not in the hands of the caller but the Agent creator.

```csharp
responseCreationOptions.Patch.Remove([.. "$."u8, .. Encoding.UTF8.GetBytes("model")]);
```

dotnet/src/Microsoft.Agents.AI.AzureAI/AgentsClientExtensions.cs
```diff
@@ -411,453 +279,261 @@ public static ChatClientAgent CreateAIAgent(
...
+            }
 
-        [JsonPropertyName("properties")]
-        public Dictionary<string, JsonElement> Properties { get; set; } = [];
+            foreach (var tool in tools)
+            {
+                // Ensure that any AIFunctions provided are In-Proc, not just the declarations.
+                if (tool is AIFunctionDeclaration and not AIFunction)
+                {
+                    throw new InvalidOperationException("When providing function avoid converting FunctionTools to AITools, use AIFunctionFactory instead.");
```
westey-m: Is this the only way in which FunctionDeclerations can be created?  Should we tell the user what is actually wrong too, so that if they got a FunctionDeclerations in a different way than what we are describing, they know how to fix the issue?

dotnet/src/Microsoft.Agents.AI.AzureAI/AgentsClientExtensions.cs
```diff
@@ -411,453 +279,261 @@ public static ChatClientAgent CreateAIAgent(
...
-                        }
-
-                        break;
-                }
-
-                return responsesMcpTool;
-
-            default:
-                throw new NotSupportedException($"Tool of type '{tool.GetType().FullName}' is not supported.");
+            throw new ArgumentException("When creating agents with prompt agent definitions use the dedicated tools parameter to provide the necessary tools instead.", nameof(tools));
```
westey-m:
```suggestion
            throw new ArgumentException("When creating agents with prompt agent definitions use the dedicated tools parameter to provide the necessary tools instead of PromptAgentDefinition.Tools.", nameof(tools));
```

dotnet/src/Microsoft.Agents.AI.AzureAI/AgentsClientExtensions.cs
rogerbarreto: I'm not being ultra strict to how to validate the tools in this method as I validate more deeply in the `CreateChatClientAgentOptions` method.

dotnet/samples/GettingStarted/AgentProviders/Agent_With_AzureAIAgent/Program.cs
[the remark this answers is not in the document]
westey-m: My point was actually about being consistent, but I would rather do all Async rather than all Sync.  I think Async is more common and all the methods we are calling here are doing I/O.

dotnet/samples/GettingStarted/AgentProviders/Agent_With_AzureAIAgent/Program.cs
```diff
@@ -18,18 +18,35 @@
...
+//      agentVersion.Name = <agentName>
 
-// You can retrieve an already created server side agent as an AIAgent.
-AIAgent existingAgent = await agentsClient.GetAIAgentAsync(agentVersion.Name);
+// You can retrieve an AIAgent for a already created server side agent version.
+AIAgent jokerAgentV1 = agentsClient.GetAIAgent(agentVersion);
 
-// You can also create a server side persistent agent and return it as an AIAgent directly.
-var createdAgent = agentsClient.CreateAIAgent(name: JokerName, model: deploymentName, instructions: JokerInstructions);
+// You can also create another AIAgent version (V2) of by providing the same name with a different definition.
```
westey-m:
```suggestion
// You can also create a second AIAgent version of the existing Agent by providing the same name with a different definition.
```

dotnet/src/Microsoft.Agents.AI.AzureAI/AgentsClientExtensions.cs
rogerbarreto: The method that takes the `AgentVersion` is called from and already existing agent.

As of now, the only in-proc tool we have is the `AIFunction`, 

So, any other **NOT in-proc** tools that sent in the `Get` will be just ignored in favor of the server side tools loaded from the definition.

> [!NOTE]
> I can revisit this `tools` parameter to be a `IList<AIFunction> functions` for the `Get`s but this can be limiting for in-proc custom tools (I can't think of now, but we may have).

dotnet/src/Microsoft.Agents.AI.AzureAI/AgentsClientExtensions.cs
rogerbarreto: Clarified the question offline, I agree, the message was improved.

dotnet/src/Microsoft.Agents.AI.AzureAI/AgentsClientExtensions.cs
rogerbarreto: Yeah good point I added the dedicated extensions for this type, there was some permutations missing from both Get and Create.

dotnet/samples/GettingStarted/AgentProviders/Agent_With_AzureAIAgent/Program.cs
westey-m: No, I meant can we use Async everywhere.  Maybe you read the change in the code from async to sync as my suggestion?  I was actually just commenting on the change, and I think async everywhere is better.  It's the more common approach since all these methods are doing I/O.

dotnet/src/Microsoft.Agents.AI.AzureAI/AzureAIAgentChatClient.cs
westey-m: Yes, that's what I was thinking as well 😄

dotnet/src/Microsoft.Agents.AI.AzureAI/AgentsClientExtensions.cs
westey-m: Let's talk more about this in the standup.  Since we can only partially support these features with V2 agents, I'd like to get everyone's thoughts on whether it makes sense to support them at all.

dotnet/samples/GettingStarted/AgentProviders/Agent_With_AzureAIAgent/Program.cs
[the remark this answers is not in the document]
rogerbarreto: I'm not sure how the I/O doing or not should dictate the availability of Sync methods.

I understand that Azure do both offerings for the API's even including `CancellationToken` for Sync, just gathering your thoughts on this subject, and correct me if I'm wrong here we would rather not have `Sync` versions given them do I/O...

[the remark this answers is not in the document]
stephentoub: That is a choice Azure SDK has made.

My preference is to only have the async methods.

PR #1896, opened 2025-11-04
.NET: AG-UI support for .NET: Support for tool calling

javiercn:
This PR adds comprehensive support for tool calling in the AG-UI implementation for .NET, including client-side and server-side function execution, conversation management, and proper serialization across boundaries.

### **1. AGUIAgent → AGUIChatClient: Architectural Shift**

**Previous Architecture:**
- The original design used `AGUIAgent` as a wrapper around an `IChatClient`
- This prevented using FunctionInvokingChatClient and required creating custom threads and so on.
  - IChatClient now handles the conversation on a given turn(s) and ChatClientAgent handles the run, memory, etc.

**New Architecture:**
- **`AGUIChatClient`** is now an `IChatClient` implementation that communicates with AG-UI compliant servers
- It properly extends `DelegatingChatClient` and wraps a `FunctionInvokingChatClient`
- The internal architecture now has clear layers:
  1. `AGUIChatClient` (outer) - manages conversation ID and ServerFunctionCallContent wrapping/unwrapping
  2. `FunctionInvokingChatClient` (middle) - handles client-side tool execution
  3. `AGUIChatClientHandler` (inner) - handles HTTP communication with AG-UI server

**Why this matters:**
- **Composability**: Leverages the existing `FunctionInvokingChatClient` for client-side tool execution
- **Separation of concerns**: HTTP communication, tool execution, and conversation management are cleanly separated
- **Standards compliance**: Implements `IChatClient` interface consistently with the rest of the ecosystem
---

### **2. ConversationId Handling in AG-UI**

AG-UI has a fundamental constraint: it **requires the full message history on every turn** and uses `ThreadId` for conversation tracking. However, the Microsoft.Extensions.AI `IChatClient` interface uses `ConversationId` to indicate stateful conversation management where the client maintains history.

**The Challenge:**
- When `ChatOptions.ConversationId` is set, `FunctionInvokingChatClient` assumes the underlying client maintains history and only sends new messages
- AG-UI always needs the full history, even across multiple turns
- We need to preserve the conversation identity for the caller while preventing `FunctionInvokingChatClient` from treating the conversation as stateful

**The Solution:**
The PR implements a sophisticated "conversation ID juggling" pattern:

1. **On Request (Client → Server):**
   ```csharp
   // If user provides a conversation ID, extract it and clear it from options
   if (options?.ConversationId != null)
   {
       conversationId = options.ConversationId;
       innerOptions = options.Clone();
       innerOptions.AdditionalProperties["agui_thread_id"] = conversationId;
       innerOptions.ConversationId = null;  // Hide from FunctionInvokingChatClient
   }
   ```

2. **During Tool Execution:**
   - For **client tool calls**: Store the thread ID in `FunctionCallContent.AdditionalProperties["agui_thread_id"]`
   - When the tool result comes back with the history, extract the thread ID from the penultimate message
   - This ensures continuity across the function invocation round-trip

3. **On Response (Server → Client):**
   ```csharp
   // Extract thread ID from first update if not already set
   if (firstUpdate != null && conversationId == null)
   {
       conversationId = firstUpdate.AdditionalProperties["agui_thread_id"];
   }
   
   // Restore conversation ID for all updates before yielding to caller
   finalUpdate.ConversationId = conversationId;
   ```

**Why this complex approach?**
- AG-UI protocol requirement: Must send full history every time
- Microsoft.Extensions.AI design: `ConversationId` implies incremental history
- `FunctionInvokingChatClient` behavior: Only sends new messages when `ConversationId` is set
- **Solution**: Hide `ConversationId` from the middleware layer, use `AdditionalProperties` as temporary storage, restore it for the caller

This ensures:
- ✅ Callers see consistent `ConversationId` across all updates
- ✅ Full message history is always sent to AG-UI server
- ✅ Multi-turn conversations work correctly
- ✅ Tool calls preserve conversation context

---

### **3. ServerFunctionCallContent: Hiding Server-Executed Tools**

**The Problem:**
When a server executes a tool, it returns a `FunctionCallContent` in the response stream. The wrapping `FunctionInvokingChatClient` sees this and tries to execute it again on the client side, causing:
- Duplicate execution attempts
- Errors when the client doesn't have the server's tool registered
- Confusion about where tools should execute

**The Solution - ServerFunctionCallContent:**
```csharp
private class ServerFunctionCallContent(FunctionCallContent functionCall) : AIContent
{
    public FunctionCallContent FunctionCallContent { get; } = functionCall;
}
```

**How it works:**

1. **Detection** (in `AGUIChatClientHandler`):
   ```csharp
   if (update.Contents[0] is FunctionCallContent fcc)
   {
       if (clientToolSet.Contains(fcc.Name))
       {
           // Client tool - let FunctionInvokingChatClient handle it
       }
       else
       {
           // Server tool - wrap it to hide from FunctionInvokingChatClient
           update.Contents[0] = new ServerFunctionCallContent(fcc);
       }
   }
   ```

2. **Unwrapping** (in `AGUIChatClient`):
   ```csharp
   for (var i = 0; i < update.Contents.Count; i++)
   {
       if (content is ServerFunctionCallContent serverFunctionCallContent)
       {
           update.Contents[i] = serverFunctionCallContent.FunctionCallContent;
       }
   }
   ```

**Why a wrapper class?**
- `FunctionInvokingChatClient` only recognizes `FunctionCallContent` as something it should execute
- By wrapping server tool calls in `ServerFunctionCallContent`, they become "invisible" to the middleware
- The outer `AGUIChatClient` then unwraps them before yielding to the caller
- **Result**: Callers see normal `FunctionCallContent`, but middleware doesn't try to execute server tools

**Benefits:**
- ✅ Server tools execute only on the server
- ✅ Client tools execute only on the client
- ✅ Mixed server/client tool scenarios work correctly
- ✅ No duplicate execution attempts
- ✅ Clean separation of execution boundaries

---

### **Message Type Hierarchy**
- Introduced proper message type hierarchy: `AGUIMessage` (base) with concrete types:
  - `AGUIUserMessage`
  - `AGUIAssistantMessage`
  - `AGUISystemMessage`
  - `AGUIDeveloperMessage`
  - `AGUIToolMessage`
- Replaced generic `AGUIMessage` with role strings to strongly-typed message classes
- Improved type safety and serialization handling

### **Tool Call Events**
- Added comprehensive tool call events:
  - `ToolCallStartEvent`
  - `ToolCallArgsEvent`
  - `ToolCallEndEvent`
  - `ToolCallResultEvent`
- Converted between `ChatResponseUpdate` (with `FunctionCallContent`) and AG-UI events
- Proper streaming of tool call arguments

### **JSON Serialization**
- Added `JsonSerializerOptions` parameter to `AGUIChatClient` constructor
- Allows custom JSON contexts for client-specific types
- Server and client can use different serialization contexts
- Merged AGUI context with custom contexts automatically
- Proper handling of cross-boundary serialization (server types → `JsonElement` on client)

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.AGUI/Microsoft.Agents.AI.AGUI.csproj
+++ b/dotnet/src/Microsoft.Agents.AI.AGUI/Microsoft.Agents.AI.AGUI.csproj
@@ -9,11 +9,6 @@
   <Import Project="$(RepoRoot)/dotnet/nuget/nuget-package.props" />
 
   <PropertyGroup>
-    <!-- Disable packing until we are ready to release this as a nuget -->
-    <IsPackable>false</IsPackable>
-  </PropertyGroup>
-
-  <PropertyGroup>
     <InjectSharedThrow>true</InjectSharedThrow>
   </PropertyGroup>
 
--- a/dotnet/src/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore.csproj
+++ b/dotnet/src/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore.csproj
@@ -13,11 +13,6 @@
   <Import Project="$(RepoRoot)/dotnet/nuget/nuget-package.props" />
 
   <PropertyGroup>
-    <!-- Disable packing until we are ready to release this as a nuget -->
-    <IsPackable>false</IsPackable>
-  </PropertyGroup>
-
-  <PropertyGroup>
     <!-- NuGet Package Settings -->
     <Title>Microsoft Agent Framework Hosting AG-UI ASP.NET Core</Title>
     <Description>Provides Microsoft Agent Framework support for hosting AG-UI agents in an ASP.NET Core context.</Description>


dotnet/samples/AGUIClientServer/AGUIClient/Program.cs
```diff
@@ -51,11 +51,21 @@ private static async Task HandleCommandsAsync(CancellationToken cancellationToke
             Timeout = TimeSpan.FromSeconds(60)
         };
 
-        AGUIAgent agent = new(
-            id: "agui-client",
+        var changeBackground = AIFunctionFactory.Create(
+            () => Console.WriteLine("Changing color to blue"),
```
stephentoub: Nit: you could actually change it :)
```suggestion
            () =>
            {
                Console.ForegroundColor = ConsoleColor.DarkBlue;
                Console.WriteLine("Changed color to blue");
            },
```

dotnet/samples/AGUIClientServer/AGUIServer/Program.cs
```diff
@@ -12,14 +13,27 @@
...
+app.MapAGUIAgent("/", (messages, tools) =>
 {
     return new AzureOpenAIClient(
             new Uri(endpoint),
             new DefaultAzureCredential())
         .GetChatClient(deploymentName)
-        .CreateAIAgent(name: "AGUIAssistant");
+        .CreateAIAgent(
+            name: "AGUIAssistant",
+            tools: [.. tools, .. new AITool[]{
```
stephentoub: Why do you need the `.. new AITool[] { ... }` part? Can't it just be `[.. tools, AIFunctionFactory.Create(...)]`?

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
```diff
@@ -0,0 +1,266 @@
...
+    public async Task<ChatResponse> GetResponseAsync(IEnumerable<ChatMessage> messages, ChatOptions? options = null, CancellationToken cancellationToken = default)
+    {
+        var updates = new List<ChatResponseUpdate>();
+        await foreach (var update in this.GetStreamingResponseAsync(messages, options, cancellationToken).ConfigureAwait(false))
+        {
+            updates.Add(update);
+        }
+
+        return updates.ToChatResponse();
+    }
```
stephentoub, line 76: Seems like this could be just:
```suggestion
    {
        return this.GetStreamingResponseAsync(messages, options, cancellationToken).ToChatResponseAsync();
    }
```
(also making the method non-`async`)

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
```diff
@@ -0,0 +1,266 @@
...
+    /// <inheritdoc />
+    public async IAsyncEnumerable<ChatResponseUpdate> GetStreamingResponseAsync(
+        IEnumerable<ChatMessage> messages,
+        ChatOptions? options = null,
+        [EnumeratorCancellation] CancellationToken cancellationToken = default)
+    {
+        await foreach (var update in this._innerClient.GetStreamingResponseAsync(messages, options, cancellationToken).ConfigureAwait(false))
+        {
+            // We bypass the underlying FunctionInvokingChatClient for any server executed function.
+            // This is because it will try to invoke it and fail because it can't find the function.
```
stephentoub: I'm not sure exactly what this is referring to, but there are configuration opportunities on an FICC to change behaviors here:
- If you set TerminateOnUnknownCalls to true, the FICC will allow an unknown FCC to pass through rather than sending an FRC back to the inner client about a failure.
- If the tool is an AIFunctionDeclaration rather than the derived AIFunction, the FICC won't try to invoke it and will just let it pass through.
- You can configure the FICC's AdditionalTools property with tools that should be available to the FICC to invoke even if they weren't provided in ChatOptions.Tools.
etc.

You could also selectively decide whether or not to employ an FICC per request, either by manufacturing one per request specific to that request, or by having two wrapped clients, one that has the FICC in the pipeline and one that doesn't.

I don't know if any of those are relevant (maybe I'll figure that out from reading further), but wanted to mention it.

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
```diff
@@ -0,0 +1,266 @@
...
+
+namespace Microsoft.Agents.AI.AGUI;
+
+/// <summary>
+/// Provides an <see cref="IChatClient"/> implementation that communicates with an AG-UI compliant server.
+/// </summary>
+public sealed class AGUIChatClient : IChatClient
+{
+    private readonly IChatClient _innerClient;
+    private readonly List<ChatResponseUpdate> _executedServerFunctionStream = [];
```
stephentoub: Generally IChatClient implementations are intended to be stateless and usable concurrently. This seems to violate that, as it's storing state specific to a particular invocation? Is that intentional or mitigated in some other way?

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
```diff
@@ -0,0 +1,266 @@
...
+using Microsoft.Extensions.Logging.Abstractions;
+
+namespace Microsoft.Agents.AI.AGUI;
+
+/// <summary>
+/// Provides an <see cref="IChatClient"/> implementation that communicates with an AG-UI compliant server.
+/// </summary>
+public sealed class AGUIChatClient : IChatClient
+{
+    private readonly IChatClient _innerClient;
```
stephentoub: Typically such a client would derive from DelegatingChatClient, which provides implementations for all the methods that forward to the inner client, and you'd then just override Get{Streaming}ResponseAsync... doesn't look like you'd need to override GetService or Dispose yourself then.

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
```diff
@@ -0,0 +1,266 @@
...
+        }
+        return this._innerClient.GetService(serviceType, serviceKey);
+    }
+
+    private sealed class AGUIChatClientHandler : IChatClient
+    {
+        private readonly AGUIHttpService _httpService;
+        private readonly JsonSerializerOptions _jsonSerializerOptions;
+
+        private readonly List<ChatResponseUpdate> _serverFunctionUpdates = [];
```
stephentoub: Same statefulness question.

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
```diff
@@ -0,0 +1,266 @@
...
+        private readonly ILogger _logger;
+
+        private List<ChatMessage>? _currentMessages;
```
stephentoub: And here as well.

Is it guaranteed that one of these chat client instances is only ever used for a single request at a time?

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
```diff
@@ -0,0 +1,266 @@
...
+        }
+
+        public ChatClientMetadata Metadata { get; }
+
+        public async Task<ChatResponse> GetResponseAsync(
+            IEnumerable<ChatMessage> messages,
+            ChatOptions? options = null,
+            CancellationToken cancellationToken = default)
+        {
+            throw new NotImplementedException("Use GetStreamingResponseAsync instead.");
```
stephentoub: Is this unreachable? Maybe add an assert?

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
```diff
@@ -0,0 +1,266 @@
...
+            {
+                if (update.Contents is { Count: 1 })
+                {
+                    if (update.Contents[0] is FunctionCallContent fcc)
+                    {
+                        if (clientToolSet.Contains(fcc.Name))
+                        {
+                            // If this is a client function, FunctionInvokingChatClient will only send the result back
+                            // so we need to keep the current messages as expected by AG-UI.
+                            this._currentMessages ??= [.. messages];
```
stephentoub: This is interesting. Is there a ConversationId set? FICC will send the entire history back to the server if there's no ConversationId; it'll send only the result if there is a ConversationId, because it assumes the inner client / service already knows about all of the previous messages (as they'd be in the tracked conversation).

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
```diff
@@ -0,0 +1,266 @@
...
+                yield return update;
+            }
+
+            this._logger.LogInformation("[AGUIChatClient] Request completed - ThreadId: {ThreadId}, RunId: {RunId}", "", runId);
+        }
+
+        public TService? GetService<TService>(object? key = null) where TService : class
+        {
+            return this as TService;
+        }
```
stephentoub: Why is this needed? There's an extension method on IChatClient that provides `GetService<T>`.

dotnet/src/Microsoft.Agents.AI.AGUI/Shared/AGUIChatMessageExtensions.cs
```diff
@@ -15,26 +17,187 @@ internal static class AGUIChatMessageExtensions
...
+                {
+                    MessageId = message.Id
+                };
+            }
+            else
+            {
+                string content = message switch
+                {
+                    AGUIDeveloperMessage dev => dev.Content,
+                    AGUISystemMessage sys => sys.Content,
```
stephentoub: I don't remember if I asked this before, but I'm still curious about what distinction AG-UI makes between system messages and developer messages.

dotnet/src/Microsoft.Agents.AI.AGUI/Shared/AGUIChatMessageExtensions.cs
```diff
@@ -44,5 +207,44 @@ public static ChatRole MapChatRole(string role) =>
         string.Equals(role, AGUIRoles.User, StringComparison.OrdinalIgnoreCase) ? ChatRole.User :
         string.Equals(role, AGUIRoles.Assistant, StringComparison.OrdinalIgnoreCase) ? ChatRole.Assistant :
         string.Equals(role, AGUIRoles.Developer, StringComparison.OrdinalIgnoreCase) ? s_developerChatRole :
+        string.Equals(role, AGUIRoles.Tool, StringComparison.OrdinalIgnoreCase) ? ChatRole.Tool :
         throw new InvalidOperationException($"Unknown chat role: {role}");
+
+    private static JsonElement ConvertToJsonElement(object value, JsonSerializerOptions jsonSerializerOptions)
```
stephentoub: cc: @eiriktsarpalis... this looks like a lot of ceremony that shouldn't be necessary?

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
javiercn: We discussed this offline, but I'm going to comment here for everyone. The challenge is that we will stream FCCs and results from the server that the client uses to render custom UI bits.

We use a `FunctionInvokingChatClient` to invoke frontend tools, and when it sees the server FCC it tries to execute it. https://github.com/dotnet/extensions/pull/7011/files covers making FunctionInvokingChatClient not fail for functions that already contain results on the stream.

In the meantime, I worked around this behavior by passing a side-channel that I use to populate with any server executed function and that I check every time I yield and update and flush before doing so. 
This was simpler than the alternative of using a custom FunctionInvoker, but the logic would be the same, we would keep a Dictionary<string,Result> and in the invoker we would check for the result before dispatching to `ctx.Function.InvokeAsync`. The challenge is that I would need to "advance" the underlying stream to access the result (if it was present) so that I could return it in the invoker.

Having two separate pipelines wouldn't work I think since I would not have a way to choose which one to use until I got the response.

All in all, the cleanest solutions are:
* https://github.com/dotnet/extensions/pull/7011/files
* Have two separate AIContent ways of representing a remotely executed function:
  * RemoteFunctionCallContent
  * RemoteFunctionResultContent

That way we can be more explicit about what's critical that someone needs to look at, and what is just additional information for consumers.

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
javiercn: I added a comment https://github.com/microsoft/agent-framework/pull/1896#discussion_r2491727460 based on our discussion. I use this as a side-channel to bypass the FunctionInvokingChatClient, but the best solution I think involves one of the two options we discussed. We can clean this up once we make up our minds.

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
javiercn: Yep, I was having trouble because of the shared piece of state that I'm passing down to the inner client, but I found a way around it. Once we have https://github.com/dotnet/extensions/pull/7011/files we can get rid of this 3-layer lasagna and have `AGUIChatClientHandler` just be `AGUIChatClient` without the need for an internal `FunctionInvokingChatClient` as CreateAIAgent will wrap that up as needed.

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
javiercn: This is the other part of the side channel where we see the server functions, and we push them to yield them to the caller before we yield any client function tool call.

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
javiercn: > Is it guaranteed that one of these chat client instances is only ever used for a single request at a time?

That's the idea as a starting point. But it is fair that if these need to be stateless, is not a constraint that we want to break.

I did it this way because of issues with the thread id that ag-ui requires sending. It was causing the `FunctionInvokingChatClient` to only send new messages (I think you commented on that below), but I found an alternative way around it where I manufacture a thread ID and I take it out from any ChatResponseUpdate if I detect the conversation ID matches the thread id. That way the thread ID only exists over the wire.

The main drawback of this is that I create a new thread id every time and essentially multiple turns have different thread IDs over the wire (I don't know how much that will confuse other AG-UI server implementations). That said, if the service returns a thread ID, then we will switch to only send new messages. This is not AG-UI compliant (it expects all messages on every turn) but I think that's fine for now. Alternatively, I don't know if there is a way to make FICC send the messages all the time despite the conversation ID.

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
javiercn: Added some additional context [here](https://github.com/microsoft/agent-framework/pull/1896#discussion_r2491873810)

Basically, to be AG-UI compliant we always need to send the full message history on every turn (there is a plan to change this on AG-UI) and our two options are:
* Manufacture a thread ID and take it out before yielding any update (what I'll be doing soon) so that FICC passes down the full augmented history on every client function call.
* Keep the list of messages from the previous turn and augment the chat history when we detect we are going to yield to a client function (what I was doing).

The main challenge being that currently, if the service sends an actual conversation id, we will start to only send newer messages and that we will be generating a new threadId on every turn, which defeats the purpose.

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
javiercn: This is gone now.

dotnet/src/Microsoft.Agents.AI.AGUI/Shared/AGUIChatMessageExtensions.cs
javiercn: Yep, I replied https://github.com/microsoft/agent-framework/pull/1776#discussion_r2477987003 here. I don't think AG-UI itself does a specific distinction. I think it reflects roles that are typically defined by other service providers.

dotnet/samples/AGUIClientServer/AGUIClient/AGUIClientSerializerContext.cs
```diff
@@ -0,0 +1,14 @@
...
+
+using System.Text.Json.Serialization;
+
+namespace AGUIClient;
+
+[JsonSerializable(typeof(SensorRequest))]
+[JsonSerializable(typeof(SensorResponse))]
+internal sealed partial class AGUIClientSerializerContext : JsonSerializerContext
+{
+}
```
stephentoub: Nit: empty types like this can just be:

```suggestion
internal sealed partial class AGUIClientSerializerContext : JsonSerializerContext;
```

dotnet/samples/AGUIClientServer/AGUIServer/ServerWeatherForecastRequest.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+namespace AGUIServer;
+
+internal sealed class ServerWeatherForecastRequest
+{
+    public DateTime Date { get; set; }
+
```
stephentoub: Nit: some of your earlier types like this don't have blank lines between props, but these do

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
```diff
@@ -0,0 +1,327 @@
...
+        ChatResponseUpdate? firstUpdate = null;
+        string? conversationId = null;
+        // AG-UI requires the full message history on every turn, so we clear the conversation id here
+        // and restore it for the caller.
+        var innerOptions = options;
+        if (options?.ConversationId != null)
+        {
+            conversationId = options.ConversationId;
+
+            // Clone the options and set the conversation ID to null so the FunctionInvokingChatClient doesn't see it.
```
DeagleGross, line 98: why do we want to avoid `FunctionInvokingChatClient` observing conversationId?

dotnet/src/Microsoft.Agents.AI.AGUI/Shared/AGUIChatMessageExtensions.cs
```diff
@@ -15,34 +16,177 @@ internal static class AGUIChatMessageExtensions
...
+        JsonSerializerOptions jsonSerializerOptions)
     {
         foreach (var message in aguiMessages)
         {
-            yield return new ChatMessage(
-                MapChatRole(message.Role),
-                message.Content);
+            var role = MapChatRole(message.Role);
+
+            if (message is AGUIToolMessage toolMessage)
```
DeagleGross: nit: use switch expression instead of multiple ifs.

dotnet/src/Microsoft.Agents.AI.AGUI/Shared/AIToolExtensions.cs
```diff
@@ -0,0 +1,56 @@
...
+    public static IEnumerable<AGUITool> AsAGUITools(this IEnumerable<AITool> tools)
+    {
+        if (tools is null)
+        {
+            yield break;
+        }
+
+        foreach (var tool in tools)
+        {
+            // Convert both AIFunctionDeclaration and AIFunction (which extends it) to AGUITool
```
DeagleGross, line 23: why do we not need to handle `AIFunction` here? Not all `AIFunctions` are `AIFunctionDeclaration`, so we need to handle it as well?

DeagleGross (approved):

dotnet/src/Microsoft.Agents.AI.AGUI/Shared/AIToolExtensions.cs
javiercn: This is for the client tools. I think it's a requirement that any client tool you register is an AIFunctionDeclaration if you want it to be invokable through AG-UI, as otherwise I don't see a way in which you can provide the Json Schema for the parameters.

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
javiercn: Because otherwise `FunctionInvokingChatClient` will only send new messages, and AG-UI expects the full message history to be sent in every turn, so we need to "hide it" from the intermediate FICC.

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
```diff
@@ -0,0 +1,325 @@
...
+        }
+
+        // Create a new JsonSerializerOptions based on the provided one
+        var combinedOptions = new JsonSerializerOptions(jsonSerializerOptions);
+
+        // Add the AGUI context to the type info resolver chain if not already present
+        if (!combinedOptions.TypeInfoResolverChain.Any(r => r == AGUIJsonSerializerContext.Default))
+        {
+            combinedOptions.TypeInfoResolverChain.Insert(0, AGUIJsonSerializerContext.Default);
+        }
```
stephentoub, line 60: What goes wrong if you just always insert it?

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
```diff
@@ -0,0 +1,325 @@
...
+            innerOptions.AdditionalProperties["agui_thread_id"] = options.ConversationId;
+            innerOptions.ConversationId = null;
+        }
+
+        await foreach (var update in base.GetStreamingResponseAsync(messages, innerOptions, cancellationToken).ConfigureAwait(false))
+        {
+            if (conversationId == null && firstUpdate == null)
+            {
+                firstUpdate = update;
+                if (firstUpdate.AdditionalProperties != null && firstUpdate.AdditionalProperties.TryGetValue("agui_thread_id", out var threadIdObj) && threadIdObj is string threadId)
```
stephentoub: Can you use the `TryGetValue<T>` method? e.g.
```suggestion
                if (firstUpdate.AdditionalProperties?.TryGetValue("agui_thread_id", out string? threadId) is true)
```

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
```diff
@@ -0,0 +1,325 @@
...
+        {
+            this._httpService = new AGUIHttpService(httpClient, endpoint);
+            this._jsonSerializerOptions = jsonSerializerOptions ?? AGUIJsonSerializerContext.Default.Options;
+            this._logger = serviceProvider?.GetService(typeof(ILogger<AGUIChatClient>)) as ILogger ?? NullLogger.Instance;
+
+            // Use BaseAddress if endpoint is empty, otherwise parse as relative or absolute
+            Uri metadataUri = string.IsNullOrEmpty(endpoint) && httpClient.BaseAddress is not null
+                ? httpClient.BaseAddress
+                : new Uri(endpoint, UriKind.RelativeOrAbsolute);
+            this.Metadata = new ChatClientMetadata("AGUI", metadataUri, null);
```
stephentoub: Nit: the provider name ends up being used in telemetry, and most providers use all lowercase, so consider "agui" or "ag-ui"

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
```diff
@@ -0,0 +1,325 @@
...
+            IEnumerable<ChatMessage> messages,
+            ChatOptions? options = null,
+            [EnumeratorCancellation] CancellationToken cancellationToken = default)
+        {
+            if (messages is null)
+            {
+                throw new ArgumentNullException(nameof(messages));
+            }
+
+            var runId = $"run_{Guid.NewGuid()}";
```
stephentoub: These guids will all have dashes in them. Consider passing in the "N" format string.

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
```diff
@@ -0,0 +1,325 @@
...
+                        // We store it on additional properties of the function call content, which will be passed down
+                        // in the next turn.
+                        fcc.AdditionalProperties ??= [];
+                        fcc.AdditionalProperties["agui_thread_id"] = threadId;
+                    }
+                    else
+                    {
+                        // Hide the server result call from the FunctionInvokingChatClient.
+                        // The wrapping client will unwrap it and present it as a normal function result.
+                        update.Contents[0] = new ServerFunctionCallContent(fcc);
```
stephentoub, line 257: @westey-m, @jozkee, I've been thinking about how we want to handle representing function call information that's notification-only rather than a request for invocation. There are a variety of cases where the server is going to invoke some function and then wants to tell the client about it; currently that requires us to introduce new content types, e.g. McpServerToolCallContent, or this ServerFunctionCallContent as @javiercn has done here. Presumably we'll need something similar for other invocations, like a server-side OpenAPI call. And while we might _want_ to add such specialized content, right now we _need_ to. We also have the issue that on the client, when you receive a FunctionCallContent, you don't actually know if you're supposed to do something with it; if you have a FunctionInvokingChatClient in the mix, it may have handled the invocation for you. So you need to check the full response / response stream to see if there's a corresponding FunctionResultContent.

I wonder if we could simplify a lot of this just by adding a FunctionCallContent.Handled property that indicates whether the call content is a request or purely informative. It'd be set to true by a leaf IChatClient that's creating one to represent some server-side call, and it'd be set to true by FunctionInvokingChatClient when it invokes the corresponding function. FunctionInvokingChatClient would also ignore any FCC where Handled was already true.

Thoughts?

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
```diff
@@ -0,0 +1,325 @@
...
+                yield return update;
+            }
+        }
+
+        // Extract the thread id from the options additional properties
+        private static string? ExtractThreadIdFromOptions(ChatOptions? options)
+        {
+            if (options?.AdditionalProperties == null ||
+              !options.AdditionalProperties.TryGetValue("agui_thread_id", out var threadIdObj) ||
+              threadIdObj is not string threadId ||
```
stephentoub: See previous comment on `TryGetValue<T>`

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
```diff
@@ -0,0 +1,325 @@
...
+            if (serviceType == typeof(ChatClientMetadata))
+            {
+                return this.Metadata;
+            }
+
+            return null;
+        }
+    }
+
+    private class ServerFunctionCallContent(FunctionCallContent functionCall) : AIContent
```
stephentoub, line 319: If I'm understanding correctly, you're just using this as an internal implementation detail to bypass the FICC's invocation? You wrap an FCC before handing it back through the FICC and then unwrap it after it passes through the FICC and before actually returning it to the caller? Just verifying because if it were to escape to anything that might try to serialize it, we'll have to do more.

dotnet/src/Microsoft.Agents.AI.AGUI/Shared/AGUIJsonSerializerContext.cs
```diff
@@ -12,18 +13,50 @@ namespace Microsoft.Agents.AI.Hosting.AGUI.AspNetCore;
 namespace Microsoft.Agents.AI.AGUI;
 #endif
 
+// All JsonSerializable attributes below are required for AG-UI functionality:
+// - AG-UI message types (AGUIMessage, AGUIUserMessage, etc.) for protocol communication
+// - Event types (BaseEvent, RunStartedEvent, etc.) for server-sent events streaming
+// - Tool-related types (AGUITool, AGUIToolCall, AGUIFunctionCall) for tool calling support
+// - Primitive and dictionary types (string, int, Dictionary, JsonElement) are required for
+//   serializing tool call parameters and results which can contain arbitrary data types
```
stephentoub, line 21: Are any Microsoft.Extensions.AI types ever referenced statically or dynamically from any of these objects? If yes, instead of having consumers use AGUIJsonSerializerContext directly, we should have a JsonSerializerOptions that first chains in the AIJsonUtilities or AgentAbstractionsJsonUtilities options, so that their resolvers are considered first.

stephentoub (approved):

dotnet/src/Microsoft.Agents.AI.AGUI/Shared/AGUIJsonSerializerContext.cs
javiercn: Nope, everything from MEAI and MAF gets mapped into AGUI "transport" types.

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
javiercn: Yep, this is just to bypass FICC and it shouldn't leak outside of it.

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
javiercn: I've gone with `ag-ui`

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
javiercn: Resolving this so I can merge.

I'll share some thoughts with you.

markwallace-microsoft (approved):

markwallace-microsoft:
Integration failure unrelated

```
  Passed OpenAIAssistant.IntegrationTests.OpenAIAssistantIRunTests.RunWithChatMessageReturnsExpectedResultAsync [6 s]
[xUnit.net 00:03:09.00]     OpenAIAssistant.IntegrationTests.OpenAIAssistantClientExtensionsTests.CreateAIAgentAsync_WithHostedCodeInterpreter_RunsCodeAsync(createMechanism: "CreateWithChatClientAgentOptionsSync") [FAIL]
```

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
jozkee: > I wonder if we could simplify a lot of this just by adding a FunctionCallContent.Handled

This wouldn't work for McpServerToolCallContent unless we choose to replace it with FunctionCallContent, but that conflicts with adding a [ServerSideContent](https://github.com/dotnet/extensions/issues/6492#issuecomment-3603938270) and use it as base for mcp contents.  I think ServerSideContent may be enough to aid cases like this one.

PR #1905, opened 2025-11-04
.NET: [BREAKING] Simplify TextSearchProvider construction and improve Mem0Provider scoping.

westey-m:
### Description

- Merge TextSearchProvider constructors to simplify default case where there may be serialized state.
- Split Mem0 scope from options so that scope options can be stored in state, while other options are provided as needed.

### Contribution Checklist


dotnet/src/Microsoft.Agents.AI/Data/TextSearchProvider.cs
```diff
@@ -44,82 +44,72 @@ public sealed class TextSearchProvider : AIContextProvider
...
+        this._searchAsync = Throw.IfNull(searchAsync);
         this._logger = loggerFactory?.CreateLogger<TextSearchProvider>();
-        this._recentMessageRolesIncluded = this._options.RecentMessageRolesIncluded ?? [ChatRole.User];
-
+        this._recentMessageMemoryLimit = Throw.IfLessThan(options?.RecentMessageMemoryLimit ?? 0, 0);
+        this._recentMessageRolesIncluded = options?.RecentMessageRolesIncluded ?? [ChatRole.User];
+        this._searchTime = options?.SearchTime ?? TextSearchProviderOptions.TextSearchBehavior.BeforeAIInvoke;
+        this._contextPrompt = options?.ContextPrompt ?? DefaultContextPrompt;
+        this._citationsPrompt = options?.CitationsPrompt ?? DefaultCitationsPrompt;
+        this._contextFormatter = options?.ContextFormatter;
```
markwallace-microsoft, line 78: What about keeping the `this._options` variable to store an instance of `TextSearchProviderOptions` and have a factory method like `TextSearchProviderOptions.CreateWithDefaults(TextSearchProviderOptions options)`

westey-m: This is just an internal implementation detail, and I think what's there works for now. Let's revisit if it becomes onerous.

PR #1907, opened 2025-11-04
.NET: Improve fidelity of OpenAI Responses server and add Conversations

ReubenBond:
### Motivation and Context

This PR adds a more complete API surface area for serving agents via the OpenAI Responses API, including adding support for the Conversations API. This only includes in-memory support so far and the intention is just that this is used for development and testing for now (since there is no external storage included, for example). Further improvements could include adding storage and other customizations.

The purpose is to support the DevUI.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Models/SortOrder.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Models/SortOrder.cs
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft. All rights reserved.
+﻿// Copyright (c) Microsoft. All rights reserved.
 
 using System;
 using System.Text.Json;
--- a/dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Converters/ItemParamConverter.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Converters/ItemParamConverter.cs
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft. All rights reserved.
+﻿// Copyright (c) Microsoft. All rights reserved.
 
 using System;
 using System.Text.Json;
--- a/dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Converters/ResponsesMessageItemParamConverter.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Converters/ResponsesMessageItemParamConverter.cs
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft. All rights reserved.
+﻿// Copyright (c) Microsoft. All rights reserved.
 
 using System;
 using System.Text.Json;
--- a/dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Converters/AgentReferenceExtensions.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Converters/AgentReferenceExtensions.cs
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft. All rights reserved.
+﻿// Copyright (c) Microsoft. All rights reserved.
 
 using Microsoft.Agents.AI.Hosting.OpenAI.Responses.Models;
 
--- a/dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Converters/ItemContentConverter.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Converters/ItemContentConverter.cs
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft. All rights reserved.
+﻿// Copyright (c) Microsoft. All rights reserved.
 
 using System;
 using Microsoft.Agents.AI.Hosting.OpenAI.Responses.Models;


dotnet/tests/Microsoft.Agents.AI.Hosting.OpenAI.UnitTests/OpenAIConversationsConformanceTests.cs
```diff
@@ -0,0 +1,1206 @@
...
+    private const string TracesBasePath = "ConformanceTraces/Conversations";
+    private WebApplication? _app;
+    private HttpClient? _httpClient;
+
+    /// <summary>
+    /// Loads a JSON file from the conformance traces directory.
+    /// </summary>
+    private static string LoadTraceFile(string relativePath)
+    {
+        var fullPath = System.IO.Path.Combine(TracesBasePath, relativePath);
```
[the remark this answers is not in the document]
ReubenBond: What's wrong with this?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Conversations/ConversationsHttpHandler.cs
```diff
@@ -0,0 +1,334 @@
...
+    /// Initializes a new instance of the <see cref="ConversationsHttpHandler"/> class.
+    /// </summary>
+    /// <param name="storage">The conversation storage service.</param>
+    /// <param name="conversationIndex">Optional conversation index service.</param>
+    public ConversationsHttpHandler(IConversationStorage storage, IAgentConversationIndex? conversationIndex)
+    {
+        this._storage = storage ?? throw new ArgumentNullException(nameof(storage));
+        this._conversationIndex = conversationIndex;
+    }
+    /// <summary>
```
stephentoub, line 35: Nit:
```suggestion

    /// <summary>
```

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Conversations/ConversationsHttpHandler.cs
```diff
@@ -0,0 +1,334 @@
...
+    /// Creates a new conversation.
+    /// </summary>
+    public async Task<IResult> CreateConversationAsync(
+        [FromBody] CreateConversationRequest request,
+        CancellationToken cancellationToken)
+    {
+        Dictionary<string, string> metadata = request.Metadata ?? [];
+        var conversation = new Conversation
+        {
+            Id = $"conv_{Guid.NewGuid():N}",
```
stephentoub: Should this be using IdGenerator? I'm unclear when we're choosing to use it vs NewGuid

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Conversations/ConversationsHttpHandler.cs
```diff
@@ -0,0 +1,334 @@
...
+    public async Task<IResult> ListItemsAsync(
+        string conversationId,
+        [FromQuery] int? limit,
+        [FromQuery] string? order,
+        [FromQuery] string? after,
+        [FromQuery] string[]? include,
+        CancellationToken cancellationToken)
+    {
+        // Validate limit parameter
+        if (limit is not null && limit.Value < 1)
```
stephentoub:
```suggestion
        if (limit is < 1)
```
?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Conversations/ConversationsHttpHandler.cs
```diff
@@ -0,0 +1,334 @@
...
+            {
+                Error = new ErrorDetails
+                {
+                    Message = $"Conversation '{conversationId}' not found.",
+                    Type = "invalid_request_error"
+                }
+            });
+        }
+
+        var result = await this._storage.ListItemsAsync(conversationId, limit ?? 20, ParseOrder(order ?? "desc"), after, cancellationToken).ConfigureAwait(false);
```
stephentoub: Where does this 20 come from? Is that matching some default OpenAI has? Consider putting it into a named const somewhere.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Conversations/IConversationStorage.cs
```diff
@@ -0,0 +1,96 @@
...
+    /// </summary>
+    /// <param name="conversationId">The conversation ID.</param>
+    /// <param name="limit">Maximum number of items to return (default: 20, max: 100).</param>
+    /// <param name="order">Sort order (default: Descending).</param>
+    /// <param name="after">Cursor for pagination - return items after this ID.</param>
+    /// <param name="cancellationToken">Cancellation token.</param>
+    /// <returns>A list response with items and pagination info.</returns>
+    Task<ListResponse<ItemResource>> ListItemsAsync(
+        string conversationId,
+        int limit = 20,
```
stephentoub: Another place that 20 shows up... presumably they could refer to the same const. Or, maybe both limit and order should be nullable instead of baking their default values into call sites?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Models/SortOrder.cs
```diff
@@ -0,0 +1,54 @@
...
+using System;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Agents.AI.Hosting.OpenAI.Conversations.Models;
+
+/// <summary>
+/// Specifies the sort order for list operations.
+/// </summary>
+[JsonConverter(typeof(SortOrderJsonConverter))]
```
stephentoub, line 12: Could this use `[JsonStringEnumMemberName]` on each member and `JsonStringEnumConverter<SortOrder>` on the type instead of a custom converter?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Conversations/InMemoryConversationStorage.cs
```diff
@@ -0,0 +1,337 @@
...
+        {
+            var afterIndex = allItems.FindIndex(m => m.Id == after);
+            if (afterIndex >= 0)
+            {
+                filtered = allItems.Skip(afterIndex + 1);
+            }
+        }
+
+        var result = filtered.Take(limit + 1).ToList();
+        var hasMore = result.Count > limit;
```
stephentoub: You could use TryGetNonEnumeratedCount to try to avoid the intermediate list. Not sure it matters, though.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Conversations/InMemoryConversationStorage.cs
```diff
@@ -0,0 +1,337 @@
...
+
+        public bool RemoveItem(string itemId)
+        {
+            lock (this._lock)
+            {
+                return this._items.Remove(itemId);
+            }
+        }
+#else
+        private readonly List<ItemResource> _items = [];
```
stephentoub, line 275: Does this lead to algorithmic complexity differences between .NET 8 and .NET 9+?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Conversations/InMemoryConversationStorage.cs
```diff
@@ -0,0 +1,337 @@
...
+        {
+            lock (this._lock)
+            {
+                var item = this._items.FirstOrDefault(i => i.Id == itemId);
+                if (item != null)
+                {
+                    this._items.Remove(item);
+                    return true;
+                }
+                return false;
```
stephentoub: Presumably an item with this id should only ever be in the list once? You could simplify this down to:
```csharp
return this._items.RemoveAll(i => i.Id == itemId) > 0;
```
?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Conversations/InMemoryConversationStorage.cs
```diff
@@ -0,0 +1,337 @@
...
+            {
+                this._conversation = conversation;
+            }
+        }
+
+        public void AddItem(ItemResource item)
+        {
+            lock (this._lock)
+            {
+                if (this._items.Any(i => i.Id == item.Id))
```
stephentoub: Nit: could use `List<T>.Exists`

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Conversations/InMemoryConversationStorage.cs
```diff
@@ -0,0 +1,337 @@
...
+                }
+                this._items.Add(item);
+            }
+        }
+
+        public ItemResource? GetItem(string itemId)
+        {
+            lock (this._lock)
+            {
+                return this._items.FirstOrDefault(i => i.Id == itemId);
```
stephentoub: Nit: could use `List<T>.Find`

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Converters/ItemParamConverter.cs
```diff
@@ -0,0 +1,59 @@
...
+            "image_generation_call" => JsonSerializer.Deserialize(jsonText, OpenAIHostingJsonContext.Default.ImageGenerationToolCallItemParam),
+            "code_interpreter_call" => JsonSerializer.Deserialize(jsonText, OpenAIHostingJsonContext.Default.CodeInterpreterToolCallItemParam),
+            "local_shell_call" => JsonSerializer.Deserialize(jsonText, OpenAIHostingJsonContext.Default.LocalShellToolCallItemParam),
+            "local_shell_call_output" => JsonSerializer.Deserialize(jsonText, OpenAIHostingJsonContext.Default.LocalShellToolCallOutputItemParam),
+            "mcp_list_tools" => JsonSerializer.Deserialize(jsonText, OpenAIHostingJsonContext.Default.MCPListToolsItemParam),
+            "mcp_approval_request" => JsonSerializer.Deserialize(jsonText, OpenAIHostingJsonContext.Default.MCPApprovalRequestItemParam),
+            "mcp_approval_response" => JsonSerializer.Deserialize(jsonText, OpenAIHostingJsonContext.Default.MCPApprovalResponseItemParam),
+            "mcp_call" => JsonSerializer.Deserialize(jsonText, OpenAIHostingJsonContext.Default.MCPCallItemParam),
+            _ => throw new JsonException($"Unknown ItemParam type: {type}")
```
stephentoub: I wonder if such unknown types should be ignored? I'm thinking about cases where OpenAI adds new kinds here but we've not yet updated this to accomodate, and whether it'd be better to accept what we can rather than blow up the whole request for something new.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Models/ConversationReference.cs
```diff
@@ -40,8 +41,10 @@ public static ConversationReference FromObject(string id, Dictionary<string, str
 /// <summary>
 /// JSON converter for ConversationReference that handles both string (conversation ID) and object representations.
 /// </summary>
+[ExcludeFromCodeCoverage]
```
stephentoub: What's driving the addition of these [ExcludeFromCodeCoverage] attributes?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Streaming/FunctionApprovalRequestEventGenerator.cs
```diff
@@ -0,0 +1,52 @@
...
+            SequenceNumber = seq.Increment(),
+            OutputIndex = outputIndex,
+            RequestId = approvalRequest.Id,
+            ItemId = idGenerator.GenerateMessageId(),
+            FunctionCall = new FunctionCallInfo
+            {
+                Id = approvalRequest.FunctionCall.CallId,
+                Name = approvalRequest.FunctionCall.Name,
+#pragma warning disable IL2026 // Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code
+#pragma warning disable IL3050 // Calling members annotated with 'RequiresDynamicCodeAttribute' may break functionality when AOT compiling.
```
stephentoub: You can avoid these if you change the line below to be more like:
```csharp
JsonSerializer.SerializeToElement(approvalRequest.FunctionCall.Arguments, jsonSerializerOptions.GetTypeInfo(typeof(IDictionary<string, object>)))
```
These suppressions are a red flag, so if we can avoid them (even if doing so doesn't actually change the correctness), we should.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Streaming/FunctionApprovalRequestEventGenerator.cs
```diff
@@ -0,0 +1,52 @@
...
+using Microsoft.Agents.AI.Hosting.OpenAI.Responses.Models;
+using Microsoft.Extensions.AI;
+
+namespace Microsoft.Agents.AI.Hosting.OpenAI.Responses.Streaming;
+
+/// <summary>
+/// A generator for streaming events from function approval request content.
+/// This is a non-standard DevUI extension for human-in-the-loop scenarios.
+/// </summary>
+#pragma warning disable MEAI001 // Type is for evaluation purposes only and is subject to change or removal in future updates.
```
stephentoub: Might be worth just suppressing this for the whole library

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/AgentRunResponseExtensions.cs
```diff
@@ -115,11 +112,35 @@ public static IEnumerable<ItemResource> ToItemResource(this ChatMessage message,
 
         if (contents.Count > 0)
         {
-            yield return new ResponsesAssistantMessageItemResource
+            List<ItemContent> contentArray = contents;
+            string messageId = idGenerator.GenerateMessageId();
+
+            yield return message.Role.Value.ToUpperInvariant() switch
```
stephentoub: I realize the string allocation here is a drop in the bucket, but I can't help but comment...

I don't think it would be meaningfully less readable to instead do:
```csharp
yield return
    message.Role == ChatRole.User ? new ResponsesUserMessageItemResource
    {
        Id = messageId,
        Status = ResponsesMessageItemResourceStatus.Completed,
        Content = contentArray
    } :
    message.Role == ChatRole.System ? ...
```
etc. and it would avoid needing to pretty much always allocate the upper-case string.

Alternatively, since the message.Role will 99.9% of the time already be lower-cased, you could suppress the analyzer that I assume drove you to use ToUpperInvariant instead of ToLowerInvariant, and just use ToLowerInvariant with the switch, as that will almost never need to allocate.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/AgentRunResponseUpdateExtensions.cs
```diff
@@ -192,4 +205,57 @@ static bool IsSameValue(string? str1, string? str2) =>
...
+            executorId = execEvent.ExecutorId;
+        }
+        JsonElement eventData;
+        if (JsonSerializer.IsReflectionEnabledByDefault)
+        {
+            JsonElement? dataElement = null;
+            if (workflowEvent.Data is not null)
+            {
+#pragma warning disable IL2026 // Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code
+#pragma warning disable IL3050 // Calling members annotated with 'RequiresDynamicCodeAttribute' may break functionality when AOT compiling.
```
stephentoub: Similar comments for all of these suppressions

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/AIAgentResponseExecutor.cs
```diff
@@ -0,0 +1,70 @@
...
+            ConversationId = request.Conversation?.Id,
+            Temperature = (float?)request.Temperature,
+            TopP = (float?)request.TopP,
+            MaxOutputTokens = request.MaxOutputTokens,
+            Instructions = request.Instructions,
+            ModelId = request.Model,
+            AdditionalProperties = new AdditionalPropertiesDictionary
+            {
+                [nameof(CreateResponse)] = request
+            }
```
stephentoub: What is looking at this? You might instead do:
```csharp
RawRepresentationFactory = _ => request,
```
though in both cases presumably it's not actually going to be particularly usable, given that CreateResponse is internal.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/AIAgentResponseExecutor.cs
```diff
@@ -0,0 +1,70 @@
...
+        var options = new ChatClientAgentRunOptions(chatOptions);
+
+        // Convert input to chat messages
+        var messages = new List<ChatMessage>();
+
+        // If instructions are provided in the request, prepend them as a system message
+        if (!string.IsNullOrWhiteSpace(request.Instructions))
+        {
+            messages.Add(new ChatMessage(ChatRole.System, request.Instructions));
+        }
```
stephentoub: For an agent that understands ChatOptions.Instructions, this is going to include the instructions twice. If we're going to include this as a system message, we should probably _not_ also store it as ChatOptions.Instructions.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/HostedAgentResponseExecutor.cs
```diff
@@ -0,0 +1,125 @@
...
+        };
+        var options = new ChatClientAgentRunOptions(chatOptions);
+
+        // Convert input to chat messages
+        var messages = new List<ChatMessage>();
+
+        // If instructions are provided in the request, prepend them as a system message
+        if (!string.IsNullOrWhiteSpace(request.Instructions))
+        {
+            messages.Add(new ChatMessage(ChatRole.System, request.Instructions));
```
stephentoub: Same Instructions comment

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/InMemoryResponsesService.cs
```diff
@@ -0,0 +1,524 @@
...
+        {
+            hostedExecutor.ValidateAgent(request);
+        }
+
+        if (request.Stream == true)
+        {
+            throw new InvalidOperationException("Cannot create a streaming response using CreateResponseAsync. Use CreateResponseStreamingAsync instead.");
+        }
+
+        var responseId = $"resp_{Guid.NewGuid():N}";
```
stephentoub: Same question as earlier about IdGenerator vs NewGuid. I'll stop asking.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/InMemoryResponsesService.cs
```diff
@@ -0,0 +1,524 @@
...
+        state?.CancellationTokenSource?.Cancel();
+
+        // Remove the response
+        this._cache.Remove(responseId);
+        return Task.FromResult(true);
+    }
+
+    public Task<ListResponse<ItemResource>> ListResponseInputItemsAsync(
+        string responseId,
+        int limit = 20,
```
stephentoub: Another one of those 20s :)

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/InMemoryResponsesService.cs
```diff
@@ -0,0 +1,524 @@
...
+            Model = request.Model ?? "default",
+            Status = initialStatus,
+            Error = null,
+            IncompleteDetails = null,
+            Output = [],
+            Instructions = request.Instructions,
+            Usage = ResponseUsage.Zero,
+            ParallelToolCalls = request.ParallelToolCalls ?? true,
+            Tools = [],
+            ToolChoice = default,
```
stephentoub: These don't get propagated from the request?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/InMemoryResponsesService.cs
```diff
@@ -0,0 +1,524 @@
...
+            var sequenceNumber = state.StreamingUpdates.Count + 1;
+            var cancelledEvent = new StreamingResponseCancelled
+            {
+                SequenceNumber = sequenceNumber,
+                Response = state.Response
+            };
+
+            state.AddStreamingEvent(cancelledEvent);
+        }
+        catch (Exception ex) when (ex is not OperationCanceledException)
```
stephentoub: This when condition is a nop, as the previous catch block already handled all OperationCanceledExceptions

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/IResponsesService.cs
```diff
@@ -0,0 +1,96 @@
...
+    /// <param name="responseId">The ID of the response.</param>
+    /// <param name="limit">Maximum number of items to return (1-100).</param>
+    /// <param name="order">Sort order ("asc" or "desc").</param>
+    /// <param name="after">Return items after this ID.</param>
+    /// <param name="before">Return items before this ID.</param>
+    /// <param name="cancellationToken">Cancellation token.</param>
+    /// <returns>A list response with items and pagination info.</returns>
+    Task<ListResponse<ItemResource>> ListResponseInputItemsAsync(
+        string responseId,
+        int limit = 20,
```
stephentoub: :)

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/MemoryCacheExtensions.cs
```diff
@@ -0,0 +1,155 @@
...
+    {
+        // Fast path: check if the value already exists
+        if (memoryCache.TryGetValue(key, out object? value))
+        {
+            return ((T?)value, false);
+        }
+
+        // Get or create a semaphore for this cache key
+        bool isOwner = false;
+        int semaphoreKey = (memoryCache, key).GetHashCode();
```
stephentoub: Why is the key the hash code of `(memoryCache, key)` rather than just being `(memoryCache, key)`? Aren't we setting ourselves up for potential collisions?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/OpenAIHostingJsonUtilities.cs
```diff
@@ -4,16 +4,62 @@
...
+    private static JsonSerializerOptions CreateDefaultOptions()
+    {
+        // Start with our source-generated context
+        JsonSerializerOptions options = new(OpenAIHostingJsonContext.Default.Options);
+
+        // Chain with agent abstraction types
+        options.TypeInfoResolverChain.Add(AgentAbstractionsJsonUtilities.DefaultOptions.TypeInfoResolver!);
+
+        // Chain with AIContent types from Microsoft.Extensions.AI
+        options.TypeInfoResolverChain.Add(AIJsonUtilities.DefaultOptions.TypeInfoResolver!);
```
stephentoub: This shouldn't be necessary, as AgentAbstractionsJsonUtilities already chains it in.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Conversations/ConversationsHttpHandler.cs
ReubenBond: Yes, the default limit is 20 according to OpenAI docs. I'll make it a named constant

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Conversations/IConversationStorage.cs
ReubenBond: I'll make them nullable

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Models/SortOrder.cs
ReubenBond: I tried but it didn't work. The docs say that it's only supported in .NET 9.0 and later: https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/source-generation#custom-enum-member-names

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Conversations/InMemoryConversationStorage.cs
ReubenBond: Yes, but I don't think it matters enough given the typical size of a conversation. This is only for in-memory use

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Converters/ItemParamConverter.cs
ReubenBond: That makes sense. I was also thinking of making this extensible (but... not now, maybe if we have internal uses for that).

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Models/ConversationReference.cs
ReubenBond: I saw it in other cases, like on the JsonContexts, and replicated it

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/AgentRunResponseExtensions.cs
ReubenBond: I'll take the ChatRole approach, and create a private readonly ChatRole for "developer"

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/AIAgentResponseExecutor.cs
ReubenBond: I'll delete this bit instead and rely on ChatOptions.Instructions

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/AIAgentResponseExecutor.cs
ReubenBond: I'll just remove it. YAGNI

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Models/ConversationReference.cs
stephentoub: Got it. It's on the JsonContexts because there's a ton of source-generated code that gets injected into those and that we don't need to test. I think we want to consider custom converters testable, though.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/MemoryCacheExtensions.cs
ReubenBond: Will fix. Collisions here are not a big deal. This code can be simplified a bit, too.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/InMemoryResponsesService.cs
ReubenBond: Fixed to propagate all properties at all call sites

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Conversations/Models/Conversation.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Agents.AI.Hosting.OpenAI.Conversations.Models;
+
+/// <summary>
+/// Represents a conversation in the system.
+/// </summary>
+internal sealed record Conversation
```
javiercn, line 12: Is the record vs class intentional?

ReubenBond: Yes, I used record sparingly, only where I wanted some benefit from it, like 'with' syntax

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Conversations/SortOrderExtensions.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using Microsoft.Agents.AI.Hosting.OpenAI.Models;
+
+namespace Microsoft.Agents.AI.Hosting.OpenAI.Conversations;
+
+/// <summary>
+/// Extension methods for <see cref="SortOrder"/>.
+/// </summary>
```
javiercn, line 9: nit: Are the `///` for the internal stuff expected?

ReubenBond: I'll remove them. There was a point where I made these types public so that I could consume them in another project, but I will most likely go with the source sharing approach for now.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Converters/ItemParamConverter.cs
```diff
@@ -0,0 +1,57 @@
...
+        using var doc = JsonDocument.ParseValue(ref reader);
+        var root = doc.RootElement;
+
+        if (!root.TryGetProperty("type", out var typeElement))
+        {
+            throw new JsonException("ItemParam must have a 'type' property");
+        }
+
+        var type = typeElement.GetString();
+        var jsonText = root.GetRawText();
```
javiercn: Not sure if you need the string conversion here

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Converters/ResponsesMessageItemParamConverter.cs
```diff
@@ -0,0 +1,59 @@
...
+        if (!root.TryGetProperty("role", out var roleElement))
+        {
+            throw new JsonException("ResponsesMessageItemParam must have a 'role' property");
+        }
+
+        var role = roleElement.GetString();
+
+        return role switch
+        {
+            "user" => JsonSerializer.Deserialize(root.GetRawText(), OpenAIHostingJsonContext.Default.ResponsesUserMessageItemParam),
```
javiercn: Same as other serializer

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/Models/WorkflowEventData.cs
```diff
@@ -0,0 +1,38 @@
...
+
+/// <summary>
+/// Represents workflow event data for serialization.
+/// </summary>
+internal sealed class WorkflowEventData
+{
+    /// <summary>
+    /// The type of the workflow event.
+    /// </summary>
+    [JsonPropertyName("event_type")]
```
javiercn, line 16: Is the snake case expected? (probably yes, but asking because I had copilot generate this erroneously a few times even after giving it the spec)

ReubenBond: Is there a spec for WorkflowEventData? I was just following the python implementation. In general, yes, they are all snake case for OpenAI

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/SseJsonResult.cs
```diff
@@ -0,0 +1,75 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Features;
+
+namespace Microsoft.Agents.AI.Hosting.OpenAI;
+
+/// <summary>
+/// IResult implementation for streaming JSON data using Server-Sent Events (SSE).
+/// </summary>
+/// <typeparam name="T">The type of items to stream.</typeparam>
+internal sealed class SseJsonResult<T> : IResult
```
javiercn, line 19: nit: Should we consider putting this in shared? I think I also have one of these on the AG-UI implementation, and I suspect you lifted it to from the asp.net repo?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Conversations/SortOrderExtensions.cs
ReubenBond: Actually, let's keep them. Many of the doc comments do serve as useful documentation even if they're not visible externally.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/SseJsonResult.cs
ReubenBond: I don't recall where this came from. I think it evolved over time (it's been through a couple of demos & experiments). We could consider lifting it in a subsequent PR if it's generally useful

javiercn (approved):

PR #1935, opened 2025-11-05
.NET: [BREAKING] refactor: Normalize WorkflowBuilder APIs

lokitoth:
### Motivation and Context

The WorkflowBuilder APIs have a few rough edges to them. For example, they reverse the order of source and target between `AddFanInEdge` and `AddFanOutEdge`, among other issues.

### Description

* "partitioner" => "assigner"
* normalize ordering so sources always to the left of targets for edges
* normalize parameter ordering so sources and targets are always first arguments

### Contribution Checklist


lokitoth:
> Wonder if we should rename partitioner to assigner in Python?

Synced offline; updated the name to better match Python's

PR #1943, opened 2025-11-05
.NET: [BREAKING] .NET: Add Workflow on-Build() Orphan Validation

lokitoth:
### Motivation and Context

To improve consistency between Python and .NET, we want to bring .NET's on-build Workflow validation closer to what happens on Python.

### Description

* Adds orphan validation to workflow builder
* Adds tests for workflow validation
* Expands on the underlying reasoning why type validation is not supported

Breaking:
* Workflows with orphan nodes will no longer pass validation unless orphan checking is disabled.

### Contribution Checklist


lokitoth:
We can also introduce this as a non-breaking change (by flipping the default behaviour), but I would rather try to get it in.

dotnet/src/Microsoft.Agents.AI.Workflows/WorkflowBuilder.cs
```diff
@@ -433,13 +494,15 @@ private Workflow BuildInternal(Activity? activity = null)
     /// <summary>
     /// Builds and returns a workflow instance.
     /// </summary>
+    /// <param name="validateOrphans">Specifies whether workflow validation should check for Executor nodes that are
+    /// not reachable from the starting executor.</param>
     /// <exception cref="InvalidOperationException">Thrown if there are unbound executors in the workflow definition,
     /// or if the start executor is not bound.</exception>
-    public Workflow Build()
+    public Workflow Build(bool validateOrphans = true)
```
adityamandaleeka, line 501: FYI, changing this default in the future will be a breaking change unless we can guarantee that all callers will be recompiled.

alliscode (approved):

crickman (commented):
Declarative workflows may have orphans and are expected to be executed anyway.

Does an orphan really mean the workflow shouldn't run?

Please don't break me! ;)

dotnet/src/Microsoft.Agents.AI.Workflows/WorkflowBuilder.cs
crickman: This parameter looks like the solution to my question.

In terms of aditya's observation, what about a `ValidateOrphans` property on the builder? (since `Build` is an instance method)

lokitoth:
> Declarative workflows may have orphans and are expected to be executed anyway.
> 
> Does an orphan really mean the workflow shouldn't run?
> 
> Please don't break me! ;)

Yep, that's why this can be turned off: I was running the Declarative Unit Tests and saw a lot of them rely on having orphans. While I would prefer the default behaviour err on the stricter side, there is nothing all that harmful about having them, which is it why it is possible to disable.

If people feel strongly enough, I can flip it to be opt-in

dotnet/src/Microsoft.Agents.AI.Workflows/WorkflowBuilder.cs
lokitoth: Yep; trying to get this in now to avoid having to flip it later.

PR #1977, opened 2025-11-06
.NET: Don't add OpenAIResponses as part of Dev UI

danroth27:
You should be able to add and remove Dev UI without impacting your other production endpoints. This is the pattern used by Swagger UI, where you need to separate set up the OpenAPI endpoint that the Swagger UI depends on.

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.DevUI/DevUIExtensions.cs
+++ b/dotnet/src/Microsoft.Agents.AI.DevUI/DevUIExtensions.cs
@@ -16,7 +16,6 @@ public static class DevUIExtensions
     {
         ArgumentNullException.ThrowIfNull(builder);
         builder.Services.AddOpenAIConversations();
-        builder.Services.AddOpenAIResponses();
 
         return builder;
     }
@@ -34,7 +33,6 @@ public static class DevUIExtensions
         group.MapDevUI(pattern: "/devui");
         group.MapEntities();
         group.MapOpenAIConversations();
-        group.MapOpenAIResponses();
         return group;
     }
 


dotnet/src/Microsoft.Agents.AI.DevUI/DevUIExtensions.cs
```diff
@@ -34,7 +33,6 @@ public static IEndpointConventionBuilder MapDevUI(
         group.MapDevUI(pattern: "/devui");
         group.MapEntities();
         group.MapOpenAIConversations();
```
jeffhandley, line 35: Presumably, we should remove `builder.Services.AddOpenAIConversations();` and `group.MapOpenAIConversations();` too if we follow this logic; right?

Ideally `MapDevUI` would validate that its needed services and mappings were in place and throw an `InvalidOperationException` if not. But I think that could come later.

stephentoub:
There are lots of DI-related APIs that add in necessary dependencies. Why is this different?

dotnet/src/Microsoft.Agents.AI.DevUI/DevUIExtensions.cs
danroth27: Yeah, I think so. And agreed on providing a good error diagnostic.

danroth27:
> There are lots of DI-related APIs that add in necessary dependencies. Why is this different?

@stephentoub With DI it's pretty easy to make the Add methods idempotent. Unfortunately, that's harder to do with Map methods in ASP.NET Core that add endpoints.

dotnet/src/Microsoft.Agents.AI.DevUI/DevUIExtensions.cs
danroth27: I think `MapEntites()` though is more Dev UI specific.

dotnet/src/Microsoft.Agents.AI.DevUI/DevUIExtensions.cs
```diff
@@ -34,7 +33,6 @@ public static IEndpointConventionBuilder MapDevUI(
...
-        group.MapOpenAIResponses();
         return group;
```
ReubenBond, line 36: We should also be adding the bits we remove to dotnet\samples\GettingStarted\DevUI\DevUI_Step01_BasicUsage\Program.cs so that the sample still works

jeffhandley:
This should be closed in favor of [(#2014) .NET: DevUI - Do not automatically add/map OpenAI services/endpoints](https://github.com/microsoft/agent-framework/pull/2014/files)

adityamandaleeka:
Closing since #2014 is in.

PR #1996, opened 2025-11-07
.NET [AG-UI]: Adds support for shared state.

javiercn:
Adds support for shared state, enables packaging the AGUI packages and adds a sample dojo server for easy testing against the AG-UI dojo.

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
+++ b/dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
@@ -321,7 +321,7 @@ public sealed class AGUIChatClient : DelegatingChatClient
             for (int i = 0; i < lastMessage.Contents.Count; i++)
             {
                 if (lastMessage.Contents[i] is DataContent dataContent &&
-                    System.Net.Http.Headers.MediaTypeHeaderValue.TryParse(dataContent.MediaType, out var mediaType) &&
+                    MediaTypeHeaderValue.TryParse(dataContent.MediaType, out var mediaType) &&
                     mediaType.Equals(s_json))
                 {
                     // Deserialize the state JSON directly from UTF-8 bytes
--- a/dotnet/samples/AGUIClientServer/AGUIDojoServer/Ingredient.cs
+++ b/dotnet/samples/AGUIClientServer/AGUIDojoServer/Ingredient.cs
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft. All rights reserved.
+﻿// Copyright (c) Microsoft. All rights reserved.
 
 using System.Text.Json.Serialization;
 
--- a/dotnet/samples/AGUIClientServer/AGUIDojoServer/Recipe.cs
+++ b/dotnet/samples/AGUIClientServer/AGUIDojoServer/Recipe.cs
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft. All rights reserved.
+﻿// Copyright (c) Microsoft. All rights reserved.
 
 using System.Text.Json.Serialization;
 
--- a/dotnet/samples/AGUIClientServer/AGUIDojoServer/RecipeResponse.cs
+++ b/dotnet/samples/AGUIClientServer/AGUIDojoServer/RecipeResponse.cs
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft. All rights reserved.
+﻿// Copyright (c) Microsoft. All rights reserved.
 
 using System.Text.Json.Serialization;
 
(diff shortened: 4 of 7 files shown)


ReubenBond (dismissed):

dotnet/samples/AGUIClientServer/AGUIDojoServer/AGUIDojoServerSerializerContext.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Text.Json.Serialization;
+
+namespace AGUIDojoServer;
+
+[JsonSerializable(typeof(WeatherInfo))]
+internal sealed partial class AGUIDojoServerSerializerContext : JsonSerializerContext
+{
+}
```
javiercn:
```suggestion
internal sealed partial class AGUIDojoServerSerializerContext : JsonSerializerContext;
```

dotnet/src/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore/AGUIEndpointRouteBuilderExtensions.cs
```diff
@@ -28,6 +31,8 @@ public static class AGUIEndpointRouteBuilderExtensions
     /// <param name="pattern">The URL pattern for the endpoint.</param>
     /// <param name="aiAgent">The agent instance.</param>
     /// <returns>An <see cref="IEndpointConventionBuilder"/> for the mapped endpoint.</returns>
+    [RequiresDynamicCode("AG-UI endpoint mapping requires dynamic code generation for JSON serialization.")]
+    [RequiresUnreferencedCode("AG-UI endpoint mapping requires unreferenced code for JSON serialization.")]
```
javiercn:
```suggestion
```

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
```diff
@@ -300,6 +307,51 @@ public async IAsyncEnumerable<ChatResponseUpdate> GetStreamingResponseAsync(
...
+            {
+                return default;
+            }
+
+            // Check the last message for state DataContent
+            ChatMessage lastMessage = messagesList[messagesList.Count - 1];
+            for (int i = 0; i < lastMessage.Contents.Count; i++)
+            {
+                if (lastMessage.Contents[i] is DataContent dataContent &&
+                    MediaTypeHeaderValue.TryParse(dataContent.MediaType, out var mediaType) &&
```
stephentoub, line 324: Do you need to parse the MediaTypeHeaderValue? Could you instead just do a starts with or the like on the MediaType, possibly with a semicolon check after if we want to be real sticklers? We could also add a comparison for the type to DataContent (not just top level) if really desirable.

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
```diff
@@ -300,6 +307,51 @@ public async IAsyncEnumerable<ChatResponseUpdate> GetStreamingResponseAsync(
...
+                        // Remove the DataContent from the message contents
+                        lastMessage.Contents.RemoveAt(i);
+
+                        // If no contents remain, remove the entire message
+                        if (lastMessage.Contents.Count == 0)
+                        {
+                            messagesList.RemoveAt(messagesList.Count - 1);
+                        }
+
+                        return stateElement;
```
stephentoub, line 343: There will never be multiple contents to handle?

stephentoub (approved):

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
stephentoub: Meh, probably doesn't matter.

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
```diff
@@ -199,6 +202,9 @@ public async IAsyncEnumerable<ChatResponseUpdate> GetStreamingResponseAsync(
             var threadId = ExtractTemporaryThreadId(messagesList) ??
                 ExtractThreadIdFromOptions(options) ?? $"thread_{Guid.NewGuid():N}";
 
+            // Extract state from the last message if it contains DataContent with application/json
+            JsonElement state = this.ExtractAndRemoveStateFromMessages(messagesList);
```
DeagleGross, line 206: why do we only supply the last message contents, not all of those? Is that because input in AG-UI allows a **single** message passing at a time?

javiercn: Discussed offline

dotnet/src/Microsoft.Agents.AI.AGUI/AGUIChatClient.cs
javiercn: There shouldn't. The convention is to pass the state for the turn as the data content with application/json on the last message so that the client can be stateless.

PR #1984, opened 2025-11-07
Python: DevUI: Use metadata.entity_id instead of model field

ReubenBond:
### Motivation and Context

Currently, DevUI uses the 'model' field of a POST /v1/responses payload to resolve the targeted entity. This is not ideal, since 'model' should be used to specify the model. This PR changes the behavior to use the 'metadata' dictionary instead.

cc @victordibia 

### Contribution Checklist

- [x] The code builds clean without any errors or warnings
- [x] The PR follows the [Contribution Guidelines](https://github.com/microsoft/agent-framework/blob/main/CONTRIBUTING.md)
- [x] All unit tests pass, and I have added new tests where possible
- [x] **Is this a breaking change?** If yes, add "[BREAKING]" prefix to the title of the PR.

PR #2003, opened 2025-11-07
.NET: OpenAI Responses: suppress 'model' parameter if it is used for agent

ReubenBond:
### Motivation and Context

The DevUI currently flows the entity id in the 'model' field, but this field has should otherwise be flowed into the implementation. When the 'model' field is used for the entity id, we should instead suppress it from flowing further down the call chain as the model field, since consumers cannot distinguish a model parameter from an entity id.

This PR implements the above logic and enhances request validation.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/HostedAgentResponseExecutor.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/HostedAgentResponseExecutor.cs
@@ -43,7 +43,7 @@ internal sealed class HostedAgentResponseExecutor : IResponseExecutor
         CancellationToken cancellationToken = default)
     {
         // Extract agent name from agent.name or model parameter
-        string? agentName = request.Agent?.Name ?? request.Model;
+        string? agentName = GetAgentName(request);
 
         if (string.IsNullOrEmpty(agentName))
         {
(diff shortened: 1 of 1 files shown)


dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/ResponsesHttpHandler.cs
```diff
@@ -55,45 +70,20 @@ public async Task<IResult> CreateResponseAsync(
...
-                    Type = "invalid_request_error",
-                    Code = "missing_required_parameter"
-                }
-            });
+            // Return InternalServerError for unexpected exceptions
+            return Results.Problem(
+                detail: ex.Message,
+                statusCode: StatusCodes.Status500InternalServerError,
+                title: "Internal Server Error");
         }
```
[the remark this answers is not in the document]
ReubenBond: Entirely intentional.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/HostedAgentResponseExecutor.cs
```diff
@@ -75,41 +111,24 @@ public async IAsyncEnumerable<StreamingResponseEvent> ExecuteAsync(
...
-        try
-        {
-            return this._serviceProvider.GetRequiredKeyedService<AIAgent>(agentName);
-        }
-        catch (InvalidOperationException ex)
-        {
-            this._logger.LogError(ex, "Failed to resolve agent with name '{AgentName}'", agentName);
-            throw new InvalidOperationException($"Agent '{agentName}' not found. Ensure the agent is registered with AddAIAgent().", ex);
-        }
+        return request.Agent?.Name ?? request.Model;
```
stephentoub, line 120: Can you remind me why DevUI doesn't just always send it in an agent property rather than using model for that?

ReubenBond: DevUI uses the official OpenAI SDK, which doesn't have the 'agent' property. I want to introduce a change there to use the 'metadata' dictionary instead of 'model', but I want to un-block @jeffhandley etc asap

stephentoub: > DevUI uses the official OpenAI SDK, which doesn't have the 'agent' property

But it does now have JSON Patch support, so you can get/set whatever properties you want to a request.

ReubenBond: Do all of the OpenAI SDKs support setting arbitrary properties in requests & responses?

ReubenBond: I have an alternate PR here that uses metadata.entity_id instead: https://github.com/microsoft/agent-framework/pull/1984

PR #2005, opened 2025-11-07
.NET: DevUI: Use relative URLs for backend API by default

ReubenBond:
### Motivation and Context

Currently, we need to customize how DevUI is built when we're consuming it for .NET because we frequently are not listening on port 8080 and the current config assumes port 8080. This PR changes the default to "" so that it DevUI assumes the host that's serving the assets is also the backend by default. This lets us remove the special casing

### Contribution Checklist


markwallace-microsoft:
<a href="https://github.com/microsoft/agent-framework/blob/13f1b4c153378ad5964a03da9af982e3bbbbea84/README.md">Python Test Coverage</a><details><summary>Python Test Coverage Report • </summary><table><tr><th>File</th><th>Stmts</th><th>Miss</th><th>Cover</th><th>Missing</th></tr><tbody><tr><td><b>TOTAL</b></td><td><b>13035</b></td><td><b>1939</b></td><td><b>85%</b></td><td>&nbsp;</td></tr></tbody></table><i>report-only-changed-files is enabled. No files were changed during this commit :)</i></details>

## Python Unit Test Overview
| 1740 | 107 :zzz: | 0 :x: | 0 :fire: | 33.712s :stopwatch: |

DeagleGross (approved):

python/packages/devui/agent_framework_devui/ui/assets/index.js
```diff
@@ -86,7 +86,7 @@ Error generating stack: `+a.message+`
...
...((e,n)=>{const{__scopeDropdownMenu:o,...s}=e,l=Pt(o);return i.jsx(xk,{...l,...s,ref:n})});Lk.displayName=zk;var Ik="DropdownMenuSeparator",g1=S.forwardRef((e,n)=>{const{__scopeDropdownMenu:o,...s}=e,l=Pt(o);return i.jsx(vk,{...l,...s,ref:n})});g1.displayName=Ik;var Hk="DropdownMenuArrow",Bk=S.forwardRef((e,n)=>{const{__scopeDropdownMenu:o,...s}=e,l=Pt(o);return i.jsx(yk,{...l,...s,ref:n})});Bk.displayName=Hk;var Uk="DropdownMenuSubTrigger",Pk=S.forwardRef((e,n)=>{const{__scopeDropdownMenu:o,...s}=e,l=Pt(o);return i.jsx(bk,{...l,...s,ref:n})});Pk.displayName=Uk;var $k="DropdownMenuSubContent",Vk=S.forwardRef((e,n)=>{const{__scopeDropdownMenu:o,...s}=e,l=Pt(o);return i.jsx(wk,{...l,...s,ref:n,style:{...e.style,"--radix-dropdown-menu-content-transform-origin":"var(--radix-popper-transform-origin)","--radix-dropdown-menu-content-available-width":"var(--radix-popper-available-width)","--radix-dropdown-menu-content-available-height":"var(--radix-popper-available-height)","--radix-dropdown-menu-trigger-width":"var(--radix-popper-anchor-width)","--radix-dropdown-menu-trigger-height":"var(--radix-popper-anchor-height)"}})});Vk.displayName=$k;var qk=l1,Yk=u1,Gk=d1,Xk=m1,Fk=h1,Zk=p1,Kk=g1;/**
```
stephentoub, line 89: I assume I don't need to review this gobbledegook :)

stephentoub (approved):

python/packages/devui/frontend/src/components/layout/settings-modal.tsx
```diff
@@ -27,7 +27,7 @@ export function SettingsModal({ open, onOpenChange, onBackendUrlChange }: Settin
   const [activeTab, setActiveTab] = useState<Tab>("settings");
 
   // Get current backend URL from localStorage or default
-  const defaultUrl = import.meta.env.VITE_API_BASE_URL || "http://localhost:8080";
+  const defaultUrl = import.meta.env.VITE_API_BASE_URL !== undefined ? import.meta.env.VITE_API_BASE_URL : "";
```
victordibia, line 30: Why dont we set this via a .env var using the [guidance here](https://github.com/microsoft/agent-framework/blob/main/python/packages/devui/frontend/README.md)?
The goal of env vars is to that the build system can configure these?

```
cd frontend
yarn install

# Create .env.local with backend URL
echo 'VITE_API_BASE_URL=http://localhost:8000' > .env.local

# Create .env.production (empty for relative URLs)
echo '' > .env.production

# Development
yarn dev

# Build (copies to backend)
yarn build
```

victordibia: For context, this is what my .env.production looks like

```
# Production Environment
# Used when NODE_ENV=production (for 'yarn build')

# In production, you'll typically override this with actual deployment URL
VITE_API_BASE_URL=""
```

and my .env.local looks like 

```
# Local Development Environment
# This file is gitignored - customize for your local setup

# Backend API URL - adjust to match your backend server
VITE_API_BASE_URL=http://localhost:8080
```

What if the .net build system set these before building the UI? 
It also provides a foundation for setting other vars as we build out devui

victordibia (changes requested):

python/packages/devui/frontend/src/components/layout/settings-modal.tsx
ReubenBond: I think being able to customize it in the build is fine, but the default being that it connects to the host which serves the assets is probably a better default, since it's likely more common to host the assets and API on the same server.

python/packages/devui/agent_framework_devui/ui/assets/index.js
ReubenBond: Yeah, ideally we would build this automatically in a central place during the build and not check it in at all, but we don't have that automation yet.

python/packages/devui/frontend/src/components/layout/settings-modal.tsx
ReubenBond: To be clear, by using "" as the default value here, you can host it on whatever port you want without rebuilding the DevUI or adding any extra config. I ran `devui` and it just works. Changing the port it binds to via the cmd line arg also just works.

I don't see a downside to this PR - do you?

victordibia: Fair enough. 
The choice here is to optimize for the common prod build case, seems fair.
I also wanted to highlight that the .env files is the way in which this is set today (I probably should make this clearer in the devui readme)

PR #2024, opened 2025-11-08
Python: DevUI: Prevent line breaks within words in the agent view

### LINKED ISSUE #1768: Python: DevUI: Agent responses have line breaks at arbitrary characters instead of at word boundaries
As seen here:

Image

This seems to be caused by the `break-all` class that is applied to the div that wraps around the agent responses:

https://github.com/microsoft/agent-framework/blob/1fbdcf82689effe2abff865dd4f31174e4868729/python/packages/devui/frontend/src/components/features/agent/agent-view.tsx#L118-L139

Within the agent responses, `break-word` is applied, but the Tailwind classes for `break-all` and `break-word` don't cascade as I would have thought, and this has the result of `break-all` applying to the content. I do not know if there will be unexpected consequences from simply removing `break-all` from the container div though.

jeffhandley:
### Motivation and Context

Fixes #1768 - Python: DevUI: Agent responses have line breaks at arbitrary characters instead of at word boundaries

### Description

The container DIV that agent messages are rendered into has the [`break-all`](https://tailscan.com/tailwind/typography/break-all) class applied to it, [which means](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/word-break#break-all) "word breaks should be inserted between any two characters." This is then inherited by all child elements. The child element containing the actual agent message text has the [`break-words`](https://tailscan.com/tailwind/typography/break-words) class applied, but that is implemented by a different CSS property than `break-all`, and it does not override the `break-all` behavior.

The end result for the CSS is the following, which still has the effect of breaking between any two characters.

```css
word-break: break-all;
overflow-wrap: break-word;
```

This PR removes `break-all` from the container element for successful agent messages, but it leaves that class on in unsuccessful containers. Error messages are more prone to have very long strings that need to have arbitrary breaks in them--and having arbitrary breaks in failure messages is acceptable while arbitrary breaks in text responses is not a good experience.

### Contribution Checklist

- [x] **Is this a breaking change?** If yes, add "[BREAKING]" prefix to the title of the PR. (No, it is not)

Before:
Image

After:
image


markwallace-microsoft:
<a href="https://github.com/microsoft/agent-framework/blob/d0346b2ce5533891ac78d85837dcb63a3133334a/README.md">Python Test Coverage</a><details><summary>Python Test Coverage Report • </summary><table><tr><th>File</th><th>Stmts</th><th>Miss</th><th>Cover</th><th>Missing</th></tr><tbody><tr><td><b>TOTAL</b></td><td><b>13035</b></td><td><b>1939</b></td><td><b>85%</b></td><td>&nbsp;</td></tr></tbody></table><i>report-only-changed-files is enabled. No files were changed during this commit :)</i></details>

## Python Unit Test Overview
| 1740 | 107 :zzz: | 0 :x: | 0 :fire: | 35.110s :stopwatch: |

javiercn (approved):
LGTM

victordibia (approved):

jeffhandley:
Thanks for the reviews, @javiercn and @victordibia. I do not have permission to merge, so I leave it to you all to merge if you think it's ready.

Issue #25, opened 2025-05-28
Add declarative agent models and factory abstractions


markwallace-microsoft:
- [x] Add support to load any type of agent from a YAML declarative representation
- [x] Create samples that show various permutations, including tool configuration
- [x] Use PowerFX to evaluate values
- [ ] Add support for all Foundry Agent tool types
- [ ] Complete all TODO's
- [ ] 80% unit test coverage and integration tests
- [ ] Use code generation [dotnet/src/Microsoft.Agents.AI.Declarative/Extensions/PropertyInfoExtensions.cs](https://github.com/microsoft/agent-framework/pull/1301/files/0b522dce698ecbbbc20711916aa37b6f2c9ab6c8#diff-a59f6b56808dd1ce3f7c2baba8cf60fbced9b5df8fea74a1884500bb00d8f181)
- [ ] Add documentation to the sample templates
- [ ] Add template samples showing all connection types
- [ ] Use Dictionary in place of IConfiguration
- [ ] Move XXXExpressionExtensions into internal helpers

matthewmeyer:
I would love to see this feature, the AutoGen Python component model and [seralization and deseralization support ](https://microsoft.github.io/autogen/stable/user-guide/agentchat-user-guide/serialize-components.html) are a really powerful tool and could help drive parity between the language specific implementations.

eavanvalkenburg:
@markwallace-microsoft is there a equivalent task of this for python?

droideronline:
+1 for this feature — would be super useful!

amit12cool:
@markwallace-microsoft  any plans to add memory//context provider support in declarative agents?

PR #2093, opened 2025-11-11
.NET: Integrate RecalcEngine to compute Env variables

markwallace-microsoft:
### Contribution Checklist

### DIFF
--- a/dotnet/samples/GettingStarted/DeclarativeAgents/OpenAI/DeclarativeOpenAIAgents.csproj
+++ b/dotnet/samples/GettingStarted/DeclarativeAgents/OpenAI/DeclarativeOpenAIAgents.csproj
@@ -11,7 +11,6 @@
   <ItemGroup>
     <PackageReference Include="Azure.Identity" />
     <PackageReference Include="Azure.AI.Agents.Persistent" />
-    <PackageReference Include="Microsoft.Extensions.Configuration.EnvironmentVariables" />
     <PackageReference Include="Microsoft.Extensions.DependencyInjection" />
     <PackageReference Include="Microsoft.Bot.ObjectModel" />
     <PackageReference Include="Microsoft.Bot.ObjectModel.Json" />


dotnet/samples/GettingStarted/DeclarativeAgents/OpenAI/DeclarativeOpenAIAgents.csproj
```diff
@@ -11,6 +11,7 @@
   <ItemGroup>
     <PackageReference Include="Azure.Identity" />
     <PackageReference Include="Azure.AI.Agents.Persistent" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.EnvironmentVariables" />
```
westey-m: Are you using this?  It seems like you are just adding the settings to an inmemory collection instead of reading them from environment variables via the config system.

agent-samples/azure/AzureOpenAIAssistants.yaml
rogerbarreto, line 1: This can be confusing for people that are looking for samples and will first look into the root/agent-samples and see those as yaml files. 

Suggest moving this to `docs/prompt-agent-definitions` to avoid any misleading sample discovery

rogerbarreto (approved):

dotnet/src/Microsoft.Agents.AI.Declarative/AgentFactory.cs
```diff
@@ -13,6 +15,29 @@ namespace Microsoft.Agents.AI;
...
+    /// Initializes a new instance of the <see cref="AgentFactory"/> class.
+    /// </summary>
+    /// <param name="configuration">The configuration.</param>
+    protected AgentFactory(IConfiguration? configuration = null)
+    {
+        this.Engine = new RecalcEngine();
+
+        if (configuration is not null)
+        {
+            foreach (var kvp in configuration.AsEnumerable())
```
westey-m, line 28: Do we need IConfiguration?  It seems like we are actually just using a dictionary.  Users could get a dictionary from wherever they want, including config if they prefer.

dotnet/src/Microsoft.Agents.AI.Declarative/Extensions/StringExpressionExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+
+using Microsoft.PowerFx;
+using Microsoft.PowerFx.Types;
+
+namespace Microsoft.Bot.ObjectModel;
+
+/// <summary>
+/// Extension methods for <see cref="StringExpression"/>.
+/// </summary>
+public static class StringExpressionExtensions
```
SergeyMenshykh, line 11: Does it need to be public?

westey-m, line 11:
```suggestion
internal static class StringExpressionExtensions
```

westey-m (approved):

dotnet/src/Microsoft.Agents.AI.Declarative/Extensions/StringExpressionExtensions.cs
markwallace-microsoft: Yes, this will be used in both declarative packages to handle expressions and variable references. In a follow up PR there will be much more use of this code.

markwallace-microsoft: Keeping this public as this will be used in both declarative packages to handle expressions and variable references. In a follow up PR there will be much more use of this code.

dotnet/src/Microsoft.Agents.AI.Declarative/AgentFactory.cs
markwallace-microsoft: I thought using IConfiguration would be more idiomatic as that is where these values would typically be set. I don't have a strong opinion on switching to a Dictionary, I will add this to issue https://github.com/microsoft/agent-framework/issues/25

agent-samples/azure/AzureOpenAIAssistants.yaml
markwallace-microsoft: We already have the workflow samples at the root

image

If we are moving the these I think we should move both. The `docs` folder doesn't seem like the right place but maybe `templates\workflows` and `templates\agents`. I will sync with Chris to make this change.

PR #2118, opened 2025-11-12
.NET: [Feature Branch] Update HTTP API to be consistent across languages

cgillum:
### Motivation and Context

Azure Functions built-in HTTP APIs between .NET and Python were inconsistent. They need to be made consistent so that documentation can be more easily created explaining how to interact with the HTTP APIs, regardless of what SDK is being used.

### Description

The following changes were made:

- The agent run HTTP API now accepts text or JSON
- The agent run HTTP API returns JSON in a particular format
- The `X-Agent-Thread` response header was renamed to `x-ms-thread-id`
- The `threadId` query string parameter was renamed to `thread_id`

Note that this PR isn't intended to create full parity. Rather, it's meant to address the highest priority gaps. Full parity will be attempted later.

FYI @anthonychu

### Contribution Checklist

### DIFF
--- a/dotnet/samples/AzureFunctions/01_SingleAgent/README.md
+++ b/dotnet/samples/AzureFunctions/01_SingleAgent/README.md
@@ -65,6 +65,25 @@ The expected `application/json` output will look something like:
 {
   "status": 200,
   "thread_id": "@dafx-joker@your-thread-id",
-  "response": "Why don't pirates ever learn the alphabet? Because they always get stuck at 'C'!"
+  "response": {
+    "Messages": [
+      {
+        "AuthorName": "Joker",
+        "CreatedAt": "2025-11-11T12:00:00.0000000Z",
+        "Role": "assistant",
+        "Contents": [
+          {
+            "Type": "text",
+            "Text": "Why don't pirates ever learn the alphabet? Because they always get stuck at 'C'!"
+          }
+        ]
+      }
+    ],
+    "Usage": {
+      "InputTokenCount": 78,
+      "OutputTokenCount": 36,
+      "TotalTokenCount": 114
+    }
+  }
 }
 ```
--- a/dotnet/samples/AzureFunctions/01_SingleAgent/README.md
+++ b/dotnet/samples/AzureFunctions/01_SingleAgent/README.md
@@ -35,8 +35,36 @@ Invoke-RestMethod -Method Post `
     -Body "Tell me a joke about a pirate."
 ```
 
-The response from the agent will be displayed in the terminal where you ran `func start`. The expected output will look something like:
+You can also send JSON requests:
+
+```bash
+curl -X POST http://localhost:7071/api/agents/Joker/run \
+    -H "Content-Type: application/json" \
+    -H "Accept: application/json" \
+    -d '{"message": "Tell me a joke about a pirate."}'
+```
+
+To continue a conversation, include the `thread_id` in the query string or JSON body:
+
+```bash
+curl -X POST "http://localhost:7071/api/agents/Joker/run?thread_id=@dafx-joker@your-thread-id" \
+    -H "Content-Type: application/json" \
+    -H "Accept: application/json" \
+    -d '{"message": "Tell me another one."}'
+```
+
+The response from the agent will be displayed in the terminal where you ran `func start`. The expected `text/plain` output will look something like:
 
 ```text
 Why don't pirates ever learn the alphabet? Because they always get stuck at "C"!
+```
+
+The expected `application/json` output will look something like:
+
+```json
+{
+  "status": 200,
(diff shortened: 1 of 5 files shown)


dotnet/src/Microsoft.Agents.AI.Hosting.AzureFunctions/BuiltInFunctions.cs
```diff
@@ -45,55 +46,101 @@ public static async Task<HttpResponseData> RunAgentHttpAsync(
...
+        // Validate that if thread_id is specified in both places, they must match
+        if (!string.IsNullOrEmpty(threadIdFromQuery) && !string.IsNullOrEmpty(threadIdFromBody) &&
+            !string.Equals(threadIdFromQuery, threadIdFromBody, StringComparison.Ordinal))
+        {
+            return await CreateErrorResponseAsync(
+                req,
+                context,
+                HttpStatusCode.BadRequest,
+                "thread_id specified in both query string and request body must match.");
+        }
```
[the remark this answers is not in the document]
cgillum: I don't understand - it _does_ end with a period already.

dotnet/samples/AzureFunctions/01_SingleAgent/README.md
```diff
@@ -35,8 +35,36 @@ Invoke-RestMethod -Method Post `
...
 Why don't pirates ever learn the alphabet? Because they always get stuck at "C"!
 ```
+
+The expected `application/json` output will look something like:
+
+```json
+{
+  "status": 200,
+  "thread_id": "@dafx-joker@your-thread-id",
+  "response": "Why don't pirates ever learn the alphabet? Because they always get stuck at 'C'!"
```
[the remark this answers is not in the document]
cgillum: Fixed

dotnet/src/Microsoft.Agents.AI.Hosting.AzureFunctions/BuiltInFunctions.cs
```diff
@@ -144,6 +291,44 @@ private static string GetAgentName(FunctionContext context)
...
+        [property: JsonPropertyName("thread_id")] string? ThreadId);
+
+    /// <summary>
+    /// Represents an error response.
+    /// </summary>
+    /// <param name="Status">The HTTP status code.</param>
+    /// <param name="Error">The error message.</param>
+    private sealed record ErrorResponse(
+        [property: JsonPropertyName("status")] int Status,
+        [property: JsonPropertyName("error")] string Error);
```
kshyju, line 310: Should this also include the "thread_id" property? I am wondering whether there is a use case where client is sending multiple requests (from different threads) and want to understand which one failed and act accordingly.

kshyju: `thread_id` or some unique id from the request, which can tell, what request this response is for.

dotnet/src/Microsoft.Agents.AI.Hosting.AzureFunctions/BuiltInFunctions.cs
```diff
@@ -45,55 +46,101 @@ public static async Task<HttpResponseData> RunAgentHttpAsync(
...
+                threadIdFromBody = requestBody.ThreadId;
+            }
+        }
+        else
+        {
+            // Plain text body
+            message = await req.ReadAsStringAsync();
+        }
+
+        // The thread ID can come from query string or JSON body
```
larohra, line 70: NIT: Want to add support for header `x-ms-thread-id` too?

larohra (approved):
one minor comment else LGTM

dotnet/src/Microsoft.Agents.AI.Hosting.AzureFunctions/BuiltInFunctions.cs
cgillum: I think OTel should be enough to help with the need for correlating error responses to requests. The thread_id value is mostly needed in cases where the thread_id is server-generated. But in these cases, the problem is that the input isn't valid, so there is no conversation to associate the error with.

dotnet/src/Microsoft.Agents.AI.Hosting.AzureFunctions/BuiltInFunctions.cs
cgillum: This would mean there are three different ways to specify the thread ID (JSON body, query string parameter, and header), which complicates request validation. If we support these, we need to come up with a set of rules for how to validate these requests to ensure consistency or decide which thread ID should be the "winner". I'd like to defer that for now since it will make the code logic more complicated and I don't want to spend time writing more tests to cover these other use cases.

PR #2120, opened 2025-11-12
Python: Fix chat middleware: add streaming support, terminate flag, and check…

t-anjan:
## Summary

This PR fixes three critical issues in the `security_filter_middleware` example in `python/samples/getting_started/devui/weather_agent_azure/agent.py` that prevent it from working correctly in streaming mode and cause unexpected behavior.

## Motivation and Context

The current sample code for chat middleware has several issues that make it non-functional in real-world scenarios:

1. **Missing termination flag**: Without `context.terminate = True`, the middleware doesn't properly stop execution after blocking a request, potentially allowing the LLM to still be called
2. **No streaming support**: Most production agents run in streaming mode, but the middleware only handles non-streaming responses, causing blank responses when blocking sensitive terms
3. **Inefficient message checking**: The middleware iterates through all conversation history instead of checking only the latest user message, which is inefficient and could trigger false positives on historical messages

This fix ensures the sample code works correctly as a reference implementation for developers learning the Agent Framework.

### Changes Made

**Imports:**
- Added `AsyncIterable` from `collections.abc` for streaming support
- Added `ChatResponseUpdate` and `TextContent` from `agent_framework` for streaming responses

**security_filter_middleware function:**
- **Streaming support**: Added `context.is_streaming` check to handle both modes:
  - Streaming: Returns async generator yielding `ChatResponseUpdate` with `TextContent`
  - Non-streaming: Returns `ChatResponse` with `ChatMessage`
- **Proper termination**: Added `context.terminate = True` to stop middleware chain execution
- **Optimized message checking**: Changed from iterating all messages to checking only `context.messages[-1]` with `Role.USER` filter

### Technical Details

The fix aligns with the [Agent Middleware documentation](https://learn.microsoft.com/en-us/agent-framework/user-guide/agents/agent-middleware?pivots=programming-language-python) "Middleware Result Override" pattern, which demonstrates handling both streaming and non-streaming scenarios by checking `context.is_streaming`.

### Testing

Manually tested with both streaming and non-streaming agent runs. The middleware now correctly:
- Blocks requests containing sensitive terms ("password", "secret", "api_key", "token")
- Displays the error message properly in both streaming and non-streaming modes
- Terminates execution without calling the LLM
- Only checks the most recent user message

## Contribution Checklist

- [x] **Is this a breaking change?** No - this is a bug fix to sample code that improves functionality without changing the API


t-anjan:

victordibia (approved):
LGTM, thanks

moonbox3 (approved):

PR #2147, opened 2025-11-12
.NET: Feature foundry agent/agent reference extension (Python Parity with Name + Version option)

rogerbarreto:
### Motivation and Context

Bring feature parity missing in .NET related to the Python option to provide `Name` + `Version` without requiring a round trip to the server side.

This update adds the option to retrieve an agent just providing the `AgentReference` which don't require a request to the `server` to create a `AgentVersion` or `AgentRecord`.

This can be ideal for scenarios where the caller just want to use the agent based in its `Id` composite of `Name:VersionNumber`.


alliscode (approved):

rogerbarreto:
@crickman 
image

PR #2214, opened 2025-11-14
.NET: Friendly error message when durable agent isn't registered

cgillum:
### Motivation and Context

From a durable orchestration, developers must call `context.GetAgent("MyAgent")` to get a `DurableAIAgent` instance. However, we don't have a way to validate whether "MyAgent" is an actual registered agent since `context.GetAgent` doesn't have access to the DI context.  If the code then calls `DurableAIAgent.RunAsync(...)`, the call will fail with an "EntityTaskNotFound" exception, which is confusing.

### Description

The fix is to catch the `EntityTaskNotFoundException` and instead throw a more helpful `AgentNotRegisteredException`.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.DurableTask/DurableAIAgent.cs
+++ b/dotnet/src/Microsoft.Agents.AI.DurableTask/DurableAIAgent.cs
@@ -5,6 +5,7 @@ using System.Runtime.CompilerServices;
 using System.Text.Json;
 using System.Text.Json.Serialization.Metadata;
 using Microsoft.DurableTask;
+using Microsoft.DurableTask.Entities;
 using Microsoft.Extensions.AI;
 
 namespace Microsoft.Agents.AI.DurableTask;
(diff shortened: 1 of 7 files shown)


larohra (approved):

dotnet/src/Microsoft.Agents.AI.DurableTask/DurableAIAgent.cs
```diff
@@ -95,7 +99,17 @@ public override async Task<AgentRunResponse> RunAsync(
...
+        try
+        {
+            return await this._context.Entities.CallEntityAsync<AgentRunResponse>(
+                durableThread.SessionId,
+                nameof(AgentEntity.RunAgentAsync),
+                request);
+        }
+        catch (EntityOperationFailedException e) when (e.FailureDetails.ErrorType == "EntityTaskNotFound")
+        {
+            throw new AgentNotRegisteredException(this._agentName, e);
```
kshyju, line 111: Apart from a missing agent registration, are there any other reasons why CallEntityAsync might throw the `EntityOperationFailedException ` exception with that ErrorType value "EntityTaskNotFound"?

kshyju (approved):

dotnet/src/Microsoft.Agents.AI.DurableTask/DurableAIAgent.cs
cgillum: This should be the only reason.

PR #2250, opened 2025-11-15
Python: Use AI Foundry evaluators for self-reflection

YusakuNo1:
### Description

Demostrate how to implement self-reflection from agent framework

### Contribution Checklist


eavanvalkenburg (commented):
Love the idea of this, but could you have a look at our ChatMiddleware and implement using that, that way it becomes a much more native way of working, and if you put the actual middleware function inside of a class then all the setup and best_score etc. can be captured and maintained, this samples shows classes, the same works for `ChatMiddleware`, but this could also be implemented with AgentMiddleware: https://github.com/microsoft/agent-framework/blob/main/python/samples/getting_started/middleware/class_based_middleware.py

YusakuNo1:
> Love the idea of this, but could you have a look at our ChatMiddleware and implement using that, that way it becomes a much more native way of working, and if you put the actual middleware function inside of a class then all the setup and best_score etc. can be captured and maintained, this samples shows classes, the same works for `ChatMiddleware`, but this could also be implemented with AgentMiddleware: https://github.com/microsoft/agent-framework/blob/main/python/samples/getting_started/middleware/class_based_middleware.py

Hi @eavanvalkenburg , thanks for the suggestion! This approach is not added into middleware because it's slow and consumes a lot of tokens, we would like the AI developers to try it as "optional" instead of in the core path. What do you think?

YusakuNo1:
> Love the idea of this, but could you have a look at our ChatMiddleware and implement using that, that way it becomes a much more native way of working, and if you put the actual middleware function inside of a class then all the setup and best_score etc. can be captured and maintained, this samples shows classes, the same works for `ChatMiddleware`, but this could also be implemented with AgentMiddleware: https://github.com/microsoft/agent-framework/blob/main/python/samples/getting_started/middleware/class_based_middleware.py

Within the team, we discussed offline for the approach of using middeware. For the use case of middleware, it'll intercept the traffic and do something extra, but for our current use case, the self-reflection code will also modify the original user input, and then run again. With this, maybe it's not a good use case for middleware, plus, this is observability use case, maybe we can make it simple for the user for now. What do you think?

python/samples/getting_started/evaluation/azure_ai_foundry/evaluation/self_reflection.py
```diff
@@ -0,0 +1,392 @@
...
+    Create a groundedness evaluator.
+
+    Args:
+        judge_model: Model deployment name for evaluation
+    Returns:
+        Configured GroundednessEvaluator
+    """
+    judge_model_config = AzureOpenAIModelConfiguration(
+        azure_endpoint=os.environ.get("AZURE_OPENAI_ENDPOINT"),
+        api_key=os.environ.get("AZURE_OPENAI_API_KEY"),
```
TaoChenOSU, line 53: Does it need `dotenv` to load the variables from a .env file?

python/samples/getting_started/observability/self_reflection.py
```diff
@@ -0,0 +1,392 @@
...
+    judge_model_config = AzureOpenAIModelConfiguration(
+        azure_endpoint=os.environ.get("AZURE_OPENAI_ENDPOINT"),
+        api_key=os.environ.get("AZURE_OPENAI_API_KEY"),
+        api_version="2024-12-01-preview",
+        azure_deployment=judge_model,
+    )
+    return GroundednessEvaluator(model_config=judge_model_config)
+
+
+async def do_query_with_self_reflection(
```
TaoChenOSU: nit:

```suggestion
async def execute_query_with_self_reflection(
```

python/samples/getting_started/evaluation/azure_ai_foundry/evaluation/self_reflection.py
```diff
@@ -0,0 +1,392 @@
...
+    best_score = 0
+    max_score = 5
+    best_response = None
+    best_iteration = 0
+    raw_response = None
+    total_groundedness_eval_time = 0.0
+    start_time = time.time()
+    iteration_scores = []  # Store all iteration scores in structured format
+
+    for i in range(max_self_reflections):
```
TaoChenOSU, line 101: Why not do `range(1, max_self_reflections+1)` instead of doing `i+1` in many other places?

python/samples/getting_started/evaluation/azure_ai_foundry/evaluation/self_reflection.py
```diff
@@ -0,0 +1,392 @@
+"""
+Self-Reflection LLM Runner
+
+This module implements a self-reflection loop for LLM responses using groundedness evaluation.
```
TaoChenOSU, line 9: Could we add more context on what self-reflection is? Maybe references to some reading materials?

YusakuNo1:
> I probably wouldn't put this sample in the observability samples because it does not have anything to do with observability.
> 
> Maybe a new `evaluation` folder?

For organization definition, our "observability" includes both tracing and evaluation... I think we can host the sample  in observability for now?

python/samples/getting_started/evaluation/azure_ai_foundry/evaluation/self_reflection.py
YusakuNo1: Yes, `load_dotenv` is in line 205

python/samples/getting_started/observability/self_reflection.py
YusakuNo1: Thanks, updated

python/samples/getting_started/evaluation/azure_ai_foundry/evaluation/self_reflection.py
YusakuNo1: As "start from 0" is a convention for software engineers, for this line we can start from 0, but in the future if any people uses the same "i", they may get into trouble... Maybe it's good to stay with the convention?

python/samples/getting_started/evaluation/azure_ai_foundry/evaluation/self_reflection.py
YusakuNo1: I heard sefl-reflection pattern is a bit famous from the research group but at early stage, also there are no "common definition" for it yet. This PR demonstrates some early attempt for it, but the definition is not super clear...

TaoChenOSU: No one will be able to maintain this sample except you if we don't know what to expect. 

It doesn't need to be a common definition but just what it means in this sample would be enough.

TaoChenOSU:
> > I probably wouldn't put this sample in the observability samples because it does not have anything to do with observability.
> > Maybe a new `evaluation` folder?
> 
> For organization definition, our "observability" includes both tracing and evaluation... I think we can host the sample in observability for now?

What do you think @eavanvalkenburg? If we want to keep evaluation under observability, could we change the name of the sample to `evaluation_with_self_reflection`?

python/samples/getting_started/evaluation/azure_ai_foundry/evaluation/self_reflection.py
YusakuNo1: Added citation from our data scientists

python/samples/getting_started/evaluation/azure_ai_foundry/evaluation/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Prerequisites
+
+### Azure Resources
+- **Azure OpenAI**: Deploy models (default: gpt-4.1 for both agent and judge)
+- **Azure CLI**: Run `az login` to authenticate
+
+### Python Environment
+```bash
+pip install agent-framework azure-ai-evaluation azure-identity pandas python-dotenv
```
eavanvalkenburg: azure-identity is a dependency of AF, so not needed, and I would only call for installing core, since that covers the use case:

```suggestion
pip install agent-framework-core azure-ai-evaluation pandas --pre
```

python/samples/getting_started/evaluation/azure_ai_foundry/evaluation/self_reflection.py
```diff
@@ -0,0 +1,397 @@
...
+    if env_file and os.path.exists(env_file):
+        load_dotenv(env_file, override=True)
+    else:
+        load_dotenv(override=True)
+
+    # Create agent
+    agent = AzureOpenAIChatClient(
+        credential=AzureCliCredential(),
+        deployment_name=agent_model,
+        api_key=os.environ.get("AZURE_OPENAI_API_KEY"),
```
eavanvalkenburg: this is loaded automatically

python/samples/getting_started/evaluation/azure_ai_foundry/evaluation/self_reflection.py
```diff
@@ -0,0 +1,397 @@
...
+        endpoint=os.environ.get("AZURE_OPENAI_ENDPOINT"),
+    ).create_agent(
+        instructions="You are a helpful agent.",
+    )
+
+    # Load input data
+    print(f"Loading prompts from: {input_file}")
+    df = pd.read_parquet(input_file)
```
eavanvalkenburg: could we make the sample resource file something readable like a csv, or json? That way you don't have to run the code to understand what's in there, if it's too large, then please add a snippet as CSV in the readme

python/samples/getting_started/evaluation/azure_ai_foundry/evaluation/self_reflection.py
```diff
@@ -0,0 +1,397 @@
...
+
+    for i in range(max_self_reflections):
+        print(f"  Self-reflection iteration {i+1}/{max_self_reflections}...")
+        
+        raw_response = await agent.run(messages=messages)
+        agent_response = raw_response.text
+
+        # Evaluate groundedness
+        start_time_eval = time.time()
+        groundedness_res = evaluator(
```
eavanvalkenburg, line 109: is there no async version of this?

python/samples/getting_started/evaluation/azure_ai_foundry/evaluation/self_reflection.py
```diff
@@ -0,0 +1,397 @@
...
+            if iterations:
+                avg_iteration = sum(iterations) / len(iterations)
+                first_try = sum(1 for it in iterations if it == 1)
+                print(f"\nIteration Statistics:")
+                print(f"  Average best iteration: {avg_iteration:.2f}")
+                print(f"  Best on first try: {first_try}/{len(iterations)} ({100*first_try/len(iterations):.1f}%)")
+
+    print("="*60)
+
+    return results_df
```
eavanvalkenburg: why do we return this if that isn't handled?

python/samples/getting_started/evaluation/azure_ai_foundry/evaluation/README.md
changliu2, line 1: @YusakuNo1 , this folder name is redundant - can we do evaluation/azure_ai_foundry/self_reflection/?

python/samples/getting_started/evaluation/azure_ai_foundry/evaluation/self_reflection.py
YusakuNo1: Thanks! I removed parameters `api_key` and `endpoint`, and added comments that AzureOpenAIChatClient will load both `AZURE_OPENAI_API_KEY` and `AZURE_OPENAI_ENDPOINT` automatically.

python/samples/getting_started/evaluation/azure_ai_foundry/evaluation/self_reflection.py
YusakuNo1: Sure, updated to JSONL

python/samples/getting_started/evaluation/azure_ai_foundry/evaluation/README.md
YusakuNo1: Yes, moved

python/samples/getting_started/evaluation/azure_ai_foundry/evaluation/self_reflection.py
YusakuNo1: This is the core processing logic which we will keep it sync and simple. In the near future, this logic will be even not user facing but only executing within our server.

python/samples/getting_started/evaluation/azure_ai_foundry/evaluation/self_reflection.py
YusakuNo1: This value is not used in this sample code but for "using run_self_reflection_batch as a library", please refer to the top section comments of "Usage as library"

python/samples/getting_started/evaluation/azure_ai_foundry/evaluation/self_reflection.py
```diff
@@ -0,0 +1,382 @@
...
+Reflexion: language agents with verbal reinforcement learning.
+Noah Shinn, Federico Cassano, Ashwin Gopinath, Karthik Narasimhan, and Shunyu Yao. 2023.
+In Proceedings of the 37th International Conference on Neural Information Processing Systems (NIPS '23). Curran Associates Inc., Red Hook, NY, USA, Article 377, 8634–8652.
+https://arxiv.org/abs/2303.11366 
+
+This module implements a self-reflection loop for LLM responses using groundedness evaluation.
+It loads prompts from a JSONL file, runs them through an LLM with self-reflection,
+and saves the results.
+
+Can be used as a library or as a standalone CLI tool.
```
eavanvalkenburg:
```suggestion
```

python/samples/getting_started/evaluation/azure_ai_foundry/evaluation/self_reflection.py
```diff
@@ -0,0 +1,382 @@
+"""
```
eavanvalkenburg, line 1: we need the microsoft copyright header

python/samples/getting_started/evaluation/azure_ai_foundry/evaluation/self_reflection.py
```diff
@@ -0,0 +1,382 @@
...
+Self-Reflection LLM Runner
```
eavanvalkenburg, line 2: and in most samples we have the comment explaining it below the imports

python/samples/getting_started/evaluation/azure_ai_foundry/evaluation/self_reflection.py
YusakuNo1: Thanks, updated!

python/samples/getting_started/evaluation/azure_ai_foundry/evaluation/self_reflection.py
YusakuNo1: I think it's better to put the comment right under the copyright, but updated for now to make it consistent with the other samples

PR #2260, opened 2025-11-17
.NET: [BREAKING] Durable agents - change thread_id from entity ID to GUID

### LINKED ISSUE #2236: .NET: Thread ID values returned by durable agents are different between .NET and Python
### Context

The Azure Functions integration for Agent Framework allows agents to be accessed over HTTP. Thread IDs are passed back and forth using `x-ms-thread-id` headers or `thread_id` JSON fields.

### Issue description

The format of these thread IDs are different between .NET and Python implementations. For example, when interacting with an agent named "joker":

| SDK | Example Thread ID |
| - | - |
| .NET | `@dafx-joker@46cd2876922345978ebcf1b170db2aa2` |
| Python | `46cd2876922345978ebcf1b170db2aa2` |

Basically, .NET is using the underlying entity ID as the thread ID whereas Python is using the entity key.

### Expected

The format should be the same, regardless of what SDK is being used. This is important for things like tooling and documentation.

As for which format to use, I'm leaning towards the Python format because including the entity name is already redundant when the URLs that accept these values already have the agent name (e.g., `/api/agents/joker/run?thread_id=xxx`). It's also probably better to expose fewer internal implementation details.

cgillum:
### Motivation and Context

Fixes #2236 

### Description

In order to make .NET and Python implementations consistent, we're changing the format and interpretation of thread_id values for the built-in HTTP triggers for durable agents.

This is a breaking change for clients that attempt to construct their own thread IDs. However, the expected common case is server-generated thread IDs.

This change also fixes some errors in the single-agent Python sample for Azure Functions.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.Hosting.AzureFunctions/BuiltInFunctions.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Hosting.AzureFunctions/BuiltInFunctions.cs
@@ -83,12 +83,13 @@ internal static class BuiltInFunctions
 
         string? threadIdValue = threadIdFromBody ?? threadIdFromQuery;
 
-        // If no session ID is provided, use a new one based on the function name and invocation ID.
-        // This may be better than a random one because it can be correlated with the function invocation.
-        // Specifying a session ID is how the caller correlates multiple calls to the same agent session.
+        // The thread_id is treated as a session key (not a full session ID).
+        // If no session key is provided, use the function invocation ID as the session key
+        // to help correlate the session with the function invocation.
+        string agentName = GetAgentName(context);
         AgentSessionId sessionId = string.IsNullOrEmpty(threadIdValue)
-            ? new AgentSessionId(GetAgentName(context), context.InvocationId)
-            : AgentSessionId.Parse(threadIdValue);
+            ? new AgentSessionId(agentName, context.InvocationId)
+            : new AgentSessionId(agentName, threadIdValue);
 
         if (string.IsNullOrWhiteSpace(message))
         {
@@ -110,7 +111,7 @@ internal static class BuiltInFunctions
             }
         }
 
-        AIAgent agentProxy = client.AsDurableAgentProxy(context, GetAgentName(context));
+        AIAgent agentProxy = client.AsDurableAgentProxy(context, agentName);
 
         DurableAgentRunOptions options = new() { IsFireAndForget = !waitForResponse };
 
@@ -126,7 +127,7 @@ internal static class BuiltInFunctions
                 req,
                 context,
                 HttpStatusCode.OK,
-                sessionId.ToString(),
+                sessionId.Key,
                 agentResponse);
         }
 
@@ -140,7 +141,7 @@ internal static class BuiltInFunctions
         return await CreateAcceptedResponseAsync(
             req,
             context,
-            sessionId.ToString());
+            sessionId.Key);
     }
 
     public static async Task<string?> RunMcpToolAsync(
--- a/python/samples/getting_started/azure_functions/01_single_agent/README.md
+++ b/python/samples/getting_started/azure_functions/01_single_agent/README.md
@@ -18,20 +18,40 @@ Follow the common setup steps in `../README.md` to install tooling, configure Az
 
 Send a prompt to the Joker agent:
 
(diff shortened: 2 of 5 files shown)


python/samples/getting_started/azure_functions/01_single_agent/README.md
```diff
@@ -18,20 +18,40 @@ Follow the common setup steps in `../README.md` to install tooling, configure Az
...
 ```bash
-curl -X POST http://localhost:7071/api/agents/Joker/run \
-     -H "Content-Type: text/plain" \
+curl -i -X POST http://localhost:7071/api/agents/Joker/run \
      -d "Tell me a short joke about cloud computing."
 ```
 
+PowerShell:
+
+```powershell
```
larohra, line 30: NIT: Do you want to add powershell examples to other samples too for Python? Currently we only document bash ones.

python/samples/getting_started/azure_functions/01_single_agent/README.md
```diff
@@ -18,20 +18,40 @@ Follow the common setup steps in `../README.md` to install tooling, configure Az
...
+HTTP/1.1 200 OK
+Content-Type: text/plain; charset=utf-8
+x-ms-thread-id: 4f205157170244bfbd80209df383757e
+
+Why did the cloud break up with the server?
+
+Because it found someone more "uplifting"!
+```
+
+When you specify the `x-ms-wait-for-response` header or include `"wait_for_response": false` in the request body, the Functions host responds with an HTTP 202 and queues the request to run in the background. A typical response body looks like the following:
```
larohra, line 54: This feels like duplicated information of line 38.

dotnet/src/Microsoft.Agents.AI.Hosting.AzureFunctions/BuiltInFunctions.cs
```diff
@@ -83,12 +83,13 @@ public static async Task<HttpResponseData> RunAgentHttpAsync(
 
         string? threadIdValue = threadIdFromBody ?? threadIdFromQuery;
 
-        // If no session ID is provided, use a new one based on the function name and invocation ID.
-        // This may be better than a random one because it can be correlated with the function invocation.
-        // Specifying a session ID is how the caller correlates multiple calls to the same agent session.
+        // The thread_id is treated as a session key (not a full session ID).
```
kshyju, line 86: I reliazed the MCP tool reads thread id from an argument named `threadId`. Should we consider changing that to `thread_id`? 

https://github.com/microsoft/agent-framework/blob/132597957abab73e7fbe1393556457b64ae2f2b2/dotnet/src/Microsoft.Agents.AI.Hosting.AzureFunctions/BuiltInFunctions.cs#L164

Needs to be fixed on the metadata transformer as well.

python/samples/getting_started/azure_functions/01_single_agent/README.md
cgillum: We should, but I didn't want to make updating all the Python samples a part of this PR. I updated this one because I needed to use it as a reference for my .NET changes.

python/samples/getting_started/azure_functions/01_single_agent/README.md
cgillum: Really? I didn't see it as duplicative, but rather complimentary. Line 38 explains what the default behavior is for _requests_ and how to change it. This line here explains what the _response_ you can expect when you use the 202 pattern.

dotnet/src/Microsoft.Agents.AI.Hosting.AzureFunctions/BuiltInFunctions.cs
cgillum: This is not part of the HTTP API, so I don't think it's necessary to change it. This is purely an LLM-facing API and how we name it isn't important.

kshyju (approved):

python/samples/getting_started/azure_functions/01_single_agent/README.md
larohra: Both of them mention returning a 202 so felt like duplicative, maybe just saying "with this header, expected response is"? 

Either way, a minor nitpick, not a blocking comment at all :)

larohra (approved):

PR #2285, opened 2025-11-17
Python: [Breaking] Python: Respond with `AgentRunResponse` with serialized structured output

larohra:
### Motivation and Context

- Fixed the `DurableAIAgent.run` function to respond with `AgentRunResponse` rather than `Any` to be more in line with agent-framework's protocols. 
- Handling deserialization of the `response.value` as part of the agent run, making it easier for the customer to author their codes.
- Remove unnecessary class wrappers (`AgentResponse`).
- Fixes https://github.com/Azure/durable-agent-framework/issues/133 and https://github.com/Azure/durable-agent-framework/issues/164

### Description

The changes bring parity with Dotnet experience and also with agent-framework in general. 

This introduces a breaking change where now we need to use `yield from` when calling the `DurableAIAgent` (vs just `yield` before) - 
```py
initial_raw = yield from writer.run(
        messages=f"Write a short article about '{payload.topic}'.",
        thread=writer_thread,
        response_format=GeneratedContent,
    )
```

Additionally, the responses are now of type `AgentRunResponse` and the structured response is stored in `AgentRunResponse.value` (as opposed to `response["structured_response"]` before). 

### Contribution Checklist


markwallace-microsoft:
<a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/README.md">Python Test Coverage</a><details><summary>Python Test Coverage Report • </summary><table><tr><th>File</th><th>Stmts</th><th>Miss</th><th>Cover</th><th>Missing</th></tr><tbody><tr><td colspan="5"><b>packages/azurefunctions/agent_framework_azurefunctions</b></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_entities.py">_entities.py</a></td><td>179</td><td>21</td><td>88%</td><td><a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_entities.py#L120">120</a>, <a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_entities.py#L160-L161">160&ndash;161</a>, <a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_entities.py#L212">212</a>, <a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_entities.py#L220">220</a>, <a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_entities.py#L225-L226">225&ndash;226</a>, <a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_entities.py#L253-L254">253&ndash;254</a>, <a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_entities.py#L263">263</a>, <a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_entities.py#L270">270</a>, <a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_entities.py#L281">281</a>, <a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_entities.py#L287-L288">287&ndash;288</a>, <a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_entities.py#L307-L308">307&ndash;308</a>, <a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_entities.py#L378">378</a>, <a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_entities.py#L410-L412">410&ndash;412</a>, <a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_entities.py#L414">414</a></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_models.py">_models.py</a></td><td>150</td><td>12</td><td>92%</td><td><a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_models.py#L204">204</a>, <a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_models.py#L207">207</a>, <a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_models.py#L210">210</a>, <a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_models.py#L222">222</a>, <a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_models.py#L225">225</a>, <a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_models.py#L244">244</a>, <a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_models.py#L247">247</a>, <a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_models.py#L252">252</a>, <a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_models.py#L257">257</a>, <a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_models.py#L274">274</a>, <a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_models.py#L329">329</a>, <a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_models.py#L348">348</a></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_orchestration.py">_orchestration.py</a></td><td>87</td><td>18</td><td>79%</td><td><a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_orchestration.py#L168">168</a>, <a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_orchestration.py#L173">173</a>, <a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_orchestration.py#L175-L176">175&ndash;176</a>, <a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_orchestration.py#L178">178</a>, <a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_orchestration.py#L193">193</a>, <a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_orchestration.py#L202-L203">202&ndash;203</a>, <a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_orchestration.py#L205">205</a>, <a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_orchestration.py#L261">261</a>, <a href="https://github.com/microsoft/agent-framework/blob/fbff50926372ba98cd7658be53553701ff1e1bfd/packages/azurefunctions/agent_framework_azurefunctions/_orchestration.py#L266-L273">266&ndash;273</a></td></tr><tr><td><b>TOTAL</b></td><td><b>15449</b></td><td><b>2360</b></td><td><b>84%</b></td><td>&nbsp;</td></tr></tbody></table></details>

## Python Unit Test Overview
| 2220 | 130 :zzz: | 0 :x: | 0 :fire: | 53.413s :stopwatch: |

python/samples/getting_started/azure_functions/05_multi_agent_orchestration_concurrency/function_app.py
```diff
@@ -51,7 +51,13 @@ def _create_agents() -> list[Any]:
...
 @app.orchestration_trigger(context_name="context")
 def multi_agent_concurrent_orchestration(context: DurableOrchestrationContext):
-    """Fan out to two domain-specific agents and aggregate their responses."""
+    """Fan out to two domain-specific agents and aggregate their responses.
+    
+    Note: This uses generator protocol to achieve true parallelism.
+    For sequential execution, you can simply use:
+        physicist_result = yield from physicist.run(messages=prompt, thread=physicist_thread)
+        chemist_result = yield from chemist.run(messages=prompt, thread=chemist_thread)
+    """
```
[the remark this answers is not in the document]
larohra: Reworded it

python/samples/getting_started/azure_functions/05_multi_agent_orchestration_concurrency/function_app.py
```diff
@@ -63,14 +69,38 @@ def multi_agent_concurrent_orchestration(context: DurableOrchestrationContext):
...
+    # Complete the generator protocol by sending results back
+    # Each generator returns its final value via StopIteration.value
+    physicist_result = None
+    chemist_result = None
+    
+    try:
+        physicist_gen.send(task_results[0])
+    except StopIteration as e:
+        physicist_result = e.value
+
```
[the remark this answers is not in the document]
larohra: Removed

python/packages/azurefunctions/agent_framework_azurefunctions/_orchestration.py
```diff
@@ -106,13 +120,13 @@ def run(
             def my_orchestration(context):
                 agent = app.get_agent(context, "MyAgent")
                 thread = agent.get_new_thread()
-                result = yield agent.run("Hello", thread=thread)
+                result = yield from agent.run("Hello", thread=thread)
         """
         message_str = self._normalize_messages(messages)
+        logger.debug(f"[DurableAIAgent] Running agent '{self.agent_name}' with message: {message_str[:100]}...")
 
         # Extract optional parameters from kwargs
```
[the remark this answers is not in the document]
larohra: Removed that code, irrelevant comment

python/packages/azurefunctions/agent_framework_azurefunctions/_entities.py
```diff
@@ -131,46 +125,38 @@ async def run_agent(
...
-                        logger.debug("Parsed structured JSON response")
-                    except json.JSONDecodeError as decode_error:
-                        logger.warning(f"Failed to parse JSON response: {decode_error}")
-                        response_text = response_str
-                else:
-                    raw_text = agent_run_response.text
-                    response_text = raw_text if raw_text else "No response"
-                    preview = response_text
-                    logger.debug(f"Response: {preview[:100]}..." if len(preview) > 100 else f"Response: {preview}")
+                raw_text = agent_run_response.text
```
gavin-aguiar: We don't need the release format anymore? How will the response be parsed if the text is in a json format?

gavin-aguiar: Also if we are not going to use the response_format lets remove the variable and the log

python/packages/azurefunctions/agent_framework_azurefunctions/_entities.py
```diff
@@ -181,16 +167,28 @@ async def run_agent(
...
-                status="error",
-                message_count=self.state.message_count,
-                error=str(exc),
-                error_type=type(exc).__name__,
+            error_text = f"Error: {exc!s}"
+            error_metadata = {
+                "status": "error",
+                "error": str(exc),
+                "error_type": type(exc).__name__,
+                "message": str(message),
```
gavin-aguiar: same here

python/packages/azurefunctions/agent_framework_azurefunctions/_entities.py
```diff
@@ -181,16 +167,28 @@ async def run_agent(
...
+                "thread_id": str(thread_id),
+                "correlation_id": correlation_id,
+                "message_count": self.state.message_count,
+            }
+
+            error_response = AgentRunResponse(
+                messages=[ChatMessage(role="assistant", text=error_text)],
```
gavin-aguiar: Use the ENUM

python/packages/azurefunctions/agent_framework_azurefunctions/_entities.py
```diff
@@ -181,16 +167,28 @@ async def run_agent(
...
+                additional_properties=error_metadata,
             )
-            return error_response.to_dict()
+
+            try:
+                self.state.add_assistant_message(error_text, error_response, correlation_id)
+            except Exception:  # pragma: no cover - defensive logging only
+                logger.warning("[AgentEntity.run_agent] Failed to record error response in state", exc_info=True)
```
gavin-aguiar: We should log the exception message too. Just this error message doesn't give much info

python/packages/azurefunctions/agent_framework_azurefunctions/_entities.py
```diff
@@ -131,46 +125,38 @@ async def run_agent(
...
-                thread_id=str(thread_id),
-                status="success",
-                message_count=self.state.message_count,
-                structured_response=structured_response,
-            )
-            result = agent_response.to_dict()
+            message_count = self.state.message_count
+            metadata: dict[str, Any] = {
+                "status": "success",
+                "message": str(message),
```
gavin-aguiar: message and thread_id are always strings. No need of casting

python/packages/azurefunctions/agent_framework_azurefunctions/_entities.py
eavanvalkenburg, line 143: Note that we parse a response format if available into the `value` field on a ChatResponse or AgentRunResponse

larohra: Thanks @eavanvalkenburg. In our use case, the LLM calls happen on a different compute from where the customer triggers their agent run (basically `agent.run` call can be on a different machine vs where the actual call to the agent happens). All this plumbing is to ensure that we are able to stream back the response from the compute that runs the agent to the compute that triggers it. This PR actually brings parity with `agent_framework` and ensures that the final response that we send back to the customer has the formatted response in the `value` property of `AgentRunResponse`.

python/packages/azurefunctions/agent_framework_azurefunctions/_entities.py
larohra: refactored

python/packages/azurefunctions/agent_framework_azurefunctions/_entities.py
```diff
@@ -164,83 +154,42 @@ async def run_agent(
...
-            )
-            result = agent_response.to_dict()
-
             logger.debug("[AgentEntity.run_agent] AgentRunResponse stored in conversation history")
 
-            return result
+            return agent_run_response
 
         except Exception as exc:
             import traceback
```
gavin-aguiar: This can be moved to the top of the file

gavin-aguiar: Also note that instead of manually formatting the traceback, you could use `logger.exception("[AgentEntity.run_agent] Agent execution failed")`  
This should automatically log the stack trace. If you want the formatted traceback string in addition to that or  what you have is fine.

python/packages/azurefunctions/agent_framework_azurefunctions/_orchestration.py
```diff
@@ -96,23 +109,24 @@ def run(
...
         Example:
             @app.orchestration_trigger(context_name="context")
             def my_orchestration(context):
                 agent = app.get_agent(context, "MyAgent")
                 thread = agent.get_new_thread()
-                result = yield agent.run("Hello", thread=thread)
+                result = yield from agent.run("Hello", thread=thread)
         """
         message_str = self._normalize_messages(messages)
+        logger.debug(f"[DurableAIAgent] Running agent '{self.agent_name}' with message: {message_str[:100]}...")
```
gavin-aguiar: This log is the same as the one on line 163. We can remove this or the other one

python/packages/azurefunctions/agent_framework_azurefunctions/_orchestration.py
```diff
@@ -96,23 +109,24 @@ def run(
...
 
         # Extract optional parameters from kwargs
         enable_tool_calls = kwargs.get("enable_tool_calls", True)
```
gavin-aguiar, line 249: Unrelated to your change but we should make this as an explicit keyword argument instead of pulling it from kwargs. You can either change it here or lets create a follow up task.

python/packages/azurefunctions/agent_framework_azurefunctions/_orchestration.py
```diff
@@ -144,7 +164,57 @@ def my_orchestration(context):
...
+            self._ensure_response_format(response_format, correlation_id, response)
+
+        return response
+
+    def _load_agent_response(self, agent_response: AgentRunResponse | dict[str, Any] | None) -> AgentRunResponse:
+        """Convert raw payloads into AgentRunResponse instance."""
+        if agent_response is None:
+            raise ValueError("agent_response cannot be None")
+
+        logger.debug(f"[load_agent_response] Loading agent response of type: {type(agent_response)}")
```
gavin-aguiar: NIT: Can we not use f strings for logging? There is a perf hit with logging and fstrings. I had to change all the python worker logs from fstring to lazy logging because of this.

gavin-aguiar: Maybe we need to do that for the whole package. We can do it in another task

larohra: Thanks for the catch, I wasnt aware of the perf impact. 
But I agree, we can do this as a separate task.

python/packages/azurefunctions/agent_framework_azurefunctions/_orchestration.py
larohra: Good point. Let's target this in a follow up task rather than here to keep changes isolated.

python/packages/azurefunctions/agent_framework_azurefunctions/_orchestration.py
larohra: Good catch, removed

python/packages/azurefunctions/agent_framework_azurefunctions/_entities.py
larohra: Done

gavin-aguiar (approved):

python/packages/azurefunctions/agent_framework_azurefunctions/_entities.py
```diff
@@ -164,83 +154,39 @@ async def run_agent(
...
-            logger.error(f"Full traceback:\n{error_traceback}")
+            logger.exception("[AgentEntity.run_agent] Agent execution failed.")
 
             # Create error message
-            error_message = DurableAgentStateMessage.from_chat_message(
-                ChatMessage(
-                    role=Role.ASSISTANT, contents=[ErrorContent(message=str(exc), error_code=type(exc).__name__)]
-                )
+            error_message = ChatMessage(
+                role=Role.ASSISTANT, contents=[ErrorContent(message=str(exc), error_code=type(exc).__name__)]
```
cgillum, line 179: It looks like you're creating a custom error response from the entity when something unexpected happens, but I don't believe we're doing this in .NET. I'm also not sure if this is the right thing for us to do since it hides the fact that an error occurred from Functions telemetry. Regardless of what design we think is best, we should do the same thing in both languages. Can you open an issue to review the implementation and track consistent error handling for agent invocations?

python/samples/getting_started/azure_functions/05_multi_agent_orchestration_concurrency/function_app.py
```diff
@@ -63,14 +71,38 @@ def multi_agent_concurrent_orchestration(context: DurableOrchestrationContext):
...
+    # Execute both tasks concurrently using task_all
+    task_results = yield context.task_all([physicist_task, chemist_task])
+
+    # Complete the generator protocol by sending results back
+    # Each generator returns its final value via StopIteration.value
+    physicist_result = None
+    chemist_result = None
+
+    try:
+        physicist_gen.send(task_results[0])
```
cgillum: I've never seen this direct access of generators and `send` in durable Python apps before, and I don't think it's intended that we make this a normal part of the programming model. Why is this needed, and can it be replaced with something more conventional?

cgillum (commented):
In addition to the feedback below, can we open an issue to track updating the snippets lab (the one we used for Ignite) to account for these breaking changes?

python/packages/azurefunctions/agent_framework_azurefunctions/_entities.py
larohra: I agree, the experiences should be consistent. 
Created an issue to track this separately - https://github.com/microsoft/agent-framework/issues/2434

python/samples/getting_started/azure_functions/05_multi_agent_orchestration_concurrency/function_app.py
larohra: I've added a comment at the top of this file, but this is basically only for the advance cases where the customers want true parallelism with invoking multi-agents with generators. 

```py
 """Fan out to two domain-specific agents and aggregate their responses.

    Note: The generator protocol lets us extract the Durable Task objects so we
    can pass them to task_all for true parallelism. If you only use `yield
    from` on each run call, the agents execute sequentially.
    For sequential execution, you can simply use:
        physicist_result = yield from physicist.run(messages=prompt, thread=physicist_thread)
        chemist_result = yield from chemist.run(messages=prompt, thread=chemist_thread)
    """
```

The conventional method is defined throughout the rest of the samples with using the `yield from` model. The only catch here is that that convention is more synchronous - 

```py
email_result_raw = yield from email_agent.run(
        messages=email_prompt,
        thread=email_thread,
        response_format=EmailResponse,
    )
```

larohra:
> In addition to the feedback below, can we open an issue to track updating the snippets lab (the one we used for Ignite) to account for these breaking changes?

Opened an issue #2435

python/samples/getting_started/azure_functions/05_multi_agent_orchestration_concurrency/function_app.py
cgillum: I understand what the sample is doing, but I don't understand why we can't use normal generator syntax. This is not something durable customers need to do in Python today: https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-overview?tabs=in-process%2Cnodejs-v3%2Cv1-model&pivots=python#fan-in-out

larohra: The difference is in the response type of the types. The `context.task_all` expects a `list[TaskBase]` but the `agent.run` responds with a `Generator[.., .., AgentRunResponse]` due to which it can't resolve properly. The main reason why this is happening is because we do some post-processing after we get the result back from the entity after we signal it using `context.signal_entity()` and return an object of `AgentRunResponse`. Earlier, we were just returning the same task that we got when executing teh `signal_entity` command without any post-processing. 

```py
result = yield self.context.call_entity(entity_id, "run_agent", run_request.to_dict())

logger.debug(
    "[DurableAIAgent] Entity call completed for correlation_id %s",
    correlation_id,
)

response = self._load_agent_response(result)

if response_format is not None:
    self._ensure_response_format(response_format, correlation_id, response)

return response
```

larohra: Rearchitected to use `TaskBase` instead of `Generators` to be consistent and simpler from a customer PoV.

python/samples/getting_started/azure_functions/05_multi_agent_orchestration_concurrency/function_app.py
```diff
@@ -63,14 +64,18 @@ def multi_agent_concurrent_orchestration(context: DurableOrchestrationContext):
     physicist_thread = physicist.get_new_thread()
     chemist_thread = chemist.get_new_thread()
 
-    physicist_task = physicist.run(messages=str(prompt), thread=physicist_thread)
-    chemist_task = chemist.run(messages=str(prompt), thread=chemist_thread)
+    # Create generators from agent.run() calls
+    physicist_gen = physicist.run(messages=str(prompt), thread=physicist_thread)
+    chemist_gen = chemist.run(messages=str(prompt), thread=chemist_thread)
```
cgillum: I think we should be able to revert these changes now that we're returning tasks again.

cgillum (commented):
The latest version looks great! Just one comment to fix up the sample.

python/packages/azurefunctions/agent_framework_azurefunctions/_orchestration.py
```diff
@@ -6,21 +6,153 @@
...
+    ) -> None:
+        """Ensure the AgentRunResponse value is parsed into the expected response_format."""
+        if response_format is not None and not isinstance(response.value, response_format):
+            response.try_parse_value(response_format)
+
+            logger.debug(
+                "[DurableAIAgent] Loaded AgentRunResponse.value for correlation_id %s with type: %s",
+                correlation_id,
+                type(response.value).__name__,
+            )
```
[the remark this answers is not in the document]
larohra: We already have this check in line 148

python/packages/azurefunctions/agent_framework_azurefunctions/_orchestration.py
```diff
@@ -6,21 +6,153 @@
...
+        # This ensures compatibility with ReplaySchema V3 which expects Action objects.
+        self.action_repr = entity_task.action_repr
+
+        # Also copy the task ID to match the entity task's identity
+        self.id = entity_task.id
+
+    def try_set_value(self, child: TaskBase):
+        """Intercept task completion and convert raw result to typed AgentRunResponse.
+
+        This method delegates to the parent WhenAllTask behavior but intercepts
```
gavin-aguiar: If we are delegating to the parent, we shoud call super().try.. 
Else the docs need to be updated

python/packages/azurefunctions/agent_framework_azurefunctions/_orchestration.py
```diff
@@ -6,21 +6,153 @@
...
+                    # Set the typed AgentRunResponse as this task's result
+                    self.set_value(is_error=False, value=response)
+                except Exception as e:
+                    logger.error(
+                        "[AgentTask] Failed to convert result for correlation_id %s: %s",
+                        self._correlation_id,
+                        e,
+                    )
+                    self.set_value(is_error=True, value=e)
+        else:  # child.state is TaskState.FAILED
```
gavin-aguiar: TaskState could be running too. If its a long running task this will count as an error.

python/packages/azurefunctions/agent_framework_azurefunctions/_orchestration.py
```diff
@@ -6,21 +6,153 @@
...
+                        self._ensure_response_format(
+                            self._response_format,
+                            self._correlation_id,
+                            response,
+                        )
+
+                    # Set the typed AgentRunResponse as this task's result
+                    self.set_value(is_error=False, value=response)
+                except Exception as e:
+                    logger.error(
```
gavin-aguiar: logger.exception or use exc_info=True to preserve the stacktrace

python/packages/azurefunctions/agent_framework_azurefunctions/_orchestration.py
```diff
@@ -6,21 +6,153 @@
...
+            if self._first_error is None:
+                self._first_error = child.result
+                self.set_value(is_error=True, value=self._first_error)
+
+    def _load_agent_response(self, agent_response: AgentRunResponse | dict[str, Any] | None) -> AgentRunResponse:
+        """Convert raw payloads into AgentRunResponse instance."""
+        if agent_response is None:
+            raise ValueError("agent_response cannot be None")
+
+        logger.debug(f"[load_agent_response] Loading agent response of type: {type(agent_response)}")
```
gavin-aguiar: Change this to lazy logging once you're making any change

python/packages/azurefunctions/agent_framework_azurefunctions/_orchestration.py
larohra: Good point. I modeled this after the existing `WhenAllTask` class in the af-durable sdk and this is how they're handling it - 

```py
def try_set_value(self, child: TaskBase):
        """Transition a WhenAll Task to a terminal state and set its value.

        Parameters
        ----------
        child : TaskBase
            A sub-task that just completed
        """
        if child.state is TaskState.SUCCEEDED:
            # A WhenAll Task only completes when it has no pending tasks
            # i.e _when all_ of its children have completed
            if len(self.pending_tasks) == 0:
                results = list(map(lambda x: x.result, self.children))
                self.set_value(is_error=False, value=results)
        else:  # child.state is TaskState.FAILED:
            # a single error is sufficient to fail this task
            if self._first_error is None:
                self._first_error = child.result
                self.set_value(is_error=True, value=self._first_error)
``` 

There's not a lot of documentation around it and given that we're thinking of moving to the new durable-sdks I am leaning towards just keeping it as is for now.

python/packages/azurefunctions/agent_framework_azurefunctions/_orchestration.py
larohra: Its already being handled by the base, let me rephrase the comments.

python/packages/azurefunctions/agent_framework_azurefunctions/_orchestration.py
larohra: Resolving it for now, we can iterate over this if needed in a follow up PR

cgillum (approved):

Issue #2179, opened 2025-11-13
.NET: Add Support for Passing Agent Name (String) Instead of AIAgent Instance to MapAGUI Method


Varorbc:
It would be great to enhance this API by allowing MapAGUI to accept an agent's name (string) instead of requiring the actual AIAgent instance. 

cc @javiercn 

https://github.com/microsoft/agent-framework/blob/7e5de8f92058e5507a216cac54eb2e723ce65849/dotnet/src/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore/AGUIEndpointRouteBuilderExtensions.cs#L31-L34

alliscode:
Thanks @Varorbc, we'll take a look.

halllo:
I also need this. Checkout https://github.com/microsoft/agent-framework/pull/2343.

```csharp
app.MapAGUI("/agents/{agentId}", async (context) =>
{
    var agentId = context.Request.RouteValues["agentId"]?.ToString() ?? string.Empty;
    return agentId switch
    {
        "agentic_chat" => ChatClientAgentFactory.CreateAgenticChat(),
        "backend_tool_rendering" => ChatClientAgentFactory.CreateBackendToolRendering(),
        "human_in_the_loop" => ChatClientAgentFactory.CreateHumanInTheLoop(),
        "tool_based_generative_ui" => ChatClientAgentFactory.CreateToolBasedGenerativeUI(),
        "agentic_generative_ui" => ChatClientAgentFactory.CreateAgenticUI(),
        "shared_state" => ChatClientAgentFactory.CreateSharedState(jsonOptions.Value.SerializerOptions),
        _ => throw new ArgumentException($"Unknown agent ID: {agentId}"),
    };
});
```

markwallace-microsoft:
@Varorbc, friendly reminder — this issue is waiting on your response. Please share any updates when you get a chance. (This is an automated message.)

Varorbc:
@markwallace-microsoft What can I do now?

javiercn:
The requested named-agent overload is available in the current hosting API.

Register the agent by name:

```csharp
builder.Services.AddAGUIServer();

builder.AddAIAgent("assistant", (_, _) => agent)
    .WithInMemorySessionStore();
```

Then map it using the string name:

```csharp
var app = builder.Build();
app.MapAGUIServer("assistant", "/agent");
```

`MapAGUIServer` resolves the keyed `AIAgent` registration and its associated session store.

[...]

PR #2343, opened 2025-11-20
.NET: MapAGUI supports per-request agent selection

halllo:
### Motivation and Context

The current AGUI agent mapping binds one agent to an endpoint route. There is currently no way to bind an agent dynamically to an endpoint route with route parameters.

With this change we can select an `AIAgent` per-request based on route parameters or anything from `HttpContext`.

This also addresses https://github.com/microsoft/agent-framework/issues/2179.

### Description

Additionally to providing an `AIAgent` instance directly during route building, we can now also provide a `Func<HttpContext, AIAgent>` to determine the `AIAgent` per-request on-demand based on route parameters or anything from `HttpContext`.

### Contribution Checklist

### DIFF
--- a/dotnet/samples/AGUIClientServer/AGUIDojoServer/Program.cs
+++ b/dotnet/samples/AGUIClientServer/AGUIDojoServer/Program.cs
@@ -40,6 +40,21 @@ app.MapAGUI("/agentic_generative_ui", ChatClientAgentFactory.CreateAgenticUI());
 var jsonOptions = app.Services.GetRequiredService<IOptions<Microsoft.AspNetCore.Http.Json.JsonOptions>>();
 app.MapAGUI("/shared_state", ChatClientAgentFactory.CreateSharedState(jsonOptions.Value.SerializerOptions));
 
+app.MapAGUI("/agents/{agentId}", async (context) =>
+{
+    var agentId = context.Request.RouteValues["agentId"]?.ToString() ?? string.Empty;
+    return agentId switch
+    {
+        "agentic_chat" => ChatClientAgentFactory.CreateAgenticChat(),
+        "backend_tool_rendering" => ChatClientAgentFactory.CreateBackendToolRendering(),
+        "human_in_the_loop" => ChatClientAgentFactory.CreateHumanInTheLoop(),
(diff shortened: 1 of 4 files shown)


halllo:

scrodde:
Nice, need this as well!

halllo:
@javiercn @moonbox3 @dmytrostruk @stephentoub what do you think?

javiercn:
@halllo this change deviates from the approach we follow in all the other hosting packages (A2A, OpenID completions/responses) so I'm not keen on doing something different here.

There is also https://github.com/microsoft/agent-framework/pull/2547 which brings additional hosting overloads and aligns with the other hosting packages, and I believe should potentially allow for doing this.

halllo:
@javiercn thanks for the feedback. I am trying to solve dynamic agent selection, where existence of certain agents does not have to be determined at application startup.
I am aware of your PR and it still requires the agent to be passed to MapAGUI() or resolved at startup time from service provider. That prevents dynamic per-request selection.
How would you implement endpoints like these?
- **/api/agents** (POST/PUT/DELETE) to manage agents at runtime
- **/api/agents/{agentid}/agui** (POST) to use a dynamically created agent

erikostling:
Is there any update on this PR? I'm facing the same situation as @halllo and would appreciate clarity on the roadmap so I can move forward.

halllo:
There is a workaround available via a `HttpContextRoutingAgent` https://github.com/microsoft/agent-framework/pull/3162#issuecomment-3902133512

chetantoshniwal:
@javiercn  can you validate if this is still required.

javiercn:
Thanks for the work on request-time agent selection. We have now consolidated the broader requirement into #7733.

The same scoped-agent and request-time selection problem exists in AG-UI, OpenAI Chat Completions, OpenAI Responses, and A2A. We do not want to establish a `Func<HttpContext, AIAgent>` pattern exclusively in the AG-UI integration and then solve the other hosting extensions differently.

#7733 will track a consistent cross-host pattern, including request-scope resolution, stable hosted-agent identity, and session-store ownership. The scenario and implementation work in this PR are useful input to that design.

Issue #2325, opened 2025-11-19
.NET: Chat Message Storage Enhancements


westey-m:
- [X] Consider support for bootstrapping new threads where the user has a pre-existing database with chat history or an external id generator.
    - [x] One option is `agent.GetNewThread(new CustomMessageStore(dbClient, id));` but this may be painful to use if dbClient needs to be accessible everywhere GetNewThread is used.
    - [ ] Another option is allowing some arbitrary payload to be passed via GetNewThread to the message store factory, although this would not be strongly typed. (Added issue #3104 to continue looking into this)
- [x] Consider creating custom messages stores at thread creation time from the factory, even if they may be removed after the first run in favor of a conversation id, to allow users to more easily update them if needed. (Since this is an alternative to item 1, it may be redundant).
- [ ] Consider supporting overriding the message store on a per run basis. (Added issue #3104 to continue looking into this)

westey-m:
Moved making the factory methods async to a separate work item: #2520

westey-m:
Closing this issue, as the all remaining work will be considered in the context of #3104

PR #2354, opened 2025-11-20
.NET: Add per run / thread feature collection support and improved custom ChatMessageStore support

westey-m:
### Motivation and Context

Today there isn't good support for scenarios where chat history is stored in a custom store and a developer wants to bootstrap new threads with the existing data. There also isn't good support for cases where a host application may want to override any chat history storage that an agent uses, to better support the protocol that the agent is exposed via.

https://github.com/microsoft/agent-framework/issues/2325

### Description

- Adding the ability to supply a ChatMessageStore when creating a new thread via ChatClientAgent.
- Adding the ability to supply override features to GetNewThread, DeserializeThread and RunAsync/RunStreamingAsync that allow per-thread or per-run chat message store overrides.
- Creating the ChatMessageStore on ChatClientAgentThreads when the ChatClientAgentThreads is created to allow retrieving and modifying it before the first run.

### Contribution Checklist


dotnet/src/Microsoft.Agents.AI.Abstractions/Features/AgentFeatureCollection.cs
```diff
@@ -0,0 +1,192 @@
...
+/// Default implementation for <see cref="IAgentFeatureCollection"/>.
+/// </summary>
+[DebuggerDisplay("Count = {GetCount()}")]
+[DebuggerTypeProxy(typeof(FeatureCollectionDebugView))]
+public class AgentFeatureCollection : IAgentFeatureCollection
+{
+    private static readonly KeyComparer s_featureKeyComparer = new();
+    private readonly IAgentFeatureCollection? _defaults;
+    private readonly int _initialCapacity;
+    private Dictionary<Type, object>? _features;
```
stephentoub: If an initial capacity is provided, why not just instantiate the dictionary then and there, rather than carrying around the initial capacity?

dotnet/src/Microsoft.Agents.AI.Abstractions/Features/AgentFeatureCollection.cs
```diff
@@ -0,0 +1,192 @@
...
+        Throw.IfLessThan(initialCapacity, 0);
+
+        this._initialCapacity = initialCapacity;
+    }
+
+    /// <summary>
+    /// Initializes a new instance of <see cref="AgentFeatureCollection"/> with the specified defaults.
+    /// </summary>
+    /// <param name="defaults">The feature defaults.</param>
+    public AgentFeatureCollection(IAgentFeatureCollection defaults)
```
stephentoub: I find it a little confusing that this is called "defaults". That suggests to me that the data from this collection is somehow being copied into this one at the time of construction, but it's not; this is a decorator that wraps it and delegates to it when a value isn't satisfied by this instance. Maybe "innerCollection"?

dotnet/src/Microsoft.Agents.AI.Abstractions/Features/AgentFeatureCollection.cs
```diff
@@ -0,0 +1,192 @@
...
+    {
+        this._defaults = defaults;
+    }
+
+    /// <inheritdoc />
+    public virtual int Revision
```
stephentoub: What is the need for this?

dotnet/src/Microsoft.Agents.AI.Abstractions/Features/AgentFeatureCollection.cs
```diff
@@ -0,0 +1,192 @@
...
+        }
+
+        if (this._defaults != null)
+        {
+            // Don't return features masked by the wrapper.
+            foreach (var pair in this._features == null ? this._defaults : this._defaults.Except(this._features, s_featureKeyComparer))
+            {
+                yield return pair;
+            }
+        }
```
stephentoub: Consider instead something like:
```csharp
if (this._features is not { Count: > 0 })
{
    IEnumerable<KeyValuePair<Type, object>> e = this._defaults ?? [];
    return e.GetEnumerator();
}

if (this._defaults is not { Count: > 0 })
{
    return this._features.GetEnumerator();
}

return YieldAll();

IEnumerator<KeyValuePair<Type, object>> YieldAll()
{
    HashSet<Type> set = [];

    foreach (var entry in this._features)
    {
        set.Add(entry.Key);
        yield return entry;
    }

    foreach (var entry in this._defaults)
    {
        if (set.Add(entry.Key))
        {
            yield return entry;
        }
    }
}
```
Then you can delete s_featureKeyComparer and the custom comparer.

dotnet/src/Microsoft.Agents.AI.Abstractions/Features/AgentFeatureCollection.cs
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    /// <inheritdoc />
+    public TFeature? Get<TFeature>()
+    {
+        if (typeof(TFeature).IsValueType)
+        {
+            var feature = this[typeof(TFeature)];
+            if (feature is null && Nullable.GetUnderlyingType(typeof(TFeature)) is null)
```
stephentoub:
```suggestion
            if (feature is null && default(TFeature) is null)
```
?

dotnet/src/Microsoft.Agents.AI.Abstractions/Features/AgentFeatureCollection.cs
```diff
@@ -0,0 +1,192 @@
...
+            // Don't return features masked by the wrapper.
+            foreach (var pair in this._features == null ? this._defaults : this._defaults.Except(this._features, s_featureKeyComparer))
+            {
+                yield return pair;
+            }
+        }
+    }
+
+    /// <inheritdoc />
+    public TFeature? Get<TFeature>()
```
stephentoub: The null issues would be addressed if this were instead TryGet, or if it were constrained to `TFeature : class`

dotnet/src/Microsoft.Agents.AI.Abstractions/Features/AgentFeatureCollection.cs
westey-m: The XML docs on the interface says "Incremented for each modification and can be used to verify cached results.", but it's unclear to me what kind of verification this may be.
@ReubenBond, do you have any more context on what the purpose is, and whether it would be useful for us?

dotnet/src/Microsoft.Agents.AI.Abstractions/Features/AgentFeatureCollection.cs
westey-m: I'm just thinking through the experience we want to enable.
Right now, when someone sets a feature to null, it removes the feature from the features list and falls back to the "default", which feels weird to me. This is not what I would have expected on setting to null. Instead I would have expected that if there is a matching feature in "defaults", they are really overriding this to say that the override is null.
This also means that there really should be a separate remove method, e.g. `myFeatureCollection.Remove<MyFeature>();`

A developer should be able to choose whether they want to override with a feature, override to null or default or not override at all (fall back).  There is potentially a fourth case, overriding with "does not exist" rather than null or default, but we could assume that null/default means "does not exist".

Not sure if changing to this behavior is more like defaults or innercollection. I think either would work for me once we make this behavior more explicit.

dotnet/src/Microsoft.Agents.AI.Abstractions/Features/AgentFeatureCollection.cs
westey-m: Thanks, updated.

dotnet/src/Microsoft.Agents.AI.Abstractions/Features/AgentFeatureCollection.cs
westey-m: I've updated the behavior to be more clear about when an override is removed, and renamed to innerCollection

dotnet/src/Microsoft.Agents.AI.Abstractions/Features/AgentFeatureCollection.cs
westey-m: As discussed, kept it for now, since it is useful when caching components to determine if they need to be retrieved from the collection again (collection has changed) or not (collection has not changed).

dotnet/src/Microsoft.Agents.AI.Abstractions/Features/AgentFeatureCollection.cs
westey-m: Thanks, updated to something similar.  Had to modify it a bit since the inner collection doesn't have count, and count is expensive due to the layering.

dotnet/src/Microsoft.Agents.AI.Abstractions/Features/AgentFeatureCollection.cs
westey-m: Thanks, switched to TryGet and notnull.

dotnet/src/Microsoft.Agents.AI.Abstractions/Features/AgentFeatureCollection.cs
westey-m: Thanks, this code is now removed entirely.

westey-m:
Closing since this has been superseded by new functionality.

PR #2356, opened 2025-11-20
.NET: sanitize redirectUrl for logs

DeagleGross:
1) Sanitizing redirect url
2) Using redirect Url from the _basePath to not use input without validation


DeagleGross:
> I am not 100% sure on the right approach for sanitizing log lines. I think it's fine to just delete that log line, tbh. It's a false alarm anyway, because `path` == `_basePath` which is a constant and cannot contain newlines.

I also consider this a false alarm, but I think DevUI should have more logs, not less, to improve dev experience and debugging. That's why decided to leave it.

Decided to do the simple regex for now - we can improve later once we find out best practice for sanitization

PR #2372, opened 2025-11-21
.NET: Redacting log data unless opting out.

westey-m:
### Description

Redacting log data unless opting out for memory components.

### Contribution Checklist


dotnet/src/Microsoft.Agents.AI.Mem0/Mem0Provider.cs
rogerbarreto, line 26: Would make sense having a decorating Logging Wrapper for the `AIContextProvider` that does that, similar how we went into Agents and ChatClients?

So anyone would just wrap their component in an enabled Logging one?

```
var myProvider = new LoggingAIContextProvider(memProvider, loggerFactory) { EnableSensitiveData = true }
```

westey-m: Unfortunately a decorator wouldn't have access to much of the information being logged.  E.g. the scopes that searches are being done with and the count of results coming back (since they are just combined into a single user message).
It would just be able to log the input and output, but I think it's more valuable to have context of how many results were matched and what scopes were used.
These two are also just coincidentally similar, while other AIContextProviders would not necessarily log the same information.

dotnet/src/Microsoft.Agents.AI.Mem0/Mem0Provider.cs
```diff
@@ -282,4 +285,6 @@ public Mem0State(Mem0ProviderScope storageScope, Mem0ProviderScope searchScope)
         public Mem0ProviderScope StorageScope { get; set; }
         public Mem0ProviderScope SearchScope { get; set; }
     }
+
+    private string? SanitizeLogData(string? data) => this._enableSensitiveTelemetryData ? data : "<redacted>";
```
stephentoub, line 289: Should this employ https://www.nuget.org/packages/Microsoft.Extensions.Compliance.Redaction ?

westey-m: Thanks for the pointer.  Added a task to investigate and employ: #2375

PR #2378, opened 2025-11-21
.NET: Fix: OpenAI ChatClients do not send entity_id in AgentChatWebEx…

### LINKED ISSUE #2376: .NET : Unable to supply metadata to OpenAI ChatClients
Since PR https://github.com/microsoft/agent-framework/pull/1984 the HostedAgentResponseExecutor expects either agent.name or metadata["entity_id"] to be sent by the requesting client. 

Unfortunately this breaks the AgentWebChat (for all OpenAI clients - also the A2A is not working but I don't know if this is related)

Image

The current IChatClient abstractions of OpenAI.ChatClient and OpenAIResponseClient don't allow passing either an Agent nor metadata.

More notes:
- OpenAI.ChatClient allows metadata via the OpenAI.Chat.ChatClientOptions parameter in the GetChatCompletionAsync method
- OpenAI.ChatClient allows metadata via the OpenAI.Responses.ResponseCreationOptions  parameter in the CreateResponseStreamingAsync method

I currently see two possible solutions:
- Roll back the requirement for agent.name or metadata["entity_id"]
- MEAI has to add support so that we can pass metadata via the IChatClient abstractions

MD-V:
Modified the OpenAI clients in AgentChatWebExample to send the needed metadata.

### Motivation and Context

Fixes #2376

### Description

I've added RawRepresentationFactory as suggested with now supplies the correct metadata for HostedAgentResponseExecutor.

### Contribution Checklist


MD-V:

dotnet/samples/AgentWebChat/AgentWebChat.Web/OpenAIResponsesAgentClient.cs
```diff
@@ -30,7 +30,14 @@ public override async IAsyncEnumerable<AgentRunResponseUpdate> RunStreamingAsync
         var openAiClient = new OpenAIResponseClient(model: agentName, credential: new ApiKeyCredential("dummy-key"), options: options).AsIChatClient();
```
arjendeblok: This also has a problem, because you will have to pass a LLM model in the model name and not the agent name.

This is quite strange that you will have to pass the model and the agent whereas the agent also defines which model to use.

MD-V: For this concrete sample you won't need to pass a correct model name. The HostedAgentResponseExecutor decides based on the entity_id which agent to call (and the agent internally has the proper fixed IChatClient)

arjendeblok: I did a lot of research because I got an error in my own application that the model could not be found. I use Azure  Foundry instead of Azure OpenAI. 

I found out that this works for **Azure OpenAI** because it does not do anything with the model, but **AI Foundry** does look at it. When the responses API is used in the agent then the agent passes the model to the LLM provider and AI Foundry then looks for the model. So this will work for this example, but if you move to Azure Foundry then it suddenly breaks.

MD-V: Sorry I don't get your concern here. 
The AgentWebChat.Web sample and the DevUI don't have the concept of a model. 
The routing is done only via the entity_id.

The AgentHost has the registered chat-model IChatClient which abstracts which model is used.

arjendeblok: Sorry, but that is not correct for Azure Foundry. 

I can see the name of model passed from Web to the Agent being send as a modelId parameter from the agent to the LLM.

See https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.ai.chatoptions.modelid?view=net-10.0-pp

This overrides the 'default' model in IChatClient. This does work in this example because it uses Azure OpenAI, which does nothing with that option.

MD-V: Which class of client are you using for AI Foundry? Did you have problems with the AgenWebChat example?

arjendeblok: I am using the `AddAzureAIFoundry `from `Aspire.Hosting.Azure.AIFoundry` and the `AddAzureChatCompletionsClient `from `Aspire.Azure.AI.Inference.`

So when I use this example and switch to AI Foundry then it does not work. But the example as is will work because Azure OpenAI does not care about the model passed.

arjendeblok (approved):
For this example it works but for other LLM providers it may fail on the name of the model.

dotnet/samples/AgentWebChat/AgentWebChat.Web/OpenAIResponsesAgentClient.cs
MD-V: So you have modified the AgenWebChat example? I'm tying to replicate your code so I can debug it.

Issue #2389, opened 2025-11-21
Python: Update OpenAI Responses and "store" parameter


dmytrostruk:
At the moment, OpenAI Responses implementation uses `store` parameter as `False` by default. This should be changed to use server default behavior in case this parameter is not configured and only send it with request when it's explicitly configured.

PR #2395, opened 2025-11-22
Python: test: add unit tests for OpenAIResponsesClient store parameter handling with proper usage mock

ishanrajsingh:
This pull request adds new unit tests to the OpenAIResponsesClient to verify the correct handling of the store parameter when making API calls.

Changes made:
Added tests to ensure the store parameter is only sent when explicitly configured.

Mocked OpenAI usage token counts correctly as integers to avoid runtime errors during tests.
Improved test reliability by patching the exact API call method used internally.

Why:
These tests validate the fix for issue #2389 where the client incorrectly defaulted store to False, causing unintended behavior in response storage.

### DIFF
--- a/python/packages/core/agent_framework/openai/_responses_client.py
+++ b/python/packages/core/agent_framework/openai/_responses_client.py
@@ -366,8 +366,12 @@ class OpenAIBaseResponsesClient(OpenAIBase, BaseChatClient):
             for key, value in additional_properties.items():
                 if value is not None:
                     run_options[key] = value
-        if "store" not in run_options:
-            run_options["store"] = False
+        if "store" in run_options:
+            # keep store param as it is
+            pass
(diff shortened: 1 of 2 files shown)


dmytrostruk:
@ishanrajsingh Thanks for your contribution! This work is already in progress, and the scope is a little bit bigger, because it also includes Azure AI V2 functionality. This is also a breaking change, so all the OpenAI Responses API and Azure AI V2 samples need to be re-tested. 

We are considering making task status public, so you would be able to see which task is already in progress and which is available for contribution. Meanwhile, if you see some task and it's already assigned to some person - it's worth to double-check with assignee if the task is available for contribution before making any changes.

I will close this PR for now and new PR will be published soon with all the changes required to resolve this task. Sorry for any inconvenience!

PR #2418, opened 2025-11-24
Python: Fix #2410: Combine text content and tool_calls into single message per OpenAI API spec

### LINKED ISSUE #2410: Python [Bug] OpenAI client incorrectly splits content and tool_calls into separate messages
## Description

  The OpenAI chat client's `_openai_chat_message_parser` method incorrectly splits text content and tool calls into separate API messages when they belong to the same `ChatMessage`. According to the OpenAI API specification, a single assistant message can contain
  both `content` and `tool_calls` fields.

  ## Current Behavior

  When a `ChatMessage` contains both text content and function calls, the parser creates **two separate OpenAI messages**:

  ```python
  # Input: ChatMessage with text + tool call
  message = ChatMessage(
      role="assistant",
      contents=[
          TextContent("I'll help you with that calculation."),
          FunctionCallContent(name="calculate", call_id="call-123", arguments={"x": 5, "y": 3})
      ]
  )

  # Current (incorrect) output:
  [
      {"role": "assistant", "content": ["I'll help you with that calculation."]},
      {"role": "assistant", "tool_calls": [{"id": "call-123", "type": "function", ...}]}
  ]

  Expected Behavior

  According to the https://platform.openai.com/docs/api-reference/chat/create, text content and tool calls should be combined into a single message:

  # Expected (correct) output:
  [
      {
          "role": "assistant",
          "content": ["I'll help you with that calculation."],
          "tool_calls": [{"id": "call-123", "type": "function", ...}]
      }
  ]

  Root Cause

  File: python/packages/core/agent_framework/openai/_chat_client.py
  Function: _openai_chat_message_parser (lines 361-390)

  The function creates a new message dictionary (args) for each content item in the loop, then appends each to all_messages:

  def _openai_chat_message_parser(self, message: ChatMessage) -> list[dict[str, Any]]:
      all_messages: list[dict[str, Any]] = []
      for content in message.contents:
          args: dict[str, Any] = {"role": ...}  # ❌ New dict for each content

          match content:
              case FunctionCallContent():
                  args["tool_calls"] = [...]
              case _:
                  args["content"] = [...]

          all_messages.append(args)  # ❌ Creates separate message for each content type
      return all_messages

  Impact

  - API Compatibility: May cause errors or unexpected behavior with OpenAI API
  - Context Accuracy: Incorrect message structure affects conversation history
  - Model Behavior: The model may receive malformed context, leading to degraded responses

  Reproduction Steps

  1. Create a ChatMessage with both text and tool call content
  2. Call _openai_chat_message_parser on the message
  3. Observe that the result contains two separate messages instead of one

  Suggested Fix

  Refactor _openai_chat_message_parser to build a single message dictionary, accumulating all content types before appending. The Anthropic client implementation (agent_framework_anthropic/_chat_client.py:324-377) demonstrates the correct pattern:

  def _openai_chat_message_parser(self, message: ChatMessage) -> list[dict[str, Any]]:
      msg: dict[str, Any] = {
          "role": message.role.value if isinstance(message.role, Role) else message.role,
      }

      for content in message.contents:
          # Skip approval content
          if isinstance(content, (FunctionApprovalRequestContent, FunctionApprovalResponseContent)):
              continue

          match content:
              case FunctionCallContent():
                  if "tool_calls" not in msg:
                      msg["tool_calls"] = []
                  msg["tool_calls"].append(self._openai_content_parser(content))
              case FunctionResultContent():
                  # FunctionResult requires separate message with role="tool"
                  return [{
                      "role": "tool",
                      "tool_call_id": content.call_id,
                      "content": prepare_function_call_results(content.result) if content.result is not None else ""
                  }]
              case _:
                  if "content" not in msg:
                      msg["content"] = []
                  msg["content"].append(self._openai_content_parser(content))

      return [msg] if ("content" in msg or "tool_calls" in msg) else []

  Additional Context

  - Similar issue may exist in _responses_client.py:405-435 (needs verification)
  - Existing test test_chat_response_content_order_text_before_tool_calls verifies the response parsing works correctly, but there's no test for request construction
  - New test case should be added to verify the fix

  Environment

  - Package: agent-framework-core
  - Files affected:
    - python/packages/core/agent_framework/openai/_chat_client.py:361-390
    - python/packages/core/agent_framework/openai/_responses_client.py:405-435 (needs review)

  Checklist for Fix

  - Fix _openai_chat_message_parser in _chat_client.py
  - Review and fix _responses_client.py if needed
  - Add test case for message construction with text + tool_calls
  - Ensure all existing tests pass
  - Verify OpenAI API compatibility

ishanrajsingh:
## Description

Fixes #2410 

The OpenAI chat client's `_openai_chat_message_parser` method was incorrectly splitting messages containing both text content and function/tool calls into separate API messages. According to the [OpenAI API specification](https://platform.openai.com/docs/api-reference/chat/create), a single assistant message can and should contain both `content` and `tool_calls` fields in the same message object.

### Problem
When a `ChatMessage` contained both text content and function calls, the parser created separate messages:
Incorrect output (before fix):
[
{"role": "assistant", "content": ["I'll help you with that calculation."]},
{"role": "assistant", "tool_calls": [{"id": "call-123", "type": "function", ...}]}
]

text

### Solution
Refactored the parser to aggregate all content types into a single message:
Correct output (after fix):
[
{
"role": "assistant",
"content": ["I'll help you with that calculation."],
"tool_calls": [{"id": "call-123", "type": "function", ...}]
}
]

text

### Core Changes
- **`_chat_client.py`**: Refactored `_openai_chat_message_parser` to build a single message dictionary that accumulates both text content and tool calls
- **`_responses_client.py`**: Applied the same fix with additional handling for Responses API-specific formats

### Additional Improvements
- Added exception handling for `FunctionResultContent` to properly format error messages as `"Error: {exception}"`
- Fixed approval content (`FunctionApprovalRequestContent` and `FunctionApprovalResponseContent`) to be top-level items in Responses API as per specification
- Ensured all content types are properly handled without being skipped

## Testing

- ✅ All existing tests pass (126 passed, 46 skipped)
- ✅ Fixed two failing tests:
  - `test_function_result_exception_handling`
  - `test_end_to_end_mcp_approval_flow`
- ✅ Verified message structure matches OpenAI API specification
- ✅ Code formatted with ruff

### API Compatibility
- **Before**: May cause errors or unexpected behavior with OpenAI API due to malformed message structure
- **After**: Fully compliant with OpenAI API specification

### Context Accuracy
- **Before**: Incorrect message structure could affect conversation history and context
- **After**: Proper message structure ensures accurate context propagation

### Model Behavior
- **Before**: Model may receive fragmented context leading to degraded responses
- **After**: Model receives complete, properly formatted context

## Breaking Changes

None. This fix corrects the API format to match the specification without changing public interfaces.

## Related Issues

Closes #2410

## Checklist

- [x] Code follows the project's style guidelines
- [x] All tests pass
- [x] Exception handling added for edge cases
- [x] Changes are backward compatible
- [x] Documentation/comments updated where necessary


python/packages/core/agent_framework/openai/_chat_client.py
```diff
@@ -359,35 +359,47 @@ def _prepare_chat_history_for_request(
...
+                            "role": "tool",
+                            "tool_call_id": content.call_id,
+                            "content": result_content,
+                        }
+                    ]
+                case FunctionApprovalRequestContent() | FunctionApprovalResponseContent():
+                    # These need special handling - add them as content
+                    if "content" not in msg:
+                        msg["content"] = []
+                    msg["content"].append(self._openai_content_parser(content))
```
eavanvalkenburg, line 396: this will add a dict based on our class to the list, I don't think that makes sense, could you 1) add test cases with this and 2) add integration tests that have this. This is used with MCP server side tool calling, so the output format should be determined by openai, not by our type.

PR #2425, opened 2025-11-24
.NET: [WIP] Remove Azure.AI.OpenAI dependency

rogerbarreto:
### Contribution Checklist


stephentoub:
I'd started on this at https://github.com/microsoft/agent-framework/pull/1167, but closed it because there were samples using Assistants, which don't work right now with the /v1 route. If we're willing to get rid of those samples, we should be able to convert everything in the repo.

rogerbarreto:
@stephentoub Will see if we consider just moving all and just keeping the `Assistant` samples dependent on the `Azure.AI.OpenAI` until those are fully discontinued. 

This may already encourage via samples a path away from the package.

rogerbarreto:
Closing for now to be discussed in a later moment.

PR #2476, opened 2025-11-26
.NET: Declarative Agents

markwallace-microsoft:
### Motivation and Context

- Update declarative agent samples
- Add the Microsoft.Agents.AI.Declarative project

### Contribution Checklist


agent-samples/azure/AzureOpenAIAssistants.yaml
```diff
@@ -1,9 +1,9 @@
 kind: Prompt
 name: Assistant
 description: Helpful assistant
-instructions: You are a helpful assistant. You answer questions is the language specified by the user. You return your answers in a JSON format. You must include Assistants as the type in your response.
+instructions: You are a helpful assistant. You answer questions in the language specified by the user. You return your answers in a JSON format. You must include Assistants as the type in your response.
 model:
-    id: =Env.AZURE_OPENAI_DEPLOYMENT_NAME
+    id: gpt-4o-mini
     provider: AzureOpenAI
     apiType: Assistants
```
rogerbarreto, line 8: I see that type is `string` and not `String`.

Are those `apiTypes` part of a enumeration or index of options, I would rather have all those indexable things as lowercase.

```suggestion
    apiType: assistants
```

agent-samples/foundry/FoundryAgent.yaml
```diff
@@ -0,0 +1,22 @@
...
+name: Assistant
+description: Helpful assistant
+instructions: You are a helpful assistant. You answer questions in the language specified by the user. You return your answers in a JSON format.
+model:
+    id: gpt-4.1-mini
+    options:
+        temperature: 0.9
+        topP: 0.95
+    connection:
+        kind: Remote
```
rogerbarreto, line 11: Same as above. For non descriptive/display text, I would stick to lower case identifiers.

```suggestion
        kind: remote
```

dotnet/samples/GettingStarted/Agents/Agent_Step19_Declarative/Program.cs
```diff
@@ -0,0 +1,54 @@
...
+                description: The language of the answer.
+            answer:
+                type: string
+                required: true
+                description: The answer text.
+    """;
+
+// Create the agent from the YAML definition.
+var agentFactory = new ChatClientAgentFactory(chatClient);
+var agent = await agentFactory.CreateFromYamlAsync(text);
```
rogerbarreto: Can we use the new static extensions C# 14 ? 

ChatClientAgent.CreateFromYamlAsync(chatClient, text) ?

```suggestion
var agent = ChatClientAgent.CreateFromYamlAsync(IChatClient chatClient, string yamlDefinition);
```

dotnet/src/Microsoft.Agents.AI.Declarative/ChatClient/ChatClientAgentFactory.cs
```diff
@@ -0,0 +1,54 @@
...
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Agents.AI;
+
+/// <summary>
+/// Provides an <see cref="AgentFactory"/> which creates instances of <see cref="ChatClientAgent"/>.
+/// </summary>
+public sealed class ChatClientAgentFactory : AgentFactory
```
rogerbarreto: This factory is quite generalistic to be in this package. Shouldn't this be in the core instead or otherwise be `internal` and created solely via `extension` ways?

```suggestion
internal sealed class ChatClientAgentFactory : AgentFactory
```

dotnet/src/Microsoft.Agents.AI.Declarative/Extensions/RecordDataTypeExtensions.cs
```diff
@@ -0,0 +1,77 @@
...
+        var json = JsonSerializer.Serialize(schemaObject, ElementSerializer.CreateOptions());
+        return JsonSerializer.Deserialize<JsonElement>(json);
+    }
+#pragma warning restore IL3050 // Calling members annotated with 'RequiresDynamicCodeAttribute' may break functionality when AOT compiling.
+#pragma warning restore IL2026 // Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code
+
+    /// <summary>
+    /// Retrieves the 'schemaName' property from a <see cref="RecordDataType"/>.
+    /// </summary>
+    internal static string? GetSchemaName(this RecordDataType recordDataType)
```
rogerbarreto: If the `internal` methods are not been used outside this class may suggest moving all to `private`.

```suggestion
    private static string? GetSchemaName(this RecordDataType recordDataType)
```

dotnet/src/Microsoft.Agents.AI.Declarative/PromptAgentFactory.cs
```diff
@@ -0,0 +1,62 @@
...
+    /// <param name="promptAgent">Definition of the agent to create.</param>
+    /// <param name="cancellationToken">Optional cancellation token.</param>
+    /// <return>The created <see cref="AIAgent"/>, if null the agent type is not supported.</return>
+    public async Task<AIAgent> CreateAsync(GptComponentMetadata promptAgent, CancellationToken cancellationToken = default)
+    {
+        Throw.IfNull(promptAgent);
+
+        var agent = await this.TryCreateAsync(promptAgent, cancellationToken).ConfigureAwait(false);
+        return agent ?? throw new NotSupportedException($"Agent type {promptAgent.Kind} is not supported.");
+    }
```
rogerbarreto, line 54: Since this factory has a `RecalcEngine` dependency, ideally I would call it `RecalcEngineAgentFactory(...)` or something not so generic, my first thought like the `CCAF (ChatClientAgentFactory)` was to move this to `Core` and add this as an `extension` method for it, asking for an optional `RecalcEngine` during the extension call that if not provided would do the `new RecalcEngine()` and lazy store as one item in a static cached list of engines. `List<object> s_Engines`.

```suggestion
```

agent-samples/azure/AzureOpenAIAssistants.yaml
markwallace-microsoft: I don't control these values, they come from `Microsoft.Bot.ObjectModel.ModelApiType`

agent-samples/foundry/FoundryAgent.yaml
markwallace-microsoft: As above the values come from `Microsoft.Bot.ObjectModel`

dotnet/src/Microsoft.Agents.AI.Declarative/ChatClient/ChatClientAgentFactory.cs
markwallace-microsoft: It cannot be `internal` because then developers cannot use it. If I move this to the core package then I also need to add the following dependencies to the core package.
```
    <PackageReference Include="Microsoft.Bot.ObjectModel" />
    <PackageReference Include="Microsoft.Bot.ObjectModel.Json" />
    <PackageReference Include="Microsoft.Bot.ObjectModel.PowerFx" />
    <PackageReference Include="Microsoft.PowerFx.Interpreter" />
```

dotnet/src/Microsoft.Agents.AI.Declarative/PromptAgentFactory.cs
markwallace-microsoft: I have made the following changes

1. Changed the class to be `PromptAgentFactory`. This will align better with the name changes planed for M.B.OM.
2. Allowed a `RecalcEngine` to be provided.

dotnet/samples/GettingStarted/Agents/Agent_Step19_Declarative/Program.cs
markwallace-microsoft: There will be other factories which also will use this method so the current extension method is the better option.

Issue #1760, opened 2025-10-28
.NET: [BackagroundResponses] ChatClientAgent.RunStreamingAsync updates loss on first run failure


SergeyMenshykh:
When Background Responses are enabled, if ChatClientAgent.RunStreamingAsync fails during the first run, the updates received within that run will never be added to the thread (client-side only; the server-side ones are added by the server). Additionally, the context provider won't receive those updates either. Only updates received in the last successful resume call of RunStreamingAsync will be added to the thread and sent to the context provider. This will leave the thread and context provider in an invalid state because they will only receive updates from the last successful resume call of the RunStreamingAsync method, while updates received in RunStreamingAsync calls before they failed will be lost.

PR #2516, opened 2025-11-28
.NET: Prevent streamed updates loss when resuming streaming

SergeyMenshykh:
This PR disables scenarios that would experience message loss, such as streaming resumption with threads configured with either a custom message store or a context provider.

Closes: https://github.com/microsoft/agent-framework/issues/1760


dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
```diff
@@ -731,6 +739,28 @@ private static Task NotifyMessageStoreOfNewMessagesAsync(ChatClientAgentThread t
...
+        if (continuationToken is null)
+        {
+            return;
+        }
+
+        // Streaming resumption is only supported with threads managed by the agent service because, currently, there's no good solution
+        // to collect updates received in failed runs and pass them to the last successful run so it can store them to the message store.
+        if (safeThread.MessageStore is not null)
+        {
+            throw new InvalidOperationException("Using non-agent service store for chat history with streaming resumption is not supported.");
```
markwallace-microsoft: Should this (and the one below) be a `NotSupportedException`

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
```diff
@@ -731,6 +739,28 @@ private static Task NotifyMessageStoreOfNewMessagesAsync(ChatClientAgentThread t
...
+    private static void ValidateStreamResumptionAllowed(ResponseContinuationToken? continuationToken, ChatClientAgentThread safeThread)
+    {
+        if (continuationToken is null)
+        {
+            return;
+        }
+
+        // Streaming resumption is only supported with threads managed by the agent service because, currently, there's no good solution
+        // to collect updates received in failed runs and pass them to the last successful run so it can store them to the message store.
+        if (safeThread.MessageStore is not null)
```
westey-m: Instead of checking if the message store is not null, should we check that the conversation id is null?
I'm working on the ability to potentially allow per-run passing of message store, in which case this might be null, but it's still a client side thread.

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
```diff
@@ -731,6 +739,28 @@ private static Task NotifyMessageStoreOfNewMessagesAsync(ChatClientAgentThread t
...
 
+    private static void ValidateStreamResumptionAllowed(ResponseContinuationToken? continuationToken, ChatClientAgentThread safeThread)
+    {
+        if (continuationToken is null)
+        {
+            return;
+        }
+
+        // Streaming resumption is only supported with threads managed by the agent service because, currently, there's no good solution
+        // to collect updates received in failed runs and pass them to the last successful run so it can store them to the message store.
```
westey-m:
```suggestion
        // Streaming resumption is only supported with chat history managed by the agent service because, currently, there's no good solution
        // to collect updates received in failed runs and pass them to the last successful run so it can store them to the message store.
```

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
```diff
@@ -731,6 +739,28 @@ private static Task NotifyMessageStoreOfNewMessagesAsync(ChatClientAgentThread t
...
+        if (safeThread.MessageStore is not null)
+        {
+            throw new NotSupportedException("Using non-agent service store for chat history with streaming resumption is not supported.");
```
westey-m:
```suggestion
            throw new NotSupportedException("Streaming resumption is only supported when chat history is stored and managed by the underlying AI service.");
```

dotnet/tests/Microsoft.Agents.AI.UnitTests/ChatClient/ChatClientAgentTests.cs
```diff
@@ -2623,6 +2628,128 @@ await Assert.ThrowsAsync<InvalidOperationException>(async () =>
...
+        mockChatClient.Verify(
+            c => c.GetStreamingResponseAsync(
+                It.IsAny<IEnumerable<ChatMessage>>(),
+                It.IsAny<ChatOptions>(),
+                It.IsAny<CancellationToken>()),
+            Times.Never);
+    }
+
+    [Fact]
+    public async Task RunStreamingAsyncThrowsWhenContinuationTokenUsedWithAIContextProviderAsync()
```
westey-m: I would generally suggest moving groups of tests out into separate files where possible when adding new ones, since this file is simply too big at this stage.

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
SergeyMenshykh: Make sense, just refactored.

dotnet/tests/Microsoft.Agents.AI.UnitTests/ChatClient/ChatClientAgentTests.cs
SergeyMenshykh: Moved all the tests related to background responses in chat client agent to a separate file.

PR #2547, opened 2025-12-01
.NET: Add name-derived MapAGUIServer overloads

### LINKED ISSUE #2179: .NET: Add Support for Passing Agent Name (String) Instead of AIAgent Instance to MapAGUI Method

javiercn:
## Summary

Adds `MapAGUIServer` overloads that derive the endpoint route from the hosted agent name, matching the current AG-UI hosting API and preserving the existing session persistence and isolation pipeline.

Fixes #2179

## Changes

- Adds `MapAGUIServer(IHostedAgentBuilder)`.
- Adds `MapAGUIServer(string agentName)` for keyed agent resolution.
- Adds `MapAGUIServer(AIAgent)`.
- Uses `/{agent.Name}/agui` as the derived route.
- Validates agent names used in derived URL routes.
- Adds focused unit coverage for route derivation, keyed resolution, null inputs, and valid/invalid names.

## Validation

- AG-UI ASP.NET Core unit tests pass on `net8.0`, `net9.0`, and `net10.0`.

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore/AGUIEndpointRouteBuilderExtensions.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore/AGUIEndpointRouteBuilderExtensions.cs
@@ -37,6 +37,21 @@ public static class AGUIEndpointRouteBuilderExtensions
     /// </summary>
     /// <param name="endpoints">The endpoint route builder.</param>
     /// <param name="agentBuilder">The hosted agent builder that identifies the agent registration.</param>
+    /// <returns>An <see cref="IEndpointConventionBuilder"/> for the mapped endpoint.</returns>
+    public static IEndpointConventionBuilder MapAGUIServer(
+        this IEndpointRouteBuilder endpoints,
+        IHostedAgentBuilder agentBuilder)
+    {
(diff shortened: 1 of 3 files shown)


halllo:
I have a similar PR out: https://github.com/microsoft/agent-framework/pull/2343

I dont think selecting agent by name is flexible enough.

chetantoshniwal:
@javiercn  is this still required ?

javiercn:
The hosted/name overload consistency in this PR is useful, but the current implementation still resolves the keyed agent from `endpoints.ServiceProvider` during endpoint mapping. It therefore does not honor scoped/transient `AddAIAgent` registrations.

We have opened #7733 to track the broader requirement across AG-UI, OpenAI Chat Completions, OpenAI Responses, and A2A.

We do not want to evolve the AG-UI hosted overloads independently. The hosted overload behavior should follow the same request-scope and request-time agent-selection pattern across every protocol integration.

PR #2564, opened 2025-12-02
Python: Port to Pydantic BaseModel for Durable State

### LINKED ISSUE #2483: .NET: Python: Timestamp in Durable Entity data is updated with every call
The Python state doesn't always include a createdAt timestamp for tool-call messages in a conversation entry.  (in .NET testing that it consistently did include those properties.) The dashboard should accommodate that, however, as they're not supposed to be required--I'll make a fix.
 
That said, the behavior where the tool call appears "completed" after a subsequent request is due to the Python state suddenly having createdAt properties for old messages when the new state is added.  Also, the createdAt properties of the conversation entries themselves are being changed, serialized with a more recent timestamp. I think this is a deserialization-serialization issue in the Python logic.

larohra:
### Motivation and Context

This PR fixes the serialization issues in the DurableState by removing all custom serialization logic and relying solely on Pydantic BaseModel to provide the serialization/deserialization support. This drastically cleans up the existing code making it much more readable and easier to maintain. 

Fixes #2483 

### Contribution Checklist


python/packages/azurefunctions/agent_framework_azurefunctions/_durable_agent_state.py
```diff
@@ -52,7 +44,10 @@
     get_logger,
 )
 from dateutil import parser as date_parser
+from pydantic import BaseModel, ConfigDict, Field, Tag, field_validator
```
moonbox3, line 47: Hi @larohra, we've deliberately moved away from Pydantic, for this type of behavior, in favor of a custom `SerializationMixin` (see code [here](https://github.com/microsoft/agent-framework/blob/cb343dd7073b07d4fd2540d28b70a395d73038db/python/packages/core/agent_framework/_serialization.py#L133)). It may seem silly to do this, but we've determined that we don't want to support Pydantic, which can have breaking changes even between minor versions, as well as tricky migrations to future major versions.

eavanvalkenburg: This!

larohra: Hey @moonbox3, thanks for sharing your concerns with me. I did look at the `SerializationMixin` but decided against it primarily because it didnt offer all the features that we need (like `snake_case` to `camelCase` serializations). Is the guidance to add all missing features to the `SerializationMixin` for using it?

larohra: I did a deeper dive into it, and here's the feature gap breakdown - 

## Feature Comparison

| Feature | Pydantic (`DurableAgentStateModel`) | `SerializationMixin` | Gap Severity |
| :--- | :--- | :--- | :--- |
| **Serialization** | Automatic `to_dict()` via `model_dump`. | `to_dict()` with deep serialization support. | Low |
| **Deserialization** | Automatic `from_dict()` via `model_validate`. | `from_dict()` with dependency injection support. | Low |
| **Case Conversion** | Automatic via `alias_generator=to_camel`. | **None**. Keys are serialized as-is (snake_case). | **High** |
| **Polymorphism** | Declarative via `Annotated[Union[...], Field(discriminator="type")]`. | **None**. Parent class must manually instantiate correct subclasses. | **High** |
| **Type Discriminator** | Configurable alias (e.g., `$type`). | Defaults to `type` field. | Medium |
| **Data Validation** | Automatic type coercion & custom validators (`field_validator`). | **None**. Must be handled in `__init__`. | Medium |
| **Forward Compatibility** | `extra="ignore"` handles unknown fields automatically. | `from_dict` passes `**kwargs`. `__init__` must explicitly handle/ignore `**kwargs`. | Low |

Apart from the `Low` severity, others are important gaps that do help us keep the code more manageable and cleaner. 

Given this gap list, is it a feasible stop-gap solution to keep the Pydantic BaseModel for now for our use case until the `SerializationMixin` is more updated with these additional features @eavanvalkenburg @moonbox3? Using `SerializationMixin` does not give us major benefit over what we currently have (i.e. customized solution with a lot of boilerplate code) so using `Pydantic` is a much more comprehensive option for us and reduces the scope for errors.

larohra:
Closing the PR since we dont want to add dependency on Pydantic. Will open a separate one with the fix for the issue.

PR #2565, opened 2025-12-02
Python: fix(python): add __str__ to ChatMessage to prevent repr in display output

### LINKED ISSUE #2562: Python: MagenticAgentExecutor produces repr string instead of message text for tool-call-only responses
## Summary

When an agent's response contains only `FunctionCallContent` (tool calls) without text content, the `MagenticAgentExecutor._invoke_agent()` method produces a `ChatMessage` with the Python object repr string (e.g., `<agent_framework._types.ChatMessage object at 0x...>`) as the message text instead of an empty string or formatted tool call info.

## Environment

- **agent-framework**: Latest from PyPI (1.0.0b*)
- **Python**: 3.11
- **Model**: Any model that returns tool calls (tested with HuggingFace Qwen/Qwen2.5-72B-Instruct)
- **Workflow**: MagenticBuilder / MagenticOne orchestration

## Root Cause

**File**: `python/packages/core/agent_framework/_workflows/_magentic.py`
**Line**: ~1799 (in `_invoke_agent` method)

```python
async def _invoke_agent(
    self,
    ctx: WorkflowContext[_MagenticResponseMessage, AgentRunResponse],
) -> ChatMessage:
    # ... agent invocation code ...
    
    if messages and len(messages) > 0:
        last: ChatMessage = messages[-1]
        author = last.author_name or self._agent_id
        role: Role = last.role if last.role else Role.ASSISTANT
        text = last.text or str(last)  # <-- BUG IS HERE
        msg = ChatMessage(role=role, text=text, author_name=author)
        await self._emit_agent_message_event(ctx, msg)
        return msg
```

**The Problem**:

1. `ChatMessage.text` is a property that only extracts text from `TextContent` items in `.contents`
2. When the message contains only `FunctionCallContent` (tool calls), `.text` returns `""`
3. Empty string is falsy, so the `or` expression evaluates `str(last)`
4. `ChatMessage` has no custom `__str__` method, so Python uses the default `__repr__`
5. Result: `text = "<agent_framework._types.ChatMessage object at 0x10c394210>"`

## Impact

1. **Display/Logging**: `MagenticAgentMessageEvent.message.text` contains garbage repr string
2. **Debugging**: Impossible to see what agents actually did in logs or UI
3. **Orchestration**: Manager agent may not properly track progress if it uses event messages
4. **User Experience**: Users see `searcher: <agent_framework._types.ChatMessage object at 0x...>` in UI

## Reproduction Steps

```python
import asyncio
from agent_framework import ChatAgent, MagenticBuilder
from agent_framework._workflows import MagenticAgentMessageEvent

# Define a tool
@ai_function
def search_database(query: str) -> str:
    """Search the database."""
    return f"Found results for: {query}"

# Create agent with tool
agent = ChatAgent(
    name="searcher",
    model_client=your_client,  # Use any client that supports tools
    tools=[search_database],
    instructions="Use the search_database tool to answer questions."
)

# Build workflow
workflow = MagenticBuilder().with_participant(agent).with_manager(...).build()

# Run and observe events
async for event in workflow.run_stream("Search for something"):
    if isinstance(event, MagenticAgentMessageEvent):
        print(f"Agent: {event.agent_id}")
        print(f"Text: {event.message.text}")  # <-- Will show repr string
        print(f"Contents: {event.message.contents}")  # <-- Will show FunctionCallContent
```

## Expected Behavior

When an agent returns a tool-call-only response, the `MagenticAgentMessageEvent.message.text` should be:
- An empty string `""`
- OR a formatted representation of the tool calls (e.g., `"[Tool: search_database(query='test')]"`)

## Suggested Fix

**Option A (Minimal)**: Use empty string instead of repr
```python
# Line ~1799
text = last.text or ""
```

**Option B (Better UX)**: Format tool calls for display
```python
# Line ~1799
text = last.text
if not text and last.contents:
    tool_calls = [c for c in last.contents if isinstance(c, FunctionCallContent)]
    if tool_calls:
        text = ", ".join(f"[Tool: {tc.name}]" for tc in tool_calls)
    else:
        text = ""
```

**Option C (Best)**: Add `__str__` to `ChatMessage` class
```python
# In _types.py, ChatMessage class
def __str__(self) -> str:
    if self.text:
        return self.text
    if self.contents:
        parts = []
        for c in self.contents:
            if isinstance(c, TextContent):
                parts.append(c.text)
            elif isinstance(c, FunctionCallContent):
                parts.append(f"[Tool: {c.name}]")
            elif isinstance(c, FunctionResultContent):
                parts.append(f"[Result: {c.call_id}]")
        return " ".join(parts)
    return ""
```

## Related Issues

- #1366 - Thread corruption with unexecuted tool calls
- #2410 - OpenAI client splits content and tool_calls
- #1288 - Tool calls not shown from workflow.run() (CLOSED)

## Additional Context

This bug is particularly problematic when using custom `ChatClient` implementations (like HuggingFace) that return tool-call-only responses. The OpenAI client may mask this issue if it always includes text with tool calls, but the bug exists in the framework regardless of client.

The-Obstacle-Is-The-Way:
## Summary

When a `ChatMessage` contains only tool calls (no text), `str(message)` returns Python's default repr (`<agent_framework._types.ChatMessage object at 0x...>`) instead of meaningful content.

This PR adds `__str__` and `__repr__` methods to `ChatMessage`:
- `__str__` returns text content when present, or a formatted summary of tool calls
- `__repr__` returns debug-friendly output with role, text preview, and content count

## The Bug

In `MagenticAgentExecutor._invoke_agent()` line 1730:
```python
text = last.text or str(last)
```

When the message has only `FunctionCallContent` (no `TextContent`), `.text` returns empty string, triggering the `str(last)` fallback—which produced the repr garbage.

## The Fix

By adding `__str__` to `ChatMessage`, the existing code works correctly without modification. This also future-proofs any other code that calls `str()` on a `ChatMessage`.

## Test Plan

- Added 4 focused tests covering text, tool-call-only, empty, and repr cases
- All existing tests pass (1145 passed)
- fmt, lint, pyright, mypy all clean

Fixes #2562

### DIFF
--- a/python/packages/core/agent_framework/_types.py
+++ b/python/packages/core/agent_framework/_types.py
@@ -2122,6 +2122,77 @@ class ChatMessage(SerializationMixin):
         """
         return " ".join(content.text for content in self.contents if isinstance(content, TextContent))
 
+    def __str__(self) -> str:
+        """Return a human-readable string representation.
+
+        This method ensures that str(ChatMessage) never returns the default
+        Python repr, which is meaningless to users and causes display bugs.
+
+        When the message contains text content, returns the text.
+        When the message contains only tool calls, returns a formatted summary.
+        Otherwise returns an empty string.
+
+        Returns:
+            A human-readable string representation of the message content.
+        """
+        # First, try the standard text property (TextContent only)
+        # Use strip() to handle edge case of whitespace-only TextContent
+        if self.text.strip():
+            return self.text
+
+        # If no text, format tool calls for readability
+        if self.contents:
+            parts: list[str] = []
+            for content in self.contents:
+                if isinstance(content, TextContent) and content.text:
+                    parts.append(content.text)
+                elif isinstance(content, FunctionCallContent):
+                    # Format: [Calling function_name(...)]
+                    args_preview = ""
+                    if content.arguments:
+                        if isinstance(content.arguments, str):
+                            args_preview = content.arguments[:50]
+                            if len(content.arguments) > 50:
+                                args_preview += "..."
+                        elif isinstance(content.arguments, dict):
+                            args_str = str(content.arguments)
+                            args_preview = args_str[:50]
+                            if len(args_str) > 50:
+                                args_preview += "..."
+                    parts.append(f"[Calling {content.name}({args_preview})]")
+                elif isinstance(content, FunctionResultContent):
+                    # Format: [Result: preview...]
+                    result_str = str(content.result) if content.result is not None else ""
+                    result_preview = result_str[:100]
+                    if len(result_str) > 100:
+                        result_preview += "..."
+                    parts.append(f"[Result: {result_preview}]")
+
+            if parts:
+                return " ".join(parts)
+
+        # Fallback: empty string (NEVER repr)
+        return ""
+
+    def __repr__(self) -> str:
+        """Return a detailed representation for debugging.
+
+        Returns:
+            A string with class name, role, text preview, content count, and author.
+        """
+        text_preview = ""
+        if self.text:
+            text_preview = self.text[:50]
+            if len(self.text) > 50:
+                text_preview += "..."
+
+        return (
+            f"ChatMessage(role={self.role!r}, "
+            f"text={text_preview!r}, "
+            f"contents=[{len(self.contents)} items], "
+            f"author_name={self.author_name!r})"
+        )
+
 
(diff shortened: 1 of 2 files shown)


The-Obstacle-Is-The-Way:

The-Obstacle-Is-The-Way:
Closing in favor of #2566 - thanks for the quick fix @moonbox3! Appreciate the fast turnaround on this.

PR #2588, opened 2025-12-02
Python: Azure AI Search Citation Extraction - V2

giles17:
### Motivation and Context
Adds extraction method to sample

### Contribution Checklist

### DIFF
--- a/python/samples/getting_started/agents/azure_ai/azure_ai_with_azure_ai_search.py
+++ b/python/samples/getting_started/agents/azure_ai/azure_ai_with_azure_ai_search.py
@@ -1,7 +1,9 @@
 # Copyright (c) Microsoft. All rights reserved.
 import asyncio
 import os
+from typing import Any
 
+from agent_framework._types import AgentRunResponse, CitationAnnotation, TextContent
 from agent_framework.azure import AzureAIClient
 from azure.identity.aio import AzureCliCredential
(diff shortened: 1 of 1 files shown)


python/samples/getting_started/agents/azure_ai/azure_ai_with_azure_ai_search.py
```diff
@@ -18,6 +20,38 @@
 """
 
 
+def extract_citations_from_response(response: AgentRunResponse) -> list[dict[str, Any]]:
+    """Extract citation information from an AgentRunResponse."""
+    citations: list[dict[str, Any]] = []
+
+    if hasattr(response, "messages") and response.messages:
```
eavanvalkenburg, line 27: you can skip a bunch of these checks (it's a sample, so we can be a less strict on typing, to improve readability) and inverse the logic for the rest:

```suggestion
    from itertools import chain
    
    if not response.messages:
        return citations
    for content in chain.from_iterable(message.contents for message in response.messages):
        if not content.annotations:
            continue
        # parse the annotations
```
Also why not just return the citations instead of this dict?

eavanvalkenburg:
Is this still current @giles17, or can we close?

giles17:
@eavanvalkenburg yes, I'll close it

PR #2618, opened 2025-12-04
Python: ADR for create/get agent API

dmytrostruk:
### Motivation and Context

There is a misalignment between the create/get agent API in the .NET and Python implementations. This PR contains ADR to compare APIs in both languages to decide what needs to be added.

### Contribution Checklist


docs/decisions/0011-create-get-agent-api.md
```diff
@@ -0,0 +1,91 @@
...
+| Microsoft.Agents.AI.AzureAI.Persistent (V1) | `GetAIAgent` (based on `PersistentAgentsClient` with `PersistentAgent`) | Creates a local instance of `ChatClientAgent`. | Partial (Python uses `create_agent` from `BaseChatClient`). |
+| Microsoft.Agents.AI.AzureAI.Persistent (V1) | `GetAIAgent`/`GetAIAgentAsync` (with `AgentId`) | Fetches `PersistentAgent` via HTTP, then creates a local `ChatClientAgent` instance. | No |
+| Microsoft.Agents.AI.AzureAI.Persistent (V1) | `CreateAIAgent`/`CreateAIAgentAsync` | Creates a remote agent first, then wraps it into a local `ChatClientAgent` instance. | No |
+| Microsoft.Agents.AI.OpenAI | `GetAIAgent` (based on `AssistantClient` with `Assistant`) | Creates a local instance of `ChatClientAgent`. | Partial (Python uses `create_agent` from `BaseChatClient`). |
+| Microsoft.Agents.AI.OpenAI | `GetAIAgent`/`GetAIAgentAsync` (with `AgentId`) | Fetches `Assistant` via HTTP, then creates a local `ChatClientAgent` instance. | No |
+| Microsoft.Agents.AI.OpenAI | `CreateAIAgent`/`CreateAIAgentAsync` (based on `AssistantClient`) | Creates a remote agent first, then wraps it into a local `ChatClientAgent` instance. | No |
```
eavanvalkenburg, line 32: OpenAIAssistantsClient also inherits from BaseChatClient, thus also has `create_agent`

docs/decisions/0011-create-get-agent-api.md
```diff
@@ -0,0 +1,91 @@
...
+```csharp
+var agent1 = new AIProjectClient(...).GetAIAgent(agentSdkInstance); // Creates a local ChatClientAgent instance
+var agent2 = new AIProjectClient(...).GetAIAgent(agentName); // Fetches agent data, creates a local ChatClientAgent instance
+var agent3 = new AIProjectClient(...).CreateAIAgent(...); // Creates a remote agent, returns a local ChatClientAgent instance
+```
+
+Python:
+
+```python
+agent1 = AIProjectClient(...).get_agent(agent_sdk_instance) # Creates a local ChatAgent instance
```
eavanvalkenburg: what exactly is the `agent_sdk_instance` here?

docs/decisions/0011-create-get-agent-api.md
```diff
@@ -0,0 +1,91 @@
...
+agent2 = AIProjectClient(...).get_agent(agent_name) # Fetches agent data, creates a local ChatAgent instance
+agent3 = AIProjectClient(...).create_agent(...) # Creates a remote agent, returns a local ChatAgent instance
```
eavanvalkenburg: This makes sense, especially if `create_agent` becomes a async call, and then we should keep this on all clients, but the behavior should be different, for agents with a agent-like service behind it, is does the work, for agents that don't have that it does the same as `get_agent`

docs/decisions/0011-create-get-agent-api.md
```diff
@@ -0,0 +1,91 @@
...
+- API should be aligned between .NET and Python.
+- API should be intuitive and consistent between backend providers in .NET and Python.
+
+## Considered Options
+
+Add missing implementations on the Python side. This should include the following:
+
+### agent-framework-azure-ai (both V1 and V2)
+
+- Add a `get_agent` method that accepts an underlying SDK agent instance and creates a local instance of `ChatAgent`.
```
eavanvalkenburg, line 82: what about `as_agent`, `get` still implies some kind of network activity

docs/decisions/0011-create-get-agent-api.md
dmytrostruk: Correct, but this `create_agent` currently doesn't actually create a remote assistant, so it behaves more like `GetAIAgent` in .NET.

docs/decisions/0011-create-get-agent-api.md
dmytrostruk: I will update this to be more descriptive, but basically it's an instance of agent that is coming from underlying SDK like `AgentReference` from Azure AI or `Assistant` from OpenAI Assistants.

docs/decisions/0011-create-get-agent-api.md
dmytrostruk: I don't have a strong preference, but `get` will be consistent with .NET naming (at least for now), so that's why I like it more.

docs/decisions/0011-create-get-agent-api.md
```diff
@@ -0,0 +1,235 @@
...
+
+.NET:
+
+```csharp
+var agent1 = new AssistantClient(...).GetAIAgent(agentInstanceFromSdkType); // Creates a local ChatClientAgent instance from OpenAI.Assistants.Assistant
+var agent2 = new AssistantClient(...).GetAIAgent(agentId); // Fetches agent data, creates a local ChatClientAgent instance
+var agent3 = new AssistantClient(...).CreateAIAgent(...); // Creates a remote agent, returns a local ChatClientAgent instance
+```
+
+### Possible Python implementations
```
eavanvalkenburg, line 108: I'm missing the option we did discuss where we do a check when a ChatClient is added to a Agent if that has already happened, seeing all the options below, that seems to me a quite simple and easy to understand for users

docs/decisions/0011-create-get-agent-api.md
```diff
@@ -0,0 +1,235 @@
...
+
+Pros:
+
+- Naturally supports async `create_agent` / `get_agent`.
+- Supports multiple agents per SDK client.
+- Closest conceptual match to .NET extension methods while staying Pythonic.
+
+Cons:
+
+- Discoverability is lower (users need to know where the functions live).
```
brettcannon, line 148: Python developers are used to using functions, so this isn't a concern in general.

docs/decisions/0011-create-get-agent-api.md
```diff
@@ -0,0 +1,235 @@
...
+
+#### Option 2: Factory object
+
+Introduce a dedicated factory type that is constructed from the underlying SDK client, and exposes async `create_agent` / `get_agent` methods.
```
brettcannon: FYI patterns are not a common thing in Python, so the term "factory" isn't going to be descriptive.

docs/decisions/0011-create-get-agent-api.md
```diff
@@ -0,0 +1,235 @@
...
+
+Example:
+
+```python
+from agent_framework.azure import AgentFactory
+
+ai_project_client = AIProjectClient(...)
+factory = AgentFactory(ai_project_client)
```
brettcannon: Is this going to do anything more than this method? If not then it feels like ceremony for the OOP gods.

docs/decisions/0011-create-get-agent-api.md
```diff
@@ -0,0 +1,235 @@
...
+- Discoverable and ergonomic call sites.
+- Mirrors the .NET “methods on the client” feeling.
+
+Cons:
+
+- Many SDK clients are not designed for inheritance; SDK upgrades can break subclasses.
+- Users must opt into subclass everywhere.
+- Typing/initialization can be tricky if the SDK client has non-trivial constructors.
+
+#### Option 4: Monkey patching
```
brettcannon, line 223: "Monkeypatching" it another term for "hack" and unless it's being done in the name of testing you shouldn't do it.

docs/decisions/0011-create-get-agent-api.md
```diff
@@ -0,0 +1,235 @@
...
+# Creates a remote agent first, then returns a local ChatAgent wrapper
+agent = await azure_agents.create_agent(
+    ai_project_client,
+    name="",
+    instructions="",
+    tools=[tool],
+)
+
+# Gets an existing remote agent and returns a local ChatAgent wrapper
+agent = await azure_agents.get_agent(ai_project_client, agent_id=agent_id)
```
eavanvalkenburg: this won't work, you can't have two methods with the same name, one sync one async, we would have to differentiate the name (could be as single as `aget_agent`, but still)

docs/decisions/0011-create-get-agent-api.md
eavanvalkenburg, line 148: my main concern with this option is that we might end up with a confusing constellation of functions, so either we make a generic function that works across multiple/all ChatClients or we need really descriptive names.

docs/decisions/0011-create-get-agent-api.md
dmytrostruk: I think it's separate and should be implemented independently from ADR and only for stateful chat clients, since chat client can be still provided to ChatAgent when invoking `ChatAgent` constructor manually. While this PR is more focused on `create_agent` API.

docs/decisions/0011-create-get-agent-api.md
dmytrostruk: At least for now, I don't think there will be more use cases except the one listed in the document.

docs/decisions/0011-create-get-agent-api.md
dmytrostruk: I think it would be hard to implement a generic function that works across multiple/all ChatClients taking into account that each function will accept a different set of parameters, but taking into account they will come from different packages, they can still have the same `create_agent` name, which should work as a common pattern to create agents.

docs/decisions/0011-create-get-agent-api.md
dmytrostruk: Makes sense, I think we can decide on the name as soon as we decide on the general approach first.

docs/decisions/0011-create-get-agent-api.md
```diff
@@ -0,0 +1,235 @@
...
+- Discoverability is lower (users need to know where the functions live).
+
+#### Option 2: Factory object
+
+Introduce a dedicated factory type that is constructed from the underlying SDK client, and exposes async `create_agent` / `get_agent` methods.
+
+Example:
+
+```python
+from agent_framework.azure import AgentFactory
```
markwallace-microsoft: How will this work if I want to create different types of persistent agent at the same time? We would need unique `AgentFractory` names, right?

docs/decisions/0011-create-get-agent-api.md
```diff
@@ -0,0 +1,235 @@
...
+
+ai_project_client = AIProjectClient(...)
+
+# Creates a remote agent first, then returns a local ChatAgent wrapper
+agent = await azure_agents.create_agent(
+    ai_project_client,
+    name="",
+    instructions="",
+    tools=[tool],
+)
```
markwallace-microsoft, line 131: This seems closest to how extension methods work, do you agree?

docs/decisions/0011-create-get-agent-api.md
```diff
@@ -0,0 +1,235 @@
...
+
+Cons:
+
+- Fragile across SDK updates and difficult to type-check.
+- Surprising behavior (global side effects), potential conflicts across packages.
+- Harder to support/debug, especially in larger apps and test suites.
+
+## Decision Outcome
+
+TBD
```
markwallace-microsoft: Do you have a recommendation?

This would be a good topic to discuss with Brett and Johan. Could you book a meeting with them as part 2 of the API review and review this decision.

docs/decisions/0011-create-get-agent-api.md
dmytrostruk: Yes, I agree.

docs/decisions/0011-create-get-agent-api.md
```diff
@@ -0,0 +1,268 @@
...
+
+Cons:
+
+- Fragile across SDK updates and difficult to type-check.
+- Surprising behavior (global side effects), potential conflicts across packages.
+- Harder to support/debug, especially in larger apps and test suites.
+
+## Decision Outcome
+
+Implement `get_agent`/`create_agent` API via `Option 1: Module-level functions`:
```
eavanvalkenburg: agreed, but not sure on the naming, as this is a function specific to Foundry (even v2?) agents, this name might lead to new confusion. I'm now thinking if we remove the `create_agent` method generally. and instead have a async classmethod on the ChatAgent then we can make this work generally, my thinking is:

1. add `async ChatAgent.create_agent` method, with the same params as the regular `init` of a ChatAgent
2. add a async `create_hosted_agent` method on chat clients (maybe define a protocol to make this more deterministic, and within the `ChatAgent.create_agent` method we call this if it exists, this also ensures we have a `name` and `description` on the agent level before we try to create the hosted agent.
3. `create_hosted_agent` takes all the arguments from the agent creation and uses them to set appropriate fields on the chatclient (and raise if there is a missing one) and then wraps the chat client in a new `ChatAgent` instance.
4. Remove the existing mechanism that syncs the name etc. to the chat client.

what do you think?

dmytrostruk: Based on internal discussions, agreed to proceed with specific methods per provider with names `create_hosted_agent` and `get_hosted_agent`.

eavanvalkenburg (approved):

docs/decisions/0011-create-get-agent-api.md
```diff
@@ -0,0 +1,268 @@
...
+
+Cons:
+
+- Fragile across SDK updates and difficult to type-check.
+- Surprising behavior (global side effects), potential conflicts across packages.
+- Harder to support/debug, especially in larger apps and test suites.
+
+## Decision Outcome
+
+Implement `get_hosted_agent`/`create_hosted_agent` API via `Option 1: Module-level functions`:
```
markwallace-microsoft: I'm not sure about these names. A hosted agents in Foundry are different from Foundry Agent Service agents, see: https://learn.microsoft.com/en-us/azure/ai-foundry/agents/concepts/hosted-agents?view=foundry&tabs=cli

Is the intent that these methods would work for Foundry V2, Foundry V1 and Foundry Hosted agents?

docs/decisions/0011-create-get-agent-api.md
```diff
@@ -0,0 +1,268 @@
...
+
+```python
+from agent_framework.azure import create_hosted_agent, get_hosted_agent
```
markwallace-microsoft: This is somewhat confusing e.g. I think this creates a Foundry V2 agent so using azure as the package name isn't clear.

Consider the case where I want to use a Foundry Agent Service agent and an Azure OpenAI Assistant agent and an Azure OpenAI ChatCompletion agent at the same time. 

- Do I just need one import and will the function behave differently based on client type that is passed in?
- How do I create a non hosted agent?

Can you do the following:

- Provide snippets for every agent type we currently support 
- Explain how new agent types will be added in future

docs/decisions/0011-create-get-agent-api.md
dmytrostruk: `Hosted` part is removed.

docs/decisions/0011-create-get-agent-api.md
dmytrostruk: Updated Decision Outcome section and provided code snippets.

giles17 (approved):

Issue #2542, opened 2025-12-01
.NET: Single Agent API review


SergeyMenshykh:
- **Microsoft.Agents.AI.Abstractions**
  - [ ] AIAgent.DisplayName
    The name of this property seems incorrect, since it's really just a convenience fallback mechanism to ensure something is always returned, with name taking precedence.
  - [ ] AIAgent.GetNewThread & Agent.DeserializeThread
    - [ ] Should these be async, for the case where a user wants to generate an id using a service in the factory method for the ChatMessageStore. Would also need a cancellation token in that case.
    - [ ] DeserializeThread param name serializedThread alternatives: threadElement, stateElement, etc. Rename the `AIContextProviderFactoryContext.SerializedState` and `ChatMessageStoreFactoryContext.SerializedState` properties as well if needed. Search codebase for `serializedState` params and rename them if necessary.
  - [ ] AgentThread.Serialize & ChatMessageStore.Serialize
    We should consider separating thread state from behavior. I.e. the agent thread is just a state object model, and behaviors are bound to objects in the tree. The AgentThread may contain behavior instances, but they can be recreated at any time. Problem is that we don't know the types of the ChatMessageStore state.  We need to do polymorphic deserialization for ChatMessageStore and AIContextProvider. Problem with current approach is we only support json, and users cannot serialize the thread using other means. Also, users expect to be able to just use JsonSerializer.
  - [ ] AIAgentMetadata & AgentThreadMetadata
    - [ ] Should these be sealed?
    - [ ] We are not using these everywhere we should, see A2AAgent for AgentMetadata, and most threads for thread metadata.
    - [ ] Why do these not have concrete properties on the abstraction, and instead are only exposed via GetService?
  - [ ] AgentRunResponse & AgentRunResponseUpdate
    - [ ] Should AgentId be required?
    - [ ] Should ResponseId be required?
    - [ ] Should CreatedAt be required?
    - [ ] JSO should be optional in Deserialize/TryDeserialize.
  - InvokingContext & InvokedContext
    - [ ] Should these be sealed?
  - InMemoryAgentThread & InMemoryChatMessageStore
    - [ ] Do they belong to Microsoft.Agents.AI.Abstractions project?

- **Microsoft.Agents.AI**
  - [ ] Either seal `ChatClientAgentOptions` or add copy constructor.
[...]

PR #2627, opened 2025-12-04
.NET: [BREAKING] Change namespaces of the Microsoft.Agents.AI.OpenAI classes 

SergeyMenshykh:
### Motivation and Context  
As part of the AF GA initiative, this PR changes the namespaces of the classes in the `Microsoft.Agents.AI.OpenAI` package. The 

`OpenAIChatClientAgent` and `OpenAIResponseClientAgent` classes are moved from the `OpenAI` namespace to the `Microsoft.Agents.AI` namespace, while all the extension classes are moved to the namespaces of the types they are extending.

Contributes to: https://github.com/microsoft/agent-framework/issues/2542

### DIFF
--- a/dotnet/tests/Microsoft.Agents.AI.OpenAI.UnitTests/Extensions/AIAgentWithOpenAIExtensionsTests.cs
+++ b/dotnet/tests/Microsoft.Agents.AI.OpenAI.UnitTests/Extensions/AIAgentWithOpenAIExtensionsTests.cs
@@ -7,8 +7,8 @@ using System.Threading;
 using System.Threading.Tasks;
 using Moq;
 using ChatMessage = Microsoft.Extensions.AI.ChatMessage;
-using OpenAIChatMessage = OpenAI.Chat.ChatMessage;
 using ChatRole = Microsoft.Extensions.AI.ChatRole;
+using OpenAIChatMessage = OpenAI.Chat.ChatMessage;
 using TextContent = Microsoft.Extensions.AI.TextContent;
 
 namespace Microsoft.Agents.AI.OpenAI.UnitTests.Extensions;


dotnet/src/Microsoft.Agents.AI.OpenAI/OpenAIChatClientAgent.cs
```diff
@@ -1,13 +1,12 @@
...
 
-using Microsoft.Agents.AI;
 using Microsoft.Extensions.AI;
 using Microsoft.Extensions.Logging;
 using Microsoft.Shared.Diagnostics;
 using OpenAI.Chat;
 using ChatMessage = OpenAI.Chat.ChatMessage;
 
-namespace OpenAI;
+namespace Microsoft.Agents.AI;
```
markwallace-microsoft: The reason this was put in the `OpenAI` namespace was so that developers who are just using OpenAI and it's primitives don't need a `using Microsoft.Agents.AI`. This change means OpenAI developers will see the MEAI and OpenAI primitives are there are conflicting names.

dotnet/src/Microsoft.Agents.AI.OpenAI/OpenAIResponseClientAgent.cs
```diff
@@ -1,13 +1,12 @@
...
 
 using System.Runtime.CompilerServices;
-using Microsoft.Agents.AI;
 using Microsoft.Extensions.AI;
 using Microsoft.Extensions.Logging;
 using Microsoft.Shared.Diagnostics;
 using OpenAI.Responses;
 
-namespace OpenAI;
+namespace Microsoft.Agents.AI;
```
markwallace-microsoft: The reason this was put in the `OpenAI` namespace was so that developers who are just using OpenAI and it's primitives don't need a `using Microsoft.Agents.AI`. This change means OpenAI developers will see the MEAI and OpenAI primitives are there are conflicting names.

dotnet/src/Microsoft.Agents.AI.OpenAI/Extensions/AIAgentWithOpenAIExtensions.cs
```diff
@@ -1,12 +1,11 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 
 using System.ClientModel;
-using Microsoft.Agents.AI;
 using Microsoft.Agents.AI.OpenAI;
 using Microsoft.Shared.Diagnostics;
 using OpenAI.Chat;
 
-namespace OpenAI;
+namespace Microsoft.Agents.AI;
```
markwallace-microsoft, line 9: One problem with this is if I an not using OpenAI I am going to see these extension methods and they may confuse developers. The intention here was that there are developers who want to use the OpenAI primitives and not the MEAI primitives and we want to make sure they don't have to deal with conflicts.

dotnet/src/Microsoft.Agents.AI.OpenAI/OpenAIChatClientAgent.cs
rogerbarreto: > The reason this was put in the OpenAI namespace was so that developers who are just using OpenAI and it's primitives don't need a using Microsoft.Agents.AI. This change means OpenAI developers will see the MEAI and OpenAI primitives are there are conflicting names.

Do we need to keep using those?

Can't we just make extensions from RunAsync and RunStreamingAsync that gives back the `OpenAI` type, do we need to have a specific `public` class for it?

dotnet/src/Microsoft.Agents.AI.OpenAI/OpenAIResponseClientAgent.cs
SergeyMenshykh: Discussed offline: no concerns here.

dotnet/src/Microsoft.Agents.AI.OpenAI/Extensions/AIAgentWithOpenAIExtensions.cs
SergeyMenshykh: Discussed offline: no concerns here.

dotnet/src/Microsoft.Agents.AI.OpenAI/OpenAIChatClientAgent.cs
SergeyMenshykh: Discussed offline: no concerns here for the first comment, and to address the second comment the classes will be moved to samples.

PR #2632, opened 2025-12-04
.NET: fix forwarding conversationId to IChatClient in OpenAI Responses

DeagleGross:
In https://github.com/microsoft/agent-framework/discussions/2477 the problem was raised: client's conversationId does not map to what `IChatClient` expects in the options. 

We should not forward the client's conversationId into `IChatClient`.

### Contribution Checklist

### DIFF
--- a/dotnet/agent-framework-dotnet.slnx
+++ b/dotnet/agent-framework-dotnet.slnx
@@ -104,6 +104,8 @@
     <File Path="samples/GettingStarted/AgentWithOpenAI/README.md" />
     <Project Path="samples/GettingStarted/AgentWithOpenAI/Agent_OpenAI_Step01_Running/Agent_OpenAI_Step01_Running.csproj" />
     <Project Path="samples/GettingStarted/AgentWithOpenAI/Agent_OpenAI_Step02_Reasoning/Agent_OpenAI_Step02_Reasoning.csproj" />
+    <Project Path="samples/GettingStarted/AgentWithOpenAI/Agent_OpenAI_Step03_CreateFromChatClient/Agent_OpenAI_Step03_CreateFromChatClient.csproj" />
+    <Project Path="samples/GettingStarted/AgentWithOpenAI/Agent_OpenAI_Step04_CreateFromOpenAIResponseClient/Agent_OpenAI_Step04_CreateFromOpenAIResponseClient.csproj" />
   </Folder>
   <Folder Name="/Samples/Purview/" />
   <Folder Name="/Samples/Purview/AgentWithPurview/">
--- a/dotnet/samples/GettingStarted/AgentWithOpenAI/Agent_OpenAI_Step03_CreateFromChatClient/Agent_OpenAI_Step03_CreateFromChatClient.csproj
+++ b/dotnet/samples/GettingStarted/AgentWithOpenAI/Agent_OpenAI_Step03_CreateFromChatClient/Agent_OpenAI_Step03_CreateFromChatClient.csproj
@@ -1,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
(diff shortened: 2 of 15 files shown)


dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/AIAgentResponseExecutor.cs
```diff
@@ -36,7 +36,13 @@ public async IAsyncEnumerable<StreamingResponseEvent> ExecuteAsync(
...
         var chatOptions = new ChatOptions
         {
-            ConversationId = request.Conversation?.Id,
+            // Note: We intentionally do NOT set ConversationId on ChatOptions here.
+            // The conversation ID from the client request is used by the hosting layer
+            // to manage conversation storage, but should not be forwarded to the underlying
+            // IChatClient as it has its own concept of conversations (or none at all).
+            // ---
+            // ConversationId = request.Conversation?.Id,
+
```
stephentoub, line 45: What about in the other direction... do the output events include a conversation Id that the client will interpret as coming from the hosting layer?

DeagleGross: Yes, output event [`response.created`](https://platform.openai.com/docs/api-reference/responses-streaming/response/created) contains conversationId.

What are your thoughts? To pass it into `IChatClient`, but instead not return the conversationId at all to the caller?

stephentoub: But are those events from the IChatClient or the hosting layer?

DeagleGross: from the hosting layer.

stephentoub: That's fine, then, assuming any conversation id in those events is a conversation id from the hosting layer and not from the IChatClient.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/HostedAgentResponseExecutor.cs
```diff
@@ -82,9 +82,16 @@ public async IAsyncEnumerable<StreamingResponseEvent> ExecuteAsync(
     {
```
stephentoub, line 82: Can you remind me what the difference is between HostedAgentResponseExecutor and AIAgentResponseExecutor?

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/AIAgentResponseExecutor.cs
DeagleGross: Yep, I commented out the code where we set the `ConversationId` from the request from the client. It is not from the `IChatClient`. And the events that we return from the hosting layer to the client we fill with the input-conversationId from incoming request.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/HostedAgentResponseExecutor.cs
DeagleGross: `AIAgentResponseExecutor` wraps the `AIAgent` and uses a specific `AIAgent` for request processing. 

`HostedAgentResponseExecutor` instead keeps a reference to `IServiceProvider` and resolves `AIAgent` per-request based on `request.Agent?.Name`

PR #2646, opened 2025-12-04
.NET: fix: RunState is sometimes Running when it should be Idle

lokitoth:
### Motivation and Context

There is a timeout-wait for messages in the inner loop driving the running of the workflow. When it gets cleared, an attempt to run a superstep is made. Currently we set the state to running as soon as the wait clears, even if we have no work to do, and clear it upon discovering this.

This causes a sporadic UnitTest failure in Workflows.

### Description

The fix is to only set the Running state when actual Workflow execution is happening when running Super-Steps, and not until the loop confirms there is work to do.

TODO: A broader-term fix would be to remove the Semaphore and timeout-wait in it, since it is working around the inability to atomically insert and release the Semaphore driving the wait. The timeout provides a race condition where the workflow was "ghosting" a Running status. See #2647

### Contribution Checklist


lokitoth:
Closing in favour of better fix post Timeout removal

Issue #2615, opened 2025-12-03
.NET & Python: [Durable Agents] Automatic thread/session cleanup TTL


cgillum:
The Durable Task extension for Agent Framework ("Durable Agents") provides built-in thread storage. However, there's no clear way to cleanup the thread storage data, which will accumulate in either the Azure Storage account or in the Durable Task Scheduler resource owned by the user.

This issue tracks a proposal to create configurable TTLs on durable agents that allow them to be "garbage collected" if there aren't any interactions after a period of time. For example, if a particular agent session is idle for 30 days, the underlying entity state (e.g., the conversation history) gets deleted. Each new interaction with the agent would reset the TTL. Any messages to an "expired" agent session will result in a new conversation history.

This is needed for each language we support:
- [ ] .NET
- [ ] Python

eavanvalkenburg:
is this still relevant @gavin-aguiar @larohra @anthonychu @cgillum

cgillum:
Yes I believe this is still relevant. @gavin-aguiar are you still the right owner or should it be @ahmedmuhsin?

gavin-aguiar:
@ahmedmuhsin  will work on this. Assigning it to him

larohra:
Feature not available in AFD package, dependency on the new durabletask package. More info here - [Entity to Entity Signaling in AFD vs new package](https://gist.github.com/larohra/6485949df0edbf1dea7280a76271028b)

Teams conversation - https://teams.microsoft.com/l/message/19:meeting_ODFiNGE3NWYtMDYwZC00NTExLTlmYzEtNTdkNWVmODM5ODVj@thread.v2/1783986889670?context=%7B%22contextType%22%3A%22chat%22%7D

PR #2669, opened 2025-12-05
.NET: [Durable Agents] Update CHANGELOG with release notes for past releases

cgillum:
Backfills the CHANGELOG.md file with the last several updates for the DurableTask extension.

No code changes.

PR #2679, opened 2025-12-06
.NET: Add TTLs to durable agent sessions

cgillum:
### Motivation and Context

The Durable Task extension for Agent Framework ("Durable Agents") provides built-in thread storage. However, there's no clear way to cleanup the thread storage data, which will accumulate in either the Azure Storage account or in the Durable Task Scheduler resource owned by the user.

Applies to #2615

### Description

This PR introduces automatic session cleanup using TTLs, implemented using scheduled entity signals. See the documentation file in the PR for more details.

### Contribution Checklist


dotnet/src/Microsoft.Agents.AI.DurableTask/AgentEntity.cs
```diff
@@ -121,4 +136,76 @@ async IAsyncEnumerable<AgentRunResponseUpdate> StreamResultsAsync()
...
+            : currentTime.Add(minimumDelay);
+
+        logger.LogTTLDeletionScheduled(sessionId, scheduledTime);
+
+        // Schedule a signal to self to check for expiration
+        this.Context.SignalEntity(
+            this.Context.Id,
+            nameof(CheckAndDeleteIfExpiredAsync),
+            options: new SignalEntityOptions { SignalTime = scheduledTime });
+    }
```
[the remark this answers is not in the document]
cgillum: Fixed in latest iteration.

dotnet/src/Microsoft.Agents.AI.DurableTask/DurableAgentsOptions.cs
```diff
@@ -9,6 +9,25 @@ public sealed class DurableAgentsOptions
...
+    private readonly Dictionary<string, TimeSpan?> _agentTimeToLive = new(StringComparer.OrdinalIgnoreCase);
+
+    /// <summary>
+    /// Gets or sets the default time-to-live (TTL) for agent entities.
+    /// </summary>
+    /// <remarks>
+    /// If an agent entity is idle for this duration, it will be automatically deleted.
+    /// Defaults to 30 days. Set to <see langword="null"/> to disable TTL for agents without explicit TTL configuration.
+    /// </remarks>
+    public TimeSpan? DefaultTimeToLive { get; set; } = TimeSpan.FromDays(30);
```
kshyju: Is setting this default value cause any breaking behavior for customers who already started an agent session and entities are present, when they redeploy with this version of the package?  I see "GetTimeToLive" method falls back to this value.

dotnet/src/Microsoft.Agents.AI.DurableTask/Logs.cs
```diff
@@ -46,4 +46,50 @@ public static partial void LogAgentResponse(
         Level = LogLevel.Information,
         Message = "Found response for agent with session ID '{SessionId}' with correlation ID '{CorrelationId}'")]
     public static partial void LogDonePollingForResponse(this ILogger logger, AgentSessionId sessionId, string correlationId);
+
+    [LoggerMessage(
+        EventId = 6,
```
kshyju, line 51: Unrelated to this change. But how do we ensure any other agent extensions (or code in agent core components in future) is not using the same EventId? Should we align with the team on a dedicated range or series of EventId values for our extension?

dotnet/src/Microsoft.Agents.AI.DurableTask/State/DurableAgentStateData.cs
```diff
@@ -17,6 +17,13 @@ internal sealed class DurableAgentStateData
     [JsonPropertyName("conversationHistory")]
     public IList<DurableAgentStateEntry> ConversationHistory { get; init; } = [];
 
+    /// <summary>
+    /// Gets or sets the expiration time for this agent entity.
```
kshyju: Should we explicitly specify "in UTC" for clarity?

dotnet/src/Microsoft.Agents.AI.DurableTask/Logs.cs
cgillum: Is it important for events across multiple providers to be unique? I assumed that it's nice to have but not required. Let me know if you think otherwise.

dotnet/src/Microsoft.Agents.AI.DurableTask/DurableAgentsOptions.cs
cgillum: Technically yes, the next time existing agent entities are touched, we'll set a TTL on them of 30 days after they upgrade to this version of the library.

dotnet/src/Microsoft.Agents.AI.DurableTask/Logs.cs
kshyju: yea, agree.

dotnet/samples/AzureFunctions/01_SingleAgent/Program.cs
```diff
@@ -32,6 +32,6 @@
 using IHost app = FunctionsApplication
     .CreateBuilder(args)
     .ConfigureFunctionsWebApplication()
-    .ConfigureDurableAgents(options => options.AddAIAgent(agent))
+    .ConfigureDurableAgents(options => options.AddAIAgent(agent, timeToLive: TimeSpan.FromHours(1)))
```
larohra, line 35: Shouldn't we add to all samples?

kshyju (approved):
Should we add an entry for this to changelogs file in this PR itself?

dotnet/src/Microsoft.Agents.AI.DurableTask/AgentEntity.cs
```diff
@@ -113,6 +107,27 @@ async IAsyncEnumerable<AgentRunResponseUpdate> StreamResultsAsync()
                     response.Usage?.TotalTokenCount);
             }
 
+            // Update TTL expiration time. Only schedule deletion check on first interaction.
+            // Subsequent interactions just update the expiration time; CheckAndDeleteIfExpiredAsync
+            // will reschedule the deletion check when it runs.
+            TimeSpan? timeToLive = this._options.GetTimeToLive(sessionId.Name);
+            if (timeToLive.HasValue)
```
larohra, line 115: Should we reset the TTL if the customer sets TTL first and then removes it (by setting it to null)?

Eg: 
User sends message at Day 0 → agent session created, expiration set to Day 30
User removes TTL at Day 1 by setting it to Null

Day 30 -> `CheckAndDeleteIfExpired` is called and the Entity will be Deleted because the `this.State.Data.ExpirationTime` is not reset

docs/features/durable-agents/durable-agents-ttl.md
```diff
@@ -0,0 +1,150 @@
...
+
+The following sections describe how TTL works in detail.
+
+### Expiration tracking
+
+Each agent session maintains an expiration timestamp in its internally managed state that is updated whenever the session processes a message:
+
+1. When a message is sent to an agent session, the expiration time is set to `current time + TTL`
+2. The runtime schedules a delete operation for the expiration time (subject to minimum delay constraints)
+3. When the delete operation runs, if the current time is past the expiration time, the session state is deleted. Otherwise, the delete operation is rescheduled for the next expiration time.
```
larohra, line 88: In this model, the entity deletion might not happen when its intended to if `MinimumTimeToLiveSignalDelay` is high which might throw the customers off.

for eg:

User sends message at Day 0 → agent session created, expiration set to Day 30 
User sends message at Day 15 → Expiration reset to Day 45 

•	Day 30: Check runs
•	expirationTime = Day 45 (from state)
•	currentTime.Add(minimumDelay) = Day 30 + 16 days = Day 46
•	Since Day 45 < Day 46, it schedules for Day 46 instead of Day 45

Maybe we can just keep a hard-coded small(ish) value for the Delay under the hood and not expose it to the cx to avoid confusion?

dotnet/samples/AzureFunctions/01_SingleAgent/Program.cs
larohra: nvm just noticed the default to 30 days

cgillum:
> Should we add an entry for this to changelogs file in this PR itself?

@kshyju yes, I'll add that after my changelog PR (#2669) gets merged, at which point hopefully Copilot will also remind me. :)

docs/features/durable-agents/durable-agents-ttl.md
cgillum: It's a good point that this could turn into some kind of footgun. However, there are two cases where I think we may need this:
1. For testing, we need to be able to disable this or set it to a low value, like 1 second, so that tests complete quickly.
2. There may be some high-throughput workflow orchestration scenarios where you want agent state to be very short-lived, in which case it would be beneficial to keep this value low.

To address your concern, how about we limit the maximum value to 5 minutes?

larohra: Yeah 5 mins sounds reasonable to me, thanks for addressing it!

dotnet/src/Microsoft.Agents.AI.DurableTask/AgentEntity.cs
cgillum: Ah, yeah, I guess there is an inconsistency here. I'll push a fix.

docs/features/durable-agents/durable-agents-ttl.md
cgillum: I've updated the implementation to enforce a max of 5 minutes.

dotnet/src/Microsoft.Agents.AI.DurableTask/State/DurableAgentStateData.cs
cgillum: Good call. I updated the property names as well.

dotnet/src/Microsoft.Agents.AI.DurableTask/AgentEntity.cs
```diff
@@ -16,25 +16,19 @@ internal class AgentEntity(IServiceProvider services, CancellationToken cancella
...
-
-        AIAgent agent = agentFactory(this._services);
+        AIAgent agent = this.GetAgent(sessionId);
         EntityAgentWrapper agentWrapper = new(agent, this.Context, request, this._services);
 
         // Logger category is Microsoft.DurableTask.Agents.{agentName}.{sessionId}
-        ILogger logger = this._loggerFactory.CreateLogger($"Microsoft.DurableTask.Agents.{agent.Name}.{sessionId.Key}");
+        ILogger logger = this.GetLogger(agent.Name!, sessionId.Key);
 
         if (request.Messages.Count == 0)
```
gavin-aguiar, line 33: If the message count is 0, we are still processing the TTL? or should we just return an empty response at this point?

dotnet/src/Microsoft.Agents.AI.DurableTask/AgentEntity.cs
```diff
@@ -113,6 +107,36 @@ async IAsyncEnumerable<AgentRunResponseUpdate> StreamResultsAsync()
                     response.Usage?.TotalTokenCount);
             }
 
+            // Update TTL expiration time. Only schedule deletion check on first interaction.
+            // Subsequent interactions just update the expiration time; CheckAndDeleteIfExpiredAsync
+            // will reschedule the deletion check when it runs.
+            TimeSpan? timeToLive = this._options.GetTimeToLive(sessionId.Name);
+            if (timeToLive.HasValue)
+            {
+                DateTime newExpirationTime = DateTime.UtcNow.Add(timeToLive.Value);
```
gavin-aguiar, line 117: Should this be set before we get a response? Right now the expiration time is based on when the run finishes, not when the request is received. If the response takes an hour for long running tools and TTL is 1hr, the session lives  up to about 2hrs from the users message

cgillum: I'm fine with the current behavior. Regarding your example, I think it would be worse if the TTL started when the request is received. If the TTL is 1 hour and the response takes 1 hour to generate, then the session would be deleted immediately, before the user can attempt to run the agent again.

dotnet/src/Microsoft.Agents.AI.DurableTask/AgentEntity.cs
cgillum: Yeah, looks like this got missed or regressed somehow. Added a return statement with an empty response.

gavin-aguiar (approved):

gavin-aguiar (approved):

Issue #1762, opened 2025-10-28
.NET: [BackagroundResponses] ChatClientAgent.RunStreamingAsync input messages loss on first run failure


SergeyMenshykh:
When Background Responses are enabled, if ChatClientAgent.RunStreamingAsync fails during the first run, the input messages provided to the run will never be added to the thread (client-side only; the server-side ones are added by the server). The context provider won't receive those updates either. This will leave the thread and context provider in an invalid state.

Related: https://github.com/microsoft/agent-framework/issues/1760

PR #2604, opened 2025-12-03
.NET: [BREAKING] Refactor ChatMessageStore methods to be similar to AIContextProvider and add filtering support

westey-m:
### Motivation and Context

#2518 
#2054

### Description

- Refactor ChatMessageStore interface to be more similar to AIContextProvider, to give more certainty as to the data being passed to it, and the circumstances under which that data is passed.
- Add a ChatMessageStore decorator that makes it easy to filter the messages being passed to the ChatMessageStore.  This allows us to easily avoid persisting messages from AIContextProviders in the chat history.

```csharp
AIAgent agent = azureOpenAIClient
    .GetChatClient(deploymentName)
    .CreateAIAgent(new ChatClientAgentOptions
    {
        ChatOptions = new() { Instructions = "You are a helpful support specialist for Contoso Outdoors. Answer questions using the provided context and cite the source document when available." },
        AIContextProviderFactory = ctx => new TextSearchProvider(SearchAdapter, ctx.SerializedState, ctx.JsonSerializerOptions, textSearchOptions),
        // Since we are using ChatCompletion which stores chat history locally, we can also add a message removal policy
        // that removes messages produced by the TextSearchProvider before they are added to the chat history, so that
        // we don't bloat chat history with all the search result messages.
        ChatMessageStoreFactory = ctx => new InMemoryChatMessageStore(ctx.SerializedState, ctx.JsonSerializerOptions)
            .WithAIContextProviderMessageRemoval(),
    });
```

### Contribution Checklist

- [ ] The code builds clean without any errors or warnings
- [ ] The PR follows the [Contribution Guidelines](https://github.com/microsoft/agent-framework/blob/main/CONTRIBUTING.md)
- [ ] All unit tests pass, and I have added new tests where possible
- [ ] **Is this a breaking change?** If yes, add "[BREAKING]" prefix to the title of the PR.

PR #2748, opened 2025-12-10
.NET: [Breaking] Prevent loss of input messages & streamed updates when resuming streaming

SergeyMenshykh:
**Context:**
- https://github.com/microsoft/agent-framework/issues/1762
- https://github.com/microsoft/agent-framework/issues/1760

**Description:**
This PR adds the `ChatClientAgentContinuationToken` token type for `ChatClientAgent`. This token type retains the agent's input messages provided during the initial call of the `RunStreamingAsync` method, as well as the streamed updates received through initial, and all subsequent stream-resumption calls of the method. `ChatClientAgent` wraps all continuation tokens received from the chat client into the new token type, and it adds both input messages and streamed updates to the token before returning it as part of the response update to the caller. Upon receiving the token back from the caller via run options, the agent retrieves the input messages and updates from the token and provides them to the configured message store and AI context provider in the last successful stream resumption call.

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.A2A/A2AAgent.cs
+++ b/dotnet/src/Microsoft.Agents.AI.A2A/A2AAgent.cs
@@ -84,9 +84,13 @@ public sealed class A2AAgent : AIAgent
         }
         else
         {
-            var a2aMessage = CreateA2AMessage(typedThread, messages);
+            MessageSendParams sendParams = new()
+            {
+                Message = CreateA2AMessage(typedThread, messages),
+                Metadata = options?.AdditionalProperties?.ToA2AMetadata()
+            };
 
-            a2aResponse = await this._a2aClient.SendMessageAsync(new MessageSendParams { Message = a2aMessage }, cancellationToken).ConfigureAwait(false);
+            a2aResponse = await this._a2aClient.SendMessageAsync(sendParams, cancellationToken).ConfigureAwait(false);
         }
 
         this._logger.LogAgentChatClientInvokedAgent(nameof(RunAsync), this.Id, this.Name);
@@ -154,9 +158,13 @@ public sealed class A2AAgent : AIAgent
             // a2aSseEvents = this._a2aClient.SubscribeToTaskAsync(token.TaskId, cancellationToken).ConfigureAwait(false);  
         }
 
-        var a2aMessage = CreateA2AMessage(typedThread, messages);
+        MessageSendParams sendParams = new()
+        {
+            Message = CreateA2AMessage(typedThread, messages),
+            Metadata = options?.AdditionalProperties?.ToA2AMetadata()
+        };
 
-        a2aSseEvents = this._a2aClient.SendMessageStreamingAsync(new MessageSendParams { Message = a2aMessage }, cancellationToken).ConfigureAwait(false);
+        a2aSseEvents = this._a2aClient.SendMessageStreamingAsync(sendParams, cancellationToken).ConfigureAwait(false);
 
         this._logger.LogAgentChatClientInvokedAgent(nameof(RunStreamingAsync), this.Id, this.Name);
 
@@ -198,10 +206,10 @@ public sealed class A2AAgent : AIAgent
     protected override string? IdCore => this._id;
 
     /// <inheritdoc/>
-    public override string? Name => this._name ?? base.Name;
+    public override string? Name => this._name;
 
     /// <inheritdoc/>
-    public override string? Description => this._description ?? base.Description;
+    public override string? Description => this._description;
 
     private A2AAgentThread GetA2AThread(AgentThread? thread, AgentRunOptions? options)
     {
--- a/dotnet/src/Microsoft.Agents.AI.A2A/Extensions/AdditionalPropertiesDictionaryExtensions.cs
+++ b/dotnet/src/Microsoft.Agents.AI.A2A/Extensions/AdditionalPropertiesDictionaryExtensions.cs
@@ -1,0 +1,44 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Collections.Generic;
+using System.Text.Json;
+using Microsoft.Agents.AI;
+
+namespace Microsoft.Extensions.AI;
+
+/// <summary>
+/// Extension methods for AdditionalPropertiesDictionary.
+/// </summary>
+internal static class AdditionalPropertiesDictionaryExtensions
+{
+    /// <summary>
+    /// Converts an <see cref="AdditionalPropertiesDictionary"/> to a dictionary of <see cref="JsonElement"/> values suitable for A2A metadata.
+    /// </summary>
+    /// <remarks>
+    /// This method can be replaced by the one from A2A SDK once it is available.
+    /// </remarks>
+    /// <param name="additionalProperties">The additional properties dictionary to convert, or <c>null</c>.</param>
+    /// <returns>A dictionary of JSON elements representing the metadata, or <c>null</c> if the input is null or empty.</returns>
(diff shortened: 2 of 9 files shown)


dotnet/src/Microsoft.Agents.AI.Abstractions/AgentRunResponse.cs
```diff
@@ -74,7 +74,6 @@ public AgentRunResponse(ChatResponse response)
         this.RawRepresentation = response;
         this.ResponseId = response.ResponseId;
         this.Usage = response.Usage;
-        this.ContinuationToken = response.ContinuationToken;
```
westey-m, line 77: Since this method is public, and anyone can use it, does it make sense to default to the response.ContinuationToken and then we can still override it where we need to in ChatClientAgent?
Same for AgentRunResponseUpdate

dotnet/src/Microsoft.Agents.AI/AgentJsonUtilities.cs
```diff
@@ -68,6 +70,8 @@ private static JsonSerializerOptions CreateDefaultOptions()
     [JsonSerializable(typeof(ChatClientAgentThread.ThreadState))]
     [JsonSerializable(typeof(TextSearchProvider.TextSearchProviderState))]
     [JsonSerializable(typeof(ChatHistoryMemoryProvider.ChatHistoryMemoryProviderState))]
+    [JsonSerializable(typeof(IReadOnlyCollection<ChatMessage>))]
+    [JsonSerializable(typeof(IReadOnlyCollection<ChatResponseUpdate>))]
```
westey-m: These should already be inherited from MEAI via AgentAbstractionsJsonUtilities.DefaultOptions

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
```diff
@@ -739,26 +750,80 @@ private static Task NotifyMessageStoreOfNewMessagesAsync(ChatClientAgentThread t
...
         }
 
-        // Similarly, streaming resumption is not supported when a context provider is used because, currently, there's no good solution
-        // to collect updates received in failed runs and pass them to the last successful run so it can notify the context provider of the updates.
-        if (safeThread.AIContextProvider is not null)
+        return new(continuationToken)
+        {
+            // Save input messages to the continuation token so they can be added to the thread and
+            // provided to the context provider in the last successful streaming resumption run.
+            // That's necessary for scenarios where initial streaming run fails and streaming is resumed later.
```
westey-m: Isn't this required for all resumption scenarios, not just for failures?

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
```diff
@@ -739,26 +750,80 @@ private static Task NotifyMessageStoreOfNewMessagesAsync(ChatClientAgentThread t
...
+            InputMessages = inputMessages?.Count > 0 ? inputMessages : null,
+
+            // Save all updates received so far to the continuation token so they can be provided to the
+            // message store and context provider in the last successful streaming resumption run.
+            // That's necessary for scenarios where a streaming run fails after some updates were received.
```
westey-m: Same as above, is this not required for all resumption scenarios?

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
```diff
@@ -739,26 +750,80 @@ private static Task NotifyMessageStoreOfNewMessagesAsync(ChatClientAgentThread t
...
+    private static void ValidateStreamResumptionAllowed(ChatClientAgentContinuationToken? continuationToken, ChatClientAgentThread safeThread)
+    {
+        if (continuationToken is null)
+        {
+            return;
+        }
+
+        // If neither input messages nor response updates are present in the token,
+        // it means it's an initial run that cannot be resumed.
+        if (continuationToken.InputMessages is not { Count: > 0 } && continuationToken.ResponseUpdates is not { Count: > 0 })
```
westey-m: Could it be possible for someone to invoke the agent without any input messages on the initial run, in which case there wouldn't be any input messages in subsequent runs either?

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
```diff
@@ -239,7 +244,7 @@ public override async IAsyncEnumerable<AgentRunResponseUpdate> RunStreamingAsync
         }
         catch (Exception ex)
         {
-            await NotifyAIContextProviderOfFailureAsync(safeThread, ex, inputMessages, aiContextProviderMessages, cancellationToken).ConfigureAwait(false);
+            await NotifyAIContextProviderOfFailureAsync(safeThread, ex, GetInputMessages(inputMessages, continuationToken), aiContextProviderMessages, cancellationToken).ConfigureAwait(false);
```
westey-m, line 250: FYI, I am also adding notifying the ChatMessageStore of failed requests in the below PR, so whoever merges second will need to update:
https://github.com/microsoft/agent-framework/pull/2604/changes#diff-560fead8693c153637744aa3f09369dd7d83ebdf77fb958cf953a840acbc670a

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
```diff
@@ -204,7 +207,9 @@ public override async IAsyncEnumerable<AgentRunResponseUpdate> RunStreamingAsync
         (ChatClientAgentThread safeThread, ChatOptions? chatOptions, List<ChatMessage> inputMessagesForChatClient, IList<ChatMessage>? aiContextProviderMessages) =
             await this.PrepareThreadAndMessagesAsync(thread, inputMessages, options, cancellationToken).ConfigureAwait(false);
 
-        ValidateStreamResumptionAllowed(chatOptions?.ContinuationToken, safeThread);
+        var continuationToken = ParseContinuationToken(options?.ContinuationToken);
```
westey-m: Below (line 604) we also check ChatOptions.Continuation token and not just AgentRunOptions.ContinuationToken.  Should both be considered here too?

Or perhaps this should just match non-streaming which uses the output chatOptions from PrepareThreadAndMessagesAsync, which should already have considered both?

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
```diff
@@ -798,20 +821,20 @@ private static void ValidatePollingAllowed(ResponseContinuationToken? continuati
...
             ResponseUpdates = responseUpdates?.Count > 0 ? responseUpdates : null
         };
     }
 
-    private static IReadOnlyCollection<ChatMessage> GetInputMessages(IReadOnlyCollection<ChatMessage> inputMessages, ChatClientAgentContinuationToken? token)
+    private static IEnumerable<ChatMessage> GetInputMessages(IEnumerable<ChatMessage> inputMessages, ChatClientAgentContinuationToken? token)
     {
         // First, use input messages if provided.
-        if (inputMessages.Count > 0)
+        if (inputMessages.Any())
```
westey-m: In the [current code](https://github.com/dotnet/dotnet/blob/b0f34d51fccc69fd334253924abd8d6853fad7aa/src/runtime/src/libraries/System.Linq/src/System/Linq/AnyAll.cs#L13C13-L42C10), Any does check internally if the IEnumerable is a Collection and checks it's count (which is more efficient than starting an enumeration), but I'm not sure if that was the case for all older (and .net framework) implementations as well.
That's why I typically try to stick to IReadlyCollection.Count where possible if we need to check if there are any messages.

PR #2759, opened 2025-12-10
.NET: Use GrpcEntityRunner instead of TaskEntityDispatcher

### LINKED ISSUE #2327: .NET: Error running an Azure Function following samples and using NuGet packages for Microsoft.Agents.AI.Hosting.AzureFunctions and Microsoft.Agents.AI.OpenAi
I'm working with the dotnet samples for Azure Functions. I created a new solution with a new function setting it up just like one of the samples referencing the NuGet packages for Microsoft.Agents.AI.OpenAi and Microsoft.Agents.AI.Hosting.AzureFunctions and I'm getting an error consistently. 

I took my new function, added it to the agent-framework-dotnet sample solution and changed the package references for Microsoft.Agents.AI.OpenAi and Microsoft.Agents.AI.Hosting.AzureFunctions to use the projects in that solution and my function started to work. 

I tried the reverse as well and took one of the sample function apps, put it in my solution and had it get those two packages from NuGet and the sample started to fail with the same error as my function app. 

Is there an issue with the NuGet packages? Has anyone else had an issue? Here is the error I see running referencing the NuGet packages.

Microsoft.DurableTask.Entities.EntityOperationFailedException: 'Operation 'RunAgentAsync' of entity '@dafx-<FUNCTIONAME>@e86fd2952ea25d8787b81c9e14acffbd' failed: Invocation of function 'dafx-<FUNCTIONAME>' failed with an exception.'

jviau:
### Motivation and Context

Resolves #2327

### Description

Durable agent framework pipeline was taking a dependency on the precise way the durable worker extension bootstrapped an entity invocation. With a recent change to the durable extension this no longer worked as that bootstrapping was deferred. This PR reduces that dependency by removing usage of `TaskEntityDispatcher` (which is no longer setup by the time this executor runs) and instead uses `GrpcEntityRunner`. `TaskEntityDispatcher` is just a wrapper around `GrpcEntityRunner` anyways.

Another reason reducing these bootstrapping dependencies is important is that it is inherently fragile. It requires the durable functions worker middleware run first, but there is nothing guaranteeing the middleware order between the extensions today. It works right now because durable _tends_ to be configured from source gen auto-start hooks, but that is not set in stone that it will always work that way. For long term stability it will be better to decouple the durable agent's invocations from these implementation details of the durable function worker extension.

### Contribution Checklist

### DIFF
--- a/python/packages/core/agent_framework/_workflows/_agent.py
+++ b/python/packages/core/agent_framework/_workflows/_agent.py
@@ -26,6 +26,7 @@ from agent_framework import (
 )
 
 from ..exceptions import AgentExecutionException
+from ._agent_executor import AgentExecutor
 from ._checkpoint import CheckpointStorage
 from ._events import (
     AgentRunUpdateEvent,
@@ -141,7 +142,8 @@ class WorkflowAgent(BaseAgent):
             checkpoint_storage: Runtime checkpoint storage. When provided with checkpoint_id,
                 used to load and restore the checkpoint. When provided without checkpoint_id,
                 enables checkpointing for this run.
-            **kwargs: Additional keyword arguments.
+            **kwargs: Additional keyword arguments passed through to underlying workflow
+                and ai_function tools.
 
         Returns:
             The final workflow response as an AgentRunResponse.
@@ -153,7 +155,7 @@ class WorkflowAgent(BaseAgent):
         response_id = str(uuid.uuid4())
 
         async for update in self._run_stream_impl(
-            input_messages, response_id, thread, checkpoint_id, checkpoint_storage
+            input_messages, response_id, thread, checkpoint_id, checkpoint_storage, **kwargs
         ):
             response_updates.append(update)
 
@@ -187,7 +189,8 @@ class WorkflowAgent(BaseAgent):
             checkpoint_storage: Runtime checkpoint storage. When provided with checkpoint_id,
                 used to load and restore the checkpoint. When provided without checkpoint_id,
                 enables checkpointing for this run.
-            **kwargs: Additional keyword arguments.
+            **kwargs: Additional keyword arguments passed through to underlying workflow
+                and ai_function tools.
 
         Yields:
             AgentRunResponseUpdate objects representing the workflow execution progress.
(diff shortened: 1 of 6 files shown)


dotnet/src/Microsoft.Agents.AI.Hosting.AzureFunctions/BuiltInFunctions.cs
```diff
@@ -22,14 +23,14 @@ internal static class BuiltInFunctions
     internal static readonly string RunAgentMcpToolFunctionEntryPoint = $"{typeof(BuiltInFunctions).FullName!}.{nameof(RunMcpToolAsync)}";
 
     // Exposed as an entity trigger via AgentFunctionsProvider
-    public static async Task InvokeAgentAsync(
-        [EntityTrigger] TaskEntityDispatcher dispatcher,
+    public static Task<string> InvokeAgentAsync(
         [DurableClient] DurableTaskClient client,
+        string dispatcher,
```
kshyju: instead of `dispatcher`, should we be considering naming this `encodedEntityRequest`? (That is the parameter name of `LoadAndRunAsync` method to which we are passing this value to.)

jviau: For binding, the name needed to match the generated function metadata. I could update that though if you want.

kshyju: We have the freedom to update the metadata value. I will leave it up to you to use the readable name for this one since you are more familiar with the durable concepts.

kshyju (commented):
Can we also update the changelogs.md?

dotnet/src/Microsoft.Agents.AI.Hosting.AzureFunctions/BuiltInFunctions.cs
jviau: Renamed it. Can you do one more round of testing? I still can't get my local environment to work due to lack of permissions.

kshyju: Validated this version and it looks good.

cgillum (approved):

Issue #2694, opened 2025-12-08
Python: MAF Feature Request: Automatic ContextId Propagation for A2A Agent Tools


moetarhinikaseya:
## Summary

When using A2A agents as tools in an LLM orchestrator, there's no automatic propagation of `ContextId` from the orchestrator's thread to the A2A agent calls. This makes it difficult to share session state (e.g., Redis) across multiple A2A agents in the same orchestration.

---

## Expected Behavior (Python Pattern)

In the Python Agent Framework, the orchestrator's thread context automatically flows to all A2A tool calls:

```python
from agent_framework import A2AAgent, AgentThread
from agent_framework.azure import AzureOpenAIChatClient
import uuid

class TriageOrchestrationService:
    def __init__(self):
        # A2A agents as tools
        self.impact_agent = A2AAgent(
            name="ImpactAssessor",
            url="https://agents.example.com/a2a/impact-assessor"
        )
        
        self.assigner_agent = A2AAgent(
            name="SmartAssigner", 
            url="https://agents.example.com/a2a/smart-assigner"
        )
        
        # LLM Orchestrator with A2A agents as tools
        self.orchestrator = AzureOpenAIChatClient().create_agent(
            name="TriageOrchestrator",
            instructions="You orchestrate ticket triage...",
            tools=[
                self.impact_agent.as_tool(name="assess_impact"),
                self.assigner_agent.as_tool(name="get_assignment")
            ]
        )
    
    async def triage_ticket(self, ticket_id: str, description: str) -> dict:
        # Create a shared context ID
        context_id = f"triage-{ticket_id}-{uuid.uuid4().hex[:8]}"
        
        # Create thread with this context ID
        thread = AgentThread(id=context_id)
        
        # LLM decides which agents to call
        # KEY: The thread's ID automatically becomes contextId for ALL A2A calls
        result = await self.orchestrator.run(
            f"Triage ticket #{ticket_id}: {description}",
            thread=thread
        )
        
        return {"context_id": context_id, "result": result.text}

# Usage - context_id "triage-12345-abc123" is automatically used for:
# - ImpactAssessor A2A calls
# - SmartAssigner A2A calls
# - All agents share the same session context
Key mechanism: When orchestrator.run(..., thread=thread) is called, the framework automatically derives contextId from the thread and passes it to all A2A agent tool calls.
Current .NET Behavior
In .NET MAF, AsAIFunction() captures the thread statically at tool creation time, not from the orchestrator's execution context:
[...]

PR #2760, opened 2025-12-10
.NET: Fix: Enable AgentContext Propagation across Async/HTTP Boundaries (Issue #2694)

clint-morgan:
This PR addresses Issue #2694 by adding support for serializing and propagating `AgentContext` across process boundaries (e.g., HTTP calls between agents).

### Problem
Currently, `AgentContext.CurrentThread` is thread-local and is lost when an agent makes an outbound call to another agent via HTTP or other transports. This breaks the ambient context pattern, causing loss of `ContextId`, `ConversationId`, and distributed tracing correlation in downstream agents.

### Changes
1. **`AgentContext.cs`**: Added `CreateSnapshot()` and `RestoreFromSnapshot()` methods to serialize/deserialize the current thread context.
2. **`AgentContextSnapshot`**: Introduced a serializable DTO for transmitting context metadata.
3. **`RestoredAgentContextThread`**: Added an internal implementation of `IAgentContextThread` to support restored contexts.

### Benefits
- Enables true distributed tracing and context propagation in microservice-style agent architectures.
- Allows downstream agents to access the original `ContextId` and `ConversationId` without explicit parameter passing.
- Preserves governance metadata across the execution chain.

### Testing
- Verified locally with a multi-agent workflow (Orchestrator -> Agent A -> Agent B).
- Confirmed that `AgentContext.CurrentThread` is correctly populated in the receiving agent.


clint-morgan:
@microsoft-github-policy-service agree [company="Kaseya"]

[cid:c10ce884-2bfc-4f51-85a4-ce44b570f650]
________________________________

clint-morgan:
@microsoft-github-policy-service agree company="Kaseya"

[cid:0adc9def-c4e9-406a-bc1b-9e8d26552b9b]
________________________________

moetarhinikaseya (approved):

dotnet/src/Microsoft.Agents.AI/AgentExtensions.cs
```diff
@@ -73,7 +73,24 @@ async Task<string> InvokeAgentAsync(
             [Description("Input query to invoke the agent.")] string query,
             CancellationToken cancellationToken)
         {
-            var response = await agent.RunAsync(query, thread: thread, cancellationToken: cancellationToken).ConfigureAwait(false);
+            // If no thread was provided at creation time, try to use the current ambient thread context.
+            var effectiveThread = thread ?? AgentContext.CurrentThread;
```
markwallace-microsoft, line 77: Providing an ambient thread will be problematic if you have heterogeneous agents which require different `AgentThread` implementations. We are in the process of redesigning our `AgentThread` abstraction and we will consider this case.

PR #2762, opened 2025-12-10
.NET: fix: Expose WorkflowErrorEvent as ErrorContent

lokitoth:
### Motivation and Context

When hosted using `AsAgent()`, `Workflow`s were not exposing inner errors coming as `Exception`s (through raised `WorkflowErrorEvent`)

### Description

The fix is to convert the `Message` property to an `ErrorContent` on the way out, rather than rely on the default "empty update" to collect the raw event.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.Workflows/WorkflowThread.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Workflows/WorkflowThread.cs
@@ -18,14 +18,16 @@ internal sealed class WorkflowThread : AgentThread
 {
     private readonly Workflow _workflow;
     private readonly IWorkflowExecutionEnvironment _executionEnvironment;
+    private readonly bool _includeExceptionDetails;
 
     private readonly CheckpointManager _checkpointManager;
     private readonly InMemoryCheckpointManager? _inMemoryCheckpointManager;
 
-    public WorkflowThread(Workflow workflow, string runId, IWorkflowExecutionEnvironment executionEnvironment, CheckpointManager? checkpointManager = null)
+    public WorkflowThread(Workflow workflow, string runId, IWorkflowExecutionEnvironment executionEnvironment, CheckpointManager? checkpointManager = null, bool includeExceptionDetails = false)
     {
         this._workflow = Throw.IfNull(workflow);
         this._executionEnvironment = Throw.IfNull(executionEnvironment);
+        this._includeExceptionDetails = includeExceptionDetails;
 
         // If the user provided an external checkpoint manager, use that, otherwise rely on an in-memory one.
         // TODO: Implement persist-only-last functionality for in-memory checkpoint manager, to avoid unbounded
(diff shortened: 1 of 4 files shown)


dotnet/tests/Microsoft.Agents.AI.Workflows.UnitTests/WorkflowHostSmokeTests.cs
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    private static Workflow CreateWorkflow(bool failByThrowing)
+    {
+        ExecutorBinding agent = new AlwaysFailsAIAgent(failByThrowing).BindAsExecutor(emitEvents: true);
+
+        return new WorkflowBuilder(agent).Build();
+    }
+
+    private async static Task InvokeAsAgentAndProcessResponseAsync(Workflow workflow)
```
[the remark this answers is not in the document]
lokitoth: This is a misfire. The `InvokeAsAgentAndProcessResponseAsync` does indeed have "Async" as the suffix.

dotnet/tests/Microsoft.Agents.AI.Workflows.UnitTests/WorkflowHostSmokeTests.cs
```diff
@@ -0,0 +1,108 @@
...
+    {
+        List<AgentRunResponseUpdate> updates = await workflow.AsAgent("WorkflowAgent")
```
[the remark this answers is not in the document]
lokitoth: `var` is not present here.

dotnet/tests/Microsoft.Agents.AI.Workflows.UnitTests/WorkflowHostSmokeTests.cs
```diff
@@ -0,0 +1,108 @@
...
+            if (failByThrowing)
+            {
+                throw new ExpectedException(ErrorMessage);
+            }
+
+            yield return new AgentRunResponseUpdate(ChatRole.Assistant, [new ErrorContent(ErrorMessage)]);
+        }
+    }
+
+    private static Workflow CreateWorkflow(bool failByThrowing)
```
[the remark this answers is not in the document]
lokitoth: `var` is not present in this code

dotnet/src/Microsoft.Agents.AI.Workflows/WorkflowThread.cs
```diff
@@ -153,10 +155,25 @@ IAsyncEnumerable<AgentRunResponseUpdate> InvokeStageAsync(
...
+                    case WorkflowErrorEvent workflowError:
+                        Exception? exception = workflowError.Exception;
+                        if (exception is TargetInvocationException tie && tie.InnerException != null)
+                        {
+                            exception = tie.InnerException;
+                        }
+
+                        if (exception != null)
+                        {
+                            ErrorContent errorContent = new(exception.Message);
```
markwallace-microsoft: Any security concerns with including the exception message here?

The error content may go back to the application and be displayed.

dotnet/src/Microsoft.Agents.AI.Workflows/WorkflowThread.cs
```diff
@@ -3,6 +3,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
```
markwallace-microsoft, line 6: Is this needed?

Any concerns about impact if we plan to support Ahead-of-Time (AOT) compilation for workflows packages.

lokitoth: This is needed for catching `TargetInvocationException`, not because we are doing actual reflection; we throw this out of `Executor`.

dotnet/src/Microsoft.Agents.AI.Workflows/WorkflowThread.cs
lokitoth: I added a way to enable showing error messages, and made it have a generic "an error occurred" by default.

Issue #2234, opened 2025-11-14
.NET: Durable Agents: platforms should use consistent entity method names


philliphoff:
Currently, the entities used to implement durable agents use different method names for invoking the agents, `RunAgentAsync` on .NET and `run_agent` on Python.  This inconsistency makes it difficult for tools (such as the DTS Dashboard) to interact with the agents at the entity-level (as opposed to having developers go through their application level for testing).

We should use a common set of entity method names across the platforms. (I'd suggest `RunAgent` but I don't have a strong opinion.)

anthonychu:
Ideally the current method name would still continue to work for at least a few more versions before removing.

cgillum:
Closing issue as PRs for both .NET and Python have been merged.

PR #2843, opened 2025-12-12
.NET: Switch to new "Run" method name.

philliphoff:
### Motivation and Context

Currently, the durable agent entities use an entity method name to invoke the agent that is specific to each supported platform (e.g. `RunAgentAsync` on .NET but `run_agent` on Python).  This makes it difficult for external tools (such as the Durable Task Scheduler Dashboard) to interact with durable agents in a platform agnostic manner.

This change updates the .NET durable agent infrastructure to use a new, more general `Run` (case-insensitive) method name. The existing `RunAgentAsync` method will still be available to accommodate older MAF clients and any pending entity signals, but considered deprecated.  After a reasonable time, the `RunAgentAsync` method will be removed.

Resolves (for .NET) #2234.  A corresponding change will be needed for Python.

### Description

Adds a new `Run` method to the `AgentEntity` that performs the same logic as the existing `RunAgentAsync` method. The `RunAgentAsync` method then just delegates to the new method (in preparation for it being removed in the future).

### Contribution Checklist


dotnet/src/Microsoft.Agents.AI.DurableTask/AgentEntity.cs
```diff
@@ -20,7 +20,16 @@ internal class AgentEntity(IServiceProvider services, CancellationToken cancella
         ? cancellationToken
         : services.GetService<IHostApplicationLifetime>()?.ApplicationStopping ?? CancellationToken.None;
 
-    public async Task<AgentRunResponse> RunAgentAsync(RunRequest request)
+    public Task<AgentRunResponse> RunAgentAsync(RunRequest request)
+    {
+        return this.RunAgent(request);
+    }
+
+#pragma warning disable IDE1006
```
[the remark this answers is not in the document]
philliphoff: Added comment.

cgillum:
@philliphoff one thing not mentioned in the PR description - what is the plan for Python?

philliphoff:
> one thing not mentioned in the PR description - what is the plan for Python?

@cgillum I could give it a go (but would rather it be a separate PR)?

cgillum:
@philliphoff sorry I should have been more clear. I'm not asking that we make the Python change in the same PR (though that would be nice). Rather, I'm trying to understand what kind of change we plan on making (now or later) in Python. In the PR description, you call out the name differences as being `RunAgentAsync` vs `run_agent`. With your change, it's now `RunAgent` vs `run_agent`. Does this mean we plan on changing Python to be `RunAgent`, which goes against Python method naming conventions?

I originally assumed we'd want to go with a name that works with the conventions of both languages. For example, using `Run` (.NET) and `run` in Python, without needing to worry about PascalCase vs. snake_case conventions.

philliphoff:
> I originally assumed we'd want to go with a name that works with the conventions of both languages. For example, using Run (.NET) and run in Python, without needing to worry about PascalCase vs. snake_case conventions.

@cgillum I suggested/chose the case-insensitive `runagent` for the common name as it's the most basic verb-object form for method naming.  (`Run`, in my opinion, is too terse and generic, but I could go along with that, too.) Since they're just strings passed to the signal method, there's no reason for it to be tied to any platform convention (other than their implementations which happen to directly bind to a method with that name).  `RunAgent` doesn't conform to the .NET guidelines (due to the missing `Async` suffix).  Luckily, the DF SDK treats entity method names as case-insensitive, which allows the .NET implementation to be *near* idiomatic.

(I was actually surprised that the DF SDK doesn't allow overriding the entity method name via attribute; that would be a bit cleaner.)

cgillum:
@philliphoff okay, that's basically what I wanted to confirm. I'd personally prefer `run` because then we only have to violate the conventions of one language and not two languages, but it's not a blocker to me as long as the Python folks are also okay with it. As far as verb-object, I'm not worried about that because the object is implied by the fact that this is an "Agent" instance that you're invoking. In that sense, "RunAgent" or "run_agent" is already redundant and renaming to "run" removes the redundancy.

/cc @larohra @gavin-aguiar

philliphoff:
The corresponding Python change is #2890.

cgillum (commented):
Looks good. Can you also update the CHANGELOG.md file with this update? I was hoping Copilot would ask for this, but it seems to not be following the new instructions I added.

philliphoff:
> Can you also update the CHANGELOG.md file with this update?

Added.

cgillum (approved):

cgillum (approved):

gavin-aguiar (approved):

dotnet/src/Microsoft.Agents.AI.DurableTask/AgentEntity.cs
```diff
@@ -21,7 +21,17 @@ internal class AgentEntity(IServiceProvider services, CancellationToken cancella
         ? cancellationToken
         : services.GetService<IHostApplicationLifetime>()?.ApplicationStopping ?? CancellationToken.None;
 
-    public async Task<AgentRunResponse> RunAgentAsync(RunRequest request)
+    public Task<AgentRunResponse> RunAgentAsync(RunRequest request)
```
kshyju, line 24: Should we add `ObsoleteAttribute` for this method and suggest consumers to consider the new `Run`?

kshyju (approved):

dotnet/src/Microsoft.Agents.AI.DurableTask/AgentEntity.cs
kshyju: Didn't realize AgentEntity is internal.

PR #2927, opened 2025-12-17
.NET: Structured Output calling work with lists/array if ResponseFormat is used

### LINKED ISSUE #2874: .NET: Structured Output calling does not work with lists/array if ResponseFormat is used, but works if .RunAsync<> is used
If you make the following agent and ask for Movies you correctly get 10 movies in an array (r1) and a list (r2) :

```cs
AzureOpenAIClient azureOpenAiClient = ...;
ChatClientAgent chatClientAgent = azureOpenAiClient.GetChatClient("gpt-4.1-mini").CreateAIAgent();

ChatClientAgentRunResponse<Movie[]> r1 = await chatClientAgent.RunAsync<Movie[]>("Give me top 10 IMDB Movies");
Movie[] r1Movies = r1.Result;

ChatClientAgentRunResponse<List<Movie>> r2 = await chatClientAgent.RunAsync<List<Movie>>("Give me top 10 IMDB Movies");
List<Movie> r2Movies = r2.Result;

public class Movie
{
    public required string Title { get; set; }
    public required int Year { get; set; }
}
```

So far so good 👍

Now try the same using the ResponseFormat way (if you don't have a ChatClientAgent due to example middleware and can't use .RunAsync<> anymore 😢)

```cs
JsonSerializerOptions jsonSerializerOptions = new()
{
    PropertyNameCaseInsensitive = true,
    TypeInfoResolver = new DefaultJsonTypeInfoResolver(),
    Converters = { new JsonStringEnumConverter() }
};

AzureOpenAIClient azureOpenAiClient = ...;
ChatClientAgent chatClientAgent = azureOpenAiClient.GetChatClient("gpt-4.1-mini").CreateAIAgent(
    new ChatClientAgentOptions
    {
        ChatOptions = new ChatOptions
        {
            ResponseFormat = ChatResponseFormat.ForJsonSchema<Movie[]>(jsonSerializerOptions)
        }
    });
AgentRunResponse agentRunResponse = await chatClientAgent.RunAsync("Give me top 10 IMDB Movies");
```

Error: 

```
System.ClientModel.ClientResultException: 'HTTP 400 (invalid_request_error: )
Parameter: response_format

Invalid schema for response_format 'Movie__': schema must be a JSON Schema of 'type: "object"', got 'type: "array"'.'
```

So the `ResponseFormat` does not work with List and Arrays, while the .RunAsync<>() does (only plain objects).

feiyun0112:
fix #2874

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
+++ b/dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
@@ -474,16 +474,19 @@ public sealed partial class ChatClientAgent : AIAgent
     {
         ChatOptions? requestChatOptions = (runOptions as ChatClientAgentRunOptions)?.ChatOptions?.Clone();
 
-        // If no agent chat options were provided, return the request chat options as is.
+        // If no agent chat options were provided, wrap array response formats if needed and return
         if (this._agentOptions?.ChatOptions is null)
         {
+            WrapArrayResponseFormatIfNeeded(requestChatOptions);
             return ApplyBackgroundResponsesProperties(requestChatOptions, runOptions);
         }
 
-        // If no request chat options were provided, use the agent's chat options clone.
+        // If no request chat options were provided, use the agent's chat options clone and wrap if needed
         if (requestChatOptions is null)
         {
-            return ApplyBackgroundResponsesProperties(this._agentOptions?.ChatOptions.Clone(), runOptions);
+            ChatOptions? agentChatOptions = this._agentOptions?.ChatOptions.Clone();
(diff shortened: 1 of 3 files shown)


stephentoub:
Thanks, but the implementation explicitly doesn't do this for the reasons I call out in that issue. If the schema is changed, it will not be desereializable using the same T, eg trying to do `JsonSerializer.Deserialize<T>` will fail. That's why it's the caller's responsibility to provide a valid T. `RunAsunc<T>` is responsible for both creating the schema and deserializing, so it can change the schema and agree with itself on what it is. I appreciate the attempt, but I think this should be closed. The right solution is instead to come up with a more holistic way that the equivalent of `RunAsync` can work with any agent.

cc: @SergeyMenshykh

PR #2894, opened 2025-12-16
Python: Flow custom kwargs to agents via Workflow SharedState

moonbox3:
### Motivation and Context

This PR enables passing custom keyword arguments (for example, `custom_data`, `user_token`, or others) from `workflow.run_stream()` or `run()` directly to agent `@ai_function` tools.

### Description

Fixes an issue where currently we aren't able to to propagate kwargs through workflow invocations to tools that may depend on certain input data (outside that of which the model would generate).

- Closes #2895

### Contribution Checklist

- [X] The code builds clean without any errors or warnings
- [X] The PR follows the [Contribution Guidelines](https://github.com/microsoft/agent-framework/blob/main/CONTRIBUTING.md)
- [X] All unit tests pass, and I have added new tests where possible
- [ ] **Is this a breaking change?** If yes, add "[BREAKING]" prefix to the title of the PR.

PR #2945, opened 2025-12-18
Python: Adds support for flowing kwargs to context and tool calls when using …

nmccann:
…workflow as agent.

### Motivation and Context

#2894 introduced support for flowing kwargs through the `Workflow` context so that they can be used by tool calls, however `WorkflowAgent` wasn't updated to provide similar functionality, this change allows WorkflowAgent to behave in a similar manner.

### Description

Applied similar solution found in #2894 to `WorkflowAgent`.

### Contribution Checklist

### DIFF
--- a/python/packages/core/agent_framework/_workflows/_agent.py
+++ b/python/packages/core/agent_framework/_workflows/_agent.py
@@ -141,7 +141,19 @@ class WorkflowAgent(BaseAgent):
             checkpoint_storage: Runtime checkpoint storage. When provided with checkpoint_id,
                 used to load and restore the checkpoint. When provided without checkpoint_id,
                 enables checkpointing for this run.
-            **kwargs: Additional keyword arguments.
+            **kwargs: Additional keyword arguments to pass through to agent invocations.
+                     These are stored in SharedState and accessible in @ai_function tools
+                     via the **kwargs parameter.
+
+            With custom context for ai_functions:
+
+            .. code-block:: python
+
+                result = await workflow_agent.run(
+                    "analyze data",
+                    custom_data={"endpoint": "https://api.example.com"},
+                    user_token={"user": "alice"},
+                )
 
         Returns:
             The final workflow response as an AgentRunResponse.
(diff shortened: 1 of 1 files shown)


nmccann:
I'd like to add some unit tests to cover this scenario, but unsure if they should go in `test_workflow_kwargs` or `test_workflow_agent` or a new file.

Also, in the course of manually testing this, I noticed that if we have an executor that calls an agent by itself (as seen in some of the "Getting Started" samples), it's not straightforward to get the `kwargs` from the workflow's context and forward those to the agent so that it has access to the same `kwargs`. My current thought of how to do that would be to import the `WORKFLOW_RUN_KWARGS_KEY` to use to extract it from the context. But that doesn't seem like the intended approach.

Maybe we shouldn't be calling agents within an executor, or if we do we should consider it to be "isolated" from the broader state of the workflow.

moonbox3:
Didn't know you were also working on this. As I was plumbing the kwargs through, I spotted another bug about not properly handling a list[ChatMessage] with a WorkflowAgent in #2946, as well as adding tests and a new sample.

python/packages/core/agent_framework/_workflows/_agent.py
```diff
@@ -153,7 +165,12 @@ async def run(
         response_id = str(uuid.uuid4())
 
         async for update in self._run_stream_impl(
-            input_messages, response_id, thread, checkpoint_id, checkpoint_storage
+            input_messages,
+            response_id,
+            thread,
+            checkpoint_id,
+            checkpoint_storage,
+            run_kwargs=kwargs if kwargs else None,
```
moonbox3, line 173: any reason we couldn't use idiomatic Python **kwargs forwarding?

python/packages/core/agent_framework/_workflows/_agent.py
```diff
@@ -255,6 +292,7 @@ async def _run_stream_impl(
                 message=None,
                 checkpoint_id=checkpoint_id,
                 checkpoint_storage=checkpoint_storage,
+                **(run_kwargs if run_kwargs else {}),
```
moonbox3, line 295: Feels like this adds some unnecessary complexity - could we simply propagate `**kwargs` as-is?

nmccann:
> Didn't know you were also working on this. As I was plumbing the kwargs through, I spotted another bug about not properly handling a list[ChatMessage] with a WorkflowAgent in #2946, as well as adding tests and a new sample.

No worries, I'm good with closing this one in favour of #2946. Was mostly looking to get familiarity with the process of running it locally and creating a PR, which I now have. Will look through your PR to see how you did the tests etc.

I would also need to wait to get approval for the CLA, and that turns out to be difficult to do this time of year :)

PR #2984, opened 2025-12-19
Python: Bug fix: OpenAI Chat Client to consume author name

TaoChenOSU:
### Motivation and Context

The OpenAI Chat client doesn't consume author names set in messages. This essentially prevents all multi-agent scenarios from working because models won't have context on who has spoken.

### Description

This PR fixes the issue in the OpenAI client. For other clients, we will do a pass separately. 

### Contribution Checklist


TaoChenOSU:
This is fixed in another PR.

PR #3042, opened 2025-12-27
.NET: Durable Agent samples and automated validation for non-Azure Functions

cgillum:
### Motivation and Context

Durable Agents are intended to be available both in Azure Functions and non-Azure Functions contexts. The capability already exists for .NET, but no docs or samples have been created yet.

### Description

This PR adds samples showing how to implement the various durable agent scenarios as simple console apps. No Azure Functions is required. Instead, apps use the Durable Task Scheduler directly.

All Azure Functions samples have been converted into their equivalent non-Azure Functions console apps with the exception of the MCP sample, which requires features specific to Azure Functions.

### Contribution Checklist


larohra (approved):

kshyju (approved):

dotnet/samples/DurableAgents/ConsoleApps/07_ReliableStreaming/Program.cs
```diff
@@ -0,0 +1,352 @@
...
+
+string? prompt = Console.ReadLine();
+if (string.IsNullOrWhiteSpace(prompt) || prompt.Equals("exit", StringComparison.OrdinalIgnoreCase))
+{
+    return;
+}
+
+// Create a new agent thread
+AgentThread thread = agentProxy.GetNewThread();
+AgentThreadMetadata? metadata = thread.GetService<AgentThreadMetadata>();
```
westey-m: We removed AgentThreadMetadata in a separate PR since it wasn't really used everywhere.  In this case, there is already an alternative service exposed by the DurableAgentThread, which can be used to fetch the conversation id instead of using AgentThreadMetadata:
```csharp
string conversationId = thread.GetService<AgentSessionId>().ToString()
```

cgillum: OK, thanks for the heads up!

kshyju (approved):

dotnet/samples/DurableAgents/ConsoleApps/01_SingleAgent/01_SingleAgent.csproj
```diff
@@ -0,0 +1,30 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
```
kshyju, line 2: Should we consider renaming the parent folder from `/samples/DurableAgents` to `samples/Durable` so that it can hold both Agents and Workflow(when it is ready) ( `/samples/Durable/Agents` and ` /samples/Durable/Workflow`)?

cgillum: Let's consider that for a separate PR that contains only file moves. I also want to move the AzureFunctions samples under this directory, so we can just move them all at once.

kshyju (approved):

kshyju (approved):

Issue #303, opened 2025-08-01
Threads: Investigate whether we can have pure state only threads, with stateless behaviors living elsewhere


westey-m:
## Background

Today threads can have an attached ChatMessageStore or AIContextProvider that both contribute state to the thread.

Having both the state and behaviors attached to the thread makes it difficult to deserialize a thread without needing an agent based factory, since both ChatMessageStores and AIContextProviders typically require some service client to be constructed.

While deserializing via the factory and serializing via a custom helper on AgentThread is not complex, it is a pattern that users don't expect.  They often assume that they should just be able to use the serializer of their choice to serialize and deserialize the thread object.

It also means that deserializing any container object that also contains thread json, cannot just be deserialized in it's entirety.  
You also cannot easily just serialize a `List<AgentThread>`.  So none of this is a great user experience.

## Option to investigate

We should investigate whether we can have AgentThreads that resemble the state tree instead of resembling the behaviors attached to it.  The AgentThread would contain state properties for each type of behavior it supports, e.g. in the case of ChatClientAgentThread, one for ChatMessageStore and one for AIContextProvider.  These objects would be directly serializable and deserializable with the AgentThread.

After a thread is deserialized, we would still need to re-create the ChatMessageStore and AIContextProvider that use these state objects and point them at the state objects.  We could also set the re-created behavior classes on the AgentThread, but mark them as non-serializable.  That way, they get recreated on first use, each time after deserialization, but doesn't stop anyone from serializing the thread as is.

A challenge here is polymorphic deserialization.  Since users can attach their own ChatMessageStore/AIContextProvider, that stores its own state, we would need to potentially have a base class for each state type.  E.g. a ChatMessageStore base type, that we can annotate with the list of inheriting types.  Plus a way for developers to extend these with their own types.

### Example

Something along the lines of this.

```csharp
public class ChatClientAgentThread
{
    public ChatMessageStoreState StoreState { get; set; }
    public AIContextProviderState AIContextProviderState { get; set; }

    [JsonIgnore]
    internal ChatMessageStore? ChatMessageStore { get; set; }
    [JsonIgnore]
    internal AIContextProvider? AIContextProvider { get; set; }
}

[JsonPolymorphic(TypeDiscriminatorPropertyName = "$type")]
[JsonDerivedType(typeof(InMemoryChatMessageStoreState), typeDiscriminator: "inmemory")]
public abstract class ChatMessageStoreState {}

[...]

PR #3062, opened 2026-01-02
.NET: AgentThread serialization alternatives ADR

westey-m:
### Motivation and Context

There are some drawbacks to the current serialization mechanism used for AgentThread instances, so creating an ADR to consider some alternatives and their implications.

#303 

### Contribution Checklist


docs/decisions/00NN-agentthread-serialization.md
```diff
@@ -0,0 +1,133 @@
...
+
+- A. Ability to continue to support custom behaviors.
+- B. Ability to serialize and deserialize AgentThreads via standard serialization mechanisms, e.g. JsonSerializer.Serialize and JsonSerializer.Deserialize.
+- C. Ability for the caller to access custom behaviors.
+
+## Considered Options
+
+- Option 1: Separate state from behavior, serialize state only and re-attach behavior on first usage
+- Option 2: Separate state from behavior, and only have state on AgentThread
+- Option 3: Keep the current approach of custom Serialize/Deserialize methods
```
SergeyMenshykh: Can Options 1 and 3 coexist? If so, we can keep all the benefits of both, allowing client code to decide on the option depending on their scenario.

westey-m: They probably could, but I'm not sure if that is better.  The behavior for the two options would differ somewhat, especially around whether the ChatMessageStore / AIContextProvider properties are instantiated or not, meaning users are not just confronted with two options for doing the same thing, these also behave differently.

westey-m: I'm Updating the ADR to reflect the latest changes, which departed quite a bit from the previous options.

PR #3071, opened 2026-01-04
Python: add voice live agent

szhaomsft:
### Motivation and Context
add voice live agent


eavanvalkenburg:
@szhaomsft while we appreciate the initiative to add this, we are working on a more broadly compatible design for voice live and other realtime services, see the WIP here: https://github.com/eavanvalkenburg/agent-framework/blob/voice_agents/docs/decisions/00XX-realtime-agents.md so we first want to decide how we want to abstract this as a whole before we ship something that we then have to change significantly.

eavanvalkenburg:
Closing now.

PR #3094, opened 2026-01-06
Python: feat(ag-ui, openai): Add TextReasoningContent to thinking events and reasoning_field to OpenAI client

chenaoxd:
### Motivation and Context

The `TextReasoningContent` type exists in agent-framework for representing extended thinking/reasoning content from LLMs (e.g., Anthropic's extended thinking, OpenAI's reasoning). However, the AG-UI Event Bridge does not handle this content type, meaning thinking content is silently dropped when streaming to AG-UI clients.

Additionally, OpenAI-compatible APIs (like Kimi, DeepSeek) return reasoning content in a custom field (e.g., `reasoning_content`) rather than the standard response format. The current `OpenAIChatClient` has no way to extract this content.

### Description

This PR adds two related features:

**1. AG-UI Event Bridge (`_events.py`)**
- Handle `TextReasoningContent` by emitting AG-UI thinking events:
  - `ThinkingStartEvent` → `ThinkingTextMessageStartEvent` → `ThinkingTextMessageContentEvent` → `ThinkingTextMessageEndEvent` → `ThinkingEndEvent`
- Support `protected_data` field extraction for OpenAI's `reasoning_details` format
- Follow the existing pattern where START/CONTENT events are emitted in the bridge, and END events are emitted by the orchestrator (consistent with `TextMessageEndEvent`)

**2. OpenAI Chat Client (`_chat_client.py`)**
- Add optional `reasoning_field` parameter to `OpenAIChatClient`
- When set (e.g., `reasoning_field="reasoning_content"`), extract content from that field and emit `TextReasoningContent`
- Enables support for Kimi, DeepSeek, and other OpenAI-compatible APIs with extended thinking

### Contribution Checklist

  - no

### DIFF
--- a/python/packages/core/agent_framework/openai/_chat_client.py
+++ b/python/packages/core/agent_framework/openai/_chat_client.py
@@ -276,6 +276,10 @@ class OpenAIBaseChatClient(OpenAIBase, BaseChatClient):
                 contents.append(text_content)
             if reasoning_details := getattr(choice.delta, "reasoning_details", None):
                 contents.append(TextReasoningContent(None, protected_data=json.dumps(reasoning_details)))
+            # Handle custom reasoning field for OpenAI-compatible APIs (e.g., Kimi, DeepSeek)
+            if reasoning_field := getattr(self, "reasoning_field", None):
+                if reasoning_content := getattr(choice.delta, reasoning_field, None):
+                    contents.append(TextReasoningContent(text=reasoning_content, raw_representation=choice))
         return ChatResponseUpdate(
             created_at=datetime.fromtimestamp(chunk.created, tz=timezone.utc).strftime("%Y-%m-%dT%H:%M:%S.%fZ"),
             contents=contents,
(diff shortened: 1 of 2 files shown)


python/packages/core/agent_framework/openai/_chat_client.py
```diff
@@ -276,6 +276,10 @@ def _parse_response_update_from_openai(
                 contents.append(text_content)
             if reasoning_details := getattr(choice.delta, "reasoning_details", None):
                 contents.append(TextReasoningContent(None, protected_data=json.dumps(reasoning_details)))
+            # Handle custom reasoning field for OpenAI-compatible APIs (e.g., Kimi, DeepSeek)
+            if reasoning_field := getattr(self, "reasoning_field", None):
```
eavanvalkenburg, line 280: How would a user discover what the reasoning field name is for other API's then openai?

chenaoxd: If I understand your question correctly, users would typically refer to the provider's documentation.

For example, DeepSeek documents reasoning_content here: https://api-docs.deepseek.com/guides/thinking_mode

Issue #3096, opened 2026-01-06
Python: API Review: feedback part 2


eavanvalkenburg:
### Language

Python

### Type of Issue

Feature Request

### Description

Comments List from Agent Framework Files

---

#### 1. AgentProtocol class name (Line 96-98)
**Location:** AgentProtocol class definition  
**Comment:** AgentLike might be better, because it might confuse with things like the A2A protocol  
**Type:** Question  
**Context:**
```python
@runtime_checkable
class AgentProtocol(
    Protocol
):  # -> AgentLike might be better, because it might confuse with things like the A2A protocol
```

---

#### 2. ChatAgent.__init__ chat_client parameter (Line 592)
**Location:** ChatAgent class, __init__ method, chat_client parameter  
**Comment:** could this be a Callable[..., ChatClientProtocol] ?  
**Type:** Question  
**Context:**
```python
def __init__(
    self,
    chat_client: ChatClientProtocol,  # -> could this be a Callable[..., ChatClientProtocol] ?
    instructions: str | None = None,
    *,
    ...
) -> None:
```

---

#### 3. ChatClientProtocol.additional_properties (Line 84)
**Location:** ChatClientProtocol protocol, additional_properties property  
**Comment:** change to attribute  
**Type:** Note  
**Context:**
```python
@property  # change to attribute
def additional_properties(self) -> dict[str, Any]:
    """Get additional properties associated with the client."""
    ...
```

---

#### 4. ChatClientProtocol.get_response messages parameter (Line 91)
**Location:** ChatClientProtocol protocol, get_response method, messages parameter  
**Comment:** is this enough? or will people call this manually?  
**Type:** Question  
**Context:**
```python
async def get_response(
    self,
    messages: list[ChatMessage],  # is this enough? or will people call this manually?
    *,
    ...
) -> ChatResponse:
```

---

#### 5. ChatClientProtocol.get_response stream parameter (Line 93-96)
**Location:** ChatClientProtocol protocol, get_response method, stream parameter and kwargs  
**Comment:** can we use this to combine both methods into one, while keeping agent.run and agent.run_stream? Use a generic typeddict as kwargs to allow provider-specific parameters? Could be specified in order to support type checking and IDE autocompletion. User could provide their own  
**Type:** Question  
**Context:**
```python
async def get_response(
    self,
    messages: list[ChatMessage],
    *,
    stream: bool = True,  # -> can we use this to combine both methods into one, while keeping agent.run and agent.run_stream?
[...]

PR #3098, opened 2026-01-06
Python: ADR for simplified get response

eavanvalkenburg:
### Motivation and Context

ADR describing the implications of a potential simplification of the `get_response` method in ChatClients, this would leverage python's type system and a `stream=True/False` param to steer the users instead of using two separate methods, that already have a lot of overlap.

Related to #3096 - Point 5

### Contribution Checklist

### DIFF
--- a/docs/decisions/0012-python-get-response-simplification.md
+++ b/docs/decisions/0012-python-get-response-simplification.md
@@ -148,7 +148,7 @@ Smaller questions to consider:
 Chosen Option: **Option 3: Consolidate + Merge Agent and Workflow Methods**
 
 Since this is the most pythonic option and it reduces the API surface and code duplication the most, we will go with this option.
-
+We will keep the default of `stream=False` for all methods to maintain backward compatibility and simplicity for new users.
 
 # Appendix
 ## Code Samples for Consolidated Method
--- a/docs/decisions/0012-python-get-response-simplification.md
+++ b/docs/decisions/0012-python-get-response-simplification.md
@@ -1,5 +1,5 @@
 ---
-status: Proposed
+status: Accepted
 contact: eavanvalkenburg
 date: 2026-01-06
 deciders: markwallace-microsoft, dmytrostruk, taochenosu, alliscode, moonbox3, sphenry
@@ -144,7 +144,10 @@ Smaller questions to consider:
     - Should this differ between ChatClient, Agent and Workflows? (e.g., Agent and Workflow defaults to streaming, ChatClient to non-streaming)
 
 ## Decision Outcome
-TBD
+
+Chosen Option: **Option 3: Consolidate + Merge Agent and Workflow Methods**
+
+Since this is the most pythonic option and it reduces the API surface and code duplication the most, we will go with this option.
 
 
 # Appendix
--- a/docs/decisions/00XX-python-get-response-simplification.md
+++ b/docs/decisions/00XX-python-get-response-simplification.md
@@ -169,7 +169,7 @@ async def main() -> None:
 
     # Non-streaming usage
     print(f"User: {message}")
-    response = await client.get_response(message, tools=get_weather, stream=False)
+    response = await client.get_response(message, tools=get_weather)
     print(f"Assistant: {response.text}")
 
     # Streaming usage - same method, different parameter
@@ -191,7 +191,7 @@ async def main() -> None:
 
     # Non-streaming agent
     print(f"\nUser: {message}")
-    result = await agent.run(message, thread=thread, stream=False)
+    result = await agent.run(message, thread=thread) # default would be stream=False
     print(f"{agent.name}: {result.text}")
 
     # Streaming agent - same method, different parameter
--- a/docs/decisions/00XX-python-get-response-simplification.md
+++ b/docs/decisions/00XX-python-get-response-simplification.md
@@ -11,7 +11,7 @@ informed:
 
 ## Context and Problem Statement
 
-Currently chat clients must implement two separate methods to get responses, one for streaming and one for non-streaming. This adds complexity to the client implementations and increases the maintenance burden. This was likely done because the dotnet version cannot do proper typing with a single method, in Python this is possible and this for instance is also how the OpenAI python client works.
+Currently chat clients must implement two separate methods to get responses, one for streaming and one for non-streaming. This adds complexity to the client implementations and increases the maintenance burden. This was likely done because the .NET version cannot do proper typing with a single method, in Python this is possible and this for instance is also how the OpenAI python client works, this would then also make it simpler to work with the Python version because there is only one method to learn about instead of two.
 
 ## Implications of this change
 
@@ -27,7 +27,7 @@ The current design has **two separate methods** at each layer:
 
 ### Key Usage Areas Identified
 
-#### 1. **ChatAgent** (_agents.py and _agents.py)
+#### 1. **ChatAgent** (_agents.py)
 - `run()` → calls `self.chat_client.get_response()`
 - `run_stream()` → calls `self.chat_client.get_streaming_response()`
 
@@ -71,7 +71,7 @@ original_get_response = chat_client.get_response
 
 #### 6. **Provider Implementations** (all subpackages)
 All subclasses implement both `_inner_*` methods, except:
-- OpenAI Assistants Client - it implements `_inner_get_response` by calling `_inner_get_streaming_response`
+- OpenAI Assistants Client (and similar clients, such as Foundry Agents V1) - it implements `_inner_get_response` by calling `_inner_get_streaming_response`
 
 ### Implications of Consolidation
 
@@ -84,20 +84,6 @@ All subclasses implement both `_inner_*` methods, except:
 | **Samples/Tests** | Many samples call `get_streaming_response()` directly - would need updates. |
 | **Protocol Simplification** | `ChatClientProtocol` goes from 2 methods + 1 property to 1 method + 1 property. |
 
-### Suggested Migration Path
-
-1. **Keep public `get_streaming_response` as an alias** for backward compatibility:
-   ```python
-   def get_streaming_response(self, messages, **kwargs):
-       return self.get_response(messages, stream=True, **kwargs)
-   ```
-
-2. **Consolidate inner methods first**: Have `_inner_get_response(stream=True/False)` in `BaseChatClient`, allowing subclasses to gradually migrate.
-
-3. **Update decorators** to handle the unified method with conditional streaming logic.
-
-4. **Deprecation warnings** on direct `get_streaming_response()` calls.
-
 ### Recommendation
 
 The consolidation makes sense architecturally, but consider:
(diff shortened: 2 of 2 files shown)


eavanvalkenburg:
CC: @brettcannon @johanste

docs/decisions/0012-python-get-response-simplification.md
```diff
@@ -0,0 +1,140 @@
...
+| **BaseChatClient** | `get_response()` (public) | `get_streaming_response()` (public) |
+| **Implementation** | `_inner_get_response()` (private) | `_inner_get_streaming_response()` (private) |
+
+### Key Usage Areas Identified
+
+#### 1. **ChatAgent** (_agents.py and _agents.py)
+- `run()` → calls `self.chat_client.get_response()`
+- `run_stream()` → calls `self.chat_client.get_streaming_response()`
+
+These are parallel methods on the agent, so consolidating the client methods would **not break** the agent API. You could keep `agent.run()` and `agent.run_stream()` unchanged while internally calling `get_response(stream=True/False)`.
```
dmytrostruk, line 34: If we unify `get_response` method to support both streaming and non-streaming scenarios, I'm wondering if we should do the same for Agent? Basically, remove `agent.run_stream()` method and put `stream` boolean parameter to `agent.run()` method?

docs/decisions/00XX-python-get-response-simplification.md
```diff
@@ -0,0 +1,140 @@
...
+#### 5. **AG-UI Client** (_client.py)
+Wraps both methods to unwrap server function calls:
+```python
+original_get_streaming_response = chat_client.get_streaming_response
+original_get_response = chat_client.get_response
+```
+
+#### 6. **Provider Implementations** (all subpackages)
+All subclasses implement both `_inner_*` methods, except:
+- OpenAI Assistants Client - it implements `_inner_get_response` by calling `_inner_get_streaming_response`
```
dmytrostruk: Azure AI V1 is also implemented in this way, so we can specify it here.

docs/decisions/00XX-python-get-response-simplification.md
```diff
@@ -0,0 +1,140 @@
...
+   ```python
+   @overload
+   async def get_response(self, messages, *, stream: Literal[True], ...) -> AsyncIterable[ChatResponseUpdate]: ...
+   @overload
+   async def get_response(self, messages, *, stream: Literal[False] = False, ...) -> ChatResponse: ...
+   ```
+
+2. **The decorator complexity** is the biggest concern. The current approach of separate decorators for separate methods is cleaner than conditional logic inside one wrapper.
+
+3. **Backward compatibility** should be maintained with deprecation, not immediate removal.
```
dmytrostruk: When exactly we would remove it, before GA or after?

docs/decisions/0012-python-get-response-simplification.md
```diff
@@ -0,0 +1,140 @@
...
+## Option 1: Status Quo
+- Good: Clear separation of streaming vs non-streaming logic
+- Good: Aligned with dotnet design
+- Bad: Code duplication in decorators and middleware
+- Bad: More complex client implementations
+
+## Option 2: Consolidate into Single Method
+- Good: Simplified public API for chat clients
+- Good: Reduced code duplication in decorators
+- Bad: Increased complexity in decorators and middleware
```
dmytrostruk, line 124: Another disadvantage: won't be aligned with .NET design. Let's discuss it with broader audience and see if this acceptable for the team.

docs/decisions/0012-python-get-response-simplification.md
eavanvalkenburg: yeah, I'm wondering the same and wanted to bring that up, I think since the assumption is that ChatClients will be used mostly by Agents, they are different, the Agent's ergonomics should be more developer focused, while ChatClient can be a bit less easy to use, but would love for other people to chime in.

docs/decisions/00XX-python-get-response-simplification.md
```diff
@@ -0,0 +1,140 @@
...
+| **Type Safety** | Overloads work well: `@overload` with `Literal[True]` → `AsyncIterable`, `Literal[False]` → `ChatResponse`. Runtime return type based on `stream` param. |
+| **Breaking Change** | **Major breaking change** for anyone implementing custom chat clients. They'd need to update from 2 methods to 1 (or 2 inner methods to 1). |
+| **Decorator Complexity** | All 3 decorator systems (function invocation, middleware, observability) would need refactoring to handle both paths in one wrapper. |
+| **Code Reduction** | Significant reduction in _tools.py (~200 lines of near-duplicate code) and other decorators. |
+| **Samples/Tests** | Many samples call `get_streaming_response()` directly - would need updates. |
+| **Protocol Simplification** | `ChatClientProtocol` goes from 2 methods + 1 property to 1 method + 1 property. |
+
+### Suggested Migration Path
+
+1. **Keep public `get_streaming_response` as an alias** for backward compatibility:
```
markwallace-microsoft: Given we are not GA is backward compatibility a concern?

If we want to keep this as a convenience method somewhere that's fine but should this be on the protocol?

docs/decisions/0012-python-get-response-simplification.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Option 1: Status Quo
+- Good: Clear separation of streaming vs non-streaming logic
+- Good: Aligned with dotnet design
+- Bad: Code duplication in decorators and middleware
+- Bad: More complex client implementations
+
+## Option 2: Consolidate into Single Method
+- Good: Simplified public API for chat clients
+- Good: Reduced code duplication in decorators
```
markwallace-microsoft, line 121: Who do these benefits accrue to?

If it is just our team and the very small number of people who might build a chat client or a decorator?

docs/decisions/0012-python-get-response-simplification.md
eavanvalkenburg: Added this addition as a separate option

docs/decisions/00XX-python-get-response-simplification.md
eavanvalkenburg: removed this (AI slop :D))

docs/decisions/0012-python-get-response-simplification.md
eavanvalkenburg: added

docs/decisions/00XX-python-get-response-simplification.md
eavanvalkenburg: removed this

docs/decisions/0012-python-get-response-simplification.md
eavanvalkenburg: the developer experience is also improved, added a note on that and more complete code sample

docs/decisions/00XX-python-get-response-simplification.md
```diff
@@ -0,0 +1,245 @@
...
+
+## Option 2: Consolidate into Single Method
+- Good: Simplified public API for chat clients
+- Good: Reduced code duplication in decorators
+- Good: Smaller API footprint for users to get familiar with
+- Good: People using OpenAI directly already expect this pattern
+- Bad: Increased complexity in decorators and middleware
+- Bad: Less alignment with .NET design
+
+## Option 3: Consolidate + Merge Agent Methods
```
moonbox3: I like the idea of a single method and passing in stream=True|False.

docs/decisions/0012-python-get-response-simplification.md
```diff
@@ -0,0 +1,245 @@
...
+
+    # Non-streaming agent
+    print(f"\nUser: {message}")
+    result = await agent.run(message, thread=thread) # default would be stream=False
+    print(f"{agent.name}: {result.text}")
+
+    # Streaming agent - same method, different parameter
+    print(f"\nUser: {message}")
+    print(f"{agent.name}: ", end="")
+    async for update in agent.run(message, thread=thread, stream=True):
```
moonbox3, line 213: Yeah, I like this...

docs/decisions/0012-python-get-response-simplification.md
```diff
@@ -0,0 +1,245 @@
...
+
+This would **reduce code duplication** but add complexity to a single function.
+
+#### 3. **Observability/Instrumentation** (observability.py)
+Same pattern as function invocation:
+- `_trace_get_response()` wraps `get_response`
+- `_trace_get_streaming_response()` wraps `get_streaming_response`
+- `use_instrumentation` decorator applies both
+
+**Impact**: Would need consolidation into a single tracing wrapper.
```
TaoChenOSU, line 60: Could we have a single wrapper but internally split into two for streaming and non-streaming? Essentially maintaining the status quote of the internals.

docs/decisions/0012-python-get-response-simplification.md
```diff
@@ -0,0 +1,245 @@
...
+
+
+async def main() -> None:
+    # Example 1: Direct ChatClient usage with single method
+    client = OpenAIChatClient()
+    message = "What's the weather in Amsterdam and in Paris?"
+
+    # Non-streaming usage
+    print(f"User: {message}")
+    response = await client.get_response(message, tools=get_weather)
```
TaoChenOSU, line 185: The default mode is non-streaming. Do we know if non-streaming is used more frequently than streaming? Perhaps something to think about: frequency of use v.s. easy to get started.

docs/decisions/0012-python-get-response-simplification.md
TaoChenOSU: Same for the other decorators.

eavanvalkenburg: all the decorators work on the public api's so those would have to be changed, we could decide to do some on the _inner methods and keep those split between streaming and non-streaming, but I would prefer to go allout, because then the code duplication is reduced by quite a bit and it becomes much simpler to understand the subtle differences between streaming and non-streaming behavior, or put differently it becomes much easier to have consistent behavior.

docs/decisions/0012-python-get-response-simplification.md
eavanvalkenburg: yeah, good question, I think ease of first use should win out here, and default=False, but good to have that discussion.

eavanvalkenburg: added this question

docs/decisions/0012-python-get-response-simplification.md
```diff
@@ -0,0 +1,255 @@
...
+- Good: People using OpenAI directly already expect this pattern
+- Good: Workflows internally already use a single method (_run_workflow_with_tracing), so would eliminate public API duplication as well, with hardly any code changes
+- Bad: More breaking changes for agent users
+- Bad: Increased complexity in agent implementation
+- Bad: More extensive misalignment with .NET design (`run(stream=True)` vs `RunStreamingAsync` in addition to `get_response` change)
+
+## Misc
+
+Smaller questions to consider:
+- Should default be `stream=False` or `stream=True`? (Current is False)
```
moonbox3, line 140: I think defaulting to False makes sense.

dmytrostruk (approved):

markwallace-microsoft (approved):

PR #3131, opened 2026-01-08
Python: .NET: Executor source gen for workflow executor routing

alliscode:
This pull request introduces a new Roslyn source generator project for the Microsoft Agent Framework Workflows, enabling compile-time route configuration for executor classes using the `[MessageHandler]` attribute. It includes the generator implementation, diagnostics, build configuration, and integration into the solution and dependency management.

---
**Follow-up PRs will update samples and docs to use this new pattern and obsolete the ReflectingExecutor**
---

**New Source Generator Project:**

- Added new project `Microsoft.Agents.AI.Workflows.Generators` targeting `netstandard2.0`, with all necessary build and NuGet packaging settings, and Roslyn dependencies for source generator development. [[1]](diffhunk://#diff-a113b457ad34bbca60cbdb5d3ee3da37f95f0046176e9be0d4086ce2a71c02dcR1-R59) [[2]](diffhunk://#diff-ce4b94dd4361b1762fdc2ee5af2f4579ec918171ab6213e27c192902e5fb7633R1-R18)

**Source Generator Implementation:**

- Implemented the `ExecutorRouteGenerator` class, a Roslyn incremental source generator that scans for `[MessageHandler]` methods and generates partial classes to configure workflow routes at compile time.
- Added the `SourceBuilder` utility to generate the actual C# code for route configuration, including overrides for `ConfigureRoutes`, `ConfigureSentTypes`, and `ConfigureYieldTypes` as needed.

**Diagnostics and Analysis:**

- Introduced `DiagnosticDescriptors` to define and register custom diagnostics for common usage errors (e.g., missing parameters, invalid return types, non-partial classes, etc.) in handler methods and executor classes.

**Solution and Dependency Integration:**

- Registered the new generator project and its corresponding unit test project in the solution file (`agent-framework-dotnet.slnx`). [[1]](diffhunk://#diff-adb6013bde168337770311d83d471b3c9baa6b8fdb9208d5af0ea4fa085fb7d3R399) [[2]](diffhunk://#diff-adb6013bde168337770311d83d471b3c9baa6b8fdb9208d5af0ea4fa085fb7d3R439)
- Updated `Directory.Packages.props` to include `Microsoft.CodeAnalysis.Analyzers` as a dependency for code analysis support.### Motivation and Context

### Contribution Checklist


dotnet/src/Microsoft.Agents.AI.Workflows.Generators/Analysis/SemanticAnalyzer.cs
```diff
@@ -0,0 +1,550 @@
...
+/// <list type="number">
+/// <item><see cref="AnalyzeMethod"/> - Called per method, extracts data and performs method-level validation only.</item>
+/// <item><see cref="CombineMethodResults"/> - Groups methods by class and performs class-level validation once.</item>
+/// </list>
+/// This avoids redundant class validation when multiple handlers exist in the same class.
+/// </remarks>
+internal static class SemanticAnalyzer
+{
+    // Fully-qualified type names used for symbol comparison
+    private const string ExecutorTypeName = "Microsoft.Agents.AI.Workflows.Executor";
```
lokitoth, line 30: [Minor] We may want to use `typeof(<T>).FullName` here, rather than hardcoding the strings, to help in refactor renames (though a lot of the tools are decent about finding the names in comments / strings, too)

dotnet/src/Microsoft.Agents.AI.Workflows.Generators/Analysis/SemanticAnalyzer.cs
```diff
@@ -0,0 +1,550 @@
...
+        if (classSymbol is null)
+        {
+            return CreateEmptyResult();
+        }
+
+        // Get the method syntax for location info
+        var methodSyntax = context.TargetNode as MethodDeclarationSyntax;
+
+        // Extract class-level info (raw facts, no validation here)
+        var classKey = GetClassKey(classSymbol);
```
lokitoth: When type name is not available on the RHS, consider using the type rather than `var`, as it makes it easier for readers without an IDE to consume the code.

dotnet/src/Microsoft.Agents.AI.Workflows.Generators/Diagnostics/DiagnosticDescriptors.cs
```diff
@@ -0,0 +1,107 @@
...
+        s_descriptorsById[descriptor.Id] = descriptor;
+        return descriptor;
+    }
+
+    /// <summary>
+    /// MAFGENWF001: Handler method must have IWorkflowContext parameter.
+    /// </summary>
+    public static readonly DiagnosticDescriptor MissingWorkflowContext = Register(new(
+        id: "MAFGENWF001",
+        title: "Handler missing IWorkflowContext parameter",
```
lokitoth, line 36: We don't need to localize these strings, right?

dotnet/src/Microsoft.Agents.AI.Workflows.Generators/Analysis/SemanticAnalyzer.cs
```diff
@@ -0,0 +1,550 @@
...
+            return AnalysisResult.WithInfoAndDiagnostics(executorInfo, allDiagnostics.ToImmutable());
+        }
+
+        return AnalysisResult.Success(executorInfo);
+    }
+
+    /// <summary>
+    /// Creates a placeholder result for invalid targets (e.g., attribute on non-method).
+    /// </summary>
+    private static MethodAnalysisResult CreateEmptyResult()
```
lokitoth: Consider moving this to MethodAnalysisResult as a static `Empty()` method/property, similar to `ImmutableEquatableArray<T>.Empty`

dotnet/src/Microsoft.Agents.AI.Workflows.Generators/ExecutorRouteGenerator.cs
```diff
@@ -0,0 +1,100 @@
...
+[Generator]
+public sealed class ExecutorRouteGenerator : IIncrementalGenerator
+{
+    private const string MessageHandlerAttributeFullName = "Microsoft.Agents.AI.Workflows.MessageHandlerAttribute";
+
+    /// <inheritdoc/>
+    public void Initialize(IncrementalGeneratorInitializationContext context)
+    {
+        // Step 1: Use ForAttributeWithMetadataName to efficiently find methods with [MessageHandler] attribute. For each method found, build a MethodAnalysisResult.
+        var methodAnalysisResults = context.SyntaxProvider
```
lokitoth: Rooting the analysis on the methods annotated by `[MessageHandler]` is likely simpler than starting with `extends(Executor)`, but it has the drawback that if we only want to use SourceGeneration for the Send/Yield types, and manually create the routes, we ignore that `Executor` class;

We should also do a bit of perf work here, after we settle on the API surface (attributes, what gets generated / not).

dotnet/src/Microsoft.Agents.AI.Workflows.Generators/Generation/SourceBuilder.cs
```diff
@@ -0,0 +1,247 @@
...
+                sb.AppendLine($"{indent}{{");
+                indent += "    ";
+            }
+        }
+
+        // Class declaration
+        sb.AppendLine($"{indent}partial class {info.ClassName}{info.GenericParameters}");
+        sb.AppendLine($"{indent}{{");
+
+        var memberIndent = indent + "    ";
```
lokitoth: We may want to consider using the `AnalyzerConfigOptionsProvider` mechanism to get the indent style / spacecount from .editorconfig, so that the generated code matches the general project style.

dotnet/src/Microsoft.Agents.AI.Workflows.Generators/Analysis/SemanticAnalyzer.cs
```diff
@@ -0,0 +1,550 @@
...
+    /// <example>
+    /// [SendsMessage(typeof(Request))]
+    /// [YieldsMessage(typeof(Response))]
+    /// public partial class MyExecutor : Executor { }
+    /// </example>
+    private static ImmutableEquatableArray<string> GetClassLevelTypes(INamedTypeSymbol classSymbol, string attributeName)
+    {
+        var builder = ImmutableArray.CreateBuilder<string>();
+
+        foreach (var attr in classSymbol.GetAttributes())
```
lokitoth: Does this always get iterated in the same order? 

Since we are creating immutable arrays that are presumably used as some form of comparator somewhere, do we want/need to ensure this is permutation invariant?

dotnet/src/Microsoft.Agents.AI.Workflows.Generators/Generation/SourceBuilder.cs
```diff
@@ -0,0 +1,247 @@
...
+            for (int i = 0; i < info.Handlers.Count; i++)
+            {
+                var handler = info.Handlers[i];
+                var isLast = i == info.Handlers.Count - 1;
+
+                sb.Append($"{bodyIndent}    .AddHandler");
+                AppendHandlerGenericArgs(sb, handler);
+                sb.Append($"(this.{handler.MethodName})");
+
+                if (isLast)
```
lokitoth: If you pull this out of the loop you can avoid the conditional and just do `sb.AppendLine(";");`.

dotnet/tests/Microsoft.Agents.AI.Workflows.Generators.UnitTests/ExecutorRouteGeneratorTests.cs
```diff
@@ -0,0 +1,916 @@
...
+        result.RunResult.Diagnostics.Should().BeEmpty();
+
+        var generated = result.RunResult.GeneratedTrees[0].ToString();
+
+        // Both handlers from different files should be registered
+        generated.Should().Contain(".AddHandler<string>(this.HandleFromFile1)");
+        generated.Should().Contain(".AddHandler<int>(this.HandleFromFile2)");
+    }
+
+    #endregion
```
lokitoth, line 767: We should also test that having class-level Send/Yield annotations mixed between multiple parts of a partial class works correctly

dotnet/src/Microsoft.Agents.AI.Workflows/Attributes/YieldsMessageAttribute.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// Gets the type of message that the executor may yield.
+    /// </summary>
+    public Type Type { get; }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="YieldsMessageAttribute"/> class.
+    /// </summary>
+    /// <param name="type">The type of message that the executor may yield.</param>
+    /// <exception cref="ArgumentNullException"><paramref name="type"/> is <see langword="null"/>.</exception>
+    public YieldsMessageAttribute(Type type)
```
lokitoth: Consider renaming this to `YieldsOutput` to be consistent with the API on `IWorkflowContext`

dotnet/src/Microsoft.Agents.AI.Workflows.Generators/Analysis/SemanticAnalyzer.cs
alliscode: I tried to do this, but we can't reasonably reference the workflows project from the source gen project.

dotnet/src/Microsoft.Agents.AI.Workflows.Generators/ExecutorRouteGenerator.cs
alliscode: Agreed that we should update this to handle Send/Yield types even with manual routing, but I don't think we'd want to root on extends(Executor). Rooting on an attribute with `ForAttributeWithMetadataName` [claims to be at least 99x faster](https://github.com/dotnet/roslyn/blob/main/docs/features/incremental-generators.cookbook.md#use-forattributewithmetadataname).

dotnet/src/Microsoft.Agents.AI.Workflows.Generators/Generation/SourceBuilder.cs
alliscode: Interesting idea. By default, these generated "files" are not written to actual files and even in VS/Code you have to look in the package dependency to see the virtual file. I'm not sure I see the value in trying to match spacecount etc.

dotnet/src/Microsoft.Agents.AI.Workflows.Generators/Analysis/SemanticAnalyzer.cs
```diff
@@ -0,0 +1,550 @@
...
+        foreach (var attr in methodSymbol.GetAttributes())
+        {
+            if (attr.AttributeClass?.ToDisplayString() != MessageHandlerAttributeName)
+            {
+                continue;
+            }
+
+            foreach (var namedArg in attr.NamedArguments)
+            {
+                if (namedArg.Key == "Yield" && !namedArg.Value.IsNull)
```
peibekwe: nit: Should this be a case sensitive comparison just to be safe?

dotnet/src/Microsoft.Agents.AI.Workflows.Generators/Analysis/SemanticAnalyzer.cs
alliscode: Good call, fixed.

dotnet/src/Microsoft.Agents.AI.Workflows.Generators/ExecutorRouteGenerator.cs
alliscode: I've updated things to handle this case while still rooting on `ForAttributeWithMetadataName `.

dotnet/src/Microsoft.Agents.AI.Workflows.Generators/Analysis/SemanticAnalyzer.cs
alliscode: Good call out. They are consistent (source order) for a given file, but the partial classes may break that. I've added explicit sort to this and other places that have this issue.

dotnet/src/Microsoft.Agents.AI.Workflows.Generators/Generation/SourceBuilder.cs
alliscode: Pulled it out and added a line to clean up the extra newline

dotnet/tests/Microsoft.Agents.AI.Workflows.Generators.UnitTests/ExecutorRouteGeneratorTests.cs
alliscode: Verified with tests, good call.

dotnet/src/Microsoft.Agents.AI.Workflows.Generators/ExecutorRouteGenerator.cs
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    /// <summary>
+    /// Generates a hint (virtual file) name for the generated source file based on the ExecutorInfo.
+    /// </summary>
+    private static string GetHintName(ExecutorInfo info)
+    {
+        var sb = new StringBuilder();
+
+        if (!string.IsNullOrEmpty(info.Namespace))
```
peibekwe: nit: Probably better to use IsNullOrWhitespace or is whitespace a valid namespace?

dotnet/src/Microsoft.Agents.AI.Workflows.Generators/Generation/SourceBuilder.cs
```diff
@@ -0,0 +1,253 @@
...
+        sb.AppendLine();
+
+        // Using directives
+        sb.AppendLine("using System;");
+        sb.AppendLine("using System.Collections.Generic;");
+        sb.AppendLine("using Microsoft.Agents.AI.Workflows;");
+        sb.AppendLine();
+
+        // Namespace
+        if (!string.IsNullOrEmpty(info.Namespace))
```
peibekwe: Same comment for this.

peibekwe (approved):

dotnet/src/Microsoft.Agents.AI.Workflows.Generators/Generation/SourceBuilder.cs
lokitoth: Makes sense.

dotnet/src/Microsoft.Agents.AI.Workflows.Generators/Diagnostics/DiagnosticDescriptors.cs
alliscode: Correct

lokitoth (approved):

PR #3149, opened 2026-01-09
.NET: fix: AIAgent - DebuggerDisplay using a non-existing Property

rwjdk:
### Motivation and Context

In a recent update, the property `DisplayName` was removed from AIAgent.cs, but the property is still used in the class DebuggerDisplay attribute.

Because of this, when debugging, you see this when hovering the mouse over the agent:

image

### Description

This PR fixes this by introducing a private DebuggerDisplay (like many other classes in AF do) that either shows Name + Id (if name is provided)

image

or just the Id if the Agent does not have a name

image

### Contribution Checklist


rwjdk:
Can see this is a duplicate of https://github.com/microsoft/agent-framework/pull/2985 so properly not needed

Issue #2520, opened 2025-11-28
.NET: Consider making ChatMessageStore and AIContextProvider Factories plus GetNewThread and DesriailzeThread async


westey-m:
User Scenario:

The developer already has a database full of chat history entries for different threads.
The developer also has a custom ChatMessageStore implementation that can read from this database.
The developer has not used Agent Framework before, so when creating a thread, needs to lookup the appropriate chat history for each user in the database, potentially using the user's name as a secondary index.

Once the developer has bootstrapped an AgentThread, they are happy to store the id from the database with the agent thread state, but to bootstrap it they need to do a lookup.
They want to do this lookup in the ChatMessageStore factory.

Implications:

Changing the factory methods to be async, will also require us to change GetNewThread and DeserializeThread to be async, since both of these call the factory.

PR #3152, opened 2026-01-09
.NET: [BREAKING] Change GetNewThread and DeserializeThread to async

westey-m:
### Motivation and Context

In some cases users need to make async calls in ChatMessageStore and AIContextProvider factories, so we need to change the factory signatures to be async, which means changing GetNewThread and DeserializeThread to async as well.

#2520

### Description

- Change GetNewThread and DeserializeThread to async

### Contribution Checklist

### DIFF
--- a/python/packages/ag-ui/agent_framework_ag_ui/_events.py
+++ b/python/packages/ag-ui/agent_framework_ag_ui/_events.py
@@ -160,10 +160,10 @@ class AgentFrameworkEventBridge:
             logger.warning(f"FunctionCallContent missing name and call_id. args_length={args_length}")
 
         tool_call_id = self._coalesce_tool_call_id(content)
+        # Only emit ToolCallStartEvent once per tool call (when it's a new tool call)
         if content.name and tool_call_id != self.current_tool_call_id:
             self.streaming_tool_args = ""
             self.state_delta_count = 0
-        if content.name:
             self.current_tool_call_id = tool_call_id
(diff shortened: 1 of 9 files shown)


dotnet/samples/GettingStarted/Agents/Agent_Step13_BackgroundResponsesWithToolsAndPersistence/Program.cs
```diff
@@ -58,13 +58,15 @@ void PersistAgentState(AgentThread thread, ResponseContinuationToken? continuati
     stateStore["continuationToken"] = JsonSerializer.SerializeToElement(continuationToken, AgentAbstractionsJsonUtilities.DefaultOptions.GetTypeInfo(typeof(ResponseContinuationToken)));
 }
 
-void RestoreAgentState(AIAgent agent, out AgentThread thread, out ResponseContinuationToken? continuationToken)
+async Task<(AgentThread Thread, ResponseContinuationToken? ContinuationToken)> RestoreAgentState(AIAgent agent)
```
markwallace-microsoft, line 61: Should these parameter names be capitalised?

westey-m: Yes, this is the recommended style for tuples, since while these look like parameters, they are really defining properties.

PR #3177, opened 2026-01-12
Python: ADR for Client Composition

eavanvalkenburg:
### Contribution Checklist

### DIFF
--- a/docs/decisions/00XX-python-client-agent-composition.md
+++ b/docs/decisions/00XX-python-client-agent-composition.md
@@ -10,9 +10,11 @@ consulted: taochenosu, moonbox3, dmytrostruk, giles17
 
 ## Context and Problem Statement
 
-In Python we currently use a set of decorators that can be applied to ChatClients and Agents, those are for function calling, telemetry and middleware. However we currently do not allow a user to compose these themselves, for example to create a ChatClient that does not do function calling, but does have tools being passed to a API. Or to only have telemetry enabled on a chat client, but not on the agent. Up to this point, that has been a sensible decision because it makes getting started very easy. However as we add more features, and more ways to customize the behavior of clients and agents, this becomes a limitation.
+In Python we currently use a set of decorators that can be applied to ChatClients and Agents: function calling, telemetry, and middleware. However, we do not allow users to compose these themselves. For example, a user cannot create a ChatClient that passes tools to an API but doesn't invoke them locally, or only enable telemetry on a chat client but not on the agent.
 
-We have also seen latency issues, and every decorator adds some overhead, so being able to compose a client or agent with only the features you need would help with that as well, and it will at least make this a very explicit tradeoff. Note all the ChatClientBuilderExtensions in the C# version [here](https://github.com/dotnet/extensions/tree/main/src/Libraries/Microsoft.Extensions.AI/ChatCompletion)
+Up to this point, this has been a sensible decision because it makes getting started very easy. However, as we add more features and more ways to customize behavior, this becomes a limitation.
+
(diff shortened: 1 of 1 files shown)


docs/decisions/00XX-python-client-agent-composition.md
```diff
@@ -0,0 +1,140 @@
...
+### Option 3: Builder pattern with fluent API
+Use a builder class with named methods for each capability. The builder constructs clients through a pipeline pattern.
+
+Example:
+```python
+client = ChatClientBuilder(OpenAIChatClient(...)) \
+    .with_telemetry(logger_factory) \
+    .with_function_calling() \
+    .with_capability(custom_capability) \
+    .build()
```
dmytrostruk: I think that in all these options except current one (Option 1), we will also need to decide which decorators makes sense to enable out of the box and which decorators delegate to the user for registration. For example, when using `OpenAIChatClient`, I think that function calling should be enabled by default. For a couple of reasons:
- We have dedicated parameter `tools` through the codebase and on abstraction level, so it would be strange if it won't work because the decorator is not applied, even though the parameter is there.
- I think OpenAI Chat Client will be used more with `tools` rather than without tools. 

So, maybe it makes sense to use some combined approach where we decide which decorators we enable by default and which we allow users to decide (in which case we expose it based on the approach we select here). 

If it makes sense, it's probably worth to update the document.

docs/decisions/00XX-python-client-agent-composition.md
```diff
@@ -0,0 +1,140 @@
...
+### Option 4: Builder pattern with wrapper-based composition
+Use a builder class with a generic method (e.g., `use`) that accepts capability decorators. Each capability is implemented as a class decorator.
+
+Example:
+```python
+client = ChatClientBuilder(OpenAIChatClient(...)) \
+    .use(TelemetryWrapper(logger_factory)) \
+    .use(FunctionCallingWrapper(...)) \
+    .use(custom_wrapper) \
+    .build()
```
dmytrostruk: Can `use` method accept `ChatClient` implementation instead of `wrapper`, so we can have `FunctionInvokingChatClient`, `TelemetryChatClient` and so on? With this approach, users will operate with chat clients only without a need to learn about new `wrapper` abstraction.

docs/decisions/00XX-python-client-agent-composition.md
```diff
@@ -0,0 +1,140 @@
...
+- Good: can provide sensible defaults while allowing customization
+- Good: can validate ordering and either raise or adjust as needed
+- Bad: all methods must be defined in core builder
+- Bad: method explosion as features grow
+- Bad: more verbose than current approach for simple cases
+- Bad: steeper learning curve compared to current approach
+- Bad: requires new builder abstraction
+- Note: A generic method like `.with(wrapper)` could be added alongside named methods to enable third-party extensibility (combining advantages of Option 4), allowing both discoverable built-in methods and flexible custom wrappers
+
+### Option 4: Builder pattern with wrapper-based composition
```
dmytrostruk: I prefer Option 3 and Option 4, and if choosing between them, I think Option 3 looks more Pythonic but wondering what other people think.

docs/decisions/00XX-python-client-agent-composition.md
eavanvalkenburg: Actually, I've been thinking about doing a CodeMode function calling feature, in which case even that one would be a choice. I'll also check his dotnet handles this, because we don't want a verbose or unexpected getting started experience

docs/decisions/00XX-python-client-agent-composition.md
eavanvalkenburg: We can name the wrappers with names like that? Or is that not what you mean?

Issue #3179, opened 2026-01-12
.NET: Copy AdditionalProperties from AgentRunOptions to ChatOptions in ChatClientAgent


westey-m:
In some cases callers may want to pass additional context to an agent and be able to access that in a function tool.

Currently, if users were to set AdditionalProperties on AgentRunOptions, and call ChatClientAgent, those additional properties are not propogated to the ChatOptions passed to the ChatClient.  You have to break out of the abstraction and set a ChatOptions on ChatClientAgentRunOptions to pass any AdditionalProperties.

We should automatically pass any AdditionalProperties on AgentRunOptions to ChatOptions in ChatClientAgent, so that users don't have to break out of the abstraction to pass additional properties all the way down the stack.

PR #3184, opened 2026-01-12
.NET: Merge AgentRunOptions.AdditionalProperties into ChatOptions.AdditionalProperties

westey-m:
### Motivation and Context

We should pass AgentRunOptions.AdditionalProperties to ChatOptions.AdditionalProperties in ChatClientAgent so that additional properties are replicated all the way down the stack.

The allows us to set a property on AgentRunOptions.AdditionalProperties without needing to break out of the abstraction and then access the property in a function tool via `FunctionInvokingChatClient.CurrentContext.Options?.AdditionalProperties`

#3179

### Description

- Replicate AgentRunOptions.AdditionalProperties to ChatOptions.AdditionalProperties in ChatClientAgent
- Move unit tests for ChatOptions Merging in a separate class to reduce the size of ChatClientAgentTests
- Modify ChatOptionsMergingPrioritizesRequestOptionsOverAgentOptionsAsync to verify the new scenario

### Contribution Checklist


dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
```diff
@@ -643,6 +643,17 @@ await thread.AIContextProvider.InvokedAsync(new(inputMessages, aiContextProvider
                 chatOptions.ContinuationToken = agentContinuationToken!.InnerToken;
             }
 
+            // Add/Replace any additional properties from the AgentRunOptions, since they should always take precedence.
```
stephentoub, line 646: Just curious what's the rationale for them always taking precedence?

If the AgentRunOptions was actually a ChatClientAgentRunOptions and included a ChatOptions, we're saying that AgentRunOptions.AdditionalProperties takes precedence over ChatClientAgentRunOptions.ChatOptions.AdditionalProperties?

I don't have a reason to disagree, but I'm not sure what that's the obvious ordering?

SergeyMenshykh (approved):

dotnet/tests/Microsoft.Agents.AI.UnitTests/ChatClient/ChatClientAgent_ChatOptionsMergingTests.cs
```diff
@@ -0,0 +1,442 @@
...
+using System.Threading.Tasks;
+using Microsoft.Extensions.AI;
+using Moq;
+
+namespace Microsoft.Agents.AI.UnitTests.ChatClient;
+
+/// <summary>
+/// Contains tests for <see cref="ChatOptions"/> merging in <see cref="ChatClientAgent"/>.
+/// </summary>
+public class ChatClientAgent_ChatOptionsMergingTests
```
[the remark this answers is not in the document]
westey-m: This is not a private class.

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
westey-m: My thinking is that AgentRunOptions and ChatClientAgentRunOptions are more directly visible to callers than ChatClientAgentRunOptions.ChatOptions.  As in, this is the main type that you discover when calling Run*, whereas ChatClientAgentRunOptions.ChatOptions is less discoverable and harder to use.  You need to break out of the abstraction and then ignore the properties on AgentRunOptions directly and use the ones on ChatOptions.  Therefore the top level type should take precedence, since it should be more commonly used.

That said, you could probably argue that the harder to use one should take precedence, since users would need to be more deliberate to use it. 😄

I don't feel strongly about it though, and happy to hear alternative arguments.

dotnet/tests/Microsoft.Agents.AI.UnitTests/ChatClient/ChatClientAgent_ChatOptionsMergingTests.cs
```diff
@@ -0,0 +1,442 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.AI;
+using Moq;
+
+namespace Microsoft.Agents.AI.UnitTests.ChatClient;
```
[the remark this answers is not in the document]
westey-m: Fixed.

dotnet/tests/Microsoft.Agents.AI.UnitTests/ChatClient/ChatClientAgent_ChatOptionsMergingTests.cs
```diff
@@ -0,0 +1,442 @@
...
+            Temperature = 0.3f,
+            AdditionalProperties = new AdditionalPropertiesDictionary { ["key2"] = "request-value", ["key3"] = "request-value" },
+            Instructions = "request instructions"
+        };
+        var agentRunOptionsAdditionalProperties = new AdditionalPropertiesDictionary { ["key3"] = "runoptions-value" };
+        var expectedChatOptionsMerge = new ChatOptions
+        {
+            MaxOutputTokens = 200, // Request value takes priority
+            Temperature = 0.3f, // Request value takes priority
+            // Check that each level of presedence is respected in AdditionalProperties
```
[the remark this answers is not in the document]
westey-m: Fixed.

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
stephentoub: I don't feel strongly about it, either.

My rationale for expecting it to be the other way was that a developer isn't going to be creating the ChatClientAgentRunOptions around a ChatOptions and specifying the same property explicitly in the AdditionalProperties on both, that just wouldn't make sense. My expectation is if they're in this situation, they've been passed the ChatOptions from somewhere higher out and are then wrapping that opaque options bag in a ChatClientAgentRunOptions, adding their own AdditionalProperties to the wrapper, at which point they could possibly conflict. Given that, I _think_ our typical policy would be to give preference to the outermost caller, treating whatever they provide as an override.

Up to you. This is a corner case that's unlikely to happen very frequently if at all.

rogerbarreto: Appreciate what are the thoughts of the `ChatClient` specialization of `AgentRunOptions` actually override the `AgentRunOptions.AdditionalProperties` pointing to the `AgentRunOptions.ChatOptions.AdditionalProperties` instead ? 

Making all those properties `virtual` in the abstraction.

westey-m: @rogerbarreto, one potential issue with this approach is if someone set `ChatClientAgentRunOptions.AllowBackgroundResponses = true` and then `ChatClientAgentRunOptions.ChatOptions = new() { Temperature = 0.5 }`.  This would cause them to essentially lose the first setting.

I think it's probably safer to have a merging mechanism, since if someone did set the same property on both, at least there would be an assumption that only one of them can ultimately be used.

westey-m: I've logged a separate task to consider the override precedence, since it'll be great to get this change merged while we continue to mull this one over.  See #3234.  Right now all overrides in the ChatClientAgent are choosing the AgentRunOptions ones over the ChatClient ones, so it is consistent, and we should change all if we change it.

Issue #526, opened 2025-08-28
.NET: Revisit Agents Level ChatOptions per Request


rogerbarreto:
## Motivation and Context

Currently when we provide ChatOptions into a `ChatClientAgent` for every `RunAsync` we merge the `Agent` level option with the provided `RunAsync` option.

This initially was a good approach when we were just using a per-request basis setting, but it turns out that providers allow the tooling/resources to be provided at the agent creation level which doesn't require re-submitting all the tooling information for each request.

This per-request approach start to get problematic when we also deal with hosted agent providers that only support one way or the other way. 

i.e, OpenAI Assistants do not allow tooling override per-request when tools where provided in the agent creation (OpenAI Assistants).

- https://github.com/dotnet/extensions/issues/6743

## Potential solution

Provide a way to disable this behavior when is not desired via `ccAgentRunOptions.AllowRunMerging` or better approach.

markwallace-microsoft:
Is there anything that needs to be done for v1.0?

Issue #266, opened 2025-07-29
.Net: Investigate how tool definitions should be specified when creating and when invoking an agent.


markwallace-microsoft:
[westey-m](https://github.com/westey-m)(westey) [19 hours ago](https://github.com/microsoft/agent-framework/pull/225#discussion_r2236841100) • 
It feels weird to me that we are advertising the tools when we create the agent, and then also passing them as the default ChatOptions to the Agent, which means they'll get advertised again on each invocation.
I suppose the main (only?) value prop of doing it when you create the agent is to avoid passing them for each invocation, saving some small amount of network bandwidth.
Not that I have a great alternative though...

PR #3218, opened 2026-01-14
.Net: ADR ChatClientAgent Options Per-Request merging optionality

rogerbarreto:
### Motivation and Context

- #526
- #266

This PR introduces an ADR addressing per-request agent configuration advertisement control in `ChatClientAgent`. It proposes solutions for providers that require strict server-side configuration authority (e.g., Azure AI Foundry agents) while maintaining backward compatibility with existing implementations.

### Problem Statement

`ChatClientAgent` currently advertises agent configuration (`AgentOptions`) with each per-request invocation, merging it with `AgentRunOptions`. While this enables per-request configuration overrides in many scenarios, it creates issues for providers where agent definition is immutable and strictly server-side controlled. Re-advertising configuration per-request violates the principle of server-side authority.

### DIFF
--- a/docs/decisions/0013-chatclient-agent-optout-per-request-config.md
+++ b/docs/decisions/0013-chatclient-agent-optout-per-request-config.md
@@ -28,7 +28,81 @@ This behavior makes it difficult for provider-specific agents and decorations to
 
 ## Considered Options
 
-### Option 1: Boolean Property in ChatClientAgentOptions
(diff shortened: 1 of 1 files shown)


rogerbarreto:
Closing this PR as the conclusion for this ADR was to keep as is, as this change clearly tries to accomodate a behavior that for the moment is exclusive of Azure Foundry Agents, which potentially will be more permissive on per-request configuration override.

PR #3219, opened 2026-01-14
.NET: Pass AdditionalProperties from parent to child when exposing an agent as a FunctionTool

westey-m:
### Motivation and Context

#3127

### Description

- Pass AdditionalProperties from parent to child when exposing an agent as a FunctionTool

### Contribution Checklist


dotnet/src/Microsoft.Agents.AI/AgentExtensions.cs
```diff
@@ -73,7 +73,12 @@ async Task<string> InvokeAgentAsync(
             [Description("Input query to invoke the agent.")] string query,
             CancellationToken cancellationToken)
         {
-            var response = await agent.RunAsync(query, thread: thread, cancellationToken: cancellationToken).ConfigureAwait(false);
+            // Propagate any additional properties from the parent agent's run to the child agent if the parent is using a FunctionInvokingChatClient.
+            AgentRunOptions? agentRunOptions = FunctionInvokingChatClient.CurrentContext?.Options?.AdditionalProperties is AdditionalPropertiesDictionary dict
+                ? new AgentRunOptions { AdditionalProperties = dict }
+                : null;
```
rogerbarreto, line 79: We might be missing the `ChatOptions` here.

I would addtionally check if the `agent` is a `ChatClientAgent` and provide the `FunctionInvokingChatClient.CurrentContext?.Options` as `new ChatClientAgentRunOptions { ChatOptions = options }` and if not use the current proposed code approach.

Suggest adding an additional test UT to covering `ChatClientAgent` (and non-chatclientagents, which seems a bit odd as I think this is exclusive to `ChatClientAgents` and `FICC` scenarios..)

westey-m: I'm not sure that we can pick the entire ChatOptions and pass it down.
E.g. we shouldn't be passing tools, conversation id, instructions, AllowBackgroundResponses, ContinuationToken or ResponseFormat, since all of these would be based on the behavior and underlying service of the parent agent, and may cause the child agent to behave incorrectly.

rogerbarreto: > I'm not sure that we can pick the entire ChatOptions and pass it down.
E.g. we shouldn't be passing tools, conversation id, instructions, AllowBackgroundResponses, ContinuationToken or ResponseFormat, since all of these would be based on the behavior and underlying service of the parent agent, and may cause the child agent to behave incorrectly.

I'm trying to understand why `AdditionalProperties` have a different criteria compared to other `Properties`, sound a bit inconsistent.

westey-m: The main difference is that they are not used by the AF/MEAI stack, but customers may need them to pass their data down the stack.

moetarhinikaseya:
I tried this @westey-m  and the E2E Scenario is working perfectly. Please try to merge it ASAP :)

westey-m:
> I tried this @westey-m and the E2E Scenario is working perfectly. Please try to merge it ASAP :)

Thanks for verifying @moetarhinikaseya.  I'm waiting to get some further input from key individuals, which should happen tomorrow.

rogerbarreto (approved):

moetarhinikaseya:
@westey-m Still on track for merging this today ?

PR #3221, opened 2026-01-14
.NET: [BREAKING] Migrate ChatMessageStores to be AIContextProviders and rename to ChatHistoryProviders

westey-m:
### Motivation and Context

ChatMessageStore is currently very similar to AIContextProvider, and just provides a subset of functionality. To reduce the concept count, we can model ChatMessageStore as an AIContextProvider.
The ChatMessageStore name is also problematic since it creates the impression that it is a full CRUD abstraction for messages storage, when it is only intended to provide chat history and store new messages for an agent.

#2542

### Description

- Replace ChatMessageStore with AIContextProvider.
- Rename ChatMessageStore implementations to ChatHistoryProvider

### Contribution Checklist


SergeyMenshykh (approved):

dotnet/src/Microsoft.Agents.AI.Abstractions/AIContextProvider.cs
```diff
@@ -176,6 +176,19 @@ public InvokedContext(IEnumerable<ChatMessage> requestMessages, IEnumerable<Chat
...
+        /// </summary>
+        /// <remarks>
+        /// Note that if chat history is stored in the underlying AI service, this property will be null.
+        /// Only chat history retrieved via a chat history provider will be provided here.
+        /// </remarks>
+        /// <value>
+        /// A collection of <see cref="ChatMessage"/> instances that were retrieved from the chat history,
+        /// and were used by the agent as part of the invocation.
+        /// </value>
+        public IEnumerable<ChatMessage>? ChatHistoryMessages { get; set; }
```
stephentoub, line 190: What is the difference between this and AIContextProviderMessages? If the idea is that a chat message store is now a context provider, then isn't such a context provider's messages the history?

dotnet/src/Microsoft.Agents.AI.Abstractions/AIContextProviderExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+    /// Adds message filtering to an existing <see cref="AIContextProvider"/>, so that data passed to to and from it
+    /// can be filtered, updated or replaced.
+    /// </summary>
+    /// <param name="innerAIContextProvider">The underlying AI context provider to be wrapped. Cannot be null.</param>
+    /// <param name="invokingContextFilter">An optional filter function to apply to the AI context before it is returned. If null, no filter is applied at this
+    /// stage.</param>
+    /// <param name="invokedContextFilter">An optional filter function to apply to the invocation context before it is consumed. If null, no
+    /// filter is applied at this stage.</param>
+    /// <returns>The <see cref="AIContextProvider"/> with filtering applied.</returns>
+    public static AIContextProvider WithMessageFilters(
```
stephentoub, line 22: Is this effectively creating a pipeline of providers as decorators?

dotnet/src/Microsoft.Agents.AI.Abstractions/InMemoryChatHistoryProvider.cs
```diff
@@ -29,40 +29,40 @@ namespace Microsoft.Agents.AI;
 /// </remarks>
 [DebuggerDisplay("Count = {Count}")]
 [DebuggerTypeProxy(typeof(DebugView))]
-public sealed class InMemoryChatMessageStore : ChatMessageStore, IList<ChatMessage>, IReadOnlyList<ChatMessage>
+public sealed class InMemoryChatHistoryProvider : AIContextProvider, IList<ChatMessage>, IReadOnlyList<ChatMessage>
```
stephentoub, line 32: As an AIContextProvider, the intent then is folks will use InMemoryChatHistoryProvider places they're using AIContextProviders? Is there an example of when you'd do that?

PR #3369, opened 2026-01-22
.NET: Added ShellTool and LocalShellExecutor

dmytrostruk:
### Motivation and Context

This PR adds shell command execution capabilities to the Agent Framework through a new `ShellTool` abstraction and `LocalShellExecutor` implementation.

- Added `ShellTool` abstraction for shell command execution with configurable security policies (Microsoft.Agents.AI.Abstractions)
- Added `LocalShellExecutor` (Microsoft.Agents.AI.Shell.Local)
- Security controls: privilege escalation blocking, dangerous pattern detection, path validation
- Support for allowlist/denylist patterns
- Configurable timeouts, output truncation, and working directory
- `AsAIFunction` conversion for use with existing AI agents

### Contribution Checklist


dotnet/src/Microsoft.Agents.AI.Abstractions/Shell/ShellToolExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+    /// <returns>An <see cref="AIFunction"/> that wraps the shell tool.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="shellTool"/> is null.</exception>
+    /// <remarks>
+    /// <para>
+    /// The returned <see cref="AIFunction"/> accepts a <c>commands</c> parameter which is an array of
+    /// shell commands to execute. The function returns a <see cref="ShellResultContent"/> containing
+    /// the output for each command.
+    /// </para>
+    /// </remarks>
+    public static AIFunction AsAIFunction(this ShellTool shellTool)
```
westey-m, line 29: Why is ShellTool not inheriting from AIFunction if we want to have it as an AIFunction anyway?

dotnet/samples/GettingStarted/Agents/Agent_Step21_ShellTool/Program.cs
```diff
@@ -0,0 +1,134 @@
...
+        WorkingDirectory = workingDirectory,
+
+        // Restrict file system access to specific paths
+        AllowedPaths = [workingDirectory],
+
+        // Block access to sensitive paths (takes priority over AllowedPaths)
+        // BlockedPaths = ["/etc", "/var"],
+
+        // Only allow specific commands (regex patterns supported)
+        AllowedCommands = ["^ls", "^dir", "^echo", "^cat", "^type", "^mkdir", "^pwd", "^cd"],
```
westey-m, line 44: Should we vary these by operating system as well? Some of these don't work in the windows command prompt shell.
Also, commands vary by shell type, so on windows powershell supports ls and pwd, but command prompt does not.

dotnet/src/Microsoft.Agents.AI.Abstractions/Shell/ShellExecutorOutput.cs
```diff
@@ -0,0 +1,44 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+namespace Microsoft.Agents.AI;
+
+/// <summary>
+/// Raw output from shell executor.
+/// </summary>
+public sealed class ShellExecutorOutput
```
westey-m, line 8: Also consider ShellExecutorResponse to match AgentRunResponse, and ChatResponse.

dotnet/src/Microsoft.Agents.AI.Shell.Local/LocalShellExecutor.cs
```diff
@@ -0,0 +1,243 @@
...
+            return (shellOverride!, command);
+        }
+
+#if NET
+        if (OperatingSystem.IsWindows())
+        {
+            return ("cmd.exe", $"/c {command}");
+        }
+
+        return ("/bin/sh", $"-c \"{command.Replace("\"", "\\\"")}\"");
```
westey-m, line 190: Do we only want to support cmd and sh, or should we make this configurable?

dotnet/src/Microsoft.Agents.AI.Abstractions/Shell/ShellResultContent.cs
```diff
@@ -0,0 +1,48 @@
...
+using Microsoft.Extensions.AI;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Agents.AI;
+
+/// <summary>
+/// Represents the result of a shell command execution.
+/// </summary>
+[DebuggerDisplay("{DebuggerDisplay,nq}")]
+public sealed class ShellResultContent : AIContent
```
westey-m, line 15: Won't this be provided to the LLM as FunctionResultContent?  What's the idea with inheriting from AIContent?
Similar question with ShellCallContent?

dotnet/src/Microsoft.Agents.AI.Abstractions/Shell/ShellTool.cs
```diff
@@ -0,0 +1,551 @@
...
+/// this tool to an <see cref="AIFunction"/> for use with AI agents.
+/// </para>
+/// </remarks>
+public class ShellTool : AITool
+{
+    private readonly ShellExecutor _executor;
+    private readonly IReadOnlyList<Regex>? _compiledAllowedPatterns;
+    private readonly IReadOnlyList<Regex>? _compiledDeniedPatterns;
+
+    private static readonly string[] s_privilegeEscalationCommands =
```
westey-m, line 35: Should these be configurable in addition to having defaults?  Same for shellWrapperCOmmands and dangerous patterns

dotnet/src/Microsoft.Agents.AI.Abstractions/Shell/ShellTool.cs
```diff
@@ -0,0 +1,551 @@
...
+    [
+        "sudo",
+        "su",
+        "runas",
+        "doas",
+        "pkexec"
+    ];
+
+    private static readonly string[] s_shellWrapperCommands =
```
westey-m, line 44: Should we also have `powershell` and `pswh`?

dotnet/src/Microsoft.Agents.AI.Abstractions/Shell/ShellToolOptions.cs
```diff
@@ -0,0 +1,127 @@
...
+/// <summary>
+/// Options for configuring shell tool behavior and security.
+/// </summary>
+public class ShellToolOptions
+{
+    /// <summary>
+    /// Gets or sets the working directory for command execution.
+    /// When null, uses the current working directory.
+    /// </summary>
+    public string? WorkingDirectory { get; set; }
```
SergeyMenshykh, line 16: To be on the safe side, I wonder if this property should be mandatory so that users must explicitly provide a working directory, rather than defaulting to Environment.CurrentDirectory. This would ensure users consider where the code is executed and reduce potential security risks.

dotnet/src/Microsoft.Agents.AI.Abstractions/Shell/ShellResultContent.cs
dmytrostruk: So, in current example, we convert `ShellTool` to `AIFunction`, just so it can be used with any provider that supports function calls. But function calling is not the only case.

There are providers that support shell tools out of the box, like OpenAI Responses Client:
https://platform.openai.com/docs/guides/tools-shell

They provide executor as example only, but they do have `"type": "shell_call"` and `"type": "shell_call_output"` message types, so the user can decide how exactly they want to process this. 

That's the primary reason for separate `ShellCallContent` and `ShellResultContent` - for the ability to work independently from function calls.

dotnet/src/Microsoft.Agents.AI.Abstractions/Shell/ShellToolExtensions.cs
dmytrostruk: Good question, see comment: https://github.com/microsoft/agent-framework/pull/3369#discussion_r2717794027

dotnet/src/Microsoft.Agents.AI.Abstractions/Shell/ShellCallContent.cs
```diff
@@ -0,0 +1,43 @@
...
+using Microsoft.Extensions.AI;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Agents.AI;
+
+/// <summary>
+/// Represents a shell command execution request.
+/// </summary>
+[DebuggerDisplay("{DebuggerDisplay,nq}")]
+public sealed class ShellCallContent : AIContent
```
stephentoub, line 15: @jozkee

stephentoub, line 15: Custom AIContent needs to be added into the AgentsAbstractionJsonUtilities' options so that it correctly participates in polymorphic serialization.

dotnet/src/Microsoft.Agents.AI.Abstractions/Shell/ShellResultContent.cs
stephentoub: Is the intent then that ShellCallContent/ShellResultContent will be used to represent the server-side shell calls? That translation would best be done in the IChatClient for responses. We should really look at pushing down these parts (at least the call and result content) of the abstractions, along with a HostedShellTool, into MEAI.

Also, it feels like for such a case the local execution would be handled by a piece of middleware akin to FunctionInvokingChatClient, but a ShellCommandInvokingChatClient, which would look for ShellCallContent and produce ShellResultContent. Or FunctionInvokingChatClient could be taught to do that.

cc: @jozkee

dotnet/src/Microsoft.Agents.AI.Abstractions/Shell/ShellTool.cs
```diff
@@ -0,0 +1,551 @@
...
+/// <para>
+/// ShellTool provides a secure way to execute shell commands with configurable
+/// allowlist/denylist patterns, privilege escalation prevention, and output limits.
+/// </para>
+/// <para>
+/// Use the <see cref="ShellToolExtensions.AsAIFunction"/> extension method to convert
+/// this tool to an <see cref="AIFunction"/> for use with AI agents.
+/// </para>
+/// </remarks>
+public class ShellTool : AITool
```
stephentoub, line 29: If this is invokable, why doesn't it inherit AIFunction?

dotnet/src/Microsoft.Agents.AI.Abstractions/Shell/ShellTool.cs
```diff
@@ -0,0 +1,551 @@
...
+    private readonly IReadOnlyList<Regex>? _compiledDeniedPatterns;
+
+    private static readonly string[] s_privilegeEscalationCommands =
+    [
+        "sudo",
+        "su",
+        "runas",
+        "doas",
+        "pkexec"
+    ];
```
stephentoub, line 42: Is there a threat model for all of this?
cc: @grabyourpitchforks

dotnet/src/Microsoft.Agents.AI.Abstractions/Shell/ShellTool.cs
```diff
@@ -0,0 +1,551 @@
...
+        // rm -rf / variants
+        new Regex(@"rm\s+(-[rRfF]+\s+)*(/|/\*|\*/)", RegexOptions.Compiled, TimeSpan.FromSeconds(1)),
+        // Format filesystem
+        new Regex(@"mkfs\.", RegexOptions.Compiled, TimeSpan.FromSeconds(1)),
+        // Direct disk write
+        new Regex(@"dd\s+.*of=/dev/", RegexOptions.Compiled, TimeSpan.FromSeconds(1)),
+        // Overwrite disk
+        new Regex(@">\s*/dev/sd", RegexOptions.Compiled, TimeSpan.FromSeconds(1)),
+        // chmod 777 /
+        new Regex(@"chmod\s+(-[rR]\s+)?777\s+/", RegexOptions.Compiled, TimeSpan.FromSeconds(1)),
```
stephentoub, line 68: If we're concerned about such dangerous commands, I'm concerned this list is insufficient.

How are we thinking about having this vs having folks treat such invocations as requiring of user approval?

(Regardless, shouldn't "dangerous patterns" be configurable?)

dotnet/src/Microsoft.Agents.AI.Abstractions/Shell/ShellTool.cs
```diff
@@ -0,0 +1,551 @@
...
+    /// <summary>
+    /// Gets the name of the tool.
+    /// </summary>
+    public override string Name => "shell";
+
+    /// <summary>
+    /// Gets the description of the tool.
+    /// </summary>
+    public override string Description =>
+        "Execute shell commands. Returns stdout, stderr, and exit code for each command.";
```
stephentoub, line 96: If this is the description provided to AI, it should be more descriptive / prescriptive, e.g. about what's supported and what's not supported, what language to use, etc. Is this bash? powershell? etc.?

dotnet/src/Microsoft.Agents.AI.Abstractions/Shell/ShellTool.cs
```diff
@@ -0,0 +1,551 @@
...
+                }
+            }
+
+            i++;
+        }
+
+        return false;
+    }
+
+    private static bool ContainsPrivilegeEscalation(string command)
```
stephentoub, line 278: If we're relying on this for security, this needs strenuous review.
cc: @grabyourpitchforks

dotnet/src/Microsoft.Agents.AI.Abstractions/Shell/ShellResultContent.cs
dmytrostruk: > Is the intent then that ShellCallContent/ShellResultContent will be used to represent the server-side shell calls? That translation would best be done in the IChatClient for responses. We should really look at pushing down these parts (at least the call and result content) of the abstractions, along with a HostedShellTool, into MEAI.

Correct, and the idea of adding this to MEAI was my initial plan :) I just added it here first, because I want to test it end-to-end, just to understand if there is anything missing.

dotnet/src/Microsoft.Agents.AI.Abstractions/Shell/ShellTool.cs
dmytrostruk: To be honest, I'm not sure that it will be possible to cover all cases in this PR. The idea here is to add initial set of security validations with a notice that this is recommended to be used only in sandboxed environments with limited access.

dotnet/src/Microsoft.Agents.AI.Abstractions/Shell/ShellTool.cs
dmytrostruk: I decided to keep it separately, because I'm not sure how well it will work with `AIFunction`-related classes like `AIFunctionArguments` and `FunctionInvokingChatClient`. 

My thinking was that if OpenAI Responses API defines `function tools` and `shell tools` as separate entities, then it's probably a good idea to keep it in the same way on abstraction level. 

But I'm looking for the best guidance here.

dotnet/src/Microsoft.Agents.AI.Abstractions/Shell/ShellCallContent.cs
jozkee: If we unsealed FCC/FRC and made this extend that, I think FunctionInvokingChatClient as-is could be able to handle these shell calls.  

Alternatively, I think these shell contents could be completely written in terms of FCC/FRC but it may be too loose i.e. some properties would need to travel in AdditionalProperties.

dotnet/src/Microsoft.Agents.AI.Abstractions/Shell/ShellTool.cs
jozkee: There's value in making this extend AIFunction since it would be needed for handling shell tool calls in FunctionInvokingChatClient.
https://github.com/dotnet/extensions/blob/ab5bdf603e04a6bbef6366cc93e6e4c5a89d29a3/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs#L854-L855
Unless there's a reason to not want shell tool calls handled there.

stephentoub: I'm still unclear on the purpose of ShellTool as it's written. It doesn't derive from AIFunction so it's not invokable as an AITool, but it has invocation APIs so it doesn't seem like it's a marker tool to be recognized by leaf IChatClients and translated into the right underlying tool for that particular API (OpenAI Responses's shell tool, Anthropic's bash tool, etc.)

stephentoub:
I thought about this a bit today, and here's what I think might make sense:
- ShellTool : AIFunction. The tool would itself advertize a schema and description that would allow for arbitrary commands to be handed to it as an argument, such that arbitrary IChatClient implementations would "just work" with it as a shell tool. However, IChatClient implementations backed by a service that has its own notion of a shell tool could instead special-case it (as they special case all the other HostedXxTool types, translating it into usage of the relevant service tool.
- ShellCallContent : FunctionCallContent. The FCC would be for the ShellTool, with the derived type having all the specific information relevant to a service's specialized shell invocation. An IChatClient that doesn't have a special shell tool would just produce the base FunctionCallContent like it would for any other tool. An IChatClient with a special shell tool would instead construct the derived ShellCallContent. When ShellTool is invoked, it would look at the FunctionCallContent; if it's of the derived type, it'd do the more specialized thing, and if it's the base, it would just proceed as would any other AIFunction.
- ShellResultContent : FunctionResultContent. When doing the special thing, ShellTool's InvokeAsync override would return a ShellResultContent rather than just returning the textual results from invoking the shell tool. We'd tweak the handling of AIFunction in FunctionInvokingChatClient so that an AIFunction returning a FunctionResultContent as the result would just use that FRC rather than creating a new one to wrap the result. That's a useful addition in general, as it gives a function more control over the FunctionResultContent instance that's created, and in this case it means the tool can flow the additional data back strongly-typed. The IChatClient that gets a ShellResultContent would translate that back to the appropriate shell result content expected by the service.

@dmytrostruk, @jozkee, thoughts? If you agree this makes sense, something we could prototype quickly?

jozkee:
@stephentoub assuming this implies the types will be moved down into MEAI.Abstractions, that approach makes sense to me.

PR #3442, opened 2026-01-26
.NET: Fix: Checkpoint Deserialization breaks when JSON metadata properties are out of order

### LINKED ISSUE #2962: .NET: Checkpoint JSON deserialization fails when using PostgreSQL jsonb
Hello all!

### Summary

When using _CheckpointManager.CreateJson_ with a persistent store backed by PostgreSQL _jsonb_, workflow resume fails during checkpoint deserialization with the following exception:

```
System.Text.Json.JsonException:
The metadata property is either not supported by the type or is not the first property in the deserialized JSON object.
```

MAF relies on _System.Text.Json_ polymorphic serialization (using _$type_) for checkpoint data.
PostgreSQL _jsonb_ (for example) does not preserve object property order.

MAF currently does not enable _JsonSerializerOptions.AllowOutOfOrderMetadataProperties_, and the _customOptions_ parameter of _CheckpointManager.CreateJson_ is not applied to the internal marshaller options used for checkpoint types. As a result, checkpoint deserialization fails during resume.

### Reprod

1. Use _CheckpointManager.CreateJson_ with a custom _ICheckpointStore<JsonElement>_
2. Persist checkpoint JSON into PostgreSQL using a _jsonb_ column
3. Resume a workflow run
4. _ResumeStreamAsync_ fails with a _JsonException_ related to _$type_ metadata ordering

### Expected

Checkpoint deserialization should succeed regardless of JSON property ordering, since checkpoints are persisted internal data rather than user-provided payloads.

### Reality

_ResumeStreamAsync_ throws during checkpoint restore when the _$type_ metadata property is not the first JSON property.

### Current workaround

When using PostgreSQL, storing checkpoint data in a _json_ or _text_ column instead of _jsonb_ avoids this issue.

lokitoth:
### Motivation and Context

JSON Checkpoint serialization is currently strict in the ordering of the metadata properties, such as the polymorphic type specifier. This causes problems when deserializing checkpoints stored as JSONB in Postgres, which provides a different order on read.

### Description

Create a way to let the user specify that a relaxed ordering is also allowed.

- Fixes #2962 

### Contribution Checklist


dotnet/src/Microsoft.Agents.AI.Workflows/CheckpointManager.cs
```diff
@@ -42,10 +42,11 @@ internal CheckpointManager(ICheckpointManager impl)
     /// <param name="store">The checkpoint store to use for persisting and retrieving checkpoint data as JSON elements. Cannot be null.</param>
     /// <param name="customOptions">Optional custom JSON serializer options to use for serialization and deserialization. Must be provided if
     /// using custom types in messages or state.</param>
+    /// <param name="checkpointOptions">Optional checkpoint manager options to configure JSON serialization behavior.</param>
     /// <returns>A CheckpointManager instance configured to serialize checkpoint data as JSON.</returns>
-    public static CheckpointManager CreateJson(ICheckpointStore<JsonElement> store, JsonSerializerOptions? customOptions = null)
+    public static CheckpointManager CreateJson(ICheckpointStore<JsonElement> store, JsonSerializerOptions? customOptions = null, JsonCheckpointManagerOptions? checkpointOptions = null)
     {
-        JsonMarshaller marshaller = new(customOptions);
+        JsonMarshaller marshaller = new(customOptions, checkpointOptions);
```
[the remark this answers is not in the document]
lokitoth: This is explicitly tested in the added test. Pure-passthrough tests are lower value.

dotnet/src/Microsoft.Agents.AI.Workflows/CheckpointManager.cs
```diff
@@ -42,10 +42,11 @@ internal CheckpointManager(ICheckpointManager impl)
     /// <param name="store">The checkpoint store to use for persisting and retrieving checkpoint data as JSON elements. Cannot be null.</param>
     /// <param name="customOptions">Optional custom JSON serializer options to use for serialization and deserialization. Must be provided if
     /// using custom types in messages or state.</param>
+    /// <param name="checkpointOptions">Optional checkpoint manager options to configure JSON serialization behavior.</param>
     /// <returns>A CheckpointManager instance configured to serialize checkpoint data as JSON.</returns>
-    public static CheckpointManager CreateJson(ICheckpointStore<JsonElement> store, JsonSerializerOptions? customOptions = null)
+    public static CheckpointManager CreateJson(ICheckpointStore<JsonElement> store, JsonSerializerOptions? customOptions = null, JsonCheckpointManagerOptions? checkpointOptions = null)
```
westey-m: Consider renaming `customOptions` to `jsonSerializerOptions` to disambiguate it better from `checkpointOptions`.

westey-m: Also, it feels a bit unintuitive to have an `AllowOutOfOrderMetadataProperties` on `checkpointOptions`, which is also on `customOptions`.  The comment for `checkpointOptions` is that it is to configure JSON serialization behavior.  I would expect that that is what the `customOptions` is for.  Without reading the code I would be confused.

lokitoth: The `customOptions` is for configuring the serialization behaviour of user types. It does not impact the serialization of system types. Will think on how to make what is going on here clearer with names / docs.

Maybe look for that field in the user-specified options and rely on it?

lokitoth: I think I like the idea of looking on the actual property on the type the user passes in. It seems unreasonable to only need to support it for one layer since we only ever give out the combined json clob to the underlying persistence layer, so if keys are randomized in some places, I'd expect them to be randomized in others.

PR #3448, opened 2026-01-27
Python: Generic Agents ADR

eavanvalkenburg:
### Contribution Checklist

### DIFF
--- a/docs/decisions/00XX-agents-generic-io.md
+++ b/docs/decisions/00XX-agents-generic-io.md
@@ -1,0 +1,436 @@
+---
+status: Proposed
+contact: eavanvalkenburg
+date: 2026-01-26
+deciders: eavanvalkenburg, markwallace-microsoft,  sphenry, alliscode, taochenosu, moonbox3, dmytrostruk, giles17
+consulted: westey-m, johanste, brettcannon, zooba
+---
+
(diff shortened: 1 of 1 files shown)


docs/decisions/00XX-agents-generic-io.md
```diff
@@ -0,0 +1,436 @@
...
+---
+
+# Making Agents generic over Input and Output types
+
+## Context and Problem Statement
+
+Currently, the Agent Framework's `AgentProtocol` and agent implementations are tightly coupled to chat-based interactions. The `run()` method accepts `str | ChatMessage | Sequence[str | ChatMessage]` and returns `AgentResponse` containing a `messages: list[ChatMessage]` field.
+
+This design limits extensibility for agents that work with different input/output types, such as:
+- Structured data agents (JSON input/output)
```
westey-m, line 16: ChatMessage is already extensible today via the AIContent list that it contains.  AIContent can be text (including JSON), images, sound, or really anything else that an Agent/AI Service requires.

I think the value of this approach is that callers will always receive a specific structure even if they don't know the type of agent they are working with.  Within that structure there may be known content types like TextContent, but also Agent/Service specific content types.  They can choose to just use the former (if they don't know the agent type) or both if they do know the agent type.

PR #3479, opened 2026-01-28
.Net: Adding copilot cli to the codespaces definition

rogerbarreto:
### Motivation and Context

Adding some features to help using copilot cli with Agent Framework via Codespaces.

Features here are similar to the ones loaded by the .NET Template + Copilot Cli.

See: 
- https://github.com/github/dotnet-codespaces/blob/main/.devcontainer/devcontainer.json
- https://github.com/devcontainers/features/tree/main/src/copilot-cli

### DIFF
--- a/.devcontainer/dotnet/devcontainer.json
+++ b/.devcontainer/dotnet/devcontainer.json
@@ -7,7 +7,6 @@
   },
   "features": {
     "ghcr.io/devcontainers/features/azure-cli:1.2.9": {},
-    "ghcr.io/devcontainers/features/docker-in-docker:2.14.0": {},
     "ghcr.io/devcontainers/features/github-cli:1": {
       "version": "2"
     },


.devcontainer/dotnet/devcontainer.json
```diff
@@ -6,10 +6,24 @@
...
-    "ghcr.io/devcontainers/features/dotnet:2.4.2": {},
-    "ghcr.io/devcontainers/features/powershell:1.5.1": {},
     "ghcr.io/devcontainers/features/azure-cli:1.2.9": {},
-    "ghcr.io/devcontainers/features/docker-in-docker:2.14.0": {}
+    "ghcr.io/devcontainers/features/docker-in-docker:2.14.0": {},
+    "ghcr.io/devcontainers/features/github-cli:1": {
+      "version": "2"
+    },
+    "ghcr.io/devcontainers/features/powershell:1": {
+      "version": "latest"
```
westey-m, line 14: Isn't latest the default? Do we have to specify it?  Same for the others.

.devcontainer/dotnet/devcontainer.json
```diff
@@ -6,10 +6,24 @@
     "dockerfile": "dotnet.Dockerfile"
   },
   "features": {
-    "ghcr.io/devcontainers/features/dotnet:2.4.2": {},
-    "ghcr.io/devcontainers/features/powershell:1.5.1": {},
     "ghcr.io/devcontainers/features/azure-cli:1.2.9": {},
-    "ghcr.io/devcontainers/features/docker-in-docker:2.14.0": {}
+    "ghcr.io/devcontainers/features/docker-in-docker:2.14.0": {},
+    "ghcr.io/devcontainers/features/github-cli:1": {
+      "version": "2"
```
westey-m, line 11: Is there no latest here?  Or hopefully the default is the latest?

rogerbarreto: I'm using the same template configuration for those settings as the official github codespaces .NET.

I'm using their approach as the "correct" way of using it.

https://github.com/github/dotnet-codespaces/blob/main/.devcontainer/devcontainer.json

.devcontainer/dotnet/devcontainer.json
rogerbarreto: Answered in the next

lokitoth (approved):

PR #3493, opened 2026-01-28
Python: Fix dependencies for durabletask

larohra:
### Motivation and Context

- Add python.dateutil dependency for `durabletask` package and remove unnecessary typing from azurefunctions package.
- Fix the `requirements.txt` for `durabletask` package. 

### Contribution Checklist


markwallace-microsoft:
<a href="https://github.com/microsoft/agent-framework/blob/0de5065d3a0435c3c87c4f3493b78a52fc3ffb74/README.md">Python Test Coverage</a><details><summary>Python Test Coverage Report • </summary><table><tr><th>File</th><th>Stmts</th><th>Miss</th><th>Cover</th><th>Missing</th></tr><tbody><tr><td><b>TOTAL</b></td><td><b>16249</b></td><td><b>2024</b></td><td><b>87%</b></td><td>&nbsp;</td></tr></tbody></table><i>report-only-changed-files is enabled. No files were changed during this commit :)</i></details>

## Python Unit Test Overview
| 3674 | 221 :zzz: | 0 :x: | 0 :fire: | 1m 9s :stopwatch: |

python/samples/getting_started/durabletask/01_single_agent/requirements.txt
```diff
@@ -1,6 +1,5 @@
-# Agent Framework packages (installing from local package until a package is published)
--e ../../../../
--e ../../../../packages/durabletask
+# Agent Framework packages
+agent-framework-durabletask
```
cgillum: Is it common for samples in this repo to pull from the published pypi bits rather than from the local copies of these libraries? The advantage to this approach is that you can just copy/paste this sample and run it. The disadvantage, unless I'm misunderstanding, is that you can't easily validate this sample against local SDK changes.

larohra: The other samples dont really specify an explicit `requirements.txt` so they kind of run the local changes (the dependencies are pulled based on the `pyproject.toml` specified in the root). 

We've designed the samples here to pull from the deployed `pypi` bits primarily so that they can be self-contained and easier to execute based on the deployed bits (this is how our existing `azurefunctions` samples are also defined). 

I do see the benefits of pulling from the local SDK changes, but historically it got complicated when the users wanted to just try out a specific sample and didnt really want to clone the whole repo but rather just wanted to download the sample zip and run that. 

In the spirit of keeping the samples more self-contained, I would prefer keeping them pinned to the deployed pypi version but happy to discuss more if you have any more suggestions around it.

cgillum: I understand the spirit, but I'm actually more concerned about whether we're breaking the convention of all other samples.

Your call how you want to go forward.

larohra: I just realized one more drawback of the existing approach is that anytime changes are made to the core package, integration tests would still run on the deployed package which might not be able to catch any breaking errors. 

Let me revert to continue using local packages over the deployed package.

larohra: I double checked the code and for tests we were always using the local packages. Either way reverted to using local packages for samples and added a comment to showcase how customers can use the production version if needed

cgillum (approved):

Issue #3020, opened 2025-12-22
.NET Workflows - Make use of `MediaType` property for message content for Declarative Workflows


crickman:
`MediaType` added to `AzureAgentMessageContent` in https://msazure.visualstudio.com/CCI/_git/ObjectModel/pullrequest/14299720.

Integrate this change for use by declarative workflows and include tests.

PR #3021, opened 2025-12-23
.NET Workflows - Support fidelity when converting to and from `ChatMessage` in declarative workflows

crickman:
### Motivation and Context

Ensures media-type and message id are persisted when converting between `ChatMessage` and the object model representation.

Fixes: https://github.com/microsoft/agent-framework/issues/3020

Note: Relies on dependency update in PR https://github.com/microsoft/agent-framework/pull/3017 (rebase this PR to `main` after merged)

### Description

- Makes use of update to Object Model update that supports `MediaType`
- Tests added

### Contribution Checklist

- [X] The code builds clean without any errors or warnings
- [X] The PR follows the [Contribution Guidelines](https://github.com/microsoft/agent-framework/blob/main/CONTRIBUTING.md)
- [X] All unit tests pass, and I have added new tests where possible
- [ ] **Is this a breaking change?** If yes, add "[BREAKING]" prefix to the title of the PR.

PR #3505, opened 2026-01-29
.NET: Workflows - Support fidelity when converting to and from ChatMessage in declarative workflows

peibekwe:
This is a rebase of PR #3021 to main. Initial description of the change is copied below and the credit for this change goes to Chris (The original PR author).

### Motivation and Context

Ensures media-type and message id are persisted when converting between ChatMessage and the object model representation.

Fixes: #3020 

### Description

- Makes use of update to Object Model update that supports MediaType
- Tests added

### Contribution Checklist


dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Extensions/ChatMessageExtensions.cs
```diff
@@ -159,25 +161,52 @@ private static IEnumerable<AIContent> GetContent(this RecordDataValue message)
...
-    private static AIContent GetImageContent(string uriText) =>
+    private static string InferMediaType(string value)
+    {
+        // Base64 encoded content includes media type
+        if (value.StartsWith("data:", StringComparison.OrdinalIgnoreCase))
+        {
+            int semicolonIndex = value.IndexOf(';');
+            if (semicolonIndex > 5)
+            {
+                return value.Substring(5, semicolonIndex - 5);
```
alliscode, line 189: This check feels fragile. A semicolon is not required in data urls with a MIME type, see https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Schemes/data

peibekwe: As discussed offline, I believe that for image content (which is what this is being used), there will be an encoding or a default encoding which means that the data url will contain a semicolon. For non-data value, file extension will be used for default value.

alliscode (approved):

PR #3519, opened 2026-01-30
.NET: [Breaking] Remove AgentResponse{Update}.Id

SergeyMenshykh:
### Motivation and Context

The property was originally added to represent an agent identifier, but at the moment, there are no benefits in having it in the agent response. It can be easily accessed either directly from the agent instance or its workflow executor wrapper.
  
### Description

This PR removes the property as part of tightening and preparing the AF public API surface for RC and GA.


SergeyMenshykh:
The AgentId property is useful in scenarios where a workflow is converted into an agent (AsAIAgent), and the caller might need to know which agent produced the workflow result.

Issue #3551, opened 2026-01-30
Python: Simplify executor registration with dict-based API


moonbox3:
## Summary

`register_executor` calls are verbose. Consider dict-based registration.

## Files

| File | Line |
|------|------|
| sample: `aggregate_results.py` | workflow building |

## Context

```python
# TODO: ask, remove the need to make these register_executor calls
# Handle this via dicts key value pair
.register_executor(lambda: Dispatcher(id="dispatcher"), name="dispatcher")
```

## Action Items

- [ ] Design simpler executor registration (e.g., dict-based)

holtvogt:
Hi @moonbox3 ! Is this issue up for grabs?

holtvogt:
@moonbox3 @TaoChenOSU Similar to this issue, would it make sense to simplify `.register_agent()` calls using the same dictionary based approach? That could be handled as a separate issue, and I'd be happy to work on it as well since I'm already simplifying the executors. Let me know what you think!

moonbox3:
> [@moonbox3](https://github.com/moonbox3) [@TaoChenOSU](https://github.com/TaoChenOSU) Similar to this issue, would it make sense to simplify `.register_agent()` calls using the same dictionary based approach? That could be handled as a separate issue, and I'd be happy to work on it as well since I'm already simplifying the executors. Let me know what you think!

Yes, that would be ideal. We'll look at if we want to leave the `register_*(...)` calls on the API - it's simple enough to instruct a user/dev to create a helper method that returns a new instance of the workflow.

moonbox3:
This is still a to-do item, but has a lower priority than other items. The idea would be to allow for a simplified edge definition, via a dict, and then from that config, build up the edge groups during build time.

moonbox3:
Superseded by #3781

PR #3565, opened 2026-01-31
[BREAKING] Python: Add bulk executor registration method to WorkflowBuilder

holtvogt:
### Motivation and Context

Registering many executors currently needs repetitive `register_executor()` calls. This PR replaces the existing API with a dict-based bulk registration method (`register_executors()`) to reduce boilerplate and verbosity. This is most useful in complex workflows with many executors.

### Description

Replaced the existing executor registration API with a bulk registration API on `WorkflowBuilder `that accepts a `dict[str, Callable[[], Executor | Awaitable[Executor]]]`, reducing boilerplate when registering multiple executors. This PR also adds async factory support to workflows. Executor factories may now return either an `Executor `or an `Awaitable[Executor]`, which is resolved via `build_async()`.

### Contribution Checklist

### Reference

- Fixes https://github.com/microsoft/agent-framework/issues/3551

### DIFF
--- a/python/packages/core/agent_framework/_workflows/_workflow_builder.py
+++ b/python/packages/core/agent_framework/_workflows/_workflow_builder.py
@@ -9,7 +9,7 @@ from typing import Any
 
 from opentelemetry import trace
 
-from .._agents import AgentProtocol
+from .._agents import SupportsAgentRun
 from .._threads import AgentThread
 from ..observability import OtelAttr, capture_exception, create_workflow_span
 from ._agent_executor import AgentExecutor
@@ -176,7 +176,7 @@ class WorkflowBuilder:
         self._max_iterations: int = max_iterations
         self._name: str | None = name
         self._description: str | None = description
-        # Maps underlying AgentProtocol object id -> wrapped Executor so we reuse the same wrapper
+        # Maps underlying SupportsAgentRun object id -> wrapped Executor so we reuse the same wrapper
         # across set_start_executor / add_edge calls. This avoids multiple AgentExecutor instances
         # being created for the same agent.
         self._agent_wrappers: dict[str, Executor] = {}
@@ -192,7 +192,7 @@ class WorkflowBuilder:
         self._executor_registry: dict[str, Callable[[], Executor | Awaitable[Executor]]] = {}
 
         # Output executors filter; if set, only outputs from these executors are yielded
-        self._output_executors: list[Executor | AgentProtocol | str] = []
+        self._output_executors: list[Executor | SupportsAgentRun | str] = []
 
     # Agents auto-wrapped by builder now always stream incremental updates.
 
@@ -213,8 +213,8 @@ class WorkflowBuilder:
 
         return executor.id
 
-    def _maybe_wrap_agent(self, candidate: Executor | AgentProtocol) -> Executor:
-        """If the provided object implements AgentProtocol, wrap it in an AgentExecutor.
+    def _maybe_wrap_agent(self, candidate: Executor | SupportsAgentRun) -> Executor:
+        """If the provided object implements SupportsAgentRun, wrap it in an AgentExecutor.
 
         This allows fluent builder APIs to directly accept agents instead of
         requiring callers to manually instantiate AgentExecutor.
@@ -226,13 +226,13 @@ class WorkflowBuilder:
             An Executor instance, wrapping the agent if necessary.
         """
         try:  # Local import to avoid hard dependency at import time
-            from agent_framework import AgentProtocol  # type: ignore
+            from agent_framework import SupportsAgentRun  # type: ignore
         except Exception:  # pragma: no cover - defensive
-            AgentProtocol = object  # type: ignore
+            SupportsAgentRun = object  # type: ignore
 
         if isinstance(candidate, Executor):  # Already an executor
             return candidate
-        if isinstance(candidate, AgentProtocol):  # type: ignore[arg-type]
+        if isinstance(candidate, SupportsAgentRun):  # type: ignore[arg-type]
             # Reuse existing wrapper for the same agent instance if present
             agent_instance_id = str(id(candidate))
             existing = self._agent_wrappers.get(agent_instance_id)
@@ -249,7 +249,7 @@ class WorkflowBuilder:
             return wrapper
 
         raise TypeError(
-            f"WorkflowBuilder expected an Executor or AgentProtocol instance; got {type(candidate).__name__}."
+            f"WorkflowBuilder expected an Executor or SupportsAgentRun instance; got {type(candidate).__name__}."
         )
 
     def register_executors(self, executor_factories: dict[str, Callable[[], Executor | Awaitable[Executor]]]) -> Self:
@@ -313,7 +313,7 @@ class WorkflowBuilder:
 
     def register_agent(
         self,
-        factory_func: Callable[[], AgentProtocol],
+        factory_func: Callable[[], SupportsAgentRun],
         name: str,
         agent_thread: AgentThread | None = None,
     ) -> Self:
@@ -324,7 +324,7 @@ class WorkflowBuilder:
         enabling deferred initialization and potentially reducing startup time.
 
         Args:
-            factory_func: A callable that returns an AgentProtocol instance when called.
+            factory_func: A callable that returns an SupportsAgentRun instance when called.
             name: The name of the registered agent factory. This doesn't have to match
                   the agent's internal name. But it must be unique within the workflow.
             agent_thread: The thread to use for running the agent. If None, a new thread will be created when
@@ -367,8 +367,8 @@ class WorkflowBuilder:
 
     def add_edge(
         self,
-        source: Executor | AgentProtocol | str,
-        target: Executor | AgentProtocol | str,
+        source: Executor | SupportsAgentRun | str,
+        target: Executor | SupportsAgentRun | str,
         condition: EdgeCondition | None = None,
     ) -> Self:
         """Add a directed edge between two executors.
@@ -437,8 +437,8 @@ class WorkflowBuilder:
             not isinstance(source, str) and isinstance(target, str)
         ):
             raise ValueError(
-                "Both source and target must be either registered factory names (str) "
-                "or Executor/AgentProtocol instances."
+                "Both source and target must be either registered factory names (str) or "
+                "Executor/SupportsAgentRun instances."
             )
 
         if isinstance(source, str) and isinstance(target, str):
@@ -446,7 +446,7 @@ class WorkflowBuilder:
             self._edge_registry.append(_EdgeRegistration(source=source, target=target, condition=condition))
             return self
 
-        # Both are Executor/AgentProtocol instances; wrap and add now
+        # Both are Executor/SupportsAgentRun instances; wrap and add now
         source_exec = self._maybe_wrap_agent(source)  # type: ignore[arg-type]
         target_exec = self._maybe_wrap_agent(target)  # type: ignore[arg-type]
         source_id = self._add_executor(source_exec)
@@ -456,8 +456,8 @@ class WorkflowBuilder:
 
     def add_fan_out_edges(
         self,
-        source: Executor | AgentProtocol | str,
-        targets: Sequence[Executor | AgentProtocol | str],
+        source: Executor | SupportsAgentRun | str,
+        targets: Sequence[Executor | SupportsAgentRun | str],
     ) -> Self:
         """Add multiple edges to the workflow where messages from the source will be sent to all targets.
 
@@ -518,8 +518,8 @@ class WorkflowBuilder:
             not isinstance(source, str) and any(isinstance(t, str) for t in targets)
         ):
             raise ValueError(
-                "Both source and targets must be either registered factory names (str) "
-                "or Executor/AgentProtocol instances."
+                "Both source and targets must be either registered factory names (str) or "
+                "Executor/SupportsAgentRun instances."
             )
 
         if isinstance(source, str) and all(isinstance(t, str) for t in targets):
@@ -527,7 +527,7 @@ class WorkflowBuilder:
             self._edge_registry.append(_FanOutEdgeRegistration(source=source, targets=list(targets)))  # type: ignore
             return self
 
-        # Both are Executor/AgentProtocol instances; wrap and add now
+        # Both are Executor/SupportsAgentRun instances; wrap and add now
         source_exec = self._maybe_wrap_agent(source)  # type: ignore[arg-type]
         target_execs = [self._maybe_wrap_agent(t) for t in targets]  # type: ignore[arg-type]
         source_id = self._add_executor(source_exec)
@@ -538,7 +538,7 @@ class WorkflowBuilder:
 
     def add_switch_case_edge_group(
         self,
-        source: Executor | AgentProtocol | str,
+        source: Executor | SupportsAgentRun | str,
         cases: Sequence[Case | Default],
     ) -> Self:
         """Add an edge group that represents a switch-case statement.
@@ -620,7 +620,7 @@ class WorkflowBuilder:
         ):
             raise ValueError(
                 "Both source and case targets must be either registered factory names (str) "
-                "or Executor/AgentProtocol instances."
+                "or Executor/SupportsAgentRun instances."
             )
 
         if isinstance(source, str) and all(isinstance(case.target, str) for case in cases):
@@ -628,7 +628,7 @@ class WorkflowBuilder:
             self._edge_registry.append(_SwitchCaseEdgeGroupRegistration(source=source, cases=list(cases)))  # type: ignore
             return self
 
-        # Source is an Executor/AgentProtocol instance; wrap and add now
+        # Source is an Executor/SupportsAgentRun instance; wrap and add now
         source_exec = self._maybe_wrap_agent(source)  # type: ignore[arg-type]
         source_id = self._add_executor(source_exec)
         # Convert case data types to internal types that only uses target_id.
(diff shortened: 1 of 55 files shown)


python/packages/core/tests/workflow/test_workflow_builder.py
```diff
@@ -259,6 +259,24 @@ def test_register_multiple_executors():
...
+    })
+
+    assert result is builder
+
+    workflow = builder.set_start_executor("ExecutorA").add_edge("ExecutorA", "ExecutorB").build()
+
+    assert "ExecutorA" in workflow.executors
+    assert "ExecutorB" in workflow.executors
+    assert workflow.start_executor_id == "ExecutorA"
+
```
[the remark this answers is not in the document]
holtvogt: Good point. We can introduce a guard variable such as `_uses_bulk_executor_registration `and set it to `True` when `register_executors()` is used. The `register_executor()` method would then check this flag accordingly?

python/packages/core/agent_framework/_workflows/_workflow_builder.py
```diff
@@ -332,6 +332,54 @@ async def log(self, message: str, ctx: WorkflowContext) -> None:
 
         return self
 
+    def register_executors(self, executor_factories: dict[str, Callable[[], Executor]]) -> Self:
```
TaoChenOSU: We can replace the original `register_executor` API with this new one.

TaoChenOSU:
Hi, thank you for your contribution!

This is actually something we wanted to improve too!

python/packages/core/agent_framework/_workflows/_workflow_builder.py
holtvogt: @TaoChenOSU Okay! Do we plan to add deprecation notes for the old approach, or are we considering a big bang removal of `register_executor()` entirely? The latter would be a larger change, since it would require touching more files.

TaoChenOSU: We can add a deprecation note to the old one.

moonbox3: I'd also simply prefer us to remove the old one. We're trying to get the API as clean as possible, as quickly as possible. Breaking changes are acceptable right now.

moonbox3: Can we also add async callable support, too? Currently when one uses the AzureAIClient, it's expected that creating an agent is an asynchronous process - therefore we should allow for an asynchronous callable, and handle it properly.

holtvogt: Should async executor factories be supported when building workflows inside an async context? If so, should we introduce an async build API (e.g., `await build_async()`), or is it acceptable to require building outside the event loop?

moonbox3: Good question. I've been thinking through this and see a few approaches:                                                                                                                                                    
        
Option 1: Add `build_async()` method                                                                                                                                                                            
                                                                                                                                                                                                                            
Keep the builder chain synchronous, but add an async variant of the terminal `build()` method:                                                                                                                              
                           
```python
workflow = await (
    WorkflowBuilder()
    .register_executors({
        "Agent": lambda: create_azure_agent()  # async factory
    })
    .build_async()
)
```

`build()` would detect async factories and raise a helpful error pointing users to `build_async()`. Users with sync-only factories continue using `build()` unchanged.

To be clear, this is not an "async builder" where every method in the chain is async (that would be ugly). The builder pattern stays fully synchronous - only the terminal build() has an async variant.

Option 2: Defer factory execution to runtime

Don't call factories at build time - store them and invoke lazily when the workflow runs. Since execution is already async, this handles async factories naturally without any API changes.

The problem: we lose build-time validation. Currently `build()` instantiates executors to validate type compatibility (handler signatures, input/output types across edges). With deferred instantiation, these become runtime errors instead of build-time errors - you wouldn't discover a type mismatch until the workflow actually runs.

Option 3: Push async instantiation to the user

Require users to pre-create async executors themselves:

```python
agent = await create_azure_agent()

workflow = (
    WorkflowBuilder()
    .add_edge(agent, processor)  # pass instance directly
    .build()
)
```

No async factories, no `build_async()`. But the problem is this breaks the isolation guarantee that factories provide. If someone creates a helper to build workflows on demand, they'd share the same executor instance
across all workflows - which is exactly what factories were designed to prevent.

I'm leaning toward Option 1. The async footprint is minimal (just the terminal method), we preserve both build-time validation and instance isolation, and users with sync-only factories are unaffected.

Thoughts?

holtvogt:
@moonbox3 The async support involved some heavier lifting, mainly to avoid duplicating existing logic and to cleanly separate reusable helpers. Does this approach look reasonable, or would you prefer handling the duplication differently? Let me know what you think!

python/packages/core/agent_framework/_workflows/_workflow_builder.py
```diff
@@ -1152,22 +1143,207 @@ def _get_executor(name: str) -> Executor:
...
+            Non-factory executors (i.e., those added directly) are not included in the returned list,
+            as they are already part of the workflow builder's internal state.
+        """
+        factory_name_to_instance: dict[str, Executor] = {}
+        for name, executor_factory in self._executor_registry.items():
+            instance = executor_factory()
+            if inspect.isawaitable(instance):
+                # Close un-awaited coroutines to avoid runtime warnings or memory leaks
+                if hasattr(instance, "close"):
+                    instance.close()  # type: ignore[reportGeneralTypeIssues]
```
moonbox3: if `instance.close()` ever throws, the `ValueError` is never raised. Should we wrap it in a try/except?

python/packages/core/agent_framework/_workflows/_workflow_builder.py
```diff
@@ -1152,22 +1143,207 @@ def _get_executor(name: str) -> Executor:
...
+                raise ValueError("Async executor factories were detected. Use build_async() instead.")
+            factory_name_to_instance[name] = instance
```
moonbox3, line 1250: Should we add a `isinstance(instance, Executor)` check here like:

```python
if not isinstance(instance, Executor):
      raise TypeError(
          f"Factory '{name}' returned {type(instance).__name__} instead of an Executor."
      )
```

Same for the new async path?

Also, when a factory in a large dict raises an exception during `build()`, the error has no indication of which factory name caused it. A try/except that adds the factory name would help improve debuggability, I think.

python/packages/core/agent_framework/_workflows/_workflow_builder.py
```diff
@@ -1152,22 +1143,207 @@ def _get_executor(name: str) -> Executor:
...
+        """
+        if not self._start_executor:
+            raise ValueError("Starting executor must be set using set_start_executor before building the workflow.")
+
+        start_executor: Executor | None = None
+        if isinstance(self._start_executor, Executor):
+            start_executor = self._start_executor
+
+        deferred_edge_groups: list[EdgeGroup] = []
+        executor_ids_seen: set[str] = set()
```
moonbox3: When a duplicate ID is found, the error says which ID collided but not which two factory names produced it. What if we use a `dict[str, str]` (id -> name) so we could say: `"ID 'agent' from factory 'B' conflicts with factory 'A'"`?

python/packages/core/tests/workflow/test_workflow_builder.py
moonbox3, line 1: Do we have tests for:
- mixed sync+async factories with build_async()?
-  coroutine `.close()` actually called on sync build with async factory?
- `build_async()` when async factory raises exception?
- multiple `register_executors()` calls with non-overlapping names?
- whitespace-only names (like "  ") failing as rejected?

python/packages/core/agent_framework/_workflows/_workflow_builder.py
```diff
@@ -273,49 +278,34 @@ async def process(self, text: str, ctx: WorkflowContext[str]) -> None:
...
-        names = [name] if isinstance(name, str) else name
+        if not executor_factories:
+            raise ValueError("Executor factories cannot be empty.")
 
-        for n in names:
-            if n in self._executor_registry:
-                raise ValueError(f"An executor factory with the name '{n}' is already registered.")
+        for name, factory_function in executor_factories.items():
+            if not name or not name.strip():
+                raise ValueError("Executor factory name cannot be empty.")
```
moonbox3:
```suggestion
                raise ValueError("Executor factory name cannot be empty or whitespace-only")
```

python/packages/core/agent_framework/_workflows/_workflow_builder.py
```diff
@@ -1152,22 +1143,207 @@ def _get_executor(name: str) -> Executor:
...
+            if inspect.isawaitable(instance):
+                # Handle maybe awaitable executor instances
+                instance = await instance
+            factory_name_to_instance[name] = instance
+
+        return self._process_instantiated_executors(factory_name_to_instance)
+
+    def _create_workflow_from_resolved_registry(
+        self,
+        span: Any,
```
moonbox3:
```suggestion
        span: trace.Span,
``` 

?

markwallace-microsoft:
<a href="https://github.com/microsoft/agent-framework/blob/f81f1ba955e4f1b8bb6b5193af8396bc2b34881b/README.md">Python Test Coverage</a><details><summary>Python Test Coverage Report • </summary><table><tr><th>File</th><th>Stmts</th><th>Miss</th><th>Cover</th><th>Missing</th></tr><tbody><tr><td colspan="5"><b>packages/core/agent_framework/_workflows</b></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/f81f1ba955e4f1b8bb6b5193af8396bc2b34881b/packages/core/agent_framework/_workflows/_workflow_builder.py">_workflow_builder.py</a></td><td>314</td><td>42</td><td>86%</td><td><a href="https://github.com/microsoft/agent-framework/blob/f81f1ba955e4f1b8bb6b5193af8396bc2b34881b/packages/core/agent_framework/_workflows/_workflow_builder.py#L249">249</a>, <a href="https://github.com/microsoft/agent-framework/blob/f81f1ba955e4f1b8bb6b5193af8396bc2b34881b/packages/core/agent_framework/_workflows/_workflow_builder.py#L518">518</a>, <a href="https://github.com/microsoft/agent-framework/blob/f81f1ba955e4f1b8bb6b5193af8396bc2b34881b/packages/core/agent_framework/_workflows/_workflow_builder.py#L619">619</a>, <a href="https://github.com/microsoft/agent-framework/blob/f81f1ba955e4f1b8bb6b5193af8396bc2b34881b/packages/core/agent_framework/_workflows/_workflow_builder.py#L626-L627">626&ndash;627</a>, <a href="https://github.com/microsoft/agent-framework/blob/f81f1ba955e4f1b8bb6b5193af8396bc2b34881b/packages/core/agent_framework/_workflows/_workflow_builder.py#L730">730</a>, <a href="https://github.com/microsoft/agent-framework/blob/f81f1ba955e4f1b8bb6b5193af8396bc2b34881b/packages/core/agent_framework/_workflows/_workflow_builder.py#L733">733</a>, <a href="https://github.com/microsoft/agent-framework/blob/f81f1ba955e4f1b8bb6b5193af8396bc2b34881b/packages/core/agent_framework/_workflows/_workflow_builder.py#L738">738</a>, <a href="https://github.com/microsoft/agent-framework/blob/f81f1ba955e4f1b8bb6b5193af8396bc2b34881b/packages/core/agent_framework/_workflows/_workflow_builder.py#L740">740</a>, <a href="https://github.com/microsoft/agent-framework/blob/f81f1ba955e4f1b8bb6b5193af8396bc2b34881b/packages/core/agent_framework/_workflows/_workflow_builder.py#L747">747</a>, <a href="https://github.com/microsoft/agent-framework/blob/f81f1ba955e4f1b8bb6b5193af8396bc2b34881b/packages/core/agent_framework/_workflows/_workflow_builder.py#L750-L754">750&ndash;754</a>, <a href="https://github.com/microsoft/agent-framework/blob/f81f1ba955e4f1b8bb6b5193af8396bc2b34881b/packages/core/agent_framework/_workflows/_workflow_builder.py#L756">756</a>, <a href="https://github.com/microsoft/agent-framework/blob/f81f1ba955e4f1b8bb6b5193af8396bc2b34881b/packages/core/agent_framework/_workflows/_workflow_builder.py#L820">820</a>, <a href="https://github.com/microsoft/agent-framework/blob/f81f1ba955e4f1b8bb6b5193af8396bc2b34881b/packages/core/agent_framework/_workflows/_workflow_builder.py#L897">897</a>, <a href="https://github.com/microsoft/agent-framework/blob/f81f1ba955e4f1b8bb6b5193af8396bc2b34881b/packages/core/agent_framework/_workflows/_workflow_builder.py#L900">900</a>, <a href="https://github.com/microsoft/agent-framework/blob/f81f1ba955e4f1b8bb6b5193af8396bc2b34881b/packages/core/agent_framework/_workflows/_workflow_builder.py#L962-L963">962&ndash;963</a>, <a href="https://github.com/microsoft/agent-framework/blob/f81f1ba955e4f1b8bb6b5193af8396bc2b34881b/packages/core/agent_framework/_workflows/_workflow_builder.py#L1119">1119</a>, <a href="https://github.com/microsoft/agent-framework/blob/f81f1ba955e4f1b8bb6b5193af8396bc2b34881b/packages/core/agent_framework/_workflows/_workflow_builder.py#L1134-L1141">1134&ndash;1141</a>, <a href="https://github.com/microsoft/agent-framework/blob/f81f1ba955e4f1b8bb6b5193af8396bc2b34881b/packages/core/agent_framework/_workflows/_workflow_builder.py#L1143">1143</a>, <a href="https://github.com/microsoft/agent-framework/blob/f81f1ba955e4f1b8bb6b5193af8396bc2b34881b/packages/core/agent_framework/_workflows/_workflow_builder.py#L1146">1146</a>, <a href="https://github.com/microsoft/agent-framework/blob/f81f1ba955e4f1b8bb6b5193af8396bc2b34881b/packages/core/agent_framework/_workflows/_workflow_builder.py#L1148-L1150">1148&ndash;1150</a>, <a href="https://github.com/microsoft/agent-framework/blob/f81f1ba955e4f1b8bb6b5193af8396bc2b34881b/packages/core/agent_framework/_workflows/_workflow_builder.py#L1192">1192</a>, <a href="https://github.com/microsoft/agent-framework/blob/f81f1ba955e4f1b8bb6b5193af8396bc2b34881b/packages/core/agent_framework/_workflows/_workflow_builder.py#L1205">1205</a>, <a href="https://github.com/microsoft/agent-framework/blob/f81f1ba955e4f1b8bb6b5193af8396bc2b34881b/packages/core/agent_framework/_workflows/_workflow_builder.py#L1479-L1480">1479&ndash;1480</a>, <a href="https://github.com/microsoft/agent-framework/blob/f81f1ba955e4f1b8bb6b5193af8396bc2b34881b/packages/core/agent_framework/_workflows/_workflow_builder.py#L1484-L1486">1484&ndash;1486</a></td></tr><tr><td><b>TOTAL</b></td><td><b>16400</b></td><td><b>1927</b></td><td><b>88%</b></td><td>&nbsp;</td></tr></tbody></table></details>

## Python Unit Test Overview
| 3994 | 221 :zzz: | 0 :x: | 0 :fire: | 1m 8s :stopwatch: |

python/packages/core/agent_framework/_workflows/_workflow_builder.py
holtvogt: Can we intentionally overwrite the error from `close()` using a `finally` block instead?

holtvogt:
@moonbox3 Thanks for the detailed feedback and the sharp eye. I addressed the points and will happily hand this back to you for another round 🔍

python/packages/core/agent_framework/_workflows/_workflow_builder.py
```diff
@@ -184,7 +189,7 @@ def __init__(
             | _MultiSelectionEdgeGroupRegistration
             | _FanInEdgeRegistration
         ] = []
-        self._executor_registry: dict[str, Callable[[], Executor]] = {}
+        self._executor_registry: dict[str, Callable[[], Executor | Awaitable[Executor]]] = {}
```
TaoChenOSU, line 192: Should we allow async factories? This will force us to have an async build method.

I intentionally left out async factories when adding the factory pattern because I'd much prefer a single build method.

AzureAIAgent requires async methods to create or retrieve an agent, which makes this very tricky. 

The `AIProjectClient` does support none-async agent creation. Should we add support for that too? @eavanvalkenburg

eavanvalkenburg: So it's nuanced,if you want the agent with Foundry to be created them you need the async 'create_agent' method, but you can always use the non-async 'get_agent' method, and then we create on the portal at first run, so I wouldn't let that determine this design because I agree, having non asyn build is preferred. @dmytrostruk can share additional insights, he did most of the work

moonbox3: The reason I suggested an asynchronous factory is because of the need for AzureAIClient. We need this to be an easy experience - why are we forcing users to first have to define an agent in the portal, AND then be able to use it in code? How is that a good dev experience?

dmytrostruk: > The AIProjectClient does support none-async agent creation. Should we add support for that too?

I don't think we can do this, using sync methods will block the event loop, can impact performance etc. I also don't think we need to force users to first have to define an agent in the portal. As @eavanvalkenburg mentioned, we have this ability in `AzureAIClient` to create an agent on first `await agent.run()` instead of doing it before the `ChatAgent` instance is created, which could be useful here, but from `AzureAIClient` perspective - it's unexpected design, which already created some confusion around it, so we should definitely revisit. 

But I also don't see a problem to have async factory. There are local agents as well as remote agents, and there should be a mechanism to create both, so you can work with them.

python/packages/core/agent_framework/_workflows/_workflow_builder.py
holtvogt: It might be worth logging `close()` failures at `WARNING` level in an `except` block to explicitly flag possible resource leaks. This seems safer than silently swallowing the error. Thoughts?

moonbox3:
Hi @holtvogt, we appreciate all the work and time you put into this PR. We're going to go a different direction by removing the factory methods. We will do this in a different PR. I will close this PR right now.

Issue #3601, opened 2026-02-02
Python: sessions: Rename ChatMessageStore to ChatHistoryProvider


eavanvalkenburg:
Rename ChatMessageStore to ChatHistoryProvider for better clarity.

**Current Name:** ChatMessageStore
**New Name:** ChatHistoryProvider

**File:** python/packages/core/agent_framework/_threads.py

Parent issue: #3575

Issue #3590, opened 2026-02-02
Python: sessions: Design orthogonal ChatMessageStore for service vs local threads


eavanvalkenburg:
Figure out if we need the always-available ability to use ChatMessageStore and make it orthogonal to service vs local threads. Consider if we should force users to handle this themselves when using service threads.

**File:** python/packages/core/agent_framework/_threads.py
**Location:** Line ~314

**Action:** Design decision needed on ChatMessageStore handling for different thread types.

Parent issue: #3575

Issue #3589, opened 2026-02-02
Python: [Python] Move serialize method into the agent


eavanvalkenburg:
The serialize method should be moved from the thread into the agent.

**File:** python/packages/core/agent_framework/_threads.py
**Location:** Line ~319

Current:
state = await local_thread.serialize()

**Action:** Move serialize logic into the agent class.

Parent issue: #3575

Issue #3588, opened 2026-02-02
Python: sessions: Add get_new_session, get_session_by_id methods


eavanvalkenburg:
Add helper methods for session/thread management:
- get_new_session() - should not have side-effects like chat history provider being called
- get_session_by_id(id)
- get_session_from_history(history)

**File:** python/packages/core/agent_framework/_threads.py
**Location:** Line ~310

**Action:** Implement these session management helper methods.

Parent issue: #3575

PR #3609, opened 2026-02-02
Python: ADR: Unifying Context Management with ContextMiddleware (Python)

eavanvalkenburg:
## Summary

This ADR proposes unifying \ContextProvider\, \ChatMessageStore\, and \AgentThread\ into a single \ContextMiddleware\ concept for the Python SDK.

## Problem

Currently, developers doing 'Context Engineering' must understand multiple abstractions:
- \ContextProvider\ - Injects instructions, messages, and tools
- \ChatMessageStore\ - Stores and retrieves conversation history
- \AgentThread\ - Coordinates between them

## Proposed Solution

A unified \ContextMiddleware\ using the onion/wrapper pattern (like existing \AgentMiddleware\):

\\\python
class RAGMiddleware(ContextMiddleware):
    async def process(self, context: SessionContext, next) -> None:
        # Pre-processing: add context
        docs = await self.retrieve_documents(context.input_messages[-1].text)
        context.add_messages(self.source_id, [ChatMessage.system(f'Context: {docs}')])

        await next(context)

        # Post-processing: store/audit
        await self.store_interaction(context.input_messages, context.response_messages)

agent = ChatAgent(
    chat_client=client,
    context_middleware=[
        InMemoryStorageMiddleware('memory'),
        RAGMiddleware('rag'),
    ]
)
\\\

## Key Decisions

| Decision | Rationale |
|----------|-----------|
| Onion/wrapper pattern | Familiar from existing middleware, natural pre/post processing |
| Agent owns config, Session owns pipeline | Enables per-session factories |
| Mandatory \source_id\ | Attribution in \context_messages\ dict |
| Default \InMemoryStorageMiddleware\ | Zero-config conversation history |
| Single \StorageContextMiddleware\ class | Configure for memory/audit/evaluation |
| Clean break (no shims) | We're in preview |

## Migration Impact

| Current | New |
|---------|-----|
| \ContextProvider\ | \ContextMiddleware\ |
| \ChatMessageStore\ | \StorageContextMiddleware\ |
| \AgentThread\ | \AgentSession\ |

See the full ADR in \docs/decisions/00XX-python-context-middleware.md\ for detailed design, code examples, and implementation workplan.

## Related Issues

This ADR addresses the following issues:
- #3587 - Rename AgentThread to AgentSession
- #3588 - Add get_existing_thread, get_new_session, get_session_by_id methods
- #3589 - Move serialize method into the agent
- #3590 - Design orthogonal ChatMessageStore for service vs local threads
- #3601 - Rename ChatMessageStore to ChatHistoryProvider


docs/decisions/00XX-python-context-middleware.md
```diff
@@ -0,0 +1,1914 @@
...
+class RAGMiddleware(ContextMiddleware):
+    async def process(self, context: SessionContext, next) -> None:
+        # Pre-processing
+        docs = await self.retrieve_documents(context.input_messages[-1].text)
+        context.add_messages(self.source_id, [ChatMessage.system(f"Context: {docs}")])
+
+        await next(context)
+
+        # Post-processing
+        await self.store_interaction(context.input_messages, context.response_messages)
```
westey-m, line 696: Can you modify response_messages here?  If you do, do those modifications get returned to the callers, both middleware higher in the stack and the user?

docs/decisions/00XX-python-context-middleware.md
```diff
@@ -0,0 +1,1914 @@
...
+- **Agent** owns `Sequence[ContextMiddlewareConfig]` (instances or factories)
+- **AgentSession** owns `ContextMiddlewarePipeline` (resolved at runtime)
+
+```python
+# Agent holds middleware configuration
+agent = ChatAgent(
+    chat_client=client,
+    context_middleware=[
+        InMemoryStorageMiddleware("memory"),
+        RAGContextMiddleware("rag"),
```
westey-m: If the user only wanted to use the user input to do a rag search, rather than that plus the chat history, how do they filter to only user input?

docs/decisions/00XX-python-context-middleware.md
```diff
@@ -0,0 +1,1914 @@
...
+agent = ChatAgent(chat_client=client, name="assistant")
+session = agent.create_session()
+response = await agent.run("Hello!", session=session)
+response = await agent.run("What did I say?", session=session)  # Remembers!
+```
+
+Default `InMemoryStorageMiddleware` is added at runtime when:
+- No `service_session_id` (service not managing storage)
+- `options.store` is not `True` (user not expecting service storage)
+- Pipeline is empty or None
```
westey-m: If the user only supplies regular context middleware and no storage context middleware, does the user get no chat history storage?

docs/decisions/00XX-python-context-middleware.md
eavanvalkenburg: I think the intent will be (and i'll clarify) that this is not the place to alter the responses, just read them and do something, the AgentMiddleware can be used for that purpose.

docs/decisions/00XX-python-context-middleware.md
eavanvalkenburg: In the setup of each middleware there are controls over which messages they want to use, and since all messages from other context providers are stored separately that is fully configurable

docs/decisions/00XX-python-context-middleware.md
eavanvalkenburg: So we can go two routes here, both only for when there is no service_session_id and store==False:

1. If the pipeline is present but does not have StorageMiddleware  -> Add InMemoryStorage
2. If there is NO middleware(pipeline)  -> Add pipeline with InMemoryStorage

The first might be easier for getting started but the second is clearer, because if you do already have a pipeline with a middleware the order suddenly matters, so we should clarify that in that case the user should do this themselves because we do not want to make assumptions about ordering. This will make sure that the simple case always works, but once people start adding their own middlewares, they should know what they want to do and thus I think approach two is the way to do.

docs/decisions/00XX-python-context-middleware.md
westey-m: Is there an example of that?

docs/decisions/00XX-python-context-middleware.md
```diff
@@ -0,0 +1,1914 @@
...
+```python
+class StorageContextMiddleware(ContextMiddleware):
+    def __init__(
+        self,
+        source_id: str,
+        *,
+        load_messages: bool | None = None,  # None = smart mode
+        store_inputs: bool = True,
+        store_responses: bool = True,
+        store_context_messages: bool = False,
```
dmytrostruk: What's the difference between `inputs/responses` and `context_messages`?

docs/decisions/00XX-python-context-middleware.md
```diff
@@ -0,0 +1,1914 @@
...
+**Pros:**
+- No migration required
+- Familiar to existing users
+
+**Cons:**
+- Cognitive overhead remains
+- No composability for context providers
+- Inconsistent with middleware pattern used elsewhere
+
+### Option 2: ContextMiddleware (Chosen)
```
victordibia: Thanks for this ADR! The unified ContextMiddleware pattern is a nice improvement for context organization and attribution.

One question: would this design support implementing **context compaction strategies**?

For example, a common need is letting agents run for arbitrary long periods by automatically compacting/replacing the message history when a max context budget is hit. E.g., agent runs calls 10s or 100s of tool calls + model call in succession and between those calls may compact context between. 

Currently this is challenging to do today because:
- ChatMessageStore.list_messages() is only called once at the start of agent.run(), not during the tool loop
- ChatMiddleware operates on a copy of messages, so modifications don't persist across tool loop iterations

see some related notes [here](https://gist.github.com/victordibia/ec3f3baf97345f7e47da025cf55b999f).

With the new ContextMiddleware, would it be possible to:

Have middleware run _during_ tool loop iterations (not just at the invocation boundary)?
Allow a compaction strategy to truly replace the context mid-execution?

If this is out of scope for this ADR, it might be worth noting as a future consideration—or confirming whether the new architecture would make such a feature easier to add later (or there is some other recommended parttern for this).

@dmytrostruk @alliscode @markwallace-microsoft

docs/decisions/00XX-python-context-middleware.md
westey-m: I agree that 1 is problematic for the same reasons as you listed.

> so we should clarify that in that case the user should do this themselves

At the same time, I think this ^ is a pit of failure, since just adding RAG middleware to a functional agent would make that agent break, and the user has to go and read the documentation to figure out why.

docs/decisions/00XX-python-context-middleware.md
eavanvalkenburg: @victordibia thanks for the reply, I added a discussion section for this topic, I'm leaning for a CompactionStrategy abstraction that can be used both for and after storage and for function call loop (with no requirement to use the same one in different places as they serve different purposes), but check the discussion in the latest update here.

victordibia: Thanks for adding the compaction discussion section, Eduard — appreciate you engaging with the feedback.

Having gone through the latest version of the ADR, I want to push a bit further on two things:

**1. Compaction still needs a concrete plan, not just a discussion section**

The open discussion section accurately restates the problem (tool loop grows unboundedly, boundary hooks can't reach it), but it stops at listing options without a decision or developer-facing API. The ADR has detailed implementation plans, workplans, and code examples for everything else; should we give compaction the same treatment?

At minimum, I think we should commit to a developer experience like:

```python
agent = Agent(
    name="investigator",
    instructions="Investigate the codebase and find the bug.",
    model_client=model_client,
    tools=[read_file, list_directory, grep],
    compaction=HeadTailStrategy(
        token_budget=100_000,
        head_ratio=0.2,
    ),
)
```

behaviour of the code above: compaction triggers inside the tool loop when context exceeds the budget, the strategy replaces the working message list (respecting tool_call/result atomicity), and developers can **bring their own compaction strategy**. Built-in options could include head/tail, sliding window, summarization, etc.

You mentioned you were leaning toward a `CompactionStrategy` abstraction, that makes sense to me. My concern is that without nailing down how it interacts with the new `ContextProvider` pipeline (does `HistoryProvider` store the compacted or uncompacted history? does compaction run before or after `after_run`?, can we replace the content of the ContextProvider as needed by a compaction strategy), we risk shipping a `ContextProvider` API that needs breaking changes when compaction lands.

Could we at minimum add a decided API shape + interaction model to this ADR, even if the implementation is a separate PR?

**2. Loop-level hooks for deterministic agent behavior**

Related but distinct from compaction — for long-running autonomous agents, we also need hooks *inside* the agentic loop:

- **Pre-iteration**: Inject deterministic behavior between tool calls. E.g., force a re-planning step every N iterations and add the plan to context.
- **Exit validation**: Before the agent exits due to no more tool calls, run checker logic that can reject the exit and resume the loop. E.g., agent declares "done" but a verifier determines the task isn't actually complete and injects feedback to continue.

These are the guardrails that make eyes-off agents viable. Without them, developers end up wrapping `agent.run()` in their own outer loop, which defeats the purpose of the framework managing the agentic loop.

I realize this might be out of scope for this specific ADR. If so, could we explicitly note it as a planned follow-up rather than leaving it unaddressed? Happy to help write up a separate ADR for loop hooks + compaction if that's cleaner.

@dmytrostruk @alliscode @markwallace-microsoft — would appreciate your thoughts on whether these should block or follow this ADR.

victordibia (approved):

docs/decisions/0016-python-context-middleware.md
```diff
@@ -0,0 +1,2575 @@
...
+- `StorageContextMiddleware(ContextMiddleware)` - for conversation history storage (in-memory, Redis, Cosmos, etc.)
+
+```python
+class ContextMiddleware(ABC):
+    def __init__(self, source_id: str, *, session_id: str | None = None):
+        self.source_id = source_id
+        self.session_id = session_id
+
+    @abstractmethod
+    async def process(self, context: SessionContext, next: ContextMiddlewareNext) -> None:
```
dmytrostruk, line 171:
```suggestion
    async def process(self, context: SessionContext, call_next: ContextMiddlewareNext) -> None:
```

dmytrostruk (approved):

PR #3658, opened 2026-02-04
.NET:[Breaking] Add support for structured output

SergeyMenshykh:
### Motivation and Context

Today, ChatClientAgent has `RunAsync<T>` instance methods to work with structured output in AF. However, if ChatClientAgent is decorated by, let's say, an Open Telemetry decorator, the agent users lose access to the `RunAsync<T>` methods because they are **instance** methods of ChatClientAgent rather than members of the AIAgent class.

Additionally, there are no RunAsync overloads that can work with SO type represented by System.Type.

### Description
This PR does the following:
- The ChatResponse property to the AgentRunOptions to communicate the agent's SO configuration.
- `RunAsync<T>` non-virtual methods to the AIAgent base abstract class.
- Removes the `bool? useJsonSchemaResponseFormat = null` parameter from existing `RunAsync<T>` signatures.
- Adds the `Clone` method to AgentRunOptions, ChatClientAgentRunOptions, DurableAgentOptions.

## Out of Scope  
  
The following will be added in the next PRs:  
- Wrapping primitives and arrays into a JSON object schema.  
- A decorator to add structured output support to agents that do not support structured output natively.  
- Functionality to check if an agent supports structured output and throw an exception if it does not.

**Note** This PR targets `feature-so` feature branch.

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.Abstractions/AgentResponse.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Abstractions/AgentResponse.cs
@@ -78,7 +78,7 @@ public class AgentResponse
     /// metadata and storing the original response in <see cref="RawRepresentation"/> for access to
     /// the underlying implementation details.
     /// </remarks>
-    public AgentResponse(AgentResponse response)
+    protected AgentResponse(AgentResponse response)
     {
         _ = Throw.IfNull(response);
 
--- a/dotnet/src/Microsoft.Agents.AI.Abstractions/AgentResponse{T}.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Abstractions/AgentResponse{T}.cs
@@ -40,36 +40,20 @@ public class AgentResponse<T> : AgentResponse
     {
         get
         {
-            var structuredOutput = this.GetResultCore(this._serializerOptions, out var failureReason);
-            return failureReason switch
+            var json = this.Text;
+            if (string.IsNullOrEmpty(json))
             {
-                FailureReason.ResultDidNotContainJson => throw new InvalidOperationException("The response did not contain JSON to be deserialized."),
-                FailureReason.DeserializationProducedNull => throw new InvalidOperationException("The deserialized response is null."),
-                _ => structuredOutput!,
-            };
+                throw new InvalidOperationException("The response did not contain JSON to be deserialized.");
+            }
+
+            T? deserialized = DeserializeFirstTopLevelObject(json!, (JsonTypeInfo<T>)this._serializerOptions.GetTypeInfo(typeof(T)));
+            if (deserialized is null)
+            {
+                throw new InvalidOperationException("The deserialized response is null.");
+            }
+
+            return deserialized;
         }
-    }
-
-    private T? GetResultCore(JsonSerializerOptions serializerOptions, out FailureReason? failureReason)
-    {
-        var json = this.Text;
-        if (string.IsNullOrEmpty(json))
-        {
-            failureReason = FailureReason.ResultDidNotContainJson;
-            return default;
-        }
-
-        // If there's an exception here, we want it to propagate, since the Result property is meant to throw directly
-        T? deserialized = DeserializeFirstTopLevelObject(json!, (JsonTypeInfo<T>)serializerOptions.GetTypeInfo(typeof(T)));
-
-        if (deserialized is null)
-        {
-            failureReason = FailureReason.DeserializationProducedNull;
-            return default;
-        }
-
-        failureReason = default;
-        return deserialized;
     }
 
     private static T? DeserializeFirstTopLevelObject(string json, JsonTypeInfo<T> typeInfo)
@@ -93,11 +77,5 @@ public class AgentResponse<T> : AgentResponse
 #else
         return JsonSerializer.Deserialize(json, typeInfo);
 #endif
-    }
-
-    private enum FailureReason
-    {
-        ResultDidNotContainJson,
-        DeserializationProducedNull
     }
 }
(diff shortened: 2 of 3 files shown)


dotnet/src/Microsoft.Agents.AI.Abstractions/AgentResponse{T}.cs
```diff
@@ -1,30 +1,117 @@
...
     /// </summary>
-    public abstract T Result { get; }
+    [JsonIgnore]
+    public virtual T Result
+    {
+        get
+        {
+            return (T)this.Deserialize(this.ResponseFormat?.SchemaType ?? typeof(T), this.ResponseFormat?.SchemaSerializerOptions ?? AgentAbstractionsJsonUtilities.DefaultOptions);
+        }
+    }
```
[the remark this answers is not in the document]
SergeyMenshykh: That's not a concern at the moment and can be done later if needed.

dotnet/src/Microsoft.Agents.AI.Abstractions/MEAI/NewChatResponseFormat.cs
```diff
@@ -0,0 +1,100 @@
...
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Agents.AI;
+
+#pragma warning disable CA1052 // Static holder types should be Static or NotInheritable
+
+/// <summary>Represents the response format that is desired by the caller.</summary>
+[JsonPolymorphic(TypeDiscriminatorPropertyName = "$type")]
+[JsonDerivedType(typeof(NewChatResponseFormatJson), typeDiscriminator: "json")]
+public partial class NewChatResponseFormat
```
rogerbarreto: Suggest putting this in the `Microsoft.Extensions.AI` namespace as this is a MEAI proposal.

dotnet/tests/Microsoft.Agents.AI.UnitTests/ChatClient/ChatClientAgent_StructuredOutput_WithFormatResponseTests.cs
rogerbarreto, line 1: Suggest keeping `StructuredOutput` in the filename.

dotnet/tests/Microsoft.Agents.AI.UnitTests/Models/Animal.cs
rogerbarreto, line 1: Similarly to Abstractions.UnitTests, suggest this two to be in different classes and internal.

```
dotnet/tests/Microsoft.Agents.AI.Abstractions.UnitTests/Models/
   ├── Animal.cs      (internal sealed class with [Description] attribute)
   └── Species.cs     (internal enum: Bear, Tiger, Walrus)
```

dotnet/src/Microsoft.Agents.AI.Abstractions/MEAI/NewChatResponseFormat.cs
SergeyMenshykh: Thanks for the suggestion, this change to the class is not necessary so it got removed.

dotnet/src/Microsoft.Agents.AI.Abstractions/AgentResponse.cs
```diff
@@ -68,6 +68,29 @@ public AgentResponse(ChatResponse response)
...
+    /// Initializes a new instance of the <see cref="AgentResponse"/> class from an existing <see cref="AgentResponse"/>.
+    /// </summary>
+    /// <param name="response">The <see cref="AgentResponse"/> from which to copy properties.</param>
+    /// <exception cref="ArgumentNullException"><paramref name="response"/> is <see langword="null"/>.</exception>
+    /// <remarks>
+    /// This constructor creates a copy of an existing agent response, preserving all
+    /// metadata and storing the original response in <see cref="RawRepresentation"/> for access to
+    /// the underlying implementation details.
+    /// </remarks>
+    public AgentResponse(AgentResponse response)
```
westey-m: Should we make this protected to match AgentRunOptions?

dotnet/src/Microsoft.Agents.AI.Abstractions/AgentResponse{T}.cs
```diff
@@ -1,30 +1,103 @@
...
+            return failureReason switch
+            {
+                FailureReason.ResultDidNotContainJson => throw new InvalidOperationException("The response did not contain JSON to be deserialized."),
+                FailureReason.DeserializationProducedNull => throw new InvalidOperationException("The deserialized response is null."),
+                _ => structuredOutput!,
+            };
+        }
+    }
+
+    private T? GetResultCore(JsonSerializerOptions serializerOptions, out FailureReason? failureReason)
```
westey-m: Not too serious, but we could probably move the contents of this method into Result.Get.  It was split previously due to the fact that we had TryDeserialize and Deserialize and this was the common logic.

dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
```diff
@@ -634,6 +634,12 @@ await session.AIContextProvider.InvokedAsync(new(inputMessages, aiContextProvide
                 chatOptions.AllowBackgroundResponses = agentRunOptions.AllowBackgroundResponses;
             }
 
+            if (agentRunOptions?.ResponseFormat is not null)
+            {
+                chatOptions ??= new ChatOptions();
+                chatOptions?.ResponseFormat = agentRunOptions.ResponseFormat;
```
westey-m: since we are initializing chatoptions on the previous line, the ? should not be required here.
```suggestion
                chatOptions.ResponseFormat = agentRunOptions.ResponseFormat;
```

PR #3694, opened 2026-02-05
.NET: Add decorator for structured output support

SergeyMenshykh:
### Motivation and Context

Currently, some of AF's AI agents don't support structured output. This is either because it's not part of the protocol, like with the A2A agent, or because the agents use LLMs without structured output capabilities. AF needs a solution to address this gap and provide structured output support for these agents.

### Description

This PR:
- Adds the StructuredOutputAgent decorator that obtains text from a decorated agent and delegates it to the configured chat client for JSON transformation.
- Adds the UseStructuredOutput extension method for the AIAgentBuilder, allowing easy registration of the decorator during the agent configuration step.
- Restructures the structured output sample to showcase different ways of working with structured output in AF.

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI/StructuredOutput/AIAgentBuilderExtensions.cs
+++ b/dotnet/src/Microsoft.Agents.AI/StructuredOutput/AIAgentBuilderExtensions.cs
@@ -20,8 +20,8 @@ public static class AIAgentBuilderExtensions
     /// The chat client used to transform text responses into structured JSON format.
     /// If <see langword="null"/>, the chat client will be resolved from the service provider.
     /// </param>
-    /// <param name="configure">
-    /// An optional callback that provides additional configuration of the <see cref="StructuredOutputAgentOptions"/> instance.
+    /// <param name="optionsFactory">
+    /// An optional factory function that returns the <see cref="StructuredOutputAgentOptions"/> instance to use.
     /// This allows for fine-tuning the structured output behavior such as setting the response format or system message.
     /// </param>
     /// <returns>The <see cref="AIAgentBuilder"/> with structured output capabilities added, enabling method chaining.</returns>
@@ -35,20 +35,12 @@ public static class AIAgentBuilderExtensions
     public static AIAgentBuilder UseStructuredOutput(
         this AIAgentBuilder builder,
         IChatClient? chatClient = null,
-        Action<StructuredOutputAgentOptions>? configure = null) =>
+        Func<StructuredOutputAgentOptions>? optionsFactory = null) =>
         Throw.IfNull(builder).Use((innerAgent, services) =>
         {
             chatClient ??= services?.GetService<IChatClient>()
                 ?? throw new InvalidOperationException($"No {nameof(IChatClient)} was provided and none could be resolved from the service provider. Either provide an {nameof(IChatClient)} explicitly or register one in the dependency injection container.");
 
-            StructuredOutputAgentOptions? options = null;
-
-            if (configure is not null)
-            {
-                options = new StructuredOutputAgentOptions();
-                configure(options);
-            }
-
-            return new StructuredOutputAgent(innerAgent, chatClient, options);
+            return new StructuredOutputAgent(innerAgent, chatClient, optionsFactory?.Invoke());
         });
 }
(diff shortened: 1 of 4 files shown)


dotnet/src/Microsoft.Agents.AI/StructuredOutput/AIAgentBuilderExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// <para>
+    /// A <see cref="ChatResponseFormatJson"/> must be specified either through the
+    /// <see cref="AgentRunOptions.ResponseFormat"/> at runtime or the <see cref="StructuredOutputAgentOptions.ChatOptions"/>
+    /// provided during configuration.
+    /// </para>
+    /// </remarks>
+    public static AIAgentBuilder UseStructuredOutput(
+        this AIAgentBuilder builder,
+        IChatClient? chatClient = null,
+        Action<StructuredOutputAgentOptions>? configure = null) =>
```
westey-m: Would it make sense to allow the implementer to return a `StructuredOutputAgentOptions` instead of having to modify one?  I expect it makes scenarios easier where the user already has an options they want to use in a few places, and they can just return the same one in those places.  Therefore `Func<StructuredOutputAgentOptions>` instead of `Action<StructuredOutputAgentOptions>`.
It's not like we are passing in anything of value, since we are just passing in a new empty StructuredOutputAgentOptions.

dotnet/src/Microsoft.Agents.AI/StructuredOutput/StructuredOutputAgentOptions.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// </summary>
+    public string? ChatClientSystemMessage { get; set; }
+
+    /// <summary>
+    /// Gets or sets the chat options to use for the structured output conversion by the chat client
+    /// used by the agent.
+    /// </summary>
+    /// <remarks>
+    /// The <see cref="ChatOptions.ResponseFormat"/> should be set to a <see cref="ChatResponseFormatJson"/>
+    /// instance to specify the expected JSON schema for the structured output.
```
westey-m: We should clarify that this is optional, and will be overridden by whatever is passed on AgentRunOptions.ResponseFormat

dotnet/src/Microsoft.Agents.AI/StructuredOutput/StructuredOutputAgentResponse.cs
```diff
@@ -0,0 +1,27 @@
...
+/// the original agent response from which the structured output was generated.
+/// </summary>
+public class StructuredOutputAgentResponse : AgentResponse
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="StructuredOutputAgentResponse"/> class.
+    /// </summary>
+    /// <param name="chatSoResponse">The <see cref="ChatResponse"/> containing the structured output.</param>
+    /// <param name="agentResponse">The original <see cref="AgentResponse"/> from the inner agent.</param>
+    internal StructuredOutputAgentResponse(ChatResponse chatSoResponse, AgentResponse agentResponse) : base(chatSoResponse)
```
westey-m: Is this a typo?
```suggestion
    /// <param name="chatResponse">The <see cref="ChatResponse"/> containing the structured output.</param>
    /// <param name="agentResponse">The original <see cref="AgentResponse"/> from the inner agent.</param>
    internal StructuredOutputAgentResponse(ChatResponse chatResponse, AgentResponse agentResponse) : base(chatResponse)
```

dotnet/src/Microsoft.Agents.AI/StructuredOutput/AIAgentBuilderExtensions.cs
SergeyMenshykh: Good point, it makes sense.

dotnet/samples/GettingStarted/Agents/Agent_Step05_StructuredOutput/Program.cs
```diff
@@ -8,64 +8,163 @@
...
+    IChatClient meaiChatClient = chatClient.AsIChatClient();
+
+    // Create the agent
+    AIAgent agent = meaiChatClient.AsAIAgent(name: "HelpfulAssistant", instructions: "You are a helpful assistant.");
 
-// Invoke the agent with some unstructured input while streaming, to extract the structured information from.
-var updates = agentWithPersonInfo.RunStreamingAsync("Please provide information about John Smith, who is a 35-year-old software engineer.");
+    // Add structured output middleware via UseStructuredOutput method to add structured output support to the agent.
+    // This middleware transforms the agent's text response into structured data using a chat client.
+    // A middleware that removes ResponseFormat is also added to emulate an agent that doesn't support structured output natively.
```
westey-m:
```suggestion
    // Since our agent does support structured output natively, we will add a middleware that removes ResponseFormat
    //  from the AgentRunOptions to emulate an agent that doesn't support structured output natively
```

PR #3726, opened 2026-02-06
.NET: Throw if structured output is not supported

SergeyMenshykh:
### Motivation and Context

Currently, to use structured output (SO) with AF agents, users need to know whether a particular agent, or in the case of the ChatClientAgent, a particular chat client supports SO. If the agent supports SO, the `RunAsync<T>` methods can be used safely. If the agent does not support SO, then SO middleware can be used; otherwise, the user may receive a plain text response instead of an SO response, causing deserialization (accessing the response.Result property) to fail.
  
If users mistakenly try to get SO from an agent that does not natively support it and no SO middleware is used, it would be helpful to fail early and notify the `RunAsync<T>` caller that SO is not supported. However, this can only be done with agents that AF "knows" do not support SO. It won't work with agents like the ChatClientAgent, whose SO capabilities depend on the chat client being used. In those cases, it is still the user's responsibility to either use an SO-capable chat client or use middleware.

### Description

- Add the SupportsStructuredOutput property to the AIAgentMetadata class.
- Set the property to `true` for two agents: ChatClientAgent and DurableAIAgent.
- Throw an exception in the `RunAsync<T>` method if SO is not supported.

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.DurableTask/CHANGELOG.md
+++ b/dotnet/src/Microsoft.Agents.AI.DurableTask/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Removed AgentThreadMetadata and used AgentSessionId directly instead ([#3067](https://github.com/microsoft/agent-framework/pull/3067));
 - Renamed AgentThread to AgentSession ([#3430](https://github.com/microsoft/agent-framework/pull/3430))
 - Updated to use base `AgentRunOptions.ResponseFormat` for structured output configuration ([#3658](https://github.com/microsoft/agent-framework/pull/3658))
+- Added structured output support detection and metadata to DurableAIAgent ([#3726](https://github.com/microsoft/agent-framework/pull/3726))
 
 ## v1.0.0-preview.251204.1
 
--- a/dotnet/tests/Microsoft.Agents.AI.Abstractions.UnitTests/AIAgentStructuredOutputTests.cs
+++ b/dotnet/tests/Microsoft.Agents.AI.Abstractions.UnitTests/AIAgentStructuredOutputTests.cs
@@ -21,6 +21,12 @@ public class AIAgentStructuredOutputTests
     public AIAgentStructuredOutputTests()
     {
         this._agentMock = new Mock<AIAgent> { CallBase = true };
+
+        // Setup GetService to return AIAgentMetadata with SupportsStructuredOutput = true
+        // This is required for RunAsync<T> to work, as it checks for structured output support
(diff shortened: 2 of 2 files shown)


dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs
```diff
@@ -97,7 +97,7 @@ public ChatClientAgent(IChatClient chatClient, ChatClientAgentOptions? options,
         // Options must be cloned since ChatClientAgentOptions is mutable.
         this._agentOptions = options?.Clone();
 
-        this._agentMetadata = new AIAgentMetadata(chatClient.GetService<ChatClientMetadata>()?.ProviderName);
+        this._agentMetadata = new AIAgentMetadata(chatClient.GetService<ChatClientMetadata>()?.ProviderName, supportsStructuredOutput: true);
```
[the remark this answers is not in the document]
SergeyMenshykh: For the ChatClientAgent case, it's still the responsibility of a developer to configure it with an SO-capable chat client.

markwallace-microsoft, line 100: Given we don't know the capabilities of the model associated with the `chatClient` what is the thinking behind setting this to be true?

westey-m: I'm also in two minds whether it makes sense.  Providing true here is technically misleading, and if the service/model does not support structured output the call will eventually fail anyway when we try and deserialize the results.  If this property was only for our own consumption to fail early, it may not be as bad, but the fact that it is public to all users isn't great.
Consider for example, if a user uses this metadata to decide whether to decorate a random AIAgent from DI with a Structured Output Agent to add structured output capabilities.  They would expect it to always work, but in some cases it may still fail, since we said true here, but the model may not actually support it.

SergeyMenshykh: There's no way the functionality is going to work reliably with the chat client agent. Even if the chat client exposes a way to check if SO is supported and the chat client agent uses it, it will still be up to the developer to configure the particular chat client with the model and specify if SO is supported by the model or not. If the developer mistakenly specifies that the model supports SO but it does not, AF will fail in the same way as it would without this functionality at the point of accessing the agent's SO result - response.Result. Therefore, I would suggest keeping it simple and not adding this feature.

PR #3732, opened 2026-02-06
.NET: Implement Task support for A2A Hosting package

### LINKED ISSUE #406: .NET: A2A Protocol/SDK integration in agentic-framework
Issue describes the A2A support in agentic-framework. 

### Description

A2A protocol has 2 different foundational entities - `Message` and `AgentTask` where the latter is a representation of a potentially long-running operation that agent takes in the work. `Message` on the other hand is a simple input-response model of communication between agent/client.

Official C#/.NET SDK for A2A Protocol is [a2aproject/a2a-dotnet](https://github.com/a2aproject/a2a-dotnet). SDK has a fundamental class `TaskManager` which allows to subscribe to different events (like `OnTaskCreated` or `OnMessageReceived`) and implement your own handling - in our case it should be invoking the agentic-framework.

A2A integration should exist in a separate package `Microsoft.Extensions.AI.Agents.A2A`. Users referencing this package should obtain API to connect their `AIAgent` to the A2A processing mechanisms (via the official C# A2A SDK), and possibly the implementation of another `AIAgent`.

A simplified integration of A2A to the AspNetCore apps should be exposed in the package `Microsoft.Extensions.AI.Agents.A2A.AspNetCore` allowing users to expose A2A endpoints in their aspnetcore apps.

### Message-based processing

Message-based processing of A2A communication is only referring to `TaskManager.OnMessageReceived`. `OnMessageReceived` theoretically can either return an `AgentTask` or a `Message`. Here we are only discussing `Message`.

I am proposing such an API (below) which allows to connect the `TaskManager` and `AIAgent` (can be any implementation).
This API requires `IActorClient` in the same fashion as currently server-side handling of user request requires `IActorClient` to map the request to agent+context. In case of A2A agent is determined based on the endpoint group (i.e. call to `/a2a/pirate` obviously calls the pirate `AIAgent`), and contextId is taken from the A2A message payload. It is mapped to `ActorId.Key`.
```diff
+ namespace Microsoft.Extensions.AI.Agents.A2A;

+ public static class AIAgentExtensions
{
+  public static void AttachA2AMessaging(this AIAgent agent, IActorClient actorClient, TaskManager taskManager, ILoggerFactory? loggerFactory = null)
}
```

And I am proposing an similar API (below) in A2A.AspNetCore package. This API is extensible (one can pass their own TaskManager to map A2A to the specific endpoint) or to rely on our implementation to connect passed `AIAgent` and `TaskManager`. First method here should call one from `AI.Agents.A2A` package.
```diff
+ namespace Microsoft.Extensions.AI.Agents.A2A.AspNetCore;

+ public static class WebApplicationExtensions
{
+ public static void AttachA2AMessaging(this WebApplication app, string agentName, string path);
+ public static void AttachA2A(this WebApplication app, TaskManager taskManager, string path);
}
```

### AgentTask-based processing

This is a more complex thing to implement, and it will be designed later (cc @SergeyMenshykh)

DeagleGross:
PR adds the support to [A2A Tasks](https://a2a-protocol.org/latest/specification/#313-get-task) per protocol. As before it uses [A2A SDK](https://github.com/a2aproject/a2a-dotnet) to subscribe to the events of receiving task/message updates and creation:
```csharp
        taskManager.OnMessageReceived += OnMessageReceivedAsync;
        taskManager.OnTaskUpdated += OnTaskUpdatedAsync;
        taskManager.OnTaskCancelled += OnTaskCancelledAsync;
```

Task sessions are stored in the `AgentSessionStore`, and continuation token for the task is stored in the session-metadata.

Closes #406


dotnet/src/Microsoft.Agents.AI.Hosting.A2A/AIAgentExtensions.cs
```diff
@@ -36,16 +39,33 @@ public static ITaskManager MapA2A(
...
         return taskManager;
 
         async Task<A2AResponse> OnMessageReceivedAsync(MessageSendParams messageSendParams, CancellationToken cancellationToken)
         {
             var contextId = messageSendParams.Message.ContextId ?? Guid.NewGuid().ToString("N");
             var session = await hostAgent.GetOrCreateSessionAsync(contextId, cancellationToken).ConfigureAwait(false);
             var options = messageSendParams.Metadata is not { Count: > 0 }
-                ? null
-                : new AgentRunOptions { AdditionalProperties = messageSendParams.Metadata.ToAdditionalProperties() };
+                ? new AgentRunOptions { AllowBackgroundResponses = true }
```
SergeyMenshykh: Enabling background responses here means that hostAgent will always return a background response if supported, and the A2A client will always receive a task as a result, even for lightweight requests that don't require task and background handling.

I think it should be configurable by users, and users can decide to: a. always reply with a message, b. always reply with a task, or c. delegate the decision to reply with a message or a task to the LLM (via IChatClient).

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/AIAgentExtensions.cs
```diff
@@ -64,6 +98,162 @@ async Task<A2AResponse> OnMessageReceivedAsync(MessageSendParams messageSendPara
...
+        {
+            AgentTask agentTask = await taskManager.CreateTaskAsync(contextId, cancellationToken: cancellationToken).ConfigureAwait(false);
+
+            // Serialize the continuation token into the task's metadata so it survives
+            // across requests and is cleaned up with the task itself.
+#pragma warning disable MEAI001 // Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.
+            agentTask.Metadata ??= [];
+            agentTask.Metadata[ContinuationTokenMetadataKey] = JsonSerializer.SerializeToElement(
+                initialResponse.ContinuationToken,
+                AIJsonUtilities.DefaultOptions.GetTypeInfo(typeof(ResponseContinuationToken)));
```
SergeyMenshykh:
```suggestion
                AgentJsonUtilities.DefaultOptions.GetTypeInfo(typeof(ResponseContinuationToken)));
```

SergeyMenshykh: JsonSerializerOptions should be configurable to serialize the continuation token for a specific agent. The continuation token may contain types that are not registered in the default options. In case JsonSerializerOptions are not provided, the code should fall back to the default options in AgentJsonUtilities. Same for deserialization part.

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/AIAgentExtensions.cs
```diff
@@ -64,6 +98,162 @@ async Task<A2AResponse> OnMessageReceivedAsync(MessageSendParams messageSendPara
                 Metadata = response.AdditionalProperties?.ToA2AMetadata()
             };
         }
+
+        async Task<AgentTask> CreateWorkingTaskAsync(
+            string contextId,
+            AgentResponse initialResponse,
+            CancellationToken cancellationToken)
+        {
+            AgentTask agentTask = await taskManager.CreateTaskAsync(contextId, cancellationToken: cancellationToken).ConfigureAwait(false);
```
SergeyMenshykh: Original messages should be added to the task history: https://github.com/a2aproject/a2a-dotnet/blob/a0533495f85a28f2b7aee60fb2a02d3e63515dbb/src/A2A/Server/TaskManager.cs#L184

SergeyMenshykh: Ideally, instead of duplicating task creation code here, it would be good if the OnMessageReceived handler took a task factory argument so we could reuse the task creation functionality.

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/AIAgentExtensions.cs
```diff
@@ -64,6 +98,162 @@ async Task<A2AResponse> OnMessageReceivedAsync(MessageSendParams messageSendPara
...
+
```
SergeyMenshykh, line 246: Consider calling UpdateStatusAsync with `TaskState.Submitted` after creating the task to communicate task creation to the client: https://a2a-protocol.org/latest/specification/#413-taskstate

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/AIAgentExtensions.cs
```diff
@@ -64,6 +98,162 @@ async Task<A2AResponse> OnMessageReceivedAsync(MessageSendParams messageSendPara
...
+                            await taskManager.UpdateStatusAsync(agentTask.Id, TaskState.Working, message: progressMessage, cancellationToken: cancellationToken).ConfigureAwait(false);
+                        }
+                    }
+                    else
+                    {
+                        // Background operation completed — remove the token from metadata
+                        agentTask.Metadata!.Remove(ContinuationTokenMetadataKey);
+
+                        var agentMessage = CreateMessageFromResponse(contextId, response);
+                        await taskManager.UpdateStatusAsync(
```
SergeyMenshykh: A2A requires task output to be returned as an artifact rather than as a message.
  
_Messages SHOULD NOT be used to deliver task outputs. Results SHOULD BE returned using artifacts associated with a task. This separation allows for a clear distinction between communication (messages) and data output (artifacts)._ Source: https://a2a-protocol.org/latest/specification/#37-messages-and-artifacts  
  
We might need a heuristic here to decide which message of the hosted agent's response should be returned as a message and which as an artifact. The simplest heuristic could be to return completed background operation output as an artifact. A more advanced approach could use IChatClient to decide what goes where.

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/AIAgentExtensions.cs
DeagleGross: makes sense, added a possibility to control `A2AResponseMode` for the user

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/AIAgentExtensions.cs
DeagleGross: added a new type `A2AHostingJsonUtilities` and allowed to pass "your own" JsonSerializerOptions to control this for complex scenarios

DeagleGross: Makes sense, allowed to do it. User now can pass their own jsonSerializerOptions and I added `A2AHostingJsonUtilities` as a public type so that they can base their jsonSerializerOptions on top of pre-setup a2a hosting one.

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/AIAgentExtensions.cs
DeagleGross: Thanks! Added

DeagleGross: refactored the code here, so should be better now - please take a look

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/AIAgentExtensions.cs
DeagleGross: done it!

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/AIAgentExtensions.cs
DeagleGross: added support for artifacts

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/A2AHostingJsonUtilities.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Text.Json;
+
+namespace Microsoft.Agents.AI.Hosting.A2A;
+
+/// <summary>
+/// Provides JSON serialization options for A2A Hosting APIs to support AOT and trimming.
+/// </summary>
+public static class A2AHostingJsonUtilities
```
SergeyMenshykh, line 10: The existing A2AJsonUtilities class in Microsoft.Agents.AI.A2A should be used instead. It already has A2A-specific types and will evolve with the agent. The only thing it misses, which is a bug, is that it doesn't chain type info resolvers from the A2A SDK like this utility class does in line 26.

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/AIAgentExtensions.cs
```diff
@@ -54,16 +86,184 @@ async Task<A2AResponse> OnMessageReceivedAsync(MessageSendParams messageSendPara
...
+        }
+
+        async Task CompleteWithArtifactAsync(string taskId, AgentResponse response, CancellationToken cancellationToken)
+        {
+            var artifact = CreateArtifactFromResponse(response);
+            await taskManager.ReturnArtifactAsync(taskId, artifact, cancellationToken).ConfigureAwait(false);
+            await taskManager.UpdateStatusAsync(taskId, TaskState.Completed, final: true, cancellationToken: cancellationToken).ConfigureAwait(false);
+        }
+
+        async Task OnTaskUpdatedAsync(AgentTask agentTask, CancellationToken cancellationToken)
```
SergeyMenshykh: nit: consider promoting this local function and the other "big" ones to separate private methods to improve readability.

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/AIAgentExtensions.cs
```diff
@@ -36,16 +43,41 @@ public static ITaskManager MapA2A(
...
+        // See https://github.com/a2aproject/a2a-dotnet/issues/275
         taskManager.OnMessageReceived += OnMessageReceivedAsync;
+
+        // Task flow for subsequent updates and cancellations
+        taskManager.OnTaskUpdated += OnTaskUpdatedAsync;
+        taskManager.OnTaskCancelled += OnTaskCancelledAsync;
+
         return taskManager;
 
         async Task<A2AResponse> OnMessageReceivedAsync(MessageSendParams messageSendParams, CancellationToken cancellationToken)
```
SergeyMenshykh: nit: consider promoting this local function and the other "big" ones to separate private methods to improve readability.

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/AIAgentExtensions.cs
```diff
@@ -54,16 +86,184 @@ async Task<A2AResponse> OnMessageReceivedAsync(MessageSendParams messageSendPara
...
+            try
+            {
+                // If this task has a pending continuation token in its metadata, check on
+                // the background operation instead of processing new messages from history.
+                if (TryExtractContinuationToken(agentTask, continuationTokenJsonOptions, out var continuationToken))
+                {
+                    var response = await hostAgent.RunAsync(
+                        session: session,
+                        options: new AgentRunOptions { ContinuationToken = continuationToken },
+                        cancellationToken: cancellationToken).ConfigureAwait(false);
```
SergeyMenshykh: The OnTaskUpdatedAsync callback is called when a user sends a new message to an existing task. However, the new message is not passed to the RunAsync method, and as a result, it will be lost. Additionally, even if the message were passed to the method, RunAsync would reject the call because AF agents don't support updating existing background responses (long-running operations). 

To address this issue, we can create a new task with RunAsync using the existing session and messages from the chat history of the original task.

CC: @westey-m

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/A2AResponseMode.cs
```diff
@@ -0,0 +1,31 @@
...
+/// <see cref="AIAgentExtensions"/>.
+/// </summary>
+public enum A2AResponseMode
+{
+    /// <summary>
+    /// The decision is delegated to the agent. Background responses are enabled
+    /// and if the agent returns a continuation token (indicating a long-running operation),
+    /// an <c>AgentTask</c> is returned. Otherwise an <c>AgentMessage</c> is returned.
+    /// </summary>
+    Auto = 0,
```
SergeyMenshykh: The Auto constant is not used anywhere at the moment, so we can maybe drop it for now.  
  
Perhaps we can redesign the mode as a class so it can support a dynamic mode, and its usage pattern looks like this:  
- A2AResponseMode.Message  
- A2AResponseMode.Task  
- A2AResponseMode.Dynamic((message) => LLM call or heuristic to decide on message/task mode)  
  
We probably won't need Dynamic now, but we might need to be ready to add it when needed without breaking changes (refactoring the A2AResponseMode enum to a class).

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/A2AHostingJsonUtilities.cs
DeagleGross: I think we should not have a dependency to `Microsoft.Agents.AI.A2A` from the `Microsoft.Agents.AI.Hosting.A2A` (aka not refercing a "client" package from the "hosting" library). Same as OpenAI.Responses Hosting package does not reference a "client" package: https://github.com/microsoft/agent-framework/blob/main/dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Microsoft.Agents.AI.Hosting.OpenAI.csproj

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/A2AResponseMode.cs
DeagleGross: > The Auto constant is not used anywhere at the moment 

it is a default mode for when registring agent to use A2A protocol - see `MapA2A()` API overloads. I dont see a difference between `Dynamic` and `Auto` - do you like the naming more or you are suggesting to drop the Auto/Dynamic mode completely for now?

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/A2AHostingJsonUtilities.cs
SergeyMenshykh: Agree, we should not.

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/A2AResponseMode.cs
SergeyMenshykh: I'm suggesting we model A2AResponseMode in a way that, when there's a need to add a Dynamic/Auto mode, we can easily do so with no breaking changes - rather than having to refactor it from an enum to a class and causing breaking changes.

Regarding Auto vs Dynamic: A2AResponseMode.Auto reads as "auto mode," which implies the absence of manual intervention but doesn't convey what is being decided. A2AResponseMode.Dynamic reads as "dynamic mode," meaning the mode is not static - it's determined at runtime. With the additional context of knowing the two other modes, Message and Task, it's easy to infer that the dynamic mode allows you to dynamically decide between them per message.

DeagleGross: Adjusted to as what we have discussed with Dynamic mode which has the context and provides the way for the user to configure the decision

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/A2AResponseMode.cs
```diff
@@ -0,0 +1,113 @@
...
+    {
+        this._value = value;
+        this._decide = decide;
+    }
+
+    /// <summary>
+    /// Always return an <c>AgentMessage</c>. Background responses are not enabled.
+    /// Suitable for lightweight, single-shot request/response interactions.
+    /// </summary>
+    public static A2AResponseMode Message { get; } = new("message");
```
SergeyMenshykh: Maybe have the strings as constants so you can reuse them.

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/A2AResponseMode.cs
```diff
@@ -0,0 +1,113 @@
...
+    /// Gets whether background (long-running) responses are enabled for this mode.
+    /// </summary>
+    internal bool AllowBackgroundResponses =>
+        !string.Equals(this._value, "message", StringComparison.OrdinalIgnoreCase);
+
+    /// <summary>
+    /// Determines whether the agent response should be returned as an <c>AgentTask</c>.
+    /// </summary>
+#pragma warning disable MEAI001 // Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.
+    internal ValueTask<bool> ShouldReturnAsTaskAsync(A2AResponseDecisionContext context)
```
SergeyMenshykh: It should also accept cancellation token so it can be cancelled.

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/AIAgentExtensions.cs
```diff
@@ -97,4 +104,210 @@ public static ITaskManager MapA2A(
...
+        ITaskManager taskManager,
+        JsonSerializerOptions continuationTokenJsonOptions,
+        CancellationToken cancellationToken)
+    {
+        var contextId = messageSendParams.Message.ContextId ?? Guid.NewGuid().ToString("N");
+        var session = await hostAgent.GetOrCreateSessionAsync(contextId, cancellationToken).ConfigureAwait(false);
+
+        // Only enable background responses when the mode allows task-based results.
+        // In Message mode, background responses are never enabled so the agent always completes synchronously.
+        bool allowBackground = responseMode.AllowBackgroundResponses;
```
SergeyMenshykh: The decision whether to allow background responses or not should be based on the incoming message and user configuration, rather than solely on user configuration. As such, the `ShouldReturnAsTaskAsync` method should be used here instead of the `AllowBackgroundResponses` property, which can be dropped.

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/AIAgentExtensions.cs
```diff
@@ -97,4 +104,210 @@ public static ITaskManager MapA2A(
...
+    {
+        var contextId = agentTask.ContextId ?? Guid.NewGuid().ToString("N");
+        var session = await hostAgent.GetOrCreateSessionAsync(contextId, cancellationToken).ConfigureAwait(false);
+
+        try
+        {
+            // Discard any stale continuation token — the incoming user message supersedes
+            // any previous background operation. AF agents do not support resuming a
+            // background run while injecting new messages; we start a fresh run from the
+            // existing session using the full chat history (which includes the new message).
```
SergeyMenshykh, line 173:
```suggestion
            // Discard any stale continuation token — the incoming user message supersedes
            // any previous background operation. AF agents don't support updating existing 
            // background responses (long-running operations); we start a fresh run from the
            // existing session using the full chat history (which includes the new message).
```

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/AIAgentExtensions.cs
DeagleGross: So as per our discussion: renamed the `A2AResponseMode` to `AgentRunMode`, having `AgentRunMode.NonBackground`, `AgentRunMode.BackgroundIfSupported` and `AgentRunMode.Dynamic` modes, and it does not impact the response, but rather the agent run.

renamed everywhere, default is `AgentRunMode.NonBackground`

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/AgentRunMode.cs
```diff
@@ -0,0 +1,102 @@
...
+    /// <summary>
+    /// The response type is decided by the supplied <paramref name="decideAsTask"/> delegate.
+    /// The delegate receives an <see cref="A2ARunDecisionContext"/> with the incoming
+    /// message and the agent response, and returns <see langword="true"/> to return an
+    /// <c>AgentTask</c> or <see langword="false"/> to return an <c>AgentMessage</c>.
+    /// Background responses are enabled.
+    /// </summary>
+    /// <param name="decideAsTask">
+    /// An async delegate that decides whether the response should be wrapped in an <c>AgentTask</c>.
+    /// </param>
```
SergeyMenshykh, line 53:
```suggestion
   /// <summary>
   /// The agent run mode is decided by the supplied <paramref name="decideAsTask"/> delegate.
   /// The delegate receives an <see cref="A2ARunDecisionContext"/> with the incoming
   /// message and returns a boolean specifying whether to run the agent in background mode.
   /// <see langword="true"/> indicates that the agent should run in background mode and return an
   /// <c>AgentTask</c> if the agent supports background mode; otherwise, it returns an <c>AgentMessage</c>
   /// if the mode is not supported. <see langword="false"/> indicates that the agent should run in
   /// non-background mode and return an <c>AgentMessage</c>.
   /// </summary>
   /// <param name="decideAsTask">
   /// An async delegate that decides whether the response should be wrapped in an <c>AgentTask</c>.
   /// </param>
```

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/AgentRunMode.cs
```diff
@@ -0,0 +1,102 @@
...
+    public static AgentRunMode Dynamic(Func<A2ARunDecisionContext, CancellationToken, ValueTask<bool>> decideAsTask)
```
SergeyMenshykh: Consider other names for the `decideAsTask` parameter that don't include "task," such as: predicate, decideBackgroundMode, runInBackground, or isBackgroundRun.

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/AgentRunMode.cs
```diff
@@ -0,0 +1,104 @@
...
+    {
+        this._value = value;
+        this._runInBackground = runInBackground;
+    }
+
+    /// <summary>
+    /// Dissallows the background responses from the agent. Is equivalent to configuring <see cref="AgentRunOptions.AllowBackgroundResponses"/> as <c>false</c>.
+    /// In the A2A protocol terminology will make responses be returned as <c>AgentMessage</c>.
+    /// </summary>
+    public static AgentRunMode NonBackground => new(MessageValue);
```
westey-m: Some additional naming to consider (to match the AgentRunOptions naming more closely):

- DisallowBackground()
- AllowBackground()
- AllowBackgroundWhen(Func<A2ARunDecisionContext, CancellationToken, ValueTask<bool>> runInBackground)

With Dynamic, where even if you return true, it doesn't mean that it would use background, it's good to make it clear using naming.

CC @SergeyMenshykh

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/AgentRunMode.cs
```diff
@@ -0,0 +1,104 @@
...
+    public static AgentRunMode Dynamic(Func<A2ARunDecisionContext, CancellationToken, ValueTask<bool>> runInBackground)
+    {
+        ArgumentNullException.ThrowIfNull(runInBackground);
+        return new(DynamicValue, runInBackground);
+    }
+
+    /// <summary>
+    /// Determines whether the agent response should be returned as an <c>AgentTask</c>.
+    /// </summary>
+#pragma warning disable MEAI001 // Type is for evaluation purposes only and is subject to change or removal in future updates. Suppress this diagnostic to proceed.
```
westey-m: Where features are experimental, we should propagate that up, so users know, rather than suppressing and then users are inadvertently using an experimental feature.
So instead of suppressing here, we should just mark AgentRunMode as experimental as well.

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/AIAgentExtensions.cs
```diff
@@ -97,4 +104,200 @@ public static ITaskManager MapA2A(
...
+        CancellationToken cancellationToken)
+    {
+        var contextId = messageSendParams.Message.ContextId ?? Guid.NewGuid().ToString("N");
+        var session = await hostAgent.GetOrCreateSessionAsync(contextId, cancellationToken).ConfigureAwait(false);
+
+        // Decide whether to run in background based on user preferences and agent capabilities
+        var decisionContext = new A2ARunDecisionContext(messageSendParams);
+        var allowBackgroundResponses = await runMode.ShouldRunInBackgroundAsync(decisionContext, cancellationToken).ConfigureAwait(false);
+
+        var options = messageSendParams.Metadata is not { Count: > 0 }
```
westey-m, line 135: I don't see us supporting `referenceTaskIds` on messages.  Should we throw if someone uses that feature?
See `https://a2a-protocol.org/latest/topics/life-of-a-task/#example-follow-up-scenario`
Or is that supported via OnTaskUpdated?

That said, continuing from some arbitrary previous point is not supported by AIAgent directly.  It's only supported by responses when using response ids and A2A after all.

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/AgentRunMode.cs
DeagleGross: It was a long discussion with Sergey when we landed on previous naming, but I like this one.

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/AIAgentExtensions.cs
DeagleGross: I added throwing. It will take some time to design continuation via `referenceTaskIds` I think, so lets keep it simple here.

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/AgentRunMode.cs
DeagleGross: Sure, added `[Experimental(DiagnosticIds.Experiments.AIResponseContinuations)]` on `AgentRunMode` and dependent components (like MapA2A() API).

PR #3792, opened 2026-02-10
.NET: [BREAKING] Implement Polymorphic Routing

### LINKED ISSUE #782: .NET: Route messages are not handled polymorphically

### LINKED ISSUE #3116: Support for polymorphic routing

### LINKED ISSUE #3242: .NET: Reconcile Executor Configure and ConfigureXYZ methods
Pending the Source Generator and Agent Hosting work, see #3131 and #3142, respectively

lokitoth:
### Motivation and Context

The initial implementation of the InProcess runtime relied on exact matches on handled types to determine routing. This was a result of the need to be able to go from a `TypeId` to a runtime `Type` after a checkpoint serialization/deserialization cycle, which loses type information. With the changes to provide a Source Generator rather than Reflection, and a more expressive `ProtocolBuilder` API to be able to define the types of sent messages and yielded outputs, we can switch the deserialization logic to rely on the source executor's information, 

### Description

Implements Polymorphic routing (closes #782, closes #3116)

- [BREAKING] Changes `ConfigureRoutes`/`ConfigureXYZTypes` to `ConfigureProtocol` (closes #3242)
- [BREAKING] Adds check for outgoing messages and outputs to match registered types
- Temporarily updates ReflectingExecutor (which is Obsolete) to properly register outgoing types
- Update FunctionExecutor to register outgoing types and listen to `[SendsMessage]` and `[YieldsOutput]` attributes
- TODO:
  - [x] Add support for attributes-based outgoing types registration to Basic Executors
  - [x] Update Declarative Workflows for missing type registrations (if any)

### Contribution Checklist

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/ObjectModel/InvokeAzureAgentExecutor.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/ObjectModel/InvokeAzureAgentExecutor.cs
@@ -17,6 +17,7 @@ using Microsoft.Shared.Diagnostics;
 
 namespace Microsoft.Agents.AI.Workflows.Declarative.ObjectModel;
 
+[SendsMessage(typeof(ExternalInputRequest))]
 internal sealed class InvokeAzureAgentExecutor(InvokeAzureAgent model, ResponseAgentProvider agentProvider, WorkflowFormulaState state) :
     DeclarativeActionExecutor<InvokeAzureAgent>(model, state)
 {


dotnet/src/Microsoft.Agents.AI.Workflows/Executor.cs
```diff
@@ -15,6 +16,134 @@
...
+    internal MessageRouter Router => router;
+
+    //public bool CanHandle(TypeId typeId) => router.CanHandle(typeId);
+    public bool CanHandle(Type type) => router.CanHandle(type);
+
+    //public bool CanOutput(TypeId typeId) => this._yieldTypes.Contains(typeId);
+    public bool CanOutput(Type type) => this._yieldTypes.Contains(new(type));
+
+    public ProtocolDescriptor Describe() => new(this.Router.IncomingTypes, yieldTypes, sendTypes, this.Router.HasCatchAll);
+}
```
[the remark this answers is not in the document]
lokitoth: This is by design: this increase in strictness on the send-side is what allows for Polymorphic routing on the receiving side.

dotnet/src/Microsoft.Agents.AI.Workflows/Execution/MessageRouter.cs
```diff
@@ -49,15 +61,45 @@ internal MessageRouter(Dictionary<Type, MessageHandlerF> handlers, HashSet<Type>
...
+                foreach (Type interfaceType in this._interfaceHandlers)
+                {
+                    if (interfaceType.IsAssignableFrom(candidateType))
+                    {
+                        handler = this._typedHandlers[interfaceType];
+                        this._typedHandlers[messageType] = handler;
+
+                        // TODO: This could cause some consternation with Checkpointing (need to ensure we surface errors well)
+                        this._runtimeTypeMap[new TypeId(messageType)] = interfaceType;
+                        return handler;
```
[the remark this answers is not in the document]
lokitoth: The `MessageRouter` is tied to a single executor instance, and any given `Executor` can only have a single logical thread. On the send side, this is single-threaded because the sending Executor only gets one thread to process messages. On the receiving side this is re-aggregated by receiving Executor and processed, again, in a single-thread-per-Executor manner.

These will only be updated from a single thread.

dotnet/src/Microsoft.Agents.AI.Workflows/Executor.cs
```diff
@@ -15,6 +16,134 @@
...
+                    this._delcaredTypeMap[messageType] = declaredTypeId;
+                }
+
+                return declaredTypeId;
+            }
+            if (!allowPolymorphism)
+            {
+                break;
+            }
+        }
```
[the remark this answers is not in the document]
lokitoth: This is vestigial - we never set `allowPolymorphism` to false. Should be removed.

dotnet/src/Microsoft.Agents.AI.Workflows/InProc/InProcessRunner.cs
```diff
@@ -201,14 +201,48 @@ private async ValueTask DeliverMessagesAsync(string receiverId, ConcurrentQueue<
...
+                    !portableValue.IsType(actualType, out value))
+                {
+                    throw new InvalidOperationException($"Cannot interpret incoming message of type '{portableValue.TypeId}' as type '{actualType.FullName}'.");
+                }
+            }
+
+            if (envelope.MessageType != messageType)
+            {
+                Console.Error.WriteLine($"Translated type of delivery: {envelope.MessageType} to {messageType}");
+            }
```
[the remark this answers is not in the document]
lokitoth: Was for debugging during development. Should be removed.

dotnet/tests/Microsoft.Agents.AI.Workflows.UnitTests/Sample/08_Subworkflow_Simple.cs
```diff
@@ -46,6 +48,22 @@ public static async ValueTask<List<TextProcessingResult>> RunAsync(TextWriter wr
...
+                                                            .Select(errorEvent => errorEvent.Exception)
+                                                            .Where(e => e is not null).ToList();
+        if (errors.Count > 0)
+        {
+            StringBuilder errorBuilder = new();
+            errorBuilder.AppendLine($"Workflow execution failed. ({errors.Count} errors.):");
+
+            foreach (Exception? error in errors)
+            {
+                errorBuilder.Append('\t').AppendLine(error!.ToString());
```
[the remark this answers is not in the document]
lokitoth: error is an `Exception`, not a `String` object.

dotnet/tests/Microsoft.Agents.AI.Workflows.UnitTests/Sample/09_Subworkflow_ExternalRequest.cs
```diff
@@ -256,6 +257,22 @@ public static async ValueTask<List<RequestFinished>> RunAsync(TextWriter writer,
...
+                                                            .Select(errorEvent => errorEvent.Exception)
+                                                            .Where(e => e is not null).ToList();
+        if (errors.Count > 0)
+        {
+            StringBuilder errorBuilder = new();
+            errorBuilder.AppendLine($"Workflow execution failed. ({errors.Count} errors.):");
+
+            foreach (Exception? error in errors)
+            {
+                errorBuilder.Append('\t').AppendLine(error!.ToString());
```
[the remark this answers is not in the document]
lokitoth: error is an Exception, not a String object.

dotnet/src/Microsoft.Agents.AI.Workflows.Generators/Generation/SourceBuilder.cs
```diff
@@ -53,38 +55,57 @@ public static string Generate(ExecutorInfo info)
             {
                 sb.AppendLine($"{indent}partial class {containingType}");
                 sb.AppendLine($"{indent}{{");
-                indent += "    ";
+                indent += IndentUnit;
```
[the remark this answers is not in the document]
lokitoth: The nesting level is unlikely to be large. The other string operations should dominate.

lokitoth:
Re:

> FanInEdgeRunner uses wrong TypeId for polymorphic type resolution

Updated the code for defense in depth, but in principle a "perf" pass should be able to remove this code.

> Source generator no longer registers handler return types as yielded outputs

This is because the `ProtocolBuilder` now universally [manages that](https://github.com/microsoft/agent-framework/pull/3792/changes#diff-42d07306f34aa886da8a448aa14adfe0d2f60d00bf41342b048e079a2abc307bR132-R142) in its `Build` method. This change should help prevent errors like accidentally forgetting to handle it from manually defined protocol builder invocations, especially if options can be provided after the fact.

> ReflectingExecutor crashes if subclass has no IMessageHandler interfaces

Added a check for this

dotnet/src/Microsoft.Agents.AI.Workflows/Execution/FanInEdgeRunner.cs
```diff
@@ -41,18 +42,51 @@ internal sealed class FanInEdgeRunner(IRunnerContext runContext, FanInEdgeData e
...
+            }
+
+            IEnumerable<(Type?, MessageEnvelope)> MapRuntimeTypes((ExecutorProtocol, IGrouping<ExecutorIdentity, MessageEnvelope>) input)
+            {
+                (ExecutorProtocol protocol, IGrouping<ExecutorIdentity, MessageEnvelope> grouping) = input;
+                return grouping.Select(envelope => (ResolveEnvelopeType(envelope), envelope));
+
+                Type? ResolveEnvelopeType(MessageEnvelope messageEnvelope)
+                {
+                    if (envelope.Message is PortableValue portableValue)
```
alliscode: Should this be `messageEnvelope`?


```suggestion
                    if (messageEnvelope.Message is PortableValue portableValue)
```

dotnet/src/Microsoft.Agents.AI.Workflows/Execution/FanInEdgeRunner.cs
```diff
@@ -41,18 +42,51 @@ internal sealed class FanInEdgeRunner(IRunnerContext runContext, FanInEdgeData e
...
+                    {
+                        return protocol.SendTypeTranslator.MapTypeId(portableValue.TypeId);
+                    }
+
+                    return envelope.Message.GetType();
```
alliscode:
```suggestion
                    return messageEnvelope.Message.GetType();
```

dotnet/src/Microsoft.Agents.AI.Workflows/Executor.cs
```diff
@@ -15,6 +16,130 @@
...
+        this._requestPorts.Add(port.Id, (port, delaySink));
+
+        return delaySink;
+    }
+}
+
+internal sealed class MessageTypeTranslator
+{
+    private readonly Dictionary<TypeId, Type> _typeLookupMap = [];
+    private readonly Dictionary<Type, TypeId> _delcaredTypeMap = [];
```
alliscode: nit: spelling

```suggestion
    private readonly Dictionary<Type, TypeId> _declaredTypeMap = [];
```

dotnet/src/Microsoft.Agents.AI.Workflows/Execution/MessageRouter.cs
```diff
@@ -49,15 +60,42 @@ internal MessageRouter(Dictionary<Type, MessageHandlerF> handlers, HashSet<Type>
...
+                    // Cache the handler for future lookups.
+                    this._typedHandlers[messageType] = handler;
+                    this._runtimeTypeMap[new TypeId(messageType)] = candidateType;
+                }
+
+                return handler;
+            }
+            else if (this._interfaceHandlers.Count > 0)
+            {
+                foreach (Type interfaceType in this._interfaceHandlers.Where(it => it.IsAssignableFrom(candidateType)))
```
alliscode, line 102: Is it a problem that this is not a deterministic iteration order? If an executor had two handlers that matched, then it could taggle back and forth randomly. Or maybe we don't allow that condition?

dotnet/src/Microsoft.Agents.AI.Workflows/FunctionExecutor.cs
```diff
@@ -70,6 +108,12 @@ ValueTask<TOutput> RunFuncAsync(TInput input, IWorkflowContext workflowContext,
         }
     }
 
+    /// <inheritdoc/>
+    protected override ProtocolBuilder ConfigureProtocol(ProtocolBuilder protocolBuilder)
+        => base.ConfigureProtocol(protocolBuilder)
+               .SendsMessageTypes(sentMessageTypes ?? [])
+               .YieldsOutputTypes(outputTypes ?? []);
```
westey-m: In `FunctionExecutor<TInput>` it also includes types from attributes.  Should it be doing the same here?

dotnet/src/Microsoft.Agents.AI.Workflows/Execution/MessageRouter.cs
```diff
@@ -49,15 +60,42 @@ internal MessageRouter(Dictionary<Type, MessageHandlerF> handlers, HashSet<Type>
...
+    public HashSet<Type> DefaultOutputTypes { get; }
 
-    public bool CanHandle(TypeId candidateType)
+    private MessageHandlerF? FindHandler(Type messageType)
     {
-        return this.HasCatchAll || this._runtimeTypeMap.ContainsKey(candidateType);
-    }
+        for (Type? candidateType = messageType; candidateType != null; candidateType = candidateType.BaseType)
+        {
+            if (this._typedHandlers.TryGetValue(candidateType, out MessageHandlerF? handler))
```
westey-m: Does this need to be thread safe?  Is it possible for this code block to execute multiple times in parallel?

dotnet/tests/Microsoft.Agents.AI.Workflows.UnitTests/Sample/01_Simple_Workflow_Sequential.cs
```diff
@@ -41,14 +41,26 @@ public static async ValueTask RunAsync(TextWriter writer, IWorkflowExecutionEnvi
...
         message.ToUpperInvariant();
 }
 
-internal sealed class ReverseTextExecutor() : ReflectingExecutor<ReverseTextExecutor>("ReverseTextExecutor", declareCrossRunShareable: true), IMessageHandler<string, string>
+//internal partial sealed class UppercaseExecutorEx() : Executor(nameof(UppercaseExecutorEx), declareCrossRunShareable: true)
+//{
+//    [MessageHandler(Send = [typeof(string)])]
+//    public async ValueTask<string> MyHandlerMethod(string message, IWorkflowContext context, CancellationToken cancellationToken = default) =>
+//        message.ToUpperInvariant();
+//}
```
westey-m: Can we delete the commented out code?
```suggestion
```

dotnet/src/Microsoft.Agents.AI.Workflows/Execution/FanInEdgeRunner.cs
lokitoth: Yes, very good catch.

dotnet/src/Microsoft.Agents.AI.Workflows/Execution/MessageRouter.cs
lokitoth: I would have expected the order to be stable once insertions to the object cease, but I suppose it could be doing clever things under the hood under memory pressure.

Since we only ever iterate it here, turning it into an array should not be an issue.

dotnet/src/Microsoft.Agents.AI.Workflows/Execution/MessageRouter.cs
lokitoth: Yes, when running in Concurrent-OffThread mode. Good catch.

dotnet/tests/Microsoft.Agents.AI.Workflows.UnitTests/Sample/01_Simple_Workflow_Sequential.cs
lokitoth: Whoops, that was from the API review discussion.

dotnet/src/Microsoft.Agents.AI.Workflows/FunctionExecutor.cs
lokitoth: Yes - and the reason that it cannot be done under the hood by the base `Executor{TIn[, TOut]}` is that the delegates get wrapped and the base cannot see the attributes on it.

PR #3802, opened 2026-02-10
ADR: Python context compaction strategy

eavanvalkenburg:
## Summary

Proposes a design for context compaction in the Python Agent Framework, extracted from the open discussion in [ADR-0016](docs/decisions/0016-python-context-middleware.md).

### Problem
Long-running agents with many tool calls accumulate unbounded message lists. The current architecture cannot compact messages during the tool loop — stores are read once, and middleware only modifies copies.

### Options Proposed
1. **Standalone `CompactionStrategy` object** — composed into `HistoryProvider` and `FunctionInvocationConfiguration`
2. **`CompactionStrategy` as a mixin** for `HistoryProvider` subclasses
3. **Separate `CompactionProvider`** as a `ContextProvider` subclass
4. **Mutable message access in `ChatMiddleware`** — refactor copy semantics

### Key Design Points
- Four compaction points: post-load, pre-write, in-run (tool loop), existing storage
- Not applicable to service-managed storage
- Atomic group preservation (tool calls + results)
- Leverages `source_id` attribution from ADR-0016
- `HistoryProvider.save_messages()` needs an overwrite mode for storage compaction

### References
- [ADR-0016: Unifying Context Management with ContextPlugin](docs/decisions/0016-python-context-middleware.md)
- [Context Compaction Limitations Analysis](https://gist.github.com/victordibia/ec3f3baf97345f7e47da025cf55b999f)


docs/decisions/00XX-python-context-compaction-strategy.md
```diff
@@ -0,0 +1,560 @@
...
+
+**All compaction discussed in this ADR is irrelevant when using only service-managed storage** (`service_session_id` is set). In that scenario:
+- The service manages message history internally — the client never holds the full conversation
+- Only new messages are sent to/from the service each turn
+- The service is responsible for its own context window management and compaction
+- The client has no message list to compact
+
+This ADR applies to two scenarios where the **client** constructs and manages the message list sent to the model:
+
+1. **With local storage** (e.g., `InMemoryHistoryProvider`, Redis, Cosmos) — compaction is needed at all four points (post-load, pre-write, in-run, existing storage)
```
markwallace-microsoft: Will our OOB memory providers always read the full chat history by default?

docs/decisions/00XX-python-context-compaction-strategy.md
```diff
@@ -0,0 +1,560 @@
...
+2. **Without any storage** (`store=False`, no `HistoryProvider`) — in-run compaction is still critical for long-running, tool-heavy agent invocations where the message list grows unbounded within a single `agent.run()` call
+
+### Compaction Strategies (Examples)
+
+A compaction strategy takes a list of messages and returns a (potentially shorter) list:
+
+- **Truncation**: Keep only the last N messages or N tokens
+- **Summarization**: Replace older messages with an LLM-generated summary
+- **Selective removal**: Remove tool call/result pairs while keeping user/assistant turns
+- **Sliding window with anchor**: Keep system message + last N messages
```
markwallace-microsoft: We **must** keep the system message in all cases

docs/decisions/00XX-python-context-compaction-strategy.md
```diff
@@ -0,0 +1,560 @@
...
+- **Token budget**: Remove oldest messages until under a token threshold
```
markwallace-microsoft: Is this the same as #1 when using N tokens

markwallace-microsoft: We also want to allow for custom compaction strategies.

docs/decisions/00XX-python-context-compaction-strategy.md
```diff
@@ -0,0 +1,560 @@
...
+
+1. **With local storage** (e.g., `InMemoryHistoryProvider`, Redis, Cosmos) — compaction is needed at all four points (post-load, pre-write, in-run, existing storage)
+2. **Without any storage** (`store=False`, no `HistoryProvider`) — in-run compaction is still critical for long-running, tool-heavy agent invocations where the message list grows unbounded within a single `agent.run()` call
+
+### Compaction Strategies (Examples)
+
+A compaction strategy takes a list of messages and returns a (potentially shorter) list:
+
+- **Truncation**: Keep only the last N messages or N tokens
+- **Summarization**: Replace older messages with an LLM-generated summary
```
markwallace-microsoft: In SK we didn't replace older messages instead

- Generate a summary of a range of memories
- Include more recent summary
- Insert the summary into the chat history in the appropriate place
- Return recent messages up to the most recent summary

We don't want to lose any data from chat history

docs/decisions/00XX-python-context-compaction-strategy.md
```diff
@@ -0,0 +1,560 @@
...
+
+This means strategies don't need to rely solely on message position or role — they can make semantically meaningful compaction decisions based on the origin of each message.
+
+### `HistoryProvider.save_messages` Is Append-Only
+
+ADR-0016's `HistoryProvider.save_messages()` is an **append** operation — `after_run` collects the new messages from the current invocation and appends them to storage. There is no built-in way to **replace** the full stored history with a compacted version.
+
+For compaction on existing storage (and pre-write compaction that rewrites history), we need a way to overwrite rather than append. Two options:
+
+1. **Add a `replace_messages()` method** to `HistoryProvider`:
```
markwallace-microsoft: Will this result in data loss?

docs/decisions/00XX-python-context-compaction-strategy.md
```diff
@@ -0,0 +1,560 @@
...
+    include_detailed_errors: bool
+    compaction_strategy: CompactionStrategy | None  # NEW
+```
+
+During the function calling loop, after each tool result is appended, the strategy is applied:
+
+```python
+# Inside the function invocation loop (e.g., in _try_execute_function_calls)
+messages.append(tool_result_message)
+if config.get("compaction_strategy"):
```
moonbox3: nit: I believe this code has a bug. Instead:

```python
compacted = await config["compaction_strategy"].compact(messages)
messages.clear()
messages.extend(compacted)
# or: messages[:] = compacted
```

The current code reassigns the local variable but doesn't mutate the list in-place. If the calling code holds a reference to the original list (which it does — prepped_messages in _tools.py), the reassignment has no effect.

docs/decisions/00XX-python-context-compaction-strategy.md
```diff
@@ -0,0 +1,560 @@
...
+        pre_write_compaction: CompactionStrategy | None = None,
+    ): ...
+
+    async def before_run(self, agent, session, context, state) -> None:
+        history = await self.get_messages(context.session_id)
+        if self.post_load_compaction:
+            history = await self.post_load_compaction.compact(history)
+        context.extend_messages(self.source_id, history)
+
+    async def after_run(self, agent, session, context, state) -> None:
```
moonbox3: `save_messages` is append-only, and `_collect_messages` only returns new messages from this run. So pre-write compaction only compacts the current turn's messages before appending, it never touches the existing history. Is that the intent? 

If you want to compact the full stored history on write, you'd need to read-compact-replace, which
requires the `replace_messages` extension discussed separately, if I am not mistaken.

Can we clarify whether "pre-write" means:
- Compact only the new messages before appending (current code), or
- Compact the full history (existing + new) and replace

These are different behaviors.

docs/decisions/00XX-python-context-compaction-strategy.md
```diff
@@ -0,0 +1,560 @@
...
+            if source == "ephemeral":
+                continue  # Remove ephemeral context
+            result.append(msg)
+        return result
+```
+
+### Open Questions
+
+1. **Naming**: Should we use `CompactionStrategy`, `ChatReducer` (for .NET alignment), or `ContextReducer`?
+2. **Trigger mechanism for in-run**: Should compaction run after every tool call, or only when a threshold is exceeded (e.g., token count, message count)?
```
moonbox3: I feel like this `trigger mechanism` open question deserves more than a mention here. It feels like one of the most important design decision for in-run compaction. Running `compact()` after every single tool call feels wasteful, because most runs won't need compaction, and summarization strategies involve an LLM call.

What if we make the trigger part of the strategy interface:

```python
class CompactionStrategy(ABC):
    def should_compact(self, messages: Sequence[Message]) -> bool:
        """Check if compaction is needed. Called after each tool result."""
        return True  # Default: always compact (subclasses override)

    @abstractmethod
    async def compact(self, messages: Sequence[Message]) -> list[Message]:
        ...
```
In this way, it keeps the trigger logic co-located with the strategy (a token-budget strategy knows its own budget) and avoids the LLM call overhead when compaction isn't needed. Thoughts?

docs/decisions/00XX-python-context-compaction-strategy.md
```diff
@@ -0,0 +1,560 @@
...
+- **Summarization**: Replace older messages with an LLM-generated summary
+- **Selective removal**: Remove tool call/result pairs while keeping user/assistant turns
+- **Sliding window with anchor**: Keep system message + last N messages
+- **Token budget**: Remove oldest messages until under a token threshold
+
+### Atomic Group Preservation
+
+A critical constraint for any compaction strategy: **tool calls and their results must be kept together**. LLM APIs (OpenAI, Azure, etc.) require that an assistant message containing `tool_calls` is always followed by corresponding `tool` result messages. A compaction strategy that removes one without the other will cause API errors.
+
+Strategies must treat `[assistant message with tool_calls] + [tool result messages]` as atomic groups — either keep the entire group or remove it entirely.
```
moonbox3: I get that [assistant with tool_calls] + [tool results] must be treated atomically, but then the doc leaves it to each strategy to implement correctly. Isn't this too error-prone? Every custom strategy would need to re-implement the same grouping logic.

What if we have a utility?

```python
def group_atomic_messages(messages: Sequence[Message]) -> list[list[Message]]:
    """Group messages into atomic units that must be kept or removed together."""
    ...
```

Or even better, make `compact()` operate on grouped messages by default, with a lower-level `compact_raw()` for strategies that need full control. This would prevent the most common class of bugs in custom strategies, I think.

docs/decisions/00XX-python-context-compaction-strategy.md
```diff
@@ -0,0 +1,560 @@
...
+                       Used for compaction workflows.
+        """
+        ...
+```
+
+Either approach enables the compaction-on-existing-storage workflow:
+
+```python
+history = await provider.get_messages(session_id)
+compacted = await strategy.compact(history)
```
moonbox3: What happens if `compact()` raises during the tool loop? I don't see this called out in the doc. For robustness, the loop should probably catch exceptions from compaction and continue with uncompacted messages and log a warning, rather than failing the entire agent.run(). Do you agree? A truncation strategy should never fail, but a summarization strategy that calls an LLM absolutely can.

docs/decisions/00XX-python-context-compaction-strategy.md
```diff
@@ -0,0 +1,560 @@
...
+            source = msg.additional_properties.get("source_id")
+            if source == "ephemeral":
+                continue  # Remove ephemeral context
+            result.append(msg)
+        return result
+```
+
+### Open Questions
+
+1. **Naming**: Should we use `CompactionStrategy`, `ChatReducer` (for .NET alignment), or `ContextReducer`?
```
moonbox3: I'd keep `CompactionStrategy`. I think it's more descriptive as to what is actually happening.

docs/decisions/00XX-python-context-compaction-strategy.md
```diff
@@ -0,0 +1,560 @@
...
+2. **Trigger mechanism for in-run**: Should compaction run after every tool call, or only when a threshold is exceeded (e.g., token count, message count)?
+3. **Chaining**: Should multiple strategies be chainable (e.g., summarize then truncate)?
```
moonbox3: I feel like this comes naturally based on Option 1's design. It should be built-in.

```python
class ChainedStrategy(CompactionStrategy):
    def __init__(self, *strategies: CompactionStrategy):
        self.strategies = strategies

    async def compact(self, messages):
        for strategy in self.strategies:
            messages = await strategy.compact(messages)
        return messages
```

docs/decisions/00XX-python-context-compaction-strategy.md
```diff
@@ -0,0 +1,560 @@
...
+
+- Standalone `CompactionStrategy` object composed into `HistoryProvider` and `FunctionInvocationConfiguration`
+- `CompactionStrategy` as a mixin for `HistoryProvider` subclasses
+- Separate `CompactionProvider` set directly on the agent
+- Mutable message access in `ChatMiddleware`
+
+
+## Pros and Cons of the Options
+
+### Option 1: Standalone `CompactionStrategy` Object
```
moonbox3: A thought in terms of observability: long-running agents with compaction will be hard to debug without visibility into when compaction happened and what was removed. Should we consider whether the strategy should return metadata (like CompactionResult with messages + removed_count + summary) or whether this should be handled via logging/events? Even a simple log line ("Compacted 150 messages to 50") could be helpful.

docs/decisions/00XX-python-context-compaction-strategy.md
eavanvalkenburg: right now, there is no filtering.

docs/decisions/00XX-python-context-compaction-strategy.md
eavanvalkenburg: yeah, it is only compacting the response messages, compacting what's already in the store is separate.

docs/decisions/00XX-python-context-compaction-strategy.md
eavanvalkenburg: makes sense

docs/decisions/00XX-python-context-compaction-strategy.md
eavanvalkenburg: you are right, but that's a implementation detail

docs/decisions/00XX-python-context-compaction-strategy.md
eavanvalkenburg: good one, will add

docs/decisions/00XX-python-context-compaction-strategy.md
eavanvalkenburg: I need to do some work there, I think it might make sense to actually separate the messages that are sent to the model (with compaction) split from the messages that are returned to the user/session, good logging should also be part indeed

docs/decisions/00XX-python-context-compaction-strategy.md
victordibia: +1 here. I think its critical that the strategy is aware of the trigger mechanism.
In my experience, all compaction strategies should be triggered on some token budget threshold (which can be tuned per use case ... too small results in data loss and repeat steps ... too large leaves efficiency on the table).

docs/decisions/00XX-python-context-compaction-strategy.md
victordibia: +1

docs/decisions/00XX-python-context-compaction-strategy.md
```diff
@@ -0,0 +1,560 @@
...
+        RAGContextProvider("rag"),               # Adds RAG context
+        SlidingWindowCompaction("compaction", max_messages=100),  # Compacts everything
+    ],
+)
+```
+
+The agent recognizes `CompactionProvider` instances and wires `compact()` into `FunctionInvocationConfiguration`:
+
+```python
+class ChatAgent:
```
victordibia: One more thing, I would like to add a clear example of a last N compaction design that shows a sawtooth pattern.
As task progresses, we count tokens, compaction kicks when a token budget is hit and then we see that total tokens drops. Ideally we should make this into a test to verify behaviour. 

Related, it might be useful to also consider if we want to emit compaction events and also support hooks (pre and post compaction)

docs/decisions/00XX-python-context-compaction-strategy.md
eavanvalkenburg: added a note on that implementation detail, there is a case to be made for both ways.

docs/decisions/00XX-python-context-compaction-strategy.md
eavanvalkenburg: clarified above

docs/decisions/00XX-python-context-compaction-strategy.md
eavanvalkenburg: My only issue with this, is that it potentially doubles the work, and if the work is expensive, like `tiktoken` invocation, then doing that twice (first to determine if we should compact, and then again to figure what to compact is a waste), we could ensure we return a tuple of dict that indicates if it did compact, so that we can log/emit a event accordingly.

PR #3761, opened 2026-02-09
.NET: [Breaking] Structured Output improvements

SergeyMenshykh:
### Motivation and Context
Currently, Structured Output (SO) in AF is supported by ChatClientAgent only and can be "enabled" either via the ResponseFormat property or the `RunAsync<T>` instance method. 

However, if ChatClientAgent is decorated by, let's say, an Open Telemetry decorator, the agent users lose access to the `RunAsync<T>` methods because they are instance methods of ChatClientAgent rather than members of the AIAgent class.  

Additionally, AF does not provide SO support for the other agents, i.e., A2A, which do not support SO natively.  

Furthermore, there's no way to provide SO configuration via the ResponseFormat property to any agent except ChatClientAgent at agent invocation because AgentRunOptions don't have the property and it only exists on ChatClientAgentRunOptions.  
  
### Description  
- Add the ResponseFormat property to the AgentRunOptions to communicate the agent's SO configuration.  
- Add `RunAsync<T>` non-virtual methods to the AIAgent base abstract class.  
- Add the StructuredOutputAgent decorator to add SO support to agents that don't natively support SO.
- Add the UseStructuredOutput extension method for the AIAgentBuilder, allowing easy registration of the decorator during the agent configuration step.  
- Restructure the SO sample to showcase different ways of working with SO in AF.

Closes: https://github.com/microsoft/agent-framework/issues/2874, https://github.com/microsoft/agent-framework/issues/1057, https://github.com/microsoft/agent-framework/issues/1642, https://github.com/microsoft/agent-framework/issues/2683, https://github.com/microsoft/agent-framework/issues/3231

PR #3803, opened 2026-02-10
.NET: Preserve caller context

SergeyMenshykh:
### Motivation and Context
The `RunAsync<T>` method in used ConfigureAwait(false), which schedules continuations on an arbitrary thread. This causes the orchestration to get stuck when the agent runs inside a durable orchestration (e.g., DurableAIAgent), whose dispatcher requires continuations to execute on the orchestration's main thread.

### Description
- Added durable-specific `RunAsync<T>` methods to DurableAIAgent (marked as new) with ConfigureAwait(true) to preserve the orchestration's synchronization context. All four overloads are provided.
- Extracted schema wrapping/unwrapping logic into a shared StructuredOutputSchemaUtilities class and made `AgentResponse<T>`.IsWrappedInObject publicly accessible.
- Added handling of primitives and arrays to DurableAIAgent while working with Structured Output.
- Removed unused constructor of `AgentResponse<T>`.

**More details:**
In this PR: https://github.com/microsoft/agent-framework/pull/3761, several durable agent samples that are run from integration tests are failing. These samples are stuck and never finish. Running them locally shows that the issue occurs in the samples related to durable orchestration scenarios with more than one agent. The first agent in the scenarios executes successfully, but [the call to the next one](https://github.com/microsoft/agent-framework/blob/977c3adfb2fa0d0b172201771316958efae74652/dotnet/samples/Durable/Agents/ConsoleApps/02_AgentOrchestration_Chaining/Program.cs#L56) never returns.

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.DurableTask/CHANGELOG.md
+++ b/dotnet/src/Microsoft.Agents.AI.DurableTask/CHANGELOG.md
@@ -11,7 +11,7 @@
 - Moved AgentSession.Serialize to AIAgent.SerializeSession ([#3650](https://github.com/microsoft/agent-framework/pull/3650))
 - Renamed serializedSession parameter to serializedState on DeserializeSessionAsync for consistency ([#3681](https://github.com/microsoft/agent-framework/pull/3681))
 - Introduce Core method pattern for Session management methods on AIAgent ([#3699](https://github.com/microsoft/agent-framework/pull/3699))
-- Refactored `RunAsync<T>` methods to use `AgentResponse<T>`, added missing overloads, and marked them as `new`. Added support for primitives and arrays ([#3803](https://github.com/microsoft/agent-framework/pull/3803))
+- Marked all `RunAsync<T>` overloads as `new`, added missing ones, and added support for primitives and arrays ([#3803](https://github.com/microsoft/agent-framework/pull/3803))
 
 ## v1.0.0-preview.251204.1
 
--- a/dotnet/src/Microsoft.Agents.AI.DurableTask/DurableAIAgent.cs
+++ b/dotnet/src/Microsoft.Agents.AI.DurableTask/DurableAIAgent.cs
@@ -5,6 +5,7 @@ using System.Text.Json;
 using Microsoft.DurableTask;
 using Microsoft.DurableTask.Entities;
 using Microsoft.Extensions.AI;
+using Microsoft.Shared.Diagnostics;
 
 namespace Microsoft.Agents.AI.DurableTask;
 
@@ -215,8 +216,12 @@ public sealed class DurableAIAgent : AIAgent
         AgentSession? session = null,
         JsonSerializerOptions? serializerOptions = null,
         AgentRunOptions? options = null,
-        CancellationToken cancellationToken = default) =>
-        this.RunAsync<T>(new ChatMessage(ChatRole.User, message), session, serializerOptions, options, cancellationToken);
+        CancellationToken cancellationToken = default)
+    {
+        _ = Throw.IfNull(message);
+
+        return this.RunAsync<T>(new ChatMessage(ChatRole.User, message), session, serializerOptions, options, cancellationToken);
+    }
 
     /// <summary>
     /// Runs the agent with a single chat message, requesting a response of the specified type <typeparamref name="T"/>.
@@ -240,8 +245,12 @@ public sealed class DurableAIAgent : AIAgent
         AgentSession? session = null,
         JsonSerializerOptions? serializerOptions = null,
         AgentRunOptions? options = null,
-        CancellationToken cancellationToken = default) =>
-        this.RunAsync<T>([message], session, serializerOptions, options, cancellationToken);
+        CancellationToken cancellationToken = default)
+    {
+        _ = Throw.IfNull(message);
+
+        return this.RunAsync<T>([message], session, serializerOptions, options, cancellationToken);
+    }
 
     /// <summary>
     /// Runs the agent with a collection of chat messages, requesting a response of the specified type <typeparamref name="T"/>.
@@ -275,12 +284,11 @@ public sealed class DurableAIAgent : AIAgent
         options = options?.Clone() ?? new DurableAgentRunOptions();
         options.ResponseFormat = responseFormat;
 
-        AgentResponse response = await this.RunAsync(messages, session, options, cancellationToken)
-            // ConfigureAwait(true) is intentional: the Durable Task Framework uses a custom
-            // synchronization context that requires all continuations to execute on the
-            // orchestration thread. Using ConfigureAwait(false) would schedule the continuation
-            // on an arbitrary thread, breaking the orchestration.
-            .ConfigureAwait(true);
+        // ConfigureAwait(false) cannot be used here because the Durable Task Framework uses
+        // a custom synchronization context that requires all continuations to execute on the
+        // orchestration thread. Scheduling the continuation on an arbitrary thread would break
+        // the orchestration.
+        AgentResponse response = await this.RunAsync(messages, session, options, cancellationToken);
 
         return new AgentResponse<T>(response, serializerOptions) { IsWrappedInObject = isWrappedInObject };
     }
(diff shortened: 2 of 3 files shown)


dotnet/src/Microsoft.Agents.AI.Abstractions/AIAgentStructuredOutput.cs
```diff
@@ -134,7 +134,13 @@ public async Task<AgentResponse<T>> RunAsync<T>(
...
         options.ResponseFormat = responseFormat;
 
-        AgentResponse response = await this.RunAsync(messages, session, options, cancellationToken).ConfigureAwait(false);
+        AgentResponse response = await this.RunAsync(messages, session, options, cancellationToken)
+            // ConfigureAwait(true) is intentional: the caller's synchronization context must be
+            // preserved because some agent implementations (e.g., DurableAIAgent) may run inside
+            // a durable orchestration whose dispatcher requires continuations to execute on the
+            // orchestration's main thread. Using ConfigureAwait(false) would schedule the
+            // continuation on an arbitrary thread, causing the orchestration to get stuck.
+            .ConfigureAwait(true);
```
SergeyMenshykh: @kshyju we're trying to understand how this change fixes the issue with agent orchestration. Could you help us understand how it works and why it's important to keep the synchronization context for agent implementations, like DurableAIAgent, when they're used in durable orchestrations?

stephentoub: And specifically why this particular ConfigureAwait(true) instead of false helps. The only thing it should impact is whether the remainder of this method is queued back to that synchronization context. Anyone awaiting this method without ConfigureAwait(false) will have their continuation queued back to the current synchronization context regardless of whether this uses ConfigureAwait(false) or ConfigureAwait(true).

cgillum: Hi @stephentoub, as you know, the Durable Task Framework has its own synchronization context when executing orchestration code, and that any `ConfigureAwait(false)` on the orchestration thread confuses the runtime and typically causes the orchestration to hang. As I understand it, The PR attempted to route the `DurableAIAgent` `RunAsync<T>` implementation through this new common code path, introducing the breaking behavior.

I can't speak for whether this is the *best* way to fix what is currently a DTFx-specific issue. Another option could be to simply not attempt to reuse this common code and have our own fork of it that does the right thing. But I suspect this could still be an issue for other users as well that are sensitive to whether a callback executes on a current synchronization context or a new one.

stephentoub: > and that any ConfigureAwait(false) on the orchestration thread confuses the runtime and typically causes the orchestration to hang.

I understand that when the work performed in the continuation is relevant, but here the work is just constructing a simple object and then completing.

Is it the case then that if I do:
```csharp
await Task.Run(() => { }); // queue a nop to the thread pool
```
in a durable flow, it might deadlock?

cgillum: One more thing that's worth mentioning: durable orchestrations will only ever use `DurableAIAgent`, which is specifically designed for running in these special synchronization contexts. Supporting other `AIAgent` implementations is out of scope for us, so usage of `ConfigureAwait(false)` in other agent types is non-impacting for us.

cgillum: > in a durable flow, it might deadlock?

Yes, can deadlock if there's any durable work that needs to happen after resuming from `await`, which is exactly what happened in the integration test. It's kind of like how WinForms requires all UI operations to happen on the UI thread - Durable requires all `context` operations to happen on the main orchestration thread.

stephentoub: > > in a durable flow, it might deadlock?
> 
> Yes, this will deadlock if there's any durable work that needs to happen after resuming from `await`.

Even though the `await` will see your SynchronizationContext.Current and post back to it to perform that work?

cgillum: > Even though the await will see your SynchronizationContext.Current and post back to it to perform that work?

I unfortunately don't know the mechanics of synchronization contexts well enough to answer that question directly. But what I can say is that any continuation that runs on a separate thread will break durable orchestrations. It doesn't really matter what work the task is actually doing.

SergeyMenshykh: @cgillum thanks for sharing the details. I've added `RunAsync<T>` methods with a few overloads to the DurableAIAgent  as "new" methods. These methods can only be invoked when the invocation happens via an instance of the DurableAIAgent, which is acceptable because it's the only agent that durable orchestration directly works with.

dotnet/src/Shared/StructuredOutput/StructuredOutputSchemaUtilities.cs
```diff
@@ -0,0 +1,104 @@
...
+        if (schemaElement.ValueKind is JsonValueKind.Object)
+        {
+            foreach (var property in schemaElement.EnumerateObject())
+            {
+                if (property.NameEquals("type"u8))
+                {
+                    return property.Value.ValueKind == JsonValueKind.String
+                        && property.Value.ValueEquals("object"u8);
+                }
+            }
```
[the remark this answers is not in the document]
SergeyMenshykh: We keep the code in sync with MEAI equivalent.

dotnet/src/Microsoft.Agents.AI.DurableTask/CHANGELOG.md
```diff
@@ -11,7 +11,7 @@
 - Moved AgentSession.Serialize to AIAgent.SerializeSession ([#3650](https://github.com/microsoft/agent-framework/pull/3650))
 - Renamed serializedSession parameter to serializedState on DeserializeSessionAsync for consistency ([#3681](https://github.com/microsoft/agent-framework/pull/3681))
 - Introduce Core method pattern for Session management methods on AIAgent ([#3699](https://github.com/microsoft/agent-framework/pull/3699))
-- Updated to use base `AgentRunOptions.ResponseFormat` for structured output configuration ([#3658](https://github.com/microsoft/agent-framework/pull/3658))
+- Refactored `RunAsync<T>` methods to use `AgentResponse<T>`, added missing overloads, and marked them as `new`. Added support for primitives and arrays ([#3803](https://github.com/microsoft/agent-framework/pull/3803))
```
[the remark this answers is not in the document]
SergeyMenshykh: There's no need in that.

dotnet/src/Microsoft.Agents.AI.DurableTask/DurableAIAgent.cs
```diff
@@ -168,4 +169,127 @@ protected override async IAsyncEnumerable<AgentResponseUpdate> RunCoreStreamingA
...
+        (responseFormat, bool isWrappedInObject) = StructuredOutputSchemaUtilities.WrapNonObjectSchema(responseFormat);
+
+        options = options?.Clone() ?? new DurableAgentRunOptions();
+        options.ResponseFormat = responseFormat;
+
+        // ConfigureAwait(false) cannot be used here because the Durable Task Framework uses
+        // a custom synchronization context that requires all continuations to execute on the
+        // orchestration thread. Scheduling the continuation on an arbitrary thread would break
+        // the orchestration.
+        AgentResponse response = await this.RunAsync(messages, session, options, cancellationToken);
```
westey-m, line 291: I assume only the last await matters? Would it work to just call base.RunAsync<T> with ConfigureAwait(true)?  That way you don't need to duplicate the structured output logic in here?

SergeyMenshykh: > I assume only the last await matters? Would it work to just call base.RunAsync with ConfigureAwait(true)? That way you don't need to duplicate the structured output logic in here?

That would allow us to avoid duplication, but it does not work. The original context should be preserved throughout the call chain down to the last call.

PR #3842, opened 2026-02-11
Python: Add request_handlers parameter for automatic HITL request handling

### LINKED ISSUE #3539: Python: Define fan-out behavior when node requests HITL info
## Summary

When a workflow fans out and one node requests info (HITL) while another node is in a loop, the behavior is undefined.

**Source:** Declarative RequestInfo notes

## Context

**Question:** Should we pause the entire workflow when any node requests info?

**Recommendation:** Pausing the entire workflow is the simplest solution.

## Action Items

- [ ] Define and document expected behavior
- [ ] Implement workflow-wide pause when any node requests info
- [ ] Consider if selective pausing is needed in the future

moonbox3:
### Motivation and Context

In fan-out workflows where one branch self-loops (async processing) and another emits request_info (HITL), callers previously had to manually collect requests, call handlers, and resubmit via run(responses=...). The new request_handlers parameter automates this with a simple dict-based API:
                    
```python                                                                                                                                                            
result = await workflow.run(                                         
    initial_data,
    request_handlers={
        ReviewRequest: handle_review,
        ApprovalRequest: handle_approval,
    },
)
```

Inline dispatch means the handler fires as soon as the request is emitted — it doesn't wait for other branches to complete first. A handler running concurrently with a long-running self-loop completes and injects its response while the loop is still iterating.

Behavior:
- Matched handler succeeds -> response submitted, request cleared -> IDLE
- Handler raises-> request stays pending -> IDLE_WITH_PENDING_REQUESTS
- No matching handler type -> warning logged, request stays pending -> IDLE_WITH_PENDING_REQUESTS
- No request_handlers provided -> existing behavior unchanged

> [!IMPORTANT]
> Compatibility with checkpointing and manual response submission: request_handlers is additive and does not change the existing HITL contract. Workflows that end as IDLE_WITH_PENDING_REQUESTS — whether because no handlers were registered, a handler failed, or a request type was unmatched — can still be resumed later via run(responses=...) 
or run(responses=..., checkpoint_id=...). Checkpointing continues to work: pending requests are captured in checkpoint state and can be restored and responded to in a separate process or session.

> [!NOTE]
> Relationship to `@response_handler`: The `@response_handler` decorator on executors is where responses get processed -- that is unchanged. `request_handlers` on `WorkflowBuilder()` or `run()` changes where the response comes from. Without it, the caller manually collects requests and resubmits via `run(responses=...)`. With it, handler functions source the response automatically and the executor's `@response_handler` runs the same as before. The two mechanisms work together, not in place of each other.

> [!NOTE]
> Type-based dispatch requires unique types: If two executors both do `ctx.request_info(request_data="some string", ...)`, there's only one slot for str in the dict -- both requests route to the same handler with no way to distinguish them. Unique dataclasses are effectively required for any non-trivial workflow:

```python
@dataclass
class ReviewRequest:
    content: str

@dataclass
class ApprovalRequest:
    task: str
```

This is the same pattern as executor input types: `str` works for toy examples but real workflows need typed messages. The `request_handlers` dispatch just makes that constraint explicit.

### Description

- Adds a request_handlers parameter to workflow.run() that maps request data types to async handler functions for automatic HITL request handling
- Handlers are dispatched inline as asyncio.create_task when request_info events are emitted, running concurrently with other workflow branches
- The runner's convergence check waits for outstanding handler tasks before declaring idle, so handler responses are processed in subsequent supersteps within the same run
- Works with both stream=True and stream=False
- Closes #3539

### Contribution Checklist


markwallace-microsoft:
<a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/README.md">Python Test Coverage</a><details><summary>Python Test Coverage Report • </summary><table><tr><th>File</th><th>Stmts</th><th>Miss</th><th>Cover</th><th>Missing</th></tr><tbody><tr><td colspan="5"><b>packages/core/agent_framework/_workflows</b></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_runner.py">_runner.py</a></td><td>223</td><td>6</td><td>97%</td><td><a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_runner.py#L142-L145">142&ndash;145</a>, <a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_runner.py#L406-L407">406&ndash;407</a></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_runner_context.py">_runner_context.py</a></td><td>153</td><td>10</td><td>93%</td><td><a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_runner_context.py#L63">63</a>, <a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_runner_context.py#L77-L78">77&ndash;78</a>, <a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_runner_context.py#L80-L81">80&ndash;81</a>, <a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_runner_context.py#L83">83</a>, <a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_runner_context.py#L369">369</a>, <a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_runner_context.py#L388">388</a>, <a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_runner_context.py#L441">441</a>, <a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_runner_context.py#L454">454</a></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_workflow.py">_workflow.py</a></td><td>271</td><td>19</td><td>92%</td><td><a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_workflow.py#L85">85</a>, <a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_workflow.py#L272-L274">272&ndash;274</a>, <a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_workflow.py#L276-L277">276&ndash;277</a>, <a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_workflow.py#L295">295</a>, <a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_workflow.py#L299">299</a>, <a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_workflow.py#L438">438</a>, <a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_workflow.py#L654">654</a>, <a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_workflow.py#L677">677</a>, <a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_workflow.py#L739">739</a>, <a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_workflow.py#L751">751</a>, <a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_workflow.py#L757">757</a>, <a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_workflow.py#L762">762</a>, <a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_workflow.py#L782-L784">782&ndash;784</a>, <a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_workflow.py#L797">797</a></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_workflow_builder.py">_workflow_builder.py</a></td><td>156</td><td>11</td><td>92%</td><td><a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_workflow_builder.py#L183">183</a>, <a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_workflow_builder.py#L460-L464">460&ndash;464</a>, <a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_workflow_builder.py#L466">466</a>, <a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_workflow_builder.py#L569">569</a>, <a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_workflow_builder.py#L584">584</a>, <a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_workflow_builder.py#L608">608</a>, <a href="https://github.com/microsoft/agent-framework/blob/29e29a111a7748a8358918c6baee121e7771c5eb/packages/core/agent_framework/_workflows/_workflow_builder.py#L619">619</a></td></tr><tr><td><b>TOTAL</b></td><td><b>27998</b></td><td><b>3428</b></td><td><b>87%</b></td><td>&nbsp;</td></tr></tbody></table></details>

## Python Unit Test Overview
| 5411 | 20 :zzz: | 0 :x: | 0 :fire: | 1m 24s :stopwatch: |

python/packages/core/agent_framework/_workflows/_workflow.py
```diff
@@ -476,6 +479,7 @@ def run(
         *,
         stream: bool = False,
         responses: dict[str, Any] | None = None,
+        response_handlers: Mapping[type, Callable[[Any], Awaitable[Any]]] | None = None,
```
eavanvalkenburg: How will users know which keys exist? Could we use a TypedDict for that?

moonbox3: The keys map to the request data types that executors emit via `ctx.request_info()`: the same types the caller already uses when wiring `run(responses=...)` manually. It's a similar discovery pattern to knowing executor input/output types when building edges. A `TypedDict` wouldn't work here since the keys are runtime type objects, not string literals. We could explore a `workflow.request_info_types` introspection property down the road if discoverability becomes a real friction point.

eavanvalkenburg (approved):

python/packages/core/agent_framework/_workflows/_workflow.py
```diff
@@ -557,15 +618,45 @@ async def _run_core(
...
+                        logger.warning(
+                            f"No response handler registered for request type "
+                            f"{request_data_type.__name__} (request_id={event.request_id}). "
+                            f"Registered types: {registered_types}"
+                        )
+
+                yield event
+        finally:
+            # Cancel any outstanding handler tasks on error/cancellation
+            if outstanding_tasks:
```
TaoChenOSU: What is going to happen if the workflow goes to idle before a handle returns?

python/packages/core/agent_framework/_workflows/_runner.py
```diff
@@ -151,6 +160,14 @@ async def run_until_convergence(self) -> AsyncGenerator[WorkflowEvent, None]:
 
                 # Check for convergence: no more messages to process
                 if not await self._ctx.has_messages():
+                    if outstanding_handler_tasks:
+                        still_running = {t for t in outstanding_handler_tasks if not t.done()}
+                        while still_running:
+                            _, still_running = await asyncio.wait(still_running, return_when=asyncio.FIRST_COMPLETED)
```
TaoChenOSU: Should there be a timeout?

python/packages/core/agent_framework/_workflows/_workflow.py
```diff
@@ -557,15 +618,45 @@ async def _run_core(
...
-            reset_context=reset_context,
-            streaming=streaming,
-            run_kwargs=kwargs if kwargs else None,
-        ):
-            if event.type == "output" and not self._should_yield_output_event(event):
-                continue
-            yield event
+        outstanding_tasks: set[asyncio.Task[None]] | None = set() if request_handlers else None
+
+        try:
```
TaoChenOSU: Is it possible to internalize the logic to the runner? Spreading it across two components seems harder to maintain.

The runner also has access to events to be emitted. If it sees a request info event that has a handler for, it can just create a task and manage inside the runner.

python/packages/core/agent_framework/_workflows/_runner.py
```diff
@@ -151,6 +160,14 @@ async def run_until_convergence(self) -> AsyncGenerator[WorkflowEvent, None]:
...
+                            if await self._ctx.has_messages():
```
TaoChenOSU: nit: a brief comment may help for the future.

I am thinking what if the other branch stop producing messages and the response handler never returns, will this become an infinite loop? Is this a valid scenario?

python/packages/core/agent_framework/_workflows/_workflow.py
```diff
@@ -193,6 +194,8 @@ def __init__(
             description: Optional description of what the workflow does.
             output_executors: Optional list of executor IDs whose outputs will be considered workflow outputs.
                               If None or empty, all executor outputs are treated as workflow outputs.
+            request_handlers: Optional default response handlers for automatic HITL request handling.
```
TaoChenOSU:
```suggestion
            request_handlers: Optional default request handlers for automatic HITL request handling.
```

I think the name will cause a lot of confusion but yet I can't come up with a better name. Can bring this up to the team.

python/packages/core/agent_framework/_workflows/_workflow.py
```diff
@@ -297,6 +303,7 @@ async def _run_workflow_with_tracing(
         reset_context: bool = True,
         streaming: bool = False,
         run_kwargs: dict[str, Any] | None = None,
+        outstanding_handler_tasks: set[asyncio.Task[None]] | None = None,
```
TaoChenOSU: This is not public but may confuse readers of this code.

python/packages/core/agent_framework/_workflows/_workflow.py
moonbox3: Similar to what happens in dotnet: the workflow stays alive and waits. In the current code, with `request_handlers` the runner intentionally waits for the handler tasks before converging. A slow handler will keep the run open.

For "days/months" HITL, the safer path is:
- don't rely on these in-run request handlers
- let the workflow emit request_info events which will end the workflow (if it converges) with IDLE_WITH_PENDING_REQUESTS.
- persist via checkpoint
- later resume with `run(responses=...)`

If we want to support both short-running and long-running behaviors we can:
- request_handler_wait_timeout or wait_for_request_handlers=False
- have a brief wait
- then finalize as IDLE_WITH_PENDING_REQUESTS if things are still pending

python/packages/core/agent_framework/_workflows/_workflow.py
moonbox3: Sure, I think that's valid feedback, thanks. Will do this.

python/packages/core/agent_framework/_workflows/_workflow.py
moonbox3: I agree. These are "requests" that need to be fulfilled, as opposed to the `@response_handler` that we currently have, which is why I went with `request_handlers`.

python/packages/core/agent_framework/_workflows/_runner.py
moonbox3: Added a comment.

To your question: similar to another answer, current behavior is intentional for "inline auto-handle in same run." For very long-latency HITL (days/months), request_handlers is not ideal; use pending requests + later `run(responses=...)` resume instead.

python/packages/core/agent_framework/_workflows/_workflow.py
moonbox3: This is removed after the refactor from workflow -> runner.

python/packages/core/agent_framework/_workflows/_workflow.py
TaoChenOSU: My understanding of this if statement is that if there are pending task, we will cancel them. Please correct me if I am wrong.

moonbox3: This is teardown safety for abnormal exits (error, cancel, stopped consumption, etc.)

TaoChenOSU:
A background run can address this issue naturally: https://github.com/microsoft/agent-framework/pull/4274. Please take a look and we can discuss.

python/packages/core/agent_framework/_workflows/_workflow.py
moonbox3: Fixed in `29e29a11`.

python/packages/core/agent_framework/_workflows/_workflow.py
moonbox3: Fixed in `29e29a11`.

moonbox3:
Closing for now - will re-open later.

PR #3981, opened 2026-02-17
.NET: bug fix for duplicate output on GitHubCopilotAgent

normalian:
### Motivation and Context

When using GitHubCopilotAgent, the final assistant message output was duplicated. The AssistantMessageDeltaEvent handlers already stream text content incrementally via TextContent, but the AssistantMessageEvent handler was also emitting a TextContent with the full message. This caused downstream consumers to render the same text twice.

### Description

Changed the ConvertToAgentResponseUpdate(AssistantMessageEvent) method in GitHubCopilotAgent.cs to emit a plain AIContent instead of a TextContent. The AssistantMessageEvent is a completion signal that carries metadata (message ID, timestamp, raw representation) — it should not re-emit the text content that was already streamed through AssistantMessageDeltaEvent. By using AIContent instead of TextContent, the event still preserves its RawRepresentation for inspection without producing duplicate text output.

### Contribution Checklist

 The code builds clean without any errors or warnings
 The PR follows the Contribution Guidelines
 All unit tests pass, and I have added new tests where possible

### DIFF
--- a/dotnet/tests/Microsoft.Agents.AI.GitHub.Copilot.UnitTests/GitHubCopilotAgentTests.cs
+++ b/dotnet/tests/Microsoft.Agents.AI.GitHub.Copilot.UnitTests/GitHubCopilotAgentTests.cs
@@ -2,7 +2,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Reflection;
 using System.Threading.Tasks;
 using GitHub.Copilot.SDK;
 using Microsoft.Extensions.AI;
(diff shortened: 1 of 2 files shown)


markwallace-microsoft:
Copilot is handing this issue https://github.com/microsoft/semantic-kernel/issues/13570 to fix the PR build

dmytrostruk:
@normalian Thanks for your contribution, wondering if you would be able to add some tests to verify this behavior? Please let me know, thanks!

normalian:
HI @dmytrostruk, thanks for your reply. I just created the test code and push here. Please review it.

dotnet/tests/Microsoft.Agents.AI.GitHub.Copilot.UnitTests/GitHubCopilotAgentTests.cs
```diff
@@ -221,4 +222,33 @@ public void CopyResumeSessionConfig_WithNullSource_ReturnsDefaults()
...
+            {
+                MessageId = "msg-456",
+                Content = "Some streamed content that was already delivered via delta events"
+            }
+        };
+        CopilotClient copilotClient = new(new CopilotClientOptions { AutoStart = false });
+        const string TestId = "agent-id";
+        var agent = new GitHubCopilotAgent(copilotClient, ownsClient: false, id: TestId, tools: null);
+        MethodInfo method = typeof(GitHubCopilotAgent).GetMethod(
+            "ConvertToAgentResponseUpdate",
```
dmytrostruk: @normalian We can make `ConvertToAgentResponseUpdate` method `internal` instead of `private` and it should be accessible in this test, thank you!

normalian:
Thanks for your review, @dmytrostruk. I just update the source code files by following your suggestion. Please check again.

dmytrostruk (approved):

PR #3989, opened 2026-02-17
.NET: Fix FoundryAgents_Step15_ComputerUse sample for Azure Agents API

rogerbarreto:
### Motivation and Context

The Azure Agents API rejects previous_response_id alongside computer_call_output items, unlike the vanilla OpenAI Responses API. This fix:

- Send all prior response output items (reasoning, computer_call, etc.) as input items in follow-up calls so the API has full conversation context
- Create a fresh session per call to avoid ConversationId/previous_response_id
- Use currentCallId instead of initialCallId for computer_call_output
- Clear ContinuationToken after polling to prevent stale tokens
- Remove unused initialCallId tracking variable

### DIFF
--- a/dotnet/samples/GettingStarted/FoundryAgents/FoundryAgents_Step15_ComputerUse/Program.cs
+++ b/dotnet/samples/GettingStarted/FoundryAgents/FoundryAgents_Step15_ComputerUse/Program.cs
@@ -94,8 +94,10 @@ internal sealed class Program
         // Initial request with screenshot - start with Bing search page
         Console.WriteLine("Starting computer automation session (initial screenshot: cua_browser_search.png)...");
 
-        // Computer-use requires a fresh session per call to avoid previous_response_id being set.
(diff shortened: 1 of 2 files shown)


dotnet/samples/GettingStarted/FoundryAgents/FoundryAgents_Step15_ComputerUse/Program.cs
```diff
@@ -96,14 +94,16 @@ private static async Task InvokeComputerUseAgentAsync(AIAgent agent)
         // Initial request with screenshot - start with Bing search page
         Console.WriteLine("Starting computer automation session (initial screenshot: cua_browser_search.png)...");
 
+        // Computer-use requires a fresh session per call to avoid previous_response_id being set.
```
westey-m: >to avoid previous_response_id being set

Why would this be bad?
Should we explain further here?

westey-m (approved):

PR #3935, opened 2026-02-13
.NET: [Feature Branch] Add Azure Functions hosting support for durable workflows

kshyju:
### Motivation and Context

This PR adds Azure Functions hosting support for durable workflows, building on the foundational workflow APIs introduced in #3648. While #3648 added console/hosted-service support for running workflows as Durable Task orchestrations, this PR enables workflows to be hosted as Azure Functions with automatic trigger registration, removing the need for manual function definitions.

> **Note:** This PR targets the `feat/durable_task` feature branch.

**Key scenarios enabled:**
- Hosting durable workflows as Azure Functions with zero boilerplate trigger definitions
- Automatic HTTP trigger registration for starting workflow orchestrations via HTTP
- Automatic activity trigger registration for workflow executors
- Automatic entity trigger registration for AI agent executors within workflows
- Deduplication of shared executors across multiple workflows

### Description

This PR introduces Azure Functions hosting infrastructure for durable workflows, along with new public APIs and shared utilities.

#### New Public APIs

Two new extension methods are added to `FunctionsApplicationBuilderExtensions` in the `Microsoft.Agents.AI.Hosting.AzureFunctions` namespace. These complement the existing `ConfigureDurableAgents` method.

| API | Description |
|-----|-------------|
| `ConfigureDurableWorkflows` | Configures the Azure Functions app to host durable workflows. Automatically registers orchestrations, generates function triggers (HTTP, activity, entity) for each workflow and its executors, and sets up the middleware pipeline. This is the recommended entry point for workflow-only applications. |
| `ConfigureDurableOptions` | Configures the Azure Functions app with full durable options, including both agents and workflows. Use this when your application needs to host both durable agents and durable workflows in the same function app. |

```csharp
public static FunctionsApplicationBuilder ConfigureDurableWorkflows(
    this FunctionsApplicationBuilder builder,
    Action<DurableWorkflowOptions> configure);

public static FunctionsApplicationBuilder ConfigureDurableOptions(
    this FunctionsApplicationBuilder builder,
    Action<DurableOptions> configure);
```

> **Note:** All other new types added in this PR (`FunctionMetadataFactory`, `DurableWorkflowsFunctionMetadataTransformer`, `BuiltInFunctionExecutor`, etc.) are `internal` and not part of the public API surface.

#### Usage Examples

**Sequential Workflow hosted in Azure Functions (Sample: `01_SequentialWorkflow`)**

```csharp
OrderLookup orderLookup = new();
OrderCancel orderCancel = new();
[...]

PR #3648, opened 2026-02-03
.NET: [Feature Branch] Add basic durable workflow support

kshyju:
### Motivation and Context

This PR adds durable workflow support to the Agent Framework, enabling workflows to run as Durable Task orchestrations. This addresses the need for long-running, reliable workflow execution that can survive process restarts, handle failures gracefully, and supports distributed workflow execution across multiple nodes.

> **Note:** This is the first in a series of PRs for durable workflow support. This PR establishes the foundational APIs and infrastructure. Follow-up PRs will add additional features such as events, yield output, human-in-the-loop patterns, and sub-workflow support.

**Key scenarios enabled:**
- Running workflows as durable orchestrations with automatic state persistence
- Sequential and concurrent workflow execution patterns
- Fan-out/fan-in workflow patterns
- Conditional routing based on runtime conditions
- Integration with existing Durable Task infrastructure

### Description

This PR introduces a new set of public APIs for running workflows as Durable Task orchestrations.

#### New Public APIs

**Workflow Execution Interfaces**

| API | Description |
|-----|-------------|
| `IWorkflowRun` | Represents a running workflow instance with event tracking |
| `IAwaitableWorkflowRun` | Extends `IWorkflowRun` with completion awaiting capability |
| `IWorkflowClient` | Client interface for starting and managing workflow executions |

```csharp
public interface IWorkflowRun
{
    string RunId { get; }
    IEnumerable<WorkflowEvent> OutgoingEvents { get; }
    IEnumerable<WorkflowEvent> NewEvents { get; }
    int NewEventCount { get; }
}

public interface IAwaitableWorkflowRun : IWorkflowRun
{
    ValueTask<TResult?> WaitForCompletionAsync<TResult>(CancellationToken cancellationToken = default);
}

public interface IWorkflowClient
{
    ValueTask<IWorkflowRun> RunAsync<TInput>(Workflow workflow, TInput input, string? runId = null, CancellationToken cancellationToken = default) where TInput : notnull;
    ValueTask<IWorkflowRun> RunAsync(Workflow workflow, string input, string? runId = null, CancellationToken cancellationToken = default);
}
```

**Configuration Options**

| API | Description |
|-----|-------------|
| `DurableOptions` | Root configuration container for durable agents and workflows |
| `DurableWorkflowOptions` | Configuration for registering and managing durable workflows |

```csharp
public sealed class DurableOptions
{
    public DurableAgentsOptions Agents { get; }
    public DurableWorkflowOptions Workflows { get; }
}

[...]

PR #4020, opened 2026-02-17
.NET: [Feature Branch] Adding support for events & shared state in durable workflows

kshyju:
### Motivation and Context

This PR adds workflow events and shared state support for durable workflows, building on the foundational workflow APIs introduced in #3648 and the Azure Functions hosting support added in #3935. While previous PRs established the core execution model, this PR enables real-time observability and cross-executor data sharing — two capabilities essential for production workflow scenarios.

> **Note:** This PR targets the `feat/durable_task` feature branch.

**Key scenarios enabled:**
- Emitting custom domain events from executors and streaming them to callers in real-time
- Sharing state across executors without passing data through the message chain (as return value) 
- Partitioning state into isolated scopes and clearing them when no longer needed
- Early workflow termination.
- Observing workflow progress via streaming (executor invocations, completions, outputs, terminal events)

### Description

This PR introduces event streaming and shared state infrastructure for durable workflows, along with new public APIs and supporting types.

#### New Public APIs

| API | Description |
|-----|-------------|
| `IWorkflowClient.StreamAsync<TInput>` | Starts a workflow and returns an `IStreamingWorkflowRun` for real-time event observation. |
| `IStreamingWorkflowRun` | Interface with `RunId` and `WatchStreamAsync()` that yields `WorkflowEvent` instances as they occur. |
| `DurableWorkflowCompletedEvent` | Terminal event emitted when the workflow completes successfully. Contains the final result. |
| `DurableWorkflowFailedEvent` | Terminal event emitted when the workflow fails. Contains the error message. |
| `DurableHaltRequestedEvent` | Event indicating early workflow halt (e.g., validation failure via `RequestHaltAsync`). |

```csharp
// Start a workflow with streaming
IStreamingWorkflowRun run = await workflowClient.StreamAsync(workflow, "ORDER-123");

// Observe events in real-time
await foreach (WorkflowEvent evt in run.WatchStreamAsync())
{
    switch (evt)
    {
        case WorkflowOutputEvent e:
            Console.WriteLine($"[Output] {e.SourceId}: {e.Data}");
            break;
        case DurableWorkflowCompletedEvent e:
            Console.WriteLine($"Completed: {e.Result}");
            break;
    }
}
```

> **Note:** All other new types added in this PR (`DurableStreamingWorkflowRun`, `DurableWorkflowCustomStatus`, `DurableWorkflowResult`, `TypedPayload`, `DurableSerialization`, `DurableRunStatus`) are `internal` and not part of the public API surface.

#### Shared State APIs (via `IWorkflowContext`)

Shared state is implemented through the existing `IWorkflowContext` interface, which executors already receive in `HandleAsync`. No new public types are needed — the following methods are now functional in the durable context:

| Method | Description |
|--------|-------------|
| `QueueStateUpdateAsync` | Write a value to shared state for downstream executors |
| `ReadStateAsync<T>` | Read a value written by a previous executor |
| `ReadOrInitStateAsync<T>` | Read or lazily initialize a state value |
| `QueueClearScopeAsync` | Clear all state entries under a scope |
| `RequestHaltAsync` | Stop the workflow early (no downstream executors run) |
| `AddEventAsync` | Emit a custom domain event observable via streaming |
| `YieldOutputAsync` | Emit an intermediate output as a `WorkflowOutputEvent` |

```csharp
// In ValidateOrder executor: write tax rate to shared state
await context.QueueStateUpdateAsync("taxRate", 0.085m);

// In ProcessPayment executor: read tax rate from shared state
decimal taxRate = await context.ReadOrInitStateAsync("taxRate", () => 0.0m);
decimal total = orderAmount + (orderAmount * taxRate);

// Use custom scopes to partition state
await context.QueueStateUpdateAsync("carrier", "Contoso Express", scopeName: "shipping");

// Clear all entries under a scope when no longer needed
await context.QueueClearScopeAsync("shipping");
```

#### Internal Infrastructure Changes

| Component | Description |
|-----------|-------------|
| `DurableActivityContext` | Full implementation of `IWorkflowContext` — state read/write, events, halt requests, and scoped state management |
| `DurableWorkflowRunner` | Orchestration carries shared state across supersteps, merges state updates from activities, publishes accumulated events via custom status, and handles halt requests |
| `DurableStreamingWorkflowRun` | Polls orchestration custom status for new events with configurable polling interval (default: 100ms) |
| `DurableActivityExecutor` | Passes shared state to activity context and returns state updates, events, and halt status in activity output |
| `DurableExecutorDispatcher` | Updated to pass shared state through to activity execution |
| `DurableWorkflowJsonContext` | Updated with new types for AOT-compatible serialization |

#### How Event Streaming Works

The orchestration accumulates events from activity outputs and publishes them via `SetCustomStatus` after each superstep. `DurableStreamingWorkflowRun` polls the orchestration metadata at a configurable interval, deserializes new events, and yields them through `WatchStreamAsync`. On completion, events are also stored in the orchestration output as a fallback (since `SerializedCustomStatus` is cleared by the framework on completion).

#### How Shared State Works

Shared state is a `Dictionary<string, string>` maintained by the orchestration. Before each activity dispatch, the current state snapshot is passed to the activity. Activities collect state updates locally and return them as part of their output. The orchestration merges updates after each superstep, making them available to the next round of executors. State keys can be partitioned into scopes (e.g., `scopeName: "shipping"`) so that different parts of a workflow can manage their state independently.

### Validation/Testing

**Samples Added:**

| Sample | Description |
|--------|-------------|
| `05_WorkflowEvents` | Demonstrates custom event emission (`AddEventAsync`) and real-time streaming (`StreamAsync`/`WatchStreamAsync`) with an order cancellation workflow. No Azure OpenAI required. |
| `07_WorkflowSharedState` | Demonstrates shared state across executors (tax rate, audit trail, scoped shipping data) in an order processing pipeline with `QueueStateUpdateAsync`, `ReadStateAsync`, `ReadOrInitStateAsync`, scoped writes, and `QueueClearScopeAsync`. No Azure OpenAI required. |

**Integration Tests:** Integration tests have been added (`WorkflowConsoleAppSamplesValidation.cs`) to validate both samples — `WorkflowEventsSampleValidationAsync` and `WorkflowSharedStateSampleValidationAsync` — ensuring end-to-end functionality works as expected.

Both samples have been manually verified against the local DTS emulator.

### Contribution Checklist


dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableActivityContext.cs
```diff
@@ -64,44 +101,193 @@ public ValueTask SendMessageAsync(
...
         Func<T> initialStateFactory,
         string? scopeName = null,
-        CancellationToken cancellationToken = default) => default;
+        CancellationToken cancellationToken = default)
+    {
+        T? value = await this.ReadStateAsync<T>(key, scopeName, cancellationToken).ConfigureAwait(false);
+
+        if (value is not null)
+        {
+            return value;
```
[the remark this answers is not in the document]
kshyju: Fixed by introducing a HasStateKey(key, scopeName) check to test key existence directly instead of relying on the null check. Added unit tests covering this scenario

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableStreamingWorkflowRun.cs
```diff
@@ -0,0 +1,359 @@
...
+
+        // Track how many events we've already read from custom status
+        int lastReadEventIndex = 0;
+
+        while (!cancellationToken.IsCancellationRequested)
+        {
+            OrchestrationMetadata? metadata = await this._client.GetInstanceAsync(
+                this.RunId,
+                getInputsAndOutputs: true,
+                cancellation: cancellationToken).ConfigureAwait(false);
```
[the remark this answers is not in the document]
kshyju: SerializedCustomStatus is only populated when getInputsAndOutputs: true. Since custom status is the mechanism used for streaming events during execution, polling  with false would break event streaming entirely (no events would ever be yielded while the workflow is running). Added an inline comment explaining this constraint.

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableStreamingWorkflowRun.cs
```diff
@@ -0,0 +1,359 @@
...
+                yield break;
+            }
+
+            if (metadata.RuntimeStatus == OrchestrationRuntimeStatus.Terminated)
+            {
+                yield return new DurableWorkflowFailedEvent("Workflow was terminated.");
+                yield break;
+            }
+
+            await Task.Delay(interval, cancellationToken).ConfigureAwait(false);
```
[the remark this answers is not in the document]
kshyju: Wrapped the Task.Delay in a try/catch (OperationCanceledException) when cancellationToken.IsCancellationRequested) → yield break so the stream completes cleanly instead of
  faulting. Added a unit test (WatchStreamAsync_Cancellation_EndsGracefullyAsync) that validates this behavior.

dotnet/src/Microsoft.Agents.AI.DurableTask/ServiceCollectionExtensions.cs
```diff
@@ -330,6 +327,15 @@ private static WorkflowRegistrationInfo BuildWorkflowRegistration(
         return new WorkflowRegistrationInfo(orchestrationName, activities);
     }
 
+    /// <summary>
+    /// Returns <see langword="true"/> for bindings that should be registered as Durable Task activities.
+    /// <see cref="AIAgentBinding"/> (Durable Entities) and <see cref="SubworkflowBinding"/> (sub-orchestrations)
+    /// use specialized dispatch and are excluded.
+    /// </summary>
```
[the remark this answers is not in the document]
kshyju: Changelog will be added when all the changes are merged to feature branch and we are ready to take it to main.

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableStreamingWorkflowRun.cs
```diff
@@ -0,0 +1,411 @@
...
+            // Not a JSON-encoded string; try direct deserialization below.
+        }
+
+        if (typeof(TResult) == typeof(string))
+        {
+            return (TResult)(object)serializedOutput;
+        }
+
+        return JsonSerializer.Deserialize<TResult>(serializedOutput);
+    }
```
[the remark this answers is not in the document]
kshyju: Fixed to use the shared Serializer option.

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableWorkflowRunner.cs
```diff
@@ -171,25 +171,43 @@ private static async Task<string> RunSuperstepLoopAsync(
...
 
-            string[] results = await DispatchExecutorsInParallelAsync(context, executorInputs, logger).ConfigureAwait(true);
+            string[] results = await DispatchExecutorsInParallelAsync(context, executorInputs, state.SharedState, logger).ConfigureAwait(true);
 
-            ProcessSuperstepResults(executorInputs, results, state, logger);
+            bool haltRequested = ProcessSuperstepResults(executorInputs, results, state, context, logger);
+
+            if (haltRequested)
+            {
+                logger.LogWorkflowCompleted();
```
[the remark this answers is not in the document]
kshyju: Removed the log entry from the loop.

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableStreamingWorkflowRun.cs
```diff
@@ -0,0 +1,411 @@
...
+                    foreach (WorkflowEvent evt in events)
+                    {
+                        hasNewEvents = true;
+                        yield return evt;
+                    }
+                }
+            }
+
+            // On terminal status, re-fetch with outputs to get the final result.
+            // Check terminal states after draining events from custom status
```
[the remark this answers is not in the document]
kshyju: fixed the comment.

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableWorkflowContext.cs
```diff
@@ -64,44 +101,193 @@ public ValueTask SendMessageAsync(
...
+        return default;
+    }
 
     /// <inheritdoc/>
     public ValueTask<T?> ReadStateAsync<T>(
         string key,
         string? scopeName = null,
-        CancellationToken cancellationToken = default) => default;
+        CancellationToken cancellationToken = default)
+    {
```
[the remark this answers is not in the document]
kshyju: Fixed.

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableStreamingWorkflowRun.cs
```diff
@@ -0,0 +1,359 @@
...
+        {
+            return default;
+        }
+
+        if (typeof(TResult) == typeof(string))
+        {
+            return (TResult)(object)resultJson;
+        }
+
+        return JsonSerializer.Deserialize<TResult>(resultJson);
```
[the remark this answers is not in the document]
kshyju: ExtractResult now falls back to deserializing serializedOutput directly as TResult when the DurableWorkflowResult wrapper is absent. It handles DurableTask's string-quoting by  first unwrapping the outer JSON string, then deserializing the inner content. Added a unit test

dotnet/tests/Microsoft.Agents.AI.DurableTask.IntegrationTests/WorkflowConsoleAppSamplesValidation.cs
```diff
@@ -181,6 +181,200 @@ private void AssertNoError(string line)
...
+            Assert.True(inputSent, "Input was not sent to the workflow.");
+            Assert.True(foundStartedRun, "Streaming run was not started.");
+            Assert.True(foundExecutorInvoked, "ExecutorInvokedEvent not found in stream.");
+            Assert.True(foundExecutorCompleted, "ExecutorCompletedEvent not found in stream.");
+            Assert.True(foundLookupStarted, "OrderLookupStartedEvent not found in stream.");
+            Assert.True(foundOrderFound, "OrderFoundEvent not found in stream.");
+            Assert.True(foundCancelProgress, "CancellationProgressEvent not found in stream.");
+            Assert.True(foundOrderCancelled, "OrderCancelledEvent not found in stream.");
+            Assert.True(foundEmailSent, "EmailSentEvent not found in stream.");
+            Assert.True(foundYieldedOutput, "WorkflowOutputEvent not found in stream.");
```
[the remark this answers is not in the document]
kshyju: @copilot Can you check again? Your last comment does not seem correct.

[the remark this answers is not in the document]
kshyju: This comment is not valid.

dotnet/samples/Durable/Workflow/ConsoleApps/06_WorkflowSharedState/06_WorkflowSharedState.csproj
```diff
@@ -0,0 +1,29 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
cgillum, line 1: I see we have a _05 and a _07 sample. What about _06?

dotnet/samples/Durable/Workflow/ConsoleApps/07_WorkflowSharedState/Executors.cs
```diff
@@ -0,0 +1,185 @@
...
+
+        // Use ReadOrInitStateAsync — only initializes if no value exists yet
+        string shippingTier = await context.ReadOrInitStateAsync(
+            "shippingTier",
+            () => "Express",
+            cancellationToken: cancellationToken);
+        Console.WriteLine($"    Read from shared state: shippingTier = {shippingTier}");
+
+        // Write shipping details under a custom "shipping" scope.
+        // Scoped keys are isolated from the default namespace, so "carrier" here
```
cgillum: Just my ignorance: What is a "default namespace" and what is a "scoped key"?

dotnet/samples/Durable/Workflow/ConsoleApps/06_WorkflowSharedState/Program.cs
```diff
@@ -0,0 +1,117 @@
...
+            {
+                case WorkflowOutputEvent e:
+                    Console.WriteLine($"  [Output] {e.SourceId}: {e.Data}");
+                    break;
+
+                case DurableWorkflowCompletedEvent e:
+                    Console.WriteLine($"  Completed: {e.Result}");
+                    break;
+
+                case DurableWorkflowFailedEvent e:
```
cgillum, line 103: Sorry if this is outside the scope of this PR, but IIRC there are some built-in MAF events for various lifecycle state transitions. Does MAF already have something for workflow completion/failed? I'm wondering why we need to have Durable-specific versions.

dotnet/samples/Durable/Workflow/ConsoleApps/07_WorkflowSharedState/Executors.cs
```diff
@@ -0,0 +1,185 @@
...
+        // won't collide with a "carrier" key in the default scope.
+        await context.QueueStateUpdateAsync("carrier", "Contoso Express", scopeName: "shipping", cancellationToken: cancellationToken);
+        await context.QueueStateUpdateAsync("estimatedDays", 2, scopeName: "shipping", cancellationToken: cancellationToken);
+        Console.WriteLine("    Wrote to shared state: shipping:carrier = Contoso Express");
+        Console.WriteLine("    Wrote to shared state: shipping:estimatedDays = 2");
+
+        // Verify we can read the audit entry from the previous step
+        AuditEntry? previousAudit = await context.ReadStateAsync<AuditEntry>("audit:validate", cancellationToken: cancellationToken);
```
cgillum: More ignorance on my part: I noticed that the update API has a separate scope and key value, but it seems the read API only has a single input type containing both the scope and the key. Is this a MAF convention? I'm wondering why we wouldn't have a more type-safe API for this.

dotnet/samples/Durable/Workflow/ConsoleApps/07_WorkflowSharedState/README.md
```diff
@@ -0,0 +1,68 @@
...
+Return values carry primary business data through the pipeline (`OrderDetails` → `OrderDetails` → payment ref → invoice string). Shared state carries side-channel data that doesn't belong in the message chain:
+
+| Executor | Returns (message flow) | Reads from State | Writes to State |
+|----------|----------------------|-----------------|-----------------|
+| **ValidateOrder** | `OrderDetails` | — | `taxRate`, `audit:validate` |
+| **EnrichOrder** | `OrderDetails` (pass-through) | `audit:validate` | `shippingTier`, `audit:enrich`, `shipping:carrier`, `shipping:estimatedDays` |
+| **ProcessPayment** | payment ref string | `taxRate` | `audit:payment` |
+| **GenerateInvoice** | invoice string | `audit:validate`, `audit:enrich`, `audit:payment` | clears `shipping` scope |
+
+> **Note:** `EnrichOrder` writes `carrier` and `estimatedDays` under the `"shipping"` scope using `scopeName: "shipping"`. Scoped keys are isolated from the default namespace, so a key like `"carrier"` in the `"shipping"` scope won't collide with a `"carrier"` key in the default scope.
```
cgillum: nit: instead of `**Note:**` we should use the standard callout syntax - e.g.
```markdown
> [!NOTE]
> blah blah blah
```
I usually have to instruct Copilot about this since it prefers the ad-hoc syntax that you have here.

dotnet/samples/Durable/Workflow/ConsoleApps/06_WorkflowSharedState/README.md
```diff
@@ -0,0 +1,68 @@
...
+    Read from shared state: taxRate = 8.5%
+    Wrote to shared state: audit:payment
+  [Output] ProcessPayment: Payment processed. Total: $271.24 (tax: $21.25). Ref: PAY-abc123def456
+    Read from shared state: 3 audit entries
+    Cleared shared state scope: shipping
+  [Output] GenerateInvoice: Invoice complete. Payment: "PAY-abc123def456". Audit trail: [ValidateOrder → EnrichOrder → ProcessPayment]
+  Completed: Invoice complete. Payment: "PAY-abc123def456". Audit trail: [ValidateOrder → EnrichOrder → ProcessPayment]
+```
+
+### Viewing Workflows in the DTS Dashboard
```
cgillum, line 65: Is there a way to view the workflow's shared state in the dashboard?

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableActivityContext.cs
```diff
@@ -11,11 +11,20 @@ namespace Microsoft.Agents.AI.DurableTask.Workflows;
 /// A workflow context for durable activity execution.
 /// </summary>
 /// <remarks>
-/// Some of the methods are returning default for this version. Those method will be updated with real implementations in follow up PRs.
+/// State is passed in from the orchestration and updates are collected for return.
+/// Events emitted during execution are collected and returned to the orchestration
+/// as part of the activity output for streaming to callers.
 /// </remarks>
 [DebuggerDisplay("Executor = {_executor.Id}, StateEntries = {_initialState.Count}")]
 internal sealed class DurableActivityContext : IWorkflowContext
```
cgillum: I'm a bit concerned about the name `DurableActivityContext` because it sounds very similar to the Durable Functions TaskActivityContext, which is completely unrelated to MAF. Should we instead give this class a workflow-specific name, like `DurableWorkflowContext`? I assume there isn't any particular reason to use "activity" in the name since this is for executors, and the fact that it runs as an activity isn't really relevant?

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableActivityContext.cs
```diff
@@ -33,12 +42,40 @@ internal DurableActivityContext(Dictionary<string, string>? initialState, Execut
     /// <summary>
     /// Gets the messages sent during activity execution via <see cref="SendMessageAsync"/>.
     /// </summary>
-    internal List<SentMessageInfo> SentMessages { get; } = [];
+    internal List<TypedPayload> SentMessages { get; } = [];
+
+    /// <summary>
+    /// Gets the events that were added during activity execution.
+    /// </summary>
+    internal List<WorkflowEvent> Events { get; } = [];
```
cgillum: Would it make sense to call this `QueuedEvents` or `OutboundEvents` to make it more specific? We use "events" in various places in Durable Task code for both "inbound" and "outbound" use cases and the usage here is a bit ambiguous.

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableActivityContext.cs
```diff
@@ -64,44 +101,227 @@ public ValueTask SendMessageAsync(
...
+
+        return default;
+    }
 
     /// <inheritdoc/>
-    public ValueTask RequestHaltAsync() => default;
+    public ValueTask RequestHaltAsync()
+    {
+        this.HaltRequested = true;
+        this.Events.Add(new DurableHaltRequestedEvent(this._executor.Id));
```
cgillum: Hmm...so we have to emit our own lifecycle events for cases like the workflow halting?

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableActivityContext.cs
```diff
@@ -64,44 +101,227 @@ public ValueTask SendMessageAsync(
...
     public ValueTask QueueClearScopeAsync(
         string? scopeName = null,
-        CancellationToken cancellationToken = default) => default;
+        CancellationToken cancellationToken = default)
+    {
+        this.ClearedScopes.Add(scopeName ?? DefaultScopeName);
+
+        // Remove any pending updates in this scope (snapshot keys to allow removal during iteration)
+        string scopePrefix = GetScopePrefix(scopeName);
+        foreach (string key in this.StateUpdates.Keys.ToArray())
```
cgillum: nit: I usually go with `ToList()` instead of `ToArray()` because `ToArray()` has a slower implementation last I checked.

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableActivityOutput.cs
```diff
@@ -3,17 +3,37 @@
...
+    /// Gets or sets the executor result.
     /// </summary>
     public string? Result { get; set; }
 
     /// <summary>
-    /// Gets or sets the collection of messages that have been sent.
+    /// Gets or sets the state updates (scope-prefixed key to value; null indicates deletion).
     /// </summary>
-    public List<SentMessageInfo> SentMessages { get; set; } = [];
+    public Dictionary<string, string?> StateUpdates { get; set; } = [];
```
cgillum: Do all these collection properties need to be settable?

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableRunStatus.cs
```diff
@@ -0,0 +1,49 @@
...
+
+namespace Microsoft.Agents.AI.DurableTask.Workflows;
+
+/// <summary>
+/// Represents the execution status of a durable workflow run.
+/// </summary>
+public enum DurableRunStatus
+{
+    /// <summary>
+    /// The orchestration instance was not found.
```
cgillum: It looks like this is a public API, so should we be using "workflow" instead of "orchestration" in these descriptions?

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableRunStatus.cs
```diff
@@ -0,0 +1,49 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+namespace Microsoft.Agents.AI.DurableTask.Workflows;
+
+/// <summary>
+/// Represents the execution status of a durable workflow run.
+/// </summary>
+public enum DurableRunStatus
```
cgillum, line 8: I'm also curious why we need to define this ourselves vs. use some workflow status enum defined in MAF?

cgillum (commented):
I haven't gotten through the full PR yet, but here's some initial feedback/questions. Mostly very minor things so far.

dotnet/samples/Durable/Workflow/ConsoleApps/06_WorkflowSharedState/06_WorkflowSharedState.csproj
kshyju: sorry! My bad. Renamed the 07 sample to 06 now.

dotnet/samples/Durable/Workflow/ConsoleApps/07_WorkflowSharedState/Executors.cs
kshyju: That comment shouldn't be here. It was more of an implementation detail (how the keys are stored behind the scene when QueueStateUpdateAsync method is called). I updated the comments to remove it, so the comments are cleaner.

dotnet/samples/Durable/Workflow/ConsoleApps/07_WorkflowSharedState/README.md
kshyju: Yea, Fixed.

dotnet/samples/Durable/Workflow/ConsoleApps/07_WorkflowSharedState/Executors.cs
kshyju: The read and update APIs actually have the same signature. Both take (string key, string? scopeName = null, ...):

   - [ReadStateAsync](https://github.com/microsoft/agent-framework/blob/b70030dae/dotnet/src/Microsoft.Agents.AI.Workflows/IWorkflowContext.cs#L68)
   - [QueueStateUpdateAsync](https://github.com/microsoft/agent-framework/blob/b70030dae/dotnet/src/Microsoft.Agents.AI.Workflows/IWorkflowContext.cs#L139)

I think my previous sample code was not the easiest to follow. The audit keys used ` :` in the key name (e.g., "audit:validate"), which made it look like a combined scope+key format. I've cleaned this up now.
   - Renamed audit keys to camelCase (auditValidate, auditEnrich, auditPayment)
   - Simplified the scoped write to a single key (carrier with scopeName: "shipping")
   - Added a scoped read in GenerateInvoice so the sample demonstrates the full write → read → clear lifecycle with scopes

dotnet/samples/Durable/Workflow/ConsoleApps/06_WorkflowSharedState/Program.cs
kshyju: Good question. MAF core currently doesn't have a `WorkflowCompletedEvent` at all, so `DurableWorkflowCompletedEvent` fills that gap. For the failure case, MAF has `WorkflowErrorEvent` but it carries an `Exception?` object. In the durable streaming context (via `WatchStreamAsync`), we only have the error message string from the orchestration status. There's no `Exception` instance available to construct a `WorkflowErrorEvent`. That's why `DurableWorkflowFailedEvent` carries `string ErrorMessage` instead.
For halt, MAF has `RequestHaltEvent` but it's `internal`, so it can't be yielded to external stream consumers.

We are using existing MAF events which are available to use in other areas of durable workflow (ex: `WorkflowOutputEvent` , `ExecutorInvokedEvent` and `ExecutorCompletedEvent`)

 All three durable specific evernts I created extend `WorkflowEvent` from MAF.
 
 I have a GH issue opened in the repo to expose these events as public. Once they are(before GA), we can switch to them and remove our copy. https://github.com/microsoft/agent-framework/issues/4063

dotnet/samples/Durable/Workflow/ConsoleApps/06_WorkflowSharedState/README.md
kshyju: Yes. Shared state is visible in the DTS dashboard, but not in a pretty/dedicated view. You can see state updates by clicking on an activity's "action->event state" and inspecting its raw input (state dictionary) and output (stateUpdates,  clearedScopes). There is no dedicated view today, but that could be a nice DTS UX enhancement in the future.

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableActivityContext.cs
kshyju: Yea, good point. I renamed it to DurableWorkflowContext. This is still an internal class.

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableActivityContext.cs
kshyju: Good point. I renamed the internal property to `OutboundEvents` to clarify these are events flowing out from the executor. I did not rename the property in DurableActivityOutput type as I think Events make sense there as this type name clearly says it is the output of an executor/activity.

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableActivityContext.cs
kshyju: There is a gap today. I have a GH issue opened in the repo to expose these events as public. Once they are(before GA), we can switch to them and remove our copy. https://github.com/microsoft/agent-framework/issues/4063

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableActivityContext.cs
kshyju: Oh okay. Switched to that. It might be interesting to benchmark in the latest .net versions though.

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableActivityOutput.cs
kshyju: Changed all properties from `set` to `init`. They only need to be settable during construction/deserialization, not afterward.

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableRunStatus.cs
kshyju: Sounds good. Updated the description.

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableRunStatus.cs
kshyju: MAF's `RunStatus` tracks in-process run lifecycle (`NotStarted`, `Idle`, `PendingRequests`, `Ended`, `Running`)
https://github.com/microsoft/agent-framework/blob/6fd50464b09b190b9384dbfd74dda508164af93e/dotnet/src/Microsoft.Agents.AI.Workflows/RunStatus.cs#L8
The  `DurableRunStatus` I added maps durable orchestration states (`Completed`, `Failed`, `Terminated`, `Suspended`, `NotFound`) which have no equivalent in MAF's `RunStatus` today. Only `Running` overlaps. Longer term, it might make sense to unify these at the MAF layer, but the current states don't map cleanly.

dotnet/samples/Durable/Workflow/ConsoleApps/06_WorkflowSharedState/Program.cs
cgillum: Thanks for the explanation. One thought I have is that we can do better than just `string ErrorMessage` and provide `TaskFailureDetails FailureDetails` from the Durable Task SDK so that the user gets the error type, message, stack trace, and inner-failure details.

dotnet/samples/Durable/Workflow/ConsoleApps/06_WorkflowSharedState/README.md
cgillum: Got it. I think this should be good enough for now. It might be worth guiding the user to look at the activity inputs to see the snapshort of the state that the executor ran with and look at the activity outputs to see what changes to the state were made.

cgillum: Oh, I just realized that you added a new `[!NOTE]` about this already. I wonder if we should just make this a regular paragraph though. I'm also noticing that we're talking about "activity" inputs/outputs in the description below, but we should probably instead talk in terms of executors since that's the terminology they're likely more familiar with.

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableActivityOutput.cs
```diff
@@ -1,19 +1,39 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
 
 namespace Microsoft.Agents.AI.DurableTask.Workflows;
 
 /// <summary>
-/// Output payload from activity execution, containing the result and other metadata.
+/// Output payload from activity execution, containing the result, state updates, and emitted events.
 /// </summary>
 internal sealed class DurableActivityOutput
```
cgillum: Maybe call this a `DurableExecutorOutput`?

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableRunStatus.cs
cgillum: Makes sense. I wonder if we could/should expose both `RunStatus` and `DurableRunStatus` to the user so that existing MAF tooling can consume the MAF status and the durable status can be used to get more information?

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableStreamingWorkflowRun.cs
```diff
@@ -0,0 +1,410 @@
...
+
+            bool hasNewEvents = false;
+
+            // Always drain any unread events from custom status before checking terminal states.
+            // The orchestration may complete before the next poll, so events would be lost if we
+            // check terminal status first.
+            if (metadata.SerializedCustomStatus is not null)
+            {
+                DurableWorkflowCustomStatus? customStatus = TryParseCustomStatus(metadata.SerializedCustomStatus);
+                if (customStatus is not null)
```
cgillum: nit: it's more conventional to have a `TryXXX` method return a `bool` and put the result as an `out` parameter. That would also allow you to combine these two lines into one and remove the null check.

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableStreamingWorkflowRun.cs
```diff
@@ -0,0 +1,410 @@
...
+                    foreach (WorkflowEvent evt in events)
+                    {
+                        yield return evt;
+                    }
+
+                    yield return new DurableWorkflowCompletedEvent(outputResult.Result);
+                }
+                else
+                {
+                    yield return new DurableWorkflowCompletedEvent(metadata.SerializedOutput);
```
cgillum: Is this ever expected? I think we should add a comment here about when this situation might occur and how we expect users to consume the raw serialized output.

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableStreamingWorkflowRun.cs
```diff
@@ -0,0 +1,410 @@
...
+                    }
+                }
+            }
+
+            // Check terminal states after draining events from custom status
+            if (metadata.RuntimeStatus == OrchestrationRuntimeStatus.Completed)
+            {
+                // The framework clears custom status on completion, so events may be in
+                // SerializedOutput as a DurableWorkflowResult wrapper.
+                DurableWorkflowResult? outputResult = TryParseWorkflowResult(metadata.SerializedOutput);
```
cgillum: nit: same comment about `TryXXX` methods.

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableStreamingWorkflowRun.cs
```diff
@@ -0,0 +1,410 @@
...
+                    yield return new DurableWorkflowCompletedEvent(metadata.SerializedOutput);
+                }
+
+                yield break;
+            }
+
+            if (metadata.RuntimeStatus == OrchestrationRuntimeStatus.Failed)
+            {
+                string errorMessage = metadata.FailureDetails?.ErrorMessage ?? "Workflow execution failed.";
+                yield return new DurableWorkflowFailedEvent(errorMessage);
```
cgillum: Related to an earlier comment - ideally, we'd send back the full `metadata.FailureDetails` object.

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableStreamingWorkflowRun.cs
```diff
@@ -0,0 +1,410 @@
...
+
+        if (metadata.RuntimeStatus == OrchestrationRuntimeStatus.Completed)
+        {
+            return ExtractResult<TResult>(metadata.SerializedOutput);
+        }
+
+        if (metadata.RuntimeStatus == OrchestrationRuntimeStatus.Failed)
+        {
+            string errorMessage = metadata.FailureDetails?.ErrorMessage ?? "Workflow execution failed.";
+            throw new InvalidOperationException(errorMessage);
```
cgillum: It might be more helpful to throw a `TaskFailedException` here and pass the `metadata.FailureDetails` is a constructor argument.

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/IStreamingWorkflowRun.cs
```diff
@@ -0,0 +1,42 @@
...
+namespace Microsoft.Agents.AI.DurableTask.Workflows;
+
+/// <summary>
+/// Represents a workflow run that supports streaming workflow events as they occur.
+/// </summary>
+/// <remarks>
+/// This interface defines the contract for streaming workflow runs in durable execution
+/// environments. Implementations provide real-time access to workflow events.
+/// </remarks>
+public interface IStreamingWorkflowRun
```
cgillum, line 14: I forget if I asked this already in the previous review, but why do we need our own interface types? Do the MAF ones not work for us?

dotnet/tests/Microsoft.Agents.AI.DurableTask.UnitTests/Workflows/DurableStreamingWorkflowRunTests.cs
```diff
@@ -0,0 +1,615 @@
...
+        List<WorkflowEvent> events = [];
+        await foreach (WorkflowEvent evt in run.WatchStreamAsync())
+        {
+            events.Add(evt);
+        }
+
+        // Assert
+        Assert.Equal(2, events.Count);
+        Assert.IsType<DurableHaltRequestedEvent>(events[0]);
+        Assert.IsType<DurableWorkflowCompletedEvent>(events[1]);
```
cgillum: Should we also evaluate the payloads for these events, or is that covered elsewhere?

cgillum (commented):
Got through the full PR! Added a few more comments.

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/IStreamingWorkflowRun.cs
kshyju: MAF's [StreamingRun](https://github.com/microsoft/agent-framework/blob/main/dotnet/src/Microsoft.Agents.AI.Workflows/StreamingRun.cs) is a concrete class tied to in process execution.It holds an AsyncRunHandle and supports SendResponseAsync, TrySendMessageAsync, and CancelRunAsync, none of which apply to the durable (polling-based) execution model. The durable streaming run polls orchestration metadata externally and has its own lifecycle (e.g., GetStatusAsync returning DurableRunStatus, WaitForCompletionAsync<T>). We defined IStreamingWorkflowRun as a minimal interface (RunId + WatchStreamAsync) that captures just the common streaming contract. If MAF core library introduces an abstraction over both in-process and durable streaming, we can switch to that.

dotnet/tests/Microsoft.Agents.AI.DurableTask.UnitTests/Workflows/DurableStreamingWorkflowRunTests.cs
kshyju: Yea, updated tests to assert that as well.

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableActivityOutput.cs
kshyju: Makes sense. Renamed to DurableExecutorOutput.

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableStreamingWorkflowRun.cs
kshyju: Fixed.

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableStreamingWorkflowRun.cs
kshyju: I added this as a defensive fallback. Ideally this line of code should not be executed as we expect the runner to always wraps output in DurableWorkflowResult. I changed this to return a more strict response , DurableWorkflowFailedEvent now with the failure details.

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableStreamingWorkflowRun.cs
kshyju: Fixed

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableStreamingWorkflowRun.cs
kshyju: Yes, updated to include FailureDetails.

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableStreamingWorkflowRun.cs
kshyju: Switched to TaskFailedException with FailureDetails

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableStreamingWorkflowRun.cs
```diff
@@ -200,11 +205,20 @@ private async IAsyncEnumerable<WorkflowEvent> WatchStreamAsync(
 
         if (metadata.RuntimeStatus == OrchestrationRuntimeStatus.Failed)
         {
-            string errorMessage = metadata.FailureDetails?.ErrorMessage ?? "Workflow execution failed.";
-            throw new InvalidOperationException(errorMessage);
+            if (metadata.FailureDetails is not null)
+            {
+                throw new TaskFailedException(
+                    taskName: this.WorkflowName,
+                    taskId: 0,
```
cgillum: nit: use `-1` instead of `0` when the task ID is not relevant.

dotnet/src/Microsoft.Agents.AI.DurableTask/Workflows/DurableStreamingWorkflowRun.cs
```diff
@@ -252,11 +268,12 @@ private static (List<WorkflowEvent> Events, int UpdatedIndex) DrainNewEvents(Lis
     /// </remarks>
     [UnconditionalSuppressMessage("AOT", "IL3050", Justification = "Deserializing workflow result wrapper.")]
     [UnconditionalSuppressMessage("Trimming", "IL2026", Justification = "Deserializing workflow result wrapper.")]
-    private static DurableWorkflowResult? TryParseWorkflowResult(string? serializedOutput)
+    private static bool TryParseWorkflowResult(string? serializedOutput, out DurableWorkflowResult result)
```
cgillum: nit: it's more conventional to do `[NotNullWhen(true)] DurableWorkflowResult? result` for an out parameter in a `TryXXX` method.

cgillum (approved):
Changes look good. My remaining comments are non-blocking.

PR #4035, opened 2026-02-18
.NET: Fix duplicate messageId in consecutive TOOL_CALL_RESULT SSE eve…

### LINKED ISSUE #3962: .NET: [Bug]: [AG-UI] MapAGUI reuses the same messageId for consecutive TOOL_CALL_RESULT SSE events
### Description

When using MapAGUI to expose an Agent Framework workflow as an AG-UI SSE endpoint, consecutive TOOL_CALL_RESULT events after the first one are emitted with the same messageId as the first result. This causes React key collisions in CopilotKit's frontend, which uses messageId as a unique key for rendering.

The issue occurs when an agent makes multiple tool calls in sequence (e.g., two tool1 calls followed by a tool2 call) all three TOOL_CALL_RESULT events share the same messageId.

### Code Sample

```markdown

```

```markdown
Backend SSE log showing three events with the same messageId:
SSE Event #9:  type=TOOL_CALL_RESULT messageId=5bc4e92c346b443ca92c3f43a9bf21ef <= 1st (OK)
SSE Event #13: type=TOOL_CALL_RESULT messageId=5bc4e92c346b443ca92c3f43a9bf21ef <= DUPLICATE
SSE Event #17: type=TOOL_CALL_RESULT messageId=5bc4e92c346b443ca92c3f43a9bf21ef <= DUPLICATE

Frontend React errors:
Encountered two children with the same key, `5bc4e92c346b443ca92c3f43a9bf21ef-custom-after`
Encountered two children with the same key, `5bc4e92c346b443ca92c3f43a9bf21ef-custom-before`
```

### Package Versions

Microsoft.Agents.AI: 1.0.0-preview.260209.1, Microsoft.Agents.AI.Hosting.AGUI.AspNetCore: 1.0.0-preview.260209.1

### .NET Version

.NET 10

### Additional Context

- Workflow type: Handoff-based (MapAGUI endpoint)
- Important: Only TOOL_CALL_RESULT events are affected. Other event types like TEXT_MESSAGE_START / TEXT_MESSAGE_CONTENT / TEXT_MESSAGE_END correctly

kallebelins:
When multiple `FunctionResultContent` items share the same `ChatResponseUpdate`, the `AsAGUIEventStreamAsync` method now generates a deterministic unique `messageId` by combining the original `MessageId` with the `CallId` (e.g. `msg1_call_1`, `msg1_call_2`), preserving traceability while satisfying the AG-UI protocol uniqueness constraint. Fixes #3962

### Motivation and Context

The AG-UI protocol specification requires every SSE event of type `TOOL_CALL_RESULT` to carry a **unique** `messageId`. However, in `Microsoft.Extensions.AI`, when an LLM returns multiple tool call results in a single response, all `FunctionResultContent` items are grouped into a single `ChatResponseUpdate` that shares the same `MessageId`. This means the `AsAGUIEventStreamAsync` method in ChatResponseUpdateAGUIExtensions.cs was emitting consecutive `TOOL_CALL_RESULT` events with **identical** `messageId` values, violating the AG-UI spec and causing client-side issues (e.g., in `@ag-ui/core` clients that key events by `messageId`).

### Description

**Root cause:** In the `AsAGUIEventStreamAsync` method, the `FunctionResultContent` → `ToolCallResultEvent` mapping was directly assigning `chatResponse.MessageId` to each event's `MessageId`. When a single `ChatResponseUpdate` contained 2+ `FunctionResultContent` items (which is the normal case when the LLM executes multiple server-side tools in parallel), all resulting SSE events had the same `messageId`.

**Fix (1 line change in source):** Instead of using `chatResponse.MessageId` directly, the `messageId` is now composed deterministically as:

```
{chatResponse.MessageId!}_{functionResultContent.CallId}
```

This approach:
- **Satisfies AG-UI uniqueness** — since `CallId` is unique per tool call, the composed `messageId` is unique per `TOOL_CALL_RESULT` event
- **Preserves traceability** — the original `MessageId` is still embedded, enabling correlation back to the source `ChatResponseUpdate`
- **Is deterministic** — the same inputs always produce the same `messageId`, supporting idempotent replays
- **Uses the null-forgiving operator (`!`)** on `chatResponse.MessageId!` for consistency with other `MessageId` usages in this file (lines 362, 375, 464)

**Tests added (4 new test cases):**

| Test | Scenario |
|------|----------|
| `ConsecutiveToolCallResults_HaveDistinctMessageIds` | 2 `FunctionResultContent` in a single `ChatResponseUpdate` → asserts distinct `messageId` on each `TOOL_CALL_RESULT` event |
| `ToolCallResultsFromSeparateUpdates_HaveDistinctMessageIds` | 2 `FunctionResultContent` in separate `ChatResponseUpdate` objects sharing the same `MessageId` → asserts distinct `messageId` |
| `ThreeConsecutiveToolCallResults_AllHaveUniqueMessageIds` | 3 `FunctionResultContent` in one update → asserts all `messageId` values are unique via `HashSet` |
| `SingleToolCallResult_HasValidMessageId` | 1 `FunctionResultContent` → asserts `messageId` is non-null and non-empty (regression guard) |

### Contribution Checklist

- [ ] **Is this a breaking change?** No — the generated `messageId` format changes from `msg1` to `msg1_call_1`, but this is a bug fix (the previous value was invalid per the AG-UI spec)


alliscode:
Hi @kallebelins thanks for the PR. Could you fix the formatting issues so we can get this in?

alliscode (approved):

markwallace-microsoft:
@kallebelins, friendly reminder — this issue is waiting on your response. Please share any updates when you get a chance. (This is an automated message.)

Issue #3499, opened 2026-01-29
Support for Agent Skills


markwallace-microsoft:
Add support for this specification https://agentskills.io/specification and have it work with any model

helloxubo:
@markwallace-microsoft   
Hello, how is the latest progress?

coldsaber:
I also want to ask about the progress. I checked the codebase, and it seems that all skill invocations currently have to go through the Anthropics agent, which means I need an Anthropics API token or account. This setup feels overly closed.

However, at the same time, I saw the explanation about agent skills in python/AGENTS.md, which ( showed below) suggests that skills can be invoked simply by placing them in the .github/skills folder. This is confusing to me.

> **Agent Skills** (`.github/skills/`) — detailed, task-specific instructions loaded on demand:
> - `python-development` — coding standards, type annotations, docstrings, logging, performance
> - `python-testing` — test structure, fixtures, async mode, running tests
> - `python-code-quality` — linting, formatting, type checking, prek hooks, CI workflow
> - ...
>

markwallace-microsoft:
@helloxubo @coldsaber We have started work on this and are planning to release something in the next couple of weeks

helloxubo:
> [@helloxubo](https://github.com/helloxubo) [@coldsaber](https://github.com/coldsaber) We have started work on this and are planning to release something in the next couple of weeks

Great! I'm really looking forward to it.

monuminu:
looking forward for this

duguankui:
What wonderful news!

sheng-jie:
@markwallace-microsoft Great  job! At present, I see that the implementation only includes `load_skills ` and `read_skill_resources` two tools. Is there any plan to support run scripts in agent skills?

PR #4122, opened 2026-02-20
.NET: Support Agent Skills

SergeyMenshykh:
## Summary

Adds `FileAgentSkillsProvider`, an `AIContextProvider` that discovers and exposes Agent Skills from filesystem directories following the [Agent Skills specification](https://agentskills.io/) progressive disclosure pattern: **advertise → load → read resources**.

## Changes

- **`FileAgentSkillsProvider`** — discovers `SKILL.md` files from configured directories, advertises skills via system prompt injection, and provides `load_skill` / `read_skill_resource` tools for on-demand access.
- **`FileAgentSkillsProviderOptions`** — configuration for custom prompt templates.
- **`FileAgentSkillLoader`** / **`FileAgentSkill`** / **`SkillFrontmatter`** — internal helpers and model classes for skill discovery, frontmatter parsing, and secure resource reading (path traversal / symlink guards).
- **Unit tests** covering discovery, loading, resource reading, and security scenarios.
- **Sample** (`Agent_Step01_BasicSkills`) demonstrating usage with an expense-report skill.
- `FileAgentSkillsProvider` is marked as **experimental** (`MAAI001`) using `DiagnosticIds`.

## Key Design Decisions

- Skills are read-only static content; no code execution.
- XML-escaping for prompt injection safety.
- Resource reads are guarded against path traversal and symlink escape.
- Only skills from trusted sources should be used.

## Out of Scope
- Scripts execution, which will be added in one of the following PRs.

Closes: https://github.com/microsoft/agent-framework/issues/3499

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI/Skills/FileAgentSkillsProvider.cs
+++ b/dotnet/src/Microsoft.Agents.AI/Skills/FileAgentSkillsProvider.cs
@@ -170,7 +170,7 @@ public sealed partial class FileAgentSkillsProvider : AIContextProvider
     {
         string promptTemplate = DefaultSkillsInstructionPrompt;
 
-        if (options?.SkillsInstructionPrompt is not null optionsInstructions)
+        if (options?.SkillsInstructionPrompt is { } optionsInstructions)
         {
             try
             {


dotnet/samples/GettingStarted/AgentSkills/Agent_Step01_BasicSkills/Program.cs
```diff
@@ -0,0 +1,49 @@
...
+using OpenAI.Responses;
+
+// --- Configuration ---
+string endpoint = Environment.GetEnvironmentVariable("AZURE_OPENAI_ENDPOINT")
+    ?? throw new InvalidOperationException("AZURE_OPENAI_ENDPOINT is not set.");
+string deploymentName = Environment.GetEnvironmentVariable("AZURE_OPENAI_DEPLOYMENT_NAME") ?? "gpt-4o-mini";
+
+// --- Skills Provider ---
+// Discovers skills from the 'skills' directory and makes them available to the agent
+var skillsProvider = new FileAgentSkillsProvider(skillPath: Path.Combine(Directory.GetCurrentDirectory(), "skills"));
```
westey-m: Consider AppContext.BaseDirectory, which is where the DLL is, rather than the current directory, which is the folder where the user is when they run the app.
```suggestion
var skillsProvider = new FileAgentSkillsProvider(skillPath: Path.Combine(AppContext.BaseDirectory, "skills"));
```

dotnet/src/Microsoft.Agents.AI/Skills/FileAgentSkillsProvider.cs
```diff
@@ -0,0 +1,224 @@
...
+                description: "Loads the full instructions for a specific skill."),
+            AIFunctionFactory.Create(
+                this.ReadSkillResourceAsync,
+                name: "read_skill_resource",
+                description: "Reads a file associated with a skill, such as references or assets."),
+        ];
+    }
+
+    /// <inheritdoc />
+    protected override ValueTask<AIContext> InvokingCoreAsync(InvokingContext context, CancellationToken cancellationToken = default)
```
westey-m: You can also override ProviderAIContextAsync, which means you don't have to merge with the input context, just return new bits, and merging will happen in the base InvokingCoreAsync.

dotnet/src/Microsoft.Agents.AI/Skills/SkillFrontmatter.cs
```diff
@@ -0,0 +1,32 @@
...
+
+/// <summary>
+/// Parsed YAML frontmatter from a SKILL.md file, containing the skill's name and description.
+/// </summary>
+internal sealed class SkillFrontmatter
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="SkillFrontmatter"/> class.
+    /// </summary>
+    /// <param name="name">Skill name (lowercase, hyphens, max 64 characters).</param>
```
lokitoth: Should we add checks for the limits?

lokitoth (approved):

dotnet/src/Microsoft.Agents.AI/Skills/SkillFrontmatter.cs
SergeyMenshykh: Yes, the limits are checked at the loader level during skill discovery. The `SkillFrontmatter` constructor doesn't need to duplicate those checks since it's internal and only called after the loader has already validated.

dotnet/src/Microsoft.Agents.AI/Skills/FileAgentSkillsProvider.cs
```diff
@@ -0,0 +1,213 @@
...
+
+    private static string? BuildSkillsInstructionPrompt(FileAgentSkillsProviderOptions? options, Dictionary<string, FileAgentSkill> skills)
+    {
+        string promptTemplate = options?.SkillsInstructionPrompt ?? DefaultSkillsInstructionPrompt;
+
+        if (options?.SkillsInstructionPrompt is not null)
+        {
+            try
+            {
+                _ = string.Format(promptTemplate, string.Empty);
```
rogerbarreto: nit
```suggestion
        string promptTemplate = DefaultSkillsInstructionPrompt;

        if (options?.SkillsInstructionPrompt is not null optionsInstructions)
        {
            try
            {
                promptTemplate = string.Format(optionsInstructions, string.Empty);
```

dotnet/src/Microsoft.Agents.AI/Skills/FileAgentSkillsProviderOptions.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+namespace Microsoft.Agents.AI;
+
+/// <summary>
+/// Configuration options for <see cref="FileAgentSkillsProvider"/>.
+/// </summary>
+public sealed class FileAgentSkillsProviderOptions
```
rogerbarreto, line 12:
```suggestion
[Experimental(DiagnosticIds.Experiments.AgentsAIExperiments)]
public sealed class FileAgentSkillsProviderOptions
```

PR #4204, opened 2026-02-24
.NET: Support InvokeMcpTool for declarative workflows

### LINKED ISSUE #3415: .NET Workflows - Implement InvokeMCPTools support in declarative workflows
### Description

Implement InvokeMCPTools action in declarative workflow. This supports the use of InvokeMCPTool through yaml and end to end execution of the workflow.

### LINKED ISSUE #3738: .NET Workflows - Create InvokeMCPTool sample
### Description

A follow-up to https://github.com/microsoft/agent-framework/issues/3415 to demonstrate usage

### Code Sample

```markdown

```

### Language/SDK

Both

peibekwe:
### Motivation and Context

This implementation supports the `InvokeMcpTool` action in a declarative workflow to call configured mcp servers directly from the workflow without going through an AI agent first. This is different from the standard `InvokeAzureAgent` pattern where the AI decides which functions to call based on the conversation.

Includes tests and samples for the new action type and sample yaml implementations.

Fixes #3415 
Fixes #3738 

### Contribution Checklist

### DIFF
--- a/dotnet/samples/GettingStarted/Workflows/Declarative/InvokeMcpTool/Program.cs
+++ b/dotnet/samples/GettingStarted/Workflows/Declarative/InvokeMcpTool/Program.cs
@@ -8,7 +8,6 @@ using Azure.AI.Projects;
 using Azure.AI.Projects.OpenAI;
 using Azure.Core;
 using Azure.Identity;
-using Microsoft.Agents.AI.Workflows.Declarative;
 using Microsoft.Agents.AI.Workflows.Declarative.Mcp;
 using Microsoft.Extensions.Configuration;
 using Shared.Foundry;


dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/ObjectModel/InvokeMcpToolExecutor.cs
```diff
@@ -0,0 +1,367 @@
...
+
+    /// <inheritdoc/>
+    protected override bool EmitResultEvent => false;
+
+    /// <inheritdoc/>
+    protected override bool IsDiscreteAction => false;
+
+    /// <inheritdoc/>
+    [SendsMessage(typeof(ExternalInputRequest))]
+    protected override async ValueTask<object?> ExecuteAsync(IWorkflowContext context, CancellationToken cancellationToken = default)
```
moonbox3, line 64: Question to make sure I understand this properly: `ExecuteAsync` resolves `serverUrl`, `toolName`, `arguments`, etc. and builds the approval request from them. Then in `CaptureResponseAsync` (after the user approves), we re-call `GetServerUrl()`, `GetToolName()`, `GetArguments()`, etc. instead of reusing the already-resolved values? If a workflow expression like `=Local.SomeValue` resolved differently between those two points, the user would have approved parameters that differ from what actually executes. Could this scenario happen?

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/DefaultMcpToolHandler.cs
```diff
@@ -0,0 +1,228 @@
...
+        string serverUrl,
+        string? serverLabel,
+        string toolName,
+        IDictionary<string, object?>? arguments,
+        IDictionary<string, string>? headers,
+        string? connectionName,
+        CancellationToken cancellationToken = default)
+    {
+        //TODO: Handle connectionName and server label appropriately when Hosted scenario supports them. For now, ignore
+        McpServerToolResultContent resultContent = new("McpServerToolcallId");
```
moonbox3: Should this be a hard-coded string?

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative.Mcp/DefaultMcpToolHandler.cs
```diff
@@ -0,0 +1,228 @@
...
+    /// The callback receives (serverUrl, cancellationToken) and should return an HttpClient
+    /// configured with any required authentication. Return <see langword="null"/> to use a default HttpClient with no auth.
+    /// </param>
+    public DefaultMcpToolHandler(Func<string, CancellationToken, Task<HttpClient?>>? httpClientProvider = null)
+    {
+        this._httpClientProvider = httpClientProvider;
+    }
+
+    /// <inheritdoc/>
+    public async Task<McpServerToolResultContent> InvokeToolAsync(
```
moonbox3, line 45: Do we need exception handling around this?

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Extensions/JsonDocumentExtensions.cs
```diff
@@ -42,6 +42,40 @@ internal static class JsonDocumentExtensions
             };
     }
 
+    /// <summary>
+    /// Creates a VariableType.List with schema inferred from the first object element in the array.
+    /// </summary>
+    internal static VariableType GetListTypeFromJson(this JsonElement arrayElement)
```
crickman: nit - This can be `public` as the class is `internal` (or `private`?)

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Extensions/JsonDocumentExtensions.cs
```diff
@@ -116,6 +150,7 @@ VariableType DetermineElementType()
                             JsonValueKind.True => typeof(bool),
                             JsonValueKind.False => typeof(bool),
                             JsonValueKind.Number => typeof(decimal),
+                            JsonValueKind.Array => (VariableType)VariableType.ListType, // Add support for nested arrays
```
crickman, line 155: I wonder about this.  As a private method isnt `ParseRecord` only invoked when the object type is `Record`/`Object`?

dotnet/tests/Microsoft.Agents.AI.Workflows.Declarative.IntegrationTests/InvokeToolWorkflowTest.cs
```diff
@@ -16,25 +16,41 @@
 namespace Microsoft.Agents.AI.Workflows.Declarative.IntegrationTests;
 
 /// <summary>
-/// Integration tests for InvokeFunctionTool action.
-/// This test pattern can be extended for other InvokeTool types.
+/// Integration tests for InvokeFunctionTool and InvokeMcpTool actions.
 /// </summary>
-public sealed class InvokeFunctionToolWorkflowTest(ITestOutputHelper output) : IntegrationTest(output)
+public sealed class InvokeToolWorkflowTest(ITestOutputHelper output) : IntegrationTest(output)
```
crickman, line 22: Nice test update!

dotnet/tests/Microsoft.Agents.AI.Workflows.Declarative.UnitTests/ObjectModel/InvokeMcpToolExecutorTest.cs
```diff
@@ -0,0 +1,845 @@
...
+using Microsoft.Extensions.AI;
+using Moq;
+using Xunit.Abstractions;
+
+namespace Microsoft.Agents.AI.Workflows.Declarative.UnitTests.ObjectModel;
+
+/// <summary>
+/// Tests for <see cref="InvokeMcpToolExecutor"/>.
+/// </summary>
+public sealed class InvokeMcpToolExecutorTest(ITestOutputHelper output) : WorkflowActionExecutorTest(output)
```
crickman, line 20: Nice coverage!

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Microsoft.Agents.AI.Workflows.Declarative.csproj
```diff
@@ -26,6 +26,7 @@
     <PackageReference Include="Microsoft.PowerFx.Interpreter" />
     <PackageReference Include="Microsoft.Extensions.Configuration" />
     <PackageReference Include="Microsoft.Extensions.Logging" />
+    <PackageReference Include="ModelContextProtocol" />
```
alliscode: nit: Would it make sense to have a `Microsoft.Agents.AI.Workflows.Declarative.Mcp` package so that people only pull in `ModelContextProtocol` when they need it?

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/DefaultMcpToolHandler.cs
```diff
@@ -0,0 +1,228 @@
...
+        this._clientLock.Dispose();
+    }
+
+    private async Task<McpClient> GetOrCreateClientAsync(
+        string serverUrl,
+        string? serverLabel,
+        IDictionary<string, string>? headers,
+        CancellationToken cancellationToken)
+    {
+        string cacheKey = $"{serverUrl.Trim().ToUpperInvariant()}";
```
alliscode: If we only use the serverUrl, there would be no way to call the same tool with different headers such as auth tokens. This might be needed for multi-tenant scenarios?

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/ObjectModel/InvokeMcpToolExecutor.cs
peibekwe: This shouldn't happen in this scenario. Both `ExecuteAsync` and `CaptureResponseAsync` are substeps of the same action that will execute before another action gets to mutate the values.

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative.Mcp/DefaultMcpToolHandler.cs
peibekwe: We thought about this. We do not want to swallow exceptions at this point. Any unexpected exception should bubble up to the calling executor.

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/DefaultMcpToolHandler.cs
peibekwe: This value isn't used. It's an Id required by the MEAI abstraction mostly for when an agent is requesting a tool call. In this case, the tool call happens outside the scope of the agent, so we can set this to whatever we want. I'm open to ideas, we could also create a random Guid here. Thoughts?

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Extensions/JsonDocumentExtensions.cs
peibekwe: Updated.

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Extensions/JsonDocumentExtensions.cs
peibekwe: Yeah, I ran into test scenarios where an object type with nested array types failed to parse because of this omission.

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/DefaultMcpToolHandler.cs
moonbox3: I'd suggest the GUID just to be safe, but either way is fine.

peibekwe: Updated to GUID. Thanks for the feedback!

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/DefaultMcpToolHandler.cs
peibekwe: Good call, I updated the cacheKey for McpClient to use hash of the headers in a deterministic way. The HttpClient can still be shared by multiple McpClient, so this should handle the scenario you mentioned.

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Microsoft.Agents.AI.Workflows.Declarative.csproj
peibekwe: Makes sense. That way, people the main package remains thin, and folks can just add the Mcp package when they want to include Mcp tools in the workflow. I've separated it.

Issue #4140, opened 2026-02-20
.NET: [Bug]: How do you in RC1 get the messages in a session?


rwjdk:
### Description

Before RC1 you could do the following to get the messages in a session:

```cs
AgentSession session = await agent.CreateSessionAsync();
IList<ChatMessage> messagesInSession = session.GetService<IList<ChatMessage>>()!;
```

I can't for the life of me see how you can do something similar in the RC1 version :-(

### Code Sample

```markdown

```

### Error Messages / Stack Traces

```markdown

```

### Package Versions

RC1

### .NET Version

.NET10

rwjdk:
Think I found the way 

```cs
InMemoryChatHistoryProvider? provider = agent.GetService<InMemoryChatHistoryProvider>();
List<ChatMessage> messagesInSession = provider?.GetMessages(session) ?? [];
```

Personal opinion: This is quite an unintuitive and cumbersome way to get these
- You need to know these things (not discoverable)
- You need to parse the agent around all over the place and not just the `AgentSession`)
- You need to ensure that an agent has a ChatHistoryProvider; otherwise, this will not work when you resume a session from a JSON-Element before you run the first `.RunAsync` that auto-creates the ChatHistoryProvider ... Is it not a bug that `agent.DeserializeSessionAsync(jsonElement)` does not initialize the ChatHistoryProvider on the agent if missing?

What is the primary reason for there not being a `session.GetMessages()`  or the old way does not work (I understand there can be multiple providers, but still, will they not all. in some form, have a list of messages?)

westey-m:
Hi @rwjdk, this is indeed the best way to do this now and I agree that it is not the most intuitive.  This is certainly the main negative side-affect of the change we made recently.

The change was to decouple `ChatHistoryProvider` and `AIContextProvider` from the `AgentSession`.  Previously the `ChatHistoryProvider` was stored as a property on `ChatClientAgentSession`, which meant that GetService could delegate to the `ChatHistoryProvider`.  Also `InMemoryChatHistoryProvider` had the messages inside itself in a field.  Having such a stateful object on the AgentSession complicates serialization and deserialization though, especially if it needs a reference to some db client or service client.  It also means that `ChatHistoryProvider` and `AIContextProvider` cannot be used in other parts of the pipeline (e.g. as a decorator on a non ChatClientAgent AIAgent), since there's no easy way to contribute state in that case at session serialization time.

[...]

PR #4224, opened 2026-02-24
.NET: Add helpers to more easily access in-memory ChatHistory and make ChatHistoryProvider management more configurable.

westey-m:
### Motivation and Context

#4140

### Description

- Add helpers to more easily access in-memory ChatHistory.
- Make ChatHistoryProvider management on ChatClientAgent more configurable.

### Contribution Checklist


dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgentOptions.cs
```diff
@@ -59,6 +59,36 @@ public sealed class ChatClientAgentOptions
...
+    public bool WarnOnChatHistoryProviderConflict { get; set; } = true;
+
+    /// <summary>
+    /// Gets or sets a value indicating whether an exception is thrown if the underlying AI service indicates that it manages chat history
+    /// (for example, by returning a conversation id in the response), but a <see cref="ChatHistoryProvider"/> is configured for the agent.
+    /// </summary>
+    /// <value>
+    /// Default is <see langword="true"/>.
+    /// </value>
+    public bool ThrowOnChatHistoryProviderConflict { get; set; } = true;
```
rogerbarreto, line 90: Since those properties can be together in the same at the same time, I think we can use a flags approach.

### What the current 3 properties do (in practice)
These three booleans model **how to react** when there’s a configuration conflict:

- The AI service says “I manage chat history” (e.g., returns a conversation id), **but**
- the agent has a `ChatHistoryProvider` configured.

The options then control *three independent reactions*:
1. **Clear** the agent’s `ChatHistoryProvider` (`ClearOn...`)
2. **Warn** (log a warning) (`WarnOn...`)
3. **Throw** (raise an exception) (`ThrowOn...`)

### Suggestion: replace with one flags property
Define an enum like:

```csharp
[Flags]
public enum ChatHistoryProviderConflictBehavior
{
    None  = 0,
    Clear = 1 << 0,
    Warn  = 1 << 1,
    Throw = 1 << 2,
}
```

Then in `ChatClientAgentOptions`:

```csharp
public ChatHistoryProviderConflictBehavior ChatHistoryProviderConflictBehavior { get; set; }
    = ChatHistoryProviderConflictBehavior.Clear
    | ChatHistoryProviderConflictBehavior.Warn
    | ChatHistoryProviderConflictBehavior.Throw;
```

Usage:

```csharp
options.ChatHistoryProviderConflictBehavior =
    ChatHistoryProviderConflictBehavior.Warn | ChatHistoryProviderConflictBehavior.Clear;
```

### How you check it in code
Wherever you currently do `if (WarnOn...)`, etc.:

```csharp
var behavior = options.ChatHistoryProviderConflictBehavior;

if (behavior.HasFlag(ChatHistoryProviderConflictBehavior.Throw))
{
    throw new InvalidOperationException(...);
}

if (behavior.HasFlag(ChatHistoryProviderConflictBehavior.Warn))
{
    logger.LogWarning(...);
}

if (behavior.HasFlag(ChatHistoryProviderConflictBehavior.Clear))
{
    agent.ChatHistoryProvider = null;
}
```

Does it make sense?

westey-m: > Does it make sense?

Yes, an enum is generally more elegant, but enums do have other issues.  If we wanted to add an additional flag, it's a breaking change with an enum, whereas adding a new bool to options is not.  That's why I chose bools rather than an enum.

dotnet/src/Microsoft.Agents.AI.Abstractions/AgentSessionExtensions.cs
```diff
@@ -0,0 +1,67 @@
...
+    /// <remarks>
+    /// This method is only applicable when using <see cref="InMemoryChatHistoryProvider"/> and if the service does not require in-service chat history storage.
+    /// </remarks>
+    /// <param name="session">The agent session from which to retrieve in-memory chat history.</param>
+    /// <param name="messages">When this method returns, contains the list of chat history messages if available; otherwise, null.</param>
+    /// <param name="stateKey">An optional key used to identify the chat history state in the session's state bag. If null, the default key for
+    /// in-memory chat history is used.</param>
+    /// <param name="jsonSerializerOptions">Optional JSON serializer options to use when accessing the session state. If null, default options are used.</param>
+    /// <returns><see langword="true"/> if the in-memory chat history messages were found and retrieved; <see langword="false"/> otherwise.</returns>
+    public static bool TryGetInMemoryChatHistory(this AgentSession session, [MaybeNullWhen(false)] out List<ChatMessage> messages, string? stateKey = null, JsonSerializerOptions? jsonSerializerOptions = null)
```
rogerbarreto, line 28: Suggest we have a `TryGetChatHistory` and it attempts to retrieve from the underlying type, abstracting this logic (even for hosted message stores), if possible.

Suggesty adding the `Experimental` for this functions as well, doesn't seem right to me adding this into the `Abstraction` level, maybe just in the `Core`.

```suggestion
    public static bool TryGetChatHistory(this AgentSession session, [MaybeNullWhen(false)] out List<ChatMessage> messages, string? stateKey = null, JsonSerializerOptions? jsonSerializerOptions = null)
```

westey-m: The intention is not to create a new abstraction layer for all possible chat history stores, including hosted ones.  While we may want to consider doing something like that in future, it would look very different to this, live somewhere else and would require someone to build and maintain implementations.  This is very much beyond the scope of this.

InMemoryChatHistoryProvider lives in Abstractions and this method is for InMemoryChatHistoryProvider.  It wouldn't make sense for this to therefore live in core.  Let's come up with a clear strategy around experimental, since it's not clear to me why we would mark this vs other things.

rogerbarreto: > InMemoryChatHistoryProvider lives in Abstractions and this method is for InMemoryChatHistoryProvider

The target of this extension points to the `AgentSession` which is not clear it will only work for `InMemoryChatHistoryProvider`.

> While we may want to consider doing something like that in future, it would look very different to this...

Suggested `Experimental` exactly given that we do consider having something different in the future (more general/abstract), so we would have wiggle room to break it without having to obsolete.

PR #4238, opened 2026-02-25
Python: (core): Add functional workflow API

### LINKED ISSUE #3537: Python: Add functional workflow API
## Summary

Feedback that we should add a functional API for defining workflows, in addition to the current graph-based API.

## Context

Currently workflows are defined using builder patterns:
```python
workflow = (
    WorkflowBuilder()
    .register_executor(...)
    .add_edge(...)
    .build()
)
```

A functional API could look like:
```python
workflow = Workflow(
    executors={...},
    edges=[...],
)
```

## Action Items

- [ ] Design functional API surface
- [ ] Consider relationship with builder pattern (complement vs replace)
- [ ] Evaluate if this aligns with "direct Workflow instantiation" feedback in builder ergonomics issue
- [ ] Prototype and gather feedback

moonbox3:
### Motivation and Context

The functional API is a stepping stone between single-agent use and the full graph API. Users write workflows as plain async functions -- no executor classes, no edges, no builder patterns.

- Add @workflow and @step decorators for writing workflows as plain async functions
- Native Python control flow (if/else, loops, asyncio.gather) replaces graph concepts
- @step is opt-in: plain functions work inside @workflow without it. Use @step on expensive operations (agent calls, API requests) to save their results and skip re-execution on
HITL resume or crash recovery
- Streaming support via run(stream=True)
- HITL support via ctx.request_info() with replay
- .as_agent() wraps a functional workflow as an agent-compatible object

A very basic example of the functional workflow API:

```python
@workflow
async def pipeline(data: str) -> str:
    upper = await to_upper(data)
    return await reverse(upper)

result = await pipeline.run("hello")
print(result.get_outputs())  # ['OLLEH']
```

Note: `@step` is opt-in for functions where per-step checkpointing matters (for example, agent calls). Without `@step`, workflows still support HITL and checkpointing — functions just re-execute on resume.

```python
@step
async def call_agent(prompt: str) -> str:
    return (await agent.run(prompt)).text

@workflow
async def pipeline(data: str, ctx: RunContext) -> str:
    result = await call_agent(data)        # saved by @step
    validated = await validate(result)      # plain function, re-runs on resume
    feedback = await ctx.request_info(...)  # HITL pause
    return await finalize(result, feedback)
```

`ctx: RunContext` is only needed when you use HITL (request_info), custom events (add_event), or state (get_state/set_state). Otherwise, omit it for a cleaner signature.

### Description

- Closes #3537

### Contribution Checklist


markwallace-microsoft:
<a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/README.md">Python Test Coverage</a><details><summary>Python Test Coverage Report • </summary><table><tr><th>File</th><th>Stmts</th><th>Miss</th><th>Cover</th><th>Missing</th></tr><tbody><tr><td colspan="5"><b>packages/core/agent_framework</b></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_feature_stage.py">_feature_stage.py</a></td><td>116</td><td>8</td><td>93%</td><td><a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_feature_stage.py#L98">98</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_feature_stage.py#L155">155</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_feature_stage.py#L166">166</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_feature_stage.py#L173">173</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_feature_stage.py#L180">180</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_feature_stage.py#L187">187</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_feature_stage.py#L206">206</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_feature_stage.py#L234">234</a></td></tr><tr><td colspan="5"><b>packages/core/agent_framework/_workflows</b></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_events.py">_events.py</a></td><td>152</td><td>21</td><td>86%</td><td><a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_events.py#L90-L91">90&ndash;91</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_events.py#L226">226</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_events.py#L230">230</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_events.py#L232">232</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_events.py#L264">264</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_events.py#L344">344</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_events.py#L359">359</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_events.py#L374">374</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_events.py#L389">389</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_events.py#L401-L403">401&ndash;403</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_events.py#L415-L417">415&ndash;417</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_events.py#L419-L420">419&ndash;420</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_events.py#L422-L423">422&ndash;423</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_events.py#L427">427</a></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_functional.py">_functional.py</a></td><td>489</td><td>31</td><td>93%</td><td><a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_functional.py#L397-L398">397&ndash;398</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_functional.py#L464">464</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_functional.py#L487">487</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_functional.py#L496-L498">496&ndash;498</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_functional.py#L501">501</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_functional.py#L531">531</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_functional.py#L712-L713">712&ndash;713</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_functional.py#L720">720</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_functional.py#L968">968</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_functional.py#L999">999</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_functional.py#L1002-L1003">1002&ndash;1003</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_functional.py#L1011-L1013">1011&ndash;1013</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_functional.py#L1088-L1089">1088&ndash;1089</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_functional.py#L1095">1095</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_functional.py#L1478-L1479">1478&ndash;1479</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_functional.py#L1481">1481</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_functional.py#L1488-L1491">1488&ndash;1491</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_functional.py#L1512">1512</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_functional.py#L1537">1537</a></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_workflow.py">_workflow.py</a></td><td>280</td><td>19</td><td>93%</td><td><a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_workflow.py#L91">91</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_workflow.py#L270-L272">270&ndash;272</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_workflow.py#L274-L275">274&ndash;275</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_workflow.py#L293">293</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_workflow.py#L297">297</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_workflow.py#L444">444</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_workflow.py#L639">639</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_workflow.py#L660">660</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_workflow.py#L716">716</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_workflow.py#L728">728</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_workflow.py#L734">734</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_workflow.py#L739">739</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_workflow.py#L759-L761">759&ndash;761</a>, <a href="https://github.com/microsoft/agent-framework/blob/7c790b06395bb4437f385f0163ff18707fa7df86/packages/core/agent_framework/_workflows/_workflow.py#L812">812</a></td></tr><tr><td><b>TOTAL</b></td><td><b>29824</b></td><td><b>3483</b></td><td><b>88%</b></td><td>&nbsp;</td></tr></tbody></table></details>

## Python Unit Test Overview
| 5998 | 30 :zzz: | 0 :x: | 0 :fire: | 1m 35s :stopwatch: |

python/samples/03-workflows/functional/basic_pipeline.py
```diff
@@ -0,0 +1,59 @@
...
+#   - .run() that returns a WorkflowRunResult with events and outputs
+#   - .run(stream=True) for streaming events in real time
+#   - .as_agent() to use this workflow anywhere an agent is expected
+#
+# The function's first parameter receives the input from .run("...").
+# Add a `ctx: RunContext` parameter only if you need HITL, state, or custom events.
+@workflow
+async def data_pipeline(url: str) -> str:
+    """A simple sequential data pipeline."""
+    raw = await fetch_data(url)
```
eavanvalkenburg, line 37: I think it would be usefull to also demonstrate that because this is just a function, you do not have to wrap everything in steps, you can do some of the manipulation just as simple code between steps, making it a lot simpler

python/samples/03-workflows/functional/hitl_review.py
```diff
@@ -0,0 +1,77 @@
...
+# @step saves the result. When the workflow resumes after the HITL pause,
+# this returns its saved result instead of running the expensive operation again.
+@step
+async def write_draft(topic: str) -> str:
+    """Simulate writing a draft — expensive, shouldn't re-run on resume."""
+    print(f"  write_draft executing for '{topic}'")
+    return f"Draft document about '{topic}': Lorem ipsum dolor sit amet..."
+
+
+@step
```
eavanvalkenburg, line 35: how this does `step` compare to `handler` is there a lot of overlap and could we reuse steps in graphs, or handler here?

python/samples/03-workflows/functional/hitl_review.py
```diff
@@ -0,0 +1,77 @@
...
+
+async def main():
+    # Phase 1: Run until the workflow pauses for human input
+    print("=== Phase 1: Initial run ===")
+    result1 = await review_pipeline.run("AI Safety")
+
+    print(f"State: {result1.get_final_state()}")
+    assert result1.get_final_state() == WorkflowRunState.IDLE_WITH_PENDING_REQUESTS
+
+    requests = result1.get_request_info_events()
```
eavanvalkenburg, line 71: what would happen here if `state != WorkflowRunState.IDLE_WITH_PENDING_REQUESTS`?

python/samples/03-workflows/functional/agent_integration.py
```diff
@@ -0,0 +1,106 @@
...
+
+# ---------------------------------------------------------------------------
+# Workflow
+# ---------------------------------------------------------------------------
+
+
+@workflow
+async def document_pipeline(document: str) -> str:
+    """Process a document through classification, summarization, and review."""
+    classification = await classify_document(document)
```
eavanvalkenburg, line 89: as far as I can tell, all of these steps are really unnecessary, just call the agent within the workflow and pass `response.text` to the next one? or am I missing something?

python/packages/core/agent_framework/_workflows/_functional.py
```diff
@@ -0,0 +1,1106 @@
...
+    def __init__(self, func: Callable[..., Awaitable[R]], *, name: str | None = None) -> None:
+        if not inspect.iscoroutinefunction(func):
+            raise TypeError(
+                f"@step can only decorate async functions, but '{func.__name__}' is not a coroutine function."
+            )
+        self._func = func
+        self._name = name or func.__name__
+        functools.update_wrapper(self, func)
+
+    @property
```
eavanvalkenburg: do we need a property for this? why not just expose `name`?

python/packages/core/agent_framework/_workflows/_functional.py
```diff
@@ -0,0 +1,1106 @@
...
+    _framework_event_origin,  # type: ignore[reportPrivateUsage]
+)
+from ._workflow import WorkflowRunResult
+
+logger = logging.getLogger(__name__)
+
+R = TypeVar("R")
+
+# ContextVar holding the active RunContext during workflow execution.
+_active_run_ctx: ContextVar[RunContext | None] = ContextVar("_active_run_ctx", default=None)
```
eavanvalkenburg, line 67: does this mean we cannot run multiple workflows simultaneously?

python/packages/core/agent_framework/_workflows/_functional.py
```diff
@@ -0,0 +1,1106 @@
...
+        streaming: bool = False,
+        **kwargs: Any,
+    ) -> AsyncIterable[WorkflowEvent[Any]]:
+        storage = checkpoint_storage or self._checkpoint_storage
+
+        # Build context
+        ctx = RunContext(self.name, streaming=streaming, run_kwargs=kwargs if kwargs else None)
+
+        # Determine effective message for this execution
+        effective_message = message
```
eavanvalkenburg: do we need this?

python/packages/core/agent_framework/_workflows/_functional.py
```diff
@@ -0,0 +1,1106 @@
...
+                hints = {}
+
+            # Build call arguments: inject RunContext where annotated,
+            # and pass `message` to the first non-ctx parameter.
+            call_args: list[Any] = []
+            message_injected = False
+
+            for param in params:
+                resolved = hints.get(param.name, param.annotation)
+                if resolved is RunContext:
```
eavanvalkenburg: anyway we can do this without relying on people typing the param? the downside of using typing for this is that it requires people to import that object, introducing another concept that they have to learn, might be good in this case, but let's be clear about the options here.

python/samples/01-get-started/06_functional_workflow_basics.py
eavanvalkenburg, line 1: could we have a 06_functional_workflow_with_agents?

python/samples/03-workflows/functional/hitl_review.py
moonbox3: I want them to be conceptually different things. A `@handler` handles a message routed to it by the graph: it's reactive, tied to the executor contract. A `@step` marks a function call in a sequential flow: it's proactive, just "I called this function as step N." Different mental models, different names.

python/samples/03-workflows/functional/hitl_review.py
moonbox3: If `request_info()` was never reached (an early return), the workflow completes normally with state `IDLE` and `get_request_info_events()` returns an empty list. Added a comment in the sample clarifying this.

python/packages/core/agent_framework/_workflows/_functional.py
moonbox3: Yes, you can. `ContextVar` is per-asyncio-Task, so concurrent workflows each get their own context. Added a comment on the contextvar to make this clear.

python/packages/core/agent_framework/_workflows/_functional.py
moonbox3: Removed

python/packages/core/agent_framework/_workflows/_functional.py
moonbox3: `ctx: RunContext` is fully optional, if you don't need HITL, state, or custom events, just leave it out of the signature entirely. When you do need it, you're already importing `RunContext` to use its methods, so the annotation adds no extra cost. 

So I understand: are you suggesting we match on parameter name instead? It feels more implicit and error-prone.

python/samples/03-workflows/functional/hitl_review.py
eavanvalkenburg: so `requests` would be None or would the `get...` call raise?

python/packages/core/agent_framework/_workflows/_functional.py
eavanvalkenburg: remember all the discussions on typed vs untyped code in the review, if we can make untyped work, then we might want to consider supporting it. You can use RunContext without ide autocomplete... the code doesn't care

moonbox3: yeah, I agree. I've changed it now to support both, type annotation and parameter name. If the param is annotated as RunContext, that's matched first. Otherwise, any parameter named ctx gets injected. So both of these work:

```python
async def my_workflow(data: str, ctx: RunContext) -> str:
async def my_workflow(data: str, ctx) -> str:
```

python/samples/03-workflows/functional/hitl_review.py
moonbox3: `get_request_info_events()` returns an empty list. It's a filter over the event stream, so no requests means `[]`. No exception raised.

python/packages/core/agent_framework/_workflows/_functional.py
```diff
@@ -0,0 +1,1102 @@
+# Copyright (c) Microsoft. All rights reserved.
+
+"""Functional workflow API for writing workflows as plain async functions.
+
+This module provides the ``@workflow`` and ``@step`` decorators that let users
+define workflows using native Python control flow (if/else, loops,
+``asyncio.gather``) instead of a graph-based topology.
+
+A ``@workflow``-decorated async function receives its input as the first
+positional argument.  If the function needs HITL (``request_info``), custom
```
TaoChenOSU, line 15: Should it be:
```suggestion
positional argument.  If a step needs HITL (``request_info``), custom
```
?

python/packages/core/agent_framework/_workflows/_functional.py
```diff
@@ -0,0 +1,1102 @@
...
+events, or key/value state, add a :class:`RunContext` parameter — otherwise it
+can be omitted.  Inside the function, plain ``async`` calls run normally.
```
TaoChenOSU: Should this be:

```suggestion
can be omitted.  Inside the workflow, plain ``async`` calls run normally.
```
?

python/packages/core/agent_framework/_workflows/_functional.py
```diff
@@ -0,0 +1,1102 @@
...
+
+# ---------------------------------------------------------------------------
+# RunContext
+# ---------------------------------------------------------------------------
+
+
+class RunContext:
+    """Execution context injected into ``@workflow`` functions.
+
+    Every ``@workflow`` invocation receives a ``RunContext`` instance that
```
TaoChenOSU: I am a bit confused on which entity receives the context. Is it the workflow or the steps or both?

python/packages/core/agent_framework/_workflows/_functional.py
```diff
@@ -0,0 +1,1102 @@
...
+            @workflow
+            async def my_pipeline(data: str) -> str:
+                return await some_step(data)
+
+
+            # Add ctx: RunContext only when you need HITL, state, or custom events:
+            @workflow
+            async def hitl_pipeline(data: str, ctx: RunContext) -> str:
+                feedback = await ctx.request_info({"draft": data}, response_type=str)
+                return feedback
```
TaoChenOSU, line 144: My brain maps `@workflow` to the graph-based `Workflow` and `@step` to `Executor`. I can see the benefit of allowing `request_info` at the workflow level. It's kind of like an executor whose sole purpose is to get user feedback. But should we also allow `request_info` inside a `@step`?

python/packages/core/agent_framework/_workflows/_functional.py
```diff
@@ -0,0 +1,1102 @@
...
+        request_data: Any,
+        response_type: type,
+        *,
+        request_id: str | None = None,
+    ) -> Any:
+        """Request external information (human-in-the-loop).
+
+        On first execution this suspends the workflow by raising an internal
+        ``WorkflowInterrupted`` signal (caught by the framework, never exposed
+        to user code).  The caller receives a ``WorkflowRunResult`` whose
```
TaoChenOSU: > The caller receives a ``WorkflowRunResult``

What about streaming?

python/packages/core/agent_framework/_workflows/_functional.py
```diff
@@ -0,0 +1,1102 @@
...
+    async def add_event(self, event: WorkflowEvent[Any]) -> None:
+        """Add a custom event to the workflow event stream.
+
+        Use this to inject application-specific events alongside the
+        framework-generated lifecycle events.
+
+        Args:
+            event: The workflow event to append.
+        """
+        self._add_event(event)
```
TaoChenOSU: nit: we can simply expose the internal method

python/packages/core/agent_framework/_workflows/_functional.py
```diff
@@ -0,0 +1,1102 @@
...
+    def _set_cached_result(self, key: tuple[str, int], value: Any) -> None:
+        self._step_cache[key] = value
+
+    def _set_responses(self, responses: dict[str, Any]) -> None:
+        self._responses = dict(responses)
+
+    def _get_response(self, request_id: str) -> tuple[bool, Any]:
+        if request_id in self._responses:
+            return True, self._responses[request_id]
+        return False, None
```
TaoChenOSU, line 363: Are we assuming that the response can be `None`?

python/packages/core/agent_framework/_workflows/_functional.py
```diff
@@ -0,0 +1,1102 @@
...
+            # Outside a workflow — pass through directly
+            return await self._func(*args, **kwargs)
+
+        cache_key = ctx._get_step_cache_key(self.name)
+        found, cached = ctx._get_cached_result(cache_key)
+        invocation_data = deepcopy({"args": args, "kwargs": kwargs}) if args or kwargs else None
+        if found:
+            # Replay path: emit events and return cached result
+            ctx._add_event(WorkflowEvent.executor_invoked(self.name, invocation_data))
+            ctx._add_event(WorkflowEvent.executor_completed(self.name, cached))
```
TaoChenOSU: Should we create a new event type (`bypass`) or something to make this more explicit?

python/packages/core/agent_framework/_workflows/_functional.py
```diff
@@ -0,0 +1,1102 @@
...
+        if not inspect.iscoroutinefunction(func):
+            raise TypeError(
+                f"@step can only decorate async functions, but '{func.__name__}' is not a coroutine function."
+            )
+        self._func = func
+        self.name: str = name or func.__name__
+        functools.update_wrapper(self, func)
+
+    async def __call__(self, *args: Any, **kwargs: Any) -> R:
+        ctx = _active_run_ctx.get()
```
TaoChenOSU, line 506: Question: will this get the same context instance each time the step is invoked? If so, what if users want to force a rerun of all steps?

python/packages/core/agent_framework/_workflows/_functional.py
```diff
@@ -0,0 +1,1102 @@
...
+        name: Optional display name.  Defaults to ``func.__name__``.
+
+    Raises:
+        TypeError: If *func* is not an async (coroutine) function.
+    """
+
+    def __init__(self, func: Callable[..., Awaitable[R]], *, name: str | None = None) -> None:
+        if not inspect.iscoroutinefunction(func):
+            raise TypeError(
+                f"@step can only decorate async functions, but '{func.__name__}' is not a coroutine function."
```
TaoChenOSU, line 445: This is probably not super important but should we also support not async methods?

python/packages/core/agent_framework/_workflows/_functional.py
```diff
@@ -0,0 +1,1102 @@
...
+        # Discover step names referenced in the function for signature hash
+        self._step_names = self._discover_step_names(func)
+
+        # Compute a stable signature hash
+        self.graph_signature_hash = self._compute_signature_hash()
+
+        functools.update_wrapper(self, func)  # type: ignore[arg-type]
+
+    # ------------------------------------------------------------------
+    # run() — same overloaded interface as graph Workflow
```
TaoChenOSU, line 732: Should we extract these methods and make both workflow types get them for free?

python/packages/core/agent_framework/_workflows/_functional.py
```diff
@@ -0,0 +1,1102 @@
...
+
+                # Emit the return value as the workflow output.
+                if return_value is not None:
+                    ctx._add_event(WorkflowEvent.output(self.name, return_value))
+
+                # Persist step cache for response-only replay
+                self._last_step_cache = dict(ctx._step_cache)
+
+                # Yield collected events
+                for event in ctx._get_events():
```
TaoChenOSU, line 997: Question: Is this true streaming?

It looks like all events have been produced at this point.

python/packages/core/agent_framework/_workflows/_functional.py
```diff
@@ -0,0 +1,1102 @@
...
+    async def __call__(self, *args: Any, **kwargs: Any) -> R:
+        ctx = _active_run_ctx.get()
+        if ctx is None:
+            # Outside a workflow — pass through directly
+            return await self._func(*args, **kwargs)
+
+        cache_key = ctx._get_step_cache_key(self.name)
+        found, cached = ctx._get_cached_result(cache_key)
+        invocation_data = deepcopy({"args": args, "kwargs": kwargs}) if args or kwargs else None
+        if found:
```
TaoChenOSU, line 513: Should we also check if the input arguments have the same values?

TaoChenOSU: There could be scenarios where some steps are checkpointed and some are not (the ones without the decorator). If a checkpointed step depends on the output of a non-checkpointed step and its output changes, it may lead to incorrect results.

python/samples/01-get-started/06_functional_workflow_basics.py
```diff
@@ -0,0 +1,57 @@
...
+No external services are required.
+"""
+
+import asyncio
+
+from agent_framework import workflow
+
+
+# Plain async functions — no decorators needed
+async def to_upper_case(text: str) -> str:
```
TaoChenOSU, line 23: nit: this sample will become even simpler if we can remove the `async`s.

python/packages/core/agent_framework/_workflows/_functional.py
moonbox3: The docstring is describing the `@workflow`-decorated function, not a `@step`. The workflow function is what receives `RunContext` and calls `request_info()`. A `@step` doesn't receive context — so "If the function needs HITL" is referring to the workflow function.

python/packages/core/agent_framework/_workflows/_functional.py
moonbox3: Good catch — updated to "Inside the workflow".

python/packages/core/agent_framework/_workflows/_functional.py
moonbox3: Only the `@workflow` function receives `RunContext`. Steps don't — they're plain async functions with caching. If a step needs HITL input, the workflow calls `request_info()` and passes the result to the step as an argument.

python/packages/core/agent_framework/_workflows/_functional.py
moonbox3: By design, `request_info` lives at the workflow level — the workflow is the orchestrator that decides when to pause for input. Steps are meant to be self-contained units of work. If a step needs human input, the workflow calls `request_info()` first and passes the result to the step. This keeps steps simple and testable in isolation (no framework dependency).

python/packages/core/agent_framework/_workflows/_functional.py
moonbox3: Good catch — updated the docstring to mention both `WorkflowRunResult` and `ResponseStream` (when `stream=True`).

python/packages/core/agent_framework/_workflows/_functional.py
moonbox3: Done — removed the wrapper and made `add_event` the single method (it now appends directly). Internal callers updated from `ctx._add_event(...)` to `await ctx.add_event(...)`.

python/packages/core/agent_framework/_workflows/_functional.py
moonbox3: The `(False, None)` tuple means "no response found for this request_id." It's using the `(found, value)` pattern — `False` signals absence, `None` is just the placeholder. When a response exists, it returns `(True, actual_value)`.

python/packages/core/agent_framework/_workflows/_functional.py
moonbox3: Current approach reuses `executor_invoked`/`executor_completed` for cached replays to keep event consumers simple — they don't need to handle a new event type. The replay is transparent from the event stream's perspective. If observability tools later need to distinguish cache hits from live execution, we can add a metadata field to the existing event rather than a new event type.

python/packages/core/agent_framework/_workflows/_functional.py
moonbox3: Yes, within a single `.run()` call all steps share the same `RunContext`. To force a full rerun (ignore all caches), just call `.run(message)` without `checkpoint_id` or `responses` — it creates a fresh `RunContext` with an empty step cache.

python/packages/core/agent_framework/_workflows/_functional.py
moonbox3: We could, but it adds complexity for limited benefit — the workflow itself is async, so wrapping a sync function in `asyncio.to_thread()` or similar is straightforward for users. For now, `@step` requires async. We can revisit if users ask for it.

python/packages/core/agent_framework/_workflows/_functional.py
moonbox3: Agree this is worth considering. The `run()` overload pattern and `_finalize_events` logic are similar between `FunctionalWorkflow` and graph `Workflow`. Worth a follow-up refactoring PR to extract a shared base, but out of scope for this one.

python/packages/core/agent_framework/_workflows/_functional.py
moonbox3: Currently events are collected during execution and yielded after. Within a single function execution there's no natural suspension point to stream intermediate events (unlike graph workflows where each superstep boundary is a yield point). However, `@step` events are collected incrementally, and the async generator yields them as they're produced. For true mid-function streaming, users can use `ctx.add_event()` — though these are still batched by the current implementation. This is a known limitation we can improve in a follow-up.

python/packages/core/agent_framework/_workflows/_functional.py
moonbox3: The cache key is `(step_name, call_index)` and relies on the documented determinism requirement — if the workflow function is deterministic w.r.t. step results, the arguments will always match on replay. Adding argument comparison would be defensive but expensive (deep equality on arbitrary objects) and could give false negatives for objects that don't implement `__eq__`.

moonbox3: This is by design and documented: workflow functions must be deterministic w.r.t. step results. If a non-`@step` function is nondeterministic and a cached `@step` depends on it, results can diverge. The fix is to use `@step` on any function whose output affects downstream cached steps. This tradeoff keeps the simple case simple.

python/samples/01-get-started/06_functional_workflow_basics.py
moonbox3: Same as the sync support discussion — supporting sync functions would simplify the getting-started sample, but `@workflow` is inherently async (it runs an async event loop internally). We can consider sync support as a follow-up.

moonbox3:
Btw, @eavanvalkenburg and @TaoChenOSU I think it would be best to stick this functional API in to its own package. We want to get some more signal around the APIs and use of it before we deem it "GA worthy," IMO.

python/samples/01-get-started/05_first_functional_workflow.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+async def reverse_text(text: str) -> str:
+    """Reverse the string."""
+    return text[::-1]
+
+
+# <create_workflow>
+@workflow
+async def text_pipeline(text: str) -> str:
```
eavanvalkenburg: for me pipeline has different annotations, the whole idea is that this is as flexible as you want it to be, since it's just code!
```suggestion
async def text_workflow(text: str) -> str:
```

python/samples/01-get-started/06_functional_workflow_with_agents.py
```diff
@@ -0,0 +1,50 @@
...
+
+reviewer = Agent(
+    name="ReviewerAgent",
+    instructions="Review the given poem in one sentence. Is it good?",
+    client=AzureOpenAIChatClient(),
+)
+
+
+@workflow
+async def poem_pipeline(topic: str) -> str:
```
eavanvalkenburg:
```suggestion
async def poem_workflow(topic: str) -> str:
```

python/samples/01-get-started/06_functional_workflow_with_agents.py
eavanvalkenburg, line 1: I think maybe we should have this first, because the 05 just makes me think: `why would I need all this stuff for something so simple?!`

python/samples/01-get-started/06_functional_workflow_with_agents.py
```diff
@@ -0,0 +1,50 @@
...
+    """Write a poem, then review it."""
+    poem = (await writer.run(f"Write a poem about: {topic}")).text
```
eavanvalkenburg, line 37: would be good to have a `step` in this workflow as well

python/samples/03-workflows/functional/agent_integration.py
```diff
@@ -0,0 +1,114 @@
...
+# Create agents
+# ---------------------------------------------------------------------------
+
+classifier_agent = Agent(
+    name="ClassifierAgent",
+    instructions=(
+        "Classify documents into one category: Technical, Legal, Marketing, or Scientific. "
+        "Reply with only the category name."
+    ),
+    client=AzureOpenAIChatClient(),
```
eavanvalkenburg: could we use `Responses` with Foundry for these?

python/samples/03-workflows/functional/hitl_review.py
```diff
@@ -0,0 +1,79 @@
...
+
+from agent_framework import RunContext, WorkflowRunState, step, workflow
+
+
+# @step saves the result. When the workflow resumes after the HITL pause,
+# this returns its saved result instead of running the expensive operation again.
+@step
+async def write_draft(topic: str) -> str:
+    """Simulate writing a draft — expensive, shouldn't re-run on resume."""
+    print(f"  write_draft executing for '{topic}'")
```
eavanvalkenburg, line 31: maybe add in a commented out `Agent.run` here to demonstrate the concept a bit more clearly

python/samples/03-workflows/README.md
```diff
@@ -30,6 +30,19 @@ Once comfortable with these, explore the rest of the samples below.
...
+
+| Sample | File | Concepts |
+|---|---|---|
+| Basic Pipeline | [functional/basic_pipeline.py](./functional/basic_pipeline.py) | Sequential steps as plain async functions |
+| Basic Streaming Pipeline | [functional/basic_streaming_pipeline.py](./functional/basic_streaming_pipeline.py) | Stream workflow events in real time with `run(stream=True)` |
+| Parallel Pipeline | [functional/parallel_pipeline.py](./functional/parallel_pipeline.py) | Fan-out/fan-in with `asyncio.gather` |
+| Steps and Checkpointing | [functional/steps_and_checkpointing.py](./functional/steps_and_checkpointing.py) | `@step` decorator for per-step checkpointing and observability |
+| Human-in-the-Loop Review | [functional/hitl_review.py](./functional/hitl_review.py) | HITL with `ctx.request_info()` and replay |
+| Agent Integration | [functional/agent_integration.py](./functional/agent_integration.py) | Calling agents inside workflow steps |
+
```
eavanvalkenburg, line 46: would it make sense to showcase a simple `naive` group chat with these workflows, I think that would be a good one to demonstrate the differences between this API and the graph and steer people to try this first

python/samples/03-workflows/functional/agent_integration.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+async def main():
+    # Simple version — agents called inline
+    result = await simple_pipeline.run("This is a technical document about machine learning...")
+    print(result.get_outputs()[0])
+
+    # .as_agent() wraps the workflow so it can be used anywhere an agent
+    # is expected — for example, as a node in a graph workflow.
+    agent = cached_pipeline.as_agent(name="doc_processor")
```
eavanvalkenburg: I would leave this out, to make sure this sample focuses on the very important conceptual difference between calling agents directly or wrapping in a step, I think that is super important so let's make sure that that is very clear, maybe also move the two `pipeline` right next to each other so you quickly see the difference

python/samples/03-workflows/functional/basic_pipeline.py
moonbox3: Fixed in `49fd9303`.

python/packages/core/agent_framework/_workflows/_functional.py
moonbox3: Fixed in `49fd9303`.

python/samples/01-get-started/05_first_functional_workflow.py
moonbox3: Fixed in `49fd9303`.

python/samples/01-get-started/06_functional_workflow_with_agents.py
moonbox3: Fixed in `49fd9303`.

python/samples/01-get-started/06_functional_workflow_with_agents.py
moonbox3: Fixed in `49fd9303`.

python/samples/01-get-started/06_functional_workflow_with_agents.py
moonbox3: Fixed in `49fd9303`.

python/samples/03-workflows/functional/agent_integration.py
moonbox3: Fixed in `49fd9303`.

python/samples/03-workflows/functional/hitl_review.py
moonbox3: Fixed in `49fd9303`.

python/samples/03-workflows/README.md
moonbox3: Fixed in `49fd9303`.

python/samples/03-workflows/functional/agent_integration.py
moonbox3: Fixed in `49fd9303`.

python/packages/core/agent_framework/_workflows/_functional.py
TaoChenOSU: How will a step let the workflow know that it needs human input?

python/packages/core/agent_framework/_workflows/_functional.py
TaoChenOSU: If `None` is not a valid response, we can simply return `None` in the absence of the key.

python/packages/core/agent_framework/_workflows/_functional.py
TaoChenOSU: How will the client distinguish between a replay and an execution with the same events?

TaoChenOSU:
> Btw, @eavanvalkenburg and @TaoChenOSU I think it would be best to stick this functional API in to its own package. We want to get some more signal around the APIs and use of it before we deem it "GA worthy," IMO.

We can use the experimental flag now :)

python/packages/core/agent_framework/_workflows/_functional.py
moonbox3: Fixed in `2392a0d3`.

python/packages/core/agent_framework/_workflows/_functional.py
moonbox3: Fixed in `2392a0d3`.

python/packages/core/agent_framework/_workflows/_functional.py
moonbox3: Fixed in `2392a0d3`.

python/packages/core/agent_framework/_workflows/_functional.py
```diff
@@ -0,0 +1,1217 @@
...
+        self.response_type = response_type
+        super().__init__(f"Workflow interrupted by request_info (request_id={request_id})")
+
+
+# ---------------------------------------------------------------------------
+# RunContext
+# ---------------------------------------------------------------------------
+
+
+class RunContext:
```
eavanvalkenburg, line 107: since this will be exposed through `from agent_framework import RunContext` I wonder if this is clear enough about what it is used for.

python/samples/01-get-started/05_functional_workflow_with_agents.py
```diff
@@ -0,0 +1,70 @@
...
+
+reviewer = Agent(
+    name="ReviewerAgent",
+    instructions="Review the given poem in one sentence. Is it good?",
+    client=client,
+)
+# </create_agents>
+
+
+# @step caches the result — the agent call won't re-execute on resume.
```
eavanvalkenburg: I love the distinction between running agent with and without step, I do wonder if this is the place to showcase that...

python/samples/01-get-started/05_functional_workflow_with_agents.py
```diff
@@ -0,0 +1,70 @@
...
+    """Write a poem, then review it."""
+    poem = await write_poem(topic)
+    review = (await reviewer.run(f"Review this poem: {poem}")).text
+    return f"Poem:\n{poem}\n\nReview: {review}"
+# </create_workflow>
+
+
+async def main() -> None:
+    result = await poem_workflow.run("a cat learning to code")
+    print(result.get_outputs()[0])
```
eavanvalkenburg, line 45: would it make sense to be able to do:

```suggestion
    print(result.text)
```

python/samples/01-get-started/06_functional_workflow_basics.py
eavanvalkenburg, line 1: bit strange that the second function workflow sample is called `first`

python/samples/03-workflows/functional/basic_streaming_pipeline.py
```diff
@@ -0,0 +1,57 @@
...
+    """Transform raw data into a summary string."""
+    return f"[{data['status']}] {data['content']}"
+
+
+async def validate_result(summary: str) -> bool:
+    """Validate the transformed result."""
+    return len(summary) > 0 and "[200]" in summary
+
+
+# @workflow enables .run(stream=True), which returns a ResponseStream
```
eavanvalkenburg, line 29: I would expect this scenario to also allow `yield` (and because of a python blocker that means you cannot also have a return with a value in async) that would be quite powerful (you would be able to yield intermediate results), but would take some additional work to make it work transparently.

python/samples/03-workflows/functional/basic_streaming_pipeline.py
```diff
@@ -0,0 +1,57 @@
...
+    # alongside application events — filter by event.type to find what you need.
+    stream = data_pipeline.run("https://example.com/api/data", stream=True)
+    async for event in stream:
+        if event.type == "output":
+            print(f"Output: {event.data}")
+
+    # After iteration, get_final_response() returns the WorkflowRunResult
+    result = await stream.get_final_response()
+    print(f"Final state: {result.get_final_state()}")
+
```
eavanvalkenburg, line 54: would be good to get a sample of the output here.

python/samples/03-workflows/functional/hitl_review.py
```diff
@@ -0,0 +1,84 @@
...
+    """Revise the draft based on feedback."""
+    return f"Revised: {draft[:50]}... [Applied feedback: {feedback}]"
+
+
+@workflow
+async def review_pipeline(topic: str, ctx: RunContext) -> str:
+    """Write a draft, get human review, then revise."""
+    draft = await write_draft(topic)
+
+    # ctx.request_info() suspends the workflow here. The caller gets back
```
eavanvalkenburg, line 46: would a step with a `input` call also work (at least for demo's)?

python/samples/03-workflows/functional/hitl_review.py
```diff
@@ -0,0 +1,84 @@
...
+
+async def main():
+    # Phase 1: Run until the workflow pauses for human input
+    print("=== Phase 1: Initial run ===")
+    result1 = await review_pipeline.run("AI Safety")
+
+    # If request_info() was reached, the state is IDLE_WITH_PENDING_REQUESTS.
+    # If the workflow completed without hitting request_info(), it would be IDLE.
+    print(f"State: {result1.get_final_state()}")
+    assert result1.get_final_state() == WorkflowRunState.IDLE_WITH_PENDING_REQUESTS
```
eavanvalkenburg:
```suggestion
    print(f"State: {final_state := result1.get_final_state()}")
    assert final_state == WorkflowRunState.IDLE_WITH_PENDING_REQUESTS
```

python/samples/03-workflows/functional/naive_group_chat.py
```diff
@@ -0,0 +1,93 @@
...
+# ---------------------------------------------------------------------------
+# A naive group chat is just a loop — no special framework needed
+# ---------------------------------------------------------------------------
+
+
+@workflow
+async def group_chat(question: str) -> str:
+    """Round-robin group chat: expert answers, critic reviews, summarizer wraps up."""
+    participants = [expert, critic, summarizer]
+    conversation: list[str] = [f"User: {question}"]
```
eavanvalkenburg: can this be a `list[Message]` or even just setup a session? in combination with `default_option={"store": False}`

python/samples/03-workflows/functional/parallel_pipeline.py
```diff
@@ -0,0 +1,62 @@
...
+# @workflow wraps the orchestration logic so you get .run(), streaming,
+# and events. The functions it calls are plain Python — no decorators
+# needed just because they're inside a workflow.
+@workflow
+async def research_pipeline(topic: str) -> str:
+    """Fan-out to three research tasks, then synthesize results."""
+    # asyncio.gather runs all three concurrently — this is standard Python,
+    # not a framework concept. Use it the same way you would anywhere else.
+    web, papers, news = await asyncio.gather(
+        research_web(topic),
```
eavanvalkenburg, line 52: does this also work when these are steps?

eavanvalkenburg (commented):
I focused mostly on the samples to understand the API shape and it looks good, with some questions about how certain things work!

python/packages/core/agent_framework/_workflows/_functional.py
moonbox3: Fixed in `2392a0d3`.

python/packages/core/agent_framework/_workflows/_functional.py
moonbox3: Fixed in `2392a0d3`.

python/packages/core/agent_framework/_workflows/_functional.py
moonbox3: Fixed in `2392a0d3`.

python/packages/core/agent_framework/_workflows/_functional.py
moonbox3: Agreed — no `@property` remains on the class; `StepWrapper.name` and `FunctionalWorkflow.name` are now plain attributes assigned in `__init__`. (Already addressed in an earlier feedback commit on this branch.)

python/samples/03-workflows/functional/agent_integration.py
moonbox3: Addressed — see the revived form of this sample: `simple_pipeline` (agents called inline, no `@step`) and `cached_pipeline` (each agent call wrapped in `@step` for result caching) sit side-by-side so you can see both options. The module docstring explains why you would add `@step` (expensive LLM calls survive HITL resume / crash recovery) and why it's opt-in.

python/samples/01-get-started/05_functional_workflow_with_agents.py
moonbox3: `WorkflowRunResult.get_outputs()` returns `list[Any]` because a workflow can emit any number of outputs of any type, and this is the convention across `samples/03-workflows/_start-here`. Kept the sample as `result.get_outputs()[0]` for consistency. Adding a `.text` convenience accessor that does the right thing in the single-string-output case is a reasonable follow-up for the `WorkflowRunResult` API, but broader than this PR — happy to do it separately.

python/samples/03-workflows/functional/basic_streaming_pipeline.py
moonbox3: Agree this would be powerful — `yield` inside `@workflow` to emit intermediate values (mirroring `ResponseStream`) is a meaningful feature in its own right (needs runtime support for async-generator functions, signature inspection, and a decision on how `@step` interacts with generators). Tracking for a follow-up PR rather than expanding this one.

python/samples/03-workflows/functional/hitl_review.py
moonbox3: Yes — a `@step` calling blocking `input()` works fine in a terminal demo. The sample deliberately uses `ctx.request_info()` because (a) it is non-blocking / asyncio-native, and (b) the response arrives on the next `run(responses=...)` call so the workflow is resumable from a different process (the HITL-with-checkpoints scenario). `input()` would demo HITL but wouldn't demo what is distinctive about this API.

python/packages/core/agent_framework/_workflows/_functional.py
moonbox3: Rewrote the docstring to lead with purpose: "Opt-in handle for workflow-only features inside a `@workflow` function." The three reasons to declare a `RunContext` (HITL, custom events, state) each get a bullet, and the examples now show both the simple case (no ctx) and the HITL case side-by-side. c0093ac0

python/samples/01-get-started/06_functional_workflow_basics.py
moonbox3: Renamed to `06_functional_workflow_basics.py` — file, docstring heading, and both README tables updated. c0093ac0

python/samples/01-get-started/05_functional_workflow_with_agents.py
moonbox3: Agreed — 05 now calls the agents directly without `@step`, keeping the `01-get-started` sample minimal. The step-vs-no-step distinction is exclusive to `03-workflows/functional/agent_integration.py`, which shows `simple_pipeline` and `cached_pipeline` side-by-side with an explanation of when `@step` is worth it. c0093ac0

python/samples/03-workflows/functional/hitl_review.py
moonbox3: Applied with parens to avoid ruff's f-string format-spec ambiguity: `print(f"State: {(final_state := result1.get_final_state())}")`. c0093ac0

python/samples/03-workflows/functional/basic_streaming_pipeline.py
moonbox3: Added an expected-output docstring block at the end of the sample. c0093ac0

python/samples/03-workflows/functional/parallel_pipeline.py
moonbox3: Yes — added a tip in the sample: `@step` composes with `asyncio.gather`, so each branch is independently cached on HITL resume or checkpoint restore. c0093ac0

python/samples/03-workflows/functional/naive_group_chat.py
moonbox3: Reworked to thread `list[Message]` between turns instead of stringifying: `conversation: list[Message] = [Message("user", [question])]`, each `agent.run(conversation)` extends it with `response.messages`. The final output joins `author_name`/role + text. Kept it session-free per Eduard's suggestion — easier to reason about for a "naive" round-robin. c0093ac0

moonbox3:
Revived this PR:

- Rebased onto `upstream/main` (c0093ac0, linear history — dropped the in-branch merge commits, resolved conflicts in `agent_framework/__init__.py`, `samples/README.md`, and `samples/01-get-started/README.md` without stomping either side)
- Addressed the still-open review threads from @eavanvalkenburg in a single commit on top (c0093ac0):
  - `RunContext` docstring rewritten to lead with purpose
  - `06_first_functional_workflow.py` → `06_functional_workflow_basics.py`
  - `05_functional_workflow_with_agents.py` simplified (step-vs-no-step contrast stays in `agent_integration.py`)
  - All functional samples moved to `FoundryChatClient` for consistency
  - `hitl_review.py` uses the walrus as suggested
  - `basic_streaming_pipeline.py` gets an expected-output block
  - `parallel_pipeline.py` clarifies `@step` composes with `asyncio.gather`
  - `naive_group_chat.py` threads `list[Message]` between turns
  - Reply posted on each thread with the resolution

Verified green: 2319 core tests pass (workflow suite: 748 pass, 1 skip, 2 xfail — no regressions).

moonbox3:
Follow-up: ran `/ultrareview` on the branch and landed a fix for every finding (commits 0b58ea8 + c32a18e).

### Fixed (10/10)
- **bug_001** — `ctx.request_info()` without an explicit `request_id` now derives a deterministic `auto::<index>` id from a per-context call counter, so HITL resume works on the documented default path (previously a fresh uuid was regenerated on every replay, making resume impossible).
- **bug_002** — `StepWrapper.__call__` no longer deepcopies args on the cache-hit branch. The copy is live-path-only, wrapped in try/except so non-deepcopyable args (locks, sockets, sessions) don't break replay.
- **bug_007** — `_set_responses` and the cache-hit branch of `request_info` now prune resolved ids from `_pending_requests`, so checkpoints stop accumulating already-answered requests and the final checkpoint reports no pending events after completion.
- **bug_008** — `_compute_signature_hash` now mixes `__code__.co_code` and `co_names` into the hash, so body changes invalidate old checkpoints even when steps are referenced via module/class attributes that `_discover_step_names` can't see statically.
- **bug_010** — `run()` docstring says "at least one of" and notes `responses + checkpoint_id` is allowed (matching the validator).
- **bug_013** — `FunctionalWorkflowAgent` now surfaces `request_info` events as `FunctionApprovalRequestContent` (mirroring graph `WorkflowAgent`), threads `responses=` and `checkpoint_id=` through to the workflow, and exposes `pending_requests`. HITL workflows are now usable via `.as_agent()`.
- **bug_014** — singleton replay state (`_last_message`, `_last_step_cache`, `_last_pending_request_ids`) is cleared on clean completion; `run()` rejects `responses=` keys that don't intersect the currently-pending set instead of silently replaying against stale data.
- **bug_015** — `FunctionalWorkflow.as_agent` signature matches graph `Workflow.as_agent` (`name`, `description`, `context_providers`, `**kwargs`). Agent stores the overrides.
- **bug_017** — `RunContext.set_state` rejects underscore-prefixed keys with a clear `ValueError`; docstring documents the reserved prefix so framework bookkeeping (`_step_cache`, `_original_message`) can't silently clobber user state.
- **merged_bug_003** — workflow arity is validated at decoration time (>1 non-ctx parameter → `ValueError`); passing a non-None `message` to a ctx-only workflow raises instead of silently dropping the message.

### Test coverage
- +18 targeted regression tests, one per bug.
- Full workflow suite: 766 passed, 1 skipped, 2 xfailed (was 748 before this pass).
- Full core suite: 2338 passed, 19 skipped, 2 xfailed.

python/packages/core/agent_framework/_workflows/_functional.py
```diff
@@ -0,0 +1,1493 @@
...
+
+@overload
+def step(func: Callable[..., Awaitable[R]]) -> StepWrapper[R]: ...
+
+
+@overload
+def step(*, name: str | None = None) -> Callable[[Callable[..., Awaitable[R]]], StepWrapper[R]]: ...
+
+
+def step(
```
eavanvalkenburg, line 568: we should add the `@experimental` markers on these ones

eavanvalkenburg (commented):
I think this looks great, let's get the experimental tag on it, and put it out there!

eavanvalkenburg (approved):

SergeyMenshykh (approved):

PR #4274, opened 2026-02-25
Python: Workflows background run

TaoChenOSU:
### Motivation and Context

Currently, we have a very limited way of handling workflow runs and responding to events. Users have to wait until a workflow converges to process events, such as requests. 

### Description

1. Create a run mode where clients can start a workflow run in the background and poll events as they desire.
2. Allow clients to respond to requests while the workflow is still running in the background.

### Contribution Checklist

### DIFF
--- a/python/packages/core/agent_framework/_workflows/_background_run.py
+++ b/python/packages/core/agent_framework/_workflows/_background_run.py
@@ -60,9 +60,18 @@ class BackgroundRunHandle:
             list if no events are available.
         """
         events: list[WorkflowEvent[Any]] = []
+        # Use get_nowait() in a loop to drain all available events.
+        # This is safe from race conditions because asyncio uses cooperative
+        # multitasking: since we never await inside this loop, no other
+        # coroutine (including the background task producing events) can
+        # execute until we finish. This guarantees we get a consistent
+        # snapshot of all events queued at the moment poll() was called.
         while True:
             try:
                 events.append(self._event_queue.get_nowait())
             except asyncio.QueueEmpty:
+                # Queue is empty — we've drained all currently available events.
+                # Any events added after this point will be picked up in the
+                # next poll() call.
                 break
         return events
--- a/python/packages/core/agent_framework/_workflows/_background_run.py
+++ b/python/packages/core/agent_framework/_workflows/_background_run.py
@@ -4,9 +4,11 @@ from __future__ import annotations
 
 import asyncio
 import logging
+from collections.abc import Awaitable, Callable
 from typing import Any
 
 from ._events import WorkflowEvent
+from ._runner_context import RunnerContext
 
 logger = logging.getLogger(__name__)
 
@@ -19,6 +21,10 @@ class BackgroundRunHandle:
     are buffered in an internal queue until the caller drains them via
     :meth:`poll`.
 
+    Responses to ``request_info`` events can be sent while the workflow is
+    still running via :meth:`respond`, enabling hot-path / cold-path
+    parallelism.
+
     Example:
         .. code-block:: python
 
@@ -26,18 +32,30 @@ class BackgroundRunHandle:
             while not handle.is_idle:
                 events = await handle.poll()
                 for event in events:
-                    print(event)
+                    if event.type == "request_info":
+                        await handle.respond({event.request_id: answer})
     """
 
-    def __init__(self, task: asyncio.Task[None], event_queue: asyncio.Queue[WorkflowEvent[Any]]) -> None:
+    def __init__(
+        self,
+        task: asyncio.Task[None],
+        event_queue: asyncio.Queue[WorkflowEvent[Any]],
+        runner_context: RunnerContext,
+        resume_fn: Callable[[], Awaitable[asyncio.Task[None]]],
+    ) -> None:
         """Initialize the background run handle.
 
         Args:
             task: The asyncio task running the workflow.
             event_queue: The queue where workflow events are buffered.
+            runner_context: The runner context for injecting responses.
+            resume_fn: Callback that creates and returns a new producer task
+                to resume the workflow after it has converged.
         """
         self._task = task
         self._event_queue = event_queue
+        self._runner_context = runner_context
+        self._resume_fn = resume_fn
 
     @property
     def is_idle(self) -> bool:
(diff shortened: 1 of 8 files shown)


python/packages/core/agent_framework/_workflows/_background_run.py
```diff
@@ -0,0 +1,116 @@
...
+from collections.abc import Awaitable, Callable
+from typing import Any
+
+from ._events import WorkflowEvent
+from ._runner_context import RunnerContext
+
+logger = logging.getLogger(__name__)
+
+
+class BackgroundRunHandle:
```
moonbox3, line 16: It feels like this class is essentially a wrapper around asyncio primitives: create_task and an asyncio.Queue. Why do we need to wrap these well-known constructs?

python/packages/core/agent_framework/_workflows/_workflow.py
```diff
@@ -537,6 +538,101 @@ def run(
             return response_stream
         return response_stream.get_final_response()
 
+    def run_in_background(
```
moonbox3, line 541: Do we genuinely need a polling-based consumption pattern? Are we getting feedback that this is missing today?

We're now pushing more concerns onto the caller. Every consumer has to:

1. Write the poll loop
2. Pick a sleep interval (and get it "wrong", too slow adds latency, too fast wastes cycles)
3. Route events by type manually
4. Track which request IDs map to which responses
5. Remember to drain after idle
6. Handle the resume-after-idle edge case

moonbox3 (commented):
Some initial thoughts.

python/packages/core/agent_framework/_workflows/_background_run.py
eavanvalkenburg: and this also adds another concept that people have to learn, I do not think this is needed

python/samples/03-workflows/background-run/respond_while_running.py
```diff
@@ -0,0 +1,173 @@
...
+        .add_edge(hot_path, hot_path)  # Self-loop for the hot path
+        .build()
+    )
+
+    handle = workflow.run_in_background(TaskInput(value=0))
+
+    # Single poll loop: process all events and respond to requests inline.
+    # The workflow continues running in the background while we process events.
+    outputs: list[str] = []
+    while not handle.is_idle:
```
eavanvalkenburg, line 136: I see almost no value in this, because it makes people have to understand this whole idle, poll, etc. Also a user having to write `asyncio.sleep` does not seem right, a much simpler pattern to solve for this scenario is to use something like a callback for response required, that's a well known concept in Python and doesn't require as much complexity as this. And when I compare these samples to existing sample that just use streaming then that is enough, that already allows you to do other stuff in the meantime, and if you really need it, a user can call that in a separate thread and then you ahve this with Python primitives instead of another new thing.

python/packages/core/agent_framework/_workflows/_workflow.py
eavanvalkenburg: I very much agree, this is not needed and leads to un-pythonic code

python/samples/03-workflows/background-run/respond_while_running.py
TaoChenOSU: Why is polling and other things not Pythonic? This is a well-known concept in Programming. OpenAI also has a similar primitive: https://developers.openai.com/api/docs/guides/background.

Using callbacks is a simpler approach but it's very limited and doesn't scale well. We had callbacks back in SK and it didn't work really well. For example, it will be difficult to integrate callbacks with a UI because the UI loop is driven by the callback and users can only handle one callback at a time. And if you don't want pre-determined actions, you will have to emit another thing from the callback and wait for user input. 

With a background run and events, we are maintaining the original event systems we have in Workflows, which reduces the concepts that people have to learn, given that they are already familiar with polling, which again is a well-known concept.

python/packages/core/agent_framework/_workflows/_workflow.py
TaoChenOSU: Why is this un-pythonic?

python/packages/core/agent_framework/_workflows/_background_run.py
TaoChenOSU: I don't agree. We are doing a lot more than just wrapping a workflow in a background task. We are handling events, requests, errors for the users. And the concept is well-known. Here is an equivalent concept from OpenAI: https://developers.openai.com/api/docs/guides/background

moonbox3: We support `background` for agents as a keyword arg right now.

python/samples/03-workflows/background-run/respond_while_running.py
eavanvalkenburg: So Polling in most other cases, means you actually exit whatever you're doing and then every so often check in to see if it is done, not building this `handle.is_idle` loop, usually a `while True` loop suffices, this is also the pattern we have now with background responses for certain api's (openai responses and a2a I think), so adding this `handle` thing, is the new concept that people have to learn, if we want to do a background run, then it needs to follow such a pattern, where you start it, you get a `id` of some sort back (continuation token in our case) and then you call `run` again to see the status. That is also the pattern that openai has on the link that you showed, not with some other thing that needs to be managed. Another concern is can you actually serialize this handler, stop the app, then restart and start polling again, probably not, and we would probably send people to checkpoints if that is a requirement. 

Callbacks are a somewhat different approach, but I think the guidance we should give, is that if you can respond directly, then use a callback, if you cannot respond quickly then do not use background, because you wont know when you'll be getting back, and then even a hot path might end up idling for long stretches, which would indicate to that the process you are building should be split into two anyway.

TaoChenOSU: You don't have to loop on `handle.is_idle`. That's just the sample. You can totally go do something else in your script and come back later to poll all the events that get produced during that period. You can also do a `while True` until an output event is produced. The `OpenAI` response is also a live handle to poll the status `while resp.status in {"queued", "in_progress"}:`.

Given that we are an SDK, returning an ID isn't a good pattern because we can give users direct access to the background task. If users are creating a service around this, they can map generated IDs to handles and create an API to allow clients to poll using IDs.


> Another concern is can you actually serialize this handler, stop the app, then restart and start polling again, probably not, and we would probably send people to checkpoints if that is a requirement.

We already have checkpoints.

> if you can respond directly, then use a callback, if you cannot respond quickly then do not use background, because you wont know when you'll be getting back, and then even a hot path might end up idling for long stretches, which would indicate to that the process you are building should be split into two anyway.

This is the hypothetical scenario we are trying to address. If we give this guidance, there is no need for callback either. The current event system will suffice.

TaoChenOSU:
Closing for now

TaoChenOSU:
Potential solution for #3539

PR #4286, opened 2026-02-26
Python: Fix #3970: Python: [Bug]: `AgentResponse.value` is `None` when streaming workflow

### LINKED ISSUE #3970: Python: [Bug]: `AgentResponse.value` is `None` when streaming workflow
### Description

If a workflow has an agent executor which has `response_format` set, the response format value is set to `None` if the workflow outputs are streamed. If the workflow is not streamed, the value is set as expected.

### Code Sample

```python
from typing import Never

from agent_framework import (
    AgentExecutorResponse,
    Workflow,
    WorkflowBuilder,
    WorkflowContext,
    executor,
)
from agent_framework.azure import AzureOpenAIResponsesClient
from azure.identity import DefaultAzureCredential
from pydantic import BaseModel


class Greeting(BaseModel):
    greeting: str


@executor
async def print_executor(message: AgentExecutorResponse, ctx: WorkflowContext[Never, str]):
    await ctx.yield_output(f"Value: {message.agent_response.value}, Text: {message.agent_response.text}")


def build_workflow() -> Workflow:
    agent = AzureOpenAIResponsesClient(credential=DefaultAzureCredential()).as_agent(
        default_options={
            "temperature": 0,
            "response_format": Greeting,
        }
    )
    return (
        WorkflowBuilder(
            start_executor=agent,
            output_executors=[print_executor],
        )
        .add_edge(agent, print_executor)
        .build()
    )


async def main() -> None:
    print("--------- Non-Streaming ---------")

    workflow = build_workflow()
    result = await workflow.run("Hello")
    for output in result.get_outputs():
        print(output)

    print("--------- Streaming ---------")

    workflow = build_workflow()
    async for output in workflow.run("Hello", stream=True):
        if output.type == "output":
            print(output.data)


if __name__ == "__main__":
    import asyncio

    asyncio.run(main())
```

### Output

```console
--------- Non-Streaming ---------
Value: greeting='Hello! How can I assist you today?', Text: {"greeting":"Hello! How can I assist you today?"}
--------- Streaming ---------
Value: None, Text: {"greeting":"Hello! How can I assist you today?"}
```

### Package Versions

1.0.0b260212

### Python Version

3.13

### Additional Context

_No response_

moonbox3:
Automated fix for #3970.

Verification passed after 0 retries.

### DIFF
--- a/python/packages/core/agent_framework/_agents.py
+++ b/python/packages/core/agent_framework/_agents.py
@@ -935,6 +935,13 @@ class RawAgent(BaseAgent, Generic[OptionsCoT]):  # type: ignore[misc]
                 session.service_session_id = conv_id
             return update
 
+        def _stream_finalizer(updates: Sequence[AgentResponseUpdate]) -> AgentResponse:
+            # Read response_format from merged chat_options (includes default_options)
+            # so that response_format set via the constructor is respected.
+            ctx = ctx_holder["ctx"]
+            response_format = ctx["chat_options"].get("response_format") if ctx else None
+            return self._finalize_response_updates(updates, response_format=response_format)
+
         return (
             ResponseStream
             .from_awaitable(_get_stream())
@@ -943,9 +950,7 @@ class RawAgent(BaseAgent, Generic[OptionsCoT]):  # type: ignore[misc]
                     map_chat_to_agent_update,
                     agent_name=self.name,
                 ),
-                finalizer=partial(
-                    self._finalize_response_updates, response_format=options.get("response_format") if options else None
-                ),
+                finalizer=_stream_finalizer,
             )
             .with_transform_hook(_propagate_conversation_id)
             .with_result_hook(_post_hook)
--- a/python/packages/core/tests/core/test_agents.py
+++ b/python/packages/core/tests/core/test_agents.py
@@ -97,6 +97,32 @@ async def test_chat_client_agent_run_streaming(client: SupportsChatGetResponse) 
     assert result.text == "test streaming response another update"
 
 
+async def test_chat_client_agent_run_streaming_response_format_via_default_options(
+    client: SupportsChatGetResponse,
+) -> None:
+    """response_format set in default_options must be available on the streamed AgentResponse.value (issue #3970)."""
+    from pydantic import BaseModel
+
+    class Greeting(BaseModel):
+        greeting: str
+
+    json_text = '{"greeting":"hello"}'
+    client.streaming_responses = [  # type: ignore[attr-defined]
+        [ChatResponseUpdate(contents=[Content.from_text(json_text)], role="assistant")],
+    ]
+
+    agent = Agent(client=client, default_options={"response_format": Greeting})
+    stream = agent.run("Hello", stream=True)
+    async for _ in stream:
+        pass
+    result = await stream.get_final_response()
+
+    assert result.text == json_text
+    assert result.value is not None
+    assert isinstance(result.value, Greeting)
+    assert result.value.greeting == "hello"
+
+
 async def test_chat_client_agent_create_session(client: SupportsChatGetResponse) -> None:
     agent = Agent(client=client)
     session = agent.create_session()
--- a/python/packages/core/tests/core/test_function_invocation_logic.py
+++ b/python/packages/core/tests/core/test_function_invocation_logic.py
@@ -937,8 +937,7 @@ async def test_max_iterations_no_orphaned_function_calls(chat_client_base: Suppo
 
     orphaned_calls = all_call_ids - all_result_ids
     assert not orphaned_calls, (
-        f"Response contains orphaned FunctionCallContent without matching "
-        f"FunctionResultContent: {orphaned_calls}."
+        f"Response contains orphaned FunctionCallContent without matching FunctionResultContent: {orphaned_calls}."
     )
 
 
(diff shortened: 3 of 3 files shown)


python/packages/core/agent_framework/_agents.py
```diff
@@ -935,6 +935,13 @@ def _propagate_conversation_id(update: AgentResponseUpdate) -> AgentResponseUpda
                 session.service_session_id = conv_id
             return update
 
+        def _stream_finalizer(updates: Sequence[AgentResponseUpdate]) -> AgentResponse:
+            # Read response_format from merged chat_options (includes default_options)
+            # so that response_format set via the constructor is respected.
+            ctx = ctx_holder["ctx"]
```
moonbox3, line 941: Minor robustness note: if `ctx` is truthy but `chat_options` is unexpectedly absent, `ctx["chat_options"]` would raise `KeyError`. Using `ctx.get("chat_options", {}).get("response_format")` would be safer, though this is likely fine if the context schema guarantees the key.

moonbox3, line 941: Nit (non-blocking): `ctx["chat_options"]` will raise `KeyError` if the key is absent. Using `ctx.get("chat_options", {}).get("response_format")` would be more defensive, though the key is expected to always be present in practice.

python/packages/core/agent_framework/_agents.py
```diff
@@ -935,6 +935,13 @@ def _propagate_conversation_id(update: AgentResponseUpdate) -> AgentResponseUpda
                 session.service_session_id = conv_id
             return update
 
+        def _stream_finalizer(updates: Sequence[AgentResponseUpdate]) -> AgentResponse:
+            # Read response_format from merged chat_options (includes default_options)
+            # so that response_format set via the constructor is respected.
```
moonbox3, line 940: This defensive `if ctx else None` check is not exercised by any test. Consider whether ctx can realistically be None here; if not, dropping the guard (or adding a test that triggers it) would improve clarity.

python/packages/core/tests/core/test_agents.py
```diff
@@ -97,6 +97,32 @@ async def test_chat_client_agent_run_streaming(client: SupportsChatGetResponse)
...
+    json_text = '{"greeting":"hello"}'
+    client.streaming_responses = [  # type: ignore[attr-defined]
+        [ChatResponseUpdate(contents=[Content.from_text(json_text)], role="assistant")],
+    ]
+
+    agent = Agent(client=client, default_options={"response_format": Greeting})
+    stream = agent.run("Hello", stream=True)
+    async for _ in stream:
+        pass
+    result = await stream.get_final_response()
```
moonbox3, line 118: Good: the test covers the exact bug scenario (default_options response_format in streaming). Consider adding a sibling test that passes response_format directly in run() options to ensure the refactored finalizer handles both paths.

moonbox3 (commented):
## Automated Code Review

**Reviewers:** 3 | **Confidence:** 88%

### ✓ Correctness
> This diff fixes a bug where `response_format` set via `default_options` in the constructor was not being respected during streaming responses. The fix correctly changes the finalizer to read `response_format` from the merged `chat_options` in the context (which includes `default_options`) rather than from the per-call `options` parameter. The closure captures `ctx_holder` and reads it at finalization time, which is appropriate since the context will be populated by then. The null check on `ctx` is present. The test correctly validates the fix. The string-join changes in `test_function_invocation_logic.py` are cosmetic and harmless.

### ✓ Security Reliability
> This diff fixes a bug where `response_format` set via `default_options` was ignored during streaming because the finalizer only read from the per-call `options`. The fix captures the merged `chat_options` from the context holder instead. The change is low-risk: it reads from an already-populated context dictionary and the closure over `ctx_holder` is safe since it's defined in the same scope. The test changes are straightforward and the string reformatting in the other test file is cosmetic. No security or reliability issues found.

### ✓ Test Coverage
> The new test directly validates the bug fix (issue #3970) where response_format set via default_options was ignored during streaming. Assertions are meaningful — checking text, parsed value type, and field content against a real Pydantic model. However, the diff lacks a complementary test verifying that response_format passed directly in per-call options still works after the refactor (the finalizer now reads from merged chat_options instead of raw options). There is also no test covering the precedence behavior when response_format appears in both default_options and per-call options.

### Suggestions
- Consider using `ctx.get("chat_options", {}).get("response_format")` instead of `ctx["chat_options"].get("response_format")` to guard against a missing `chat_options` key, though this may be unnecessary if the context contract guarantees its presence.
- In `_stream_finalizer`, if `ctx_holder["ctx"]` is unexpectedly missing the `"chat_options"` key (e.g., due to a future refactor), a `KeyError` would propagate uncaught. Consider using `ctx.get("chat_options", {}).get("response_format")` for defensive access, though this is minor given the controlled internal usage.
- Add a test for response_format passed directly via per-call options in streaming mode (e.g., agent.run('Hello', stream=True, response_format=Greeting)) to confirm the refactored finalizer still handles this path correctly.
- Add a test where response_format is specified in both default_options and per-call options to verify correct merge/override precedence in streaming mode.
- Consider a test where ctx_holder['ctx'] is None (the defensive branch on line 940) — though this is unlikely in practice, it would exercise the fallback path.

---
*Automated review by maf-dashboard agent*

moonbox3:
Currently testing.. closing PR.

PR #4326, opened 2026-02-26
Python: Add OpenTelemetry instrumentation to ClaudeAgent (#4278)

### LINKED ISSUE #4278: Python: [Bug]: ClaudeAgent missing telemetry -  inherits from BaseAgent instead of Agent
### Description

ClaudeAgent inherits directly from BaseAgent, bypassing AgentTelemetryLayer, AgentMiddlewareLayer, and RawAgent. As a result, enable_instrumentation() has no effect on ClaudeAgent — no OpenTelemetry spans are emitted when calling agent.run(), and agents are invisible in Azure Application Insights Agents (Preview) view.

### Code Sample

```markdown
from agent_framework_claude import ClaudeAgent

# Check 1: issubclass is False
print(issubclass(ClaudeAgent, Agent))  # False ← should be True

# Check 2: Compare MRO chains
print([c.__name__ for c in ClaudeAgent.__mro__])
# ['ClaudeAgent', 'BaseAgent', 'SerializationMixin', 'Generic', 'object']

print([c.__name__ for c in Agent.__mro__])
# ['Agent', 'AgentTelemetryLayer', 'AgentMiddlewareLayer', 'RawAgent', 'BaseAgent', 'SerializationMixin', 'Generic', 'object']
```

```markdown

```

### Package Versions

Package: agent-framework-claude v1.0.0b260212 Dependency: agent-framework-core v1.0.0b260212 Python: 3.12.10 (Windows ARM64)

### Python Version

_No response_

### Additional Context

Missing Layers
ClaudeAgent is missing 4 layers that Agent includes:

AgentTelemetryLayer — emits gen_ai.invoke_agent spans, token/duration histograms
AgentMiddlewareLayer — middleware pipeline support
RawAgent — raw agent contract
Agent — the complete agent class

Expected Behavior:
ClaudeAgent should inherit from Agent (not BaseAgent), so that:

1. enable_instrumentation() automatically wraps ClaudeAgent.run() with telemetry
2. gen_ai.invoke_agent spans are emitted with gen_ai.agent.id and gen_ai.agent.name attributes
3. Agents appear in Azure Application Insights Agents (Preview) view
4. Token usage histograms and duration metrics are recorded

Actual Behavior:
enable_instrumentation() has no effect on ClaudeAgent
No gen_ai.* spans are emitted for ClaudeAgent invocations
Agents are invisible in App Insights Agents (Preview) view
Only custom spans manually created via get_tracer() appear

Impact
Any application using agent-framework-claude with Azure Application Insights observability (Option 3: Third party setup with configure_azure_monitor() + enable_instrumentation()) cannot see Claude agents in the Agents (Preview) monitoring view.

Suggested Fix
In agent_framework_claude/_agent.py, change:

class ClaudeAgent(BaseAgent):  # ← current

to:

class ClaudeAgent(Agent):  # ← fix (or at minimum RawAgent)

This aligns ClaudeAgent with how other agent implementations (e.g., OpenAI-based) inherit from the full Agent class.

amitmukh:
## Summary
- Fixes #4278: ClaudeAgent inherits from `BaseAgent` directly, bypassing `AgentTelemetryLayer`, so `enable_instrumentation()` has no effect
- Adds inline OpenTelemetry telemetry to `ClaudeAgent.run()` using the same observability helpers as `AgentTelemetryLayer`
- Covers both streaming and non-streaming paths with `invoke_agent` spans, duration tracking, exception capture, and sensitive data support

## Approach
Instead of changing the inheritance chain (which would cause MRO conflicts since `ClaudeAgent` defines its own `run()` and doesn't use a standard chat client), this PR integrates telemetry directly into `ClaudeAgent.run()` using the framework's existing observability helpers (`_get_span`, `_get_span_attributes`, `_capture_messages`, `_capture_response`, `capture_exception`). This is the safest approach with zero risk of breaking existing behavior.

## Changes
- `python/packages/claude/agent_framework_claude/_agent.py`: Add telemetry wrapping in `run()`, with two new private methods `_run_with_telemetry()` and `_run_with_telemetry_stream()`
- `python/packages/claude/tests/test_claude_agent.py`: Add 5 new tests covering span emission, disabled instrumentation, streaming spans, exception capture, and provider name verification

## Test plan
- [x] All 49 existing tests pass unchanged
- [x] 5 new telemetry tests pass (54 total)
- [ ] Verify `invoke_agent` span appears in App Insights when `enable_instrumentation()` is called
- [ ] Verify no spans are emitted when instrumentation is disabled

## Future Work: Tool-level telemetry granularity

This PR provides the outer `invoke_agent` span for ClaudeAgent. However, visibility into 
individual tool calls (Read, Write, Bash, etc.) and per-LLM-call spans that happen *inside* 
the Claude CLI subprocess is not possible at the MAF layer — these are opaque to the framework.

A related issue has been raised upstream: [anthropics/claude-agent-sdk-python#611](https://github.com/anthropics/claude-agent-sdk-python/issues/611) 
requesting OpenTelemetry instrumentation in the Claude Agent SDK itself.

Once Anthropic ships SDK-level telemetry, a follow-up PR can wire those events as child spans 
under the `invoke_agent` span created here. The trace context parent is already in place — 
the integration should be a small, targeted change once the Anthropic SDK defines its 
telemetry surface.

**Note:** Custom `@tool`-decorated functions already get tool-level spans today via 
`FunctionTool.invoke()` in core. Only Claude's built-in CLI tools require the upstream fix.


markwallace-microsoft:
<a href="https://github.com/microsoft/agent-framework/blob/ba06771dc260644b1b0a97e8c1eef8ac67dcfa4b/README.md">Python Test Coverage</a><details><summary>Python Test Coverage Report • </summary><table><tr><th>File</th><th>Stmts</th><th>Miss</th><th>Cover</th><th>Missing</th></tr><tbody><tr><td><b>TOTAL</b></td><td><b>22475</b></td><td><b>2778</b></td><td><b>87%</b></td><td>&nbsp;</td></tr></tbody></table><i>report-only-changed-files is enabled. No files were changed during this commit :)</i></details>

## Python Unit Test Overview
| 4559 | 25 :zzz: | 0 :x: | 0 :fire: | 1m 18s :stopwatch: |

python/packages/claude/agent_framework_claude/_agent.py
```diff
@@ -27,6 +29,19 @@
...
+    get_tracer,
+)
+from agent_framework.observability import (
+    _capture_messages as capture_messages,
+    _capture_response as capture_response,
+    _get_response_attributes as get_response_attributes,
+    _get_span as get_span,
+    _get_span_attributes as get_span_attributes,
+    capture_exception,
+)
```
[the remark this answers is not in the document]
amitmukh: Added a 4-line justification comment explaining why these internal helpers are needed (MRO conflict prevents using AgentTelemetryLayer) and noting they should be updated if public equivalents are exposed.

python/packages/claude/agent_framework_claude/_agent.py
```diff
@@ -620,9 +635,109 @@ def run(
...
+    ) -> ResponseStream[AgentResponseUpdate, AgentResponse[Any]]:
+        """Wrap a streaming run with OpenTelemetry tracing."""
+        operation = attributes.get(OtelAttr.OPERATION, "operation")
+        span_name = attributes.get(OtelAttr.AGENT_NAME, "unknown")
+        span = get_tracer().start_span(f"{operation} {span_name}")
+        span.set_attributes(attributes)
+
+        if OBSERVABILITY_SETTINGS.SENSITIVE_DATA_ENABLED and messages:
+            capture_messages(span=span, provider_name=provider_name, messages=messages)
+
```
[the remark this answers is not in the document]
amitmukh: Added system_instructions=self._default_options.get("system_prompt") to the capture_messages() call in _run_with_telemetry_stream().

python/packages/claude/agent_framework_claude/_agent.py
```diff
@@ -620,9 +635,109 @@ def run(
...
+        self,
+        result_stream: ResponseStream[AgentResponseUpdate, AgentResponse[Any]],
+        attributes: dict[str, Any],
+        provider_name: str,
+        messages: AgentRunInputs | None,
+    ) -> AgentResponse[Any]:
+        """Wrap a non-streaming run with OpenTelemetry tracing."""
+        with get_span(attributes=attributes, span_name_attribute=OtelAttr.AGENT_NAME) as span:
+            if OBSERVABILITY_SETTINGS.SENSITIVE_DATA_ENABLED and messages:
+                capture_messages(span=span, provider_name=provider_name, messages=messages)
```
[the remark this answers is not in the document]
amitmukh: Same fix applied to _run_with_telemetry().

python/packages/claude/tests/test_claude_agent.py
```diff
@@ -945,3 +945,211 @@ async def test_structured_output_with_error_does_not_propagate(self) -> None:
...
+                agent = ClaudeAgent(name="test-agent")
+                response = await agent.run("Hello")
+
+                assert response.text == "Hello!"
+                mock_get_span.assert_called_once()
+                call_kwargs = mock_get_span.call_args[1]
+                assert call_kwargs["attributes"]["gen_ai.agent.name"] == "test-agent"
+                assert call_kwargs["attributes"]["gen_ai.operation.name"] == "invoke_agent"
+        finally:
+            OBSERVABILITY_SETTINGS.enable_instrumentation = original_enabled
```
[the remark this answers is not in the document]
amitmukh: Replaced all manual try/finally save/restore patterns with monkeypatch.setattr(OBSERVABILITY_SETTINGS, "enable_instrumentation", ...) across all 5 telemetry tests.

python/packages/claude/agent_framework_claude/_agent.py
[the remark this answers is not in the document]
amitmukh: Acknowledged. These helpers are the same ones used internally by AgentTelemetryLayer and are the only way to produce consistent telemetry spans. Promoting them to public API is a core-package decision best handled in a follow-up PR. Added a comment in code noting the dependency and intent to update when public equivalents exist.

eavanvalkenburg:
I would prefer this to adopt the agent telemetry layer instead is reimplementing a bunch, if that means we need to do some updates to it thats fine, but otel is a changing spec so we don't want to do double work

amitmukh:
> I would prefer this to adopt the agent telemetry layer instead is reimplementing a bunch, if that means we need to do some updates to it thats fine, but otel is a changing spec so we don't want to do double work

Restructured to adopt `AgentTelemetryLayer` via MRO instead of inlining telemetry.

**Changes in latest commit:**
- ClaudeAgent now inherits from `AgentTelemetryLayer` via a `_ClaudeAgentRunImpl` mixin
- MRO: `ClaudeAgent → AgentTelemetryLayer.run() → _ClaudeAgentRunImpl.run()`
- Removed all private API imports (`_capture_messages`, `_get_span`, etc.)
- Removed duplicated `_run_with_telemetry` / `_run_with_telemetry_stream` methods
- Added `default_options` property to map `system_prompt` → `instructions` for telemetry layer compatibility
- Net -105 lines

python/packages/claude/agent_framework_claude/_agent.py
```diff
@@ -171,7 +172,72 @@ class ClaudeAgentOptions(TypedDict, total=False):
...
+        **kwargs: Any,
+    ) -> AgentResponse[Any]: ...
+
+    def run(
+        self,
+        messages: AgentRunInputs | None = None,
+        *,
+        stream: bool = False,
+        session: AgentSession | None = None,
+        options: Any = None,
```
dmytrostruk: Should this be `options: OptionsT | MutableMapping[str, Any] | None = None,`?

python/packages/claude/agent_framework_claude/_agent.py
```diff
@@ -171,7 +172,72 @@ class ClaudeAgentOptions(TypedDict, total=False):
 )
 
 
-class ClaudeAgent(BaseAgent, Generic[OptionsT]):
+class _ClaudeAgentRunImpl:
```
dmytrostruk: I'm wondering if we should introduce `RawClaudeAgent` instead, so users can use it directly in case the telemetry layer is not needed. cc: @eavanvalkenburg

python/packages/claude/agent_framework_claude/_agent.py
amitmukh: This has been addressed — the options parameter was removed from the _ClaudeAgentRunImpl.run() signature entirely to match AgentTelemetryLayer.run(), which passes options via **kwargs. The options are now extracted inside the method body with options = kwargs.pop("options", None). See commit aba0691.

python/packages/claude/agent_framework_claude/_agent.py
amitmukh: Good suggestion - the RawAgent/Agent pattern from the core package would fit well here. The current _ClaudeAgentRunImpl mixin is essentially a RawClaudeAgent already (holds all core run() logic without telemetry). Renaming it to RawClaudeAgent, making it public, and moving the shared init/lifecycle code into it would align with the framework convention.

Would you prefer this refactor in this PR, or as a follow-up? The current PR scope is adding telemetry instrumentation - the RawClaudeAgent split would be an architectural change that might be cleaner as a separate PR.

eavanvalkenburg: I was about to comment with that, indeed, let's use `RawClaudeAgent` and make sure that that thing works, and then for most people the instrumented one can be used, but if they want to skip that or provider their own telemetry implementation they can. The same architectural change is already here, because the class is built-up different then before, so no need to do that separately.

amitmukh: Thanks for the alignment! I've pushed the refactor in the latest commit. The architecture now follows the RawAgent/Agent pattern:

RawClaudeAgent(BaseAgent) — contains all core logic (init, run, lifecycle, tools, session management) without any telemetry
ClaudeAgent(AgentTelemetryLayer, RawClaudeAgent) — thin wrapper that adds OpenTelemetry instrumentation via MRO
Users who want to skip telemetry or provide their own implementation can use RawClaudeAgent directly. Both classes are exported from the package.

python/packages/claude/agent_framework_claude/_agent.py
```diff
@@ -188,9 +192,9 @@ class ClaudeAgent(BaseAgent, Generic[OptionsT]):
 
         .. code-block:: python
 
-            from agent_framework_claude import ClaudeAgent
+            from agent_framework_claude import RawClaudeAgent
 
-            async with ClaudeAgent(
+            async with RawClaudeAgent(
```
eavanvalkenburg, line 197: this docstring does not need to have all these examples, the common use case will still be using `ClaudeAgent`, this is for completeness, rather then lots of people using it.

python/packages/claude/agent_framework_claude/_agent.py
```diff
@@ -721,3 +738,25 @@ async def _get_stream(
...
+    This is the recommended agent class for most use cases. It includes
+    OpenTelemetry-based telemetry for observability. For a minimal
+    implementation without telemetry, use :class:`RawClaudeAgent`.
+
+    Examples:
+        Basic usage with context manager:
+
+        .. code-block:: python
+
+            from agent_framework_claude import ClaudeAgent
```
eavanvalkenburg: this should show the `agent_framework.anthropic import` instead of the path

eavanvalkenburg (approved):
two small nits, otherwise good to go

python/packages/claude/agent_framework_claude/_agent.py
amitmukh: Done. Trim RawClaudeAgent docstring examples (keep it minimal since it's for completeness, not common use)

python/packages/claude/agent_framework_claude/_agent.py
amitmukh: Done. Change ClaudeAgent docstring import to from agent_framework.anthropic import ClaudeAgent
Bonus: Add RawClaudeAgent to the agent_framework.anthropic lazy imports

PR #4330, opened 2026-02-26
.NET: Wait for input before starting

westey-m:
RunLoopAsync(CancellationToken) uses a 1-second timeout for WaitForInputAsync(CancellationToken) (line 130). When the timeout fires without new input, SemaphoreSlim.WaitAsync returns but the boolean result (indicating timeout vs. signal) was discarded. The loop would then unconditionally create a new workflow_invoke Activity, even though there were no messages to process. On a slow Windows CI machine running net472, multiple timeouts fire before the consumer disposes the run, creating 3+ spurious activities instead of 1.

Changes:

1. StreamingRunEventStream.cs (production fix): Added a guard at the top of the while loop that checks HasUnprocessedMessages before creating a new run-stage activity. If there are no messages (timeout wakeup with no new input), the loop re-enters the wait without creating a spurious activity, halt signal, or telemetry span.

### Contribution Checklist


westey-m:
Closing, since fix is still being investigated.

PR #4361, opened 2026-02-27
.NET: Pass through external input request and handle response conversion for workflow as agent scenario

### LINKED ISSUE #3029: .NET Workflows - Workflow as Agent how to respond to external input requests
Hi,

Trying to follow the documentation https://learn.microsoft.com/en-us/agent-framework/user-guide/workflows/as-agents?pivots=programming-language-csharp#handling-external-input-requests.

My intent is to get user input (just console right now) for my workflow that is converted to Agent.

The document above stops with a statement: "Provide response in the next message".
The docs don't cover how to do it.

The expectation for the docs from me would be to give complete guidance how to achieve a described outcome. In this case how to provide response in the next message.

I assumed that i must ensure the following things:
1. get a thread from the agent and use it in all calls
2. configure checkpointing
3. In a second call to agent.RunStreamingAsync, provide a new ChatMessage(ChatRole.User, [new FunctionResultContent(callId, userInput)]). Since the ExternalRequest is handled as FunctionCallContent.

I failed to progress past my RequestPort.

Stepping through code with a debugger and searching in repo i was expecting that something converts the FunctionResultContent into a different type, such as ExternalResponse, that is used by Workflow to work with responses to requests. I could not find such a place.

It is also not obvious how to create an ExternalResponse myself or if this is the right idea.

Please advise, Regards, Antero

peibekwe:
### Description

This PR handles external input request for non-Function call contents the right way and also fixes issue with external input response conversion from checkpoints.

Fixes #3029 

### Contribution Checklist


dotnet/src/Microsoft.Agents.AI.Workflows/WorkflowSession.cs
```diff
@@ -154,18 +158,135 @@ await this._executionEnvironment
...
+    /// <summary>
+    /// Tries to get a pending request from the state bag by content ID.
+    /// </summary>
+    private ExternalRequest? TryGetPendingRequest(string contentId) =>
+        this.StateBag.GetValue<ExternalRequest>(PendingRequestKeyPrefix + contentId);
+
+    /// <summary>
+    /// Adds a pending request to the state bag.
+    /// </summary>
+    private void AddPendingRequest(string contentId, ExternalRequest request) =>
```
NivedhaChellam: AddPendingRequest  - Stores request in StateBag
RemovePendingRequest - only happens - IF the caller sends back a matching response.

•	What happens if a response is never received? The pending request remains in **StateBag** indefinitely.  This could lead to memory leak.

Please consider adding timeout logic or session cleanup to remove stale pending requests.

dotnet/src/Microsoft.Agents.AI.Workflows/WorkflowSession.cs
```diff
@@ -25,6 +25,10 @@ internal sealed class WorkflowSession : AgentSession
 
     private InMemoryCheckpointManager? _inMemoryCheckpointManager;
 
+    // Key prefix used in StateBag to track pending external requests by their content ID
+    // This enables converting incoming response content back to ExternalResponse when resuming.
+    private const string PendingRequestKeyPrefix = "workflow.pendingRequest:";
```
westey-m: Since you own the WorkflowSession type, you can also use your own typed properties, instead of having to rely on the untyped StateBag.  See the SessionState class below.

peibekwe: Makes sense, updated the change to store pending requests in a typed Dict.

dotnet/src/Microsoft.Agents.AI.Workflows/WorkflowSession.cs
peibekwe: Great callout - I looked into this.. The pending requests are now stored in a typed Dictionary (per Weslie's suggestion), scoped entirely to the WorkflowSession instance — this is not global or shared state, so there's no cross-session accumulation. A few things about the lifecycle:

- Entries are bounded by the number of outstanding external requests in a single workflow run and are removed as soon as a matching response is delivered.
- If the session is abandoned, everything is released (including pending requests) with the session object and can be GC'ed.

Some frameworks seem to support request level timeout by modeling them in the workflow definition itself (e.g., a timer racing against an external event), rather than in the session infrastructure. That feels like the right layer for it if we are to add that but seems out of scope for this bugfix.

I've added comments to the code to capture this lifecycle. Let me know if you think we should discuss further.

NivedhaChellam: Thanks for the detailed clarification. With the pending requests scoped to WorkflowSession and released on session abandonment, this looks safe from a memory-leak perspective. Agree that timeout handling belongs at the workflow-definition layer. The added comments help—no further concerns from my side.

NivedhaChellam (approved):

dotnet/src/Microsoft.Agents.AI.Workflows/WorkflowSession.cs
```diff
@@ -175,12 +300,19 @@ IAsyncEnumerable<AgentResponseUpdate> InvokeStageAsync(
...
                 await this.CreateOrResumeRunAsync(messages, cancellationToken).ConfigureAwait(false);
+#pragma warning disable CA2007 // Analyzer misfiring.
+            await using StreamingRun run = resumeResult.Run;
 #pragma warning restore CA2007
 
-            await run.TrySendMessageAsync(new TurnToken(emitEvents: true)).ConfigureAwait(false);
+            // Send a TurnToken only when no external responses were delivered.
+            // External response handlers already drive continuation turns and can merge
+            // buffered regular messages, so an extra TurnToken would cause a redundant turn.
+            if (!resumeResult.HasMatchedExternalResponses)
```
lokitoth: We may still need to send a `TurnToken`:

If there were also normal messages, they will be queued up, but will not run unless the starting executor works without a `TurnToken`, or the starting executor is the one that had the request.

dotnet/src/Microsoft.Agents.AI.Workflows/WorkflowSession.cs
```diff
@@ -192,8 +324,20 @@ IAsyncEnumerable<AgentResponseUpdate> InvokeStageAsync(
...
+                        {
+                            ExternalRequest externalRequest when externalRequest.TryGetDataAs(out FunctionCallContent? fcc) => (fcc, fcc.CallId),
+                            ExternalRequest externalRequest when externalRequest.TryGetDataAs(out UserInputRequestContent? uic) => (uic, uic.Id),
+                            ExternalRequest externalRequest => ((AIContent)externalRequest.ToFunctionCall(), externalRequest.RequestId)
+                        };
+
+                        // Track the pending request so we can convert incoming responses back to ExternalResponse
+                        if (contentId != null)
+                        {
+                            this.AddPendingRequest(contentId, requestInfo.Request);
```
lokitoth: Is it possible for two agents to independently generate a request with the same `contentId`? We may need to scope this by source agent.

peibekwe: I have updated this to take the portId into account to prevent possible collision between two agents.

dotnet/src/Microsoft.Agents.AI.Workflows/WorkflowSession.cs
peibekwe: I have updated the solution to handle the scenario. So `TurnToken` will be sent if there's no external response or if a regular message exists where the start executor isn't the one with the request.

crickman (approved):

dotnet/src/Microsoft.Agents.AI.Workflows/WorkflowSession.cs
```diff
@@ -154,18 +175,155 @@ await this._executionEnvironment
...
+                // Skip duplicate response content for an already-matched content ID
+                if (contentId != null && matchedContentIds?.Contains(contentId) == true)
+                {
+                    continue;
+                }
+
+                if (contentId != null
+                    && this.TryGetPendingRequest(contentId) is ExternalRequest pendingRequest)
+                {
+                    if (!run.TryGetResponsePortExecutorId(pendingRequest.PortInfo.PortId, out string? responseExecutorId))
```
lokitoth: This might not work right if the request gets intercepted along the way, because the outgoing `RequestPort` may not be the same as the underlying agent. This also adds the complication that the agent executor could get the response several turns later, too.

dotnet/src/Microsoft.Agents.AI.Workflows/WorkflowsJsonUtilities.cs
```diff
@@ -90,6 +90,7 @@ private static JsonSerializerOptions CreateDefaultOptions()
     [JsonSerializable(typeof(ChatMessage))]
     [JsonSerializable(typeof(ExternalRequest))]
     [JsonSerializable(typeof(ExternalResponse))]
+    [JsonSerializable(typeof(Dictionary<string, ExternalRequest>))]
```
lokitoth: This should not be necessary if ExternalRequest is already in the list. Were you seeing issues?

peibekwe: You're right, this is redundant. Removed.

dotnet/src/Microsoft.Agents.AI.Workflows/WorkflowSession.cs
peibekwe: I believe that intercepted requests use `Context.SendMessageAsync`, so they don't create `ExternalRequest` so shouldn't be impacted. However, I updated this to handle a gap when a resume contained only an external response directed at a non-start executor.

lokitoth: Good point

lokitoth (approved):

PR #4401, opened 2026-03-02
.NET: Discover skill resources from directory instead of markdown links

SergeyMenshykh:
## Summary

Switches `FileAgentSkillLoader` from parsing markdown links (`[text](path)`) to discover resources to scanning the skill directory for files with matching extensions. This is more robust and aligns with the convention that all files in a skill folder are part of that skill.

## Changes

- **Resource discovery**: Replace regex-based markdown link extraction with recursive directory scanning using configurable allowed extensions (`.md`, `.json`, `.yaml`, `.yml`, `.csv`, `.xml`, `.txt` by default)
- **Graceful degradation**: Invalid resources (path traversal, symlink escape, disallowed extensions) are now skipped with warnings instead of rejecting the entire skill
- **Configurable extensions**: `FileAgentSkillLoader` accepts an optional `allowedResourceExtensions` parameter; `FileAgentSkillsProviderOptions` exposes a corresponding property
- **Log safety**: Added `SanitizePathForLog` to prevent log injection via crafted filenames containing control characters
- **Tests**: Updated and expanded unit tests to reflect the new discovery-based approach

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI/Skills/FileAgentSkillLoader.cs
+++ b/dotnet/src/Microsoft.Agents.AI/Skills/FileAgentSkillLoader.cs
@@ -408,34 +408,17 @@ internal sealed partial class FileAgentSkillLoader
     /// </summary>
     private static string SanitizePathForLog(string path)
     {
-        foreach (char c in path)
+        char[]? chars = null;
+        for (int i = 0; i < path.Length; i++)
         {
-            if (char.IsControl(c))
+            if (char.IsControl(path[i]))
             {
-#if NET
-                return string.Create(path.Length, path, static (span, source) =>
-                {
-                    for (int i = 0; i < source.Length; i++)
-                    {
-                        span[i] = char.IsControl(source[i]) ? '?' : source[i];
-                    }
-                });
-#else
-                char[] chars = path.ToCharArray();
-                for (int i = 0; i < chars.Length; i++)
-                {
-                    if (char.IsControl(chars[i]))
-                    {
-                        chars[i] = '?';
-                    }
-                }
-
-                return new string(chars);
-#endif
+                chars ??= path.ToCharArray();
+                chars[i] = '?';
             }
         }
 
-        return path;
+        return chars is null ? path : new string(chars);
     }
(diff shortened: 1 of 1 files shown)


dotnet/src/Microsoft.Agents.AI/Skills/FileAgentSkillLoader.cs
```diff
@@ -270,34 +266,64 @@ private bool TryParseSkillDocument(string content, string skillFilePath, out Ski
...
+        foreach (string filePath in Directory.EnumerateFiles(skillDirectoryFullPath, "*", SearchOption.AllDirectories))
         {
-            string fullPath = Path.GetFullPath(Path.Combine(skillDirectoryPath, resourceName));
+            string fileName = Path.GetFileName(filePath);
+
+            // Exclude SKILL.md itself
+            if (string.Equals(fileName, SkillFileName, StringComparison.OrdinalIgnoreCase))
+            {
+                continue;
+            }
```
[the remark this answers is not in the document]
SergeyMenshykh: It will be addressed separately later.

dotnet/src/Microsoft.Agents.AI/Skills/FileAgentSkillLoader.cs
```diff
@@ -372,6 +402,60 @@ private static string NormalizeResourcePath(string path)
...
+                char[] chars = path.ToCharArray();
+                for (int i = 0; i < chars.Length; i++)
+                {
+                    if (char.IsControl(chars[i]))
+                    {
+                        chars[i] = '?';
+                    }
+                }
+
+                return new string(chars);
```
westey-m: This code seems a bit weird, in that you are looping through path on line 411, but here (inside the loop) looping through it again and existing the method.  It's probably fine, but it feels a bit unintuitive.

SergeyMenshykh: Yes, it does. It somehow slipped my attention. Thanks for pointing that out. Fixed.

PR #4403, opened 2026-03-02
Python: Fix OllamaChatClient passing unsupported kwargs to ollama.AsyncClient…

### LINKED ISSUE #4402: Python: [Bug]: OllamaChatClient passes unsupported kwargs (e.g. allow_multiple_tool_calls) to ollama.AsyncClient.chat(), causing TypeError
## Description

When using `OllamaChatClient` with `HandoffBuilder` (from `agent-framework-orchestrations`), the orchestration layer injects additional keyword arguments — notably `allow_multiple_tool_calls=True` — into `_inner_get_response()`. These kwargs are passed through unfiltered via `**kwargs` to `ollama.AsyncClient.chat()`, which does not accept them, resulting in a `TypeError`.

**What happened?**

Running a `HandoffBuilder` workflow with `OllamaChatClient` raises:
```
TypeError: AsyncClient.chat() got an unexpected keyword argument 'allow_multiple_tool_calls'
```

`OllamaChatClient._inner_get_response()` should filter out kwargs that are not supported by `ollama.AsyncClient.chat()` before forwarding them, or the orchestration layer should not pass provider-specific kwargs to adapters that don't support them.

**Steps to reproduce:**
1. Create agents using `OllamaChatClient`
2. Build a workflow using `HandoffBuilder` with those agents
3. Invoke the workflow — the handoff mechanism injects `allow_multiple_tool_calls=True`
4. The call to `ollama.AsyncClient.chat()` fails with `TypeError`

## Code Sample

Minimal reproduction:

```python
import asyncio
from agent_framework.agents import Agent
from agent_framework_ollama import OllamaChatClient
from agent_framework_orchestrations.handoffs import HandoffBuilder, HandoffAgentUserRequest

async def main():
    client = OllamaChatClient(model_id="qwen2.5:3b")

    coordinator = Agent(
        name="Coordinator",
        instructions="Route incidents to IncidentAgent.",
        client=client,
    )
    incident_agent = Agent(
        name="IncidentAgent",
        instructions="Handle incidents.",
        client=client,
    )

    handoff = (
        HandoffBuilder()
        .with_client(client)
        .participants([coordinator, incident_agent])
        .build()
    )

    # This triggers the bug — HandoffBuilder passes allow_multiple_tool_calls=True
    # through to OllamaChatClient._inner_get_response(), which forwards it to
    # ollama.AsyncClient.chat()
    async for msg in handoff.invoke("Report an incident"):
        print(msg)

asyncio.run(main())
```

```
TypeError: AsyncClient.chat() got an unexpected keyword argument 'allow_multiple_tool_calls'
```

The root cause is in `_chat_client.py`, method `_inner_get_response()`, lines ~360 and ~381:

```python
# Streaming path (line ~360):
response_object = await self.client.chat(
    stream=True,
    **options_dict,
    **kwargs,          # <-- allow_multiple_tool_calls leaks through here
)

# Non-streaming path (line ~378):
response = await self.client.chat(
    stream=False,
    **options_dict,
    **kwargs,          # <-- same issue here
)
```

The `**kwargs` from the method signature is forwarded directly to `ollama.AsyncClient.chat()` without filtering out kwargs that the Ollama Python client doesn't support.

## Workaround (Monkey Patch)

We found two functional workarounds. The most reliable one is monkey-patching `ollama.AsyncClient.chat()` directly to filter kwargs by inspecting its signature:

```python
"""
Monkey-patch for agent_framework_ollama to strip unsupported kwargs.

The MAF Ollama adapter passes **kwargs directly to ollama.AsyncClient.chat(),
but HandoffBuilder injects kwargs like `allow_multiple_tool_calls` that the
Ollama Python client doesn't support, causing a TypeError.

This patch wraps ollama.AsyncClient.chat() itself to strip any kwargs not in
its explicit signature, so no matter how MAF calls it, invalid kwargs are
dropped.

Usage:
    import patch_ollama  # noqa: F401  — must import before using OllamaChatClient
"""

import functools
import inspect
from typing import Any


def _apply_patch() -> None:
    import ollama

    original_chat = ollama.AsyncClient.chat

    if getattr(original_chat, "_patched_for_kwargs", False):
        return  # Already patched

    # Get the valid parameter names from the original signature
    valid_params = set(inspect.signature(original_chat).parameters.keys())
    valid_params.discard("self")

    @functools.wraps(original_chat)
    async def _patched_chat(self, *args, **kwargs):
        filtered = {k: v for k, v in kwargs.items() if k in valid_params}
        dropped = set(kwargs.keys()) - set(filtered.keys())
        if dropped:
            pass  # silently drop unsupported kwargs
        return await original_chat(self, *args, **filtered)

    _patched_chat._patched_for_kwargs = True
    ollama.AsyncClient.chat = _patched_chat

_apply_patch()
```

### Suggested proper fix

A cleaner fix would be inside `OllamaChatClient._inner_get_response()` itself — pop/discard any kwargs that aren't valid for `ollama.AsyncClient.chat()` before forwarding them. For example:

```python
# In _inner_get_response, before calling self.client.chat():
kwargs.pop("allow_multiple_tool_calls", None)
# Or more generically, filter kwargs against ollama.AsyncClient.chat()'s signature
```

## Package Versions

- agent-framework-core: 1.0.0rc2
- agent-framework: 1.0.0rc2
- agent-framework-ollama: 1.0.0b260225
- agent-framework-orchestrations: 1.0.0b260225
- ollama (Python client): 0.6.1

## Python Version

Python 3.13.9

## Additional Context

- Ollama server version: 0.15.6
- Model tested: qwen2.5:3b
- The issue affects both streaming and non-streaming code paths in `_inner_get_response()`
- The `allow_multiple_tool_calls` kwarg is injected by the handoff orchestration layer and is meaningful for providers like OpenAI/Azure, but the Ollama adapter does not handle or strip it
- We also attempted patching `_inner_get_response` directly, but the kwargs flow through closures (`_stream()` and `_get_response()`), making the `ollama.AsyncClient.chat()` monkey-patch more reliable

max-montes:
### Motivation and Context

Stop forwarding **kwargs from _inner_get_response() to ollama.AsyncClient.chat(). When the orchestration layer (e.g. HandoffBuilder) injects kwargs like allow_multiple_tool_calls=True, these were passed through unfiltered, causing a TypeError since ollama.AsyncClient.chat() does not accept them.

### Description

This aligns with how other clients (e.g. OpenAIChatClient) handle kwargs — they only pass the prepared options_dict to the provider's API, not the raw **kwargs from the middleware chain.

Fixes #4402

### Contribution Checklist

- [ X] The code builds clean without any errors or warnings
- [ X] The PR follows the [Contribution Guidelines](https://github.com/microsoft/agent-framework/blob/main/CONTRIBUTING.md)
- [X ] All unit tests pass, and I have added new tests where possible

### DIFF
--- a/.github/workflows/merge-gatekeeper.yml
+++ b/.github/workflows/merge-gatekeeper.yml
@@ -29,4 +29,7 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           timeout: 3600
           interval: 30
-          ignored: CodeQL,CodeQL analysis (csharp)
+          # "Cleanup artifacts", "Agent", "Prepare", and "Upload results" are check runs
+          # created by an org-level GitHub App (MSDO), not by any workflow in this repo.
+          # They are outside our control and their transient failures should not block merges.
+          ignored: CodeQL,CodeQL analysis (csharp),Cleanup artifacts,Agent,Prepare,Upload results
--- a/dotnet/Directory.Packages.props
+++ b/dotnet/Directory.Packages.props
@@ -19,8 +19,8 @@
     <PackageVersion Include="Aspire.Microsoft.Azure.Cosmos" Version="$(AspireAppHostSdkVersion)" />
     <PackageVersion Include="CommunityToolkit.Aspire.OllamaSharp" Version="13.0.0" />
     <!-- Azure.* -->
-    <PackageVersion Include="Azure.AI.Projects" Version="1.2.0-beta.5" />
-    <PackageVersion Include="Azure.AI.Projects.OpenAI" Version="1.0.0-beta.5" />
+    <PackageVersion Include="Azure.AI.Projects" Version="2.0.0-beta.1" />
+    <PackageVersion Include="Azure.AI.Projects.OpenAI" Version="2.0.0-beta.1" />
     <PackageVersion Include="Azure.AI.Agents.Persistent" Version="1.2.0-beta.8" />
     <PackageVersion Include="Azure.AI.OpenAI" Version="2.8.0-beta.1" />
     <PackageVersion Include="Azure.Identity" Version="1.17.1" />
@@ -35,7 +35,7 @@
     <!-- System.* -->
     <PackageVersion Include="Microsoft.Bcl.AsyncInterfaces" Version="10.0.3" />
     <PackageVersion Include="Microsoft.Bcl.HashCode" Version="6.0.0" />
-    <PackageVersion Include="System.ClientModel" Version="1.8.1" />
+    <PackageVersion Include="System.ClientModel" Version="1.9.0" />
     <PackageVersion Include="System.CodeDom" Version="10.0.0" />
     <PackageVersion Include="System.Collections.Immutable" Version="10.0.1" />
     <PackageVersion Include="System.CommandLine" Version="2.0.0-rc.2.25502.107" />
@@ -94,7 +94,7 @@
     <PackageVersion Include="Microsoft.SemanticKernel.Agents.AzureAI" Version="1.67.0-preview" />
     <PackageVersion Include="Microsoft.SemanticKernel.Plugins.OpenApi" Version="1.67.0" />
     <!-- Agent SDKs -->
-    <PackageVersion Include="GitHub.Copilot.SDK" Version="0.1.23" />
+    <PackageVersion Include="GitHub.Copilot.SDK" Version="0.1.29" />
     <PackageVersion Include="Microsoft.Agents.CopilotStudio.Client" Version="1.3.171-beta" />
     <!-- M365 Agents SDK -->
     <PackageVersion Include="AdaptiveCards" Version="3.1.0" />
--- a/dotnet/samples/02-agents/FoundryAgents/FoundryAgents_Evaluations_Step01_RedTeaming/Program.cs
+++ b/dotnet/samples/02-agents/FoundryAgents/FoundryAgents_Evaluations_Step01_RedTeaming/Program.cs
@@ -60,7 +60,7 @@ Console.WriteLine();
 
 // Submit the red team run to the service
 Console.WriteLine("Submitting red team run...");
-RedTeam redTeamRun = await aiProjectClient.RedTeams.CreateAsync(redTeamConfig);
+RedTeam redTeamRun = await aiProjectClient.RedTeams.CreateAsync(redTeamConfig, options: null);
 
 Console.WriteLine($"Red team run created: {redTeamRun.Name}");
 Console.WriteLine($"Status: {redTeamRun.Status}");
--- a/dotnet/samples/02-agents/FoundryAgents/FoundryAgents_Step22_MemorySearch/Program.cs
+++ b/dotnet/samples/02-agents/FoundryAgents/FoundryAgents_Step22_MemorySearch/Program.cs
@@ -35,7 +35,7 @@ string userScope = $"user_{Environment.MachineName}";
 AIProjectClient aiProjectClient = new(new Uri(endpoint), new AzureCliCredential());
 
 // Create the Memory Search tool configuration
-MemorySearchTool memorySearchTool = new(memoryStoreName, userScope)
+MemorySearchPreviewTool memorySearchTool = new(memoryStoreName, userScope)
 {
     // Optional: Configure how quickly new memories are indexed (in seconds)
     UpdateDelay = 1,
--- a/dotnet/samples/03-workflows/Declarative/HostedWorkflow/Program.cs
+++ b/dotnet/samples/03-workflows/Declarative/HostedWorkflow/Program.cs
@@ -88,7 +88,9 @@ internal sealed class Program
     {
         string workflowYaml = File.ReadAllText("MathChat.yaml");
 
+#pragma warning disable AAIP001 // WorkflowAgentDefinition is experimental
         WorkflowAgentDefinition workflowAgentDefinition = WorkflowAgentDefinition.FromYaml(workflowYaml);
+#pragma warning restore AAIP001
 
         return
             await agentClient.CreateAgentAsync(
--- a/dotnet/src/Microsoft.Agents.AI.AzureAI/AzureAIProjectChatClientExtensions.cs
+++ b/dotnet/src/Microsoft.Agents.AI.AzureAI/AzureAIProjectChatClientExtensions.cs
@@ -39,7 +39,7 @@ public static partial class AzureAIProjectChatClientExtensions
     /// <exception cref="InvalidOperationException">The agent with the specified name was not found.</exception>
     /// <remarks>
     /// When instantiating a <see cref="ChatClientAgent"/> by using an <see cref="AgentReference"/>, minimal information will be available about the agent in the instance level, and any logic that relies
-    /// on <see cref="AIAgent.GetService(Type, object?)"/> to retrieve information about the agent like <see cref="AgentVersion" /> will receive <see langword="null"/> as the result.
+    /// on <see cref="AIAgent.GetService{TService}(object?)"/> to retrieve information about the agent like <see cref="AgentVersion" /> will receive <see langword="null"/> as the result.
     /// </remarks>
     public static ChatClientAgent AsAIAgent(
         this AIProjectClient aiProjectClient,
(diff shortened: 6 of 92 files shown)


markwallace-microsoft:
<a href="https://github.com/microsoft/agent-framework/blob/fe62436830e80c3982aab1e0f39cd43ddfeffd2c/README.md">Python Test Coverage</a><details><summary>Python Test Coverage Report • </summary><table><tr><th>File</th><th>Stmts</th><th>Miss</th><th>Cover</th><th>Missing</th></tr><tbody><tr><td><b>TOTAL</b></td><td><b>22544</b></td><td><b>2798</b></td><td><b>87%</b></td><td>&nbsp;</td></tr></tbody></table><i>report-only-changed-files is enabled. No files were changed during this commit :)</i></details>

## Python Unit Test Overview
| 4587 | 25 :zzz: | 0 :x: | 0 :fire: | 1m 20s :stopwatch: |

TaoChenOSU:
This looks like a duplicate of #4404. Please confirm and we can close this.

python/packages/ollama/tests/test_ollama_chat_client.py
```diff
@@ -248,6 +248,67 @@ async def test_cmc(
...
+    ollama_client = OllamaChatClient()
+
+    # Call _inner_get_response with an unsupported kwarg that the orchestration layer would inject
+    coro = ollama_client._inner_get_response(
+        messages=chat_history,
+        options={"model_id": "test"},
+        stream=False,
+        allow_multiple_tool_calls=True,
+    )
+    result = await coro
```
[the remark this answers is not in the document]
max-montes: Tests now use public get_response() API instead of private 
  _inner_get_response(), matching the real middleware/orchestration call path

python/packages/ollama/tests/test_ollama_chat_client.py
```diff
@@ -248,6 +248,67 @@ async def test_cmc(
...
+    chat_history.append(Message(text="hello world", role="user"))
+
+    ollama_client = OllamaChatClient()
+
+    stream = ollama_client._inner_get_response(
+        messages=chat_history,
+        options={"model_id": "test"},
+        stream=True,
+        allow_multiple_tool_calls=True,
+    )
```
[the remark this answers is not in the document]
max-montes: Added # noqa: ARG002 annotation on the unused **kwargs parameter to explicitly document it's accepted for interface compatibility but intentionally not forwarded to Ollama

python/packages/ollama/agent_framework_ollama/_chat_client.py
```diff
@@ -360,7 +360,6 @@ async def _stream() -> AsyncIterable[ChatResponseUpdate]:
                     response_object: AsyncIterable[OllamaChatResponse] = await self.client.chat(  # type: ignore[misc]
                         stream=True,
                         **options_dict,
-                        **kwargs,
                     )
```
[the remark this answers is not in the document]
max-montes: Added allow_multiple_tool_calls to exclude_keys in _prepare_options() — this covers the case where the flag arrives via the options dict (e.g., from Agent.default_options or workflow cloning), not just as a top-level kwarg

python/packages/ollama/agent_framework_ollama/_chat_client.py
```diff
@@ -378,7 +377,6 @@ async def _get_response() -> ChatResponse:
                 response: OllamaChatResponse = await self.client.chat(  # type: ignore[misc]
                     stream=False,
                     **options_dict,
-                    **kwargs,
                 )
```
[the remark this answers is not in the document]
max-montes: Added a third test (test_cmc_ignores_unsupported_options) verifying the options dict path is also filtered

max-montes:
> This looks like a duplicate of #4404. Please confirm and we can close this.

Hi @TaoChenOSU — this PR was actually submitted first (#4403 at 22:33 UTC vs
  #4404 at 22:41 UTC).

  Beyond timing, the approaches differ:

   - #4404 uses a blocklist (_UNSUPPORTED_CHAT_KWARGS) to filter specific kwargs
   before chat() — but only covers the **kwargs path. New unsupported kwargs 
  would need to be added to the set manually.
   - This PR removes **kwargs forwarding to chat() entirely (it was never 
  needed) AND adds allow_multiple_tool_calls to exclude_keys in 
  _prepare_options(), which covers the options dict path too (e.g., from 
  Agent.default_options or workflow cloning). This is a more complete fix with 
  less code.

  Happy to make any adjustments you'd like, but if you've chose to go with the other fix, I'll go ahead and close.

python/packages/ollama/tests/test_ollama_chat_client.py
```diff
@@ -248,6 +248,88 @@ async def test_cmc(
...
+    """
+    mock_chat.return_value = mock_chat_completion_response
+    chat_history.append(Message(text="hello world", role="user"))
+
+    ollama_client = OllamaChatClient()
+    await ollama_client.get_response(
+        messages=chat_history,
+        options={"allow_multiple_tool_calls": True},
+    )
+
```
moonbox3: Minor: the result of `get_response` is discarded here. Adding `result = await ...` and `assert result.text == "test"` would be consistent with the sibling tests and would confirm the call still succeeds end-to-end after filtering.

```suggestion
    result = await ollama_client.get_response(
        messages=chat_history,
        options={"allow_multiple_tool_calls": True},
    )

    assert result.text == "test"
```

moonbox3 (commented):
## Automated Code Review

**Reviewers:** 3 | **Confidence:** 91%

### ✓ Correctness
> The diff correctly fixes issue #4402 by stopping the forwarding of unsupported `**kwargs` to `ollama.AsyncClient.chat()` and adding `allow_multiple_tool_calls` to the set of keys excluded during option preparation. The three new tests cover the non-streaming kwargs path, the streaming kwargs path, and the options-dict path. The changes are focused, logically sound, and consistent with the existing codebase patterns.

### ✓ Security Reliability
> This diff is a security improvement that removes the blind forwarding of **kwargs to the underlying Ollama AsyncClient.chat() call, preventing callers from injecting arbitrary, unvalidated parameters into the downstream API. The addition of 'allow_multiple_tool_calls' to exclude_keys and the accompanying regression tests are well-structured. No security or reliability issues are introduced by this change.

### ✓ Test Coverage
> The diff removes forwarding of **kwargs to ollama.AsyncClient.chat() and adds allow_multiple_tool_calls to the exclude_keys set in _prepare_options. Three new tests cover the non-streaming kwargs path, streaming kwargs path, and options-dict path respectively. Coverage is solid for the specific regression case (issue #4402). Minor gaps: no test verifies that arbitrary/unknown kwargs (beyond allow_multiple_tool_calls) are also silently dropped, which would better document the broader behavioral change of removing **kwargs forwarding entirely. test_cmc_ignores_unsupported_options omits a result assertion but the intent is clear.

### Suggestions
- Consider logging a debug-level warning when kwargs are silently dropped, so that users who accidentally pass an unsupported parameter get some feedback during development rather than silent no-op behavior.
- Consider logging a debug-level warning when kwargs are silently dropped so that callers can diagnose unexpected behaviour during development, rather than having parameters disappear without any signal.
- Consider adding a test that passes an arbitrary unknown kwarg (e.g., `some_future_flag=True`) to `get_response` and asserts it does not appear in the `self.client.chat()` call. This would document the broader behavioral change (all kwargs are now dropped, not just `allow_multiple_tool_calls`).
- In `test_cmc_ignores_unsupported_options`, consider adding `assert result.text == "test"` for consistency with the other two new tests, to confirm the response is still valid after filtering.

---

python/packages/ollama/agent_framework_ollama/_chat_client.py
```diff
@@ -395,8 +393,9 @@ def _prepare_options(self, messages: Sequence[Message], options: Mapping[str, An
 
             messages = prepend_instructions_to_messages(list(messages), instructions, role="system")
 
-        # Keys to exclude from processing
-        exclude_keys = {"instructions", "tool_choice"}
+        # Keys to exclude from processing — these are either handled separately
+        # or not supported by the Ollama API.
+        exclude_keys = {"instructions", "tool_choice", "allow_multiple_tool_calls"}
```
eavanvalkenburg: this is not a approach we want to take, since that parameter is already set to None in the OllamaChatOptions, people will get notified by their IDE if that is set, if that is being set by something like a workflow then we need to update that. The reason we don't want to hardcode filtering like this is because we want to not block future updates, let's say at some point Ollama does implement support for this parameter, then that will not be usable, because we filter it out, so we decided we would prefer to use the None above to notify the user beforehand that it is likely not right, and then we just let the API tell the user exactly what's wrong.

max-montes: @eavanvalkenburg: Reverted the exclude_keys change — agreed that the fix belongs in the workflow layer, not in client-side filtering. The core fix (removing **kwargs forwarding to chat()) still prevents the TypeError on the kwargs path.

python/packages/ollama/tests/test_ollama_chat_client.py
max-montes: @moonbox3: Removed the options-dict test since we're no longer filtering there. Both remaining tests assign the result and assert on it, consistent with the sibling tests.

max-montes:
Hey @eavanvalkenburg, gentle ping that I made the changes you requested

python/packages/ollama/tests/test_ollama_chat_client.py
```diff
@@ -248,6 +248,62 @@ async def test_cmc(
...
+    Regression test for: https://github.com/microsoft/agent-framework/issues/4402
+    """
+    mock_chat.return_value = mock_streaming_chat_completion_response
+    chat_history.append(Message(text="hello world", role="user"))
+
+    ollama_client = OllamaChatClient()
+    result = ollama_client.get_response(
+        messages=chat_history,
+        stream=True,
+        allow_multiple_tool_calls=True,
```
eavanvalkenburg, line 296: this is no longer a valid pattern, one should always use `options` and we have a PR open to remove `kwargs` in most places(#4581), I think we can close this PR for now

eavanvalkenburg:
This is superseded by https://github.com/microsoft/agent-framework/pull/4581 but thanks for helping out @max-montes

PR #4404, opened 2026-03-02
Python: fix(python): Filter unsupported kwargs in OllamaChatClient._inner_get_response

### LINKED ISSUE #4402: Python: [Bug]: OllamaChatClient passes unsupported kwargs (e.g. allow_multiple_tool_calls) to ollama.AsyncClient.chat(), causing TypeError

LEDazzio01:
## Summary

Fixes #4402

`OllamaChatClient._inner_get_response()` forwards `**kwargs` directly to `ollama.AsyncClient.chat()`. When used with `HandoffBuilder`, kwargs like `allow_multiple_tool_calls=True` leak through, causing a `TypeError` since the Ollama Python client doesn't accept this parameter.

### `_chat_client.py`
- Added `_UNSUPPORTED_CHAT_KWARGS` set containing kwargs that orchestration layers inject but Ollama doesn't support
- Filter `kwargs` before passing to `self.client.chat()` in both streaming and non-streaming paths

### `test_ollama_chat_client.py`
- Added `test_cmc_filters_unsupported_kwargs` — verifies `allow_multiple_tool_calls` is stripped in non-streaming mode
- Added `test_cmc_streaming_filters_unsupported_kwargs` — verifies the same for streaming mode

## Motivation

When `HandoffBuilder` is used with `OllamaChatClient`, it injects `allow_multiple_tool_calls=True` as a kwarg. `OllamaChatOptions` already documents this as unsupported (`allow_multiple_tool_calls: None`), but the kwarg was still forwarded to the Ollama API.

The fix matches the existing pattern in `_prepare_options()` where `tool_choice` is excluded via `exclude_keys`.

### DIFF
--- a/python/packages/ollama/agent_framework_ollama/_chat_client.py
+++ b/python/packages/ollama/agent_framework_ollama/_chat_client.py
@@ -268,9 +268,9 @@ OLLAMA_MODEL_OPTION_TRANSLATIONS: dict[str, str] = {
 }
 """Maps ChatOptions keys to Ollama model option parameter names."""
 
-# Kwargs that may be injected by orchestration layers (e.g. HandoffBuilder)
-# but are not supported by ollama.AsyncClient.chat().  These are silently
-# stripped in _inner_get_response so they never reach the Ollama API.
+# Framework-level kwargs that are not supported by ollama.AsyncClient.chat().
+# These are silently stripped in _inner_get_response so they never reach the
+# Ollama API.
 _UNSUPPORTED_CHAT_KWARGS: set[str] = {
     "allow_multiple_tool_calls",
 }
@@ -358,9 +358,8 @@ class OllamaChatClient(
         stream: bool = False,
         **kwargs: Any,
     ) -> Awaitable[ChatResponse] | ResponseStream[ChatResponseUpdate, ChatResponse]:
-        # Filter out kwargs that are not supported by ollama.AsyncClient.chat().
-        # Orchestration layers (e.g. HandoffBuilder) may inject kwargs like
-        # allow_multiple_tool_calls that the Ollama Python client doesn't accept.
+        # Filter out framework-level kwargs that are not supported by
+        # ollama.AsyncClient.chat().
         filtered_kwargs = {k: v for k, v in kwargs.items() if k not in _UNSUPPORTED_CHAT_KWARGS}
 
         if stream:


LEDazzio01 (commented):
Good catches! Pushed a follow-up commit (`f9b01dc`) addressing both:

1. **Options path** — Added `allow_multiple_tool_calls` to `exclude_keys` in `_prepare_options()`, so it's stripped whether it arrives via `**kwargs` *or* the `options` dict (e.g. from `Agent.default_options` / workflow cloning).

2. **Options-path test** — Added `test_cmc_filters_unsupported_options` which passes `allow_multiple_tool_calls` inside the `options` dict and asserts it doesn't reach `ollama.AsyncClient.chat()`.

Both entry points are now covered.

markwallace-microsoft:
<a href="https://github.com/microsoft/agent-framework/blob/f9b01dcb276916183ea715ec1e2d356ee206ddc0/README.md">Python Test Coverage</a><details><summary>Python Test Coverage Report • </summary><table><tr><th>File</th><th>Stmts</th><th>Miss</th><th>Cover</th><th>Missing</th></tr><tbody><tr><td><b>TOTAL</b></td><td><b>22222</b></td><td><b>2757</b></td><td><b>87%</b></td><td>&nbsp;</td></tr></tbody></table><i>report-only-changed-files is enabled. No files were changed during this commit :)</i></details>

## Python Unit Test Overview
| 4720 | 247 :zzz: | 0 :x: | 0 :fire: | 1m 16s :stopwatch: |

python/packages/ollama/agent_framework_ollama/_chat_client.py
```diff
@@ -268,6 +268,13 @@ class OllamaChatOptions(ChatOptions[ResponseModelT], Generic[ResponseModelT], to
 }
 """Maps ChatOptions keys to Ollama model option parameter names."""
 
+# Kwargs that may be injected by orchestration layers (e.g. HandoffBuilder)
```
TaoChenOSU: nit: We shouldn't mention downstream modules in the comment. We can just say this framework level key is not supported by the ollama client or it's handled separately.

python/packages/ollama/agent_framework_ollama/_chat_client.py
```diff
@@ -268,6 +268,13 @@ class OllamaChatOptions(ChatOptions[ResponseModelT], Generic[ResponseModelT], to
 }
 """Maps ChatOptions keys to Ollama model option parameter names."""
 
+# Framework-level kwargs that are not supported by ollama.AsyncClient.chat().
+# These are silently stripped in _inner_get_response so they never reach the
+# Ollama API.
+_UNSUPPORTED_CHAT_KWARGS: set[str] = {
+    "allow_multiple_tool_calls",
```
eavanvalkenburg, line 275: this is not a approach we want to take, since that parameter is already set to `None` in the OllamaChatOptions, people will get notified by their IDE if that is set, if that is being set by something like a workflow then we need to update that. The reason we don't want to hardcode filtering like this is because we want to not block future updates, let's say at some point Ollama does implement support for this parameter, then that will not be usable, because we filter it out, so we decided we would prefer to use the `None` above to notify the user beforehand that it is likely not right, and then we just let the API tell the user exactly what's wrong.

eavanvalkenburg: So reading more of the above, the real issue is that the HandoffBuilder set's parameters that are not universally applicable, which is what we would need to fix.

moonbox3:
Closing due this being a duplicate of #4403.

Issue #4446, opened 2026-03-04
.NET: Recommended way to register an AIAgent in the Service Collection


marcominerva:
I'm working on a Web API. In the **Microsoft.Agents.AI.Hosting** namespace, there is the `AddAIAgent` extension method on `IServiceCollection`, which allows you to create and register an AIAgent in the service collection as a singleton.

However, in the new end-to-end example introduced in issue #1488, the AIAgent is explicitly registered using `AddScoped<AIAgent>`:

https://github.com/microsoft/agent-framework/blob/4dc20c6be404dacdac4d58f6741985b22c87e0b1/dotnet/samples/05-end-to-end/AspNetAgentAuthorization/Service/Program.cs#L81

What is the recommended lifetime for AIAgent in an ASP.NET Core application? Should it be registered as a singleton via `AddAIAgent`, or as scoped as shown in the end-to-end sample?

westey-m:
@marcominerva, the scope with which an agent should be registered is very much dependent on the dependencies that the agent has and how they are implemented.  In the referenced sample, the agent depends on some function tools in ExpenseService, which in turn depend on KeycloakUserContext which uses IHttpContextAccessor.

The important thing here is in how KeycloakUserContext uses IHttpContextAccessor.  It reads the user in the constructor, which means that each KeycloakUserContext instance is tied to a specific user from that http request and therefore has to be registered as scoped.  This means all classes that use it up the stack, also needs to be scoped, otherwise they will end up referencing the first user that called the service for all subsequent requests.

Of course, it doesn't have to be this way, and previously I had intended to showcase a singleton based approach, but ended up getting sidetracked 😄.  I'll update the KeycloakUserContext to be singleton compatible by reading values from the IHttpContextAccessor on each property get and update the sample.  I generally think singletons are preferable to avoid unnecessary allocations, but each class in the stack needs to ensure that they don't store / cache user info in fields anywhere and do auth checks for each method call.

The approach you should take in your app will therefore very much depend on the dependencies you have.  If some of them need to be recreated for each request, you may be forced to use scoped regardless.  Note as well, that the hosting code is still work in progress and needs further review.  We certainly need to make it possible to allow the caller to choose the scoping of the registered agent.

marcominerva:
[...]

PR #4454, opened 2026-03-04
.NET: Switch auth sample to use Singletons

westey-m:
### Motivation and Context

#4446

### Description

- Switch auth sample to use Singletons

### Contribution Checklist


dotnet/samples/05-end-to-end/AspNetAgentAuthorization/Service/UserContext.cs
```diff
@@ -27,43 +27,67 @@ public interface IUserContext
...
-            : new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+    public IReadOnlySet<string> Scopes
+    {
+        get
+        {
+            string? scopeClaim = this.CurrentUser?.FindFirstValue("scope");
+            return scopeClaim is not null
+                ? new HashSet<string>(scopeClaim.Split(' ', StringSplitOptions.RemoveEmptyEntries), StringComparer.OrdinalIgnoreCase)
+                : new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+        }
```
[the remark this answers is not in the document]
westey-m: Fixed.

dotnet/samples/05-end-to-end/AspNetAgentAuthorization/Service/Program.cs
```diff
@@ -76,9 +76,9 @@
 string model = builder.Configuration["OPENAI_MODEL"] ?? "gpt-4.1-mini";
 
 builder.Services.AddHttpContextAccessor();
-builder.Services.AddScoped<IUserContext, KeycloakUserContext>();
-builder.Services.AddScoped<ExpenseService>();
-builder.Services.AddScoped<AIAgent>(sp =>
+builder.Services.AddSingleton<IUserContext, KeycloakUserContext>();
```
rogerbarreto, line 84: I think UserContext & PII should live only during the HttpContext lifecycle and not persist into the memory via a Singleton.

We might consider having the same for agents unless we want to have a single agent per service which I don't think is the common scenario for `Service` IMHO.

westey-m: The KeycloakUserContext is stateless and doesn't persist anything across requests.  It reads the current request's HttpContext each time it is used, and caches the results of parsing the token in the per request HttpContext.
Having Singletons does have the benefit of fewer allocations, since we don't rebuild the agent stack on each request.

PR #4492, opened 2026-03-05
Python: Add Google AI (Gemini) connector

Jeyaramjj:
## Summary

Adds a new Google AI connector (`agent-framework-google`) for the Gemini API, following the same patterns established by the Anthropic connector.

- **`GoogleAIChatClient`** — full-featured chat client supporting text, streaming, function calling, images (base64 + URI), and Google-hosted tools (Google Search, Code Execution)
- **`GoogleAISettings`** — TypedDict-based settings via `load_settings()` with `GOOGLE_AI_` env prefix
- **`GoogleAIChatOptions`** — chat options with Google-specific fields (`top_k`, `candidate_count`)
- **59 unit tests** covering settings, client init, message conversion, tool preparation, config creation, response/stream processing, and async API
- **5 integration tests** (marker-gated, require `GOOGLE_AI_API_KEY`)

### Key design decisions

- Inherits from `ChatMiddlewareLayer`, `FunctionInvocationLayer`, `ChatTelemetryLayer`, `BaseChatClient` — matching the framework's class hierarchy
- Uses `google-genai` SDK (`google.genai.Client`) with async API (`aio.models.generate_content`)
- Tracks function names via `_function_name_map` since Google requires function name (not call_id) in `function_response` parts
- Shared `_parse_parts_from_google()` for both streaming and non-streaming response parsing
- Google-specific usage metadata: `google.cached_content_token_count`, `google.thoughts_token_count`
- OTEL provider set to `gcp.gemini`
- Passes through `response_format` from options to `ChatResponse` for structured output support

### Files changed

| File | Description |
|------|-------------|
| `python/packages/google/agent_framework_google/_chat_client.py` | Main connector implementation (868 lines) |
| `python/packages/google/agent_framework_google/__init__.py` | Public exports (`GoogleAIChatClient`, `GoogleAIChatOptions`) |
| `python/packages/google/pyproject.toml` | Package metadata, deps: `agent-framework-core>=1.0.0rc3`, `google-genai>=0.2,<1` |
| `python/packages/google/tests/test_google_chat_client.py` | 54 unit tests + 5 integration tests |
| `python/packages/google/tests/test_settings.py` | 5 settings unit tests |
| `python/packages/google/tests/conftest.py` | Test fixtures (env setup, mock Google client) |
| `python/pyproject.toml` | Workspace source entries for new packages |

## Test plan

- [x] All 59 unit tests pass (`pytest --cov`)
- [x] 5 integration tests skipped without API key (marker-gated)
- [x] Ruff lint clean
- [x] Follows Anthropic connector patterns for consistency
- [x] Reviewed against Google GenAI SDK docs and best practices

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Jeyaramjj:
@eavanvalkenburg Please help review

python/packages/google/agent_framework_google/_chat_client.py
```diff
@@ -0,0 +1,868 @@
...
+    bound=TypedDict,  # type: ignore[valid-type]
+    default="GoogleAIChatOptions",
+    covariant=True,
+)
+
+
+# region GoogleAIChatClient
+
+
+class GoogleAIChatClient(
```
eavanvalkenburg, line 158: we should apply the pattern that we have in other clients, with a `RawGoogleAIChatClient` that has the implementation and then the "regular" `GoogleAIChatClient` that has the layers built-in and adds the parameters for function invocation, middleware and telemetry, see any of the others for a example

python/packages/google/agent_framework_google/_chat_client.py
```diff
@@ -0,0 +1,868 @@
...
+
+        # Initialize instance variables
+        self.google_client = google_client
+        self.model_id = google_settings["chat_model_id"]
+        self._function_name_map: dict[str, str] = {}
+
+    # region Static factory methods for hosted tools
+
+    @staticmethod
+    def get_google_search_tool() -> dict[str, Any]:
```
eavanvalkenburg, line 288: this should be `get_web_search_tool` to align with others, see the protocol for that in the _clients file

python/packages/google/agent_framework_google/_chat_client.py
```diff
@@ -0,0 +1,868 @@
...
+
+                from agent_framework_google import GoogleAIChatClient
+
+                tool = GoogleAIChatClient.get_google_search_tool()
+                agent = GoogleAIChatClient().as_agent(tools=[tool])
+        """
+        return {"google_search": {}}
+
+    @staticmethod
+    def get_code_execution_tool() -> dict[str, Any]:
```
eavanvalkenburg, line 305: same with this one

python/packages/google/agent_framework_google/_chat_client.py
```diff
@@ -0,0 +1,868 @@
...
+                # Or passing in an existing client
+                from google import genai
+
+                google_client = genai.Client(api_key="your_api_key")
+                client = GoogleAIChatClient(
+                    model_id="gemini-2.5-flash",
+                    google_client=google_client,
+                )
+        """
+        google_settings = load_settings(
```
eavanvalkenburg, line 252: you can set a conditional `required` based on the presence of the client, that way you do not have to raise below

python/packages/google/agent_framework_google/_chat_client.py
```diff
@@ -0,0 +1,868 @@
...
+
+        if google_client is None:
+            if not google_settings["api_key"]:
+                raise ValueError(
+                    "Google AI API key is required. Set via 'api_key' parameter "
+                    "or 'GOOGLE_AI_API_KEY' environment variable."
+                )
+
+            google_client = genai.Client(
+                api_key=google_settings["api_key"].get_secret_value(),
```
eavanvalkenburg, line 269: just ignore the type error that this will raise

python/pyproject.toml
```diff
@@ -23,17 +23,19 @@ classifiers = [
   "Typing :: Typed",
 ]
 dependencies = [
-    "agent-framework-core[all]==1.0.0rc3",
```
eavanvalkenburg, line 26: do not change this!

python/packages/google/pyproject.toml
```diff
@@ -0,0 +1,93 @@
...
+extend = "../../pyproject.toml"
+
+[tool.coverage.run]
+omit = [
+    "**/__init__.py"
+]
+
+[tool.pyright]
+extends = "../../pyproject.toml"
+exclude = ['tests']
```
eavanvalkenburg, line 64: this needs a `include` for `agent_framework_google` because of a recent change

python/pyproject.toml
eavanvalkenburg, line 1: please make sure to update to main, there should only be a change in this file for the new package in the workspace

python/packages/google/tests/test_settings.py
eavanvalkenburg, line 1: this name is potentially conflicting with other test files with the same name, make sure to include google in the name or move the tests into `tests/google`

python/packages/google/agent_framework_google/_chat_client.py
```diff
@@ -0,0 +1,868 @@
...
+
+        Returns:
+            A dictionary with model, contents, and config keys for the Google AI client.
+        """
+        # Prepend instructions from options if they exist
+        instructions = options.get("instructions")
+        if instructions:
+            from agent_framework._types import prepend_instructions_to_messages
+
+            messages = prepend_instructions_to_messages(list(messages), instructions, role="system")
```
eavanvalkenburg, line 390: why do this if they are skipped later anyway, if anything system message should be pulled out of the list and set as instructions

python/packages/google/agent_framework_google/_chat_client.py
```diff
@@ -0,0 +1,868 @@
...
+        """
+        config_params: dict[str, Any] = {}
+
+        # Extract system instruction from all system messages
+        if messages:
+            system_instructions = [msg.text for msg in messages if msg.role == "system" and msg.text]
+            if system_instructions:
+                config_params["system_instruction"] = "\n".join(system_instructions)
+
+        # Map Agent Framework options to Google AI config
```
eavanvalkenburg, line 433: we prefer to do this more dynamically so that people can call with customized Options dicts, those options should be passed through, so in the other `_prepare_options` methods, you can see that there is usually only a translation step (like with max_tokens here) or a more complex step with tools and tool_choice and everything else is kept as is, that way we do not limit which parameters can be sent to the api and we therefore are not blocking when google adds a new parameter

python/packages/google/agent_framework_google/_chat_client.py
```diff
@@ -0,0 +1,868 @@
...
+        if not model:
+            raise ValueError("model_id must be a non-empty string")
+
+        return {
+            "model": model,
+            "contents": contents,
+            "config": config,
+        }
+
+    def _create_config(
```
eavanvalkenburg, line 409:
```suggestion
    def _prepare_options_dict(
```

python/packages/google/agent_framework_google/_chat_client.py
```diff
@@ -0,0 +1,868 @@
...
+            if isinstance(tool_item, FunctionTool):
+                # FunctionTool has name, description, and parameters() method
+                function_declarations.append(
+                    types.FunctionDeclaration(
+                        name=tool_item.name,
+                        description=tool_item.description or "",
+                        parameters=tool_item.parameters(),
+                    )
+                )
+            elif isinstance(tool_item, dict):
```
eavanvalkenburg, line 498: this should not be needed, the above `get...tool` methods should be able to pass back any object that then is just passed through, again making sure that we are not blocking other functionality by limiting tools

python/packages/google/agent_framework_google/_chat_client.py
```diff
@@ -0,0 +1,868 @@
...
+
+        for content in message.contents:
+            match content.type:
+                case "text":
+                    if content.text:
+                        google_parts.append({"text": content.text})
+                case "function_call":
+                    args = content.arguments
+                    if isinstance(args, str):
+                        from contextlib import suppress
```
eavanvalkenburg, line 609: don't import this here, at the top, this loop is called many times and it's standard lib function

python/packages/google/agent_framework_google/_chat_client.py
```diff
@@ -0,0 +1,868 @@
...
+        """
+        google_parts: list[dict[str, Any]] = []
+
+        for content in message.contents:
+            match content.type:
+                case "text":
+                    if content.text:
+                        google_parts.append({"text": content.text})
+                case "function_call":
+                    args = content.arguments
```
eavanvalkenburg, line 607: you can also just leverage `content.parse_arguments()` for this

python/packages/google/agent_framework_google/_chat_client.py
```diff
@@ -0,0 +1,868 @@
...
+                    if isinstance(args, str):
+                        from contextlib import suppress
+
+                        with suppress(json.JSONDecodeError):
+                            args = json.loads(args)
+                    if isinstance(args, Mapping):
+                        args = dict(args)
+                    google_parts.append({
+                        "function_call": {"name": content.name, "args": args or {}}
```
eavanvalkenburg, line 616: will this not be json serialized anyway, any way we can just pass through the str?

python/packages/google/agent_framework_google/_chat_client.py
```diff
@@ -0,0 +1,868 @@
...
+                    Content.from_text(
+                        text=part.text,
+                        raw_representation=part,
+                    )
+                )
+            # Check for function call content
+            elif hasattr(part, "function_call") and part.function_call:
+                fc = part.function_call
+                # Google doesn't provide a call ID, so we generate one
+                call_id = str(uuid.uuid4())
```
eavanvalkenburg, line 702: call_id's are not transferrable between providers, so just using `name` with a index is probably better, then we can still track multiple calls, but you can then strip off the index from the call id in the prepare messages func and be done.

python/packages/google/agent_framework_google/_chat_client.py
```diff
@@ -0,0 +1,868 @@
...
+            response: The Google AI response.
+            options: The options dict used for the request.
+
+        Returns:
+            The Agent Framework chat response.
+        """
+        contents: list[Content] = []
+        finish_reason: FinishReasonLiteral | None = "stop"
+
+        if hasattr(response, "candidates") and response.candidates:
```
eavanvalkenburg, line 734: is there a feasible scenario where it does not have that `attribute`, wouldn't that indicate something majorly wrong?

python/packages/google/agent_framework_google/_chat_client.py
```diff
@@ -0,0 +1,868 @@
...
+        Args:
+            response: The Google AI response or chunk.
+
+        Returns:
+            UsageDetails dict, or None if no usage metadata is present.
+        """
+        if not hasattr(response, "usage_metadata") or not response.usage_metadata:
+            return None
+
+        usage = response.usage_metadata
```
eavanvalkenburg, line 846: this all looks quite defensive, we know the type, does it have those attributes or not?

python/packages/google/agent_framework_google/_chat_client.py
```diff
@@ -0,0 +1,868 @@
...
+        if cached_count is not None:
+            result["google.cached_content_token_count"] = cached_count  # type: ignore[typeddict-unknown-key]
+        thoughts_count = getattr(usage, "thoughts_token_count", None)
+        if thoughts_count is not None:
+            result["google.thoughts_token_count"] = thoughts_count  # type: ignore[typeddict-unknown-key]
+        return result
+
+    def service_url(self) -> str:
+        """Get the service URL for the chat client."""
+        return "https://generativelanguage.googleapis.com"
```
eavanvalkenburg, line 866: is this the only option? or should we get something from the google_client?

python/packages/google/agent_framework_google/_chat_client.py
```diff
@@ -0,0 +1,868 @@
...
+            if tools_list:
+                config_params["tools"] = tools_list
+
+        # Add tool choice if provided
+        if options.get("tool_choice") is not None:
+            tool_config = self._prepare_tool_config(options)
+            if tool_config:
+                config_params["tool_config"] = tool_config
+
+        extra_params = {k: v for k, v in kwargs.items() if not k.startswith("_") and k not in {"thread", "middleware"}}
```
eavanvalkenburg, line 464: We no longer want this mechanism for extra arguments, the options themselves are used for that and can be dynamically extended, (see my other comments on that) and see this sample for how that works: https://github.com/microsoft/agent-framework/blob/main/python/samples/02-agents/typed_options.py

python/packages/google/.gitignore
eavanvalkenburg, line 1: there shouldn't be extra .gitignore files, if we need something excluded add it to the main one

eavanvalkenburg (commented):
lots of small comments, some slightly bigger. Overall looks decent, some work to do on the options typeddict handling and make sure to rebase to main

eavanvalkenburg:
no activity and we already have a google package now, this probably overlaps.

PR #4484, opened 2026-03-04
Python: Fix `as_agent()` not defaulting name/description from client properties

giles17:
## Summary

`AzureAIClient.as_agent()` and `AzureAIAgentClient.as_agent()` now fall back to `self.agent_name` and `self.agent_description` when `name`/`description` are not explicitly passed. This ensures `Agent.name` is populated for telemetry spans without requiring callers to repeat the name.

## Problem

When creating an agent via `AzureAIClient(agent_name="my_agent").as_agent(...)`, the resulting `Agent.name` was `None` because `as_agent()` never consulted the client's stored `agent_name`. This caused OpenTelemetry spans to record `agent_name=None` instead of the actual agent name.

## Fix

One-line change in each client's `as_agent()` override:
- `name=name or self.agent_name`
- `description=description or self.agent_description`

This matches the pattern already used by `AzureAIProjectAgentProvider._to_chat_agent_from_details()`, which explicitly passes the same name to both the client constructor and the `Agent` constructor.

## Tests

Added 6 new unit tests (3 per client) covering:
- Default name/description inherited from client properties
- Explicit name/description overrides client properties
- No name set anywhere results in `None`

All 191 azure-ai tests pass.

Fixes #4471

PR #4498, opened 2026-03-05
Python: fix: use agent_name for ChatAgent.name in AzureAIAgentClient.as_agent()

hobostay:
Fixes an issue where AzureAIClient.as_agent() does not use agent_name for ChatAgent.name when the name parameter is not provided.

Problem: When a user creates an AzureAIChatClient with agent_name parameter, then calls as_agent() without providing a name parameter, the agent_name should be used as the default value for the ChatAgent name.

Solution: Added a check in the as_agent() method to use self.agent_name when name is None.

File changed: python/packages/azure-ai/agent_framework_azure_ai/_chat_client.py

### DIFF
--- a/python/packages/azure-ai/agent_framework_azure_ai/_chat_client.py
+++ b/python/packages/azure-ai/agent_framework_azure_ai/_chat_client.py
@@ -1486,6 +1486,9 @@ class AzureAIAgentClient(
         Returns:
             A Agent instance configured with this chat client.
         """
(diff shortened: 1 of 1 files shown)


dmytrostruk:
Looks like a duplicate of https://github.com/microsoft/agent-framework/pull/4484

giles17:
Thanks for the contribution but this is a duplicate of the fix made in #4484

PR #4502, opened 2026-03-05
.NET: [Breaking] Update Foundry Agents for Responses API

rogerbarreto:
### Motivation and Context

Introduce self-contained Foundry agent types (`FoundryAgent`, `FoundryVersionedAgent`, `FoundryAITool`) that eliminate the need for callers to manually construct `AIProjectClient` instances. The existing `AIProjectClient` extension methods are preserved for backward compatibility but marked obsolete.

See **ADR 0020** (`docs/decisions/0020-foundry-agent-type-naming.md`) for the full decision record.

#### New types in `Microsoft.Agents.AI.AzureAI`

| Type | Purpose |
|------|---------|
| `FoundryAgent` | Responses API wrapper — no server-side agent. Explicit `Uri endpoint` + `AuthenticationTokenProvider` + `model` constructors. |
| `FoundryVersionedAgent` | Server-side versioned agent wrapper — private ctor, async static factories (`CreateAIAgentAsync`, `GetAIAgentAsync`, `AsAIAgent`), plus `DeleteAIAgentAsync` / `DeleteAIAgentVersionAsync`. |
| `FoundryAITool` | Static `AITool` factory wrapping `AgentTool.Create*` / `ResponseTool.Create*`, removing the `((ResponseTool)...).AsAITool()` casting ceremony. |

#### Key design decisions

- **No env-var auto-discovery** — constructors and factory methods require explicit endpoint, credential, and model parameters. Samples read env vars at the top of `Program.cs` and pass them in.
- **Obsolete extension methods** — all 8 public `AzureAIProjectChatClientExtensions` methods (`CreateAIAgentAsync` ×3, `GetAIAgentAsync` ×2, `AsAIAgent` ×3) are marked `[Obsolete]` pointing to `FoundryVersionedAgent`.
- **`FoundryAgent` options ctor validates `ModelId`** — prevents constructing an invalid agent that would fail later at runtime.
- **`CreateConversationSessionAsync`** — convenience on both types that creates a server-side `ProjectConversation` and returns a linked `ChatClientAgentSession`.

#### Samples

- Renamed `FoundryResponseAgents/` → `FoundryAgents/`, old `FoundryAgents/` → `FoundryVersionedAgents/`
- All 49 Foundry sample programs updated to explicit endpoint/credential/model wiring
- READMEs rewritten to match the current API (no stale auto-discovery language)
- Non-Foundry samples that used obsolete extensions (`Agent_With_AzureAIProject`, `AgentWithRAG_Step04`, `AgentWithMemory_Step04`, `HostedWorkflow`) migrated to `FoundryVersionedAgent`

#### Tests

- **Unit tests**: New `FoundryAgentTests.cs` and `FoundryVersionedAgentTests.cs` covering constructor/factory validation, metadata, `GetService`, functional HTTP behavior, and user-agent headers.
- **Integration tests**: New conformance test suites for both `FoundryAgent` and `FoundryVersionedAgent` (run, streaming, chat-client, structured output, agent creation with tools). Old extension-method integration tests marked `[Obsolete]`. Full suite: 71 passed, 0 failed, 28 skipped.

#### Docs

- ADR 0020 updated to "accepted" status with the final explicit-configuration design (Option 6).

### Contribution Checklist

- [ ] **Is this a breaking change?** No — obsolete extension methods still compile and work; this is additive.


dotnet/samples/02-agents/FoundryResponseAgents/FoundryAgentsRAPI_Step02_MultiturnConversation/Program.cs
```diff
@@ -0,0 +1,41 @@
...
+
+using Azure.Identity;
+using Microsoft.Agents.AI;
+using Microsoft.Agents.AI.AzureAI;
+
+string endpoint = Environment.GetEnvironmentVariable("AZURE_AI_PROJECT_ENDPOINT") ?? throw new InvalidOperationException("AZURE_AI_PROJECT_ENDPOINT is not set.");
+string deploymentName = Environment.GetEnvironmentVariable("AZURE_AI_MODEL_DEPLOYMENT_NAME") ?? "gpt-4o-mini";
+
+// Create a FoundryAgentClient that uses the Responses API directly.
+// No server-side agent is created — instructions and model are provided locally.
```
alliscode: nit: We should avoid talking about what the samples used to do like 'no server-side agent`. Lot's more of this issue elsewhere.
```suggestion
// Create a FoundryAgentClient
```

docs/decisions/0020-foundry-agent-type-naming.md
```diff
@@ -0,0 +1,231 @@
...
+2. **Versioned Agents** — Server-side agent definitions (managed by the Foundry service) that are created, versioned, and persisted. These come in four `AgentKind` variants: `Prompt`, `ContainerApp`, `Hosted`, and `Workflow`.
+
+We currently have `FoundryResponsesAgent` (wrapping the Responses API path). We need a clear, consistent naming scheme that:
+- Distinguishes between the two interaction models
+- Is intuitive for developers new to the framework
+- Leaves room for future agent kinds without naming conflicts
+- Aligns with the broader MAF naming conventions (`ChatClientAgent`, `AzureAIProjectChatClient`, etc.)
+
+## Decision Drivers
+
```
qubitron, line 22: The consistency principle should include consistency _within_ agent framework on how agents work with other providers. Other providers have no equivalent of a `FoundryResponsesAgent` or `FoundryAgent` class (see [Anthropic example](https://github.com/microsoft/agent-framework/blob/main/dotnet/samples/02-agents/AgentWithAnthropic/Agent_Anthropic_Step01_Running/Program.cs), [OpenAI example](https://github.com/microsoft/agent-framework/blob/main/dotnet/samples/02-agents/AgentWithOpenAI/Agent_OpenAI_Step01_Running/Program.cs)). 

This also violates the clarity principle, because it's not an accurate reflection on what it does. The agent is not a foundry agent, it's a locally agent framework agent that uses foundry models and tools (via a foundry responses chat client).

Instead, use a FoundryResponsesChatClient class and use this to create an AIAgent: 
```csharp
AIAgent agent = new FoundryResponsesChatClient(endpoint=, credential=, model=)
    .AsAIAgent(instructions: "You are good at telling jokes.", name: "Joker");
```
This hits the simplicity, you're looking for while retaining clarity of what's actually happening at each layer. This is a local agent that is using a foundry model. I can swap the FoundryResponsesChatClient to an Anthropic or OpenAI chat client and the rest is the code is the same.

In this case, it's totally fine to make FoundryResponsesChatClient() hide the AIProjectClient, because it's a higher-level abstraction on a subset of the AIProjectClient capabilities, and complements without duplicating the existing functionality in Foundry SDK.

Use FoundryAgent exclusively for calling an existing Foundry agent:
```csharp
AIAgent agent = new FoundryAgent(endpoint=, credential=, agent_name=, version=)
```
Duplicating the entire create agent surface area from Foundry is a slippery slope and creates too much duplicate surface area.

dotnet/samples/02-agents/AgentsWithFoundry/Agent_Step01_Basics/Program.cs
```diff
@@ -0,0 +1,26 @@
...
+
+AIAgent agent =
+    new AIProjectClient(new Uri(endpoint), new DefaultAzureCredential())
+    .AsAIAgent(model: deploymentName, instructions: "You are good at telling jokes.", name: "JokerAgent");
+
+ProjectResponsesClient projectResponsesClient = new AIProjectClient(new Uri(endpoint), new DefaultAzureCredential())
+    .GetProjectOpenAIClient()
+    .GetProjectResponsesClientForModel(deploymentName);
+
+AIAgent agent2 = new ChatClientAgent(chatClient: projectResponsesClient.AsIChatClient(deploymentName), instructions: "You are good at telling jokes.", name: "JokerAgent2");
```
westey-m: Should we be calling out here that these two are equivalent and just alternative ways of doing the same thing?
I don't think it's clear otherwise why we have two agents in the sample.

dotnet/src/Microsoft.Agents.AI.AzureAI/FoundryAgent.cs
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    #region Convenience methods
+
+    /// <summary>
+    /// Creates a server-side conversation session that appears in the Foundry Project UI.
+    /// </summary>
+    /// <param name="cancellationToken">A token to monitor for cancellation requests.</param>
+    /// <returns>A <see cref="ChatClientAgentSession"/> linked to the newly created server-side conversation.</returns>
+    public async Task<ChatClientAgentSession> CreateConversationSessionAsync(CancellationToken cancellationToken = default)
```
westey-m, line 113: I'm wondering if this should be the default for CreateSessionCoreAsync as well?
E.g. when people are using a FoundryAgent like this, do they expect to use foundry conversations by default?

dotnet/src/Microsoft.Agents.AI.AzureAI/FoundryAgent.cs
```diff
@@ -0,0 +1,215 @@
...
+/// Foundry-specific features such as server-side conversation management via
+/// <see cref="CreateConversationSessionAsync(CancellationToken)"/>.
+/// </para>
+/// <para>
+/// Instances can be created directly via public constructors or through
+/// <c>AsAIAgent</c> extension methods on <see cref="AIProjectClient"/>.
+/// </para>
+/// </remarks>
+[Experimental(DiagnosticIds.Experiments.AIOpenAIResponses)]
+public sealed class FoundryAgent : AIAgent
```
westey-m: Not sure if this would cause issues with the constructors, but also consider inheriting from DelegatingAIAgent, so that you don't have to implement all the delegations to the innerAgent yourself.

dotnet/src/Microsoft.Agents.AI.AzureAI/FoundryAgent.cs
rogerbarreto: Since we have this `Conversation` dedicated method proposed as part of the API I would keep the other method doing what the Core does so it's clear both are doing differently, and the behavior of the other method aligns with all our other provider samples that uses the RAPI.

dotnet/src/Microsoft.Agents.AI.AzureAI/FoundryAgent.cs
rogerbarreto: Yeah, I think I did considered that, not sure why is not using it, good call out I will ensure the impl uses it.

dotnet/samples/02-agents/AgentsWithFoundry/Agent_Step01_Basics/Program.cs
rogerbarreto: Good point, I may just remove from the sample for simplicity and put this in the README.md as composable way of doing it.

dotnet/src/Microsoft.Agents.AI.AzureAI/AzureAIProjectChatClientExtensions.cs
```diff
@@ -42,7 +43,7 @@ public static partial class AzureAIProjectChatClientExtensions
     /// When instantiating a <see cref="ChatClientAgent"/> by using an <see cref="AgentReference"/>, minimal information will be available about the agent in the instance level, and any logic that relies
     /// on <see cref="AIAgent.GetService{TService}(object?)"/> to retrieve information about the agent like <see cref="AgentVersion" /> will receive <see langword="null"/> as the result.
     /// </remarks>
-    public static ChatClientAgent AsAIAgent(
+    public static FoundryAgent AsAIAgent(
```
SergeyMenshykh, line 46: This is a breaking change, so the PR title should indicate that.

dotnet/src/Microsoft.Agents.AI.AzureAI/FoundryAgent.cs
```diff
@@ -0,0 +1,202 @@
...
+    /// <param name="clientOptions">Optional configuration options for the <see cref="AIProjectClient"/>.</param>
+    /// <param name="name">Optional name for the agent.</param>
+    /// <param name="description">Optional description for the agent.</param>
+    /// <param name="tools">Optional tools to use when interacting with the agent.</param>
+    /// <param name="clientFactory">Provides a way to customize the creation of the underlying <see cref="IChatClient"/>.</param>
+    /// <param name="loggerFactory">Optional logger factory for creating loggers used by the agent.</param>
+    /// <param name="services">Optional service provider for resolving dependencies required by AI functions.</param>
+    public FoundryAgent(
+        Uri projectEndpoint,
+        AuthenticationTokenProvider credential,
```
SergeyMenshykh, line 55: I wonder if the `credential` parameter should go after the `instructions` one? This will group `projectEndpoint`, `model`, and `instructions` as "what" while the `credential` will follow as "how"

dotnet/samples/02-agents/FoundryAgents/FoundryAgents_Step04_UsingFunctionToolsWithApprovals/Program.cs
```diff
@@ -1,45 +1,34 @@
...
 
-const string AssistantInstructions = "You are a helpful assistant that can get weather information.";
-const string AssistantName = "WeatherAssistant";
+string endpoint = Environment.GetEnvironmentVariable("AZURE_AI_PROJECT_ENDPOINT") ?? throw new InvalidOperationException("AZURE_AI_PROJECT_ENDPOINT is not set.");
+string deploymentName = Environment.GetEnvironmentVariable("AZURE_AI_MODEL_DEPLOYMENT_NAME") ?? "gpt-4o-mini";
 
-// Get a client to create/retrieve/delete server side agents with Azure Foundry Agents.
-// WARNING: DefaultAzureCredential is convenient for development but requires careful consideration in production.
-// In production, consider using a specific credential (e.g., ManagedIdentityCredential) to avoid
-// latency issues, unintended credential probing, and potential security risks from fallback mechanisms.
```
westey-m, line 28: Should we keep the warning?

dotnet/samples/02-agents/AgentsWithFoundry/Agent_Step06_PersistedConversations/Program.cs
```diff
@@ -10,16 +10,11 @@
 string endpoint = Environment.GetEnvironmentVariable("AZURE_AI_PROJECT_ENDPOINT") ?? throw new InvalidOperationException("AZURE_AI_PROJECT_ENDPOINT is not set.");
 string deploymentName = Environment.GetEnvironmentVariable("AZURE_AI_MODEL_DEPLOYMENT_NAME") ?? "gpt-4o-mini";
 
-const string JokerInstructions = "You are good at telling jokes.";
-const string JokerName = "JokerAgent";
-
-// Get a client to create/retrieve/delete server side agents with Azure Foundry Agents.
-// WARNING: DefaultAzureCredential is convenient for development but requires careful consideration in production.
```
westey-m, line 17: Should we keep the warning?

dotnet/samples/02-agents/AgentsWithFoundry/Agent_Step07_Observability/Program.cs
```diff
@@ -28,14 +24,13 @@
 }
 using var tracerProvider = tracerProviderBuilder.Build();
 
-// Get a client to create/retrieve/delete server side agents with Azure Foundry Agents.
-// WARNING: DefaultAzureCredential is convenient for development but requires careful consideration in production.
```
westey-m, line 32: Looks like there's a few places where we may not having the Default credential warning anymore.

dotnet/samples/02-agents/AgentsWithFoundry/Agent_Step09_UsingMcpClientAsTools/README.md
```diff
@@ -0,0 +1,29 @@
+﻿# Using MCP Client as Tools with the Responses API
+
+This sample shows how to use MCP (Model Context Protocol) client tools with a `ChatClientAgent` using the Responses API directly.
+
+## What this sample demonstrates
+
+- Connecting to an MCP server (GitHub) via stdio transport
```
westey-m:
```suggestion
- Connecting to an MCP server (GitHub) via http client transport
```

dotnet/samples/02-agents/AgentsWithFoundry/Agent_Step02_MultiturnConversation/Program.cs
```diff
@@ -0,0 +1,43 @@
...
+    deploymentName,
+    instructions: "You are good at telling jokes.",
+    name: "JokerAgent");
+
+ProjectConversation conversation = await aiProjectClient
+    .GetProjectOpenAIClient()
+    .GetProjectConversationsClient()
+    .CreateProjectConversationAsync();
+
+AgentSession session = await agent.GetService<ChatClientAgent>()!.CreateSessionAsync(conversation.Id);
```
westey-m: We use session in quite a few places without conversations.  This seems to be the one place we use conversations, and the sample is about multi-turn conversations.
This may lead reader to think that to use multi-turn conversations they have to use the conversations service, but that's not the case.  Consider separating this into something that demonstrates session with response ids (simple), and session with the conversations service (more complex)

dotnet/samples/02-agents/AgentsWithFoundry/Agent_Step23_LocalMCP/Program.cs
```diff
@@ -0,0 +1,74 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+// This sample demonstrates how to use a local MCP (Model Context Protocol) client with a ChatClientAgent.
```
westey-m: This sample seems pretty much the same as Step09

dotnet/samples/02-agents/AgentWithAnthropic/Agent_Anthropic_Step01_Running/Program.cs
```diff
@@ -10,15 +10,13 @@
...
-AIAgent agent = new AnthropicClient(new ClientOptions { ApiKey = apiKey })
+AIAgent agent =
+    new AnthropicClient(new ClientOptions { ApiKey = apiKey })
     .AsAIAgent(model: model, instructions: "You are good at telling jokes.", name: "Joker");
 
-// Invoke the agent and output the text result.
-var response = await agent.RunAsync("Tell me a joke about a pirate.");
-Console.WriteLine(response);
+AnthropicClient anthropicClient = new(new ClientOptions { ApiKey = apiKey });
+AIAgent agent2 = new ChatClientAgent(anthropicClient.AsIChatClient());
```
westey-m: Similar to the earlier comment, we have two agents here.  We should maybe explain why or simplify.

dotnet/samples/02-agents/AgentWithOpenAI/Agent_OpenAI_Step01_Running/Program.cs
```diff
@@ -4,28 +4,20 @@
...
-AIAgent agent = new OpenAIClient(apiKey)
-    .GetChatClient(model)
-    .AsAIAgent(instructions: "You are good at telling jokes.", name: "Joker");
+AIAgent agent =
+    new ResponsesClient(new ApiKeyCredential(apiKey))
+    .AsAIAgent(model: model, instructions: "You are good at telling jokes.", name: "Joker");
 
-UserChatMessage chatMessage = new("Tell me a joke about a pirate.");
+var responsesClient = new ResponsesClient(new ApiKeyCredential(apiKey));
+AIAgent agent2 = new ChatClientAgent(responsesClient.AsIChatClient());
```
westey-m: Same as above, two agents.

dotnet/samples/02-agents/AgentWithRAG/AgentWithRAG_Step04_FoundryServiceRAG/Program.cs
```diff
@@ -39,12 +42,15 @@
...
-        instructions: "You are a helpful support specialist for Contoso Outdoors. Answer questions using the provided context and cite the source document when available.",
-        tools: [fileSearchTool]);
+AgentVersion agentVersion = await aiProjectClient.Agents.CreateAgentVersionAsync(
+    "AskContoso",
+    new AgentVersionCreationOptions(
+        new PromptAgentDefinition(model: deploymentName)
+        {
+            Instructions = "You are a helpful support specialist for Contoso Outdoors. Answer questions using the provided context and cite the source document when available.",
+            Tools = { fileSearchTool.GetService<ResponseTool>() ?? fileSearchTool.AsOpenAIResponseTool() ?? throw new InvalidOperationException("Unable to convert hosted file search tool to a ResponseTool.") }
+        }));
```
westey-m, line 54: Don't we have our own version of CreateAgentVersionAsync anymore?  If so, one thing that is problematic with only having the project client version is that when creating an agent with function tools, there isn't an easy way of generating the tool definitions from native code.

dotnet/samples/02-agents/Agents/Agent_Step11_Middleware/Program.cs
```diff
@@ -309,3 +309,9 @@ protected override ValueTask<IEnumerable<ChatMessage>> ProvideMessagesAsync(
             ]);
     }
 }
+
+internal partial class Program
```
SergeyMenshykh, line 313: Is it necessary to generate only one regex out of three?

dotnet/samples/02-agents/ModelContextProtocol/FoundryAgent_Hosted_MCP/Program.cs
```diff
@@ -32,17 +33,14 @@
...
+    "MicrosoftLearnAgent",
+    new AgentVersionCreationOptions(
+        new PromptAgentDefinition(model: model)
         {
             Instructions = "You answer questions by searching the Microsoft Learn content only.",
-            Tools = [mcpTool]
-        },
-    });
+        }));
+AIAgent agent = aiProjectClient.AsAIAgent(agentVersion, tools: [mcpTool]);
```
westey-m: Does this work?  We didn't advertise the mcp tool when we created the server side agent, so I'm expecting that just passing it here would fail when we run the sample?

dotnet/src/Microsoft.Agents.AI.AzureAI/FoundryAgent.cs
```diff
@@ -0,0 +1,202 @@
...
+using Microsoft.Extensions.Logging;
+using Microsoft.Shared.DiagnosticIds;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Agents.AI.AzureAI;
+
+/// <summary>
+/// Provides an <see cref="AIAgent"/> that uses Microsoft Foundry for AI agent capabilities.
+/// </summary>
+/// <remarks>
```
westey-m, line 17: Might be useful to explain here in comments how this differs from a foundry based agent that doesn't use the persistent service side prompt agent.

dotnet/samples/02-agents/AgentsWithFoundry/Agent_Step06_PersistedConversations/Program.cs
```diff
@@ -42,6 +37,3 @@
 
 // Run the agent again with the resumed session.
 Console.WriteLine(await agent.RunAsync("Now tell the same joke in the voice of a pirate, and add some emojis to the joke.", resumedSession));
-
-// Cleanup by agent name removes the agent version created.
-await aiProjectClient.Agents.DeleteAgentAsync(agent.Name);
```
SergeyMenshykh, line 47: Don't we need to remove the agent anymore?

dotnet/src/Microsoft.Agents.AI.AzureAI/FoundryAgent.cs
rogerbarreto: I put those together only for the overrides where we own the projectClient as those two are tight correlated and required.

This definition also may collide for optional vs and required parameters, where for example the agent name is what, coming as nullable after how `required` credential.

dotnet/samples/02-agents/ModelContextProtocol/FoundryAgent_Hosted_MCP/Program.cs
rogerbarreto: Good point, I updated the sample using the non-RAPI approach.

dotnet/samples/02-agents/AgentWithRAG/AgentWithRAG_Step04_FoundryServiceRAG/Program.cs
rogerbarreto: Actually there's, I will remove the `AITool` from the sample, we should use the SDK Native for the creation process.

dotnet/samples/02-agents/AgentsWithFoundry/Agent_Step23_LocalMCP/Program.cs
rogerbarreto: - Step09: Basic MCP: connects to Microsoft Learn MCP server, passes McpClientTool directly as AITool to the agent. Simplest MCP usage.
 - Step23: MCP with tool wrapping: same MCP server, but wraps each tool with LoggingMcpTool (a DelegatingAIFunction) to log local invocations. 
Demonstrates the DelegatingAIFunction pattern for diagnostics/metering/approval around tool calls.

dotnet/samples/02-agents/AgentsWithFoundry/Agent_Step02_MultiturnConversation/Program.cs
rogerbarreto: good point, splitter the step02 in two for each scenario.

dotnet/src/Microsoft.Agents.AI.AzureAI/FoundryAgent.cs
rogerbarreto: Make sense, addressed

docs/decisions/0020-foundry-agent-type-naming.md
```diff
@@ -0,0 +1,125 @@
...
+deciders: rogerbarreto, alliscode
+consulted: ""
+informed: ""
+---
+
+# Foundry agent surface stays centered on `ChatClientAgent`
+
+## Context
+
+The Azure AI Foundry integration exposes two distinct usage patterns:
```
SergeyMenshykh: I still see the name "Azure AI Foundry" being used. Please replace it with the new one.

dotnet/samples/02-agents/AgentsWithFoundry/Agent_Step02.2_MultiturnWithServerConversations/README.md
```diff
@@ -0,0 +1,36 @@
...
+Set the following environment variables:
+
+```powershell
+$env:AZURE_AI_PROJECT_ENDPOINT="https://your-foundry-service.services.ai.azure.com/api/projects/your-foundry-project"
+$env:AZURE_AI_MODEL_DEPLOYMENT_NAME="gpt-4o-mini"
+```
+
+## Run the sample
+
+Navigate to the ChatClientAgents sample directory and run:
```
SergeyMenshykh: Shouldn't it be Navigate to the **AgentsWithFoundry** sample directory and run: ? Same in several other readmes.

dotnet/samples/02-agents/AgentsWithFoundry/Agent_Step00_FoundryAgentLifecycle/Program.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
SergeyMenshykh, line 1: This sample name is ﻿Inconsistent with Agent_StepNN_* samples name pattern and can be renamed to smth like: ﻿Agent_Step00_FoundryAgentLifecycle

dotnet/samples/02-agents/AgentsWithFoundry/Agent_Step02.2_MultiturnWithServerConversations/Program.cs
```diff
@@ -0,0 +1,34 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+// This sample shows how to use server-side conversations with a FoundryAgent.
+// Server-side conversations persist on the Foundry service and are visible in the Foundry Project UI.
+// Use this when you need conversation history to be stored and accessible server-side.
```
westey-m, line 5: Can probably do this in a follow up PR, but we need to clearly differentiate between the conversations service and response ids, which are both server-side conversations.
Right now step 2.1 is also a server side conversation, but using response ids, whereas this sample uses the conversations service.

PR #4533, opened 2026-03-06
.NET Compaction - Introducing compaction strategies and pipeline

crickman:
### Motivation and Context

Context management (compaction) is one of the key features in the "dev-harness" effort. This change introduces structured handling of long-running AI chat conversations by compacting historical context while preserving key decisions and intent. By reducing token growth and context drift, it improves response quality, performance, and cost predictability over extended sessions. The implementation is designed to be extensible and transparent, making context lifecycle management a first‑class concern for agent development.

[[Spec](https://github.com/microsoft/agent-framework/blob/main/docs/decisions/0019-python-context-compaction-strategy.md)]

### Description

The goal of this approach is inject compaction as part of the `IChatClient` architecture.  This approach maintains an index that allows for presenting a compacted version of the conversation without modifying the source "chat history".  

**Features**
- Built-in compaction strategies 
- Supports pipeline strategy
- Supports direct invocation
- Supports `IChatReducer`
- State retained with session serialization
- Incremental message processing
- Atomic tool-call grouping
- Includes tool loops
- Retains original chat history

### Details

Compaction occurs via a `CompactionStrategy`.  A set of strategies are incuded as part of this initial release, including a _pipeline_ strategy that is able to sequentially apply one or more strategies:

| Strategy | Aggressiveness | Preserves context | Requires LLM | Best for |
|---|---|---|---|---|
| `ToolResultCompactionStrategy` | Low | High — only collapses tool results | No | Reclaiming space from verbose tool output |
| `SummarizationCompactionStrategy` | Medium | Medium — replaces history with a summary | Yes | Long conversations where context matters |
| `SlidingWindowCompactionStrategy` | High | Low — drops entire turns | No | Hard turn-count limits |
| `TruncationCompactionStrategy` | High | Low — drops oldest groups | No | Emergency token-budget backstops |
| `PipelineCompactionStrategy` | Configurable | Depends on child strategies | Depends | Layered compaction with multiple fallbacks |
| `ChatReducerCompactionStrategy` | Configurable | Depends on the `IChatReducer` | Depends | Compact using an existing `IChatReducer` |

### Code

```c#
// Setup a chat client for summarization
IChatClient summarizingChatClient = ...;

// Configure the compaction pipeline with one of each strategy, ordered least to most aggressive.
PipelineCompactionStrategy compactionPipeline =
    new(// 1. Gentle: collapse old tool-call groups into short summaries like "[Tool calls: LookupPrice]"
        new ToolResultCompactionStrategy(CompactionTriggers.TokensExceed(0x200)),

        // 2. Moderate: use an LLM to summarize older conversation spans into a concise message
        new SummarizationCompactionStrategy(summarizingChatClient, CompactionTriggers.TokensExceed(0x8000)),

        // 3. Aggressive: keep only the last N user turns and their responses
        new SlidingWindowCompactionStrategy(CompactionTriggers.TurnsExceed(4)),

        // 4. Emergency: drop oldest groups until under the token budget
        new TruncationCompactionStrategy(CompactionTriggers.GroupsExceed(0x1000)));

AIAgent agent =
    agentChatClient.AsAIAgent(
        new ChatClientAgentOptions
        {
            Name = "ShoppingAssistant",
            ChatOptions = new()
            {
                Instructions = "...",
                Tools = [...],
            },
            AIContextProviders = [new CompactionProvider(compactionPipeline)],
        });
```

or

```c#
AIAgent agent =
    agentChatClient
        .AsBuilder()
        .UseAIContextProviders(new CompactionProvider(compactionPipeline))
        .BuildAIAgent(
            new ChatClientAgentOptions
            {
                Name = "ShoppingAssistant",
                ChatOptions = new()
                {
                    Instructions = "...",
                    Tools = [...],
                }
            });
```

### Contribution Checklist

### DIFF
--- a/python/packages/purview/README.md
+++ b/python/packages/purview/README.md
@@ -29,7 +29,7 @@ Add Purview when you need to:
 
 - Microsoft Azure subscription with Microsoft Purview configured.
 - Microsoft 365 subscription with an E5 license and pay-as-you-go billing setup.
-  - For testing, you can use a Microsoft 365 Developer Program tenant. For more information, see [Join the Microsoft 365 Developer Program](https://developer.microsoft.com/en-us/microsoft-365/dev-program).
+  - For testing, you can use a Microsoft 365 Developer Program tenant. For more information, see [Join the Microsoft 365 Developer Program](https://learn.microsoft.com/en-us/office/developer-program/microsoft-365-developer-program).
 
 ### Authentication
 


dotnet/src/Microsoft.Agents.AI/Compaction/PipelineCompactionStrategy.cs
```diff
@@ -0,0 +1,62 @@
...
+    /// Initializes a new instance of the <see cref="PipelineCompactionStrategy"/> class.
+    /// </summary>
+    /// <param name="strategies">The ordered sequence of strategies to execute.</param>
+    public PipelineCompactionStrategy(params IEnumerable<CompactionStrategy> strategies)
+        : base(CompactionTriggers.Always)
+    {
+        this.Strategies = [.. Throw.IfNull(strategies)];
+    }
+
+    /// <summary>
```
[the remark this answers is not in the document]
crickman: `params IEnumerable<>` works fine.

dotnet/src/Microsoft.Agents.AI/Compaction/MessageIndex.cs
```diff
@@ -0,0 +1,381 @@
...
+
+    /// <summary>
+    /// Gets the total number of groups across all included (non-excluded) groups that are not <see cref="MessageGroupKind.System"/>.
+    /// </summary>
+    public int IncludedNonSystemGroupCount => this.Groups.Count(g => !g.IsExcluded && g.Kind != MessageGroupKind.System);
+
+    /// <summary>
+    /// Returns all groups that belong to the specified user turn.
+    /// </summary>
+    /// <param name="turnIndex">The zero-based turn index.</param>
```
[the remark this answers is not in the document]
crickman: Its zero based...anything before the first user message is "turn 0"

dotnet/src/Microsoft.Agents.AI/Compaction/MessageIndex.cs
```diff
@@ -0,0 +1,381 @@
...
+    {
+        int total = 0;
+        for (int i = 0; i < messages.Count; i++)
+        {
+            string text = messages[i].Text;
+            if (text.Length > 0)
+            {
+                total += Encoding.UTF8.GetByteCount(text);
+            }
+        }
```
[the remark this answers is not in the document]
crickman: Its never null...maybe those other spots care if its empty...we don't.

dotnet/src/Microsoft.Agents.AI/Compaction/CompactionMessageIndex.cs
```diff
@@ -0,0 +1,381 @@
...
+    {
+        foreach (AIContent content in message.Contents)
+        {
+            if (content is FunctionCallContent)
+            {
+                return true;
+            }
+        }
+
+        return false;
```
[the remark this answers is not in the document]
crickman: Contents is never null, the get accessor ensures that

dotnet/src/Microsoft.Agents.AI/Compaction/MessageIndex.cs
```diff
@@ -0,0 +1,381 @@
...
+    {
+        int total = 0;
+        for (int i = 0; i < messages.Count; i++)
+        {
+            string text = messages[i].Text;
+            if (text.Length > 0)
+            {
+                total += tokenizer.CountTokens(text);
+            }
+        }
```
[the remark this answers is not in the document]
crickman: Text is never null.

dotnet/samples/02-agents/Agents/Agent_Step18_CompactionPipeline/Program.cs
```diff
@@ -0,0 +1,113 @@
...
+                Instructions =
+                    """
+                    You are a helpful, but long winded, shopping assistant.
+                    Help the user look up prices and compare products.
+                    When responding, Be sure to be extra descriptive and use as
+                    many words as possible without sounding ridiculous.
+                    """,
+                Tools = [AIFunctionFactory.Create(LookupPrice)],
+            },
+            AIContextProviders = [new MessageCompactionContextProvider(compactionPipeline)],
```
westey-m: Would be nice to show how to add this to the ChatClient as well (maybe as a commented out snippet) with a comment around the fact that putting it there would also compact during the function calling loop.

dotnet/src/Microsoft.Agents.AI.Abstractions/InMemoryChatHistoryProvider.cs
```diff
@@ -79,20 +79,21 @@ public List<ChatMessage> GetMessages(AgentSession? session)
...
         state.Messages = messages;
     }
 
     /// <inheritdoc />
     protected override async ValueTask<IEnumerable<ChatMessage>> ProvideChatHistoryAsync(InvokingContext context, CancellationToken cancellationToken = default)
     {
-        var state = this._sessionState.GetOrInitializeState(context.Session);
+        State state = this._sessionState.GetOrInitializeState(context.Session);
 
         if (this.ReducerTriggerEvent is InMemoryChatHistoryProviderOptions.ChatReducerTriggerEvent.BeforeMessagesRetrieval && this.ChatReducer is not null)
```
westey-m, line 93: Since you moved the chat reducer null check to CompactMessagesAsync, the check here is no redundant.
```suggestion
        if (this.ReducerTriggerEvent is InMemoryChatHistoryProviderOptions.ChatReducerTriggerEvent.BeforeMessagesRetrieval)
```

dotnet/src/Microsoft.Agents.AI/Compaction/CompactionStrategy.cs
```diff
@@ -0,0 +1,130 @@
...
+    /// <remarks>
+    /// This method is called by <see cref="CompactAsync"/> only when the <see cref="Trigger"/>
+    /// returns <see langword="true"/>. Implementations do not need to evaluate the trigger or
+    /// report metrics — the base class handles both. Implementations should use <see cref="Target"/>
+    /// to determine when to stop compacting incrementally.
+    /// </remarks>
+    /// <param name="index">The message index to compact. The strategy mutates this collection in place.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests.</param>
+    /// <returns>A task whose result is <see langword="true"/> if any compaction was performed, <see langword="false"/> otherwise.</returns>
+    protected abstract Task<bool> ApplyCompactionAsync(MessageIndex index, CancellationToken cancellationToken);
```
westey-m: We've typically been following the Microsoft.Extensions practice in other places in Microsoft.Agents.AI.Abstractions, where the overridable method is called `*CoreAsync()`.
So in this case would be `CompactCoreAsync`.

dotnet/src/Microsoft.Agents.AI/Compaction/CompactionStrategy.cs
```diff
@@ -0,0 +1,130 @@
...
+
+        Stopwatch stopwatch = Stopwatch.StartNew();
+
+        bool compacted = await this.ApplyCompactionAsync(index, cancellationToken).ConfigureAwait(false);
+
+        stopwatch.Stop();
+
+        if (compacted)
+        {
+            Debug.WriteLine(
```
westey-m: Should we support logging here?  I imagine this info may be useful in production as well.

dotnet/src/Microsoft.Agents.AI/Compaction/CompactionStrategy.cs
```diff
@@ -0,0 +1,130 @@
...
+    /// <summary>
+    /// Gets the trigger predicate that controls when compaction proceeds.
+    /// </summary>
+    protected CompactionTrigger Trigger { get; }
+
+    /// <summary>
+    /// Gets the target predicate that controls when compaction stops.
+    /// Strategies re-evaluate this after each incremental exclusion and stop when it returns <see langword="true"/>.
+    /// </summary>
+    protected CompactionTrigger Target { get; }
```
westey-m, line 79: This class isn't using target, and some implementations aren't either.  Consider removing it here and only having it on those classes that actually require it.

dotnet/src/Microsoft.Agents.AI/Compaction/CompactionTriggers.cs
```diff
@@ -0,0 +1,117 @@
...
+    /// Always trigger compaction, regardless of the message index state.
+    /// </summary>
+    public static readonly CompactionTrigger Always =
+        _ => true;
+
+    /// <summary>
+    /// Never trigger compaction, regardless of the message index state.
+    /// </summary>
+    public static readonly CompactionTrigger Never =
+        _ => false;
```
westey-m, line 35: In what situation would you use never?

dotnet/src/Microsoft.Agents.AI/Compaction/CompactionTriggers.cs
```diff
@@ -0,0 +1,117 @@
...
+    /// <summary>
+    /// Creates a trigger that fires when the included message count exceeds the specified maximum.
+    /// </summary>
+    /// <param name="maxMessages">The message threshold. Compaction proceeds when included messages exceed this value.</param>
+    /// <returns>A <see cref="CompactionTrigger"/> that evaluates included message count.</returns>
+    public static CompactionTrigger MessagesExceed(int maxMessages) =>
+        index => index.IncludedMessageCount > maxMessages;
+
+    /// <summary>
+    /// Creates a trigger that fires when the included user turn count exceeds the specified maximum.
```
westey-m, line 62: Does this refer to only agent runs or also function call loop iterations.  Or perhaps a better way of asking is, is it agent runs or downstream service calls?
If the latter, we may want to avoid referring to it as *user* turn count, since the user is not necessarily involved.

dotnet/src/Microsoft.Agents.AI/Compaction/CompactionTriggers.cs
```diff
@@ -0,0 +1,117 @@
...
+    /// </summary>
+    /// <param name="maxTurns">The turn threshold. Compaction proceeds when included turns exceed this value.</param>
+    /// <returns>A <see cref="CompactionTrigger"/> that evaluates included turn count.</returns>
+    public static CompactionTrigger TurnsExceed(int maxTurns) =>
+        index => index.IncludedTurnCount > maxTurns;
+
+    /// <summary>
+    /// Creates a trigger that fires when the included group count exceeds the specified maximum.
```
westey-m, line 82: I expect this will be the main place where users encounter groups if they are just using the built in functionality. i.e. they are unlikely to encounter MessageGroup if they are not building their own compacter.
Consider either explaining in more detail what a group is here, or add a reference to MessageGroup so that they can read more about it there.

dotnet/src/Microsoft.Agents.AI/Compaction/MessageCompactionContextProvider.cs
```diff
@@ -0,0 +1,136 @@
...
+/// to the agent's underlying chat client.
+/// </para>
+/// <para>
+/// The <see cref="MessageCompactionContextProvider"/> can be added to an agent's context provider pipeline
+/// via <see cref="ChatClientAgentOptions.AIContextProviders"/> or via <c>UseAIContextProviders</c>
+/// on a <see cref="ChatClientBuilder"/> or <see cref="AIAgentBuilder"/>.
+/// </para>
+/// </remarks>
+[Experimental(DiagnosticIds.Experiments.AgentsAIExperiments)]
+public sealed class MessageCompactionContextProvider : MessageAIContextProvider
```
westey-m: For the other AIContextProviders, we typically just shorten the implementation postfix to Provider, e.g. TextSearchProvider, FileAgentSkillsProvider, Mem0Provider, etc.

dotnet/src/Microsoft.Agents.AI/Compaction/MessageCompactionContextProvider.cs
```diff
@@ -0,0 +1,136 @@
...
+    /// An optional key used to store the provider state in the <see cref="AgentSession.StateBag"/>.
+    /// Defaults to <c>"MessageCompactionContextProvider"</c>.
+    /// </param>
+    /// <exception cref="ArgumentNullException"><paramref name="compactionStrategy"/> is <see langword="null"/>.</exception>
+    public MessageCompactionContextProvider(CompactionStrategy compactionStrategy, string? stateKey = null)
+    {
+        this._compactionStrategy = Throw.IfNull(compactionStrategy);
+        this._sessionState = new ProviderSessionState<State>(
+            _ => new State(),
+            stateKey ?? $"{nameof(MessageCompactionContextProvider)}:{Convert.ToBase64String(BitConverter.GetBytes(compactionStrategy.GetHashCode()))}",
```
westey-m: Could this cause a memory/storage (size) leak if a dev changes the compaction strategy often or if they pass the session to different agents with different strategies?  And if the dev did that, would all but the last used index in the session not be stale?

dotnet/src/Microsoft.Agents.AI/Compaction/MessageCompactionContextProvider.cs
```diff
@@ -0,0 +1,136 @@
...
+            // First pass — initialize the message index from scratch.
+            messageIndex = MessageIndex.Create(messageList);
+        }
+
+        // Apply compaction
+        Stopwatch stopwatch = Stopwatch.StartNew();
+        bool wasCompacted = await this._compactionStrategy.CompactAsync(messageIndex, cancellationToken).ConfigureAwait(false);
+        stopwatch.Stop();
+
+        Debug.WriteLine($"COMPACTION: {wasCompacted} - {stopwatch.ElapsedMilliseconds}ms");
```
westey-m: Is this needed, considering that CompactionStrategy already does timing?
We may just need to add wasCompacted to CompactionStrategy as well.

dotnet/src/Microsoft.Agents.AI/Compaction/MessageCompactionContextProvider.cs
```diff
@@ -0,0 +1,136 @@
...
+    /// <summary>
+    /// Represents the persisted state of a <see cref="MessageCompactionContextProvider"/> stored in the <see cref="AgentSession.StateBag"/>.
+    /// </summary>
+    public sealed class State
+    {
+        /// <summary>
+        /// Gets or sets the message index groups used for incremental compaction updates.
+        /// </summary>
+        [JsonPropertyName("messages")]
+        public List<MessageGroup> MessageIndex { get; set; } = [];
```
westey-m: Considering that there is a class called MessageIndex too, calling the variable here MessageIndex made me think it actually is one, when I saw it referenced elsewhere.  Also consider MessageGroups for a name to clarify.

dotnet/src/Microsoft.Agents.AI/Compaction/CompactionGroupKind.cs
```diff
@@ -0,0 +1,55 @@
...
+
+    /// <summary>
+    /// An atomic tool call group containing an assistant message with tool calls
+    /// followed by the corresponding tool result messages.
+    /// </summary>
+    /// <remarks>
+    /// This group must be treated as an atomic unit during compaction. Removing the assistant
+    /// message without its tool results (or vice versa) will cause LLM API errors.
+    /// </remarks>
+    ToolCall,
```
westey-m, line 42: If we get back a response from the service which contains both FunctionCallContent and TextContent in one message, does go into a ToolCall group, even though it has Assistant Text as well?

dotnet/src/Microsoft.Agents.AI/Microsoft.Agents.AI.csproj
```diff
@@ -18,10 +18,14 @@
...
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.Agents.AI.Abstractions\Microsoft.Agents.AI.Abstractions.csproj" />
+  </ItemGroup>
+  
+  <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.AI" />
     <PackageReference Include="Microsoft.Extensions.VectorData.Abstractions" />
     <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" />
+    <PackageReference Include="Microsoft.ML.Tokenizers" />
```
westey-m, line 28: This package pulls in Google.Protobuf as a transient dependency.  Not the best, considering that it is used very widely, and we'd neither want to be a blocker nor a forcing function for version upgrades for users.
Too bad these folks don't have an Abstraction library for Tokenizer.  We could abstract it away ourselves, but that won't lead to a great experience either.  Let's maybe discuss further to see what others think as well.

dotnet/src/Microsoft.Agents.AI/Compaction/MessageIndex.cs
```diff
@@ -0,0 +1,381 @@
...
+    /// If the message count is less than <see cref="ProcessedMessageCount"/> (e.g., after storage compaction
+    /// replaced messages with summaries), all groups are cleared and rebuilt from scratch.
+    /// </para>
+    /// <para>
+    /// If the message count equals <see cref="ProcessedMessageCount"/>, no work is performed.
+    /// </para>
+    /// </remarks>
+    public void Update(IList<ChatMessage> allMessages)
+    {
+        if (allMessages.Count == this.ProcessedMessageCount)
```
westey-m: There is also the scenario where the ChatHistoryProvider always returns at most the last x messages, and there is 1 new user message.  If we reach the x limit, the number of messages in allMessages would not change across multiple runs, but they would have changed.

dotnet/src/Microsoft.Agents.AI/Compaction/MessageIndex.cs
```diff
@@ -0,0 +1,381 @@
...
+    /// <returns>The groups belonging to the turn, in order.</returns>
+    public IEnumerable<MessageGroup> GetTurnGroups(int turnIndex) =>
+        this.Groups.Where(g => g.TurnIndex == turnIndex);
+
+    /// <summary>
+    /// Computes the UTF-8 byte count for a set of messages.
+    /// </summary>
+    /// <param name="messages">The messages to compute byte count for.</param>
+    /// <returns>The total UTF-8 byte count of all message text content.</returns>
+    public static int ComputeByteCount(IReadOnlyList<ChatMessage> messages)
```
westey-m: Should this be private?

dotnet/src/Microsoft.Agents.AI/Compaction/MessageIndex.cs
```diff
@@ -0,0 +1,381 @@
...
+        return total;
+    }
+
+    /// <summary>
+    /// Computes the token count for a set of messages using the specified tokenizer.
+    /// </summary>
+    /// <param name="messages">The messages to compute token count for.</param>
+    /// <param name="tokenizer">The tokenizer to use for counting tokens.</param>
+    /// <returns>The total token count across all message text content.</returns>
+    public static int ComputeTokenCount(IReadOnlyList<ChatMessage> messages, Tokenizer tokenizer)
```
westey-m: Should this be private?

dotnet/src/Microsoft.Agents.AI/Compaction/CompactionStrategy.cs
```diff
@@ -0,0 +1,130 @@
...
+
+    /// <summary>
+    /// Evaluates the <see cref="Trigger"/> and, when it fires, delegates to
+    /// <see cref="ApplyCompactionAsync"/> and reports compaction metrics.
+    /// </summary>
+    /// <param name="index">The message index to compact. The strategy mutates this collection in place.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests.</param>
+    /// <returns>A task representing the asynchronous operation. The task result is <see langword="true"/> if compaction occurred, <see langword="false"/> otherwise.</returns>
+    public async Task<bool> CompactAsync(MessageIndex index, CancellationToken cancellationToken = default)
```
westey-m: Also consider a ValueTask response type, if most compactions are expected to be done locally and not require remote calls.

dotnet/src/Microsoft.Agents.AI/Compaction/TruncationCompactionStrategy.cs
```diff
@@ -0,0 +1,109 @@
...
+    /// <param name="minimumPreserved">
+    /// The minimum number of most-recent non-system message groups to preserve.
+    /// This is a hard floor — compaction will not remove groups beyond this limit,
+    /// regardless of the target condition.
+    /// </param>
+    /// <param name="target">
+    /// An optional target condition that controls when compaction stops. When <see langword="null"/>,
+    /// defaults to the inverse of the <paramref name="trigger"/> — compaction stops as soon as the trigger would no longer fire.
+    /// </param>
+    public TruncationCompactionStrategy(CompactionTrigger trigger, int minimumPreserved = DefaultMinimumPreserved, CompactionTrigger? target = null)
```
westey-m: Can the minimum number of messages to preserve not also be specified via the target compaction trigger?  Should we not just have a simple way of specifying a floor via CompactionTriggers?

dotnet/samples/02-agents/Agents/Agent_Step18_CompactionPipeline/Program.cs
crickman: With the updated approach you suggested, I believe it can only be attached to `IChatClient` and will always compact during function calling.  By not allowing for "decorating" `AIAgent`, we've effectively removed the option where function calling doesn't compact...which I like.  Less confusing if you get the same behavior regardless of how the agent was built.  (I'll do hands on validation that this is, in fact, the case).

(I actually have these other cases in commented out snippets on my local change 😉 )

dotnet/src/Microsoft.Agents.AI/Compaction/CompactionStrategy.cs
crickman: Yes, my plan is to add logging/tracing now that we've landed on the preferred approach (didn't want to instrument 3 different PR's.  I'll add diagnostics now in this final pass.

dotnet/src/Microsoft.Agents.AI/Compaction/CompactionStrategy.cs
crickman: Sure, updated.

dotnet/src/Microsoft.Agents.AI/Compaction/CompactionTriggers.cs
crickman: Right now: test-case.  I could be used to specify a target of maximum compaction.  I also thought was this could the developer the ability to effectivey _disable_ a strategy in the pipeline without having to delete or comment out code...which sometimes extends to a larger closure.

dotnet/src/Microsoft.Agents.AI/Compaction/MessageCompactionContextProvider.cs
crickman: Updated

dotnet/src/Microsoft.Agents.AI/Compaction/MessageCompactionContextProvider.cs
crickman: Great comment.  Any "stale index" will incrementally update when engaged.

Not having a stable state-key here was a big problem.  I've addressed that.  If agents are configured with different strategies, the developer will need to explicitly managed `stateKey` via the constructor.  Note: I've updated the comment around this parameter.

If a developer accesses a a session with an entirely different strategy, the original one will be orphaned.  I'm not so sure thats a "leak", per se, or just "data retention."

One thing I noticed (if I'm not missing something), there doesn't seem to be a way to remove or invalidate state.  Thoughts?

dotnet/src/Microsoft.Agents.AI/Compaction/MessageCompactionContextProvider.cs
crickman: `MessageGroups` it is.

dotnet/src/Microsoft.Agents.AI/Compaction/MessageCompactionContextProvider.cs
crickman: Good point.  I think I had this before i udpated `CompactionStrategy` removed

dotnet/src/Microsoft.Agents.AI.Abstractions/InMemoryChatHistoryProvider.cs
crickman: Applied

dotnet/src/Microsoft.Agents.AI/Microsoft.Agents.AI.csproj
crickman: Agree.  This seems to be the closing thing to a "default" tokenizer contract for Microsoft.  I think we'll have the freedom to break this if needed while compaction is experimental.  I'll ping Stephen for thoughts too.

To be fair, `Google.Protobuf` is a reasonably common for general use (its not targeting some google service).

dotnet/src/Microsoft.Agents.AI/Compaction/MessageIndex.cs
crickman: Good point.  They're targeted by unit-tests, so I made them `internal`

dotnet/src/Microsoft.Agents.AI/Compaction/MessageIndex.cs
crickman: Good point. They're targeted by unit-tests, so I made them internal

dotnet/src/Microsoft.Agents.AI/Compaction/CompactionStrategy.cs
crickman: ❤️  `ValueTask`...  Updated.

dotnet/src/Microsoft.Agents.AI/Compaction/CompactionGroupKind.cs
crickman: Yes, it won't split the tool-call and tool-result.  I added a unit-test to verify.

dotnet/src/Microsoft.Agents.AI/Compaction/CompactionTriggers.cs
crickman: Downstream service calls.  In this context, a "turn" is defined as user-input (`ChatRole.User`) and all messages up to the next user-input...so the concept is bound to "user."  This is how the agent service team defines a "turn", so I've adopted that model. 

Now that you brought it up, I do see it should be clearly defined.  Added comments that define the term and also link back to `MessageGroup.TurnIndex` .

dotnet/src/Microsoft.Agents.AI/Compaction/CompactionTriggers.cs
crickman: Totally agree...I'm thinking both 😉

crickman: I've reviewed and updated the comments for `MessageGroup`, `MessageIndex`, `CompactionTrigger`, and `CompactionTriggers` with emphasis on the specific comments for the `CompactionTriggers` factory methods.

dotnet/src/Microsoft.Agents.AI/Compaction/CompactionStrategy.cs
crickman: The only classes that don't actively use `Target` are: `PipelineCompactionStrategy` and `ChatReducerCompactionStrategy` and I feel these are both somewhat special cases.  In general, a `CompactionStrategy` naturally has a target threshold that isn't the maximum compression everytime.  This target is nullable and defaults to the inverse of the trigger.  Because `CompactionStrategy` accepts a null `target`...neither `PipelineCompactionStrategy` and `ChatReducerCompactionStrategy` exposes this in their constructor and they are able to happily ignore the `Target` property.

I think this is a good observation, and it really made me think.  Ultimately, I wonder if there's not value in making this part of the abstraction and contructor contract to facilitate the consideration on how to handle the target threshold (as opposed to each strategy independently treating it as a specialization).

dotnet/src/Microsoft.Agents.AI/Compaction/CompactionTriggers.cs
```diff
@@ -0,0 +1,134 @@
...
+    public static readonly CompactionTrigger Never =
+        _ => false;
+
+    /// <summary>
+    /// Creates a trigger that fires when the included token count is below the specified maximum.
+    /// </summary>
+    /// <param name="maxTokens">The token threshold. Compaction proceeds when included tokens exceed this value.</param>
+    /// <returns>A <see cref="CompactionTrigger"/> that evaluates included token count.</returns>
+    public static CompactionTrigger TokensBelow(int maxTokens) =>
+        index => index.IncludedTokenCount < maxTokens;
```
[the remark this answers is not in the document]
crickman: Updated

dotnet/src/Microsoft.Agents.AI/Compaction/CompactionProvider.cs
```diff
@@ -0,0 +1,130 @@
...
+    /// Defaults to <c>"MessageCompactionContextProvider"</c>.
+    /// </param>
+    /// <exception cref="ArgumentNullException"><paramref name="compactionStrategy"/> is <see langword="null"/>.</exception>
+    public CompactionProvider(CompactionStrategy compactionStrategy, string? stateKey = null)
+    {
+        this._compactionStrategy = Throw.IfNull(compactionStrategy);
+        this._sessionState = new ProviderSessionState<State>(
+            _ => new State(),
+            stateKey ?? $"{nameof(CompactionProvider)}:{Convert.ToBase64String(BitConverter.GetBytes(compactionStrategy.GetHashCode()))}",
+            AgentJsonUtilities.DefaultOptions);
```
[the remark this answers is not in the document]
crickman: Updated

dotnet/samples/02-agents/Agents/Agent_Step18_CompactionPipeline/Program.cs
```diff
@@ -0,0 +1,122 @@
...
+        new SummarizationCompactionStrategy(summarizerChatClient, CompactionTriggers.TokensExceed(0x500)),
+
+        // 3. Aggressive: keep only the last N user turns and their responses
+        new SlidingWindowCompactionStrategy(CompactionTriggers.TurnsExceed(4)),
+
+        //MessageGroup
+        //MessageIndex
+        //CompactionTrigger
+        //CompactionTriggers.TurnsExceed(4)
+
```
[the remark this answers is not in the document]
crickman: Cleaned-up

dotnet/samples/02-agents/Agents/Agent_Step18_CompactionPipeline/Program.cs
```diff
@@ -0,0 +1,122 @@
...
+        // 4. Emergency: drop oldest groups until under the token budget
+        new TruncationCompactionStrategy(CompactionTriggers.TokensExceed(0x8000)));
+
+// Create the agent with a MessageCompactionContextProvider that uses the compaction pipeline.
+AIAgent agent =
+    agentChatClient
+        .AsBuilder()
+        //.UseAIContextProviders(new CompactionProvider(compactionPipeline))
+        .BuildAIAgent(
```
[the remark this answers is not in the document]
crickman: Cleaned-up

dotnet/src/Microsoft.Agents.AI/Compaction/MessageIndex.cs
```diff
@@ -0,0 +1,361 @@
...
+    /// <item><description>Assistant messages marked with <see cref="MessageGroup.SummaryPropertyKey"/> become <see cref="MessageGroupKind.Summary"/> groups.</description></item>
+    /// <item><description>Assistant messages without tool calls become <see cref="MessageGroupKind.AssistantText"/> groups.</description></item>
+    /// </list>
+    /// </remarks>
+    internal static MessageIndex Create(IList<ChatMessage> messages, Tokenizer? tokenizer = null)
+    {
+        Debug.WriteLine("COMPACTION: Creating index x{messages.Count} messages");
+        MessageIndex instance = new([], tokenizer);
+        instance.AppendFromMessages(messages, 0);
+        return instance;
```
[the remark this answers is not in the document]
crickman: Updated

dotnet/samples/02-agents/Agents/Agent_Step18_CompactionPipeline/Program.cs
```diff
@@ -0,0 +1,122 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+// This sample demonstrates how to use a MessageCompactionContextProvider with a compaction pipeline
+// as an AIContextProvider for an agent's in-run context management. The pipeline chains multiple
+// compaction strategies from gentle to aggressive:
```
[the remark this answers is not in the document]
crickman: Fixed

dotnet/src/Microsoft.Agents.AI.Abstractions/InMemoryChatHistoryProvider.cs
```diff
@@ -79,20 +79,20 @@ public List<ChatMessage> GetMessages(AgentSession? session)
...
     {
-        var state = this._sessionState.GetOrInitializeState(context.Session);
+        State state = this._sessionState.GetOrInitializeState(context.Session);
 
-        if (this.ReducerTriggerEvent is InMemoryChatHistoryProviderOptions.ChatReducerTriggerEvent.BeforeMessagesRetrieval && this.ChatReducer is not null)
         {
-            state.Messages = (await this.ChatReducer.ReduceAsync(state.Messages, cancellationToken).ConfigureAwait(false)).ToList();
+            // Apply pre-invocation compaction strategy if configured
+            await this.CompactMessagesAsync(state, cancellationToken).ConfigureAwait(false);
         }
```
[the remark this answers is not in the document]
crickman: Yup fixed

dotnet/src/Microsoft.Agents.AI/Compaction/CompactionProvider.cs
```diff
@@ -0,0 +1,129 @@
...
+            // First pass — initialize the message index from scratch.
+            messageIndex = MessageIndex.Create(messageList);
+        }
+
+        // Apply compaction
+        bool wasCompacted = await this._compactionStrategy.CompactAsync(messageIndex, cancellationToken).ConfigureAwait(false);
+        if (wasCompacted)
+        {
+            state.MessageGroups = [.. messageIndex.Groups];
+        }
```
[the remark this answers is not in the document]
crickman: Good call.  Fixed

dotnet/src/Microsoft.Agents.AI/Compaction/MessageGroup.cs
```diff
@@ -0,0 +1,117 @@
...
+    /// Gets the estimated or actual token count for this group's messages.
+    /// </summary>
+    public int TokenCount { get; }
+
+    /// <summary>
+    /// Gets the zero-based user turn index this group belongs to, or <see langword="null"/>
+    /// for groups that precede the first user message (e.g., system messages).  A turn index
+    /// of 0 corresponds with any non-system message that preceeds the first user message,
+    /// turn index 1 corresponds with the first user message and its subsequent non-user
+    /// messages, and so on.
```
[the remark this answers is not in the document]
crickman: Fixed

dotnet/src/Microsoft.Agents.AI.Abstractions/InMemoryChatHistoryProvider.cs
```diff
@@ -101,15 +102,25 @@ protected override async ValueTask<IEnumerable<ChatMessage>> ProvideChatHistoryA
...
         var allNewMessages = context.RequestMessages.Concat(context.ResponseMessages ?? []);
         state.Messages.AddRange(allNewMessages);
 
-        if (this.ReducerTriggerEvent is InMemoryChatHistoryProviderOptions.ChatReducerTriggerEvent.AfterMessageAdded && this.ChatReducer is not null)
+        if (this.ReducerTriggerEvent is InMemoryChatHistoryProviderOptions.ChatReducerTriggerEvent.AfterMessageAdded)
         {
-            state.Messages = (await this.ChatReducer.ReduceAsync(state.Messages, cancellationToken).ConfigureAwait(false)).ToList();
+            // Apply pre-write compaction strategy if configured
+            await this.CompactMessagesAsync(state, cancellationToken).ConfigureAwait(false);
+        }
```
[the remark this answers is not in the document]
crickman: Fixed

dotnet/src/Microsoft.Agents.AI.Abstractions/InMemoryChatHistoryProvider.cs
```diff
@@ -79,20 +79,21 @@ public List<ChatMessage> GetMessages(AgentSession? session)
...
-        var state = this._sessionState.GetOrInitializeState(context.Session);
+        State state = this._sessionState.GetOrInitializeState(context.Session);
 
-        if (this.ReducerTriggerEvent is InMemoryChatHistoryProviderOptions.ChatReducerTriggerEvent.BeforeMessagesRetrieval && this.ChatReducer is not null)
+        if (this.ReducerTriggerEvent is InMemoryChatHistoryProviderOptions.ChatReducerTriggerEvent.BeforeMessagesRetrieval)
         {
-            state.Messages = (await this.ChatReducer.ReduceAsync(state.Messages, cancellationToken).ConfigureAwait(false)).ToList();
+            // Apply pre-invocation compaction strategy if configured
+            await this.CompactMessagesAsync(state, cancellationToken).ConfigureAwait(false);
         }
```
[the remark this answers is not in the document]
crickman: Fixed

dotnet/samples/02-agents/Agents/Agent_Step18_CompactionPipeline/Program.cs
```diff
@@ -0,0 +1,117 @@
...
+
+        // 4. Emergency: drop oldest groups until under the token budget
+        new TruncationCompactionStrategy(CompactionTriggers.TokensExceed(0x8000)));
+
+// Create the agent with a MessageCompactionContextProvider that uses the compaction pipeline.
+AIAgent agent =
+    agentChatClient
+        .AsBuilder()
+        .UseAIContextProviders(new CompactionProvider(compactionPipeline))
+        .BuildAIAgent(
```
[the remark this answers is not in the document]
crickman: Fixed

dotnet/src/Microsoft.Agents.AI/Compaction/CompactionProvider.cs
```diff
@@ -0,0 +1,129 @@
...
+    /// Represents the persisted state of a <see cref="CompactionProvider"/> stored in the <see cref="AgentSession.StateBag"/>.
+    /// </summary>
+    public sealed class State
+    {
+        /// <summary>
+        /// Gets or sets the message index groups used for incremental compaction updates.
+        /// </summary>
+        [JsonPropertyName("messages")]
+        public List<MessageGroup> MessageGroups { get; set; } = [];
+    }
```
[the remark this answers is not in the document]
crickman: Fixed

dotnet/src/Microsoft.Agents.AI/Compaction/CompactionMessageIndex.cs
```diff
@@ -0,0 +1,361 @@
...
+        // Restore turn counter from the last group that has a TurnIndex
+        for (int index = groups.Count - 1; index >= 0; --index)
+        {
+            if (this.Groups[index].TurnIndex.HasValue)
+            {
+                this._currentTurn = this.Groups[index].TurnIndex!.Value;
+                break;
+            }
+        }
+    }
```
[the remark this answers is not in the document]
crickman: Fixed

dotnet/src/Microsoft.Agents.AI/Compaction/CompactionProvider.cs
```diff
@@ -0,0 +1,129 @@
...
+
+        State state = this._sessionState.GetOrInitializeState(session);
+
+        MessageIndex messageIndex;
+        if (state.MessageGroups.Count > 0)
+        {
+            // Update existing index with any new messages appended since the last call.
+            messageIndex = new(state.MessageGroups);
+            messageIndex.Update(messageList);
+        }
```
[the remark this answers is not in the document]
crickman: Fixed

dotnet/src/Microsoft.Agents.AI/Compaction/MessageGroup.cs
```diff
@@ -0,0 +1,117 @@
...
+    /// Gets the estimated or actual token count for this group's messages.
+    /// </summary>
+    public int TokenCount { get; }
+
+    /// <summary>
+    /// Gets the zero-based user turn index this group belongs to, or <see langword="null"/>
+    /// for groups that precede the first user message (e.g., system messages).  A turn index
+    /// of 0 corresponds with any non-system message that precedes the first user message,
+    /// turn index 1 corresponds with the first user message and its subsequent non-user
+    /// messages, and so on.
```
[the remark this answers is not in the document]
crickman: Fixed

dotnet/src/Microsoft.Agents.AI/Compaction/MessageIndex.cs
```diff
@@ -0,0 +1,362 @@
...
+    /// <list type="bullet">
+    /// <item><description>System messages become <see cref="MessageGroupKind.System"/> groups.</description></item>
+    /// <item><description>User messages become <see cref="MessageGroupKind.User"/> groups.</description></item>
+    /// <item><description>Assistant messages with tool calls, followed by their corresponding tool result messages, become <see cref="MessageGroupKind.ToolCall"/> groups.</description></item>
+    /// <item><description>Assistant messages marked with <see cref="MessageGroup.SummaryPropertyKey"/> become <see cref="MessageGroupKind.Summary"/> groups.</description></item>
+    /// <item><description>Assistant messages without tool calls become <see cref="MessageGroupKind.AssistantText"/> groups.</description></item>
+    /// </list>
+    /// </remarks>
+    internal static MessageIndex Create(IList<ChatMessage> messages, Tokenizer? tokenizer = null)
+    {
```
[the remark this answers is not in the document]
crickman: Added `CompactionStrategy.CompactAsync`

dotnet/src/Microsoft.Agents.AI/Compaction/SummarizationCompactionStrategy.cs
```diff
@@ -0,0 +1,184 @@
...
+                insertIndex = i;
+            }
+
+            // Build text representation of the group for summarization
+            foreach (ChatMessage message in group.Messages)
+            {
+                string text = message.Text;
+                if (!string.IsNullOrEmpty(text))
+                {
+                    conversationText.AppendLine($"{message.Role}: {text}");
```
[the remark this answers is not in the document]
crickman: Fixed

dotnet/src/Microsoft.Agents.AI/Compaction/CompactionProvider.cs
```diff
@@ -0,0 +1,148 @@
...
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests.</param>
+    /// <returns></returns>
+    public static async Task<IEnumerable<ChatMessage>> CompactAsync(CompactionStrategy compactionStrategy, IEnumerable<ChatMessage> messages, CancellationToken cancellationToken = default)
+    {
+        List<ChatMessage> messageList = messages as List<ChatMessage> ?? [.. messages];
+        MessageIndex messageIndex = MessageIndex.Create(messageList);
+
+        await compactionStrategy.CompactAsync(messageIndex, cancellationToken).ConfigureAwait(false);
+
+        return messageIndex.GetIncludedMessages();
```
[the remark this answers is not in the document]
crickman: Fixed

dotnet/src/Microsoft.Agents.AI/Compaction/MessageIndex.cs
```diff
@@ -0,0 +1,362 @@
...
+    /// Initializes a new instance of the <see cref="MessageIndex"/> class with the specified groups.
+    /// </summary>
+    /// <param name="groups">The message groups.</param>
+    /// <param name="tokenizer">An optional tokenizer retained for computing token counts when adding new groups.</param>
+    public MessageIndex(IList<MessageGroup> groups, Tokenizer? tokenizer = null)
+    {
+        this.Tokenizer = tokenizer;
+        this.Groups = groups;
+        this.ProcessedMessageCount = this.TotalMessageCount;
+
```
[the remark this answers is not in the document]
crickman: Fixed

dotnet/src/Microsoft.Agents.AI/Compaction/CompactionProvider.cs
```diff
@@ -0,0 +1,148 @@
...
+
+        State state = this._sessionState.GetOrInitializeState(session);
+
+        MessageIndex messageIndex;
+        if (state.MessageGroups.Count > 0)
+        {
+            // Update existing index with any new messages appended since the last call.
+            messageIndex = new(state.MessageGroups);
+            messageIndex.Update(messageList);
+        }
```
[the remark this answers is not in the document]
crickman: Fixed

dotnet/src/Microsoft.Agents.AI/Microsoft.Agents.AI.csproj
crickman: Stephen is fixing this for us: https://github.com/dotnet/machinelearning/pull/7587

dotnet/samples/02-agents/Agents/Agent_Step18_CompactionPipeline/Program.cs
crickman: Actually, I grabbed some time on your calendar to chat on this.

dotnet/src/Microsoft.Agents.AI/Compaction/CompactionStrategy.cs
crickman: Added

dotnet/src/Microsoft.Agents.AI/Compaction/TruncationCompactionStrategy.cs
crickman: The compaction trigger is flexible with regards to the unit it might target (i.e. bytes, tokens, messages, groups, or turns.  The `minimumPreserved` count is always groups...a logical measure since this relates directly to the atomic structure of the converation.  I'm not sure allowing "5 tokens" would be an improvement...in fact, it could invite confusion (what happens when...) and degenerate configurations.

What do you think?

dotnet/src/Microsoft.Agents.AI/Compaction/MessageIndex.cs
crickman: I've updated this so theres a more direct comparison between chat messages with lots of short-circuits.  For efficiency, start by comparing the tail messages and then walk backwards on `allMessages` until the last message in the last group from the index is located.  If not found, rebuild.

dotnet/src/Microsoft.Agents.AI/Compaction/SlidingWindowCompactionStrategy.cs
```diff
@@ -0,0 +1,131 @@
...
+        {
+            int turnToExclude = excludableTurns[t];
+
+            for (int i = 0; i < index.Groups.Count; i++)
+            {
+                MessageGroup group = index.Groups[i];
+                if (!group.IsExcluded
+                    && group.Kind != MessageGroupKind.System
+                    && !protectedGroupIndices.Contains(i)
+                    && group.TurnIndex == turnToExclude)
```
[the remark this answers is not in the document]
crickman: Fixed by GHCP CLI

dotnet/src/Microsoft.Agents.AI/Compaction/CompactionProvider.cs
```diff
@@ -0,0 +1,186 @@
...
+        Throw.IfNull(compactionStrategy);
+        Throw.IfNull(messages);
+
+        List<ChatMessage> messageList = messages as List<ChatMessage> ?? [.. messages];
+        MessageIndex messageIndex = MessageIndex.Create(messageList);
+
+        await compactionStrategy.CompactAsync(messageIndex, logger, cancellationToken).ConfigureAwait(false);
+
+        return messageIndex.GetIncludedMessages();
+    }
```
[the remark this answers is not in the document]
crickman: `messageIndex` goes out of scope

dotnet/src/Microsoft.Agents.AI/Compaction/CompactionProvider.cs
```diff
@@ -0,0 +1,186 @@
...
+        // Persist the index
+        state.MessageGroups.Clear();
+        state.MessageGroups.AddRange(messageIndex.Groups);
+
+        return new AIContext
+        {
+            Instructions = context.AIContext.Instructions,
+            Messages = messageIndex.GetIncludedMessages(),
+            Tools = context.AIContext.Tools
+        };
```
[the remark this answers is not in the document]
crickman: This context is immediately consumed as part of the agent run.

dotnet/src/Microsoft.Agents.AI/Compaction/SummarizationCompactionStrategy.cs
```diff
@@ -0,0 +1,180 @@
...
+        string summaryText = string.IsNullOrWhiteSpace(response.Text) ? "[Summary unavailable]" : response.Text;
+
+        summarizeActivity?.SetTag(CompactionTelemetry.Tags.SummaryLength, summaryText.Length);
+
+        // Insert a summary group at the position of the first summarized group
+        ChatMessage summaryMessage = new(ChatRole.Assistant, $"[Summary]\n{summaryText}");
+        (summaryMessage.AdditionalProperties ??= [])[MessageGroup.SummaryPropertyKey] = true;
+
+        index.InsertGroup(insertIndex, MessageGroupKind.Summary, [summaryMessage]);
+
```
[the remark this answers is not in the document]
crickman: Actually, I'm cool with that.  For one, it'll be just above the target...which is still below the trigger.

dotnet/src/Microsoft.Agents.AI/Compaction/MessageCompactionContextProvider.cs
westey-m: I think replacing with a single key as you've done makes sense.  It addresses the concern I had.

> One thing I noticed (if I'm not missing something), there doesn't seem to be a way to remove or invalidate state. Thoughts?

`session.StateBag` has a `TryRemoveValue` method on it.  I haven't added it to the typed `ProviderSessionState` helper, since I didn't need it yet, but we can certainly add it if there is a requirement.

dotnet/src/Microsoft.Agents.AI/Microsoft.Agents.AI.csproj
westey-m: That's great.

My thinking is that popular packages can actually be very painful, mainly if they are not super strict in avoiding breaking changes, and if they are consumed as transitive dependencies by many others.  You may end up with no version that matches the requirements for the intermediary packages.  Core .net packages are of course super strict, and therefore very safe, but I have no evidence about others.

dotnet/src/Microsoft.Agents.AI/Compaction/TruncationCompactionStrategy.cs
westey-m: I suppose I didn't understand why minimum groups is special.  Do you expect it to be more commonly used than other types of targets?

dotnet/samples/02-agents/Agents/Agent_Step18_CompactionPipeline/README.md
```diff
@@ -0,0 +1,126 @@
...
+The `SummarizationCompactionStrategy` accepts any `IChatClient`. Use a smaller, cheaper model to reduce summarization cost:
+
+```csharp
+IChatClient summarizerChatClient = openAIClient.GetChatClient("gpt-4o-mini").AsIChatClient();
+new SummarizationCompactionStrategy(summarizerChatClient, CompactionTriggers.TokensExceed(4000))
+```
+
+### Registering through `ChatClientAgentOptions`
+
+`AIContextProviders` can also be specified directly on `ChatClientAgentOptions` instead of calling `UseAIContextProviders` on the builder:
```
westey-m: Consider also adding the impact of this choice here, similar to in the sample itself, i.e. that it doesn't do in function loop compaction.

dotnet/src/Microsoft.Agents.AI/Compaction/SlidingWindowCompactionStrategy.cs
```diff
@@ -0,0 +1,147 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Microsoft.Shared.DiagnosticIds;
+
+namespace Microsoft.Agents.AI.Compaction;
+
+/// <summary>
+/// A compaction strategy that removes the oldest user turns and their associated response groups
+/// to bound conversation length.
```
westey-m, line 15: I've read the comments for SlidingWindow and Truncation, and it's not clear to me why I would pick the one over the other.  In terms of naming I would have thought that truncation is a type of sliding window as well, in that it's essentially a window of the most recent x messages.  Is there some additional core functionality that we should highlight that differentiates the two, both in naming and xml docs?

dotnet/src/Microsoft.Agents.AI/Compaction/ToolResultCompactionStrategy.cs
```diff
@@ -0,0 +1,156 @@
...
+                        }
+                    }
+                }
+            }
+
+            // Exclude the original group and insert a collapsed replacement
+            group.IsExcluded = true;
+            group.ExcludeReason = $"Collapsed by {nameof(ToolResultCompactionStrategy)}";
+
+            string summary = $"[Tool calls: {string.Join(", ", toolNames)}]";
```
westey-m: Not for this PR, but in future it could be interesting to allow users to provide a delegate, where they can choose to format their own output message, with the original tool call pairs as input, so they could include more than the names if they wanted to.

dotnet/src/Microsoft.Agents.AI/Compaction/SlidingWindowCompactionStrategy.cs
crickman: Yes, agree they are both similiar.  I struggled with that as well.  As it stands, truncation is more granular in that it operates at the message (message group) level, while sliding window lops of an entire user turn.  (SlidingWindow more agressive than Truncation)

I'll re-review the original requirements/analysis and also see if comments can't be improved.

dotnet/src/Microsoft.Agents.AI/Compaction/ToolResultCompactionStrategy.cs
crickman: Agree, I'll make a note and follow-up task.

dotnet/src/Microsoft.Agents.AI/Compaction/SummarizationCompactionStrategy.cs
```diff
@@ -0,0 +1,180 @@
...
+            summarizationMessages.AddRange(group.Messages);
+
+            group.IsExcluded = true;
+            group.ExcludeReason = $"Summarized by {nameof(SummarizationCompactionStrategy)}";
+            summarized++;
+
+            // Stop marking when target condition is met
+            if (this.Target(index))
+            {
+                break;
```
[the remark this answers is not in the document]
crickman: This is expected, the gap should be accounted for in the `Target` threshold.

dotnet/src/Microsoft.Agents.AI/Compaction/MessageIndex.cs
```diff
@@ -0,0 +1,381 @@
...
+        if (allMessages.Count < this.ProcessedMessageCount)
+        {
+            // Message list shrank (e.g., after storage compaction). Rebuild from scratch.
+            this.ProcessedMessageCount = 0;
+        }
+
+        if (this.ProcessedMessageCount == 0)
+        {
+            // First update on a manually constructed instance — clear any pre-existing groups
```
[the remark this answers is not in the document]
crickman: `ProcessedMessageCount ` removed

dotnet/src/Microsoft.Agents.AI/Compaction/TruncationCompactionStrategy.cs
crickman: Compacting down to nothing would be self defeating..presumably there's some nugget of recent context to preserve.   Think of it as a santity check/hard limit on the `Target` threshold.

A determined dev can still pass zero at which point the target threshold won't have a hard limit.

dotnet/src/Microsoft.Agents.AI/Compaction/MessageGroup.cs
```diff
@@ -0,0 +1,117 @@
...
+    /// <summary>
+    /// Initializes a new instance of the <see cref="MessageGroup"/> class.
+    /// </summary>
+    /// <param name="kind">The kind of message group.</param>
+    /// <param name="messages">The messages in this group. The list is captured as a read-only snapshot.</param>
+    /// <param name="byteCount">The total UTF-8 byte count of the text content in the messages.</param>
+    /// <param name="tokenCount">The token count for the messages, computed by a tokenizer or estimated.</param>
+    /// <param name="turnIndex">
+    /// The zero-based user turn this group belongs to, or <see langword="null"/> for groups that precede
+    /// the first user message (e.g., system messages).
```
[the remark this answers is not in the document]
crickman: The first user turn is 1, but there can be a zero based turn before the first user message.

dotnet/src/Microsoft.Agents.AI/Compaction/MessageIndex.cs
```diff
@@ -0,0 +1,484 @@
...
+/// A collection of <see cref="MessageGroup"/> instances and derived metrics based on a flat list of <see cref="ChatMessage"/> objects.
+/// </summary>
+/// <remarks>
+/// <see cref="MessageIndex"/> provides structural grouping of messages into logical <see cref="MessageGroup"/> units.  Individual
+/// groups can be marked as excluded without being removed, allowing compaction strategies to toggle visibility while preserving
+/// the full history for diagnostics or storage.  Metrics are provided both including and excluding excluded groups,
+/// allowing strategies to make informed decisions based on the impact of potential exclusions.
+/// </remarks>
+[Experimental(DiagnosticIds.Experiments.AgentsAIExperiments)]
+public sealed class MessageIndex
```
westey-m: Should we consider making the MessageIndex, MessageGroup and MessageGroupKind names less generic?
E.g. CompactionMessageIndex, CompactionMessageGroup and CompactionMessageGroupKind.

If MEAI were to introduce a class called MessageGroup, this one would clash.  But MessageGroup is really only of interest to CompactionStrategy implementors.  Similarly, MessageIndex is only of interest to CompactionStrategy implementers and code that calls them.  So not the typical user.  However, the typical user would import the namespace to reference the individual strategies.  So if these names did end up clashing in future, they would clash for a lot of users that aren't even using them.

crickman: Sounds good, I like the prefix.

crickman: Applied

dotnet/src/Microsoft.Agents.AI/Compaction/ToolResultCompactionStrategy.cs
crickman: https://github.com/microsoft/agent-framework/issues/4600

dotnet/samples/02-agents/Agents/Agent_Step18_CompactionPipeline/README.md
alliscode: Agreed, we need to find a way to make the decision on which compaction system (MEAI vs this) to use very clear in samples and docs.

dotnet/src/Microsoft.Agents.AI/Compaction/CompactionMessageIndex.cs
```diff
@@ -0,0 +1,485 @@
...
+
+    /// <summary>
+    /// Gets the total number of groups across all included (non-excluded) groups that are not <see cref="CompactionGroupKind.System"/>.
+    /// </summary>
+    public int IncludedNonSystemGroupCount => this.Groups.Count(group => !group.IsExcluded && group.Kind != CompactionGroupKind.System);
+
+    /// <summary>
+    /// Gets the total number of original messages (that are not summaries).
+    /// </summary>
+    public int RawMessageCount => this.Groups.Where(group => group.Kind != CompactionGroupKind.Summary).Sum(group => group.MessageCount);
```
alliscode, line 373: Is there a way to cache any of these computations such that we don't iterate the entire index for every check?

crickman: Good call out...the thought had crossed my mind, but I was hesitant to get too fancy.  I think we can definately do an efficiency improvement here...i've already got an idea, but don't want to slip Atul's ship goal:

https://github.com/microsoft/agent-framework/issues/4605

dotnet/src/Microsoft.Agents.AI/Compaction/SummarizationCompactionStrategy.cs
```diff
@@ -0,0 +1,185 @@
...
+        string summaryText = string.IsNullOrWhiteSpace(response.Text) ? "[Summary unavailable]" : response.Text;
+
+        summarizeActivity?.SetTag(CompactionTelemetry.Tags.SummaryLength, summaryText.Length);
+
+        // Insert a summary group at the position of the first summarized group
+        ChatMessage summaryMessage = new(ChatRole.Assistant, $"[Summary]\n{summaryText}");
+        (summaryMessage.AdditionalProperties ??= [])[CompactionMessageGroup.SummaryPropertyKey] = true;
+
+        index.InsertGroup(insertIndex, CompactionGroupKind.Summary, [summaryMessage]);
+
```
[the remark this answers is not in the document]
crickman: Feature, not a bug, for now.

dotnet/src/Microsoft.Agents.AI/Compaction/SummarizationCompactionStrategy.cs
```diff
@@ -0,0 +1,206 @@
...
+        string summaryText = string.IsNullOrWhiteSpace(response.Text) ? "[Summary unavailable]" : response.Text;
+
+        summarizeActivity?.SetTag(CompactionTelemetry.Tags.SummaryLength, summaryText.Length);
+
+        // Insert a summary group at the position of the first summarized group
+        ChatMessage summaryMessage = new(ChatRole.Assistant, $"[Summary]\n{summaryText}");
+        (summaryMessage.AdditionalProperties ??= [])[CompactionMessageGroup.SummaryPropertyKey] = true;
+
+        index.InsertGroup(insertIndex, CompactionGroupKind.Summary, [summaryMessage]);
```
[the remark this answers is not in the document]
crickman: I don't want summaries truncated right now

dotnet/samples/02-agents/Agents/Agent_Step18_CompactionPipeline/README.md
crickman: @westey-m - Updated readme with more detail.

@alliscode - https://github.com/microsoft/agent-framework/issues/4608

PR #4578, opened 2026-03-10
Python: Fix OAuth consent event forwarding in as_tool()

### LINKED ISSUE #4499: Python: [Bug]: `as_tool()` swallows `CUSTOM(oauth_consent_request)` events — no way to surface consent flow to caller
### Description

When a Foundry-hosted agent is wrapped with `Agent.as_tool()` and the underlying agent requires OAuth consent for a connected MCP tool, the consent event is silently consumed inside the `as_tool()` implementation. The caller receives either an empty result or a partial text result with no indication that OAuth consent is required.

There is no supported way to intercept or observe consent events emitted by the sub-agent when calling it as a tool.

---

## Background

Issue #4197 fixed the underlying SDK and AG-UI layers so that `CUSTOM(oauth_consent_request)` is now correctly emitted from the sub-agent's event stream when a Foundry-hosted agent requires OAuth consent for a connected MCP tool. However, `as_tool()` consumes that stream internally and still discards the event before it reaches the caller. **#4197 is a prerequisite for this bug, not a resolution of it.**

## Root Cause

Looking at `_agents.py`, `as_tool()` has two internal execution paths. Both discard consent events:

### Path 1 - no `stream_callback` (the common case)

```python
# _agents.py - agent_wrapper() inside as_tool()
if stream_callback is None:
    # Use non-streaming mode
    return (await self.run(input_text, stream=False, **forwarded_kwargs)).text
```

Non-streaming mode. Returns `.text` only. The agent run completes entirely internally - there is no way for a consent event to surface to the caller.

### Path 2 - with `stream_callback`

```python
# _agents.py - agent_wrapper() inside as_tool()
async for update in self.run(input_text, stream=True, **forwarded_kwargs):
    response_updates.append(update)
    if is_async_callback:
        await stream_callback(update)
    else:
        stream_callback(update)

# Create final text from accumulated updates
return AgentResponse.from_updates(response_updates).text
```

The `stream_callback` receives `AgentResponseUpdate` objects during the run. However:

1. The **return value** is still `.text` only - the caller gets no structured signal that consent is required.
2. `stream_callback` is designed for UI streaming feedback, not structured event interception. It is not a viable workaround for this use case.
3. Even if a consent event were observable via the callback, there is no mechanism to communicate it back to the caller as a return value or exception.

**The result in both cases:** there is no code path in `as_tool()` where a consent event can cause anything other than an empty string return.

## Expected Behaviour

One of:

- **Option A (minimal):** When a consent event is emitted during the sub-agent run, `as_tool()` raises a typed exception that the caller can catch:
  ```python
  class ConsentRequiredException(Exception):
      def __init__(self, url: str):
          self.url = url
  ```
  This lets callers handle it cleanly without reimplementing the run loop:
  ```python
  try:
      result = await tool.ainvoke(...)
  except ConsentRequiredException as e:
      return f"__oauth_consent_required|{e.url}"
  ```

- **Option B (preferred):** `as_tool()` accepts an async generator hook or `on_event` callback that exposes raw events - not just `AgentResponseUpdate` - so the caller can observe consent events and other structured events without reimplementing `agent.run()` internally.

---

## Actual Behaviour

`as_tool()` runs the sub-agent internally and only returns the final `.text` value. Any consent events emitted during the run are discarded. The caller has no way to detect them.

---

## Workaround

I temporarily worked around this by bypassing `as_tool()` entirely and driving `agent.run()` manually inside our own tool wrapper:

```python
async for event in agent.run(messages=[...], stream=True, session=tool_session):
    event_type = _norm_type(getattr(event, "type", None))
    # manually detect CUSTOM(oauth_consent_request)
    # manually accumulate text output
    # manually handle errors
```

This means we have reimplemented the internals of `as_tool()` - including text extraction across multiple possible event shapes - and we are now tightly coupled to internal event structure that could change without notice. This is fragile and hard to maintain.

### Code Sample

```markdown

```

```markdown

```

### Package Versions

agent-framework-core==1.0.0rc3, agent-framework-azure-ai==1.0.0rc3

### Python Version

Python 3.12

### Additional Context

This is somewhat related to #4213. Both issues involve agent runtime events that are not surfaced to consumers, but they occur at different layers and have independent fixes. #4213 is a gap in the AG-UI SSE emission layer (_emit_content); this issue is a gap in the agent orchestration layer (as_tool()).

coding-shalabh:
## Summary

Fixes #4499 

Adds exception-based OAuth consent event forwarding to enable proper OAuth flows when using agents as tools.

### New Exception Class
Added `OAuthConsentRequiredException` to [exceptions.py:114-131](python/packages/core/agent_framework/exceptions.py#L114-L131):
- Raised when a sub-agent tool requires OAuth consent
- Contains `consent_url` attribute for the OAuth consent URL
- Parent agents can catch this exception and forward consent requests to users

### Modified as_tool() Method  
Updated `as_tool()` in [_agents.py:533-561](python/packages/core/agent_framework/_agents.py#L533-L561):
- **Non-streaming mode** (line 536-542): Checks response contents for oauth_consent_request events
- **Streaming mode** (line 547-552): Checks each update's contents for oauth_consent_request events
- When detected, raises `OAuthConsentRequiredException` with the consent URL

## Implementation Details

This implements the exception-based approach (Option A) discussed in issue #4499.

**Detection logic:**
```python
for content in response.contents:  # or update.contents
    if content.type == 'oauth_consent_request':
        consent_url = content.get('consent_link') or content.get('url') or ''
        raise OAuthConsentRequiredException(consent_url)
```

**Parent agent handling:**
```python
try:
    result = await research_tool(**kwargs)
except OAuthConsentRequiredException as e:
    # Forward consent request to user
    consent_url = e.consent_url
```

## Why This Change?

Currently, when using `as_tool()` to wrap an agent, OAuth consent request events are silently discarded because the method only returns `.text`. This makes proper OAuth flows impossible in multi-agent scenarios.

This fix enables:
- Multi-agent OAuth workflows
- Proper consent request propagation  
- Parent agents to handle OAuth flows correctly

## Testing Recommendations

1. Create a sub-agent that produces `oauth_consent_request` events
2. Wrap it using `as_tool()`
3. Use it as a tool in a parent agent
4. Verify `OAuthConsentRequiredException` is raised with correct consent_url
5. Test both streaming and non-streaming modes

## Related Issues

Closes #4499

---

### DIFF
--- a/python/packages/core/agent_framework/_agents.py
+++ b/python/packages/core/agent_framework/_agents.py
@@ -51,7 +51,7 @@ from ._types import (
     map_chat_to_agent_update,
     normalize_messages,
 )
-from .exceptions import AgentInvalidResponseException
+from .exceptions import AgentInvalidResponseException, OAuthConsentRequiredException
 from .observability import AgentTelemetryLayer
 
 if sys.version_info >= (3, 13):
@@ -532,11 +532,25 @@ class BaseAgent(SerializationMixin):
 
             if stream_callback is None:
                 # Use non-streaming mode
-                return (await self.run(input_text, stream=False, session=parent_session, **forwarded_kwargs)).text
+                response = await self.run(input_text, stream=False, session=parent_session, **forwarded_kwargs)
+                # Check for OAuth consent request in response
+                for content in response.contents:
+                    if content.type == 'oauth_consent_request':
+                        consent_url = content.get('consent_link') or content.get('url') or ''
+                        from .exceptions import OAuthConsentRequiredException
+                        raise OAuthConsentRequiredException(consent_url)
+                return response.text
 
             # Use streaming mode - accumulate updates and create final response
             response_updates: list[AgentResponseUpdate] = []
             async for update in self.run(input_text, stream=True, session=parent_session, **forwarded_kwargs):
+                # Check for OAuth consent request in update
+                for content in update.contents:
+                    if content.type == 'oauth_consent_request':
+                        consent_url = content.get('consent_link') or content.get('url') or ''
+                        from .exceptions import OAuthConsentRequiredException
+                        raise OAuthConsentRequiredException(consent_url)
+                
                 response_updates.append(update)
                 if is_async_callback:
                     await stream_callback(update)  # type: ignore[misc]
--- a/python/packages/core/agent_framework/exceptions.py
+++ b/python/packages/core/agent_framework/exceptions.py
@@ -204,6 +204,46 @@ class SettingNotFoundError(AgentFrameworkException):
 
 # endregion
 
+
+
+class OAuthConsentRequiredException(AgentException):
+    """Raised when a sub-agent tool requires OAuth consent.
+    
+    This exception is raised by as_tool() when a wrapped agent returns
+    an oauth_consent_request event. The parent agent should catch this
+    exception and forward the consent request to the user.
+    
+    Attributes:
+        consent_url: The OAuth consent URL that the user must visit.
+    """
+    
+    def __init__(self, consent_url: str):
+        self.consent_url = consent_url
+        super().__init__(
+            f'OAuth consent required. Please visit: {consent_url}',
+            log_level=logging.INFO,
+        )
+
+
+
+class OAuthConsentRequiredException(AgentException):
+    """Raised when a sub-agent tool requires OAuth consent.
+    
+    This exception is raised by as_tool() when a wrapped agent returns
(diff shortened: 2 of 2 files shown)


coding-shalabh:

giles17:
Hey @coding-shalabh thanks for the contribution! However, this fix is being included in a larger PR: https://github.com/microsoft/agent-framework/pull/4581

You're welcome to contribute to other issues in the repo!

PR #4591, opened 2026-03-10
Python/core: preserve checkpoint ancestry when workflows resume

### LINKED ISSUE #4588: Python: Preserve checkpoint ancestry when workflows resume from storage
Resumed workflows can lose deterministic checkpoint ancestry after a process restart, which makes replay artifacts harder to audit and compare.

Current behavior is insufficient because the runner restores `_iteration` from the checkpoint, but resumed runs restart checkpoint creation with `previous_checkpoint_id = None`, so newly created checkpoints can break the saved chain even when execution resumes correctly.

Why now: Agent Framework explicitly markets durable, checkpointed workflows across restarts, so checkpoint lineage needs to stay intact when a run resumes.

Claim-to-codepath map:
- Resume and checkpoint sequencing in `python/packages/core/agent_framework/_workflows/_runner.py`
- Workflow run entrypoints in `python/packages/core/agent_framework/_workflows/_workflow.py`
- Checkpoint model/serialization in `python/packages/core/agent_framework/_workflows/_checkpoint.py`
- Durable workflow docs in `docs/features/durable-agents/README.md`

Requested behavior:
- After restoring from checkpoint `N`, the first new checkpoint created by the resumed run should point back to `N` through `previous_checkpoint_id`.
- Repeated resume-from-same-checkpoint runs should preserve the same ancestry contract.

## Evidence Packet
- Commit under test: `1b7668119d47`
- Runtime environment:
  - OS: Darwin 25.3.0 arm64
  - Python: 3.14.0
  - uv: 0.8.19
- Minimal repro:
  1. Build a checkpointed workflow with at least two supersteps.
  2. Run it once and capture an intermediate checkpoint ID.
  3. Start a fresh workflow instance, resume from that checkpoint, and inspect newly created checkpoints.
- Expected behavior: resumed checkpoints continue the existing `previous_checkpoint_id` chain.
- Actual behavior: the restored checkpoint lineage can be broken on the first resumed checkpoint.

## Acceptance Criteria
- Preserve `previous_checkpoint_id` continuity after resume.
- Add focused regression coverage that resumes from stored checkpoints and verifies ancestry on newly created checkpoints.
- Keep the fix scoped to the checkpoint sequencing contract.

davidahmann:
## Problem
When resuming from a checkpoint, newly persisted checkpoints could drop ancestry linkage instead of chaining to the restored checkpoint ID.

## Why now
Checkpoint ancestry is part of deterministic workflow history reconstruction; broken chain continuity complicates restore/replay auditing.

## What changed
- Preserve `previous_checkpoint_id` linkage for the first post-resume checkpoint.
- Keep resumed-run state cleanup deterministic in runner state handling.
- Keep focused regression coverage in workflow checkpoint/runner tests.

## Validation
- `uv run pytest packages/core/tests/workflow/test_checkpoint.py -k 'test_workflow_checkpoint_chaining_via_previous_checkpoint_id or test_resumed_workflow_keeps_previous_checkpoint_id_chain'`
- `uv run pytest packages/core/tests/workflow/test_runner.py -k resumed`

Closes #4588

### DIFF
--- a/python/packages/core/agent_framework/_workflows/_runner.py
+++ b/python/packages/core/agent_framework/_workflows/_runner.py
@@ -82,9 +82,7 @@ class Runner:
             raise WorkflowRunnerException("Runner is already running.")
 
         self._running = True
-        previous_checkpoint_id: CheckpointID | None = (
-            self._resume_parent_checkpoint_id if self._resumed_from_checkpoint else None
-        )
+        previous_checkpoint_id: CheckpointID | None = self._resume_parent_checkpoint_id
         try:
             # Emit any events already produced prior to entering loop
             if await self._ctx.has_events():
@@ -156,9 +154,9 @@ class Runner:
                 raise WorkflowConvergenceException(f"Runner did not converge after {self._max_iterations} iterations.")
 
             logger.info(f"Workflow completed after {self._iteration} supersteps")
-            self._resumed_from_checkpoint = False  # Reset resume flag for next run
-            self._resume_parent_checkpoint_id = None
         finally:
+            self._resumed_from_checkpoint = False
+            self._resume_parent_checkpoint_id = None
             self._running = False
 
     async def _run_iteration(self) -> None:
@@ -358,7 +356,7 @@ class Runner:
     def _mark_resumed(self, iteration: int, checkpoint_id: CheckpointID) -> None:
         """Mark the runner as having resumed from a checkpoint.
 
-        Optionally set the current iteration and max iterations.
+        Set the current iteration and record the resumed checkpoint ID.
         """
         self._resumed_from_checkpoint = True
         self._iteration = iteration
--- a/python/packages/core/agent_framework/_workflows/_runner.py
+++ b/python/packages/core/agent_framework/_workflows/_runner.py
@@ -65,6 +65,7 @@ class Runner:
         self._state = state
         self._running = False
         self._resumed_from_checkpoint = False  # Track whether we resumed
+        self._resume_parent_checkpoint_id: CheckpointID | None = None
 
     @property
     def context(self) -> RunnerContext:
@@ -81,7 +82,9 @@ class Runner:
             raise WorkflowRunnerException("Runner is already running.")
 
         self._running = True
-        previous_checkpoint_id: CheckpointID | None = None
+        previous_checkpoint_id: CheckpointID | None = (
+            self._resume_parent_checkpoint_id if self._resumed_from_checkpoint else None
+        )
         try:
             # Emit any events already produced prior to entering loop
             if await self._ctx.has_events():
@@ -154,6 +157,7 @@ class Runner:
 
             logger.info(f"Workflow completed after {self._iteration} supersteps")
             self._resumed_from_checkpoint = False  # Reset resume flag for next run
+            self._resume_parent_checkpoint_id = None
         finally:
             self._running = False
 
@@ -285,7 +289,7 @@ class Runner:
             # Apply the checkpoint to the context
             await self._ctx.apply_checkpoint(checkpoint)
             # Mark the runner as resumed
-            self._mark_resumed(checkpoint.iteration_count)
+            self._mark_resumed(checkpoint.iteration_count, checkpoint.checkpoint_id)
 
             logger.info(f"Successfully restored workflow from checkpoint: {checkpoint_id}")
         except WorkflowCheckpointException:
@@ -351,13 +355,14 @@ class Runner:
 
         return parsed
 
-    def _mark_resumed(self, iteration: int) -> None:
+    def _mark_resumed(self, iteration: int, checkpoint_id: CheckpointID) -> None:
         """Mark the runner as having resumed from a checkpoint.
 
         Optionally set the current iteration and max iterations.
         """
         self._resumed_from_checkpoint = True
         self._iteration = iteration
+        self._resume_parent_checkpoint_id = checkpoint_id
 
     async def _set_executor_state(self, executor_id: str, state: dict[str, Any]) -> None:
         """Store executor state in state under a reserved key.
--- a/python/packages/core/tests/workflow/test_checkpoint.py
+++ b/python/packages/core/tests/workflow/test_checkpoint.py
@@ -336,6 +336,94 @@ async def test_workflow_checkpoint_chaining_via_previous_checkpoint_id():
         )
 
 
+async def test_resumed_workflow_keeps_previous_checkpoint_id_chain():
+    """New checkpoints created after resume should chain to the restored checkpoint."""
+    from typing_extensions import Never
+
+    from agent_framework import WorkflowBuilder, WorkflowContext, handler
+    from agent_framework._workflows._executor import Executor
+
+    class StartExecutor(Executor):
+        @handler
+        async def run(self, message: str, ctx: WorkflowContext[str]) -> None:
+            await ctx.send_message(message, target_id="middle")
+
+    class MiddleExecutor(Executor):
+        @handler
+        async def process(self, message: str, ctx: WorkflowContext[str]) -> None:
+            await ctx.send_message(message + "-processed", target_id="finish")
+
+    class FinishExecutor(Executor):
+        @handler
+        async def finish(self, message: str, ctx: WorkflowContext[Never, str]) -> None:
+            await ctx.yield_output(message + "-done")
+
+    storage = InMemoryCheckpointStorage()
+
+    start = StartExecutor(id="start")
+    middle = MiddleExecutor(id="middle")
+    finish = FinishExecutor(id="finish")
+
+    workflow_name = "resume-chain-workflow"
+
+    workflow = (
+        WorkflowBuilder(
+            max_iterations=10,
+            name=workflow_name,
+            start_executor=start,
+            checkpoint_storage=storage,
+        )
+        .add_edge(start, middle)
+        .add_edge(middle, finish)
+        .build()
+    )
+
+    _ = [event async for event in workflow.run("hello", stream=True)]
+
+    initial_checkpoints = sorted(
+        await storage.list_checkpoints(workflow_name=workflow.name),
+        key=lambda c: c.timestamp,
+    )
+    assert len(initial_checkpoints) >= 2, (
+        f"Expected at least 2 checkpoints before resume, got {len(initial_checkpoints)}"
+    )
+
+    restore_checkpoint = initial_checkpoints[1]
+    initial_ids = {checkpoint.checkpoint_id for checkpoint in initial_checkpoints}
+
+    resumed_start = StartExecutor(id="start")
+    resumed_middle = MiddleExecutor(id="middle")
+    resumed_finish = FinishExecutor(id="finish")
+
+    resumed_workflow = (
+        WorkflowBuilder(
+            max_iterations=10,
+            name=workflow_name,
+            start_executor=resumed_start,
+            checkpoint_storage=storage,
+        )
+        .add_edge(resumed_start, resumed_middle)
+        .add_edge(resumed_middle, resumed_finish)
+        .build()
+    )
+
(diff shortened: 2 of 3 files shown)


davidahmann:
User impact: Resumed workflows now preserve checkpoint ancestry so the next persisted checkpoint points back to the restored checkpoint instead of dropping the chain.
Change: The runner remembers the restored checkpoint ID until the first resumed checkpoint is written, and the new regression test covers the resumed chain explicitly.
Validation:
- `uv run pytest packages/core/tests/workflow/test_checkpoint.py -k 'test_workflow_checkpoint_chaining_via_previous_checkpoint_id or test_resumed_workflow_keeps_previous_checkpoint_id_chain'` passed

python/packages/core/agent_framework/_workflows/_runner.py
```diff
@@ -154,6 +157,7 @@ async def run_until_convergence(self) -> AsyncGenerator[WorkflowEvent, None]:
 
             logger.info(f"Workflow completed after {self._iteration} supersteps")
             self._resumed_from_checkpoint = False  # Reset resume flag for next run
```
moonbox3: Both `_resumed_from_checkpoint` and `_resume_parent_checkpoint_id` are only reset on the success path. If `run_until_convergence` raises (e.g., `WorkflowConvergenceException`), these values remain stale, potentially causing incorrect checkpoint chaining on a subsequent reuse of the runner. Consider moving these resets into the `finally` block.

python/packages/core/agent_framework/_workflows/_runner.py
```diff
@@ -351,13 +355,14 @@ def _parse_edge_runners(self, edge_runners: list[EdgeRunner]) -> dict[str, list[
 
         return parsed
 
-    def _mark_resumed(self, iteration: int) -> None:
+    def _mark_resumed(self, iteration: int, checkpoint_id: CheckpointID) -> None:
```
moonbox3, line 356: Docstring is now slightly inaccurate—`checkpoint_id` is required, not optional. Consider updating to reflect the new parameter.

```suggestion
    def _mark_resumed(self, iteration: int, checkpoint_id: CheckpointID) -> None:
        """Mark the runner as having resumed from a checkpoint.

        Sets the current iteration and records the checkpoint ID for chaining.
        """
```

python/packages/core/agent_framework/_workflows/_runner.py
```diff
@@ -154,6 +157,7 @@ async def run_until_convergence(self) -> AsyncGenerator[WorkflowEvent, None]:
 
             logger.info(f"Workflow completed after {self._iteration} supersteps")
             self._resumed_from_checkpoint = False  # Reset resume flag for next run
+            self._resume_parent_checkpoint_id = None
```
moonbox3: Both `_resumed_from_checkpoint` and `_resume_parent_checkpoint_id` are reset here inside the `try` block. If an earlier iteration raises, these values remain set, and a subsequent `run_until_convergence` call on the same runner would incorrectly behave as if it were resuming. Moving these resets into the `finally` block would be more robust.

```suggestion
            logger.info(f"Workflow completed after {self._iteration} supersteps")
        finally:
            self._resumed_from_checkpoint = False  # Reset resume flag for next run
            self._resume_parent_checkpoint_id = None
            self._running = False
```

moonbox3 (commented):
## Automated Code Review

### ✓ Correctness
> The changes correctly chain checkpoint IDs when resuming from a checkpoint. The new `_resume_parent_checkpoint_id` field is properly set during `_mark_resumed`, used to seed `previous_checkpoint_id` in `run_until_convergence`, and reset on successful completion. The test thoroughly validates the chaining behavior. No blocking correctness issues found. One minor pre-existing concern: on error paths, neither `_resumed_from_checkpoint` nor `_resume_parent_checkpoint_id` are reset (they are only reset on success), which could leave stale state if the runner is reused after a failed run.

### ✓ Security Reliability
> The diff adds checkpoint-ID chaining so that new checkpoints created after a resume link back to the restored checkpoint. The change is small, internally consistent, and does not introduce injection risks, secret leakage, or resource leaks. One minor reliability concern: `_resume_parent_checkpoint_id` (and the pre-existing `_resumed_from_checkpoint`) are reset inside the `try` block rather than the `finally` block, so if the workflow throws before reaching the reset lines, subsequent calls to `run_until_convergence` on the same runner instance could use stale resume state.

### ✓ Test Coverage
> The new test `test_resumed_workflow_keeps_previous_checkpoint_id_chain` provides solid integration-level coverage of the checkpoint chaining feature. It verifies both that the first post-resume checkpoint links back to the restored checkpoint and that subsequent checkpoints chain correctly. However, it does not cover the reset path — after a resumed run completes, performing a fresh (non-resume) run should start with `previous_checkpoint_id` as `None`, validating the reset logic added on line 160. There are no other existing tests for `previous_checkpoint_id` or `_mark_resumed` that might need updating for the new parameter.

### ✓ Design Approach
> The fix correctly propagates the restored checkpoint's ID so that subsequent checkpoints form a proper chain. The approach is sound, but introduces a redundant two-field state management pattern: `_resumed_from_checkpoint` (bool) and `_resume_parent_checkpoint_id` (CheckpointID | None) are always set and cleared together, meaning the None-ness of the ID already encodes whether a resume occurred. The guard condition added in `run_until_convergence` (`if self._resumed_from_checkpoint`) is therefore always vacuously true whenever `_resume_parent_checkpoint_id` is not None, and always vacuously false otherwise — making the boolean flag carry no additional information. The test is well-structured and covers the chain assertion correctly.

### Suggestions
- Consider resetting `_resumed_from_checkpoint` and `_resume_parent_checkpoint_id` in the `finally` block (or a dedicated cleanup path) rather than only on the success path, so that a failed `run_until_convergence` does not leave stale resume state if the runner is reused. This is a pre-existing issue that also applies to `_resumed_from_checkpoint`.
- The docstring for `_mark_resumed` says 'Optionally set the current iteration and max iterations' but `checkpoint_id` is now a required parameter—consider updating the docstring to reflect the new signature.
- Consider resetting `_resumed_from_checkpoint` and `_resume_parent_checkpoint_id` in the `finally` block (alongside `self._running = False`) so they are cleared even when an exception interrupts the run. This prevents stale resume state if the runner instance is reused after a failure.
- Add a follow-up assertion in `test_resumed_workflow_keeps_previous_checkpoint_id_chain` (or a separate test) that runs the workflow again without a checkpoint_id after the resumed run, and verifies the first checkpoint of that fresh run has `previous_checkpoint_id == None`. This would cover the reset logic on lines 157-160 of `_runner.py`.
- Consider adding a focused unit test for `_mark_resumed` to verify that it correctly sets `_resume_parent_checkpoint_id`, `_resumed_from_checkpoint`, and `_iteration`, since its signature changed from `(iteration)` to `(iteration, checkpoint_id)`.
- The boolean flag `_resumed_from_checkpoint` is now fully redundant with `_resume_parent_checkpoint_id is not None`. All usages of the flag (including the pre-existing one that guards message-skipping on resume) could be replaced with a check on the ID field, reducing the number of synchronized state variables to maintain and eliminating the possibility of them drifting out of sync in future edits.

---

python/packages/core/tests/workflow/test_checkpoint.py
```diff
@@ -336,6 +336,94 @@ async def finish(self, message: str, ctx: WorkflowContext[Never, str]) -> None:
...
+    resumed_checkpoints = [
+        checkpoint for checkpoint in all_checkpoints if checkpoint.checkpoint_id not in initial_ids
+    ]
+
+    assert resumed_checkpoints, "Expected at least one new checkpoint after resume"
+    assert resumed_checkpoints[0].previous_checkpoint_id == restore_checkpoint.checkpoint_id
+
+    for i in range(1, len(resumed_checkpoints)):
+        assert resumed_checkpoints[i].previous_checkpoint_id == resumed_checkpoints[i - 1].checkpoint_id
+
```
moonbox3, line 425: Consider extending this test (or adding a new one) to also verify the reset behavior: after the resumed run completes, do a fresh `workflow.run("hello", stream=True)` and assert the first new checkpoint has `previous_checkpoint_id is None`. This covers the reset of `_resume_parent_checkpoint_id` on line 160 of `_runner.py`.

python/packages/core/agent_framework/_workflows/_runner.py
```diff
@@ -81,7 +82,9 @@ async def run_until_convergence(self) -> AsyncGenerator[WorkflowEvent, None]:
             raise WorkflowRunnerException("Runner is already running.")
 
         self._running = True
-        previous_checkpoint_id: CheckpointID | None = None
+        previous_checkpoint_id: CheckpointID | None = (
```
moonbox3: This condition is logically equivalent to just `self._resume_parent_checkpoint_id` because the two fields are always set and cleared together — `_resume_parent_checkpoint_id` is non-None iff `_resumed_from_checkpoint` is True. The guard adds noise without adding safety. Simplifying to the single expression also makes it obvious that the ID is the only thing that matters here.

```suggestion
        previous_checkpoint_id: CheckpointID | None = self._resume_parent_checkpoint_id
```

moonbox3:
Doesn't look like the proper changes from main were merged to this branch? There shouldn't be .NET changes in it.

davidahmann:
Yes, you were right: that branch had unrelated .NET commits mixed in.
I force-pushed a cleaned history rebased to `main` that now contains only the two Python workflow-checkpoint commits for #4588.
Validation: `uv run pytest packages/core/tests/workflow/test_checkpoint.py -k 'test_workflow_checkpoint_chaining_via_previous_checkpoint_id or test_resumed_workflow_keeps_previous_checkpoint_id_chain'` (pass), `uv run pytest packages/core/tests/workflow/test_runner.py -k resumed` (pass).
Blockers/Risks: CLA policy check is still pending.

python/packages/core/tests/workflow/test_checkpoint.py
```diff
@@ -336,6 +336,94 @@ async def finish(self, message: str, ctx: WorkflowContext[Never, str]) -> None:
...
+    restore_checkpoint = initial_checkpoints[1]
+    initial_ids = {checkpoint.checkpoint_id for checkpoint in initial_checkpoints}
+
+    resumed_start = StartExecutor(id="start")
+    resumed_middle = MiddleExecutor(id="middle")
+    resumed_finish = FinishExecutor(id="finish")
+
+    resumed_workflow = (
+        WorkflowBuilder(
```
TaoChenOSU, line 399: Since the executors are stateless, you can just create a workflow builder then build new workflow instances from it.

markwallace-microsoft:
<a href="https://github.com/microsoft/agent-framework/blob/ef5b41cfc5faf1c1c7c42397427b255b5fb6cb60/README.md">Python Test Coverage</a><details><summary>Python Test Coverage Report • </summary><table><tr><th>File</th><th>Stmts</th><th>Miss</th><th>Cover</th><th>Missing</th></tr><tbody><tr><td colspan="5"><b>packages/core/agent_framework/_workflows</b></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/ef5b41cfc5faf1c1c7c42397427b255b5fb6cb60/packages/core/agent_framework/_workflows/_runner.py">_runner.py</a></td><td>176</td><td>2</td><td>98%</td><td><a href="https://github.com/microsoft/agent-framework/blob/ef5b41cfc5faf1c1c7c42397427b255b5fb6cb60/packages/core/agent_framework/_workflows/_runner.py#L306-L307">306&ndash;307</a></td></tr><tr><td><b>TOTAL</b></td><td><b>23589</b></td><td><b>2663</b></td><td><b>88%</b></td><td>&nbsp;</td></tr></tbody></table></details>

## Python Unit Test Overview
| 5144 | 20 :zzz: | 0 :x: | 0 :fire: | 1m 23s :stopwatch: |

eavanvalkenburg:

PR #4601, opened 2026-03-10
Python: Added AgentMiddleware to Copilot and Claude agents

dmytrostruk:
### Motivation and Context

Added `AgentMiddleware` support to Copilot and Claude agents.

### Contribution Checklist


markwallace-microsoft:
<a href="https://github.com/microsoft/agent-framework/blob/5b10bd424db728d9a2d1acb4a9d63cd43b9b1fd2/README.md">Python Test Coverage</a><details><summary>Python Test Coverage Report • </summary><table><tr><th>File</th><th>Stmts</th><th>Miss</th><th>Cover</th><th>Missing</th></tr><tbody><tr><td><b>TOTAL</b></td><td><b>22700</b></td><td><b>2601</b></td><td><b>88%</b></td><td>&nbsp;</td></tr></tbody></table><i>report-only-changed-files is enabled. No files were changed during this commit :)</i></details>

## Python Unit Test Overview
| 4964 | 20 :zzz: | 0 :x: | 0 :fire: | 1m 19s :stopwatch: |

python/packages/github_copilot/agent_framework_github_copilot/_agent.py
```diff
@@ -640,3 +648,11 @@ async def _resume_session(self, session_id: str, streaming: bool) -> CopilotSess
             config["mcp_servers"] = self._mcp_servers
 
         return await self._client.resume_session(session_id, config)
+
+
+class GitHubCopilotAgent(
+    AgentMiddlewareLayer,
+    RawGitHubCopilotAgent[OptionsT],
```
eavanvalkenburg, line 661: Should this also have the telemetry layer?

python/packages/claude/agent_framework_claude/_agent.py
```diff
@@ -620,15 +625,24 @@ def run(
             When stream=False: An Awaitable[AgentResponse] with the complete response.
         """
```
moonbox3, line 626: `middleware` is accepted as a named parameter in `run()` but never forwarded here. `AgentMiddlewareLayer` passes combined function/chat middleware via `kwargs["middleware"]`, which is now intercepted by the named param and silently dropped instead of reaching `_get_stream`.

```suggestion
        response = self._run_stream_impl(messages=messages, session=session, options=options, middleware=middleware, **kwargs)
```

python/packages/github_copilot/agent_framework_github_copilot/_agent.py
```diff
@@ -348,16 +351,26 @@ def run(
         Raises:
             AgentException: If the request fails.
         """
+        options = cast(OptionsT | None, kwargs.pop("options", None))
         if stream:
+            return self._run_stream_impl(messages=messages, session=session, options=options, **kwargs)
```
moonbox3, line 356: Same issue: `middleware` (containing combined function/chat middleware from `AgentMiddlewareLayer`) is captured by the named param in `run()` and never forwarded to `_run_stream_impl`, silently discarding any function/chat middleware.

```suggestion
            return self._run_stream_impl(messages=messages, session=session, options=options, middleware=middleware, **kwargs)
```

python/packages/github_copilot/agent_framework_github_copilot/_agent.py
```diff
@@ -348,16 +351,26 @@ def run(
...
+        return self._run_impl(messages=messages, session=session, options=options, **kwargs)
```
moonbox3, line 357: `middleware` not forwarded to `_run_impl` either — same silent drop for the non-streaming path.

```suggestion
        return self._run_impl(messages=messages, session=session, options=options, middleware=middleware, **kwargs)
```

python/packages/github_copilot/agent_framework_github_copilot/_agent.py
```diff
@@ -302,7 +304,7 @@ def run(
         *,
```
moonbox3, line 304: `options` was previously an explicitly typed parameter in this overload and is now completely absent from the signature. `AgentMiddlewareLayer._middleware_handler` still passes `options=context.options` explicitly, which works at runtime (falls into `**kwargs`), but this is a type-breaking change for any typed callers of `GitHubCopilotAgent.run()`.

python/packages/github_copilot/agent_framework_github_copilot/_agent.py
```diff
@@ -348,16 +351,26 @@ def run(
         Raises:
             AgentException: If the request fails.
         """
+        options = cast(OptionsT | None, kwargs.pop("options", None))
```
moonbox3, line 354: `cast()` silently coerces whatever is in kwargs to `OptionsT | None` with no runtime check. If a caller passes `options` with a wrong type, the error will surface much later (or not at all). Consider validating the extracted value before use.

```suggestion
        options = kwargs.pop("options", None)
        if options is not None and not isinstance(options, (dict, type(None))):
            raise TypeError(f"Expected options to be a mapping or None, got {type(options).__name__}")
        options = cast(OptionsT | None, options)
```

python/packages/github_copilot/tests/test_github_copilot_agent.py
```diff
@@ -459,6 +461,97 @@ def mock_on(handler: Any) -> Any:
...
+        agent = GitHubCopilotAgent(client=mock_client, middleware=[tracking_middleware])
+        response = await agent.run("Hello")
+
+        assert isinstance(response, AgentResponse)
+        assert response.text == "Test response"
+        assert calls == ["before", "after"]
+
+
 class TestGitHubCopilotAgentSessionManagement:
     """Test cases for session management."""
```
moonbox3, line 557: The per-run `middleware` parameter added to `run()` is never exercised. This is the primary new API surface introduced by the diff and has no test coverage.

```suggestion
async def test_run_supports_per_run_middleware(
    self,
    mock_client: MagicMock,
    mock_session: MagicMock,
    assistant_message_event: SessionEvent,
) -> None:
    """Test that middleware passed directly to run() is executed."""
    mock_session.send_and_wait.return_value = assistant_message_event
    calls: list[str] = []

    async def run_middleware(context: AgentContext, call_next: Any) -> None:
        calls.append("run_before")
        await call_next()
        calls.append("run_after")

    agent = GitHubCopilotAgent(client=mock_client)
    response = await agent.run("Hello", middleware=[run_middleware])

    assert isinstance(response, AgentResponse)
    assert calls == ["run_before", "run_after"]
```

python/packages/claude/tests/test_claude_agent.py
```diff
@@ -434,6 +443,168 @@ async def test_run_stream_raises_on_result_message_error(self) -> None:
...
+                            message_id=update.message_id,
+                            raw_representation=update.raw_representation,
+                        )
+
+                    context.stream_transform_hooks.append(add_prefix)
+                await call_next()
+
+        with patch("agent_framework_claude._agent.ClaudeSDKClient", return_value=mock_client):
+            agent = ClaudeAgent(middleware=[PrefixDeltaMiddleware()])
+            updates: list[AgentResponseUpdate] = []
```
moonbox3, line 557: No test covers passing `middleware` to `run()` at call time. The constructor-level middleware path is tested, but the per-run path added to the `run()` signature is dead code from a testing perspective.

```suggestion
async def test_run_supports_per_run_middleware(self) -> None:
    """Test that middleware passed to run() at call time is executed."""
    from claude_agent_sdk import AssistantMessage, ResultMessage, TextBlock
    from claude_agent_sdk.types import StreamEvent

    messages = [
        StreamEvent(
            event={"type": "content_block_delta", "delta": {"type": "text_delta", "text": "Hi"}},
            uuid="e1", session_id="s1",
        ),
        AssistantMessage(content=[TextBlock(text="Hi")], model="claude-sonnet"),
        ResultMessage(subtype="success", duration_ms=100, duration_api_ms=50,
                      is_error=False, num_turns=1, session_id="s1"),
    ]
    mock_client = self._create_mock_client(messages)
    calls: list[str] = []

    async def run_middleware(context: AgentContext, call_next: Any) -> None:
        calls.append("run_before")
        await call_next()
        calls.append("run_after")

    with patch("agent_framework_claude._agent.ClaudeSDKClient", return_value=mock_client):
        agent = ClaudeAgent()
        response = await agent.run("Hello", middleware=[run_middleware])

    assert isinstance(response, AgentResponse)
    assert calls == ["run_before", "run_after"]
```

python/packages/github_copilot/agent_framework_github_copilot/_agent.py
```diff
@@ -302,7 +304,7 @@ def run(
         *,
         stream: Literal[False] = False,
         session: AgentSession | None = None,
-        options: OptionsT | None = None,
+        middleware: Sequence[AgentMiddlewareTypes] | None = None,
```
moonbox3, line 307: Replacing `options` with `middleware` here (and in the other two overloads) is a breaking API change. `options` was a typed parameter that provided compile-time safety for runtime configuration; hiding it in **kwargs removes that safety. `RawGitHubCopilotAgent` also has no need to declare `middleware` explicitly—AgentMiddlewareLayer intercepts `run()` before the raw class sees it, and any residual `middleware` kwarg forwarded through context.kwargs is already absorbed by **kwargs. Restore `options` as an explicit parameter and drop the `middleware` declaration from all three overloads.

```suggestion
        options: OptionsT | None = None,
```

python/packages/github_copilot/agent_framework_github_copilot/_agent.py
```diff
@@ -313,7 +315,7 @@ def run(
         *,
         stream: Literal[True],
         session: AgentSession | None = None,
-        options: OptionsT | None = None,
+        middleware: Sequence[AgentMiddlewareTypes] | None = None,
```
moonbox3, line 318: Same issue as the non-streaming overload: `options` should remain explicit and typed here; `middleware` should not appear in the raw class signature.

```suggestion
        options: OptionsT | None = None,
```

moonbox3 (commented):
## Automated Code Review

### ✗ Correctness
> The refactoring introduces `AgentMiddlewareLayer` into both `ClaudeAgent` and `GitHubCopilotAgent` MROs, which is conceptually correct. However, there is a concrete correctness bug in both raw agents: `AgentMiddlewareLayer.run()` deliberately forwards combined function/chat middleware via `combined_kwargs["middleware"]` and expects it to flow downstream (through `_run_stream_impl`/`_run_impl`) to the underlying chat client. The new `middleware` named parameter added to `RawClaudeAgent.run()` and `RawGitHubCopilotAgent.run()` intercepts and silently drops this value — it is never forwarded to `_run_stream_impl` or `_run_impl`. Additionally, `options` was removed as an explicit typed parameter from `RawGitHubCopilotAgent.run()`'s overloads (replaced by `middleware`), while `AgentMiddlewareLayer._middleware_handler` still passes `options=` explicitly; this works at runtime via `**kwargs` extraction but is a type-safety regression for direct callers. The tests only cover agent-level middleware and would not catch the dropped function/chat middleware.

### ✓ Security Reliability
> The diff introduces a middleware layer (AgentMiddlewareLayer) into both ClaudeAgent and GitHubCopilotAgent via multiple inheritance. From a security and reliability standpoint the changes are largely sound, but there is one notable reliability issue: in GitHubCopilotAgent.run(), the previous overloads declared `options: OptionsT | None = None` as an explicit typed parameter, but the new overloads replace it with `middleware` and move `options` into **kwargs, relying on `kwargs.pop('options', None)` with a cast. This silent removal of a typed parameter breaks the public API contract without deprecation and can cause silent option-dropping if callers pass options as a keyword argument positionally, though this is a breaking-change / API-reliability concern rather than a security risk. No secrets are hardcoded, no unsafe deserialization is present, no injection risks are introduced, and the middleware callable path uses `Any` typing for `call_next` in tests which is acceptable in test code. The `cast()` used for extracting options from kwargs bypasses type safety and could silently discard a wrongly typed value without raising, which is a mild reliability concern.

### ✗ Test Coverage
> The diff adds middleware support (AgentMiddlewareLayer) to both ClaudeAgent and GitHubCopilotAgent, and includes 6 new tests covering class-based middleware, function-based middleware, and streaming transform hooks. However, the new per-run `middleware` parameter added to `run()` in both agents is never tested—there are zero tests that pass `middleware` at call time rather than construction time, and none that verify combined constructor+per-run middleware ordering. Additionally, the GitHubCopilotAgent `run()` overloads silently dropped the explicit `options` parameter (moved to `**kwargs`), but no test verifies that runtime options still route correctly through the refactored `_run_impl`/`_run_stream_impl` path. Multiple-middleware chaining order is also not covered by any test.

### ✗ Design Approach
> The diff adds AgentMiddlewareLayer to both ClaudeAgent and GitHubCopilotAgent by splitting each into a raw base class and a composed final class. The approach is architecturally sound for the Claude agent, but contains a breaking API change in RawGitHubCopilotAgent: the typed `options` parameter is removed from all three run() overloads and replaced with `middleware`, pushing `options` into untyped `**kwargs`. This silently removes type-checking and IDE support for `options` from the public API. Additionally, adding `middleware` explicitly to the raw agent signatures is a leaky abstraction—the middleware routing concern belongs exclusively to AgentMiddlewareLayer, and the raw classes should accept it transparently via `**kwargs` rather than declaring it as a named parameter.

### Flagged Issues
- [ ] In `RawClaudeAgent.run()` and `RawGitHubCopilotAgent.run()`, the explicit `middleware` named parameter captures — and silently drops — the combined function/chat middleware forwarded by `AgentMiddlewareLayer` via `combined_kwargs["middleware"]`. Since `middleware` is now an explicit named param rather than flowing through `**kwargs`, it is never forwarded to `_run_stream_impl` or `_run_impl`, breaking function/chat middleware for composed agents.
- [ ] Per-run `middleware` parameter added to `run()` in both agents is entirely untested — there is no test that calls `agent.run(..., middleware=[...])` and verifies the per-run middleware executes, making it impossible to confirm the feature works at all.
- [ ] The GitHubCopilotAgent `run()` overloads removed the explicit `options` parameter (replaced by `middleware`), with options now consumed via `kwargs.pop('options', None)`. No regression test verifies that runtime options (e.g., model, timeout) still reach `_run_impl`/`_run_stream_impl` correctly after this refactor.
- [ ] In RawGitHubCopilotAgent.run(), the `options` parameter is removed from all three typed overload signatures and replaced with `middleware`. This is a breaking change: callers of GitHubCopilotAgent.run(options=...) lose static type-checking and IDE autocomplete. The options kwarg is silently buried in **kwargs and extracted via `cast(OptionsT | None, kwargs.pop('options', None))`, which bypasses the TypedDict-based typing that made it safe. `options` should remain an explicit typed parameter in all overloads, and `middleware` should not appear in the raw class's overloads at all—it is AgentMiddlewareLayer's concern and can flow through **kwargs without an explicit declaration.

### Suggestions
- Remove `middleware` from the explicit named parameters of `RawClaudeAgent.run()` and `RawGitHubCopilotAgent.run()` so it continues to flow via `**kwargs` to `_run_stream_impl`/`_run_impl`, consistent with how `AgentMiddlewareLayer` expects to forward it. If the intent is only to prevent 'unexpected keyword argument' errors, forward it explicitly instead.
- `options` was a typed, explicitly documented parameter in `RawGitHubCopilotAgent.run()` overloads before this diff and has been silently replaced by `middleware`. Consider keeping `options` in the overloads (even alongside `middleware`) to avoid a breaking type-checking regression for existing callers who relied on IDE autocomplete or mypy checking for `options`.
- In `RawGitHubCopilotAgent.run()`, `options` was previously an explicit, typed overload parameter and is now silently moved into `**kwargs` with a `cast()`. Consider keeping `options` as an explicit keyword parameter in the overloads (alongside `middleware`) to preserve type safety and avoid silent option-dropping for existing callers.
- The `cast(OptionsT | None, kwargs.pop('options', None))` in `_agent.py` (GitHub Copilot) bypasses runtime type checking. If a caller accidentally passes a wrong type, it will be silently accepted. A runtime isinstance guard or at least a comment explaining the invariant would improve reliability.
- The new `_run_stream_impl` / `_run_impl` split on both agents leaves `**kwargs` forwarded without any validation. Consider documenting (or asserting) that no unexpected kwargs are passed, to avoid silently ignoring unexpected arguments.
- Add a test that passes `middleware` directly to `run()` (e.g., `agent.run('Hello', middleware=[TrackingMiddleware()])`) and asserts the middleware executes.
- Add a test that combines constructor middleware and per-run middleware and asserts both execute in the correct order (constructor first, then per-run).
- Add a regression test for GitHubCopilotAgent that passes `options` via `run('Hello', options={...})` and asserts the options are forwarded to the underlying session/client call, to guard against the kwarg extraction breakage.
- Add a test with two middleware in sequence to verify chaining order — currently all tests use exactly one middleware.
- The `middleware` parameter added to RawClaudeAgent.run()'s three overloads (non-streaming, streaming, and base) is a leaky abstraction. AgentMiddlewareLayer._middleware_handler() passes `middleware` to super().run() via **context.kwargs, so it already arrives through **kwargs in the raw class. Declaring it explicitly makes the raw class aware of a concern it doesn't implement and silently drops—this should be left as an implicit **kwargs passthrough with a brief inline comment explaining why it's intentionally ignored.
- The new `_run_stream_impl` helper in RawClaudeAgent silently drops the `middleware` kwarg received from AgentMiddlewareLayer (via context.kwargs → super().run() → _run_stream_impl's **kwargs). Since this will confuse future maintainers, a brief comment noting that `middleware` is intentionally absorbed and not forwarded—because it is handled by AgentMiddlewareLayer before reaching this point—would prevent future bugs from someone trying to 'fix' the drop.

---

python/packages/github_copilot/agent_framework_github_copilot/_agent.py
```diff
@@ -323,7 +325,7 @@ def run(
         *,
         stream: bool = False,
         session: AgentSession | None = None,
-        options: OptionsT | None = None,
+        middleware: Sequence[AgentMiddlewareTypes] | None = None,
```
moonbox3, line 328: Same issue on the base overload. Restoring `options` and removing `middleware` here also lets you remove the `cast(OptionsT | None, kwargs.pop('options', None))` inside the body and go back to the direct typed `options` parameter.

```suggestion
        options: OptionsT | None = None,
```

python/packages/claude/agent_framework_claude/_agent.py
```diff
@@ -581,6 +582,7 @@ def run(
         *,
         stream: Literal[False] = ...,
         session: AgentSession | None = None,
```
moonbox3, line 584: Declaring `middleware` explicitly in RawClaudeAgent's overloads leaks AgentMiddlewareLayer's routing concern into the raw class. The parameter is accepted and immediately discarded (never forwarded to _run_stream_impl). Since it arrives through **kwargs anyway when AgentMiddlewareLayer calls super().run(), it can simply be absorbed there without an explicit declaration.

```suggestion
        **kwargs: Any,
```

python/packages/claude/agent_framework_claude/_agent.py
```diff
@@ -591,6 +593,7 @@ def run(
         *,
         stream: Literal[True],
         session: AgentSession | None = None,
```
moonbox3, line 595: Same leaky abstraction as the non-streaming overload: `middleware` is declared but silently dropped. Remove it from this overload as well.

python/packages/claude/agent_framework_claude/_agent.py
```diff
@@ -600,6 +603,7 @@ def run(
         *,
         stream: bool = False,
         session: AgentSession | None = None,
```
moonbox3, line 605: Same issue on the base overload. Dropping the explicit `middleware` declaration from all three overloads keeps the raw class unaware of the middleware routing mechanism.

python/packages/claude/agent_framework_claude/_agent.py
```diff
@@ -620,15 +625,24 @@ def run(
             When stream=False: An Awaitable[AgentResponse] with the complete response.
         """
         options = kwargs.pop("options", None)
-        response = ResponseStream(
-            self._get_stream(messages, session=session, options=options, **kwargs),
-            finalizer=self._finalize_response,
-        )
-
+        response = self._run_stream_impl(messages=messages, session=session, options=options, **kwargs)
```
eavanvalkenburg, line 628: is this change needed? we are already wrapping a call to another function, no real need to do that twice (_get_stream and _run_stream_impl)

python/packages/github_copilot/agent_framework_github_copilot/_agent.py
```diff
@@ -302,7 +304,7 @@ def run(
         *,
         stream: Literal[False] = False,
         session: AgentSession | None = None,
-        options: OptionsT | None = None,
```
eavanvalkenburg, line 305: this shouldn't be removed, should it?

python/packages/github_copilot/agent_framework_github_copilot/_agent.py
```diff
@@ -338,8 +340,9 @@ def run(
         Keyword Args:
             stream: Whether to stream the response. Defaults to False.
             session: The conversation session associated with the message(s).
-            options: Runtime options (model, timeout, etc.).
-            kwargs: Additional keyword arguments.
+            middleware: Runtime middleware parameter accepted for compatibility with middleware layer routing.
+            kwargs: Additional keyword arguments, including ``options`` for runtime options
```
eavanvalkenburg, line 344: we should not reintroduce kwargs, we are trying to get rid of them, use `options`

Issue #4556, opened 2026-03-09
.NET: Consider adding finish reason to AgentResponse and AgentResponseUpdate


westey-m:
It's hard to tell whether a run completed due to content filtering rules without having a finish reason, so we should consider replicating that from ChatResponse across

PR #4617, opened 2026-03-11
.NET: Add FinishReason to AgentResponses

westey-m:
### Motivation and Context

It's currently difficult to determine when a ChatClientAgent completed because of content filtering, so adding FinishReason to AgentRunResponse, and mapping where we have enough data to do so.

#4556

### Description

- Add FinishReason to AgentResponse and AgentResponseUpdate

### Contribution Checklist

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/AgentResponseExtensions.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/AgentResponseExtensions.cs
@@ -34,7 +34,7 @@ internal static class AgentResponseExtensions
         var chatCompletionChoices = new List<ChatCompletionChoice>();
         var index = 0;
 
-        var finishReason = agentResponse.FinishReason?.ToString() ?? "stop"; // "stop" is a natural stop point; returning this by-default
+        var finishReason = agentResponse.FinishReason?.ToString() ?? ChatFinishReason.Stop.Value; // "stop" is a natural stop point; returning this by-default
 
         foreach (var message in agentResponse.Messages)
         {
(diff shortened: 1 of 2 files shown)


rogerbarreto (approved):

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/AgentResponseExtensions.cs
```diff
@@ -34,9 +34,7 @@ public static List<ChatCompletionChoice> ToChoices(this AgentResponse agentRespo
         var chatCompletionChoices = new List<ChatCompletionChoice>();
         var index = 0;
 
-        var finishReason = (agentResponse.RawRepresentation is ChatResponse { FinishReason: not null } chatResponse)
-            ? chatResponse.FinishReason.ToString()
-            : "stop"; // "stop" is a natural stop point; returning this by-default
+        var finishReason = agentResponse.FinishReason?.ToString() ?? "stop"; // "stop" is a natural stop point; returning this by-default
```
rogerbarreto: Suggest moving the `stop` to a const in a default place or so that could be later on cofigured if necessary.

Assuming all empty finishReasons are `stop` feels very opiniated, I'm not sure this is a good approach. Thought?

dotnet/src/Microsoft.Agents.AI.Workflows/MessageMerger.cs
```diff
@@ -156,6 +157,7 @@ public AgentResponse ComputeMerged(string primaryResponseId, string? primaryAgen
                 createdTimes.Add(response.CreatedAt.Value);
             }
 
+            finishReason = response.FinishReason ?? finishReason;
```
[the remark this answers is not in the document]
westey-m: Fixed.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/AgentResponseExtensions.cs
westey-m: Switched the code to use `ChatFinishReason.Stop.Value`.  

> Assuming all empty finishReasons are stop feels very opiniated, I'm not sure this is a good approach. Thought?

I believe this value is required, and there is no unknown/null. I think stop is the best option out of the possible ones considering that.  Also note this is existing behavior.

Issue #4618, opened 2026-03-11
.NET: Python/.NET HITL: make approval tokens single-use and retry-safe


davidahmann:
## Problem
Repeated submission of the same human-approval response can double-apply governed actions or leave workflow approval state ambiguous.

## Why now
Human-in-the-loop approval is a first-class feature and governance boundary in Agent Framework. Replay-unsafe approval handling weakens auditability and deterministic workflow behavior.

## Current insufficiency
Approval handling is documented and emitted via workflow events, but there is no explicit cross-surface contract guaranteeing that one approval token can only be consumed once with stable replay semantics.

## Expected behavior
- First valid approval token submission commits exactly one decision.
- Replays of the same token are rejected or treated as idempotent no-ops with a stable reason.
- Approval receipts remain correlated to the original accepted submission.

## Validation requirements
- Add tests that replay the same approval response token.
- Assert single state transition and single correlated approval receipt.
- Verify deterministic behavior in Python and .NET event/contract surfaces where applicable.

## Scope map
- `docs/decisions/0006-userapproval.md`
- `python/packages/core/agent_framework/_middleware.py`
- `python/packages/core/agent_framework/_workflows/_events.py`
- `dotnet/src/Microsoft.Agents.AI.Workflows/RequestInfoEvent.cs`

lokitoth:
This should already be the case for .NET, because all are modeled as requests which get checked for RequestId matching a pending request (which is removed) [here](https://github.com/microsoft/agent-framework/blob/main/dotnet/src/Microsoft.Agents.AI.Workflows/InProc/InProcessRunnerContext.cs#L157-L160).

PR #4620, opened 2026-03-11
Python/core: make request-info approval tokens replay-safe

davidahmann:
## Problem
Request-info approval request IDs are intended to be single-use. There was no direct regression test that proves a second submission for the same request ID is rejected after the first accepted response.

## Why now
Human-in-the-loop retries can replay stale request IDs; without explicit regression coverage, single-use token behavior can regress silently.

## What changed
- Added a workflow test that sends one valid response for a pending request ID.
- Asserted the request ID is removed from pending state after first use.
- Asserted a second response for the same request ID fails with a deterministic error.

## Validation
- `uv run pytest packages/core/tests/workflow/test_request_info_event_rehydrate.py -k replay --maxfail=1`

Refs #4618

### DIFF
--- a/python/packages/core/tests/workflow/test_request_info_event_rehydrate.py
+++ b/python/packages/core/tests/workflow/test_request_info_event_rehydrate.py
@@ -4,6 +4,8 @@ import json
 from dataclasses import dataclass, field
 from datetime import datetime, timezone
 
+import pytest
+
 from agent_framework import (
     FileCheckpointStorage,
     InMemoryCheckpointStorage,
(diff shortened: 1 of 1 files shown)


davidahmann:
This change protects HITL operator flow from replaying a previously accepted request-info token.
The patch is minimal and test-only: one regression test verifies the request ID is consumed on first response and a second submission fails with a deterministic error.
Validation: `uv run pytest packages/core/tests/workflow/test_request_info_event_rehydrate.py -k replay --maxfail=1` (pass).
Blockers/Risks: none currently identified.

eavanvalkenburg:

PR #4642, opened 2026-03-12
.NET: Validate SkillsInstructionPrompt contains {0} placeholder in FileAgentSkillsProvider

### LINKED ISSUE #4638: .NET: [Bug]: FileAgentSkillsProvider accepts SkillsInstructionPrompt without {0} and drops the generated skills list
## Summary
`FileAgentSkillsProvider` documents that `SkillsInstructionPrompt` must include a `{0}` placeholder for the generated skills list, but the current validation only checks whether `string.Format(template, string.Empty)` succeeds. As a result, templates with no placeholder are accepted and silently omit the generated skills list from the final instructions.

## Affected code
- `dotnet/src/Microsoft.Agents.AI/Skills/FileAgentSkillsProvider.cs`
- `dotnet/src/Microsoft.Agents.AI/Skills/FileAgentSkillsProviderOptions.cs`
- `dotnet/tests/Microsoft.Agents.AI.UnitTests/AgentSkills/FileAgentSkillsProviderTests.cs`

## Current behavior
A template like:

```csharp
new FileAgentSkillsProviderOptions
{
    SkillsInstructionPrompt = "Custom template"
}
```

is accepted, and the final instructions become just `Custom template` rather than including the discovered skills.

## Expected behavior
If a custom `SkillsInstructionPrompt` does not contain a usable `{0}` placeholder, construction should fail with `ArgumentException`, matching the documented contract.

## Why this is a bug
The public option docs say to use `{0}` as the placeholder for the generated skills list. The implementation currently enforces only valid format-string syntax, not presence of the placeholder itself, so valid-but-placeholder-free templates are accepted and produce incorrect output.

## Likely fix
Tighten validation in `BuildSkillsInstructionPrompt(...)` so it rejects templates that do not actually reference argument 0, and add a unit test for a placeholder-free template.

## Notes
I checked the current open `.NET` issues in this repo on 2026-03-12 and did not find a clear duplicate for this specific bug.

moonbox3:
### Motivation and Context

When a `SkillsInstructionPrompt` was provided without a `{0}` placeholder, `FileAgentSkillsProvider` silently accepted it and the generated skills list was dropped from the final prompt, producing incorrect AI instructions with no error.

Fixes #4638

### Description

The root cause was that the constructor only validated the prompt was a valid format string (via `string.Format` with an empty arg), but never checked that the `{0}` placeholder was actually present. The fix formats the prompt with a unique sentinel value and verifies the sentinel appears in the output; if it doesn't, an `ArgumentException` is thrown telling the caller the `{0}` placeholder is required. A new test confirms that a prompt without a placeholder is rejected.

### Contribution Checklist

---


dotnet/src/Microsoft.Agents.AI/Skills/FileAgentSkillsProvider.cs
```diff
@@ -172,10 +172,12 @@ private async Task<string> ReadSkillResourceAsync(string skillName, string resou
 
         if (options?.SkillsInstructionPrompt is { } optionsInstructions)
```
moonbox3, line 173: The sentinel round-trip is an indirect way to check for `{0}`. The simpler and more readable approach is to check the raw template directly with `optionsInstructions.Contains("{0}", StringComparison.Ordinal)` after the format-validation try/catch, removing the sentinel variable entirely.

```suggestion
            try
            {
                _ = string.Format(optionsInstructions, string.Empty);
            }
            catch (FormatException ex)
            {
                throw new ArgumentException(
                    "The provided SkillsInstructionPrompt is not a valid format string.",
                    nameof(options),
                    ex);
            }

            if (!optionsInstructions.Contains("{0}", StringComparison.Ordinal))
            {
                throw new ArgumentException(
                    "The provided SkillsInstructionPrompt must contain a '{0}' placeholder for the generated skills list.",
                    nameof(options));
            }

            promptTemplate = optionsInstructions;
```

dotnet/tests/Microsoft.Agents.AI.UnitTests/AgentSkills/FileAgentSkillsProviderTests.cs
```diff
@@ -127,6 +127,20 @@ public void Constructor_InvalidPromptTemplate_ThrowsArgumentException()
...
+        var options = new FileAgentSkillsProviderOptions
+        {
+            SkillsInstructionPrompt = "No placeholder here"
+        };
+
+        var ex = Assert.Throws<ArgumentException>(() => new FileAgentSkillsProvider(this._testRoot, options));
+        Assert.Contains("{0}", ex.Message);
+        Assert.Equal("options", ex.ParamName);
+    }
+
```
moonbox3, line 143: Good negative test. Consider adding a complementary positive test that verifies a `SkillsInstructionPrompt` with `{0}` is accepted and the custom template is actually applied to the output (e.g., call `GetSkillsPromptAsync` and assert the result contains the custom text). This would protect against regressions in the `promptTemplate = optionsInstructions` assignment.

dotnet/src/Microsoft.Agents.AI/Skills/FileAgentSkillsProvider.cs
```diff
@@ -184,6 +186,15 @@ private async Task<string> ReadSkillResourceAsync(string skillName, string resou
                     nameof(options),
                     ex);
             }
+
+            if (!formatted.Contains(sentinel, StringComparison.Ordinal))
```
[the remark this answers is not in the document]
moonbox3: Fixed in `aa5bb2ca`.

dotnet/src/Microsoft.Agents.AI/Skills/FileAgentSkillsProvider.cs
moonbox3: Fixed in `aa5bb2ca`.

dotnet/tests/Microsoft.Agents.AI.UnitTests/AgentSkills/FileAgentSkillsProviderTests.cs
moonbox3: Fixed in `aa5bb2ca`.

PR #4668, opened 2026-03-12
[BREAKING] Python: Add context mode to AgentExecutor

### LINKED ISSUE #3688: Python: Add context_mode to control what gets sent to agents in workflows
## Summary
Investigate adding a `context_mode` option to control what context gets passed between agents in a workflow.

## Background
From a user question:

> My scenario involves a deterministic sequential flow with multiple agents, so I'm using the Agent framework's sequential workflow. In my case, I don't need to pass the whole context to each agent in the workflow; I only want each agent to receive the final output from the previous one as input. From what I understand, the default behavior is to pass the entire context. How can I adjust this?

## Current Workaround
Today, users need to create a custom executor that transforms the conversation between agents:

```python
class LastOutputOnlyExecutor(Executor):
    @handler
    async def transform(
        self,
        messages: list[ChatMessage],
        ctx: WorkflowContext[list[ChatMessage]]
    ) -> None:
        # Extract only the last assistant message
        last_assistant = [m for m in messages if m.role == "assistant"][-1:]
        await ctx.send_message(last_assistant)
```

Then build the flow like:

```python
workflow = SequentialBuilder().participants([
    agent1,
    LastOutputOnlyExecutor(id="filter-1"),
    agent2,
    LastOutputOnlyExecutor(id="filter-2"),
    agent3,
]).build()
```

## Proposed Solution
Investigate adding a `context_mode` parameter to reduce verbosity. Possible modes:
- `full` (default) - pass entire conversation context
- `last_output` - pass only the last assistant message
- `last_n` - pass last N messages
- Custom filter function

## Scope
This feature should be evaluated for all orchestration patterns, not just sequential:
- Sequential workflows
- Concurrent workflows
- Group chat
- Other orchestration patterns

TaoChenOSU:
### Motivation and Context

Closes #3688 

### Description

1. Add `context_mode` and `context_filter` to `AgentExecutor`. These two control knobs are applied on the agent executor when it receives an `AgentExecutorResponse` that carries an agent response and the full conversation.
2. Update `sequential` to allow customers to configure if they want to chain the full conversation or not.
3. Add tests and a new sample.

> This is a breaking change because `full_conversation` is now a required argument in `AgentExecutorResponse`.

### Contribution Checklist

### DIFF
--- a/python/packages/core/agent_framework/_workflows/_agent_executor.py
+++ b/python/packages/core/agent_framework/_workflows/_agent_executor.py
@@ -169,7 +169,9 @@ class AgentExecutor(Executor):
         elif self._context_mode == "last_agent":
             self._cache.extend(prior.agent_response.messages)
         else:
-            assert self._context_filter is not None
+            if not self._context_filter:
+                # This should never happen due to validation in __init__, but mypy doesn't track that well
+                raise ValueError("context_filter function must be provided for 'custom' context_mode.")
             self._cache.extend(self._context_filter(prior.full_conversation))
 
         await self._run_agent_and_emit(ctx)


python/packages/core/agent_framework/_workflows/_agent_executor.py
```diff
@@ -57,7 +57,71 @@ class AgentExecutorResponse:
...
+class ContextMode:
+    """Configuration for how AgentExecutor should manage conversation context when receiving new messages.
+
+    Attributes:
+        filter_mode: Determines which incoming messages are included in the context provided to the agent.
+            - "full": Include all incoming messages (default).
+            - "last_agent": Include only messages from the agent response.
+            - "custom": Use the provided messages_filter callable to filter messages.
+        retain_cache: A flag indicating whether the executor should retain its internal cache upon receiving new
+            messages. If False, the cache will be cleared before processing new messages.
```
alliscode: Is this here to allow the `AgentExecutor` to queue up messages for a future round (e.g. tool response)? This feels like something we should avoid as it causes the local and potentially remote conversation state to diverge.

TaoChenOSU: No, it's for things like groupchat or handoff where the agent needs to cache responses from other agents.

python/packages/core/agent_framework/_workflows/_agent_executor.py
```diff
@@ -57,7 +57,71 @@ class AgentExecutorResponse:
...
+        after which the cache is cleared and the messages are added to the full conversation context. It's important
+        for the executor to be able to stage messages because in many multiple agent workflows, one agent may receive
+        many messages from different executors before it runs.
+    """
+
+    filter_mode: Literal["full", "last_agent", "custom"]
+    retain_cache: bool
+    messages_filter: Callable[[list[Message]], list[Message]] | None = None
+
+    def __init__(
```
eavanvalkenburg: does this need to be a dataclass, if you're already doing a manual `__init__`?

python/packages/core/agent_framework/_workflows/_agent_executor.py
```diff
@@ -57,7 +57,71 @@ class AgentExecutorResponse:
...
+        messages_filter: Callable[[list[Message]], list[Message]] | None = None,
+    ):
+        self.filter_mode = filter_mode
+        self.retain_cache = retain_cache
+        self.messages_filter = messages_filter
+        if filter_mode == "custom" and messages_filter is None:
+            raise ValueError("messages_filter must be provided when filter_mode is 'custom'.")
+
+    # Some common context modes
+    @staticmethod
```
eavanvalkenburg: these would make more sense as classmethods

python/packages/core/agent_framework/_workflows/_agent_executor.py
```diff
@@ -57,7 +57,71 @@ class AgentExecutorResponse:
...
+    def __init__(
+        self,
+        filter_mode: Literal["full", "last_agent", "custom"] = "full",
+        retain_cache: bool = True,
+        messages_filter: Callable[[list[Message]], list[Message]] | None = None,
+    ):
+        self.filter_mode = filter_mode
+        self.retain_cache = retain_cache
+        self.messages_filter = messages_filter
+        if filter_mode == "custom" and messages_filter is None:
```
eavanvalkenburg: if you use a dataclass, this is typically in a `_post_init_` method. But I would switch to a plain class

python/packages/core/agent_framework/_workflows/_agent_executor.py
```diff
@@ -83,13 +147,16 @@ def __init__(
         *,
         session: AgentSession | None = None,
         id: str | None = None,
+        context_mode: ContextMode | None = None,
```
eavanvalkenburg: this does mean people need to import another class, could we make this more lightweight, maybe a set of string literals/typed dicts?

eavanvalkenburg: this parameter is optional, so not sure if this is a breaking change perse

TaoChenOSU: They normally don't have to use the AgentExecutor directly.

TaoChenOSU: The breaking change comes from the `AgentExecutorResponse`.

markwallace-microsoft:
<a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/README.md">Python Test Coverage</a><details><summary>Python Test Coverage Report • </summary><table><tr><th>File</th><th>Stmts</th><th>Miss</th><th>Cover</th><th>Missing</th></tr><tbody><tr><td colspan="5"><b>packages/core/agent_framework/_workflows</b></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/core/agent_framework/_workflows/_agent_executor.py">_agent_executor.py</a></td><td>204</td><td>19</td><td>90%</td><td><a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/core/agent_framework/_workflows/_agent_executor.py#L109">109</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/core/agent_framework/_workflows/_agent_executor.py#L133">133</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/core/agent_framework/_workflows/_agent_executor.py#L174">174</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/core/agent_framework/_workflows/_agent_executor.py#L200-L201">200&ndash;201</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/core/agent_framework/_workflows/_agent_executor.py#L256-L257">256&ndash;257</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/core/agent_framework/_workflows/_agent_executor.py#L259-L260">259&ndash;260</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/core/agent_framework/_workflows/_agent_executor.py#L296-L298">296&ndash;298</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/core/agent_framework/_workflows/_agent_executor.py#L300">300</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/core/agent_framework/_workflows/_agent_executor.py#L310-L311">310&ndash;311</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/core/agent_framework/_workflows/_agent_executor.py#L410-L411">410&ndash;411</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/core/agent_framework/_workflows/_agent_executor.py#L476">476</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/core/agent_framework/_workflows/_agent_executor.py#L495">495</a></td></tr><tr><td colspan="5"><b>packages/orchestrations/agent_framework_orchestrations</b></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_concurrent.py">_concurrent.py</a></td><td>152</td><td>27</td><td>82%</td><td><a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_concurrent.py#L51">51</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_concurrent.py#L60-L61">60&ndash;61</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_concurrent.py#L69-L70">69&ndash;70</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_concurrent.py#L87-L88">87&ndash;88</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_concurrent.py#L93">93</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_concurrent.py#L117">117</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_concurrent.py#L122">122</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_concurrent.py#L127-L128">127&ndash;128</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_concurrent.py#L134">134</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_concurrent.py#L156">156</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_concurrent.py#L166">166</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_concurrent.py#L173">173</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_concurrent.py#L240">240</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_concurrent.py#L256">256</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_concurrent.py#L259">259</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_concurrent.py#L316">316</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_concurrent.py#L346">346</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_concurrent.py#L348-L349">348&ndash;349</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_concurrent.py#L351">351</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_concurrent.py#L356">356</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_concurrent.py#L369">369</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_concurrent.py#L373">373</a></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_orchestration_request_info.py">_orchestration_request_info.py</a></td><td>56</td><td>0</td><td>100%</td><td>&nbsp;</td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_sequential.py">_sequential.py</a></td><td>88</td><td>11</td><td>87%</td><td><a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_sequential.py#L73">73</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_sequential.py#L171">171</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_sequential.py#L182">182</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_sequential.py#L188">188</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_sequential.py#L219">219</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_sequential.py#L221-L222">221&ndash;222</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_sequential.py#L224">224</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_sequential.py#L229">229</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_sequential.py#L246">246</a>, <a href="https://github.com/microsoft/agent-framework/blob/9caba77c408f764cb3e64d6528dbb91021d540a9/packages/orchestrations/agent_framework_orchestrations/_sequential.py#L250">250</a></td></tr><tr><td><b>TOTAL</b></td><td><b>27099</b></td><td><b>3214</b></td><td><b>88%</b></td><td>&nbsp;</td></tr></tbody></table></details>

## Python Unit Test Overview
| 5276 | 20 :zzz: | 0 :x: | 0 :fire: | 1m 28s :stopwatch: |

moonbox3 (commented):
Question around: checkpoint serialization that can crash with custom `ContextMode`

`on_checkpoint_save` stores `self._context_mode` directly in the state dict. The checkpoint encoding system (`_checkpoint_encoding.py`) uses `pickle.dumps()` for non-JSON-native types. However, **Python's standard `pickle` cannot serialize nested functions or closures**.

`ContextMode.last_n()` creates a closure (`_last_n_messages`) as the `messages_filter` attribute. Any attempt to checkpoint an executor using `ContextMode.last_n()` — or any user-provided lambda/closure as `messages_filter` — will crash:

```
PicklingError: Can't pickle local object 'ContextMode.last_n.<locals>._last_n_messages'
```

Per `_runner.py:302-307`, this propagates as a `WorkflowCheckpointException` and **fails the entire checkpoint save**, not just the context_mode field.

Verified locally:
```python
import pickle

class Foo:
    @classmethod
    def with_closure(cls, n):
        def _inner(x):
            return x[-n:]
        return cls(_inner)
    def __init__(self, fn):
        self.fn = fn

pickle.dumps(Foo.with_closure(3))
# -> PicklingError: Can't pickle local object
```

Some ideas: either (a) don't include `context_mode` in checkpoint state — reconstruct it from constructor args on restore, or (b) implement a custom `__getstate__`/`__setstate__` on `ContextMode` that serializes the filter mode + parameters (e.g., store `("last_n", 5)`) rather than the closure itself.

Minor secondary note: the `on_checkpoint_restore` try/except around `self._context_mode = context_mode_payload` is dead code — a simple assignment can never raise, so the except branch is unreachable.

TaoChenOSU:
> Question around: checkpoint serialization that can crash with custom `ContextMode`
> 
> `on_checkpoint_save` stores `self._context_mode` directly in the state dict. The checkpoint encoding system (`_checkpoint_encoding.py`) uses `pickle.dumps()` for non-JSON-native types. However, **Python's standard `pickle` cannot serialize nested functions or closures**.
> 
> `ContextMode.last_n()` creates a closure (`_last_n_messages`) as the `messages_filter` attribute. Any attempt to checkpoint an executor using `ContextMode.last_n()` — or any user-provided lambda/closure as `messages_filter` — will crash:
> 
> ```
> PicklingError: Can't pickle local object 'ContextMode.last_n.<locals>._last_n_messages'
> ```
> 
> Per `_runner.py:302-307`, this propagates as a `WorkflowCheckpointException` and **fails the entire checkpoint save**, not just the context_mode field.
> 
> Verified locally:
> 
> ```python
> import pickle
> 
> class Foo:
>     @classmethod
>     def with_closure(cls, n):
>         def _inner(x):
>             return x[-n:]
>         return cls(_inner)
>     def __init__(self, fn):
>         self.fn = fn
> 
> pickle.dumps(Foo.with_closure(3))
> # -> PicklingError: Can't pickle local object
> ```
> 
> Some ideas: either (a) don't include `context_mode` in checkpoint state — reconstruct it from constructor args on restore, or (b) implement a custom `__getstate__`/`__setstate__` on `ContextMode` that serializes the filter mode + parameters (e.g., store `("last_n", 5)`) rather than the closure itself.
> 
> Minor secondary note: the `on_checkpoint_restore` try/except around `self._context_mode = context_mode_payload` is dead code — a simple assignment can never raise, so the except branch is unreachable.

Good catch! I think we can probably drop the `custom` mode and the `messages_filter` param. If complex message filtering is needed, custom executors is a better solution.

python/packages/core/agent_framework/_workflows/_agent_executor.py
```diff
@@ -237,6 +318,7 @@ async def on_checkpoint_save(self) -> dict[str, Any]:
             "cache": self._cache,
             "full_conversation": self._full_conversation,
             "agent_session": serialized_session,
+            "context_mode": self._context_mode,
```
moonbox3: **Pickle crash**: `self._context_mode` is stored directly in the checkpoint state dict and will be pickled. When `ContextMode.last_n()` is used, `messages_filter` is a closure that cannot be pickled, crashing checkpoint save with `PicklingError`. Serialize only the picklable fields (filter_mode, retain_cache, and for last_n the value of n), and reconstruct on restore — or implement `__getstate__`/`__setstate__` on ContextMode. A checkpoint round-trip test with each ContextMode variant would immediately surface this.

```suggestion
            "context_mode": {"filter_mode": self._context_mode.filter_mode, "retain_cache": self._context_mode.retain_cache},
```

python/packages/core/agent_framework/_workflows/_agent_executor.py
```diff
@@ -278,6 +360,16 @@ async def on_checkpoint_restore(self, state: dict[str, Any]) -> None:
...
             self._session = self._agent.create_session()
 
+        context_mode_payload = state.get("context_mode")
+        if context_mode_payload:
+            try:
+                self._context_mode = context_mode_payload
+            except Exception as exc:
+                logger.warning("Failed to restore context mode: %s", exc)
+                self._context_mode = ContextMode.default()
+        else:
```
moonbox3: **Dead code + missing validation**: `self._context_mode = context_mode_payload` is a plain attribute assignment that cannot raise — the try/except is unreachable dead code. Additionally, the payload is assigned without any type validation, so a corrupted checkpoint could assign an arbitrary object. Replace with an isinstance check, falling back to ContextMode.default() with a warning on unexpected types.

```suggestion
        context_mode_payload = state.get("context_mode")
        if isinstance(context_mode_payload, ContextMode):
            self._context_mode = context_mode_payload
        else:
            if context_mode_payload is not None:
                logger.warning("Unexpected context_mode type in checkpoint: %s", type(context_mode_payload))
            self._context_mode = ContextMode.default()
```

python/packages/core/agent_framework/_workflows/_agent_executor.py
```diff
@@ -143,19 +208,26 @@ async def from_response(
...
-        # Replace cache with full conversation if available, else fall back to agent_response messages.
-        source_messages = (
-            prior.full_conversation if prior.full_conversation is not None else prior.agent_response.messages
-        )
-        self._cache = list(source_messages)
+        if not self._context_mode.retain_cache:
+            self._cache.clear()
+
+        if self._context_mode.filter_mode == "full":
+            self._cache.extend(list(prior.full_conversation))
```
moonbox3: **Behavioral regression**: The old code did `self._cache = list(source_messages)` (full replace). Now with default ContextMode (retain_cache=True), existing cache entries are preserved and new messages are appended via extend. Since full_conversation is the complete history, this will produce duplicate messages if the cache already contains entries (e.g., from a prior AgentExecutorRequest with should_respond=False). The old replace semantics should be preserved for from_response to maintain backward compatibility, or this must be documented as a breaking change. Also add an else branch with a warning for unrecognized filter_mode values.

```suggestion
        if not self._context_mode.retain_cache:
            self._cache.clear()

        if self._context_mode.filter_mode == "full":
            self._cache.extend(list(prior.full_conversation))
        elif self._context_mode.filter_mode == "last_agent":
            self._cache.extend(list(prior.agent_response.messages))
        elif self._context_mode.filter_mode == "custom":
            self._cache.extend(list(self._context_mode.messages_filter(prior.full_conversation)))  # type: ignore
        else:
            logger.warning("Unknown filter_mode '%s', falling back to full conversation", self._context_mode.filter_mode)
            self._cache.extend(list(prior.full_conversation))
```

moonbox3 (commented):
## Automated Code Review

**Reviewers:** 4 | **Confidence:** 91%

### ✗ Correctness
> The diff introduces a ContextMode class for controlling conversation context in AgentExecutor, makes full_conversation required on AgentExecutorResponse, and updates all handlers to use the new context mode. There are two acknowledged-but-unfixed bugs (checkpoint pickle crash with closures, dead try/except) and a silent default-behavior change in from_response that switches from cache-replacement to cache-append semantics, which could cause duplicate messages in existing workflows.

### ✗ Security Reliability
> This diff introduces ContextMode for managing conversation context in AgentExecutor. There are two blocking reliability issues: (1) storing self._context_mode directly in checkpoint state will crash with PicklingError when ContextMode.last_n() is used because it captures a closure, and (2) the from_response handler changed from cache-replace to cache-extend semantics under the default ContextMode, which silently breaks existing workflows. Additionally, the on_checkpoint_restore has a dead try/except that cannot execute, and the restored context_mode payload is assigned without any type validation, which undermines checkpoint integrity.

### ✗ Test Coverage
> The diff introduces a substantial new ContextMode class with three filter modes, changes the default cache behavior in all four message handlers (extend instead of replace), adds checkpoint serialization for context_mode, and makes full_conversation non-optional — yet there are zero new tests for any of this functionality. All test changes in the diff are purely mechanical (adding the now-required full_conversation argument to existing constructors). There are no unit tests for ContextMode construction/validation, no integration tests verifying handler behavior under different context modes, no checkpoint round-trip tests for context_mode (which is known to crash with closures from last_n), and no tests covering the behavioral change from cache-replace to cache-extend.

### ✗ Design Approach
> The diff introduces a `ContextMode` class for controlling conversation context in `AgentExecutor`, makes `full_conversation` non-optional on `AgentExecutorResponse`, and updates all callsites. Three design problems remain unfixed from prior review: (1) `last_n()` embeds an unpicklable closure that will crash checkpoint save, (2) the try/except in `on_checkpoint_restore` wraps an assignment that cannot raise, making the except branch dead code. Two new design problems appear: (3) the default behaviour of `from_response` silently changed from cache-replace to cache-extend, which will duplicate messages for callers relying on the old replace semantics; and (4) `last_agent` filter mode is handled only in `from_response`—the mode is silently ignored in `from_str`, `from_message`, and `from_messages`, making the abstraction inconsistently applied and misleading to users.

### Flagged Issues
- [ ] on_checkpoint_save stores self._context_mode directly, but ContextMode.last_n() captures a closure (_last_n_messages) that cannot be pickled — any checkpoint save with last_n mode will crash with PicklingError, killing the entire checkpoint operation. Either serialize only the picklable constructor arguments (filter_mode, retain_cache, and for last_n the value of n) and reconstruct on restore, or implement __getstate__/__setstate__ on ContextMode.
- [ ] from_response default behavior silently changed from self._cache = list(source_messages) (replace) to self._cache.extend(...) (append). Under the default ContextMode (retain_cache=True), existing cache content — including staged messages from AgentExecutorRequest(should_respond=False) — is now preserved instead of replaced, which is a behavioral regression for all current callers that relied on from_response resetting the cache.
- [ ] last_agent filter_mode is silently a no-op in from_str, from_message, and from_messages. These handlers extend the cache unconditionally with all incoming messages and never branch on filter_mode=='last_agent'. Users who set ContextMode.last_agent() will get different and undocumented behavior depending on which handler fires.
- [ ] No tests exist for the ContextMode class (default(), last_agent(), last_n(n), ValueError on custom without filter) or for checkpoint round-trip serialization of context_mode — a round-trip test with ContextMode.last_n() would immediately surface the pickle crash.
- [ ] No tests cover handler behavior under different context modes (from_response with full/last_agent/custom, from_str/from_message/from_messages with retain_cache=False) or validate the changed cache extend-vs-replace semantics.

### Suggestions
- Remove the dead try/except in on_checkpoint_restore — self._context_mode = context_mode_payload is a plain assignment that cannot raise. The except branch is unreachable dead code. Replace with a straightforward conditional assignment.
- Add type validation when restoring context_mode from checkpoint state (e.g., isinstance check for ContextMode) to prevent corrupted or tampered checkpoint data from assigning an arbitrary object. Fall back to ContextMode.default() with a warning on type mismatch.
- Add a fallback else branch in from_response's filter_mode chain that logs a warning and falls back to full conversation, so that an unrecognized filter_mode doesn't silently drop all messages.
- Warn or raise when filter_mode='last_agent' or filter_mode='custom' is used with from_str/from_message/from_messages, since these handlers silently ignore filtering — the mode behaves identically to 'full' with no indication to the user.
- Centralize filtering logic in a single helper method so every handler (from_response, from_str, from_message, from_messages) applies filter_mode and retain_cache uniformly, rather than each handler reimplementing the logic inconsistently.
- Add a test that calls from_response twice in succession to verify and document the extend-vs-replace semantics under default ContextMode.
- Add unit tests for ContextMode covering all factories (default, last_agent, last_n including edge cases n=0 and n > len) and the ValueError on custom mode without a filter function.
- Add a checkpoint round-trip test for each ContextMode variant, including ContextMode.last_n() to document and catch the pickle failure.

---

python/packages/core/agent_framework/_workflows/_agent_executor.py
```diff
@@ -143,19 +208,26 @@ async def from_response(
...
+        elif self._context_mode.filter_mode == "last_agent":
+            self._cache.extend(list(prior.agent_response.messages))
+        elif self._context_mode.filter_mode == "custom":
+            self._cache.extend(list(self._context_mode.messages_filter(prior.full_conversation)))  # type: ignore
```
moonbox3: `messages_filter` is typed as `Callable | None`. The `# type: ignore` suppresses a real type-safety concern — if a ContextMode is constructed incorrectly or mutated, this will crash with `TypeError: 'NoneType' is not callable`. Add a runtime guard.

```suggestion
            filtered = self._context_mode.messages_filter(prior.full_conversation) if self._context_mode.messages_filter else prior.full_conversation
            self._cache.extend(list(filtered))
```

python/packages/core/agent_framework/_workflows/_agent_executor.py
```diff
@@ -143,19 +208,26 @@ async def from_response(
...
 
     @handler
     async def from_str(
         self, text: str, ctx: WorkflowContext[AgentExecutorResponse, AgentResponse | AgentResponseUpdate]
     ) -> None:
         """Accept a raw user prompt string and run the agent (one-shot)."""
-        self._cache = normalize_messages_input(text)
+        if not self._context_mode.retain_cache:
+            self._cache.clear()
+        self._cache.extend(normalize_messages_input(text))
```
moonbox3: **Inconsistent filter application**: `from_str` respects `retain_cache` but ignores `filter_mode` entirely — neither `last_agent` nor `custom` filtering is applied. A user setting `ContextMode(filter_mode='custom', messages_filter=my_fn)` would see filtering in `from_response` and `from_messages` but silently skipped here.

python/packages/core/agent_framework/_workflows/_agent_executor.py
```diff
@@ -165,7 +237,9 @@ async def from_message(
         ctx: WorkflowContext[AgentExecutorResponse, AgentResponse | AgentResponseUpdate],
     ) -> None:
         """Accept a single Message as input."""
-        self._cache = normalize_messages_input(message)
+        if not self._context_mode.retain_cache:
+            self._cache.clear()
+        self._cache.extend(normalize_messages_input(message))
```
moonbox3: Same inconsistency as `from_str`: `from_message` respects `retain_cache` but silently ignores `filter_mode`. Custom and last_agent modes have no effect.

python/packages/core/agent_framework/_workflows/_agent_executor.py
```diff
@@ -175,7 +249,14 @@ async def from_messages(
         ctx: WorkflowContext[AgentExecutorResponse, AgentResponse | AgentResponseUpdate],
     ) -> None:
         """Accept a list of chat inputs (strings or Message) as conversation context."""
-        self._cache = normalize_messages_input(messages)
+        if not self._context_mode.retain_cache:
+            self._cache.clear()
+
+        normalized_messages = normalize_messages_input(messages)
+        if self._context_mode.filter_mode == "custom":
+            normalized_messages = self._context_mode.messages_filter(normalized_messages)  # type: ignore
```
moonbox3: Same `# type: ignore` concern as in `from_response` — `messages_filter` could be None at runtime. Additionally, `from_messages` applies the custom filter while `from_str` and `from_message` skip it entirely, so the same ContextMode behaves differently depending on the handler. Centralise filtering in a shared helper.

```suggestion
            normalized_messages = self._context_mode.messages_filter(normalized_messages) if self._context_mode.messages_filter else normalized_messages
```

python/packages/core/agent_framework/_workflows/_agent_executor.py
```diff
@@ -175,7 +249,14 @@ async def from_messages(
...
+
+        self._cache.extend(normalized_messages)
         await self._run_agent_and_emit(ctx)
 
     @response_handler
```
moonbox3, line 216: `last_agent` filter_mode is never handled in `from_messages`. Users who construct `ContextMode.last_agent()` and trigger this handler will silently receive all messages instead. Either raise/warn for unsupported modes in each handler, or apply a shared filtering helper that maps each mode uniformly.

python/packages/core/tests/workflow/test_runner.py
```diff
@@ -341,7 +341,7 @@ async def test_runner_emits_runner_completion_for_agent_response_without_targets
 
     await ctx.send_message(
         WorkflowMessage(
-            data=AgentExecutorResponse("agent", AgentResponse()),
+            data=AgentExecutorResponse("agent", AgentResponse(), []),
```
moonbox3, line 344: This test was updated to pass the now-required empty list for full_conversation, but no new tests validate behavior when full_conversation contains actual messages or differs from agent_response.messages. Consider adding substantive tests for the new parameter.

python/packages/core/agent_framework/_workflows/_agent_executor.py
TaoChenOSU: `custom` mode and `messages_filter` have been dropped.

python/packages/core/agent_framework/_workflows/_agent_executor.py
TaoChenOSU: `retain_cache` together with `full` context mode should be used with caution.

python/packages/core/agent_framework/_workflows/_agent_executor.py
TaoChenOSU: filter_mode doesn't apply here

python/packages/core/agent_framework/_workflows/_agent_executor.py
TaoChenOSU: No way to know the messages from the last agents in `last_agent` mode

python/packages/core/agent_framework/_workflows/_agent_executor.py
```diff
@@ -83,13 +83,25 @@ def __init__(
         *,
         session: AgentSession | None = None,
         id: str | None = None,
+        context_mode: Literal["full", "last_agent", "custom"] | None = None,
```
alliscode, line 86: If filter is provided but context_mode != "custom" will the filter be ignored? Would it be better to have it so that you provide a custom filter, or you specify a mode, but not both?

TaoChenOSU: Yes, but I think if we try to be smart about it, it may lead to a leaky API, where the mode setting is split into two parameters: `full` and `last_agent` via the `context_mode` param and implicit `custom` mode via the filter param. And even with this, we will still have to potentially ignore one of the parameters if filter is provided but the mode says `last_agent`.

PR #4703, opened 2026-03-14
Python: fix: include `status` field on `function_call` and `function_call_output` items

### LINKED ISSUE #4701: Python: Responses API: missing status field on function_call / function_call_output items
When using AG-UI client-side tools with `AzureOpenAIResponsesClient`, follow-up requests that include tool call history fail with `400 Bad Request` because the Responses API requires a `status` field on `function_call` and `function_call_output` items.

The root cause is in `RawOpenAIResponsesClient._prepare_content_for_openai`:
- `function_call` items get `"status": None` (line ~1168) — API rejects null
- `function_call_output` items omit `status` entirely (line ~1194) — API requires it

The API expects `status` to be `"in_progress"`, `"completed"`, or `"incomplete"`.

This only surfaces with AG-UI client-side (declaration-only) tools because server-side tools use `previous_response_id` chaining and never re-serialize tool history as input.

**Workaround:** monkey-patch to add `"status": "completed"` on both item types.

**Versions:** agent-framework-core 1.0.0rc3, agent-framework-ag-ui 1.0.0b260304, Python 3.13

LEDazzio01:
## Motivation and Context

Fixes #4701.

When replaying conversation history that includes tool calls through the Responses API, the `_prepare_content_for_openai` method produced items that violated the API schema:

1. **`function_call` items** set `"status": None` unconditionally. For AG-UI client-side tools where `additional_properties` does not carry an explicit `status`, this resulted in `null` being sent to the API, causing a `400 Bad Request`.
2. **`function_call_output` items** omitted the `status` field entirely, also causing API rejections.

### `_responses_client.py`
- **`function_call`**: Resolve `status` from `content.additional_properties` when available, falling back to `"completed"` when absent or `None`.
- **`function_call_output`**: Always include `"status": "completed"`.

### `test_openai_responses_client.py`
Added 4 unit tests:
- `test_prepare_content_function_call_defaults_status_to_completed` — verifies default fallback
- `test_prepare_content_function_call_preserves_explicit_status` — verifies explicit status is preserved
- `test_prepare_content_function_result_includes_status_completed` — verifies function_call_output always has status
- `test_prepare_content_function_result_none_result` — verifies None result maps to empty string with status

### DIFF
--- a/python/packages/core/agent_framework/openai/_responses_client.py
+++ b/python/packages/core/agent_framework/openai/_responses_client.py
@@ -1036,16 +1036,20 @@ class RawOpenAIResponsesClient(  # type: ignore[misc]
                     "type": "function_call",
                     "name": content.name,
                     "arguments": content.arguments,
-                    "status": None,
+                    "status": (
+                        content.additional_properties.get("status")
+                        if content.additional_properties
+                        else None
+                    ) or "completed",
                 }
             case "function_result":
                 # call_id for the result needs to be the same as the call_id for the function call
-                args: dict[str, Any] = {
+                return {
                     "call_id": content.call_id,
                     "type": "function_call_output",
                     "output": content.result if content.result is not None else "",
+                    "status": "completed",
                 }
-                return args
             case "function_approval_request":
                 return {
                     "type": "mcp_approval_request",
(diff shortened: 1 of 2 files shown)


LEDazzio01:
Closing this in favor of #4704 by @ltwlf, who originally reported the issue and has a more comprehensive fix that also includes filename passthrough for media parts. Thanks for the great catch and fix, @ltwlf! 🙌

PR #4716, opened 2026-03-16
Python: Fix _deduplicate_messages catch-all branch dropping valid repeated messages

### LINKED ISSUE #4682: Python: ag-ui _deduplicate_messages catch-all branch can drop valid repeated messages
The catch-all branch in `_deduplicate_messages` (lines 244-253 of `_message_adapters.py`) deduplicates by `(role, hash(content_str))`, which means any two messages with the same role and identical content are treated as duplicates. This can incorrectly drop valid repeated messages — e.g., a user saying "yes" to confirm two separate things.

The tool-specific dedup branches (tool results by `call_id`, assistant tool calls by `call_id` tuple) are correct and should remain. The fix should either remove the catch-all branch or scope it more tightly (e.g., only dedup adjacent identical messages, or use message IDs from the AG-UI protocol).

Ref: https://github.com/microsoft/agent-framework/discussions/4640

moonbox3:
### Motivation and Context

The catch-all `else` branch in `_deduplicate_messages` hashed message content and skipped any message whose role+content matched a previously seen message. This incorrectly dropped legitimate repeated user messages (e.g., repeated "yes" confirmations) and repeated system messages, silently altering conversation history.

Fixes #4682

### Description

The root cause was that the `else` branch applied content-based deduplication to all message types not handled by the earlier `tool_call_id` / assistant-tool-call branches, using a `(role, hash(content))` key. This is wrong because identical content at different positions in a conversation carries distinct semantic meaning. The fix removes the hash-based deduplication from the catch-all branch so that non-tool, non-assistant-tool-call messages are always preserved. Regression tests verify that repeated user confirmations, identical user messages, and duplicate system messages are all kept.

### Contribution Checklist

---


python/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py
```diff
@@ -242,14 +242,6 @@ def _deduplicate_messages(messages: list[Message]) -> list[Message]:
             unique_messages.append(msg)
```
moonbox3, line 242: Removing all deduplication for non-tool messages is too broad. The reported problem (legitimate repeated confirmations being dropped) is better addressed by switching to consecutive-duplicate detection rather than eliminating the check entirely. As written, any upstream bug that replays the full message history will now silently send duplicate messages to the LLM.

```suggestion
        else:
            content_str = str([str(c) for c in msg.contents]) if msg.contents else ""
            prev = unique_messages[-1] if unique_messages else None
            if prev is not None and get_role_value(prev) == role_value and (str([str(c) for c in prev.contents]) if prev.contents else "") == content_str:
                logger.info(f"Skipping consecutive duplicate message at index {idx}: role={role_value}")
                continue
            unique_messages.append(msg)
```

moonbox3 (commented):
## Automated Code Review

**Reviewers:** 4 | **Confidence:** 87%

### ✓ Correctness
> This diff removes content-based deduplication for general (non-tool-call, non-tool-result) messages, so identical user/system/assistant messages at different points in a conversation are preserved rather than collapsed. The change is logically correct: the `seen_keys` dictionary remains properly used by the two tool-related branches, and the else branch now simply appends. The updated and new tests correctly validate the new behavior. No correctness issues found.

### ✓ Security Reliability
> The diff removes hash-based deduplication of general (non-tool) messages, which previously could silently drop legitimate repeated user messages (e.g., repeated confirmations). The change improves reliability by preserving all general messages and removes use of Python's non-stable `hash()` on stringified content. No new security or reliability issues are introduced. Tests are updated to reflect the corrected behavior.

### ✓ Test Coverage
> The diff removes general-message deduplication logic and updates tests accordingly. The three tests covering the new behaviour (duplicate user messages preserved, interleaved confirmations preserved, repeated system messages preserved) are reasonable in scope, but all of them assert only on list length without verifying message content or ordering. For `test_deduplicate_preserves_repeated_confirmations` in particular, which mixes roles, a length-only check does not prove the messages are in the correct order or that the right messages survived. Additionally, there is no test confirming that messages with `contents=None` (empty content) still pass through the now-simplified else branch without error, which is a reachable edge case.

### ✗ Design Approach
> The diff removes deduplication logic for the 'else' branch of `_deduplicate_messages`, which previously deduplicated general messages (user, system, etc.) by content hash. The fix is justified if the original deduplication was causing legitimate messages to be dropped (e.g., repeated user confirmations). However, the approach is a complete removal rather than a targeted fix — it swings from 'always deduplicate same-content messages' to 'never deduplicate same-content messages'. The deeper question is: what is the correct deduplication scope? The tool-call branches deduplicate by structural identity (call_id, tool_call_ids), which is meaningful because duplicate tool results/calls are usually protocol artifacts. For regular messages, the same-content check was catching a different class of duplicates (e.g., the same message appended twice due to a bug upstream). Silently dropping that check means any upstream bug that sends duplicate messages will now silently propagate. A better approach would be to narrow the scope — e.g., only deduplicate consecutive identical messages, or only deduplicate when role is 'system', rather than removing the check entirely. Additionally, `seen_keys` is now populated only for tool and assistant-tool-call branches, but it is still declared and passed around unused for the else branch, which is a latent confusion.

### Flagged Issues
- [ ] The fix completely removes content-based deduplication for non-tool messages rather than narrowing it. Any upstream source that accidentally sends duplicate user/system messages (a real scenario during retries or reconnects) will silently propagate duplicates to the LLM, potentially causing unexpected behavior. A scoped fix—e.g., deduplicating only consecutive identical messages—would address the reported regression without losing all protection.

### Suggestions
- Consider replacing the blanket removal with consecutive-duplicate detection: only skip a message if the immediately preceding message has the same role and content hash. This preserves protection against the common upstream-resend artifact while allowing the same message to appear legitimately at different points in the conversation (e.g., repeated user confirmations after different assistant turns).
- The `seen_keys` dict is now unused for the else branch but is still declared and maintained for tool branches. Add a comment or restructure to make clear why the else branch intentionally does no deduplication, so future readers don't treat it as an oversight.
- In the new tests, assert on message identity and order (e.g., `assert result == [confirm1, assistant, confirm2]`) rather than just `len(result)`. A length-only assertion would still pass if the function returned the wrong messages or reordered them.
- Consider adding a test for messages with `contents=None` going through the simplified else branch, to verify no `AttributeError` or unexpected filtering occurs.
- Consider whether unbounded preservation of duplicate general messages could be a resource-exhaustion vector if message lists originate from untrusted input. A configurable upper bound on total message count may be warranted as a defense-in-depth measure.

---

python/packages/ag-ui/tests/ag_ui/test_message_adapters.py
```diff
@@ -1016,14 +1016,36 @@ def test_deduplicate_assistant_tool_calls():
...
+    result = _deduplicate_messages([confirm1, assistant, confirm2])
+    assert len(result) == 3
+
+
+def test_deduplicate_preserves_repeated_system_messages():
+    """Multiple identical system messages are preserved."""
+    from agent_framework_ag_ui._message_adapters import _deduplicate_messages
+
+    msgs = [Message(role="system", contents=[Content.from_text(text="You are a helpful assistant.")]) for _ in range(3)]
+
```
moonbox3, line 1071: Length-only assertions don't verify ordering or identity. Since this test exercises an interleaved conversation (confirm, assistant, confirm), assert the exact sequence to make the test more meaningful.

```suggestion
    assert result == [confirm1, assistant, confirm2]
```

python/packages/ag-ui/tests/ag_ui/test_message_adapters.py
```diff
@@ -1016,14 +1016,36 @@ def test_deduplicate_assistant_tool_calls():
...
-    assert len(result) == 1
+    assert len(result) == 2
+
+
+def test_deduplicate_preserves_repeated_confirmations():
+    """Identical user confirmations at different points in conversation are preserved."""
+    from agent_framework_ag_ui._message_adapters import _deduplicate_messages
+
+    assistant = Message(role="assistant", contents=[Content.from_text(text="Are you sure?")])
+    confirm1 = Message(role="user", contents=[Content.from_text(text="yes")])
```
moonbox3, line 1038: Consider also asserting content identity to guard against the function returning the wrong messages or reordering them.

```suggestion
    assert len(result) == 2
    assert result == [msg1, msg2]
```

markwallace-microsoft:
<a href="https://github.com/microsoft/agent-framework/blob/74678900fc1542b33a5e22101fa38de298cbe734/README.md">Python Test Coverage</a><details><summary>Python Test Coverage Report • </summary><table><tr><th>File</th><th>Stmts</th><th>Miss</th><th>Cover</th><th>Missing</th></tr><tbody><tr><td colspan="5"><b>packages/ag-ui/agent_framework_ag_ui</b></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/74678900fc1542b33a5e22101fa38de298cbe734/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py">_message_adapters.py</a></td><td>592</td><td>55</td><td>90%</td><td><a href="https://github.com/microsoft/agent-framework/blob/74678900fc1542b33a5e22101fa38de298cbe734/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py#L102-L103">102&ndash;103</a>, <a href="https://github.com/microsoft/agent-framework/blob/74678900fc1542b33a5e22101fa38de298cbe734/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py#L112-L115">112&ndash;115</a>, <a href="https://github.com/microsoft/agent-framework/blob/74678900fc1542b33a5e22101fa38de298cbe734/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py#L118-L122">118&ndash;122</a>, <a href="https://github.com/microsoft/agent-framework/blob/74678900fc1542b33a5e22101fa38de298cbe734/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py#L124-L129">124&ndash;129</a>, <a href="https://github.com/microsoft/agent-framework/blob/74678900fc1542b33a5e22101fa38de298cbe734/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py#L132">132</a>, <a href="https://github.com/microsoft/agent-framework/blob/74678900fc1542b33a5e22101fa38de298cbe734/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py#L141-L147">141&ndash;147</a>, <a href="https://github.com/microsoft/agent-framework/blob/74678900fc1542b33a5e22101fa38de298cbe734/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py#L187">187</a>, <a href="https://github.com/microsoft/agent-framework/blob/74678900fc1542b33a5e22101fa38de298cbe734/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py#L318">318</a>, <a href="https://github.com/microsoft/agent-framework/blob/74678900fc1542b33a5e22101fa38de298cbe734/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py#L408-L411">408&ndash;411</a>, <a href="https://github.com/microsoft/agent-framework/blob/74678900fc1542b33a5e22101fa38de298cbe734/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py#L413-L415">413&ndash;415</a>, <a href="https://github.com/microsoft/agent-framework/blob/74678900fc1542b33a5e22101fa38de298cbe734/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py#L462-L464">462&ndash;464</a>, <a href="https://github.com/microsoft/agent-framework/blob/74678900fc1542b33a5e22101fa38de298cbe734/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py#L518">518</a>, <a href="https://github.com/microsoft/agent-framework/blob/74678900fc1542b33a5e22101fa38de298cbe734/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py#L521">521</a>, <a href="https://github.com/microsoft/agent-framework/blob/74678900fc1542b33a5e22101fa38de298cbe734/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py#L523">523</a>, <a href="https://github.com/microsoft/agent-framework/blob/74678900fc1542b33a5e22101fa38de298cbe734/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py#L526">526</a>, <a href="https://github.com/microsoft/agent-framework/blob/74678900fc1542b33a5e22101fa38de298cbe734/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py#L529">529</a>, <a href="https://github.com/microsoft/agent-framework/blob/74678900fc1542b33a5e22101fa38de298cbe734/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py#L545">545</a>, <a href="https://github.com/microsoft/agent-framework/blob/74678900fc1542b33a5e22101fa38de298cbe734/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py#L562">562</a>, <a href="https://github.com/microsoft/agent-framework/blob/74678900fc1542b33a5e22101fa38de298cbe734/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py#L584">584</a>, <a href="https://github.com/microsoft/agent-framework/blob/74678900fc1542b33a5e22101fa38de298cbe734/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py#L680">680</a>, <a href="https://github.com/microsoft/agent-framework/blob/74678900fc1542b33a5e22101fa38de298cbe734/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py#L696-L697">696&ndash;697</a>, <a href="https://github.com/microsoft/agent-framework/blob/74678900fc1542b33a5e22101fa38de298cbe734/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py#L768">768</a>, <a href="https://github.com/microsoft/agent-framework/blob/74678900fc1542b33a5e22101fa38de298cbe734/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py#L790">790</a>, <a href="https://github.com/microsoft/agent-framework/blob/74678900fc1542b33a5e22101fa38de298cbe734/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py#L860">860</a>, <a href="https://github.com/microsoft/agent-framework/blob/74678900fc1542b33a5e22101fa38de298cbe734/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py#L895-L896">895&ndash;896</a>, <a href="https://github.com/microsoft/agent-framework/blob/74678900fc1542b33a5e22101fa38de298cbe734/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py#L964">964</a>, <a href="https://github.com/microsoft/agent-framework/blob/74678900fc1542b33a5e22101fa38de298cbe734/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py#L1007">1007</a></td></tr><tr><td><b>TOTAL</b></td><td><b>24000</b></td><td><b>2642</b></td><td><b>88%</b></td><td>&nbsp;</td></tr></tbody></table></details>

## Python Unit Test Overview
| 5230 | 20 :zzz: | 0 :x: | 0 :fire: | 1m 24s :stopwatch: |

moonbox3 (commented):
## Automated Code Review

**Reviewers:** 4 | **Confidence:** 96% | **Result:** All clear

Reviewed: Correctness, Security Reliability, Test Coverage, Design Approach

---

python/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py
moonbox3: Fixed in `83cfd01d`.

python/packages/ag-ui/tests/ag_ui/test_message_adapters.py
moonbox3: Fixed in `83cfd01d`.

python/packages/ag-ui/tests/ag_ui/test_message_adapters.py
moonbox3: Fixed in `83cfd01d`.

python/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py
```diff
@@ -242,14 +242,17 @@ def _deduplicate_messages(messages: list[Message]) -> list[Message]:
...
-            content_str = str([str(c) for c in msg.contents]) if msg.contents else ""
-            key = (role_value, hash(content_str))
-
-            if key in seen_keys:
-                logger.info(f"Skipping duplicate message at index {idx}: role={role_value}")
-                continue
+            # Use message_id for deduplication when available — two messages with the
+            # same id are definitively the same message (e.g. upstream replays), while
+            # different messages that happen to share identical content (e.g. repeated
+            # "yes" confirmations) will have distinct ids and be preserved.
```
moonbox3, line 248: If `message_id` can ever be an empty string, this check treats it as a valid dedup key, collapsing unrelated messages. A truthy check (`if msg.message_id:`) is more defensive. Also consider adding a test for the `""` edge case.

```suggestion
            if msg.message_id:
```

python/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py
```diff
@@ -242,14 +242,17 @@ def _deduplicate_messages(messages: list[Message]) -> list[Message]:
...
+            if msg.message_id is not None:
+                key = ("id", msg.message_id)
+                if key in seen_keys:
+                    logger.info(f"Skipping duplicate message at index {idx}: message_id={msg.message_id}")
+                    continue
+                seen_keys[key] = len(unique_messages)
```
moonbox3, line 259: Minor log-injection surface: if `message_id` contains newline or control characters from an untrusted source, it can split or corrupt log output. Use `!r` formatting.

```suggestion
                    logger.info(f"Skipping duplicate message at index {idx}: message_id={msg.message_id!r}")
```

python/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py
```diff
@@ -242,14 +242,17 @@ def _deduplicate_messages(messages: list[Message]) -> list[Message]:
...
 
-            seen_keys[key] = len(unique_messages)
             unique_messages.append(msg)
 
     return unique_messages
```
moonbox3, line 263: When `msg.message_id` is None the code skips deduplication entirely, creating false negatives for genuine duplicates from integrations that don't set message_id. Prefer a two-tier strategy: use the ID key when present, fall back to content-hash key otherwise.

```suggestion
            if msg.message_id is not None:
                key = ("id", msg.message_id)
            else:
                content_str = str([str(c) for c in msg.contents]) if msg.contents else ""
                key = ("content", role_value, hash(content_str))

            if key in seen_keys:
                logger.info(f"Skipping duplicate message at index {idx}: role={role_value}")
                continue

            seen_keys[key] = len(unique_messages)
            unique_messages.append(msg)
```

moonbox3 (commented):
## Automated Code Review

**Reviewers:** 4 | **Confidence:** 85%

### ✓ Correctness
> The change replaces content-hash-based deduplication with message_id-based deduplication in the else branch. The logic is correct: messages with a message_id are deduplicated by that id, and messages without a message_id are always preserved (no false-positive dedup of repeated identical content like 'yes' confirmations). The key tuple ('id', msg.message_id) avoids collisions with keys from the tool-result and assistant-tool-call branches (which use role_value as the first element). Tests comprehensively cover the new behavior. No correctness bugs found.

### ✓ Security Reliability
> The change replaces content-hash deduplication with message_id-based deduplication. This is a sound design improvement that avoids false-positive dedup of intentionally repeated messages. However, it introduces a reliability gap: messages without a message_id are now never deduplicated, meaning a misbehaving or replaying client that omits message_ids can cause unbounded message accumulation. The logging of raw message_id values also presents a minor log-injection surface. No blocking security issues found.

### ✓ Test Coverage
> Test coverage for the new message_id-based deduplication logic is thorough. The tests cover the main branches well: dedup by same message_id, preservation with distinct ids, no-id passthrough, None contents, and mixed id/no-id scenarios. One minor gap: there is no test verifying behavior when `message_id` is an empty string (which is not None and would be used as a dedup key, potentially causing unintended collisions).

### ✗ Design Approach
> The change replaces content-hash-based deduplication with ID-based deduplication for general messages. The motivation is correct — content hashing caused false positives (deduplicating legitimate repeated messages like "yes" confirmations). However, the fix goes too far in the other direction: when `message_id` is absent (None), the new code performs *no* deduplication at all, silently abandoning the protection the function was designed to provide. The test `test_deduplicate_without_message_id_always_preserves` explicitly enshrines this as intended behavior, but it means the function's deduplication guarantee is now conditional on callers setting `message_id`. A better approach is to use ID-based deduplication when IDs are present and fall back to content-based deduplication when they are not, preserving best-effort protection without regressing on the no-ID case.

### Flagged Issues
- [ ] When message_id is None, all deduplication is silently disabled. Any call path that does not populate message_id (e.g., older integrations, third-party adapters) will accumulate genuine duplicates the old code correctly suppressed. Prefer a two-tier strategy: use ID-based deduplication when message_id is present, and fall back to content-based deduplication when it is absent.

### Suggestions
- Guard against empty-string message_id values: `msg.message_id is not None` treats `message_id=""` as a valid dedup key, collapsing all such messages into one. A truthy check (`if msg.message_id:`) would be safer. Add a corresponding test for the empty-string edge case to document expected behavior.
- Sanitise message_id before logging (line 255). If message_id originates from an untrusted source and contains newlines or control characters, it can split or corrupt log output. Use `repr(msg.message_id)` or strip control characters.
- Consider logging a debug-level warning when a message lacks a message_id, so operators can detect integrations that silently lose deduplication and risk unbounded context growth.

---

python/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py
```diff
@@ -242,14 +242,17 @@ def _deduplicate_messages(messages: list[Message]) -> list[Message]:
...
+            # Use message_id for deduplication when available — two messages with the
+            # same id are definitively the same message (e.g. upstream replays), while
+            # different messages that happen to share identical content (e.g. repeated
+            # "yes" confirmations) will have distinct ids and be preserved.
+            if msg.message_id is not None:
+                key = ("id", msg.message_id)
+                if key in seen_keys:
+                    logger.info(f"Skipping duplicate message at index {idx}: message_id={msg.message_id}")
+                    continue
+                seen_keys[key] = len(unique_messages)
```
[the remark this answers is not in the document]
moonbox3: Fixed in `74678900`.

python/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py
moonbox3: Fixed in `74678900`.

python/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py
moonbox3: Fixed in `74678900`.

python/packages/ag-ui/agent_framework_ag_ui/_message_adapters.py
moonbox3: Fixed in `74678900`.

SogoKato:
Thanks for adding the fallback! One concern: using a global seen_keys for content hashes might accidentally remove intentional repeats (e.g., a user saying "Yes" or "Confirm" multiple times in one session).

What if we limit the fallback check to consecutive messages only? This catches UI glitches while preserving legitimate repeats later in the chat:

```py
last_key = None

for msg in messages:
    # ... (existing logic) ...

    elif msg.message_id:
        # ... (message_id logic)

    # Fallback: check only if it matches the immediately preceding message
    else:
        key = (role_value, hash(content_str))
        if key == last_key:
            continue

        last_key = key
        unique_messages.append(msg)
```

This seems safer for natural conversation flow. What do you think?

eavanvalkenburg (approved):

giles17 (approved):

PR #4731, opened 2026-03-16
Add ADR-0020: Foundry Evals integration

alliscode:
Captures the design for integrating Azure AI Foundry Evaluations with agent-framework. Key decisions:

- EvalItem with conversation (list[Message]) as single source of truth
- query/response derived from configurable conversation split strategies
- Tools as list[FunctionTool] (including auto-extracted MCP tools)
- FoundryEvals provider with auto-detection of evaluator capabilities
- LocalEvaluator with @function_evaluator decorator for local checks
- Consistent Python/C# APIs: evaluate_agent, evaluate_workflow

### Contribution Checklist

### DIFF
--- a/docs/decisions/0020-foundry-evals-integration.md
+++ b/docs/decisions/0020-foundry-evals-integration.md
@@ -1,5 +1,5 @@
 ---
-status: proposed
+status: accepted
 contact: bentho
 date: 2026-02-27
 deciders: bentho, markwallace-microsoft, westey-m


docs/decisions/0020-foundry-evals-integration.md
```diff
@@ -0,0 +1,732 @@
...
+However, using Foundry Evals with an agent-framework agent today requires significant manual effort. Developers must:
+
+1. Transform agent-framework's `Message`/`Content` types into the OpenAI-style agent message schema that Foundry evaluators expect
+2. Map tool definitions from agent-framework's `FunctionTool` format to evaluator-compatible schemas
+3. Manually wire up the correct Foundry data source type (`azure_ai_traces`, `jsonl`, `azure_ai_target_completions`, etc.) depending on their scenario
+4. Handle App Insights trace ID queries, response ID collection, and eval polling
+
+Additionally, evaluation is a concern that extends beyond any single provider. Developers may want to use local evaluators (LLM-as-judge, regex, keyword matching), third-party evaluation libraries, or multiple providers in combination. The architecture must support this without creating a Foundry-specific lock-in at the API level.
+
+### What Developers Need from Agent Evaluation
```
moonbox3: This section seems really long for an ADR? Are the "decision drivers" capturing these concerns already?

docs/decisions/0020-foundry-evals-integration.md
```diff
@@ -0,0 +1,732 @@
...
+    evaluators=evals,
+)
+for r in results:
+    r.assert_passed()
+```
+
+**C#:**
+
+```csharp
+var evals = new FoundryEvaluator(chatConfiguration, FoundryEvals.Relevance, FoundryEvals.Coherence);
```
moonbox3: In the C# examples, evaluator constants are referenced as `FoundryEvals.Relevance`, but the C# evaluator class is `FoundryEvaluator`. Are the constants on `FoundryEvaluator` itself, or is there a separate `FoundryEvals` static class in .NET? If separate, what's the rationale for the different naming vs Python where they live on the same class?

docs/decisions/0020-foundry-evals-integration.md
```diff
@@ -0,0 +1,732 @@
...
+
+    query: str       # property — derived from conversation split
+    response: str    # property — derived from conversation split
+```
+
+`conversation` is the single source of truth. `query` and `response` are derived properties — splitting the conversation at the last user message (default) and extracting text from each side. Changing the `split_strategy` consistently changes all derived values.
+
+`tools` provides typed `FunctionTool` objects — including MCP tools, which are automatically extracted after agent runs.
+
+`EvalItem.to_dict()` serializes to flat dict: splits conversation, converts each half to OpenAI chat format (`query_messages`/`response_messages`), derives `query`/`response` strings, and serializes `tools` as `tool_definitions`.
```
moonbox3: Does this implementation detail need to be included? Wouldn't this be internal to the Foundry provider path?

docs/decisions/0020-foundry-evals-integration.md
```diff
@@ -0,0 +1,732 @@
...
+    public static EvalCheck Create(string name, Func<EvalItem, bool> check);         // full item
+    public static EvalCheck Create(string name, Func<EvalItem, CheckResult> check);  // full control
+    public static EvalCheck Create(string name, Func<string, Task<bool>> check);     // async
+}
+```
+
+`EvalItem` is a lightweight record used only by `FunctionEvaluator` and `LocalEvaluator` to pass context to check functions. It is not part of the `IEvaluator` interface:
+
+```csharp
+public record EvalItem(
```
moonbox3: The Python `EvalItem` has `response_id` and `split_strategy`, which are missing from the .NET version?

docs/decisions/0020-foundry-evals-integration.md
```diff
@@ -0,0 +1,732 @@
...
+| `EvalResults` | `AgentEvaluationResults` |
+| `EvalItemResult` / `EvalScoreResult` | MEAI `EvaluationResult` / `EvaluationMetric` (reused) |
+| `LocalEvaluator` | `LocalEvaluator` (implements `IEvaluator`) |
+| `@function_evaluator` | `FunctionEvaluator.Create()` overloads |
+| `keyword_check()` / `tool_called_check()` | `EvalChecks.KeywordCheck()` / `EvalChecks.ToolCalledCheck()` |
+| `FoundryEvals` | `FoundryEvaluator` (implements `IEvaluator`) |
+| `evaluate_agent()` | `agent.EvaluateAsync(queries, evaluator)` extension method |
+| `evaluate_agent(responses=)` | `agent.EvaluateAsync(responses, evaluator)` extension method |
+| `evaluate_workflow()` | `run.EvaluateAsync()` extension method |
+| `AgentEvalConverter` | Internal to extension methods |
```
moonbox3: Can we trim this table at all? Looks like calling out `AgentEvalConverter` shows an absence of mapping?

docs/decisions/0020-foundry-evals-integration.md
```diff
@@ -0,0 +1,732 @@
...
+By default, evaluators see only the last turn (final user message → final assistant response). For multi-turn conversations, you can control how the conversation is factored for evaluation:
+
+**Python:**
+
+```python
+results = await evaluate_agent(
+    agent=agent,
+    queries=["Plan a 3-day trip to Paris"],
+    evaluators=evals,
+    conversation_split=ConversationSplit.FULL,      # evaluate entire trajectory
```
eavanvalkenburg, line 186: are these universal across different evals?

docs/decisions/0020-foundry-evals-integration.md
```diff
@@ -0,0 +1,732 @@
...
+    "planner":  EvalResults(passed=1, total=1)
+    "researcher": EvalResults(passed=1, total=1)
+```
+
+#### Mix multiple providers
+
+**Python:**
+
+```python
+@function_evaluator
```
eavanvalkenburg:
```suggestion
@evaluator
```

docs/decisions/0020-foundry-evals-integration.md
```diff
@@ -0,0 +1,732 @@
...
+
+Multiple evaluators return one result each — `results[0]` is the local evaluator, `results[1]` is Foundry.
+
+#### Custom function evaluators
+
+**Python:**
+
+```python
+@function_evaluator
+def mentions_city(response: str, expected: str) -> bool:
```
eavanvalkenburg: what is the expected signature for functions evaluators? because I see two different ones here?

docs/decisions/0020-foundry-evals-integration.md
```diff
@@ -0,0 +1,732 @@
...
+
+### Core: Evaluator Protocol
+
+A runtime-checkable protocol that any evaluation provider implements:
+
+```python
+@runtime_checkable
+class Evaluator(Protocol):
+    @property
+    def name(self) -> str: ...
```
eavanvalkenburg:
```suggestion
    name: str
```

docs/decisions/0020-foundry-evals-integration.md
eavanvalkenburg: should we use something as generic as `to_dict` for something that makes a very specific version of itself?

docs/decisions/0020-foundry-evals-integration.md
alliscode: The naming should be the same. I'll fix that.

docs/decisions/0020-foundry-evals-integration.md
```diff
@@ -0,0 +1,732 @@
...
+    return len(response.split()) < 500
+
+# Tier 2: Ground truth — compare against expected output
+@function_evaluator
+def mentions_city(response: str, expected: str) -> bool:
+    return expected.lower() in response.lower()
+
+# Tier 3: Full context — inspect conversation and tools
+@function_evaluator
+def used_tools(conversation: list, tools: list) -> float:
```
eavanvalkenburg, line 506: could we demonstrate how `GAIA` would fit into this? It has a lot of knobs, which is probably useful, so would be good to understand if we can accommodate all of those settings with the proposed abstraction.

docs/decisions/0020-foundry-evals-integration.md
alliscode: Fixed. Dotnet has split strategy, but it was missing from the ADR. I removed response_id as that can be internal details of providers.

docs/decisions/0020-foundry-evals-integration.md
alliscode: Yes. The basic eval protocol is a (query, response) -> BOOL | float. The split strategies are at the MAF layer and do the splitting before the evaluators run.

docs/decisions/0020-foundry-evals-integration.md
alliscode: Agree, changed to `to_eval_dict`. I've also removed this detail from the ADR as it's internal implementation as @moonbox3 suggested.

docs/decisions/0020-foundry-evals-integration.md
alliscode: Done

docs/decisions/0020-foundry-evals-integration.md
alliscode: Added more info about the signature below:

`@evaluator` uses **parameter name injection** — the function's parameter names determine what data it receives from the `EvalItem`. Supported names: `query`, `response`, `expected`, `conversation`, `tools`, `context`. Any combination is valid.

docs/decisions/0020-foundry-evals-integration.md
```diff
@@ -0,0 +1,753 @@
...
+- **Lowest concept count**: Introduce the fewest possible new types, abstractions, and APIs for developers to learn.
+- **Leverage existing knowledge**: The framework already knows which agents exist, what tools they have, and what conversations occurred. Evals should use this automatically rather than requiring the developer to re-specify it.
+- **Foundry-native results**: When using Foundry, results should be viewable in the Foundry portal with dashboards and comparison views.
+- **Progressive disclosure**: Simple scenarios should be near-zero code. Advanced scenarios should build on the same primitives.
+- **Cross-language parity**: Design must be implementable in both Python and .NET.
+
+## Considered Options
+
+1. **Provider-specific functions** — Build Foundry-specific helper functions (`evaluate_agent()`, etc.) directly in the Azure package. All eval functions take Foundry connection parameters.
+2. **Evaluator protocol with core orchestration** — Define a provider-agnostic `Evaluator` protocol in core. Orchestration functions live in core. Providers implement the protocol.
```
TaoChenOSU: Is this Python only? My first impression is this will live in the core package but .Net doesn't have one. This will probably benefit from more details on `core`.

docs/decisions/0020-foundry-evals-integration.md
```diff
@@ -0,0 +1,753 @@
...
+
+results.AssertAllPassed();
+```
+
+`evaluate_agent` returns one `EvalResults` per evaluator. Each result contains per-item scores:
+
+```
+# results[0] (FoundryEvals)
+EvalResults(status="completed", passed=1, failed=0, total=1)
+  items[0]: EvalItemResult(query="What's the weather?", scores={"relevance": 5, "coherence": 5})
```
TaoChenOSU: Does an item include the agent response and tool usage for auditing purposes?

docs/decisions/0020-foundry-evals-integration.md
```diff
@@ -0,0 +1,753 @@
...
+
+**Python:**
+
+```python
+query = "What's the weather?"
+response = await agent.run([Message("user", [query])])
+
+results = await evaluate_agent(
+    agent=agent,
+    responses=response,
```
TaoChenOSU: Should this also be a list of responses too?

docs/decisions/0020-foundry-evals-integration.md
```diff
@@ -0,0 +1,753 @@
...
+
+```python
+evals = FoundryEvals(
+    project_client=client,
+    model_deployment="gpt-4o",
+    evaluators=[FoundryEvals.RELEVANCE, FoundryEvals.COHERENCE],
+)
+
+results = await evaluate_agent(
+    agent=my_agent,
```
TaoChenOSU, line 71: Does this invoke the agent to get responses for evaluation? If so, how many time will it invoke the agent?

docs/decisions/0020-foundry-evals-integration.md
```diff
@@ -0,0 +1,753 @@
...
+
+foreach (var r in evalResults)
+{
+    Console.WriteLine($"  overall: {r.Passed}/{r.Total}");
+    foreach (var (name, sub) in r.SubResults)
+        Console.WriteLine($"    {name}: {sub.Passed}/{sub.Total}");
+}
+```
+
+Workflows return one result per evaluator, with sub-results per agent in the workflow:
```
TaoChenOSU, line 254: What happens if the workflow doesn't have agents?

docs/decisions/0020-foundry-evals-integration.md
```diff
@@ -0,0 +1,753 @@
...
+```
+
+Categories: Agent behavior, Tool usage, Quality, Safety.
+
+### Azure AI: Foundry-Specific Functions
+
+| Function | What it does |
+|---|---|
+| `evaluate_traces()` | Evaluate from stored response IDs or OTel traces |
+| `evaluate_foundry_target()` | Evaluate a Foundry-registered agent or deployment |
```
TaoChenOSU, line 477: Are these prompt agents and hosted agents?

docs/decisions/0020-foundry-evals-integration.md
alliscode: It will skip non-agent executors in the workflow.

docs/decisions/0020-foundry-evals-integration.md
alliscode: Various options can be used to support prompt agents, hosted agents, and responses agents e.g. trace based can be used for anything, Response based can be used for anything using Foundry Responses, etc.

docs/decisions/0020-foundry-evals-integration.md
alliscode: Yeah, the ADR is for both python and dotnet but this section uses mostly python terminology, I'll clean it up.

docs/decisions/0020-foundry-evals-integration.md
alliscode: Yes it does if you provide queries. It will invoke the agent once per query.

TaoChenOSU: It sounds less convincing that it only invokes once. If it only invokes it once, the results can hardly be a performance indicator for the agent. Would it make more sense to say the agent is invoked 10 times and the results indicate that the agent perform well 9 (just a random number) times?

docs/decisions/0020-foundry-evals-integration.md
alliscode: Currently it does not. Do you think it needs to?

docs/decisions/0020-foundry-evals-integration.md
alliscode: Discussed offline, the confusion is from the badly formatted sample, will fix.

docs/decisions/0020-foundry-evals-integration.md
```diff
@@ -0,0 +1,766 @@
...
+
+However, using Foundry Evals with an agent-framework agent today requires significant manual effort. Developers must:
+
+1. Transform agent-framework's `Message`/`Content` types into the OpenAI-style agent message schema that Foundry evaluators expect
+2. Map tool definitions from agent-framework's `FunctionTool` format to evaluator-compatible schemas
+3. Manually wire up the correct Foundry data source type (`azure_ai_traces`, `jsonl`, `azure_ai_target_completions`, etc.) depending on their scenario
+4. Handle App Insights trace ID queries, response ID collection, and eval polling
+
+Additionally, evaluation is a concern that extends beyond any single provider. Developers may want to use local evaluators (LLM-as-judge, regex, keyword matching), third-party evaluation libraries, or multiple providers in combination. The architecture must support this without creating a Foundry-specific lock-in at the API level.
```
crickman, line 23: Is there a list of third-party libraries to support?

docs/decisions/0020-foundry-evals-integration.md
```diff
@@ -0,0 +1,766 @@
...
+    public static EvalCheck Create(string name, Func<string, string?, bool> check);  // expectedOutput
+    public static EvalCheck Create(string name, Func<EvalItem, bool> check);         // full item
+    public static EvalCheck Create(string name, Func<EvalItem, CheckResult> check);  // full control
+    public static EvalCheck Create(string name, Func<string, Task<bool>> check);     // async
+}
+```
+
+`EvalItem` is a lightweight record used only by `FunctionEvaluator` and `LocalEvaluator` to pass context to check functions. It is not part of the `IEvaluator` interface:
+
+```csharp
```
crickman, line 760: are multi-modal responses to be supported for evalution?

docs/decisions/0020-foundry-evals-integration.md
alliscode: Updated for clarity

docs/decisions/0020-foundry-evals-integration.md
TaoChenOSU: I feel like there should just for completeness.

docs/decisions/0020-foundry-evals-integration.md
alliscode: Added context around supporting 3rd parties as well as examples.

docs/decisions/0020-foundry-evals-integration.md
alliscode: The are supported as the evaluator gets access to the full conversation. I haven't shown any example of multi-modal evaluators though and we likely won't ship any in the first round.

docs/decisions/0020-foundry-evals-integration.md
alliscode: I misspoke, it does include the response, but it does not include the tools at this time. I think we should add that as we see the need for it.

docs/decisions/0020-foundry-evals-integration.md
alliscode: I've added an optional `num_repetitions` parameter that allow for this.

TaoChenOSU (approved):

Issue #3705, opened 2026-02-05
Python: [Bug]: Group Chat orchestrator message cleanup issue


leelakarthik:
### Description

When using groupchatbuilder, messages with empty content are not being cleaned like in the _handoff.py 'clean_conversation_for_handoff'

This breaks the conversation flow 
Error message
{'type': 'error', 'error': {'type': 'invalid_request_error', 'message': 'messages.11: all messages must have non-empty content except for the optional final assistant message'}, 'request_id': 'req_011CXq5rCQUPFti495Nrvn6f'}

### Code Sample

```markdown

```

### Error Messages / Stack Traces

```markdown
Traceback (most recent call last):
  File "/Users/karthik/Code bases/Jan/MAF/main.py", line 221, in _interactive_mode
    response = await workflow.run(query)
               ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/karthik/Code bases/Jan/MAF/codebase_orchestrator_agent/agent/codebase_orchestrator.py", line 203, in run
    async for event in self._run_with_events(query):
  File "/Users/karthik/Code bases/Jan/MAF/codebase_orchestrator_agent/agent/codebase_orchestrator.py", line 296, in _run_with_events
    events = await self.workflow.run(clean_query)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/karthik/Code bases/Jan/MAF/.venv/lib/python3.11/site-packages/agent_framework/_workflows/_workflow.py", line 660, in run
    raw_events = [
                 ^
  File "/Users/karthik/Code bases/Jan/MAF/.venv/lib/python3.11/site-packages/agent_framework/_workflows/_workflow.py", line 660, in <listcomp>
    raw_events = [
                 ^
  File "/Users/karthik/Code bases/Jan/MAF/.venv/lib/python3.11/site-packages/agent_framework/_workflows/_workflow.py", line 346, in _run_workflow_with_tracing
    async for event in self._runner.run_until_convergence():
  File "/Users/karthik/Code bases/Jan/MAF/.venv/lib/python3.11/site-packages/agent_framework/_workflows/_runner.py", line 120, in run_until_convergence
    await iteration_task
  File "/Users/karthik/Code bases/Jan/MAF/.venv/lib/python3.11/site-packages/agent_framework/_workflows/_runner.py", line 189, in _run_iteration
    await asyncio.gather(*tasks)
[...]

PR #4736, opened 2026-03-17
Python: Fix group chat broadcast to include next speaker when it re-speaks

yashy797:
### Motivation and Context

When the orchestrator selects the same participant to speak in consecutive turns, the broadcast step after the first response skips that participant (because it was the one who just responded). This means the participant's AgentExecutor cache is not repopulated with its own response before the follow-up request arrives, causing the agent to miss context or potentially triggering errors from empty message payloads.

Related to #3705 — that issue addressed empty message content in the same `_handle_response` methods by adding `clean_conversation_for_handoff`. This fix addresses a separate broadcast filtering bug in the same code path.

### Description

Fixes the message broadcasting filter in both `GroupChatOrchestrator._handle_response` and `AgentBasedGroupChatOrchestrator._handle_response` so that when the next speaker is the same participant that just responded, it still receives the broadcast.

**Changes in `_group_chat.py`:**

1. **`GroupChatOrchestrator._handle_response`** — the participant filter now uses `p != participant or p == next_speaker` instead of `p != participant`, ensuring the next speaker is always included in the broadcast even if it just responded.

2. **`AgentBasedGroupChatOrchestrator._handle_response`** — same filtering logic applied. Additionally, `next_speaker` is extracted into a local variable for clarity and reused in both the broadcast filter and the subsequent `_send_request_to_participant` call.

Both orchestrators now share the comment explaining the rationale:

> "…except the participant that just responded (unless it is also the next speaker, in which case it must receive the broadcast so its executor cache is repopulated before the follow-up request arrives)."

### Contribution Checklist

- [x] **Is this a breaking change?** No — this is a bug fix with no public API changes.

### DIFF
--- a/python/packages/orchestrations/agent_framework_orchestrations/_group_chat.py
+++ b/python/packages/orchestrations/agent_framework_orchestrations/_group_chat.py
@@ -206,12 +206,16 @@ class GroupChatOrchestrator(BaseGroupChatOrchestrator):
         next_speaker = await self._get_next_speaker()
 
         # Broadcast participant messages to all participants for context, except
-        # the participant that just responded
+        # the participant that just responded (unless it is also the next speaker,
+        # in which case it must receive the broadcast so its executor cache is
+        # repopulated before the follow-up request arrives).
         participant = ctx.get_source_executor_id()
         await self._broadcast_messages_to_participants(
             messages,
             cast(WorkflowContext[AgentExecutorRequest | GroupChatParticipantMessage], ctx),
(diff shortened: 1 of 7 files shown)


TaoChenOSU:
Hi @yashy797,

Thank you for your contribution!

The agent that just responded will have the response in its context, thus we don't need to prepopulate the cache with its own response. Are you an issue where the agent is missing its own response? Could you provide a reproduction?

eavanvalkenburg:

PR #4741, opened 2026-03-17
Python: Normalize OpenAI tool-call argument envelopes on parse

### LINKED ISSUE #4740: Python: Normalize provider tool-call argument envelopes across chat backends
## Problem
Equivalent tool calls do not produce the same internal argument envelope across providers. Anthropic and Ollama surface structured arguments as mappings, while OpenAI/Azure Responses surface the same arguments as raw JSON strings, which leaks provider differences into workflow and middleware logic.

## Why now
The framework explicitly promises provider choice. Tool-call handling should not require callers to branch on provider-specific argument shapes for the same interaction.

## Version under test
- Commit: `cdb51e6a41a2ea6f353360a353943da8bcb18eb3` (`origin/main` on 2026-03-17)
- Runtime: Python `3.14.0`, uv `0.8.19`, macOS `darwin/arm64`

## Minimal repro
1. Produce one synthetic tool call with identical JSON arguments through OpenAI Responses, Azure Responses, Anthropic, and Ollama clients.
2. Inspect the emitted `Content(type="function_call")` items.

## Expected behavior
The same tool-call interaction should expose one normalized internal argument representation across supported providers.

## Actual behavior
OpenAI/Azure preserve raw JSON strings in `arguments`, while Anthropic/Ollama already return structured mappings for the same semantic payload.

## Code paths
- `python/packages/core/agent_framework/openai/_responses_client.py`
- `python/packages/core/agent_framework/azure/_responses_client.py`
- `python/packages/core/agent_framework/anthropic/__init__.py`
- `python/packages/core/agent_framework/ollama/__init__.py`

## Validation target
A cross-provider regression test should prove that one synthetic tool-call interaction yields the same internal argument envelope across the supported backends covered by the framework.

davidahmann:
Closes #4740

## Problem
OpenAI chat and responses parsers return function-call arguments as raw JSON strings, while the other backends already normalize equivalent tool-call payloads to mappings. That backend-dependent envelope shape leaks into downstream handling.

## What changed
- add a helper that normalizes provider function-call arguments to mappings when the payload is a JSON object
- use that helper in the OpenAI chat and responses parse paths
- add/update focused tests covering the normalized parse behavior

## Validation
- `uv run pytest packages/core/tests/openai/test_openai_chat_client.py::test_parse_tool_calls_from_openai_normalizes_json_object_arguments packages/core/tests/openai/test_openai_responses_client.py::test_response_content_creation_with_function_call packages/core/tests/openai/test_openai_responses_client.py::test_parse_response_from_openai_function_call_includes_status -m "not integration"`

### DIFF
--- a/python/packages/core/agent_framework/_types.py
+++ b/python/packages/core/agent_framework/_types.py
@@ -70,6 +70,21 @@ def _parse_content_list(contents_data: Sequence[Any]) -> list[Content]:
     return contents
 
 
+def normalize_function_call_arguments(
+    arguments: str | Mapping[str, Any] | None,
+) -> str | dict[str, Any] | None:
+    """Normalize provider tool-call arguments to a mapping when possible."""
+    if not isinstance(arguments, str):
+        return dict(arguments) if isinstance(arguments, Mapping) else arguments
+
+    try:
+        loaded = json.loads(arguments)
+    except (json.JSONDecodeError, TypeError):
+        return arguments
+
+    return loaded if isinstance(loaded, dict) else arguments
+
+
 # region Internal Helper functions for unified Content
 
 
--- a/python/packages/core/agent_framework/openai/_chat_client.py
+++ b/python/packages/core/agent_framework/openai/_chat_client.py
@@ -49,6 +49,7 @@ from .._types import (
     Message,
     ResponseStream,
     UsageDetails,
+    normalize_function_call_arguments,
 )
 from ..exceptions import (
     ChatClientException,
@@ -556,7 +557,7 @@ class RawOpenAIChatClient(  # type: ignore[misc]
                     fcc = Content.from_function_call(
                         call_id=tool.id if tool.id else "",
                         name=tool.function.name if tool.function.name else "",
-                        arguments=tool.function.arguments if tool.function.arguments else "",
+                        arguments=normalize_function_call_arguments(tool.function.arguments if tool.function.arguments else ""),
                         raw_representation=tool.function,
                     )
                     resp.append(fcc)
--- a/python/packages/core/tests/openai/test_openai_chat_client.py
+++ b/python/packages/core/tests/openai/test_openai_chat_client.py
@@ -862,6 +862,32 @@ def test_function_approval_content_is_skipped_in_preparation(
     assert prepared_mixed[0]["content"] == "I need approval for this action."
 
 
+def test_parse_tool_calls_from_openai_normalizes_json_object_arguments(
+    openai_unit_test_env: dict[str, str],
+) -> None:
+    client = OpenAIChatClient()
+
+    mock_tool_function = MagicMock()
+    mock_tool_function.name = "get_weather"
+    mock_tool_function.arguments = '{"city": "Seattle"}'
+
+    mock_tool = MagicMock()
+    mock_tool.id = "call_123"
+    mock_tool.function = mock_tool_function
+
+    mock_message = MagicMock()
+    mock_message.tool_calls = [mock_tool]
+
+    mock_choice = MagicMock()
+    mock_choice.delta = mock_message
+
+    contents = client._parse_tool_calls_from_openai(mock_choice)
+
+    assert len(contents) == 1
+    assert contents[0].type == "function_call"
+    assert contents[0].arguments == {"city": "Seattle"}
+
+
 def test_usage_content_in_streaming_response(
     openai_unit_test_env: dict[str, str],
 ) -> None:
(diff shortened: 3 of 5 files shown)


davidahmann:
This removes an OpenAI-only envelope mismatch where tool-call arguments arrived as raw JSON strings instead of the mapping shape other backends already emit.

The change stays on the parse boundary: OpenAI chat and responses payloads now normalize JSON-object arguments immediately, and the targeted tests were updated to lock that in.

Validation:
- `uv run pytest packages/core/tests/openai/test_openai_chat_client.py::test_parse_tool_calls_from_openai_normalizes_json_object_arguments packages/core/tests/openai/test_openai_responses_client.py::test_response_content_creation_with_function_call packages/core/tests/openai/test_openai_responses_client.py::test_parse_response_from_openai_function_call_includes_status -m "not integration"` (pass)

python/packages/core/agent_framework/openai/_chat_client.py
```diff
@@ -556,7 +557,7 @@ def _parse_tool_calls_from_openai(self, choice: Choice | ChunkChoice) -> list[Co
                     fcc = Content.from_function_call(
                         call_id=tool.id if tool.id else "",
                         name=tool.function.name if tool.function.name else "",
-                        arguments=tool.function.arguments if tool.function.arguments else "",
+                        arguments=normalize_function_call_arguments(tool.function.arguments if tool.function.arguments else ""),
```
moonbox3, line 560: This method is called for both non-streaming `Choice` and streaming `ChunkChoice`. Normalizing here means streaming chunks with partial JSON stay as strings (fine), but if a chunk contains complete valid JSON it becomes a `dict`, and a subsequent string chunk causes `TypeError('Incompatible argument types')` in `_add_function_call_content`. Consider gating normalization on the non-streaming path (`isinstance(choice, Choice)`) or applying it after accumulation.

python/packages/core/agent_framework/_types.py
```diff
@@ -70,6 +70,21 @@ def _parse_content_list(contents_data: Sequence[Any]) -> list[Content]:
...
 
 
+def normalize_function_call_arguments(
+    arguments: str | Mapping[str, Any] | None,
+) -> str | dict[str, Any] | None:
+    """Normalize provider tool-call arguments to a mapping when possible."""
+    if not isinstance(arguments, str):
+        return dict(arguments) if isinstance(arguments, Mapping) else arguments
+
+    try:
```
moonbox3, line 80: The normalization belongs inside `Content.from_function_call()` rather than as a standalone helper. All current and future call sites automatically benefit, and the existing `parse_arguments()` method's JSON-parsing logic is no longer duplicated. Applying it only at selected call sites leads to a `Content` whose `arguments` type depends on which parsing path produced it.

python/packages/core/agent_framework/_types.py
```diff
@@ -70,6 +70,21 @@ def _parse_content_list(contents_data: Sequence[Any]) -> list[Content]:
...
+        loaded = json.loads(arguments)
+    except (json.JSONDecodeError, TypeError):
+        return arguments
+
+    return loaded if isinstance(loaded, dict) else arguments
+
+
 # region Internal Helper functions for unified Content
```
moonbox3, line 90: This utility function has five distinct code paths (None, non-str Mapping, non-str non-Mapping, valid JSON dict string, and non-dict-JSON/invalid-JSON string) but no direct unit tests. Add a parametrized test covering at least: `None` → `None`, `'{"a":1}'` → `{"a":1}`, `'not json'` → `'not json'`, `'[1,2]'` → `'[1,2]'`, `{"a":1}` (dict) → `{"a":1}`, and an `OrderedDict` → plain dict.

python/packages/core/tests/openai/test_openai_chat_client.py
```diff
@@ -862,6 +862,32 @@ def test_function_approval_content_is_skipped_in_preparation(
...
+    mock_choice.delta = mock_message
+
+    contents = client._parse_tool_calls_from_openai(mock_choice)
+
+    assert len(contents) == 1
+    assert contents[0].type == "function_call"
+    assert contents[0].arguments == {"city": "Seattle"}
+
+
 def test_usage_content_in_streaming_response(
```
moonbox3, line 891: This test only covers the JSON-object-string → dict path. Consider adding cases for non-JSON arguments (e.g., `arguments = 'plain text'` should pass through as-is) and empty string (`arguments = ''`) to verify the fallback behavior.

python/packages/core/tests/openai/test_openai_responses_client.py
```diff
@@ -1037,7 +1037,7 @@ def test_response_content_creation_with_function_call() -> None:
     function_call = response.messages[0].contents[0]
```
moonbox3, line 1037: This is the only responses-client test covering function_call argument normalization in the non-streaming path. Consider adding a case where `arguments` is not valid JSON (e.g., partial/truncated) to confirm passthrough behavior.

PR #4743, opened 2026-03-17
Python: Fix ENABLE_SENSITIVE_DATA env var ignored when set after module import

### LINKED ISSUE #4119: Python: [Bug]: ENABLE_SENSITIVE_DATA env var ignored when load_dotenv() runs after module import
## Description

**What happened?**
When setting `ENABLE_SENSITIVE_DATA=true` in a `.env` file and using `load_dotenv()` before calling `configure_otel_providers()` or `enable_instrumentation()`, sensitive data logging remains disabled (`enable_sensitive_data` stays `False`).

`ENABLE_SENSITIVE_DATA=true` should be respected regardless of whether it was set before or after the module was imported, since `load_dotenv()` populates `os.environ` before the user calls `configure_otel_providers()`.

**Steps to reproduce:**
1. Create a `.env` file with `ENABLE_SENSITIVE_DATA=true`
2. In your script: `import agent_framework`, then call `load_dotenv()`, then call `configure_otel_providers()`
3. Check `OBSERVABILITY_SETTINGS.enable_sensitive_data` — it is `False`

## Code Sample

```python
load_dotenv()  # populates os.environ with ENABLE_SENSITIVE_DATA=true

from agent_framework.observability import configure_otel_providers, OBSERVABILITY_SETTINGS

configure_otel_providers()
print(OBSERVABILITY_SETTINGS.enable_sensitive_data)  # False — should be True
```

## Root Cause

`OBSERVABILITY_SETTINGS = ObservabilitySettings()` is constructed at **module import time** (line 884 of `observability.py`). At that point, `os.environ` does not yet contain the `.env` values because `load_dotenv()` hasn't been called. The cached `enable_sensitive_data=False` is never refreshed.

`configure_otel_providers()` and `enable_instrumentation()` force `enable_instrumentation=True` but do **not** re-read `ENABLE_SENSITIVE_DATA` from `os.environ` when the parameter is `None`.

The same timing issue affects `VS_CODE_EXTENSION_PORT` in `configure_otel_providers()`.

## Package Versions

agent-framework-core: latest (main branch)

## Python Version

Python 3.13

## Additional Context

This is the standard usage pattern shown in samples — `load_dotenv()` is called after imports but before `configure_otel_providers()`. The `env_file_path` branch of `configure_otel_providers()` works correctly because it constructs a fresh `ObservabilitySettings(**settings_kwargs)`, but the common `else` branch (no `env_file_path`) does not.

eavanvalkenburg:
### Motivation and Context

When users call `load_dotenv()` after importing the `agent_framework.observability` module, environment variables like `ENABLE_SENSITIVE_DATA` and `VS_CODE_EXTENSION_PORT` were ignored because `ObservabilitySettings` was only read once at import time. This meant the module-level singleton retained stale values even after the environment changed.

Fixes #4119

### Description

The root cause was that `enable_instrumentation()` and `configure_otel_providers()` only updated settings when explicit parameters were passed, falling back to the already-stale module-level `OBSERVABILITY_SETTINGS` singleton otherwise. The fix re-reads environment variables by constructing a fresh `ObservabilitySettings()` instance at call time, using those refreshed values as defaults when no explicit parameter is provided. Explicit parameters still take precedence over environment variables. Tests verify the re-read behavior for `ENABLE_SENSITIVE_DATA`, `VS_CODE_EXTENSION_PORT`, and that explicit arguments override env vars.

### Contribution Checklist

---

### DIFF
--- a/python/packages/core/agent_framework/observability.py
+++ b/python/packages/core/agent_framework/observability.py
@@ -942,8 +942,6 @@ def enable_instrumentation(
     else:
         # Re-read from current environment in case env vars were set after import (e.g. load_dotenv())
         OBSERVABILITY_SETTINGS.enable_sensitive_data = _read_bool_env("ENABLE_SENSITIVE_DATA")
-    # Also refresh enable_console_exporters so the global stays in sync with the environment
-    OBSERVABILITY_SETTINGS.enable_console_exporters = _read_bool_env("ENABLE_CONSOLE_EXPORTERS")
 
 
 def configure_otel_providers(
--- a/python/packages/core/agent_framework/observability.py
+++ b/python/packages/core/agent_framework/observability.py
@@ -920,6 +920,9 @@ def enable_instrumentation(
     OBSERVABILITY_SETTINGS.enable_instrumentation = True
     if enable_sensitive_data is not None:
         OBSERVABILITY_SETTINGS.enable_sensitive_data = enable_sensitive_data
+    else:
+        # Re-read from current environment in case env vars were set after import (e.g. load_dotenv())
+        OBSERVABILITY_SETTINGS.enable_sensitive_data = ObservabilitySettings().enable_sensitive_data
 
 
 def configure_otel_providers(
@@ -1064,12 +1067,17 @@ def configure_otel_providers(
         OBSERVABILITY_SETTINGS._resource = updated_settings._resource  # type: ignore[reportPrivateUsage]
         OBSERVABILITY_SETTINGS._executed_setup = False  # type: ignore[reportPrivateUsage]
     else:
-        # Update the observability settings with the provided values
+        # Re-read settings from current environment in case env vars were set
+        # after import (e.g. via load_dotenv()). Explicit parameters take precedence.
+        refreshed = ObservabilitySettings()
         OBSERVABILITY_SETTINGS.enable_instrumentation = True
-        if enable_sensitive_data is not None:
-            OBSERVABILITY_SETTINGS.enable_sensitive_data = enable_sensitive_data
-        if vs_code_extension_port is not None:
-            OBSERVABILITY_SETTINGS.vs_code_extension_port = vs_code_extension_port
+        OBSERVABILITY_SETTINGS.enable_sensitive_data = (
+            enable_sensitive_data if enable_sensitive_data is not None else refreshed.enable_sensitive_data
+        )
+        OBSERVABILITY_SETTINGS.enable_console_exporters = refreshed.enable_console_exporters
+        OBSERVABILITY_SETTINGS.vs_code_extension_port = (
+            vs_code_extension_port if vs_code_extension_port is not None else refreshed.vs_code_extension_port
+        )
 
     OBSERVABILITY_SETTINGS._configure(  # type: ignore[reportPrivateUsage]
         additional_exporters=exporters,
--- a/python/packages/core/agent_framework/observability.py
+++ b/python/packages/core/agent_framework/observability.py
@@ -942,6 +942,8 @@ def enable_instrumentation(
     else:
         # Re-read from current environment in case env vars were set after import (e.g. load_dotenv())
         OBSERVABILITY_SETTINGS.enable_sensitive_data = _read_bool_env("ENABLE_SENSITIVE_DATA")
+    # Also refresh enable_console_exporters so the global stays in sync with the environment
+    OBSERVABILITY_SETTINGS.enable_console_exporters = _read_bool_env("ENABLE_CONSOLE_EXPORTERS")
 
 
 def configure_otel_providers(
--- a/python/packages/core/agent_framework/observability.py
+++ b/python/packages/core/agent_framework/observability.py
@@ -899,6 +899,25 @@ def get_meter(
 OBSERVABILITY_SETTINGS: ObservabilitySettings = ObservabilitySettings()
 
 
+def _read_bool_env(name: str, *, default: bool = False) -> bool:
+    """Read a boolean from an environment variable."""
+    value = os.getenv(name)
+    if value is None:
+        return default
+    return value.lower() in ("true", "1", "yes", "on")
+
+
+def _read_int_env(name: str, *, default: int | None = None) -> int | None:
+    """Read an optional integer from an environment variable."""
+    value = os.getenv(name)
+    if value is None:
+        return default
+    try:
+        return int(value)
+    except ValueError:
+        return default
+
+
 def enable_instrumentation(
     *,
     enable_sensitive_data: bool | None = None,
@@ -922,12 +941,13 @@ def enable_instrumentation(
         OBSERVABILITY_SETTINGS.enable_sensitive_data = enable_sensitive_data
     else:
         # Re-read from current environment in case env vars were set after import (e.g. load_dotenv())
-        OBSERVABILITY_SETTINGS.enable_sensitive_data = ObservabilitySettings().enable_sensitive_data
+        OBSERVABILITY_SETTINGS.enable_sensitive_data = _read_bool_env("ENABLE_SENSITIVE_DATA")
 
 
 def configure_otel_providers(
     *,
     enable_sensitive_data: bool | None = None,
+    enable_console_exporters: bool | None = None,
     exporters: list[LogRecordExporter | SpanExporter | MetricExporter] | None = None,
     views: list[View] | None = None,
     vs_code_extension_port: int | None = None,
@@ -966,6 +986,8 @@ def configure_otel_providers(
     Keyword Args:
         enable_sensitive_data: Enable OpenTelemetry sensitive events. Overrides
             the environment variable ENABLE_SENSITIVE_DATA if set. Default is None.
+        enable_console_exporters: Enable console exporters for traces, logs, and metrics.
+            Overrides the environment variable ENABLE_CONSOLE_EXPORTERS if set. Default is None.
         exporters: A list of custom exporters for logs, metrics or spans, or any combination.
             These will be added in addition to exporters configured via environment variables.
             Default is None.
@@ -1054,6 +1076,8 @@ def configure_otel_providers(
             settings_kwargs["env_file_encoding"] = env_file_encoding
         if enable_sensitive_data is not None:
             settings_kwargs["enable_sensitive_data"] = enable_sensitive_data
+        if enable_console_exporters is not None:
+            settings_kwargs["enable_console_exporters"] = enable_console_exporters
         if vs_code_extension_port is not None:
             settings_kwargs["vs_code_extension_port"] = vs_code_extension_port
 
@@ -1069,15 +1093,20 @@ def configure_otel_providers(
     else:
         # Re-read settings from current environment in case env vars were set
         # after import (e.g. via load_dotenv()). Explicit parameters take precedence.
-        refreshed = ObservabilitySettings()
         OBSERVABILITY_SETTINGS.enable_instrumentation = True
         OBSERVABILITY_SETTINGS.enable_sensitive_data = (
-            enable_sensitive_data if enable_sensitive_data is not None else refreshed.enable_sensitive_data
+            enable_sensitive_data if enable_sensitive_data is not None else _read_bool_env("ENABLE_SENSITIVE_DATA")
         )
-        OBSERVABILITY_SETTINGS.enable_console_exporters = refreshed.enable_console_exporters
+        OBSERVABILITY_SETTINGS.enable_console_exporters = (
+            enable_console_exporters
+            if enable_console_exporters is not None
+            else _read_bool_env("ENABLE_CONSOLE_EXPORTERS")
+        )
         OBSERVABILITY_SETTINGS.vs_code_extension_port = (
-            vs_code_extension_port if vs_code_extension_port is not None else refreshed.vs_code_extension_port
+            vs_code_extension_port if vs_code_extension_port is not None else _read_int_env("VS_CODE_EXTENSION_PORT")
         )
+        OBSERVABILITY_SETTINGS._resource = create_resource()  # type: ignore[reportPrivateUsage]
+        OBSERVABILITY_SETTINGS._executed_setup = False  # type: ignore[reportPrivateUsage]
 
     OBSERVABILITY_SETTINGS._configure(  # type: ignore[reportPrivateUsage]
         additional_exporters=exporters,
--- a/python/packages/core/tests/core/test_observability.py
+++ b/python/packages/core/tests/core/test_observability.py
@@ -1059,6 +1059,8 @@ def test_configure_otel_providers_reads_env_sensitive_data(monkeypatch):
 
     monkeypatch.setenv("ENABLE_INSTRUMENTATION", "false")
     monkeypatch.setenv("ENABLE_SENSITIVE_DATA", "false")
+    monkeypatch.delenv("VS_CODE_EXTENSION_PORT", raising=False)
+    monkeypatch.delenv("ENABLE_CONSOLE_EXPORTERS", raising=False)
     for key in [
         "OTEL_EXPORTER_OTLP_ENDPOINT",
         "OTEL_EXPORTER_OTLP_TRACES_ENDPOINT",
@@ -1115,6 +1117,8 @@ def test_configure_otel_providers_explicit_param_overrides_env(monkeypatch):
 
     monkeypatch.setenv("ENABLE_INSTRUMENTATION", "false")
     monkeypatch.setenv("ENABLE_SENSITIVE_DATA", "true")
+    monkeypatch.delenv("VS_CODE_EXTENSION_PORT", raising=False)
+    monkeypatch.delenv("ENABLE_CONSOLE_EXPORTERS", raising=False)
     for key in [
         "OTEL_EXPORTER_OTLP_ENDPOINT",
         "OTEL_EXPORTER_OTLP_TRACES_ENDPOINT",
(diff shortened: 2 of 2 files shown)


markwallace-microsoft:
<a href="https://github.com/microsoft/agent-framework/blob/8af8bd0743abacd8e8fe4518cf92e6bfea1d72e8/README.md">Python Test Coverage</a><details><summary>Python Test Coverage Report • </summary><table><tr><th>File</th><th>Stmts</th><th>Miss</th><th>Cover</th><th>Missing</th></tr><tbody><tr><td colspan="5"><b>packages/core/agent_framework</b></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/8af8bd0743abacd8e8fe4518cf92e6bfea1d72e8/packages/core/agent_framework/observability.py">observability.py</a></td><td>693</td><td>25</td><td>96%</td><td><a href="https://github.com/microsoft/agent-framework/blob/8af8bd0743abacd8e8fe4518cf92e6bfea1d72e8/packages/core/agent_framework/observability.py#L370-L371">370&ndash;371</a>, <a href="https://github.com/microsoft/agent-framework/blob/8af8bd0743abacd8e8fe4518cf92e6bfea1d72e8/packages/core/agent_framework/observability.py#L398">398</a>, <a href="https://github.com/microsoft/agent-framework/blob/8af8bd0743abacd8e8fe4518cf92e6bfea1d72e8/packages/core/agent_framework/observability.py#L400-L402">400&ndash;402</a>, <a href="https://github.com/microsoft/agent-framework/blob/8af8bd0743abacd8e8fe4518cf92e6bfea1d72e8/packages/core/agent_framework/observability.py#L405-L407">405&ndash;407</a>, <a href="https://github.com/microsoft/agent-framework/blob/8af8bd0743abacd8e8fe4518cf92e6bfea1d72e8/packages/core/agent_framework/observability.py#L412-L413">412&ndash;413</a>, <a href="https://github.com/microsoft/agent-framework/blob/8af8bd0743abacd8e8fe4518cf92e6bfea1d72e8/packages/core/agent_framework/observability.py#L419-L420">419&ndash;420</a>, <a href="https://github.com/microsoft/agent-framework/blob/8af8bd0743abacd8e8fe4518cf92e6bfea1d72e8/packages/core/agent_framework/observability.py#L426-L427">426&ndash;427</a>, <a href="https://github.com/microsoft/agent-framework/blob/8af8bd0743abacd8e8fe4518cf92e6bfea1d72e8/packages/core/agent_framework/observability.py#L717">717</a>, <a href="https://github.com/microsoft/agent-framework/blob/8af8bd0743abacd8e8fe4518cf92e6bfea1d72e8/packages/core/agent_framework/observability.py#L917-L918">917&ndash;918</a>, <a href="https://github.com/microsoft/agent-framework/blob/8af8bd0743abacd8e8fe4518cf92e6bfea1d72e8/packages/core/agent_framework/observability.py#L1080">1080</a>, <a href="https://github.com/microsoft/agent-framework/blob/8af8bd0743abacd8e8fe4518cf92e6bfea1d72e8/packages/core/agent_framework/observability.py#L1325-L1326">1325&ndash;1326</a>, <a href="https://github.com/microsoft/agent-framework/blob/8af8bd0743abacd8e8fe4518cf92e6bfea1d72e8/packages/core/agent_framework/observability.py#L1572">1572</a>, <a href="https://github.com/microsoft/agent-framework/blob/8af8bd0743abacd8e8fe4518cf92e6bfea1d72e8/packages/core/agent_framework/observability.py#L1754">1754</a>, <a href="https://github.com/microsoft/agent-framework/blob/8af8bd0743abacd8e8fe4518cf92e6bfea1d72e8/packages/core/agent_framework/observability.py#L2138">2138</a>, <a href="https://github.com/microsoft/agent-framework/blob/8af8bd0743abacd8e8fe4518cf92e6bfea1d72e8/packages/core/agent_framework/observability.py#L2140">2140</a></td></tr><tr><td><b>TOTAL</b></td><td><b>24023</b></td><td><b>2634</b></td><td><b>89%</b></td><td>&nbsp;</td></tr></tbody></table></details>

## Python Unit Test Overview
| 5247 | 20 :zzz: | 0 :x: | 0 :fire: | 1m 23s :stopwatch: |

python/packages/core/agent_framework/observability.py
```diff
@@ -1064,12 +1067,17 @@ def configure_otel_providers(
...
+        refreshed = ObservabilitySettings()
         OBSERVABILITY_SETTINGS.enable_instrumentation = True
-        if enable_sensitive_data is not None:
-            OBSERVABILITY_SETTINGS.enable_sensitive_data = enable_sensitive_data
-        if vs_code_extension_port is not None:
-            OBSERVABILITY_SETTINGS.vs_code_extension_port = vs_code_extension_port
+        OBSERVABILITY_SETTINGS.enable_sensitive_data = (
+            enable_sensitive_data if enable_sensitive_data is not None else refreshed.enable_sensitive_data
+        )
+        OBSERVABILITY_SETTINGS.enable_console_exporters = refreshed.enable_console_exporters
```
eavanvalkenburg: This line unconditionally overwrites `enable_console_exporters` from the environment, but `configure_otel_providers` does not accept an explicit `enable_console_exporters` parameter to allow callers to override it (unlike `enable_sensitive_data` and `vs_code_extension_port`). Any prior programmatic assignment to `OBSERVABILITY_SETTINGS.enable_console_exporters` is silently discarded, and there is no test covering this behaviour. Consider adding both an override parameter and a test.

python/packages/core/tests/core/test_observability.py
```diff
@@ -1033,6 +1033,104 @@ def test_enable_instrumentation_with_sensitive_data(monkeypatch):
...
+
+    observability = importlib.import_module("agent_framework.observability")
+    importlib.reload(observability)
+
+    # Explicit False should override the env var True
+    observability.configure_otel_providers(enable_sensitive_data=False)
+    assert observability.OBSERVABILITY_SETTINGS.enable_sensitive_data is False
+
+
 # region Test _to_otel_part content types
```
eavanvalkenburg, line 1304: This test suite covers `enable_sensitive_data` and `vs_code_extension_port` env-var refresh but has no test for the new `enable_console_exporters` refresh added at line 1077 of observability.py. Please add a test that sets `ENABLE_CONSOLE_EXPORTERS=true` after module reload and asserts the setting is picked up by `configure_otel_providers()`.

python/packages/core/agent_framework/observability.py
```diff
@@ -920,6 +920,9 @@ def enable_instrumentation(
     OBSERVABILITY_SETTINGS.enable_instrumentation = True
     if enable_sensitive_data is not None:
         OBSERVABILITY_SETTINGS.enable_sensitive_data = enable_sensitive_data
+    else:
+        # Re-read from current environment in case env vars were set after import (e.g. load_dotenv())
+        OBSERVABILITY_SETTINGS.enable_sensitive_data = ObservabilitySettings().enable_sensitive_data
```
eavanvalkenburg: Creating a full `ObservabilitySettings()` here just to read `enable_sensitive_data` also constructs a throw-away `_resource` via `create_resource()`. A better fix is making `enable_sensitive_data`, `enable_console_exporters`, and `vs_code_extension_port` lazy properties that read from `os.environ` on every access, so no caller—public API or direct—can observe a stale value.

```suggestion
        OBSERVABILITY_SETTINGS.enable_sensitive_data = _read_bool_env("ENABLE_SENSITIVE_DATA", default=False)
```

python/packages/core/agent_framework/observability.py
```diff
@@ -1064,12 +1067,17 @@ def configure_otel_providers(
         OBSERVABILITY_SETTINGS._resource = updated_settings._resource  # type: ignore[reportPrivateUsage]
         OBSERVABILITY_SETTINGS._executed_setup = False  # type: ignore[reportPrivateUsage]
     else:
-        # Update the observability settings with the provided values
+        # Re-read settings from current environment in case env vars were set
+        # after import (e.g. via load_dotenv()). Explicit parameters take precedence.
+        refreshed = ObservabilitySettings()
```
eavanvalkenburg: Same concern as in `enable_instrumentation()`: `ObservabilitySettings()` construction calls `create_resource()` and the resulting resource is silently discarded. Separating settings parsing from resource construction in `ObservabilitySettings.__init__` would eliminate this waste and make the refresh pattern safe to call repeatedly.

eavanvalkenburg (commented):
## Automated Code Review

**Reviewers:** 4 | **Confidence:** 88%

### ✓ Correctness
> The changes correctly implement re-reading environment variables when `enable_instrumentation()` and `configure_otel_providers()` are called without explicit parameters, addressing the use case where `load_dotenv()` sets env vars after module import. The logic for explicit-parameter-overrides-env is sound, and the new tests cover the key scenarios. No correctness issues found.

### ✓ Security Reliability
> The changes allow `enable_instrumentation()` and `configure_otel_providers()` to re-read environment variables at call time instead of relying only on import-time values, supporting `load_dotenv()` workflows. The security-sensitive `enable_sensitive_data` flag correctly preserves explicit-parameter-takes-precedence semantics. No injection risks, resource leaks, or unsafe deserialization found. One minor behavioral change—`enable_console_exporters` is now unconditionally refreshed from the environment in the `else` branch of `configure_otel_providers`, which was not previously updated there—is consistent with the `env_file_path` branch but lacks test coverage.

### ✗ Test Coverage
> The new tests cover the main happy-path scenarios for the env-var re-read behaviour in both `enable_instrumentation` and `configure_otel_providers`. However, the newly added `enable_console_exporters` refresh line in `configure_otel_providers` has no test coverage at all, and there is no test verifying that `enable_instrumentation(enable_sensitive_data=False)` still takes precedence over a truthy env var (the symmetric explicit-override case for that function).

### ✗ Design Approach
> The change fixes a real problem — `OBSERVABILITY_SETTINGS` is constructed at module import time (line 899), so env vars set later via `load_dotenv()` are invisible to it. The fix is a symptom-level workaround: it instantiates a fresh `ObservabilitySettings()` inside each public API function to re-read env vars. This works only for the two code paths covered, but `ObservabilitySettings` is a public global, so any caller accessing it directly still sees the stale values. Additionally, `ObservabilitySettings()` construction calls `create_resource()`, which builds an OTel `Resource` object that is immediately discarded; the fix uses heavy object creation as a mechanism for what is essentially an env-var read. The root cause — eager construction of the singleton at import time — is not addressed. A cleaner fix would be to make the env-var-backed fields lazy properties on `ObservabilitySettings` (reading from `os.environ` on access), or to delay singleton construction to first use. There is also a minor inconsistency: `enable_instrumentation()` refreshes `enable_sensitive_data` but not `enable_console_exporters`, whereas `configure_otel_providers()` refreshes both; if the user sets `ENABLE_CONSOLE_EXPORTERS` after import and only calls `enable_instrumentation()`, that flag is silently ignored.

### Flagged Issues
- [ ] No test covers the new `OBSERVABILITY_SETTINGS.enable_console_exporters = refreshed.enable_console_exporters` line (~line 1077 in observability.py). If this line were deleted or regressed, no test would fail.
- [ ] The fix patches only the two public API code paths; `OBSERVABILITY_SETTINGS` remains a module-level public global whose env-var-backed fields are stale for any direct caller. The root cause—eager singleton construction at import time—is not addressed. Consider making these fields lazy properties that read `os.environ` on access, or lazily constructing the singleton on first use.
- [ ] Each call to `enable_instrumentation()` or `configure_otel_providers()` now instantiates a full `ObservabilitySettings()` whose constructor also calls `create_resource()`, building an OTel `Resource` that is immediately discarded. Settings parsing should be separated from resource creation so the refresh pattern does not carry unnecessary side effects.

### Suggestions
- Add a test for `enable_instrumentation(enable_sensitive_data=False)` when `ENABLE_SENSITIVE_DATA=true` is in the environment, to verify the explicit parameter still wins (mirrors the existing `test_configure_otel_providers_explicit_param_overrides_env` but for the simpler function).
- In `enable_instrumentation()`, `enable_console_exporters` is not refreshed from the environment (only `enable_sensitive_data` is), whereas `configure_otel_providers()` refreshes both. If a user sets `ENABLE_CONSOLE_EXPORTERS` after import and calls only `enable_instrumentation()`, the env var is silently ignored. Either document this intentional asymmetry or refresh `enable_console_exporters` in both functions.
- The tests use `importlib.reload(observability)`, which re-runs module-level code but can leave stale references in other already-imported modules. This fragility should be noted if the module gains cross-module references.

---

eavanvalkenburg (commented):
## Automated Code Review

### ✓ Correctness
> The change correctly adds a re-read of ENABLE_CONSOLE_EXPORTERS from the environment inside enable_instrumentation(), mirroring the existing pattern for ENABLE_SENSITIVE_DATA. The new line in the existing test (delenv for ENABLE_CONSOLE_EXPORTERS) prevents test pollution from the new behavior. The new test is well-structured and follows the same pattern as the surrounding tests. No correctness issues found.

### ✓ Security Reliability
> The change adds a single line to `enable_instrumentation()` that refreshes the `enable_console_exporters` setting from the environment, matching the existing pattern used for `enable_sensitive_data`. The helper `_read_bool_env` performs a safe `os.getenv` followed by a lowercase string comparison—no injection, deserialization, or resource-leak risk. The new test properly isolates environment variables via `monkeypatch` and validates the round-trip. The existing test gains a defensive `delenv` for the new variable. No security or reliability issues found.

### ✓ Test Coverage
> The diff adds a one-line production fix to refresh `enable_console_exporters` from the environment inside `enable_instrumentation()`, plus a well-structured test that covers the primary scenario (env var absent at import, set before calling `enable_instrumentation`). The existing `test_configure_otel_providers_reads_env_vs_code_port` test also gets a minor cleanup to delete `ENABLE_CONSOLE_EXPORTERS` to avoid cross-contamination. The new test follows the same reload-and-assert pattern used by sibling tests and its assertions are meaningful.

### ✗ Design Approach
> The fix unconditionally overwrites `OBSERVABILITY_SETTINGS.enable_console_exporters` from the environment every time `enable_instrumentation()` is called. This is problematic for two reasons. First, it is a leaky abstraction: `enable_instrumentation()` is documented as a function that only flips the instrumentation flag and does not configure exporters or providers; folding console-exporter state management into it bleeds a provider-configuration concern across a documented layer boundary. Second, `configure_otel_providers()` already correctly re-reads `ENABLE_CONSOLE_EXPORTERS` from the environment at call time (lines 1102–1106), so for the primary load_dotenv use case the fix is redundant on that path. On the `enable_instrumentation()`-only path (users who wire their own providers, e.g. Azure Monitor), the value written is never read by any exporter-creation code—`enable_console_exporters` is only consumed inside `_configure()`, which `enable_instrumentation()` never calls. The unconditional overwrite also introduces an asymmetry: `enable_sensitive_data` can be protected by passing it explicitly, but there is no corresponding `enable_console_exporters` parameter to `enable_instrumentation()`, so any value previously set programmatically (e.g., via a prior `configure_otel_providers(enable_console_exporters=True)` call) can be silently clobbered by a subsequent `enable_instrumentation()` call when the env var is absent.

### Flagged Issues
- [ ] Unconditional overwrite in `enable_instrumentation()` silently resets `enable_console_exporters` to `False` whenever the env var is absent, clobbering any value previously set by `configure_otel_providers(enable_console_exporters=True)`. This is a destructive side-effect on a function documented as only enabling the instrumentation flag.
- [ ] `enable_console_exporters` is only consumed inside `_configure()`, which `enable_instrumentation()` never calls. The refresh has no effect on the `enable_instrumentation()`-only code path and is redundant for the `configure_otel_providers()` path (which already re-reads the env var), so it does not solve a real problem but introduces regression risk.

### Suggestions
- If the goal is to let `enable_instrumentation()` influence console exporters, add an explicit `enable_console_exporters: bool | None = None` parameter (mirroring `enable_sensitive_data`) with a conditional pattern: use the explicit value when provided, fall back to `_read_bool_env` otherwise. This preserves caller intent and maintains API symmetry. If the goal is solely to fix the `load_dotenv` timing issue, no change to `enable_instrumentation()` is needed—`configure_otel_providers()` already re-reads the env var at call time.
- Consider adding a reverse-direction test (set ENABLE_CONSOLE_EXPORTERS=true before reload, then delete it, call `enable_instrumentation()`, and assert the setting reverts to False) to verify the re-read works in both directions.
- Consider adding a test that verifies `enable_console_exporters` is refreshed even when `enable_sensitive_data` is explicitly passed, since the new line runs unconditionally regardless of that branch.

---

python/packages/core/agent_framework/observability.py
```diff
@@ -920,11 +939,17 @@ def enable_instrumentation(
     OBSERVABILITY_SETTINGS.enable_instrumentation = True
     if enable_sensitive_data is not None:
         OBSERVABILITY_SETTINGS.enable_sensitive_data = enable_sensitive_data
+    else:
+        # Re-read from current environment in case env vars were set after import (e.g. load_dotenv())
+        OBSERVABILITY_SETTINGS.enable_sensitive_data = _read_bool_env("ENABLE_SENSITIVE_DATA")
+    # Also refresh enable_console_exporters so the global stays in sync with the environment
+    OBSERVABILITY_SETTINGS.enable_console_exporters = _read_bool_env("ENABLE_CONSOLE_EXPORTERS")
```
eavanvalkenburg: This unconditional overwrite is both a leaky abstraction and a regression risk. `enable_instrumentation()` is documented as not configuring exporters, so managing `enable_console_exporters` here bleeds provider-layer concerns into it. More critically, this silently resets a value that may have been deliberately set by a prior `configure_otel_providers(enable_console_exporters=True)` call when `ENABLE_CONSOLE_EXPORTERS` is absent from env. If an explicit override parameter is added (matching the `enable_sensitive_data` pattern), the env fallback should be conditional on the parameter being `None`.

```suggestion
    if enable_console_exporters is not None:
        OBSERVABILITY_SETTINGS.enable_console_exporters = enable_console_exporters
    else:
        # Re-read from current environment in case env vars were set after import (e.g. load_dotenv())
        OBSERVABILITY_SETTINGS.enable_console_exporters = _read_bool_env("ENABLE_CONSOLE_EXPORTERS")
```

python/packages/core/agent_framework/observability.py
```diff
@@ -1064,12 +1067,17 @@ def configure_otel_providers(
...
-            OBSERVABILITY_SETTINGS.enable_sensitive_data = enable_sensitive_data
-        if vs_code_extension_port is not None:
-            OBSERVABILITY_SETTINGS.vs_code_extension_port = vs_code_extension_port
+        OBSERVABILITY_SETTINGS.enable_sensitive_data = (
+            enable_sensitive_data if enable_sensitive_data is not None else refreshed.enable_sensitive_data
+        )
+        OBSERVABILITY_SETTINGS.enable_console_exporters = refreshed.enable_console_exporters
+        OBSERVABILITY_SETTINGS.vs_code_extension_port = (
+            vs_code_extension_port if vs_code_extension_port is not None else refreshed.vs_code_extension_port
+        )
```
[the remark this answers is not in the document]
eavanvalkenburg: Fixed in `a318552b`.

python/packages/core/tests/core/test_observability.py
```diff
@@ -1033,6 +1033,104 @@ def test_enable_instrumentation_with_sensitive_data(monkeypatch):
...
+
+    assert observability.OBSERVABILITY_SETTINGS.enable_sensitive_data is False
+
+    # Simulate load_dotenv() setting env var after import
+    monkeypatch.setenv("ENABLE_SENSITIVE_DATA", "true")
+
+    observability.configure_otel_providers()
+    assert observability.OBSERVABILITY_SETTINGS.enable_instrumentation is True
+    assert observability.OBSERVABILITY_SETTINGS.enable_sensitive_data is True
+
```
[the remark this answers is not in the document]
eavanvalkenburg: Fixed in `a318552b`.

python/packages/core/tests/core/test_observability.py
```diff
@@ -1033,6 +1033,104 @@ def test_enable_instrumentation_with_sensitive_data(monkeypatch):
...
+    ]:
+        monkeypatch.delenv(key, raising=False)
+
+    observability = importlib.import_module("agent_framework.observability")
+    importlib.reload(observability)
+
+    # Explicit False should override the env var True
+    observability.configure_otel_providers(enable_sensitive_data=False)
+    assert observability.OBSERVABILITY_SETTINGS.enable_sensitive_data is False
+
```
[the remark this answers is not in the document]
eavanvalkenburg: Fixed in `a318552b`.

python/packages/core/agent_framework/observability.py
eavanvalkenburg: Fixed in `a318552b`.

python/packages/core/tests/core/test_observability.py
eavanvalkenburg: Fixed in `a318552b`.

python/packages/core/agent_framework/observability.py
eavanvalkenburg: Fixed in `a318552b`.

python/packages/core/agent_framework/observability.py
eavanvalkenburg: Fixed in `a318552b`.

python/packages/core/agent_framework/observability.py
eavanvalkenburg: Fixed in `a318552b`.

eavanvalkenburg (commented):
## Automated Code Review

**Reviewers:** 4 | **Confidence:** 94%

### ✓ Correctness
> The diff correctly removes the side-effect where enable_instrumentation() was unconditionally overwriting enable_console_exporters from the environment, which could clobber a value previously set by configure_otel_providers(). The responsibility for managing enable_console_exporters is properly scoped to configure_otel_providers(), which still reads the env var (lines 1100-1104). The four new tests thoroughly verify the fix: env var set after import is not picked up, prior configure_otel_providers value is preserved, explicit enable_sensitive_data param doesn't clobber, and env var removal after reload doesn't reset. No correctness issues found.

### ✓ Security Reliability
> This diff removes a line from `enable_instrumentation()` that was unconditionally overwriting `OBSERVABILITY_SETTINGS.enable_console_exporters` by re-reading from the environment. This was a bug: it clobbered values previously set by `configure_otel_providers()`. The fix is correct—`enable_console_exporters` is an exporter concern managed exclusively by `configure_otel_providers`, which already handles both explicit parameters and env-var fallback (lines 1079-1104). The new tests thoroughly verify that `enable_instrumentation` no longer touches the console-exporters setting. No security or reliability issues found.

### ✓ Test Coverage
> The diff removes a two-line side-effect from `enable_instrumentation` that was clobbering `enable_console_exporters`, and replaces one existing test with four well-targeted tests covering the new behavior. The original test was updated to assert the opposite outcome (value stays `False` instead of being set to `True`), and three new tests cover: clobbering after `configure_otel_providers`, clobbering when `enable_sensitive_data` is passed, and preservation when the env var is removed post-reload. Assertions are meaningful and specific. The test coverage for the behavioral change is thorough.

### ✓ Design Approach
> The diff removes an unconditional re-read of `ENABLE_CONSOLE_EXPORTERS` from `enable_instrumentation()`, which previously clobbered any value set by `configure_otel_providers`. This is a well-motivated fix: `enable_instrumentation()` is documented as not configuring exporters or providers, yet the removed code was silently overwriting the exporter-level setting on every call, making it impossible to preserve a value set by `configure_otel_providers(enable_console_exporters=True)`. The fix correctly narrows `enable_instrumentation` to its stated contract. The four new tests are clear and cover the key scenarios (env var post-reload, clobbering a prior configure call, combined with explicit sensitive_data param). No blocking design issues found.

### Suggestions
- The first test (`test_enable_instrumentation_does_not_touch_console_exporters`) cleans up fewer env vars than the other three new tests (it doesn't clear `ENABLE_SENSITIVE_DATA`, `VS_CODE_EXTENSION_PORT`, or the `OTEL_EXPORTER_OTLP_*` vars). While this doesn't cause a correctness issue today, aligning the env cleanup across all four tests would make them more robust against future refactors.
- The same clobbering risk now exists for `enable_sensitive_data`: calling `configure_otel_providers(enable_sensitive_data=True)` followed by `enable_instrumentation()` (without the explicit param) resets it to the env-var value. This is pre-existing and outside this PR's scope, but worth a follow-up to make the two settings consistent in how they handle values already set by `configure_otel_providers`.

---

moonbox3 (approved):

giles17 (approved):

Issue #2752, opened 2025-12-10
Python: What am I doing wrong here? The latency on MAF is unbelievable


moelk-atea:
**I wrote a script were I compared calling the same model through 3 different endpoints/clients:**
1. Using the OpenAI library
2. Using the Azure AI foundry library
3. Using the MAF library

```
Here are the results:

===============================================================================================
AGGREGATE RESULTS
===============================================================================================

Metric                                      Foundry                  MAF               OpenAI
────────────────────────────── ──────────────────── ──────────────────── ────────────────────
Time to First Token (ms)                     1775.1               2638.6               1401.0
  └─ Winner                                                                                 ✓

Total Time (s)                                 1.89                12.86                 1.52
  └─ Winner                                                                                 ✓

Throughput (tokens/s)                         14.22                 1.69                14.97
  └─ Winner                                                                                 ✓

───────────────────────────────────────────────────────────────────────────────────────────────

Errors:
  Foundry: 0/9
  MAF:     0/9
  OpenAI:  0/9

===============================================================================================
WINNER BREAKDOWN
===============================================================================================

🏆 OpenAI wins 3/3 metrics

Metric Winners:
  • Time to First Token: OpenAI (1401ms)
  • Total Time:          OpenAI (1.52s)
  • Throughput:          OpenAI (15.0 tok/s)

===============================================================================================
```

**The different between azure ai foundry and openai is marginal but MAF is not even comparable, am I doing something wrong here? Look at the code below:**

```
"""
Three-Way Latency Comparison: Azure AI Foundry vs MAF vs Azure OpenAI

Compares streaming latency across three different approaches:
1. Azure AI Foundry (Direct) - azure.ai.inference SDK for direct model inference
2. Microsoft Agent Framework (MAF) - MAF ChatAgent with Foundry project endpoint
3. Azure OpenAI (Direct) - OpenAI SDK with Azure OpenAI endpoint

All three use the same model deployment to isolate endpoint/framework differences.
"""

import os
import time
import asyncio
import statistics
from typing import List, Dict
from azure.identity.aio import DefaultAzureCredential, get_bearer_token_provider
from azure.ai.inference.aio import ChatCompletionsClient
from azure.ai.inference.models import SystemMessage, UserMessage
from agent_framework.azure import AzureAIClient
from agent_framework import ChatAgent
from openai import AsyncAzureOpenAI
from dotenv import load_dotenv

load_dotenv("XXX")

# Endpoint configurations
[...]

PR #4744, opened 2026-03-17
Python: Reduce Azure chat client import overhead

eavanvalkenburg:
### Motivation and Context

Latency benchmarking on current `main` showed `import_azure_openai_chat_client` regressed relative to the latest comparable historical baseline. Profiling pointed at `agent_framework.azure._chat_client` eagerly importing heavier OpenAI runtime modules and top-level package surfaces during module import. This change narrows that import path and helps address the latency work tracked in #2752.

### Description

- move `AsyncAzureOpenAI` behind `TYPE_CHECKING` so it is not imported at module load time
- stop eagerly importing OpenAI `Choice` and `ChunkChoice` runtime types
- import `OpenAIChatOptions` directly from `agent_framework.openai._chat_client` instead of the heavier `agent_framework.openai` package surface
- update `_parse_text_from_openai` to read `message` and `delta` via `getattr`, preserving the existing parsing behavior without those runtime type imports

This keeps the change isolated to import-time behavior and avoids changing the public Azure chat client API surface. The change was exercised with the non-integration Azure chat client test suite.

### Contribution Checklist

### DIFF
--- a/python/packages/core/agent_framework/azure/_chat_client.py
+++ b/python/packages/core/agent_framework/azure/_chat_client.py
@@ -45,6 +45,8 @@ else:
 
 if TYPE_CHECKING:
     from openai.lib.azure import AsyncAzureOpenAI
+    from openai.types.chat.chat_completion import Choice
+    from openai.types.chat.chat_completion_chunk import Choice as ChunkChoice
 
     from agent_framework._middleware import MiddlewareTypes
 
@@ -288,7 +290,7 @@ class AzureOpenAIChatClient(  # type: ignore[misc]
         )
 
     @override
-    def _parse_text_from_openai(self, choice: Any) -> Content | None:
+    def _parse_text_from_openai(self, choice: Choice | ChunkChoice) -> Content | None:
         """Parse the choice into a Content object with type='text'.
 
         Overwritten from RawOpenAIChatClient to deal with Azure On Your Data function.
--- a/python/packages/core/agent_framework/azure/_chat_client.py
+++ b/python/packages/core/agent_framework/azure/_chat_client.py
@@ -8,9 +8,6 @@ import sys
 from collections.abc import Mapping, Sequence
 from typing import TYPE_CHECKING, Any, Generic, cast
 
-from openai.lib.azure import AsyncAzureOpenAI
-from openai.types.chat.chat_completion import Choice
-from openai.types.chat.chat_completion_chunk import Choice as ChunkChoice
 from pydantic import BaseModel
 
 from agent_framework import (
@@ -23,8 +20,7 @@ from agent_framework import (
     FunctionInvocationLayer,
 )
 from agent_framework.observability import ChatTelemetryLayer
-from agent_framework.openai import OpenAIChatOptions
-from agent_framework.openai._chat_client import RawOpenAIChatClient
+from agent_framework.openai._chat_client import OpenAIChatOptions, RawOpenAIChatClient
 
 from .._settings import load_settings
 from ._entra_id_authentication import AzureCredentialTypes, AzureTokenProvider
@@ -48,6 +44,8 @@ else:
     from typing_extensions import TypedDict  # type: ignore # pragma: no cover
 
 if TYPE_CHECKING:
+    from openai.lib.azure import AsyncAzureOpenAI
+
     from agent_framework._middleware import MiddlewareTypes
 
 logger: logging.Logger = logging.getLogger(__name__)
@@ -290,14 +288,16 @@ class AzureOpenAIChatClient(  # type: ignore[misc]
         )
 
     @override
-    def _parse_text_from_openai(self, choice: Choice | ChunkChoice) -> Content | None:
+    def _parse_text_from_openai(self, choice: Any) -> Content | None:
         """Parse the choice into a Content object with type='text'.
 
         Overwritten from RawOpenAIChatClient to deal with Azure On Your Data function.
         For docs see:
         https://learn.microsoft.com/en-us/azure/ai-foundry/openai/references/on-your-data?tabs=python#context
         """
-        message = choice.message if isinstance(choice, Choice) else choice.delta
+        message = getattr(choice, "message", None)
+        if message is None:
+            message = getattr(choice, "delta", None)
         # When you enable asynchronous content filtering in Azure OpenAI, you may receive empty deltas
         if message is None:  # type: ignore
             return None
--- a/python/packages/core/tests/azure/test_azure_chat_client.py
+++ b/python/packages/core/tests/azure/test_azure_chat_client.py
@@ -652,6 +652,88 @@ async def test_streaming_with_none_delta(
     assert any(msg.contents for msg in results)
 
 
+# region _parse_text_from_openai direct unit tests
+
+
+def test_parse_text_from_openai_with_choice_message(azure_openai_unit_test_env: dict[str, str]) -> None:
+    """Test _parse_text_from_openai correctly reads message from a Choice."""
+    client = AzureOpenAIChatClient()
+    choice = Choice(
+        index=0,
+        message=ChatCompletionMessage(content="hello", role="assistant"),
+        finish_reason="stop",
+    )
+    result = client._parse_text_from_openai(choice)
+    assert result is not None
+    assert result.type == "text"
+    assert result.text == "hello"
+
+
+def test_parse_text_from_openai_with_chunk_choice_delta(azure_openai_unit_test_env: dict[str, str]) -> None:
+    """Test _parse_text_from_openai correctly reads delta from a ChunkChoice."""
+    client = AzureOpenAIChatClient()
+    choice = ChunkChoice(
+        index=0,
+        delta=ChunkChoiceDelta(content="streamed", role="assistant"),
+        finish_reason=None,
+    )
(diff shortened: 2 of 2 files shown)


markwallace-microsoft:
<a href="https://github.com/microsoft/agent-framework/blob/569ea9b2d285599ae56205404b896d34d9a6454d/README.md">Python Test Coverage</a><details><summary>Python Test Coverage Report • </summary><table><tr><th>File</th><th>Stmts</th><th>Miss</th><th>Cover</th><th>Missing</th></tr><tbody><tr><td colspan="5"><b>packages/core/agent_framework/azure</b></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/569ea9b2d285599ae56205404b896d34d9a6454d/packages/core/agent_framework/azure/_chat_client.py">_chat_client.py</a></td><td>87</td><td>3</td><td>96%</td><td><a href="https://github.com/microsoft/agent-framework/blob/569ea9b2d285599ae56205404b896d34d9a6454d/packages/core/agent_framework/azure/_chat_client.py#L322-L323">322&ndash;323</a>, <a href="https://github.com/microsoft/agent-framework/blob/569ea9b2d285599ae56205404b896d34d9a6454d/packages/core/agent_framework/azure/_chat_client.py#L334">334</a></td></tr><tr><td><b>TOTAL</b></td><td><b>24004</b></td><td><b>2638</b></td><td><b>89%</b></td><td>&nbsp;</td></tr></tbody></table></details>

## Python Unit Test Overview
| 5242 | 20 :zzz: | 0 :x: | 0 :fire: | 1m 24s :stopwatch: |

python/packages/core/agent_framework/azure/_chat_client.py
```diff
@@ -290,14 +288,16 @@ class MyOptions(AzureOpenAIChatOptions, total=False):
...
-    def _parse_text_from_openai(self, choice: Choice | ChunkChoice) -> Content | None:
+    def _parse_text_from_openai(self, choice: Any) -> Content | None:
         """Parse the choice into a Content object with type='text'.
 
         Overwritten from RawOpenAIChatClient to deal with Azure On Your Data function.
         For docs see:
         https://learn.microsoft.com/en-us/azure/ai-foundry/openai/references/on-your-data?tabs=python#context
         """
-        message = choice.message if isinstance(choice, Choice) else choice.delta
+        message = getattr(choice, "message", None)
```
eavanvalkenburg, line 300: The switch from `isinstance(choice, Choice)` to `getattr` fallback introduces a subtle behavioral change: if `choice.message` exists but is falsy (e.g., `0`, `''`, `False`), the old code would use it, but the new code falls through to `delta`. This is likely fine for the known OpenAI types, but please add a direct unit test for `_parse_text_from_openai` on the Azure client to lock in the expected behavior—especially the refusal branch that has no test coverage at all.

python/packages/core/agent_framework/azure/_chat_client.py
```diff
@@ -48,6 +44,8 @@
     from typing_extensions import TypedDict  # type: ignore # pragma: no cover
 
 if TYPE_CHECKING:
+    from openai.lib.azure import AsyncAzureOpenAI
```
eavanvalkenburg, line 47: Since `from __future__ import annotations` is already in this file, annotations are never evaluated at runtime. Add `Choice` and `ChunkChoice` here under `TYPE_CHECKING` so the override can keep the proper typed signature without any runtime import cost.

```suggestion
if TYPE_CHECKING:
    from openai.lib.azure import AsyncAzureOpenAI
    from openai.types.chat.chat_completion import Choice
    from openai.types.chat.chat_completion_chunk import Choice as ChunkChoice

    from agent_framework._middleware import MiddlewareTypes
```

python/packages/core/agent_framework/azure/_chat_client.py
```diff
@@ -290,14 +288,16 @@ class MyOptions(AzureOpenAIChatOptions, total=False):
...
         For docs see:
         https://learn.microsoft.com/en-us/azure/ai-foundry/openai/references/on-your-data?tabs=python#context
         """
-        message = choice.message if isinstance(choice, Choice) else choice.delta
+        message = getattr(choice, "message", None)
+        if message is None:
+            message = getattr(choice, "delta", None)
         # When you enable asynchronous content filtering in Azure OpenAI, you may receive empty deltas
         if message is None:  # type: ignore
             return None
```
eavanvalkenburg, line 305: Using `Any` as the parameter type discards the type contract from the base class and is unnecessary here. With `from __future__ import annotations` already present and `Choice`/`ChunkChoice` available under `TYPE_CHECKING`, the signature can remain `choice: Choice | ChunkChoice`.

```suggestion
    def _parse_text_from_openai(self, choice: Choice | ChunkChoice) -> Content | None:
```

eavanvalkenburg (commented):
## Automated Code Review

**Reviewers:** 3 | **Confidence:** 89%

### ✓ Correctness
> The diff makes two changes: (1) moves openai type imports behind TYPE_CHECKING or removes them, and (2) replaces the isinstance-based dispatch in `_parse_text_from_openai` with a getattr-based duck-typing approach. Both changes are logically correct. The `from __future__ import annotations` import ensures `AsyncAzureOpenAI` can safely move to TYPE_CHECKING. The getattr fallback from `message` to `delta` correctly mirrors the original isinstance logic since `Choice` has `message` and `ChunkChoice` has `delta` (and not the other). The re-export path change for `OpenAIChatOptions` is valid as it's defined in `_chat_client.py` and also re-exported from `__init__.py`. No correctness issues found.

### ✗ Test Coverage
> The diff changes `_parse_text_from_openai` in `AzureOpenAIChatClient` from isinstance-based dispatch to duck-typing via `getattr`. Existing tests cover the non-streaming path (via `test_azure_on_your_data*`) and streaming with `None` delta (via `test_streaming_with_none_delta`), but all exercise the method indirectly through the full `get_response` flow. There is no direct unit test for the Azure override of `_parse_text_from_openai`, and notably the refusal branch inside this override has zero test coverage—the existing `test_parse_text_with_refusal` only tests the base `RawOpenAIChatClient`. Adding targeted tests for the Azure-specific method would catch any regressions from the isinstance-to-getattr refactor.

### ✗ Design Approach
> The diff makes three changes: (1) moves `AsyncAzureOpenAI` to the `TYPE_CHECKING` block (correct), (2) consolidates `OpenAIChatOptions` import from the public `agent_framework.openai` package into the private `agent_framework.openai._chat_client` module, and (3) replaces the typed `choice: Choice | ChunkChoice` override signature with `choice: Any` and substitutes `isinstance(choice, Choice)` discrimination with `getattr` duck-typing. The `getattr` approach at runtime is fine given the imports are being removed, but widening the signature to `Any` is unnecessary and the wrong tradeoff: since the file already has `from __future__ import annotations` (PEP 563), adding `Choice` and `ChunkChoice` to the `TYPE_CHECKING` block would let the type annotation remain fully typed without any runtime import cost. The import of `OpenAIChatOptions` from a private module (`_chat_client`) instead of the public package API is also a leaky abstraction, though less severe.

### Flagged Issues
- [ ] The Azure-specific `_parse_text_from_openai` override changed dispatch logic (isinstance → getattr fallback) but has no direct unit test. The refusal branch (`if hasattr(message, 'refusal') and message.refusal`) in this override is completely untested. A regression in the getattr fallback (e.g., a ChunkChoice that unexpectedly has a `message` attribute) or the refusal path would go undetected.
- [ ] The override signature `_parse_text_from_openai(self, choice: Any)` unnecessarily discards the type contract established by the base class (`Choice | ChunkChoice`). Because `from __future__ import annotations` is already present in this file, annotations are never evaluated at runtime — `Choice` and `ChunkChoice` only need to be added to the `TYPE_CHECKING` block (not imported unconditionally) to keep the typed signature. Using `Any` here silences the type checker and breaks the override contract without any actual benefit.

### Suggestions
- Add a direct unit test for `AzureOpenAIChatClient._parse_text_from_openai` covering: (1) a Choice-like object with only `message`, (2) a ChunkChoice-like object with only `delta`, (3) an object with both `message=None` and `delta` (verifying fallback), (4) the refusal path, and (5) an object with neither attribute (returns None).
- `OpenAIChatOptions` is part of the public API of `agent_framework.openai` (listed in `__all__`). The diff changes its import to come from the internal `agent_framework.openai._chat_client` module, bypassing the public surface. Import it from the public package (`from agent_framework.openai import OpenAIChatOptions`) to avoid coupling to implementation internals.

---

giles17 (approved):

python/packages/core/agent_framework/azure/_chat_client.py
eavanvalkenburg: Fixed in `569ea9b2`.

python/packages/core/agent_framework/azure/_chat_client.py
eavanvalkenburg: Fixed in `569ea9b2`.

python/packages/core/agent_framework/azure/_chat_client.py
eavanvalkenburg: Fixed in `569ea9b2`.

moonbox3 (approved):

PR #4745, opened 2026-03-17
.NET: fix: HandoffAgentExecutor does not output any response when non-streaming

### LINKED ISSUE #2423: .NET - Assistant responses are not stored when using Workflow as Agent with Handoff orchestration
**Issue Summary**
When implementing the Handoff workflow in the Microsoft Agent Framework using the official documentation example and Workflow as Agent pattern, only user messages are recorded in the thread. Assistant responses are not stored.

**Expected Behavior**
Both user and assistant messages should be stored to ensure a complete conversation history.

**Actual Behavior**
Only user messages appear in the log; assistant messages are missing.

**Reproduction Steps**
1. Use the [Handoff workflow example](https://learn.microsoft.com/en-us/agent-framework/user-guide/workflows/orchestrations/handoff?pivots=programming-language-csharp#set-up-the-azure-openai-client) provided in the official Microsoft documentation but with [Workflow as Agent](https://learn.microsoft.com/en-us/agent-framework/user-guide/workflows/as-agents?pivots=programming-language-csharp#creating-a-workflow-agent) as a result.
2. Trigger a handoff scenario.
3. Review the stored conversation history in thread.

**Impact**
Incomplete history limits context availability for agents and affects traceability.

Please advise whether assistant message logging can be enabled or if this is an unintended limitation.

Code snippet:
   ```
 static async Task Main(string[] args)
    {
        var client = new AzureOpenAIClient(new Uri(""),
            new System.ClientModel.ApiKeyCredential(""))
            .GetChatClient("gpt-5-mini")
            .AsIChatClient();

        ChatClientAgent historyTutor = new(client,
            "You provide assistance with historical queries. Explain important events and context clearly. Only respond about history.",
            "history_tutor",
            "Specialist agent for historical questions");

        ChatClientAgent mathTutor = new(client,
            "You provide help with math problems. Explain your reasoning at each step and include examples. Only respond about math.",
            "math_tutor",
            "Specialist agent for math questions");

        ChatClientAgent triageAgent = new(client,
            "You determine which agent to use based on the user's homework question. ALWAYS handoff to another agent.",
            "triage_agent",
            "Routes messages to the appropriate specialist agent");

        var workflowAgent = AgentWorkflowBuilder.CreateHandoffBuilderWith(triageAgent)
            .WithHandoffs(triageAgent, [mathTutor, historyTutor])
            .WithHandoff(mathTutor, triageAgent)                 
            .WithHandoff(historyTutor, triageAgent)               
            .Build()
            .AsAgent();

        var workflowAgentThread = workflowAgent.GetNewThread();

        while (true)
        {
            Console.WriteLine();
            Console.Write("User (or 'exit' to quit): ");
            string? input = Console.ReadLine();
            if (string.IsNullOrWhiteSpace(input) || input.Equals("exit", StringComparison.OrdinalIgnoreCase))
            {
                break;
            }

            await ProcessInputAsync(workflowAgent, workflowAgentThread, input);
        }
    }

    static async Task ProcessInputAsync(AIAgent agent, AgentThread thread, string input)
    {
        Dictionary<string, List<AgentRunResponseUpdate>> buffer = [];
        await foreach (AgentRunResponseUpdate update in agent.RunStreamingAsync(input, thread))
        {
            if (update.MessageId is null || string.IsNullOrEmpty(update.Text))
            {
                // skip updates that don't have a message ID or text
                continue;
            }

            if (!buffer.TryGetValue(update.MessageId, out List<AgentRunResponseUpdate>? value))
            {
                value = [];
                buffer[update.MessageId] = value;
            }
            value.Add(update);

            foreach (var (messageId, segments) in buffer)
            {
                string combinedText = string.Concat(segments);
                Console.WriteLine($"{segments[0].AuthorName}: {combinedText}");
                Console.WriteLine();
            }
        }
    }

```
Image

### PR DESCRIPTION
**Pull request by lokitoth**
### Motivation and Context

Due to needing custom logic and a more delocalized state, the `HandoffAgentExecutor` is still separate from the general `AIAgentHostExecutor`. The fixes made to the `AIAgentHostExecutor` to better support both streaming and non-streaming agent invocation modes and separate configuration for outputs (streaming vs. nonstream, output vs. no output) in PR #3142 did not get ported to `HandoffAgentExecutor` in hopes of unifying the two.

### Description

To better support Handoff workflows running in non-streaming mode (and in particular when hosted as an AIAgent), we change the `HandoffAgentExecutor` to always emit an `AgentRunResponse` if it is not configured to `emitEvents` on the `TurnToken`, which should be more properly called `emitStreamingUpdates` now that `AgentRunResponseEvent` exists.

Fixes #2423 

### Contribution Checklist

### CODE REVIEW

**Review by alliscode — APPROVED** [review]

File: dotnet/src/Microsoft.Agents.AI.Workflows/WorkflowHostAgent.cs [hunk 1]
```diff
@@ -138,11 +142,18 @@ IAsyncEnumerable<AgentResponseUpdate> RunCoreStreamingAsync(
...
                                                                       .ConfigureAwait(false)
                                                                       .WithCancellation(cancellationToken))
         {
+            merger.AddUpdate(update);
             yield return update;
         }
+
+        AgentResponse response = merger.ComputeMerged(workflowSession.LastResponseId!, this.Id, this.Name);
+        workflowSession.ChatHistoryProvider.AddMessages(workflowSession, response.Messages);
+        workflowSession.ChatHistoryProvider.UpdateBookmark(workflowSession);
```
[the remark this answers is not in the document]
lokitoth: Yes, but the correct fix is to remove the partial update persistence in the middle of `InvokeStageAsync` and instead rely on the canonical response generated in `RunAsync`/`RunStreamingAsync`.

dotnet/src/Microsoft.Agents.AI.Workflows/WorkflowChatHistoryProvider.cs
```diff
@@ -62,6 +62,12 @@ public IEnumerable<ChatMessage> GetFromBookmark(AgentSession session)
         }
     }
 
+    public IEnumerable<ChatMessage> GetAllMessages(AgentSession session)
+    {
+        var state = this._sessionState.GetOrInitializeState(session);
+        return state.Messages;
+    }
```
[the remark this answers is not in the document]
lokitoth: Yes, but using `.AsReadOnly()` is a better approach.

dotnet/src/Microsoft.Agents.AI.Workflows/WorkflowHostAgent.cs
chetantoshniwal, line 157: `RunCoreStreamingAsync` is an async iterator. If the caller breaks out of enumeration early (cancellation, `break`, exception), code after the `await foreach` loop is skipped—only pending `finally` blocks execute. This means `AddMessages` and `UpdateBookmark` are silently skipped on early termination. Worse, `WorkflowSession.LastCheckpoint` can already be advanced mid-stream (when `SuperStepCompletedEvent` fires), so the next call to `GetFromBookmark` will return messages the workflow has already consumed, causing a checkpoint/bookmark divergence. The old `InvokeStageAsync` `finally` block prevented this. Wrap the loop and post-loop code in a `try/finally`:

```suggestion
        WorkflowSession workflowSession = await this.UpdateSessionAsync(messages, session, cancellationToken).ConfigureAwait(false);
        MessageMerger merger = new();

        try
        {
            await foreach (AgentResponseUpdate update in workflowSession.InvokeStageAsync(cancellationToken)
                                                                          .ConfigureAwait(false)
                                                                          .WithCancellation(cancellationToken))
            {
                merger.AddUpdate(update);
                yield return update;
            }
        }
        finally
        {
            AgentResponse response = merger.ComputeMerged(workflowSession.LastResponseId!, this.Id, this.Name);
            workflowSession.ChatHistoryProvider.AddMessages(workflowSession, response.Messages);
            workflowSession.ChatHistoryProvider.UpdateBookmark(workflowSession);
        }
```

dotnet/tests/Microsoft.Agents.AI.Workflows.UnitTests/WorkflowHostSmokeTests.cs
```diff
@@ -112,4 +112,50 @@ public async Task Test_AsAgent_ErrorContentStreamedOutAsync(bool includeExceptio
...
+        AgentResponse response = await workflowAgent.RunAsync(session);
+
+        // Assert
+        WorkflowSession workflowSession = session.Should().BeOfType<WorkflowSession>().Subject;
+
+        ChatMessage[] responseMessages = response.Messages.ToArray();
+        ChatMessage[] sessionMessages = workflowSession.ChatHistoryProvider.GetAllMessages(workflowSession).ToArray();
+
+        // Since we never sent an incoming message, the expectation is that there should be nothing in the session
+        // except the response
```
chetantoshniwal, line 775: This test validates the non-streaming path (`RunAsync`). The streaming path in `RunCoreStreamingAsync` now has identical `AddMessages` + `UpdateBookmark` logic (new in this PR) but no test. Please add a streaming variant that calls `RunStreamingAsync`, consumes all updates, then asserts `GetAllMessages` matches the streamed content.

dotnet/tests/Microsoft.Agents.AI.Workflows.UnitTests/WorkflowHostSmokeTests.cs
```diff
@@ -112,4 +112,50 @@ public async Task Test_AsAgent_ErrorContentStreamedOutAsync(bool includeExceptio
...
+        AgentSession session = await workflowAgent.CreateSessionAsync();
+        AgentResponse response = await workflowAgent.RunAsync(session);
+
+        // Assert
+        WorkflowSession workflowSession = session.Should().BeOfType<WorkflowSession>().Subject;
+
+        ChatMessage[] responseMessages = response.Messages.ToArray();
+        ChatMessage[] sessionMessages = workflowSession.ChatHistoryProvider.GetAllMessages(workflowSession).ToArray();
+
+        // Since we never sent an incoming message, the expectation is that there should be nothing in the session
```
chetantoshniwal, line 774: Same gap: the handoff-workflow history test only covers `RunAsync`. A streaming variant is needed to cover the new `RunCoreStreamingAsync` bookmark/history logic.

dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/HandoffAgentExecutor.cs
```diff
@@ -250,7 +250,18 @@ await AddUpdateAsync(
...
+        AgentResponse agentResponse = updates.ToAgentResponse();
+
+        // Since there is no good way to configure the agent output behaviour due to how we add it to Handoff orchestration
+        // configurations, treat the emitEvents flag as simply determining whether to stream updates or to emit the whole response
+        // It would make little sense to avoid emitting any agent responses since this is only used in Orchestration workflows,
+        // which are Agent-only, and thus would do nothing.
+        if (message.TurnToken.EmitEvents is not true)
+        {
+            await context.YieldOutputAsync(agentResponse, cancellationToken).ConfigureAwait(false);
+        }
```
chetantoshniwal: `TurnToken.EmitEvents` is documented as controlling whether agent run events (streaming updates) are raised. Here it is repurposed to also decide whether to emit a full batched `AgentResponse`—a distinct concept. This will break callers that set `EmitEvents = false` to silence streaming events but do not expect a batched response injected into the output stream. A dedicated option on `HandoffAgentExecutorOptions` that controls batch-vs-streaming output mode would express the intent without overloading an unrelated flag's semantics.

chetantoshniwal (commented):
## Automated Code Review

**Reviewers:** 4 | **Confidence:** 86%

### ✗ Correctness
> This PR refactors chat history management by moving `AddMessages` and `UpdateBookmark` calls from per-update (inside `WorkflowSession.CreateUpdate` and `InvokeStageAsync`'s finally block) to post-completion (in `WorkflowHostAgent`'s `RunCoreAsync`/`RunCoreStreamingAsync`). It also adds a new output mode for `HandoffAgentExecutor` where the full `AgentResponse` is yielded when streaming updates are not requested. The changes are largely sound, but the streaming path in `RunCoreStreamingAsync` lacks a try/finally guard, meaning chat history and bookmark updates are silently skipped if the consumer disposes the async enumerator early or cancellation occurs—a behavioral regression from the old `finally` block in `InvokeStageAsync`.

### ✗ Security Reliability
> The diff moves chat-history persistence (AddMessages + UpdateBookmark) out of WorkflowSession.InvokeStageAsync (which had a try/finally) into the callers in WorkflowHostAgent. The non-streaming path (RunCoreAsync) is fine because the method is not an async iterator and fully consumes the enumerable. However, RunCoreStreamingAsync is an async iterator (it uses yield return), and the post-loop AddMessages/UpdateBookmark calls are placed after the await foreach loop without a try/finally. In C#, when a consumer disposes an async enumerator early (break, cancellation, exception), code after the yield-return loop is skipped—only pending finally blocks execute. This means on early termination the response messages are never written to history and the bookmark is never advanced, leaving the session in an inconsistent state. This is a reliability regression from the old code whose finally block guaranteed bookmark updates on any exit path.

### ✗ Test Coverage
> The diff moves chat-history bookkeeping out of WorkflowSession.CreateUpdate/InvokeStageAsync and into WorkflowHostAgent.RunCoreAsync and the new RunCoreStreamingAsync path. Two new smoke tests verify that outgoing messages appear in the chat history after RunAsync (non-streaming), and a new HandoffAgentExecutor test covers the emit-events logic. However, there is no corresponding streaming-path test for the chat-history/bookmark change in RunCoreStreamingAsync, which is the most significant behavioural addition in this PR. The removal of the try/finally around UpdateBookmark also changes error-path behaviour that is untested.

### ✗ Design Approach
> The diff moves chat history management from incremental per-update writes (inside `WorkflowSession.CreateUpdate`) to a deferred batch write after the full turn completes (in `WorkflowHostAgent`). This is a cleaner design in principle, but the streaming path has a correctness gap: history and bookmark updates only run after the `await foreach` loop, with no `try/finally` guard. If a consumer disposes the streaming `IAsyncEnumerable` early, those commits never execute—yet `LastCheckpoint` can be advanced mid-stream (via `SuperStepCompletedEvent`). This creates a divergence between workflow checkpoint state (advanced) and chat history bookmark (stale), potentially causing re-delivery of already-processed messages on the next turn. The old `finally` block in `InvokeStageAsync` existed precisely to prevent this. Separately, `HandoffAgentExecutor` repurposes `TurnToken.EmitEvents`—documented as controlling whether agent run events are raised—to also mean 'emit a batched full response instead of streaming updates'. The inline comment acknowledges this is a workaround due to a configuration limitation, which is itself a flag for a deeper design issue rather than a fix.

### Flagged Issues
- [ ] In `WorkflowHostAgent.RunCoreStreamingAsync`, `AddMessages` and `UpdateBookmark` are placed after the `await foreach` loop with no `try/finally`. Since this is an async iterator (uses `yield return`), if the consumer disposes the enumerator early (cancellation, `break`, or exception), post-loop code is skipped—only pending `finally` blocks execute. Meanwhile, `WorkflowSession.LastCheckpoint` can already be advanced mid-stream (by `SuperStepCompletedEvent`), so skipping `UpdateBookmark` leaves a divergence: the workflow has moved past a checkpoint but the history bookmark has not, causing re-delivery of already-consumed messages on the next turn. The old `finally` block in `InvokeStageAsync` prevented exactly this. A `try/finally` around the loop is needed to ensure chat history is updated on all exit paths.
- [ ] The new smoke tests (`Test_SingleAgent_AsAgent_OutgoingMessagesInHistoryAsync`, `Test_Handoffs_AsAgent_OutgoingMessagesInHistoryAsync`) both call `RunAsync`, exercising only the non-streaming path. The streaming path in `RunCoreStreamingAsync` contains identical new logic (`MessageMerger` accumulation, `AddMessages`, `UpdateBookmark`) but has zero direct test coverage.

### Suggestions
- Add streaming-path equivalents of the two new smoke tests that call `RunStreamingAsync`, consume all updates, and verify the session's chat history and bookmark are updated afterwards.
- The refactor also changes error-path behaviour: `UpdateBookmark` was previously in a `finally` block inside `InvokeStageAsync`, guaranteeing it ran even when the enumeration threw. It now sits after the `foreach` loop. Consider adding a test that verifies the intended behaviour when the inner workflow throws—is it intentional that the bookmark is not updated in that case?
- `HandoffAgentExecutor` repurposes `TurnToken.EmitEvents` (documented as controlling whether agent run events are raised) to also decide whether to emit a full batched `AgentResponse`—a distinct concept. This will break callers that set `EmitEvents = false` to silence streaming events but do not expect a batched response injected into the output stream. A dedicated option on `HandoffAgentExecutorOptions` that controls batch-vs-streaming output mode would express the intent without overloading an unrelated flag's semantics.

---

dotnet/src/Microsoft.Agents.AI.Workflows/WorkflowHostAgent.cs
lokitoth: Even with moving this logic, the history would be inconsistent - not because of bookmark alignment, but due to failing to process already "committed" (from the workflow agents' point of view) messages still in the iterator. Generally speaking, cancellation should be immediate and leave the state minimally modified. Otherwise we give users the illusion that they have access to the correct state, which is not true. If a user sends another message post-cancellation on that Session, the behaviour will not be as though there were no further messages past the ones that were processed.

If we want to get back to a consistent place, the correct fix is not to try to patch it up after the fact, but to use the checkpointing mechanism to restore to the pre-cancellation checkpoint - but that would be beyond the scope of this PR.

dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/HandoffAgentExecutor.cs
lokitoth: Implemented the same configuration mechanism as for other agents; the caveat is that since the user is not configuring individual agents directly, the settings apply to all.

dotnet/src/Microsoft.Agents.AI.Workflows/WorkflowHostAgent.cs
lokitoth: I am going to not make changes here for now. Can put a feature issue (for post v1) to make cancellation return to last checkpoint.

dotnet/src/Microsoft.Agents.AI.Workflows/HandoffsWorkflowBuilder.cs
```diff
@@ -47,16 +55,43 @@ in your conversation with the user.
...
 
+    /// <summary>
+    /// Sets a value indicating whether agent streaming update events should be emitted during execution.
+    /// If <see langword="null"/>, the value will be taken from the <see cref="TurnToken"/>
+    /// </summary>
+    /// <param name="emitAgentResponseUpdateEvents"></param>
+    /// <returns></returns>
+    public HandoffsWorkflowBuilder EmitAgentResponseUpdateEvents(bool emitAgentResponseUpdateEvents)
+    {
+        this._emitAgentResponseUpdateEvents = true;
```
SergeyMenshykh: Should not the `emitAgentResponseUpdateEvents` parameter to be assigned to the `_emitAgentResponseUpdateEvents` field? Same for the `EmitAgentResponseEvents` property below.

SergeyMenshykh (approved):

dotnet/src/Microsoft.Agents.AI.Workflows/HandoffsWorkflowBuilder.cs
lokitoth: Yeah, good catch!

dotnet/src/Microsoft.Agents.AI.Workflows/HandoffsWorkflowBuilder.cs
```diff
@@ -47,16 +55,43 @@ in your conversation with the user.
               """;
 
     /// <summary>
-    /// Sets additional instructions to provide to an agent that has handoffs about how and when to
-    /// perform them.
+    /// Sets instructions to provide to each agent that has handoffs about how and when perform them.
```
peibekwe: nit" missing "to" between perform and them.

peibekwe (approved):

dotnet/src/Microsoft.Agents.AI.Workflows/HandoffsWorkflowBuilder.cs
lokitoth: Good catch

peibekwe (approved):

PR #4763, opened 2026-03-18
Python: Support structuredContent in MCP tool results and fix sampling options type

### LINKED ISSUE #4625: Python: [Feature]: Support for MCP sampling tool
### Description

Agent Framework should advertise the sampling.tools capability, taking in sampling tools from MCP that enforce a data model through result_type

Example from DevUI run
Image

### Code Sample

```markdown
agent.py (client)

import os

from azure.identity.aio import DefaultAzureCredential
from agent_framework.azure import AzureOpenAIResponsesClient
from agent_framework import Agent, MCPStreamableHTTPTool
from agent_framework.devui import serve

load_dotenv()


credential = DefaultAzureCredential()

client = AzureOpenAIResponsesClient(
    credential=credential,
    project_endpoint=os.environ.get("AZURE_AI_PROJECT_ENDPOINT"),
    deployment_name=os.environ.get("AZURE_AI_MODEL_DEPLOYMENT_NAME"),
)

mcp_tools = MCPStreamableHTTPTool(
    name="recipe-server",
    url="http://localhost:8000/mcp",
    description="Local FastMCP server tools",
    client=client,
)

agent = Agent(
    client=client,
    name="assistant",
    instructions="""
        You are a helpful recipe assistant. Use the generate_recipe tool to create recipes for any dish the user asks about.
        Do not respond with your own knowledge of recipes, always use the generate_recipe tool to get the recipe information.
    """,
    tools=[mcp_tools],
)


if __name__ == "__main__":
    serve(entities=[agent], auto_open=True, port=8090)


main.py (mcp server)

from pydantic import BaseModel
from fastmcp import FastMCP, Context


class Ingredient(BaseModel):
    name: str
    quantity: str
    unit: str


class Recipe(BaseModel):
    name: str
    description: str
    servings: int
    prep_time_minutes: int
    cook_time_minutes: int
    ingredients: list[Ingredient]
    steps: list[str]


mcp = FastMCP("agent-framework-mcp-test")


@mcp.tool()
async def generate_recipe(dish: str, servings: int, ctx: Context) -> Recipe:
    """Generate a structured recipe for a given dish and number of servings using the client's LLM."""
    result = await ctx.sample(
        f"Create a recipe for {dish} that serves {servings}", result_type=Recipe
    )

    if not isinstance(result, Recipe):
        raise ValueError("Expected result to be of type Recipe")

    return result


if __name__ == "__main__":
    mcp.run(transport="streamable-http", host="0.0.0.0", port=8000)


OS: Windows 11
AF version: 1.0.0rc3
Language: Python
```

### Language/SDK

Python

eavanvalkenburg:
### Motivation and Context

MCP servers can return `structuredContent` in `CallToolResult`, but the agent framework silently dropped this data. Additionally, the sampling handler's `options` dict lacked proper typing, making it inconsistent with the rest of the codebase.

Fixes #4625

### Description

The `_parse_tool_result_from_mcp` function did not handle the `structuredContent` field on `CallToolResult`, so any structured data returned by MCP tools was lost. This fix appends `structuredContent` as a JSON-serialized `Content.from_text` entry to the parsed result list, and updates the sampling handler's `options` variable to use the typed `ChatOptions[None]` instead of a bare `dict[str, Any]`. Tests are added to verify that `structuredContent` is correctly included both alongside regular content and when it is the only content present.

### Contribution Checklist

---


python/packages/core/agent_framework/_mcp.py
```diff
@@ -192,6 +193,9 @@ def _parse_tool_result_from_mcp(
             case _:
                 result.append(Content.from_text(str(item)))
 
+    if mcp_type.structuredContent is not None:
```
eavanvalkenburg: This unconditionally appends `structuredContent` even when `content[]` already holds the same data serialised as text (FastMCP's default for `result_type`). The model then sees duplicate JSON. A safer design: only fall back to `structuredContent` when `content` produced no items (i.e. swap this block with the `if not result` guard below, or gate it on `not result`).

```suggestion
    if not result and mcp_type.structuredContent is not None:
        result.append(Content.from_text(json.dumps(mcp_type.structuredContent)))
```

eavanvalkenburg (commented):
## Automated Code Review

**Reviewers:** 4 | **Confidence:** 88%

### ✓ Correctness
> The changes are correct and well-tested. The `_parse_tool_result_from_mcp` function now handles `structuredContent` by serializing it to JSON text, placed correctly after the content loop and before the empty-result fallback. The `ChatOptions[None]` type annotation is a safe improvement — it's a local variable annotation in function scope, so it's never evaluated at runtime, and during type-checking the Generic branch provides proper type safety. Both test cases cover the relevant scenarios (structured content alone, and combined with regular content). No bugs, no missing error handling, no incorrect assumptions.

### ✓ Security Reliability
> The changes are minimal and low-risk. The `structuredContent` handling uses `json.dumps` on a value typed as `dict[str, Any] | None` by the MCP SDK, which is safe. The `ChatOptions[None]` type annotation is a purely cosmetic improvement. No injection risks, resource leaks, or unhandled failure modes introduced. One minor reliability concern: `json.dumps` can raise `TypeError` on non-serializable values if the MCP server returns unexpected types within the dict, but this is an edge case given the MCP protocol guarantees JSON-compatible content.

### ✓ Test Coverage
> The two new tests for `structuredContent` handling in `_parse_tool_result_from_mcp` are well-written and cover the key scenarios (structured content alongside regular content, and structured content alone). However, the `ChatOptions[None]` type annotation change in `sampling_callback` has no corresponding test coverage. A minor missing edge case is that `structuredContent=None` (explicit) isn't tested, though it's implicitly covered by all existing tests that omit it.

### ✗ Design Approach
> The diff makes two changes: (1) appends `structuredContent` as a JSON text `Content` item after regular content in `_parse_tool_result_from_mcp`, and (2) tightens the type annotation of `options` from `dict[str, Any]` to `ChatOptions[None]`. The structured-content parsing is a reasonable pragmatic approach given the framework's limited `Content` type set, but has a meaningful design flaw: `structuredContent` is appended unconditionally even when the regular `content[]` array already carries the same data serialised as text (which is FastMCP's default behaviour when `result_type` is used). The LLM will then see the same JSON payload twice, one in human-readable position and once as the appended structured blob. The `ChatOptions[None]` annotation is type-correct but purely cosmetic — `None` as the type parameter explicitly means 'no response model', so it does not fulfil the stated goal from the prior discussion of forwarding the desired output schema to `get_response`.

### Flagged Issues
- [ ] When a FastMCP server uses `result_type`, it typically serialises the structured value into both `content[0]` (as a text item) and `structuredContent`. The new code appends `structuredContent` unconditionally after all `content` items, causing the LLM to receive the same JSON payload twice. The fix should either (a) only emit `structuredContent` when `content` is empty (replacing the existing 'null' fallback), or (b) at minimum skip `content` items that are identical to the JSON-serialised structured content. Emitting both by default risks confusing the model and inflating context.

### Suggestions
- Consider whether `structuredContent` should be distinguished from regular text content (e.g., via a metadata flag or a different Content type) so downstream consumers can programmatically identify structured output without having to speculatively parse JSON.
- Consider adding a test for `structuredContent` with nested complex types (e.g., nested dicts/lists) to verify `json.dumps` handles deeper structures correctly.
- The `ChatOptions[None]` annotation is accurate but the `None` type parameter signals 'no structured output type', so it does not advance the goal (raised in prior discussion) of passing the server's expected schema as `response_format` to `get_response`. If that goal is deferred intentionally, add a TODO comment so it is not silently dropped. Existing sampling callback tests already exercise this code path, but a type-checking CI step (mypy/pyright) would catch any incompatibility.

---

python/packages/core/agent_framework/_mcp.py
```diff
@@ -192,6 +196,9 @@ def _parse_tool_result_from_mcp(
             case _:
                 result.append(Content.from_text(str(item)))
 
+    if not result and mcp_type.structuredContent is not None:
+        result.append(Content.from_text(json.dumps(mcp_type.structuredContent, default=str)))
```
eavanvalkenburg: `default=str` silently converts non-JSON-serializable values to their `str()` representation instead of raising `TypeError`. Since `structuredContent` is defined as `dict[str, Any]` and comes from MCP SDK JSON deserialization, it should always be JSON-round-trippable. Consider omitting `default=str` to surface upstream bugs sooner, or keeping it with a logged warning when the fallback fires.

```suggestion
        result.append(Content.from_text(json.dumps(mcp_type.structuredContent)))
```

python/packages/core/tests/core/test_mcp.py
```diff
@@ -246,6 +247,57 @@ def test_parse_tool_result_from_mcp_blob_plain_base64():
...
+    mcp_result = types.CallToolResult(
+        content=[],
+        structuredContent=structured,
+    )
+    result = _parse_tool_result_from_mcp(mcp_result)
+
+    assert isinstance(result, list)
+    assert len(result) == 1
+    assert result[0].type == "text"
+    assert json.loads(result[0].text) == structured
```
eavanvalkenburg: This test exercises nested dicts/lists but doesn't test the `default=str` parameter added to `json.dumps`. Consider adding a test where `structuredContent` contains a non-JSON-serializable value (e.g. a `datetime` object) to verify the fallback produces valid JSON text rather than raising `TypeError`.

python/packages/core/agent_framework/_mcp.py
```diff
@@ -732,12 +745,31 @@ async def sampling_callback(
         messages: list[Message] = []
         for msg in params.messages:
             messages.append(_parse_message_from_mcp(msg))
+
+        # TODO(Copilot): forward server's expected result schema as response_format
+        # when result_type support is added to ChatOptions.
+        options: ChatOptions[None] = {}
+        if params.systemPrompt is not None:
```
eavanvalkenburg, line 750: The change from `if params.systemPrompt:` to `if params.systemPrompt is not None:` now forwards an empty string `""` as `instructions`. The existing test only uses a non-empty string. Add a test with `params.systemPrompt = ""` asserting `options["instructions"] == ""`.

python/packages/core/agent_framework/_mcp.py
```diff
@@ -732,12 +745,31 @@ async def sampling_callback(
...
+            options["instructions"] = params.systemPrompt
+        if params.tools is not None:
```
eavanvalkenburg, line 752: Similarly, the change from `if params.tools:` to `if params.tools is not None:` now forwards an empty list `[]` as `tools`. Add a test with `params.tools = []` asserting `options["tools"] == []`.

python/packages/core/agent_framework/_mcp.py
```diff
@@ -732,12 +745,31 @@ async def sampling_callback(
         messages: list[Message] = []
         for msg in params.messages:
             messages.append(_parse_message_from_mcp(msg))
+
+        # TODO(Copilot): forward server's expected result schema as response_format
```
eavanvalkenburg: The TODO correctly defers `result_type` forwarding, but since `_parse_tool_result_from_mcp` returns `list[Content]` with no structured-data channel, retrofitting typed result support will require a parallel code path or richer return type. Consider clarifying this in the TODO so the next author doesn't attempt to bolt it onto the existing `list[Content]` signature.

```suggestion
        # TODO(Copilot): forward server's expected result schema as response_format
        # when result_type support is added to ChatOptions. Note: this will also
        # require a parallel parse path (or richer return type) in _parse_tool_result_from_mcp
        # since structuredContent is currently discarded when content items are present.
```

eavanvalkenburg (commented):
## Automated Code Review

**Reviewers:** 4 | **Confidence:** 88%

### ✓ Correctness
> The changes are logically correct. The structuredContent fallback change correctly prioritizes content items over structuredContent (avoiding duplication), matching the MCP spec intent. The truthiness-to-None checks for `params.systemPrompt` and `params.tools` are a genuine bugfix: since both default to `None`, the old `if params.systemPrompt:` would silently drop an empty string system prompt, and `if params.tools:` would drop an empty tools list — both valid values a server might send. The `default=str` addition to `json.dumps` is a reasonable defensive measure. Tests are updated consistently with the behavior changes.

### ✓ Security Reliability
> The changes are small and well-scoped. The `structuredContent` fallback logic and the `is not None` truthiness fixes are correct improvements. The `default=str` in `json.dumps` silently coerces arbitrary types to strings, which could mask data-corruption bugs rather than failing loudly, but this is a low-severity concern since `structuredContent` is a dict from the MCP SDK. The error message in the sampling callback at line 777 still exposes raw exception details (`f"Failed to get chat message content: {ex}"`) which is returned as an MCP `ErrorData`—this could leak internal details to an untrusted MCP server, but this is pre-existing, not introduced by this diff. No injection, deserialization, secret, or resource-leak issues found.

### ✗ Test Coverage
> The diff modifies two behaviors: (1) structuredContent is now a fallback (only used when no content items exist) with `default=str` added to json.dumps, and (2) the sampling callback now uses `is not None` checks instead of truthiness for systemPrompt and tools. The existing tests for structuredContent are updated and a new nested-object test is added, which is good. However, there are three test coverage gaps: no test for the `default=str` behavior on json.dumps (e.g., structuredContent containing a non-serializable type like datetime), no test for an empty-string systemPrompt being forwarded (the semantic change from truthiness to `is not None`), and no test for an empty tools list being forwarded (same truthiness-to-identity change).

### ✓ Design Approach
> The diff makes three independent improvements: (1) making `structuredContent` a fallback-only when `content` items are absent, (2) adding `default=str` to the `json.dumps` call, and (3) tightening `systemPrompt`/`tools` checks from truthiness to explicit `None` comparisons. All three changes are directionally correct. The one design concern worth calling out is that permanently discarding `structuredContent` when any `content` items are present means structured data is irreversibly lost at the parsing boundary — even when the structured payload contains richer information than the accompanying human-readable text. The function signature `-> list[Content]` has no way to carry structured data forward, so the TODO about `result_type`/`response_format` will require a different entry point (e.g., a parallel `_parse_tool_result_from_mcp_typed` or a richer return type) rather than being retrofittable into this function. That said, the current fix is correct for the immediate LLM-consumption use case, and the limitation is explicitly acknowledged.

### Flagged Issues
- [ ] No tests cover the behavioral changes from truthiness to `is not None` checks. An empty string `params.systemPrompt = ""` is now forwarded as `instructions`, and an empty list `params.tools = []` is now forwarded as `tools`, but no tests verify these edge cases.

### Suggestions
- Consider whether `default=str` in `json.dumps(mcp_type.structuredContent, default=str)` is the right tradeoff. Since `structuredContent` is typed as `dict[str, Any]` by the MCP SDK and comes from JSON deserialization, it should always be JSON-serializable. Omitting `default=str` would fail fast on unexpected types; alternatively, keep it but log a warning when the fallback is actually invoked.
- Pre-existing: the sampling callback at line 777 returns `f"Failed to get chat message content: {ex}"` as an MCP `ErrorData` message. If the MCP server is untrusted, this could leak internal exception details (file paths, credentials in connection strings). Consider sanitizing the error message.
- Add a test for `structuredContent` containing a non-JSON-serializable value (e.g. a `datetime` object) to exercise the `default=str` fallback in `json.dumps`.
- Add tests for the empty-value edge cases: `params.systemPrompt = ""` should set `options["instructions"] = ""`, and `params.tools = []` should set `options["tools"] = []`.
- The `structuredContent` is permanently discarded when `content` is non-empty, making it impossible to add typed result support (the `result_type`/`response_format` TODO) through `_parse_tool_result_from_mcp` without a breaking signature change. Consider noting this constraint in the TODO comment so the next author knows a parallel code path or richer return type will be needed.
- An MCP server sending `tools=[]` will now produce `options["tools"] = []` rather than omitting it. Downstream LLM clients likely filter via truthiness, but verify no middleware path treats `options.get("tools") is not None` as semantically meaningful.

---

python/packages/core/agent_framework/_mcp.py
eavanvalkenburg: Fixed in `9d30e44e`.

python/packages/core/agent_framework/_mcp.py
```diff
@@ -192,6 +193,9 @@ def _parse_tool_result_from_mcp(
             case _:
                 result.append(Content.from_text(str(item)))
 
+    if mcp_type.structuredContent is not None:
+        result.append(Content.from_text(json.dumps(mcp_type.structuredContent)))
```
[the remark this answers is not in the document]
eavanvalkenburg: Fixed in `9d30e44e`.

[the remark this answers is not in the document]
eavanvalkenburg: Fixed in `9d30e44e`.

python/packages/core/agent_framework/_mcp.py
```diff
@@ -732,12 +742,29 @@ async def sampling_callback(
...
+            options["tools"] = [
+                FunctionTool(
+                    name=tool.name,
+                    description=tool.description or "",
+                    input_model=tool.inputSchema,
+                )
+                for tool in params.tools
+            ]
+        if params.toolChoice is not None and params.toolChoice.mode is not None:
+            options["tool_choice"] = params.toolChoice.mode
```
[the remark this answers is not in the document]
eavanvalkenburg: Fixed in `9d30e44e`.

python/packages/core/tests/core/test_mcp.py
```diff
@@ -246,6 +247,38 @@ def test_parse_tool_result_from_mcp_blob_plain_base64():
...
+        structuredContent=structured,
+    )
+    result = _parse_tool_result_from_mcp(mcp_result)
+
+    assert isinstance(result, list)
+    assert len(result) == 2
+    assert result[0].type == "text"
+    assert result[0].text == "Here is a recipe"
+    assert result[1].type == "text"
+    assert result[1].text == json.dumps(structured)
```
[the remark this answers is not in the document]
eavanvalkenburg: Fixed in `9d30e44e`.

python/packages/core/tests/core/test_mcp.py
```diff
@@ -246,6 +247,38 @@ def test_parse_tool_result_from_mcp_blob_plain_base64():
...
+    mcp_result = types.CallToolResult(
+        content=[],
+        structuredContent=structured,
+    )
+    result = _parse_tool_result_from_mcp(mcp_result)
+
+    assert isinstance(result, list)
+    assert len(result) == 1
+    assert result[0].type == "text"
+    assert result[0].text == json.dumps(structured)
```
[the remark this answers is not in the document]
eavanvalkenburg: Fixed in `9d30e44e`.

python/packages/core/agent_framework/_mcp.py
eavanvalkenburg: Fixed in `9d30e44e`.

python/packages/core/tests/core/test_mcp.py
eavanvalkenburg: Fixed in `9d30e44e`.

python/packages/core/agent_framework/_mcp.py
eavanvalkenburg: Fixed in `9d30e44e`.

python/packages/core/agent_framework/_mcp.py
eavanvalkenburg: Fixed in `9d30e44e`.

python/packages/core/agent_framework/_mcp.py
eavanvalkenburg: Fixed in `9d30e44e`.

python/packages/core/tests/core/test_mcp.py
```diff
@@ -1562,12 +1627,15 @@ async def test_mcp_tool_sampling_callback_chat_client_exception():
...
+    params.systemPrompt = None
+    params.tools = None
+    params.toolChoice = None
 
     result = await tool.sampling_callback(Mock(), params)
 
     assert isinstance(result, types.ErrorData)
     assert result.code == types.INTERNAL_ERROR
-    assert "Failed to get chat message content: Chat client error" in result.message
+    assert "Failed to get chat message content" in result.message
```
eavanvalkenburg, line 1707: This test confirms the error message no longer leaks exception details, but it should also assert that the exception *is* logged at DEBUG level (the new behavior in `_mcp.py:778`). Use `caplog` or patch the logger to verify `"Sampling callback error"` appears in the debug output.

```suggestion
    assert "Failed to get chat message content" in result.message
    assert "Chat client error" not in result.message  # exception detail must not leak


async def test_mcp_tool_sampling_callback_exception_is_debug_logged(caplog):
    """Test that sampling callback logs exception details at DEBUG level."""
    import logging

    tool = MCPStdioTool(name="test_tool", command="python")

    mock_chat_client = AsyncMock()
    mock_chat_client.get_response.side_effect = RuntimeError("Chat client error")
    tool.client = mock_chat_client

    params = Mock()
    mock_message = Mock()
    mock_message.role = "user"
    mock_message.content = Mock()
    mock_message.content.text = "Test question"
    params.messages = [mock_message]
    params.temperature = None
    params.maxTokens = None
    params.stopSequences = None
    params.systemPrompt = None
    params.tools = None
    params.toolChoice = None

    with caplog.at_level(logging.DEBUG, logger="agent_framework._mcp"):
        result = await tool.sampling_callback(Mock(), params)

    assert isinstance(result, types.ErrorData)
    assert any("Sampling callback error" in record.message for record in caplog.records)
```

python/packages/core/agent_framework/_mcp.py
```diff
@@ -732,17 +745,40 @@ async def sampling_callback(
...
         try:
             response = await self.client.get_response(
                 messages,
                 temperature=params.temperature,
                 max_tokens=params.maxTokens,
                 stop=params.stopSequences,
+                options=options or None,
             )
         except Exception as ex:
+            logger.debug("Sampling callback error: %s", ex, exc_info=True)
```
eavanvalkenburg, line 776: Logging the exception at DEBUG means it will be invisible in most production deployments. Since a sampling callback failure is a real operational error, WARNING or ERROR would be more appropriate to ensure operators are alerted.

```suggestion
            logger.warning("Sampling callback error: %s", ex, exc_info=True)
```

eavanvalkenburg (commented):
## Automated Code Review

### ✓ Correctness
> This is a small, correct change that improves error handling hygiene in the MCP sampling callback. It removes the raw exception message from the error response returned to the MCP server (preventing potential information leakage), and instead logs it locally at DEBUG level with full traceback via exc_info=True. The test is updated accordingly. The logger is already defined in the module. No correctness issues found.

### ✓ Security Reliability
> This is a good security hardening change. It removes the internal exception message (which could leak implementation details like stack traces, internal service names, or credential fragments) from the error response sent back to the MCP server, and instead logs it server-side at debug level. The change is minimal, correct, and the test is properly updated. No issues found.

### ✗ Test Coverage
> The diff strips exception details from the user-facing error message (good security practice) and adds debug logging with `exc_info=True`. The existing test `test_mcp_tool_sampling_callback_chat_client_exception` is updated to match the new message but does not verify that the debug log is actually emitted. Since the logging is the new behavior introduced by this change, it should be tested.

### ✓ Design Approach
> The change strips the exception message from the user-facing error string and logs it at DEBUG level instead. This is a reasonable security/information-leakage improvement — internal exception details (which might include credentials, tokens, or internal hostnames from the chat client) should not be forwarded to the MCP server/caller in the error message. Logging at DEBUG with exc_info=True preserves full diagnostics for operators. The test is updated consistently. No design-level concerns: the fix is narrowly scoped, doesn't mask a deeper root cause, and doesn't extend a questionable pattern.

### Flagged Issues
- [ ] No test verifies the new `logger.debug("Sampling callback error: ...")` call. Since the purpose of this change is to move exception details from the user-facing message into the debug log, a test should confirm the debug log is actually produced (e.g., via `caplog` or by patching the logger).

### Suggestions
- Consider logging at WARNING or ERROR level rather than DEBUG, since a sampling callback failure is an operational error that operators need to notice in production logs where DEBUG is typically suppressed.

---

python/packages/core/tests/core/test_mcp.py
eavanvalkenburg: Fixed in `c1a742e5`.

python/packages/core/agent_framework/_mcp.py
eavanvalkenburg: Fixed in `c1a742e5`.

python/packages/core/agent_framework/_mcp.py
```diff
@@ -732,17 +745,40 @@ async def sampling_callback(
         messages: list[Message] = []
         for msg in params.messages:
             messages.append(_parse_message_from_mcp(msg))
+
+        # TODO(Copilot): forward server's expected result schema as response_format
```
eavanvalkenburg: do this now

python/packages/core/agent_framework/_mcp.py
```diff
@@ -732,17 +745,40 @@ async def sampling_callback(
...
                 temperature=params.temperature,
                 max_tokens=params.maxTokens,
                 stop=params.stopSequences,
+                options=options or None,
             )
         except Exception as ex:
+            logger.debug("Sampling callback error: %s", ex, exc_info=True)
             return types.ErrorData(
                 code=types.INTERNAL_ERROR,
-                message=f"Failed to get chat message content: {ex}",
```
eavanvalkenburg, line 746: why is this removed, that seems useful info

python/packages/core/agent_framework/_mcp.py
```diff
@@ -192,6 +196,9 @@ def _parse_tool_result_from_mcp(
             case _:
                 result.append(Content.from_text(str(item)))
 
+    if not result and mcp_type.structuredContent is not None:
+        result.append(Content.from_text(json.dumps(mcp_type.structuredContent)))
```
eavanvalkenburg: let's set the raw object as a `additional_property` as well

python/packages/core/agent_framework/_mcp.py
```diff
@@ -732,17 +745,40 @@ async def sampling_callback(
...
+                )
+                for tool in params.tools
+            ]
+        if params.toolChoice is not None and params.toolChoice.mode is not None:
+            options["tool_choice"] = params.toolChoice.mode
+
         try:
             response = await self.client.get_response(
                 messages,
                 temperature=params.temperature,
```
eavanvalkenburg: all of these parameters should be in `options`, they are no longer part of the contract and wouldn't be merged. Also temperature should not be set unless it is passed since that param is not supported by all models

python/packages/core/agent_framework/_mcp.py
eavanvalkenburg: Fixed in `1ff7f1f7`.

python/packages/core/agent_framework/_mcp.py
eavanvalkenburg: Fixed in `1ff7f1f7`.

python/packages/core/agent_framework/_mcp.py
eavanvalkenburg: Fixed in `1ff7f1f7`.

python/packages/core/agent_framework/_mcp.py
eavanvalkenburg: Fixed in `1ff7f1f7`.

markwallace-microsoft:
<a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/README.md">Python Test Coverage</a><details><summary>Python Test Coverage Report • </summary><table><tr><th>File</th><th>Stmts</th><th>Miss</th><th>Cover</th><th>Missing</th></tr><tbody><tr><td colspan="5"><b>packages/core/agent_framework</b></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/core/agent_framework/_mcp.py">_mcp.py</a></td><td>575</td><td>29</td><td>94%</td><td><a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/core/agent_framework/_mcp.py#L273">273</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/core/agent_framework/_mcp.py#L333-L334">333&ndash;334</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/core/agent_framework/_mcp.py#L463">463</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/core/agent_framework/_mcp.py#L525">525</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/core/agent_framework/_mcp.py#L538">538</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/core/agent_framework/_mcp.py#L540-L543">540&ndash;543</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/core/agent_framework/_mcp.py#L562-L563">562&ndash;563</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/core/agent_framework/_mcp.py#L576-L579">576&ndash;579</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/core/agent_framework/_mcp.py#L581-L582">581&ndash;582</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/core/agent_framework/_mcp.py#L586">586</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/core/agent_framework/_mcp.py#L649-L651">649&ndash;651</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/core/agent_framework/_mcp.py#L998">998</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/core/agent_framework/_mcp.py#L1111">1111</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/core/agent_framework/_mcp.py#L1163-L1164">1163&ndash;1164</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/core/agent_framework/_mcp.py#L1167">1167</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/core/agent_framework/_mcp.py#L1185">1185</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/core/agent_framework/_mcp.py#L1590">1590</a></td></tr><tr><td colspan="5"><b>packages/openai/agent_framework_openai</b></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/openai/agent_framework_openai/_shared.py">_shared.py</a></td><td>188</td><td>27</td><td>85%</td><td><a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/openai/agent_framework_openai/_shared.py#L65">65</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/openai/agent_framework_openai/_shared.py#L71-L74">71&ndash;74</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/openai/agent_framework_openai/_shared.py#L127-L128">127&ndash;128</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/openai/agent_framework_openai/_shared.py#L138">138</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/openai/agent_framework_openai/_shared.py#L194">194</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/openai/agent_framework_openai/_shared.py#L236">236</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/openai/agent_framework_openai/_shared.py#L288">288</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/openai/agent_framework_openai/_shared.py#L290-L291">290&ndash;291</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/openai/agent_framework_openai/_shared.py#L307">307</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/openai/agent_framework_openai/_shared.py#L309">309</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/openai/agent_framework_openai/_shared.py#L311">311</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/openai/agent_framework_openai/_shared.py#L318">318</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/openai/agent_framework_openai/_shared.py#L320">320</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/openai/agent_framework_openai/_shared.py#L333">333</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/openai/agent_framework_openai/_shared.py#L414">414</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/openai/agent_framework_openai/_shared.py#L419">419</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/openai/agent_framework_openai/_shared.py#L438">438</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/openai/agent_framework_openai/_shared.py#L497-L500">497&ndash;500</a>, <a href="https://github.com/microsoft/agent-framework/blob/8cdae40e3d8efc58d5c94c8f64f5b62bb5747fea/packages/openai/agent_framework_openai/_shared.py#L502">502</a></td></tr><tr><td><b>TOTAL</b></td><td><b>27980</b></td><td><b>3412</b></td><td><b>87%</b></td><td>&nbsp;</td></tr></tbody></table></details>

## Python Unit Test Overview
| 5427 | 20 :zzz: | 0 :x: | 0 :fire: | 1m 29s :stopwatch: |

python/packages/core/tests/core/test_mcp.py
```diff
@@ -1613,6 +1686,323 @@ async def test_mcp_tool_sampling_callback_no_valid_content():
...
+    result = await tool.sampling_callback(Mock(), params)
+
+    assert isinstance(result, types.CreateMessageResult)
+    call_kwargs = mock_chat_client.get_response.call_args
+    options = call_kwargs.kwargs.get("options") or {}
+    assert "temperature" not in options
+    assert options.get("max_tokens") == 100
+    assert "stop" not in options
+    """Test connect() passes sampling_capabilities to ClientSession when client is set."""
+    tool = MCPStdioTool(name="test", command="test-command", load_tools=False, load_prompts=False)
```
chetantoshniwal, line 2100: Bug: This docstring and the subsequent connect-test code block are inside `test_mcp_tool_sampling_callback_omits_temperature_when_none`. A separate `async def test_connect_sampling_capabilities_with_client():` function definition is missing here. As-is, the connect test is not discoverable by pytest as its own test case and the temperature-omission test is corrupted.

```suggestion
    assert "stop" not in options


async def test_connect_sampling_capabilities_with_client():
    """Test connect() passes sampling_capabilities to ClientSession when client is set."""
```

python/packages/core/agent_framework/_mcp.py
```diff
@@ -192,6 +196,14 @@ def _parse_tool_result_from_mcp(
...
                 result.append(Content.from_text(str(item)))
 
+    if not result and mcp_type.structuredContent is not None:
+        result.append(
+            Content.from_text(
+                json.dumps(mcp_type.structuredContent),
+                additional_properties={"structured_content": mcp_type.structuredContent},
+            )
+        )
+
```
chetantoshniwal: If `structuredContent` contains values that are not JSON-serializable (e.g., custom objects injected by a faulty MCP server), `json.dumps` raises `TypeError` which will propagate unhandled. Consider wrapping in try/except to fall back gracefully.

```suggestion
    if not result and mcp_type.structuredContent is not None:
        try:
            text = json.dumps(mcp_type.structuredContent)
        except (TypeError, ValueError):
            text = str(mcp_type.structuredContent)
        result.append(
            Content.from_text(
                text,
                additional_properties={"structured_content": mcp_type.structuredContent},
            )
        )
```

python/packages/core/agent_framework/_mcp.py
```diff
@@ -732,14 +750,35 @@ async def sampling_callback(
...
+                    input_model=tool.inputSchema,
+                )
+                for tool in params.tools
+            ]
+        if params.toolChoice is not None and params.toolChoice.mode is not None:
+            options["tool_choice"] = params.toolChoice.mode
+
+        if params.temperature is not None:
+            options["temperature"] = params.temperature
+        options["max_tokens"] = params.maxTokens
```
chetantoshniwal, line 766: `options["max_tokens"]` is set unconditionally, so options always has at least one key and `options or None` on line 778 never evaluates to `None`. It also passes `max_tokens=None` to backends that may treat absence and explicit-None differently. Guard it like temperature.

```suggestion
        if params.maxTokens is not None:
            options["max_tokens"] = params.maxTokens
```

chetantoshniwal (commented):
## Automated Code Review

**Reviewers:** 4 | **Confidence:** 81%

### ✓ Correctness
> 

### ✗ Security Reliability
> The diff adds structured content fallback handling, sampling capability negotiation, and richer option forwarding in the MCP sampling callback. There is one blocking reliability issue in the test file: the test for `test_connect_sampling_capabilities_with_client` is missing its `async def` function definition and its body is accidentally embedded inside `test_mcp_tool_sampling_callback_omits_temperature_when_none`, making it an unreliable merged test that won't be discoverable as a separate test case. In the production code, the `json.dumps` call on `structuredContent` can raise an unhandled `TypeError` for non-serializable payloads, which would propagate as an unexpected exception to the caller instead of being caught and returned gracefully. Additionally, the error logging in the sampling callback includes `exc_info=True`, which is fine at DEBUG level but the error message returned to the MCP server still includes the raw exception string, potentially leaking internal details across the trust boundary.

### ✗ Test Coverage
> The diff adds good test coverage for the new `structuredContent` fallback and the expanded `sampling_callback` options forwarding. However, there is a critical structural bug: the test for `test_connect_sampling_capabilities_with_client` is missing its `async def` function definition, causing its body to be appended inside the unrelated `test_mcp_tool_sampling_callback_omits_temperature_when_none` test. This means the sampling-capabilities-with-client scenario will not run as a discoverable test, and it will corrupt the temperature-omission test. Additionally, the `options or None` expression on the `get_response` call is effectively dead code because `options` always contains at least `max_tokens`, making the dict truthy; a test verifying the 'no options' path would be misleading.

### ✗ Design Approach
> The refactoring of `sampling_callback` to use a `ChatOptions` dict is the right direction, but contains one inconsistency that makes a guard expression permanently dead: `options["max_tokens"] = params.maxTokens` is unconditional, always inserting a key into the dict (even with a `None` value), which means `options or None` on the `get_response` call always evaluates to `options` — the guard never fires. Beyond the dead guard, passing an explicit `None` for `max_tokens` is semantically different from omitting the key, and some backend implementations may treat them differently. The `structuredContent` fallback and `sampling_capabilities` wiring are clean and well-scoped. However, there is a test-file bug: the `test_connect_with_sampling_capabilities` test function is missing its `async def` declaration — its body (starting with a bare docstring) is appended as dead code inside `test_mcp_tool_sampling_callback_omits_temperature_when_none`, so that test never runs.

### Flagged Issues
- [ ] python/packages/core/tests/core/test_mcp.py: The test for `connect() with sampling_capabilities` is missing its `async def` function definition. Its body is accidentally embedded at the end of `test_mcp_tool_sampling_callback_omits_temperature_when_none`, so (1) pytest won't discover it as a separate test, (2) it runs under a misleading test name, and (3) the temperature-omission test is corrupted by the appended code.
- [ ] python/packages/core/agent_framework/_mcp.py: In `sampling_callback`, `options["max_tokens"] = params.maxTokens` is always executed unconditionally (even when `params.maxTokens is None`), inserting a `None`-valued key into the dict. This makes the `options or None` guard on the `get_response` call permanently dead (a dict with any key is truthy) and silently propagates `max_tokens=None` to the underlying client. Guard it the same way temperature is guarded: `if params.maxTokens is not None: options["max_tokens"] = params.maxTokens`.

### Suggestions
- In `_parse_tool_result_from_mcp`, wrap the `json.dumps(mcp_type.structuredContent)` call in a try/except to handle non-serializable payloads gracefully (e.g., fall back to `str()`) rather than letting `TypeError` propagate as an unhandled exception to the caller.
- Consider sanitizing the exception message in the sampling callback error response (`message=f"Failed to get chat message content: {ex}"`) to avoid leaking internal details (stack traces, file paths, config values) across the MCP trust boundary back to the server.
- Add a test for the edge case where `params.toolChoice` is set but `params.toolChoice.mode` is `None`, to verify `tool_choice` is not added to options.
- After fixing the max_tokens guard, consider simplifying or dropping the `options or None` expression entirely, depending on the client contract for handling an empty dict.

---

python/packages/core/tests/core/test_mcp.py
eavanvalkenburg: Fixed in `73ec3378`.

python/packages/core/agent_framework/_mcp.py
eavanvalkenburg: Fixed in `73ec3378`.

python/packages/core/agent_framework/_mcp.py
eavanvalkenburg: Fixed in `73ec3378`.

chetantoshniwal (approved):

giles17 (approved):

moonbox3 (approved):

PR #4768, opened 2026-03-18
.NET: Expose workflows as MCP tools when hosting on Azure functions

kshyju:
### Motivation and Context

This PR adds support for exposing durable workflows as [MCP (Model Context Protocol)](https://modelcontextprotocol.io/) tools in `Microsoft.Agents.AI.Hosting.AzureFunctions`. This  mirrors the existing agent MCP tool support (`AddAIAgent(enableMcpToolTrigger: true)`) but for workflows.

   When enabled, the Azure Functions host automatically registers an MCP tool trigger for the workflow at the `/runtime/webhooks/mcp` endpoint, allowing MCP-compatible clients to
  discover and invoke the workflow as a tool.

   ### Usage

   ```csharp
   builder.ConfigureDurableWorkflows(workflows =>
   {
       workflows.AddWorkflow(myWorkflow, exposeStatusEndpoint: false, exposeMcpToolTrigger: true);
   });
   ```

   ### New Public APIs

   | Type | Member | Description |
   |------|--------|-------------|
   | `DurableWorkflowOptionsExtensions` | `AddWorkflow(DurableWorkflowOptions, Workflow, bool exposeStatusEndpoint, bool exposeMcpToolTrigger)` | New overload that configures MCP tool trigger for the workflow |

   ### Implementation details

   - **`FunctionsDurableOptions`** — Tracks which workflows have MCP tool triggers enabled (internal, mirrors status endpoint pattern)
   - **`FunctionMetadataFactory.CreateWorkflowMcpToolTrigger`** — Generates function metadata using `System.Text.Json.Nodes.JsonObject` (AOT-safe, no manual string escaping)
   - **`BuiltInFunctions.RunWorkflowMcpToolAsync`** — Schedules an orchestration, waits for completion, validates `RuntimeStatus`, and returns the `DurableWorkflowResult.Result`
   - **`BuiltInFunctionExecutor`** — Routes the new entry point to `RunWorkflowMcpToolAsync`
   - **`DurableWorkflowsFunctionMetadataTransformer`** — Registers the MCP tool trigger when enabled
   - **`FunctionsApplicationBuilderExtensions`** — Adds the new entry point to the middleware filter

   ### Sample app

   Added `04_WorkflowMcpTool` under `dotnet/samples/04-hosting/DurableWorkflows/AzureFunctions/` with two workflows:
   - **Translate** — Simple string-in/string-out workflow (TranslateText → FormatOutput)
   - **OrderLookup** — Returns a POCO (LookupOrder → EnrichOrder → OrderSummary)

   ### Tests

   - **Unit tests** (`FunctionMetadataFactoryTests.cs`) — 7 tests covering all factory methods, including JSON validity validation via `JsonDocument.Parse()`
   - **Integration test** (`WorkflowSamplesValidation.cs`) — Connects via MCP client (`HttpClientTransport`), lists tools, invokes both workflows, and asserts results

### Contribution Checklist


dotnet/src/Microsoft.Agents.AI.Hosting.AzureFunctions/BuiltInFunctions.cs
```diff
@@ -378,6 +379,54 @@ await agentProxy.RunAsync(
...
+            cancellation: functionContext.CancellationToken);
+
+        if (metadata is null)
+        {
+            throw new InvalidOperationException($"Workflow orchestration '{instanceId}' returned no metadata.");
+        }
+
+        if (metadata.RuntimeStatus is OrchestrationRuntimeStatus.Failed)
+        {
+            throw new InvalidOperationException($"Workflow orchestration '{instanceId}' failed: {metadata.ReadOutputAs<string>()}");
```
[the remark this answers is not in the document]
kshyju: Fixed.

dotnet/src/Microsoft.Agents.AI.Hosting.AzureFunctions/Workflows/DurableWorkflowsFunctionMetadataTransformer.cs
```diff
@@ -113,6 +113,17 @@ public void Transform(IList<IFunctionMetadata> original)
...
             }
 
+            // Register an MCP tool trigger if opted in via AddWorkflow(exposeMcpToolTrigger: true).
+            if (this._options.IsMcpToolTriggerEnabled(workflow.Key))
+            {
+                string mcpToolFunctionName = $"{BuiltInFunctions.McpToolPrefix}{workflow.Key}";
+                if (registeredFunctions.Add(mcpToolFunctionName))
+                {
+                    this._logger.LogRegisteringWorkflowTrigger(workflow.Key, mcpToolFunctionName, "mcpTool");
+                    original.Add(FunctionMetadataFactory.CreateWorkflowMcpToolTrigger(workflow.Key, workflow.Value.Description));
```
cgillum, line 126: Is it possible for an agent and a workflow to have the same key value? And if so, does that mean it's possible to have a duplicate MCP function name if both enable MCP triggering?

cgillum (approved):

dotnet/src/Microsoft.Agents.AI.Hosting.AzureFunctions/Workflows/DurableWorkflowsFunctionMetadataTransformer.cs
kshyju: Currently, this fails during worker indexing (at app startup). The resulting error message is not very helpful for the user.

> Error indexing method 'Functions.dafx-Joker'. Microsoft.Azure.WebJobs.Host: Method overloads are not supported. There are multiple methods with the name 'Host.Functions.dafx-Joker'."

The root cause is that we use the same prefix (dafx-) for both agents and workflows, which leads to duplicate function names. This issue also exists for HTTP triggers when an agent and a workflow share the same name.

One potential solution is to use a different prefix for workflows (e.g., dafxw-). We can address that separately once we decide on the approach. Let me follow up offline on this.

While validating this scenario, I also discovered another bug where having both workflows and agents together did not behave as expected in azure functions hosting scenario. I’ve fixed this issue and added a new sample to demonstrate the correct behavior.

ahmedmuhsin (approved):

cgillum (approved):

PR #4787, opened 2026-03-19
ADR to support a multi-source architecture for agent skills

SergeyMenshykh:
This PR adds an ADR proposing a new design to support a multi-source architecture for agent skills.

### DIFF
--- a/docs/decisions/0021-agent-skills-design.md
+++ b/docs/decisions/0021-agent-skills-design.md
@@ -1,7 +1,7 @@
 status: proposed
 date: 2026-03-23
 contact: sergeymenshykh
-deciders: rbarreto, westey-m
+deciders: rbarreto, westey-m, eavanvalkenburg
 ---
 
 # Agent Skills: Multi-Source Architecture
@@ -29,7 +29,7 @@ Skills are presented to the model as up to three tools that progressively disclo
 - **`read_skill_resource(skillName, resourceName)`** — reads a supplementary resource (file-based or code-defined) associated with a skill
 - **`run_skill_script(skillName, scriptName, arguments?)`** — executes a script associated with a skill; only registered when at least one skill contains scripts
 
-Each tool delegates to the corresponding method on the resolved `AgentSkill` — calling `Resource.ReadAsync()` or `Script.ExecuteAsync()` respectively.
+Each tool delegates to the corresponding method on the resolved `AgentSkill` — calling `Resource.ReadAsync()` or `Script.RunAsync()` respectively.
 
 If skills have no scripts defined, the `run_skill_script` tool is **not advertised** to the model and instructions related to script execution are **not included** in the default skills instructions.
 
@@ -50,14 +50,14 @@ public abstract class AgentSkillResource
 {
     public string Name { get; }
     public string? Description { get; }
-    public abstract Task<object?> ReadAsync(AIFunctionArguments arguments, CancellationToken cancellationToken = default);
+    public abstract Task<object?> ReadAsync(IServiceProvider? serviceProvider = null, CancellationToken cancellationToken = default);
 }
 
 public abstract class AgentSkillScript
 {
     public string Name { get; }
     public string? Description { get; }
-    public abstract Task<object?> ExecuteAsync(AgentSkill skill, AIFunctionArguments arguments, CancellationToken cancellationToken = default);
+    public abstract Task<object?> RunAsync(AgentSkill skill, AIFunctionArguments arguments, CancellationToken cancellationToken = default);
 }
 
 public abstract class AgentSkillsSource
@@ -88,7 +88,7 @@ The type hierarchy at a glance:
 AgentSkill (abstract)                   AgentSkillsSource (abstract)
 ├── AgentFileSkill                      ├── AgentFileSkillsSource          (public)
 └── [Programmatic]                      ├── AgentInMemorySkillsSource      (public)
-    ├── AgentInlineSkill                ├── AggregateAgentSkillsSource     (public)
+    ├── AgentInlineSkill                ├── AggregatingAgentSkillsSource     (public)
     └── AgentClassSkill (abstract)      └── DelegatingAgentSkillsSource    (abstract, public)
                                             ├── FilteringAgentSkillsSource  (public)
 AgentSkillResource (abstract)               ├── CachingAgentSkillsSource    (public)
@@ -133,28 +133,28 @@ internal sealed class AgentFileSkillResource : AgentSkillResource
 
     public string FullPath { get; }
 
-    public override Task<object?> ReadAsync(AIFunctionArguments arguments, CancellationToken cancellationToken = default)
+    public override Task<object?> ReadAsync(IServiceProvider? serviceProvider = null, CancellationToken cancellationToken = default)
     {
         return File.ReadAllTextAsync(FullPath, Encoding.UTF8, cancellationToken);
     }
 }
 ```
 
-**`AgentFileSkillScript`** — A file-based skill script that represents a script file on disk. Delegates execution to an external `AgentFileSkillScriptExecutor` callback (e.g., runs Python/shell via `Process.Start`). Throws `NotSupportedException` if no executor is configured:
+**`AgentFileSkillScript`** — A file-based skill script that represents a script file on disk. Delegates execution to an external `AgentFileSkillScriptRunner` callback (e.g., runs Python/shell via `Process.Start`). Throws `NotSupportedException` if no executor is configured:
 
 ```csharp
-public delegate Task<object?> AgentFileSkillScriptExecutor(
+public delegate Task<object?> AgentFileSkillScriptRunner(
     AgentFileSkill skill, AgentFileSkillScript script,
     AIFunctionArguments arguments, CancellationToken cancellationToken);
 
 public sealed class AgentFileSkillScript : AgentSkillScript
 {
-    private readonly AgentFileSkillScriptExecutor _executor;
+    private readonly AgentFileSkillScriptRunner _executor;
 
-    internal AgentFileSkillScript(string name, string fullPath, AgentFileSkillScriptExecutor executor)
+    internal AgentFileSkillScript(string name, string fullPath, AgentFileSkillScriptRunner executor)
         : base(name) { ... }
 
-    public override async Task<object?> ExecuteAsync(AgentSkill skill, AIFunctionArguments arguments, ...)
+    public override async Task<object?> RunAsync(AgentSkill skill, AIFunctionArguments arguments, ...)
     {
 
         return await _executor(fileSkill, this, arguments, cancellationToken);
@@ -162,7 +162,7 @@ public sealed class AgentFileSkillScript : AgentSkillScript
 }
 ```
 
-The executor can be provided at the **provider level** via `AgentSkillsProviderBuilder.UseFileScriptExecutor(executor)` and optionally overridden for a **particular file skill** or for a **set of skills** at the file skill source level, giving fine-grained control over how different scripts are executed.
+The executor can be provided at the **provider level** via `AgentSkillsProviderBuilder.UseFileScriptRunner(executor)` and optionally overridden for a **particular file skill** or for a **set of skills** at the file skill source level, giving fine-grained control over how different scripts are executed.
 
 **`AgentFileSkillsSource`** — A skill source that discovers skills from filesystem directories containing `SKILL.md` files. Recursively scans directories (max 2 levels), validates frontmatter, and enforces path traversal and symlink security checks:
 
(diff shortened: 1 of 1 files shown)


docs/decisions/0021-agent-skills-design.md
```diff
@@ -0,0 +1,529 @@
...
+public sealed class CompositeAgentSkillsSource : AgentSkillsSource
+{
+    public override async Task<IReadOnlyList<AgentSkill>> GetSkillsAsync(...)
+    {
+        var allSkills = new List<AgentSkill>();
+        foreach (var source in _sources)
+        {
+            var skills = await source.GetSkillsAsync(cancellationToken);
+            allSkills.AddRange(skills);
+        }
```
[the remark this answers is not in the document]
SergeyMenshykh: I want the claim to be there to tell that it's possible and don't want to clutter the adr with unnecessary details.

docs/decisions/0021-agent-skills-design.md
```diff
@@ -0,0 +1,529 @@
...
+    public CachingSkillsSource(AgentSkillsSource inner)
+    {
+        _inner = inner;
+    }
+
+    public override async Task<IReadOnlyList<AgentSkill>> GetSkillsAsync(CancellationToken cancellationToken = default)
+    {
+        return _cached ??= await _inner.GetSkillsAsync(cancellationToken);
+    }
+}
```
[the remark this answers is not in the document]
SergeyMenshykh: That's OK for the ADR.

docs/decisions/0021-agent-skills-design.md
```diff
@@ -0,0 +1,529 @@
...
+    .WithPromptTemplate(customTemplate)                  // optional prompt customization
+    .Build();
+
+AIAgent agent = chatClient.AsAIAgent(new ChatClientAgentOptions
+{
+    AIContextProviders = [provider],
+});
+```
+
+## Decision Outcome
```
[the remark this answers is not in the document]
SergeyMenshykh: That's OK and it'll be updated later.

docs/decisions/0021-agent-skills-design.md
```diff
@@ -0,0 +1,591 @@
...
+    public string? Description { get; }
+    public abstract Task<object?> ReadAsync(AIFunctionArguments arguments, CancellationToken cancellationToken = default);
+}
+
+public abstract class AgentSkillScript
+{
+    public string Name { get; }
+    public string? Description { get; }
+    public abstract Task<object?> ExecuteAsync(AgentSkill skill, AIFunctionArguments arguments, CancellationToken cancellationToken = default);
+}
```
rogerbarreto, line 61: Would make sense, making it an `AIFunction` ? It seems to address this purpose, given the `AIFunctionArguments` so you just add AgentSkillScriptAIFunction to the AgentSkillScript.

And use the `InvokeAsync` as is? Including all the IServiceProvider capabilities, etc?

docs/decisions/0021-agent-skills-design.md
```diff
@@ -0,0 +1,591 @@
...
+
+```csharp
+public sealed class AgentFileSkillResource : AgentSkillResource
+{
+    public AgentFileSkillResource(string name, string path) { ... }
+
+    public override Task<object?> ReadAsync(AIFunctionArguments arguments, CancellationToken cancellationToken = default)
+    {
+        return File.ReadAllTextAsync(path, cancellationToken);
+    }
```
rogerbarreto: Is `AIFunctionArguments` intended here?

```suggestion
    public override Task<object?> ReadAsync(CancellationToken cancellationToken = default)
    {
        return File.ReadAllTextAsync(path, cancellationToken);
    }
```

docs/decisions/0021-agent-skills-design.md
```diff
@@ -0,0 +1,591 @@
...
+```
+
+**`AgentFileSkillScript`** — A file-based skill script that represents a script file on disk. Delegates execution to an external `AgentFileSkillScriptExecutor` callback (e.g., runs Python/shell via `Process.Start`). Throws `NotSupportedException` if no executor is configured:
+
+```csharp
+public delegate Task<object?> AgentFileSkillScriptExecutor(
+    AgentSkill skill, AgentFileSkillScript script,
+    AIFunctionArguments arguments, CancellationToken cancellationToken);
+
+public sealed class AgentFileSkillScript : AgentSkillScript
```
rogerbarreto, line 150: Maybe we don't need the abstract `AgentSkillScript` at all if we can use the `AIFunction` directly and implement one specific.

```suggestion
public sealed class AgentFileSkillScriptAIFunction : AIFunction
```

docs/decisions/0021-agent-skills-design.md
```diff
@@ -0,0 +1,591 @@
...
+            var response = await query.ReadNextAsync(cancellationToken);
+
+            foreach (var doc in response)
+            {
+                var frontmatter = new AgentSkillFrontmatter(doc.Name, doc.Description);
+                var resources = doc.Resources?.Select(
+                    r => new AgentCodeSkillResource(r.Content, r.Name, r.Description)).ToList();
+
+                skills.Add(new AgentCodeSkill(frontmatter, doc.Instructions)
+                    .AddResources(resources));
```
rogerbarreto: Any special reason in this example for retrieving a CodeSkillResource from the CosmosDB ?

Maybe a real `Delegate` sample would fit best for this type.

docs/decisions/0021-agent-skills-design.md
```diff
@@ -0,0 +1,591 @@
...
+
+The builder internally decides how to wire up the object graph: it creates the appropriate source instances, applies caching and filtering, and returns a fully configured `AgentSkillsProvider`. This keeps the setup code concise while still allowing fine-grained control when needed.
+
+**Example** — Using the builder to combine multiple source types with configuration:
+
+```csharp
+var provider = new AgentSkillsProviderBuilder()
+    .AddFileSkills("./skills")                           // file-based source
+    .AddCodeSkills(codeSkill)                            // code-defined source
+    .AddClassSkills(new ClassSkill())                    // class-based source
```
rogerbarreto: Our other builders have the `Use` naming, we might consider `UseFileSkills(...)`. 

`UseFileScriptExecutor()`
`UseScriptApproval()`
`UsePromptTemplate()`.

docs/decisions/0021-agent-skills-design.md
```diff
@@ -0,0 +1,591 @@
...
+// Composed with other sources and wrapped with caching
+var compositeSource = new CompositeAgentSkillsSource([
+    new CachingSkillsSource(new CosmosDbSkillsSource(cosmosClient)),
+    new AgentFileSkillsSource(["./skills"]),
+]);
+var provider = new AgentSkillsProvider(compositeSource);
+
+// Or via the builder (requires registering the source type with the builder)
+var provider = new AgentSkillsProviderBuilder()
+    .AddSource(new CosmosDbSkillsSource(cosmosClient))
```
rogerbarreto: I would drop the `CompositeAgentSkillSource` and suggest we keep via builder similar how we do for `AIAgents`, `Middlewares`, etc.

i.e
```suggestion
var provider = new AgentSkillsProvider(compositeSource);

// Or via the builder (requires registering the source type with the builder)
var provider = new AgentSkillsProviderBuilder()
    .UseSource(new CachingSkillsSource(new CosmosDbSkillsSource(cosmosClient))
    .UseSource(new AgentFileSkillsSource(["./skills"])
```

docs/decisions/0021-agent-skills-design.md
SergeyMenshykh: Yes, it's intentional for a few reasons:
 1. DI support: `AIFunctionArguments` contains an `IServiceProvider`, enabling resource functions to access services/dependencies from the service collection. We could use `IServiceProvider` directly, but that would break symmetry (see point 2) and wouldn't allow us to support LLM-provided arguments in the future without a breaking change.
 2. Symmetry: It keeps the signature consistent with `AgentSkillScript.ExecuteAsync(AgentSkill skill, AIFunctionArguments arguments, CancellationToken cancellationToken = default)`.

The base class defines the contract broadly; concrete implementations use what they need. The arguments are used by scripts for code-defined skills. For file-based skills, they are not relevant and not used.

docs/decisions/0021-agent-skills-design.md
SergeyMenshykh: I've tried it both ways. I prefer the expressiveness of the ExecuteAsync signature - as an implementer of a custom script, you clearly see the parameters the method accepts (e.g., the owning AgentSkill). With AIFunction as the base, users would override InvokeCoreAsync(AIFunctionArguments arguments, CancellationToken cancellationToken) and rely on documentation or reverse engineering to discover which arguments it's called with and how to access them.

docs/decisions/0021-agent-skills-design.md
SergeyMenshykh: Similar to the reasoning above - with AIFunction, the owning AgentSkill has to be smuggled through AIFunctionArguments via a magic string key:

```csharp
// With AgentSkillScript - contract is explicit
public override Task<object?> ExecuteAsync(
    AgentSkill skill,              // ← owning skill is right here
    AIFunctionArguments arguments,
    CancellationToken ct)
{
    var scriptPath = Path.Combine(skill.SourcePath, this.Path);
    return this._executor(skill, this, arguments, ct);
}

// With AIFunction - skill is hidden behind a convention
protected override ValueTask<object?> InvokeCoreAsync(
    AIFunctionArguments arguments,
    CancellationToken ct)
{
    // Caller must remember to do: arguments["__agentSkill"] = skill
    var skill = arguments["__agentSkill"] as AgentSkill
        ?? throw new InvalidOperationException("...");
    var scriptPath = Path.Combine(skill.SourcePath, this.Path);
    return this._executor(skill, this, arguments, ct);
}
 ```
 ExecuteAsync makes the skill-script relationship a compile-time contract. With AIFunction, it becomes a runtime convention.

Comment (rogerbarreto) [reply docs/decisions/0021-agent-skills-design.md:150]: CAn't we use the `ctor`?

`AgentFileSkillScriptAIFunction(AgentSkill) : AIFunction` ?

Unless we have a AgentFileSkillScript instance that may be shared across multiple instances I would guess that this enforces (maybe for a good reason) the script to a given skill only in runtime.

File: docs/decisions/0021-agent-skills-design.md (hunk 4 above)
Comment (rogerbarreto) [reply docs/decisions/0021-agent-skills-design.md:61]: > I've tried it both ways. I prefer the expressiveness of the ExecuteAsync signature - as an implementer of a custom script, you clearly see the parameters the method accepts (e.g., the owning AgentSkill). With AIFunction as the base, users would override InvokeCoreAsync(AIFunctionArguments arguments, CancellationToken cancellationToken) and rely on documentation or reverse engineering to discover which arguments it's called with and how to access them.

Not follow

File: docs/decisions/0021-agent-skills-design.md [hunk 10]
```diff
@@ -0,0 +1,734 @@
...
+            Refer to the conversion-table resource for supported unit pairs.
+            Run the convert script to perform conversions.
+            """
+    )
+    .AddResource("kg=2.205lb, m=3.281ft, L=0.264gal", "conversion-table", "Supported unit pairs")
+    .AddScript(Convert, "convert", "Converts a value between units");
+
+var source = new AgentInMemorySkillsSource([skill]);
+
+var provider = new AgentSkillsProvider(source);
```
rogerbarreto, line 321: I wonder if we could just hide all this composition into a single `AgentSkillProviderBuilder` and seal it if not already.

so the path suggested will be always like step 5.

```
var provider = new AgentSkillsProviderBuilder()
    .UseSource(new AgentInMemorySkillsSource([skill]))
    // Equals to
    .UseInMemorySkill(source)
    .Build();
```

docs/decisions/0021-agent-skills-design.md
SergeyMenshykh: Passing `AgentSkill` via the constructor would introduce a circular dependency: `AgentSkill` takes scripts in its constructor, and scripts would take `AgentSkill` in theirs - so neither can be constructed first. In theory, the `AgentSkill` constructor could inject `this` into each script it receives (e.g., `script.Skill = this`), but that does not feel right. Keeping `AgentSkill` as a parameter of `ExecuteAsync` avoids both issues cleanly.

docs/decisions/0021-agent-skills-design.md
SergeyMenshykh: The `AgentInMemorySkillsSource` is needed to collect and keep non-file-based skills like inline skills, which are skills defined in code, and class-based skills, which are skills defined as a class. Potentially, custom skills could be added by users.  
  
We can hide this implementation and expose `UseSkill(s)` from the builder, which will use the in-memory skill source behind the scenes. If we do that, then we can proceed with my current idea: hiding all implementations of agent skills abstractions, including deduping, filtering, and caching decorators, to enforce the use of the agent skills builder. This approach should cover the majority of use cases but would limit scenarios like caching only file-based skills and not caching the rest. To support those scenarios, we can start opening up certain internal APIs when needed.

PR #4798, opened 2026-03-19
Python: Fix A2AAgent to surface message content from in-progress TaskStatusUpdateEvents

### LINKED ISSUE #4783: Python: [Bug]: A2AAgent failed to parse message content from TaskStatusUpdateEvent
### Description

The correct message content cannot be parsed from the TaskStatusUpdateEvent as expected.
For example, the A2A Server sends messages using the code format below:

```python
               await event_queue.enqueue_event(
                    TaskStatusUpdateEvent(
                        status=TaskStatus(
                            state=TaskState.working,
                            message=new_agent_text_message(
                                text_content, # here is the content
                                task.context_id,
                                task.id,
                            ),
                        ),
                        final=False,
                        context_id=task.context_id,
                        task_id=task.id,
                    )
                )
```

However, the A2AAgent fails to extract the message content correctly.

Additionally, the test case for streaming output is also unreasonable: it only generates a single message and fails to cover scenarios with multiple messages.

### Code Sample

```markdown

```

```markdown

```

### Package Versions

agent-framework-1.0.0rc4

### Python Version

_No response_

### Additional Context

_No response_

giles17:
### Motivation and Context

When an A2A remote agent sent `TaskStatusUpdateEvent`s in a working/submitted state with `status.message` content (e.g., intermediate streaming text), `A2AAgent._parse_task_into_updates` silently discarded those messages. This caused callers to lose intermediate streaming content during non-background task execution.

Fixes #4783

### Description

The root cause was that `_parse_task_into_updates` only handled terminal task states and background continuation tokens, returning an empty list for all other in-progress updates — even when `task.status.message` carried meaningful text parts. The fix adds a check after the background-token branch: if the in-progress task's status has a non-empty message with parts, those parts are parsed into an `AgentResponseUpdate` and yielded to the caller. Three regression tests verify that working updates with message content are surfaced, that multiple intermediate messages stream correctly, and that working updates without messages remain silently skipped.

### Contribution Checklist

---

### DIFF
--- a/python/packages/a2a/agent_framework_a2a/_agent.py
+++ b/python/packages/a2a/agent_framework_a2a/_agent.py
@@ -313,6 +313,7 @@ class A2AAgent(AgentTelemetryLayer, BaseAgent):
             self._map_a2a_stream(
                 a2a_stream,
                 background=background,
+                emit_intermediate=stream,
                 session=provider_session,
                 session_context=session_context,
             ),
@@ -327,6 +328,7 @@ class A2AAgent(AgentTelemetryLayer, BaseAgent):
         a2a_stream: AsyncIterable[A2AStreamItem],
         *,
         background: bool = False,
+        emit_intermediate: bool = False,
         session: AgentSession | None = None,
         session_context: SessionContext | None = None,
     ) -> AsyncIterable[AgentResponseUpdate]:
@@ -339,6 +341,10 @@ class A2AAgent(AgentTelemetryLayer, BaseAgent):
             background: When False, in-progress task updates are silently
                 consumed (the stream keeps iterating until a terminal state).
                 When True, they are yielded with a continuation token.
+            emit_intermediate: When True, in-progress status updates that
+                carry message content are yielded to the caller.  Typically
+                set for streaming callers so non-streaming consumers only
+                receive terminal task outputs.
             session: The agent session for context providers.
             session_context: The session context for context providers.
         """
@@ -373,7 +379,11 @@ class A2AAgent(AgentTelemetryLayer, BaseAgent):
                 yield update
             elif isinstance(item, tuple) and len(item) == 2 and isinstance(item[0], Task):
                 task, _update_event = item
-                for update in self._updates_from_task(task, background=background):
+                for update in self._updates_from_task(
+                    task,
+                    background=background,
+                    emit_intermediate=emit_intermediate,
+                ):
                     all_updates.append(update)
                     yield update
             else:
@@ -389,17 +399,26 @@ class A2AAgent(AgentTelemetryLayer, BaseAgent):
     # Task helpers
     # ------------------------------------------------------------------
 
-    def _updates_from_task(self, task: Task, *, background: bool = False) -> list[AgentResponseUpdate]:
+    def _updates_from_task(
+        self,
+        task: Task,
+        *,
+        background: bool = False,
+        emit_intermediate: bool = False,
+    ) -> list[AgentResponseUpdate]:
         """Convert an A2A Task into AgentResponseUpdate(s).
 
         Terminal tasks produce updates from their artifacts/history.
         In-progress tasks produce a continuation token update when
-        ``background=True``.  When ``background=False``, any message
-        content attached to the status update is surfaced; otherwise
-        the update is silently skipped so the caller keeps consuming
-        the stream until completion.
+        ``background=True``.  When ``emit_intermediate=True`` (typically
+        set for streaming callers), any message content attached to an
+        in-progress status update is surfaced; otherwise the update is
+        silently skipped so the caller keeps consuming the stream until
+        completion.
         """
-        if task.status.state in TERMINAL_TASK_STATES:
+        status = task.status
+
+        if status.state in TERMINAL_TASK_STATES:
             task_messages = self._parse_messages_from_task(task)
             if task_messages:
                 return [
@@ -414,7 +433,7 @@ class A2AAgent(AgentTelemetryLayer, BaseAgent):
                 ]
             return [AgentResponseUpdate(contents=[], role="assistant", response_id=task.id, raw_representation=task)]
 
-        if background and task.status.state in IN_PROGRESS_TASK_STATES:
+        if background and status.state in IN_PROGRESS_TASK_STATES:
             token = self._build_continuation_token(task)
             return [
                 AgentResponseUpdate(
@@ -427,13 +446,20 @@ class A2AAgent(AgentTelemetryLayer, BaseAgent):
             ]
 
         # Surface message content from in-progress status updates (e.g. working state)
-        if task.status.message is not None and task.status.message.parts:
-            contents = self._parse_contents_from_a2a(task.status.message.parts)
+        # Only emitted when the caller opts in (streaming), so non-streaming
+        # consumers keep receiving only terminal task outputs.
+        if (
+            emit_intermediate
+            and status.state in IN_PROGRESS_TASK_STATES
+            and status.message is not None
+            and status.message.parts
+        ):
+            contents = self._parse_contents_from_a2a(status.message.parts)
             if contents:
                 return [
                     AgentResponseUpdate(
                         contents=contents,
-                        role="assistant" if task.status.message.role == A2ARole.agent else "user",
+                        role="assistant" if status.message.role == A2ARole.agent else "user",
                         response_id=task.id,
                         raw_representation=task,
                     )
--- a/python/packages/a2a/tests/test_a2a_agent.py
+++ b/python/packages/a2a/tests/test_a2a_agent.py
@@ -92,13 +92,14 @@ class MockA2AClient:
         context_id: str = "test-context",
         state: TaskState = TaskState.working,
         text: str | None = None,
+        role: A2ARole = A2ARole.agent,
     ) -> None:
         """Add a mock in-progress Task response (non-terminal)."""
         message = None
         if text is not None:
             message = A2AMessage(
                 message_id=str(uuid4()),
-                role=A2ARole.agent,
+                role=role,
                 parts=[Part(root=TextPart(text=text))],
             )
         status = TaskStatus(state=state, message=message)
@@ -110,6 +111,7 @@ class MockA2AClient:
         """Mock send_message method that yields responses."""
         self.call_count += 1
 
+        # All queued responses are delivered as a single streaming batch per call.
         for response in self.responses:
             yield response
         self.responses.clear()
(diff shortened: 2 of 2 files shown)


python/packages/a2a/tests/test_a2a_agent.py
```diff
@@ -88,9 +88,17 @@ def add_in_progress_task_response(
         task_id: str,
         context_id: str = "test-context",
         state: TaskState = TaskState.working,
+        text: str | None = None,
```
giles17, line 94: Add a `role` parameter (defaulting to `A2ARole.agent`) so tests can exercise the `"user"` role-mapping branch in `_updates_from_task`.

```suggestion
        text: str | None = None,
        role: A2ARole = A2ARole.agent,
```

python/packages/a2a/tests/test_a2a_agent.py
```diff
@@ -88,9 +88,17 @@ def add_in_progress_task_response(
...
     ) -> None:
         """Add a mock in-progress Task response (non-terminal)."""
-        status = TaskStatus(state=state, message=None)
+        message = None
+        if text is not None:
+            message = A2AMessage(
+                message_id=str(uuid4()),
+                role=A2ARole.agent,
```
giles17: Use the new `role` parameter when constructing the message so callers can test both role-mapping branches.

```suggestion
            message = A2AMessage(
                message_id=str(uuid4()),
                role=role,
                parts=[Part(root=TextPart(text=text))],
            )
```

python/packages/a2a/tests/test_a2a_agent.py
```diff
@@ -851,3 +859,58 @@ async def test_poll_task_completed(a2a_agent: A2AAgent, mock_a2a_client: MockA2A
...
+
+    updates: list[AgentResponseUpdate] = []
+    async for update in a2a_agent.run("Hello", stream=True):
+        updates.append(update)
+
+    assert len(updates) == 1
+    assert updates[0].contents[0].text == "Result"
+
+
+# endregion
```
giles17, line 1192: Missing test coverage: (1) call `add_in_progress_task_response` with `role=A2ARole.user` and assert `updates[0].role == "user"` to cover the else branch; (2) test `background=True` + status message to verify the continuation-token path takes precedence; (3) add a test where the terminal task has no artifacts after working-state messages with content, to confirm `_parse_messages_from_task`'s `task.history[-1]` fallback does not re-emit the last working-state message.

markwallace-microsoft:
<a href="https://github.com/microsoft/agent-framework/blob/3f838146ecaa1dfd9b925b46a57f0fe28c581ed0/README.md">Python Test Coverage</a><details><summary>Python Test Coverage Report • </summary><table><tr><th>File</th><th>Stmts</th><th>Miss</th><th>Cover</th><th>Missing</th></tr><tbody><tr><td colspan="5"><b>packages/a2a/agent_framework_a2a</b></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/3f838146ecaa1dfd9b925b46a57f0fe28c581ed0/packages/a2a/agent_framework_a2a/_agent.py">_agent.py</a></td><td>210</td><td>16</td><td>92%</td><td><a href="https://github.com/microsoft/agent-framework/blob/3f838146ecaa1dfd9b925b46a57f0fe28c581ed0/packages/a2a/agent_framework_a2a/_agent.py#L352">352</a>, <a href="https://github.com/microsoft/agent-framework/blob/3f838146ecaa1dfd9b925b46a57f0fe28c581ed0/packages/a2a/agent_framework_a2a/_agent.py#L357">357</a>, <a href="https://github.com/microsoft/agent-framework/blob/3f838146ecaa1dfd9b925b46a57f0fe28c581ed0/packages/a2a/agent_framework_a2a/_agent.py#L359">359</a>, <a href="https://github.com/microsoft/agent-framework/blob/3f838146ecaa1dfd9b925b46a57f0fe28c581ed0/packages/a2a/agent_framework_a2a/_agent.py#L475">475</a>, <a href="https://github.com/microsoft/agent-framework/blob/3f838146ecaa1dfd9b925b46a57f0fe28c581ed0/packages/a2a/agent_framework_a2a/_agent.py#L496">496</a>, <a href="https://github.com/microsoft/agent-framework/blob/3f838146ecaa1dfd9b925b46a57f0fe28c581ed0/packages/a2a/agent_framework_a2a/_agent.py#L517">517</a>, <a href="https://github.com/microsoft/agent-framework/blob/3f838146ecaa1dfd9b925b46a57f0fe28c581ed0/packages/a2a/agent_framework_a2a/_agent.py#L537">537</a>, <a href="https://github.com/microsoft/agent-framework/blob/3f838146ecaa1dfd9b925b46a57f0fe28c581ed0/packages/a2a/agent_framework_a2a/_agent.py#L551">551</a>, <a href="https://github.com/microsoft/agent-framework/blob/3f838146ecaa1dfd9b925b46a57f0fe28c581ed0/packages/a2a/agent_framework_a2a/_agent.py#L565">565</a>, <a href="https://github.com/microsoft/agent-framework/blob/3f838146ecaa1dfd9b925b46a57f0fe28c581ed0/packages/a2a/agent_framework_a2a/_agent.py#L577-L578">577&ndash;578</a>, <a href="https://github.com/microsoft/agent-framework/blob/3f838146ecaa1dfd9b925b46a57f0fe28c581ed0/packages/a2a/agent_framework_a2a/_agent.py#L620-L621">620&ndash;621</a>, <a href="https://github.com/microsoft/agent-framework/blob/3f838146ecaa1dfd9b925b46a57f0fe28c581ed0/packages/a2a/agent_framework_a2a/_agent.py#L650-L652">650&ndash;652</a></td></tr><tr><td><b>TOTAL</b></td><td><b>27300</b></td><td><b>3218</b></td><td><b>88%</b></td><td>&nbsp;</td></tr></tbody></table></details>

## Python Unit Test Overview
| 5355 | 20 :zzz: | 0 :x: | 0 :fire: | 1m 27s :stopwatch: |

python/packages/a2a/agent_framework_a2a/_agent.py
```diff
@@ -373,7 +379,7 @@ async def _map_a2a_stream(
                 yield update
```
giles17, line 379: The `_update_event` (`TaskStatusUpdateEvent`) is still discarded. The fix reads `task.status.message` as a proxy, but the event's own `status.message` is the authoritative payload for this streaming frame. If a real client provides a non-None event whose status diverges from the task snapshot (e.g., stale snapshot), intermediate content will be silently wrong. Prefer reading from the event when available: `source = _update_event.status if _update_event is not None else task.status`.

```suggestion
                task, _update_event = item
                effective_status = _update_event.status if _update_event is not None else task.status
                for update in self._updates_from_task(task, background=background, emit_intermediate=emit_intermediate, status_override=effective_status):
```

python/packages/a2a/agent_framework_a2a/_agent.py
```diff
@@ -424,6 +435,21 @@ def _updates_from_task(self, task: Task, *, background: bool = False) -> list[Ag
...
+        # Surface message content from in-progress status updates (e.g. working state)
+        # Only emitted when the caller opts in (streaming), so non-streaming
+        # consumers keep receiving only terminal task outputs.
+        if emit_intermediate and task.status.message is not None and task.status.message.parts:
+            contents = self._parse_contents_from_a2a(task.status.message.parts)
+            if contents:
+                return [
+                    AgentResponseUpdate(
+                        contents=contents,
+                        role="assistant" if task.status.message.role == A2ARole.agent else "user",
```
giles17: Defensive: Consider adding `task.status.state in IN_PROGRESS_TASK_STATES` to this condition, matching the pattern used by the `background` branch above. Without it, any unknown future non-terminal task state would emit intermediate content when streaming, which may not be intended.

```suggestion
        if emit_intermediate and task.status.state in IN_PROGRESS_TASK_STATES and task.status.message is not None and task.status.message.parts:
```

python/packages/a2a/tests/test_a2a_agent.py
```diff
@@ -102,9 +111,9 @@ async def send_message(self, message: Any) -> AsyncIterator[Any]:
         """Mock send_message method that yields responses."""
```
giles17, line 111: The semantics changed from 'pop one response per `send_message` call' to 'yield all queued responses and clear'. This is correct for multi-event streaming but silently removes support for multi-call test scenarios. A short comment documenting the new contract would prevent future surprises.

```suggestion
        # All queued responses are delivered as a single streaming batch per call.
        for response in self.responses:
            yield response
        self.responses.clear()
```

python/packages/a2a/tests/test_a2a_agent.py
```diff
@@ -1039,3 +1048,121 @@ async def test_run_with_continuation_token_does_not_require_messages(mock_a2a_cl
...
+
+
+async def test_streaming_working_update_without_message_is_skipped(
+    a2a_agent: A2AAgent, mock_a2a_client: MockA2AClient
+) -> None:
+    """Test that working updates without status.message are still silently skipped."""
+    mock_a2a_client.add_in_progress_task_response("task-n", context_id="ctx-n")
+    mock_a2a_client.add_task_response("task-n", [{"id": "art-n", "content": "Result"}])
+
+    updates: list[AgentResponseUpdate] = []
```
giles17, line 1098: This test verifies `stream=False` doesn't surface intermediate messages. Consider also adding a test for the case where `status.message` exists but `parts` is an empty list `[]`, to confirm the truthiness guard in `_updates_from_task` works as expected (distinct from `message=None`).

giles17 (commented):
## Automated Code Review

### ✓ Correctness
> The fix correctly addresses the bug where streaming in-progress TaskStatusUpdateEvents with message content were silently dropped. The new `emit_intermediate` flag gates intermediate content emission on `stream=True`, preserving backward compatibility for non-streaming callers. The mock `send_message` change from pop-one to yield-all is necessary for multi-event streaming tests and is backward-compatible since all existing tests add only one response. Tests are thorough, covering role mapping, background precedence, and edge cases. No correctness issues found.

### ✓ Security Reliability
> This PR fixes a real bug where streaming in-progress status updates with message content were silently dropped. The implementation is clean: it adds an `emit_intermediate` flag propagated from `stream` in `run()`, and surfaces `task.status.message` parts for non-background streaming callers. The `background` check correctly takes precedence (returning a continuation token instead of content). The mock `send_message` change from single-pop to yield-all is necessary for multi-event streaming tests and is safe since all existing tests add exactly one response. No security issues (no injection, deserialization, secrets, or resource leak risks). One minor defensive coding suggestion regarding an explicit state guard on the new intermediate block.

### ✓ Test Coverage
> The new tests provide solid coverage for the `emit_intermediate` feature: multiple streaming updates, single update, no-message skip, role mapping, background precedence, non-streaming exclusion, and terminal-after-working. The mock `send_message` change from yield-one to yield-all is backward-compatible since all existing tests add a single response. One minor gap: there is no test for the edge case where `task.status.message` has an empty `parts` list (not `None` but `[]`), which exercises the `task.status.message.parts` truthiness check. Additionally, the `response_id` field on intermediate updates is never asserted.

### ✓ Design Approach
> The fix correctly addresses the root cause: streaming `TaskStatusUpdateEvent` items with non-terminal (`working`) states were silently dropped because `_updates_from_task` had no path to yield intermediate message content. Adding `emit_intermediate` (tied to `stream=True`) and reading `task.status.message` is a clean, minimal change. Two design concerns remain: (1) the `_update_event` from the stream tuple is still unconditionally discarded — the fix reads the task snapshot (`task.status.message`) rather than the event's own payload (`_update_event.status.message`), which is the authoritative source for streaming content and works only because the client happens to embed the event data into the task snapshot; (2) the mock's `send_message` is silently changed from 'pop-one-per-call' to 'yield-all-then-clear', which is necessary for multi-event tests but would silently break any future test that calls `run()` twice against accumulated responses — the mock now behaves differently from the old design without documentation of that contract change.

### Suggestions
- Consider whether `run(stream=True, background=True)` should surface both the intermediate message content and the continuation token rather than silently discarding the message. The test `test_background_with_status_message_yields_continuation_token` verifies that background takes precedence and content is lost, which is a reasonable default but may surprise callers who want both.
- The `emit_intermediate` block (around line 447) does not verify `task.status.state in IN_PROGRESS_TASK_STATES`, unlike the `background` block above it. Adding an explicit state check would be more defensive and consistent: if the A2A protocol adds a new non-terminal, non-in-progress state in the future, content could be unexpectedly emitted.
- The `_update_event` from `task, _update_event = item` in `_map_a2a_stream` is still discarded. The `TaskStatusUpdateEvent` is the authoritative source for streamed content, not the task snapshot. If a real A2A client ever provides a non-None event whose `status.message` differs from `task.status.message`, intermediate content will be read from the wrong source. Consider reading from the event when available and falling back to the task snapshot.
- The `send_message` mock change from pop-one-per-call to yield-all-then-clear is implicitly required by the new tests but isn't documented. Adding a docstring comment explaining the new contract ('all queued responses are consumed by a single `send_message` call') would prevent future test authors from relying on the removed per-call behavior.
- Add a test where `task.status.message` exists but has an empty `parts` list (`[]`), verifying the update is skipped. This exercises the `task.status.message.parts` falsiness branch distinctly from `message=None`.
- Consider asserting `response_id` on intermediate updates (e.g., `assert updates[0].response_id == "task-w"`) in at least one streaming test to verify task ID propagation.
- In `test_terminal_no_artifacts_after_working_with_content`, the assertion `updates[1].contents == []` documents that a terminal task with no artifacts always yields an empty-content update. Consider whether streaming consumers should receive this empty sentinel at all; if not, the fix to suppress it is out of scope here but worth a follow-up issue.

---

PR #4849, opened 2026-03-23
Python: Add header_provider to Streamable HTTP MCP servers

### LINKED ISSUE #4808: Python: [Bug]: Unable to pass AgentContext to MCPStreamableHTTPTool
### Issue
Unable to pass AgentContext values from Agent Middleware into MCPStreamableHttpTool

### Description
I am unsure if this is a missing feature or simply a missing sample. Raising this as a Bug to get clarification.

### Setup:
- My Agent is hosted as a Foundry Hosted Agent using HostingAdapter.from_agent_framework(agent).run().
- In my Agent-level Middleware, I set some kwargs values. These are populated from the /responses API metadata field (which is sent to _request_headers).
- I want these kwargs values to be available to my Tool.
- I attempted to use function_invocation_kwargs, but that failed at runtime (not the main issue here).

MCP Server Integration:
- My Agent has tools accessible from an MCP server hosted externally.
- The MCP server requires certain headers.
- Currently, I send these headers using http_client.
- Ideally, I would prefer to send headers without creating a separate http_client, but that’s secondary.

Main Issue: When the MCP server is called from my Agent, it should receive the AgentContext variables I set in kwargs (from _request_headers). These values should be passed as additional headers to the MCP server, since they are required on every /responses API call.

I could not find any example showing how to send values extracted in middleware into MCPStreamableHttpTool.

Question
How can I pass AgentContext values (set in Agent Middleware) into MCPStreamableHttpTool so that they are forwarded as headers to the MCP server?

Note:
I used Agent Middleware instead of Function Middleware because _request_headers from /responses API is only populated in Agent Middleware, not in Function Middleware.

### Code Sample

```markdown
Middleware Code:


class UserContextAgentMiddleware(AgentMiddleware):
    """Agent middleware that obtains user context."""

    async def process(
        self,
        context: AgentContext,
        next: Callable[[], Awaitable[None]],
    ) -> None:
        if hasattr(context.agent, '_request_headers'):
            request_data = context.agent._request_headers

            # Runtime values that will be forwarded to tools
            if 'x-some-token' in request_data:
                context.kwargs.setdefault("some_token", request_data['x-some-token'])
            
        await next()


MCP Server Tool Setup Code:


    def create_mcp_tool(self, kwargs: Any = None) -> MCPStreamableHTTPTool:
        """Create and configure MCP tool."""
        
        print("Running MCP tool with URL:", kwargs)
        # Set up HTTP client if not already done
        if not self.http_client:
            self.setup_http_client()
            
        logger.info(f"Creating MCP tool for server: {self.config.mcp_server_url}")
        
        self.mcp_tool = MCPStreamableHTTPTool(
            name="University MCP Tool",
            description="Access to resources in ",
            url=self.config.mcp_server_url,
            http_client=self.http_client,
            load_tools=True,
            load_prompts=False,
            approval_mode="never_require",
        )
        
        return self.mcp_tool


Http Client needs some fixed headers and some dynamic per request headers. Code to set fixed header is something like below (its not exact code, but so far fixed header works):


    def setup_http_client(self) -> None:
        """Create HTTP client with dynamic authentication for MCP tools."""        
        keyvault_manager = KeyVaultManager(
            key_vault_url=self.config.azure_keyvault_url
        )
        
        # Retrieve API keys using the same KeyVault client
        mcp_subscription_key = keyvault_manager.get_secret(self.config.mcp_key_subscription_key)

        self.http_client = AsyncClient(auth=mcp_subscription_key, timeout=30.0)


Main question is - How can I send per call additional headers that goes to MCP Server. These headers will be dynamically populated from /responses api metadata field.
```

```markdown

```

### Package Versions

agent-framework-azure-ai latest

### Python Version

3.12

### Additional Context

_No response_

eavanvalkenburg:
### Motivation and Context

When `header_provider` was set on `MCPStreamableHTTPTool`, the code used `create_mcp_http_client()` from the MCP library to create an HTTP client. This function may not be available in all versions of the MCP SDK and introduces an unnecessary dependency on an internal MCP helper, causing failures when trying to pass `AgentContext` headers.

Fixes #4808

### Description

The root cause was that `MCPStreamableHTTPTool._get_client()` relied on `mcp.client.streamable_http.create_mcp_http_client` to construct the default `httpx.AsyncClient` when a `header_provider` was configured but no explicit `httpx_client` was supplied. This was replaced with a direct `httpx.AsyncClient(follow_redirects=True, timeout=httpx.Timeout(30.0, read=300.0))` construction, which matches the equivalent defaults and eliminates the problematic import. This ensures header injection (used for passing `AgentContext`) works reliably without depending on the MCP SDK's internal client factory.

### Contribution Checklist

---

### DIFF
--- a/python/packages/core/agent_framework/_mcp.py
+++ b/python/packages/core/agent_framework/_mcp.py
@@ -143,7 +143,18 @@ def _inject_otel_into_mcp_meta(meta: dict[str, Any] | None = None) -> dict[str, 
 
 def streamable_http_client(*args: Any, **kwargs: Any) -> _AsyncGeneratorContextManager[Any, None]:
     """Lazily import the MCP streamable HTTP transport."""
-    from mcp.client.streamable_http import streamable_http_client as _streamable_http_client
+    try:
+        from mcp.client.streamable_http import streamable_http_client as _streamable_http_client
+    except ModuleNotFoundError as ex:
+        missing_name = ex.name or str(ex)
+        if missing_name == "mcp" or missing_name.startswith("mcp."):
+            raise ModuleNotFoundError("`MCPStreamableHTTPTool` requires `mcp`. Please install `mcp`.") from ex
+        if "mcp" in missing_name:
+            raise ModuleNotFoundError("`MCPStreamableHTTPTool` requires `mcp`. Please install `mcp`.") from ex
+        raise ModuleNotFoundError(
+            f"`MCPStreamableHTTPTool` requires streamable HTTP transport support. "
+            f"The optional dependency `{missing_name}` is not installed. Please update your dependencies."
+        ) from ex
 
     return cast(_AsyncGeneratorContextManager[Any, None], _streamable_http_client(*args, **kwargs))
 
@@ -962,9 +973,17 @@ class MCPTool:
                 input_schema = dict(tool.inputSchema or {})
                 if input_schema.get("type") == "object" and "properties" not in input_schema:
                     input_schema["properties"] = {}
+
+                async def _call_tool_with_runtime_kwargs(
+                    *,
+                    _remote_tool_name: str = tool.name,
+                    **kwargs: Any,
+                ) -> str | list[Content]:
+                    return await self.call_tool(_remote_tool_name, **kwargs)
+
                 # Create FunctionTools out of each tool
                 func: FunctionTool = FunctionTool(
-                    func=partial(self.call_tool, tool.name),
+                    func=_call_tool_with_runtime_kwargs,
                     name=local_name,
                     description=tool.description or "",
                     approval_mode=approval_mode,
--- a/python/packages/core/agent_framework/_mcp.py
+++ b/python/packages/core/agent_framework/_mcp.py
@@ -148,16 +148,14 @@ def streamable_http_client(*args: Any, **kwargs: Any) -> _AsyncGeneratorContextM
         from mcp.client.streamable_http import streamable_http_client as _streamable_http_client
     except ModuleNotFoundError as ex:
         missing_name = ex.name or str(ex)
-        if missing_name == "mcp" or missing_name.startswith("mcp."):
-            raise ModuleNotFoundError("`MCPStreamableHTTPTool` requires `mcp`. Please install `mcp`.") from ex
-        if "mcp" in missing_name:
+        if missing_name == "mcp" or missing_name.startswith("mcp.") or "mcp" in missing_name:
             raise ModuleNotFoundError("`MCPStreamableHTTPTool` requires `mcp`. Please install `mcp`.") from ex
         raise ModuleNotFoundError(
             f"`MCPStreamableHTTPTool` requires streamable HTTP transport support. "
             f"The optional dependency `{missing_name}` is not installed. Please update your dependencies."
         ) from ex
 
-    return cast(_AsyncGeneratorContextManager[Any, None], _streamable_http_client(*args, **kwargs))
+    return _streamable_http_client(*args, **kwargs)  # type: ignore[return-value]
 
 
 # region: MCP Plugin
--- a/python/packages/core/agent_framework/_mcp.py
+++ b/python/packages/core/agent_framework/_mcp.py
@@ -39,6 +39,7 @@ if TYPE_CHECKING:
     from mcp.shared.session import RequestResponder
 
     from ._clients import SupportsChatGetResponse
+    from ._middleware import FunctionInvocationContext
 
 
 logger = logging.getLogger(__name__)
@@ -975,11 +976,14 @@ class MCPTool:
                     input_schema["properties"] = {}
 
                 async def _call_tool_with_runtime_kwargs(
+                    ctx: FunctionInvocationContext,
                     *,
                     _remote_tool_name: str = tool.name,
                     **kwargs: Any,
                 ) -> str | list[Content]:
-                    return await self.call_tool(_remote_tool_name, **kwargs)
+                    call_kwargs = dict(ctx.kwargs)
+                    call_kwargs.update(kwargs)
+                    return await self.call_tool(_remote_tool_name, **call_kwargs)
 
                 # Create FunctionTools out of each tool
                 func: FunctionTool = FunctionTool(
@@ -993,7 +997,6 @@ class MCPTool:
                         _MCP_NORMALIZED_NAME_KEY: normalized_name,
                     },
                 )
-                func.__dict__["_forward_runtime_kwargs"] = True
                 self._functions.append(func)
                 existing_names.add(local_name)
 
--- a/python/packages/core/agent_framework/_mcp.py
+++ b/python/packages/core/agent_framework/_mcp.py
@@ -61,6 +61,8 @@ class MCPSpecificApproval(TypedDict, total=False):
 _MCP_REMOTE_NAME_KEY = "_mcp_remote_name"
 _MCP_NORMALIZED_NAME_KEY = "_mcp_normalized_name"
 _mcp_call_headers: contextvars.ContextVar[dict[str, str]] = contextvars.ContextVar("_mcp_call_headers")
+MCP_DEFAULT_TIMEOUT = 30
+MCP_DEFAULT_SSE_READ_TIMEOUT = 60 * 5
 
 # region: Helpers
 
@@ -137,6 +139,13 @@ def _inject_otel_into_mcp_meta(meta: dict[str, Any] | None = None) -> dict[str, 
             meta[key] = value
 
     return meta
+
+
+def streamable_http_client(*args: Any, **kwargs: Any) -> _AsyncGeneratorContextManager[Any, None]:
+    """Lazily import the MCP streamable HTTP transport."""
+    from mcp.client.streamable_http import streamable_http_client as _streamable_http_client
+
+    return cast(_AsyncGeneratorContextManager[Any, None], _streamable_http_client(*args, **kwargs))
 
 
 # region: MCP Plugin
@@ -1469,21 +1478,20 @@ class MCPStreamableHTTPTool(MCPTool):
         Returns:
             An async context manager for the streamable HTTP client transport.
         """
-        from httpx import AsyncClient
-        from mcp.client.streamable_http import streamable_http_client
+        from httpx import AsyncClient, Request, Timeout
 
         http_client = self._httpx_client
         if self._header_provider is not None:
             if http_client is None:
                 http_client = AsyncClient(
                     follow_redirects=True,
-                    timeout=httpx.Timeout(MCP_DEFAULT_TIMEOUT, read=MCP_DEFAULT_SSE_READ_TIMEOUT),
+                    timeout=Timeout(MCP_DEFAULT_TIMEOUT, read=MCP_DEFAULT_SSE_READ_TIMEOUT),
                 )
                 self._httpx_client = http_client
 
             if not hasattr(self, "_inject_headers_hook"):
 
-                async def _inject_headers(request: httpx.Request) -> None:  # noqa: RUF029
+                async def _inject_headers(request: Request) -> None:  # noqa: RUF029
                     headers = _mcp_call_headers.get({})
                     for key, value in headers.items():
                         request.headers[key] = value
--- a/python/packages/core/agent_framework/_mcp.py
+++ b/python/packages/core/agent_framework/_mcp.py
@@ -1469,22 +1469,27 @@ class MCPStreamableHTTPTool(MCPTool):
         Returns:
             An async context manager for the streamable HTTP client transport.
         """
+        from httpx import AsyncClient
         from mcp.client.streamable_http import streamable_http_client
 
         http_client = self._httpx_client
         if self._header_provider is not None:
             if http_client is None:
-                from mcp.shared._httpx_utils import create_mcp_http_client
-
-                http_client = create_mcp_http_client()
+                http_client = AsyncClient(
+                    follow_redirects=True,
+                    timeout=httpx.Timeout(MCP_DEFAULT_TIMEOUT, read=MCP_DEFAULT_SSE_READ_TIMEOUT),
+                )
                 self._httpx_client = http_client
 
-            async def _inject_headers(request: httpx.Request) -> None:  # noqa: RUF029
-                headers = _mcp_call_headers.get({})
-                for key, value in headers.items():
-                    request.headers[key] = value
+            if not hasattr(self, "_inject_headers_hook"):
 
-            http_client.event_hooks["request"].append(_inject_headers)
+                async def _inject_headers(request: httpx.Request) -> None:  # noqa: RUF029
+                    headers = _mcp_call_headers.get({})
+                    for key, value in headers.items():
+                        request.headers[key] = value
+
+                self._inject_headers_hook = _inject_headers  # type: ignore[attr-defined]
+                http_client.event_hooks["request"].append(self._inject_headers_hook)  # type: ignore[attr-defined]
 
         return streamable_http_client(
             url=self.url,
--- a/python/packages/core/agent_framework/_mcp.py
+++ b/python/packages/core/agent_framework/_mcp.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import asyncio
 import base64
+import contextvars
 import json
 import logging
 import re
@@ -59,6 +60,7 @@ class MCPSpecificApproval(TypedDict, total=False):
 
 _MCP_REMOTE_NAME_KEY = "_mcp_remote_name"
 _MCP_NORMALIZED_NAME_KEY = "_mcp_normalized_name"
+_mcp_call_headers: contextvars.ContextVar[dict[str, str]] = contextvars.ContextVar("_mcp_call_headers")
 
 # region: Helpers
 
@@ -1386,6 +1388,7 @@ class MCPStreamableHTTPTool(MCPTool):
         client: SupportsChatGetResponse | None = None,
         additional_properties: dict[str, Any] | None = None,
         http_client: AsyncClient | None = None,
+        header_provider: Callable[[dict[str, Any]], dict[str, str]] | None = None,
         **kwargs: Any,
     ) -> None:
         """Initialize the MCP streamable HTTP tool.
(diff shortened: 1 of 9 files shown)


markwallace-microsoft:
<a href="https://github.com/microsoft/agent-framework/blob/2f3afadede7456ea354770c8aec15b175a3240a5/README.md">Python Test Coverage</a><details><summary>Python Test Coverage Report • </summary><table><tr><th>File</th><th>Stmts</th><th>Miss</th><th>Cover</th><th>Missing</th></tr><tbody><tr><td colspan="5"><b>packages/core/agent_framework</b></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/2f3afadede7456ea354770c8aec15b175a3240a5/packages/core/agent_framework/_mcp.py">_mcp.py</a></td><td>611</td><td>30</td><td>95%</td><td><a href="https://github.com/microsoft/agent-framework/blob/2f3afadede7456ea354770c8aec15b175a3240a5/packages/core/agent_framework/_mcp.py#L153">153</a>, <a href="https://github.com/microsoft/agent-framework/blob/2f3afadede7456ea354770c8aec15b175a3240a5/packages/core/agent_framework/_mcp.py#L294">294</a>, <a href="https://github.com/microsoft/agent-framework/blob/2f3afadede7456ea354770c8aec15b175a3240a5/packages/core/agent_framework/_mcp.py#L354-L355">354&ndash;355</a>, <a href="https://github.com/microsoft/agent-framework/blob/2f3afadede7456ea354770c8aec15b175a3240a5/packages/core/agent_framework/_mcp.py#L484">484</a>, <a href="https://github.com/microsoft/agent-framework/blob/2f3afadede7456ea354770c8aec15b175a3240a5/packages/core/agent_framework/_mcp.py#L546">546</a>, <a href="https://github.com/microsoft/agent-framework/blob/2f3afadede7456ea354770c8aec15b175a3240a5/packages/core/agent_framework/_mcp.py#L559">559</a>, <a href="https://github.com/microsoft/agent-framework/blob/2f3afadede7456ea354770c8aec15b175a3240a5/packages/core/agent_framework/_mcp.py#L561-L564">561&ndash;564</a>, <a href="https://github.com/microsoft/agent-framework/blob/2f3afadede7456ea354770c8aec15b175a3240a5/packages/core/agent_framework/_mcp.py#L583-L584">583&ndash;584</a>, <a href="https://github.com/microsoft/agent-framework/blob/2f3afadede7456ea354770c8aec15b175a3240a5/packages/core/agent_framework/_mcp.py#L597-L600">597&ndash;600</a>, <a href="https://github.com/microsoft/agent-framework/blob/2f3afadede7456ea354770c8aec15b175a3240a5/packages/core/agent_framework/_mcp.py#L602-L603">602&ndash;603</a>, <a href="https://github.com/microsoft/agent-framework/blob/2f3afadede7456ea354770c8aec15b175a3240a5/packages/core/agent_framework/_mcp.py#L607">607</a>, <a href="https://github.com/microsoft/agent-framework/blob/2f3afadede7456ea354770c8aec15b175a3240a5/packages/core/agent_framework/_mcp.py#L670-L672">670&ndash;672</a>, <a href="https://github.com/microsoft/agent-framework/blob/2f3afadede7456ea354770c8aec15b175a3240a5/packages/core/agent_framework/_mcp.py#L1031">1031</a>, <a href="https://github.com/microsoft/agent-framework/blob/2f3afadede7456ea354770c8aec15b175a3240a5/packages/core/agent_framework/_mcp.py#L1144">1144</a>, <a href="https://github.com/microsoft/agent-framework/blob/2f3afadede7456ea354770c8aec15b175a3240a5/packages/core/agent_framework/_mcp.py#L1196-L1197">1196&ndash;1197</a>, <a href="https://github.com/microsoft/agent-framework/blob/2f3afadede7456ea354770c8aec15b175a3240a5/packages/core/agent_framework/_mcp.py#L1200">1200</a>, <a href="https://github.com/microsoft/agent-framework/blob/2f3afadede7456ea354770c8aec15b175a3240a5/packages/core/agent_framework/_mcp.py#L1218">1218</a>, <a href="https://github.com/microsoft/agent-framework/blob/2f3afadede7456ea354770c8aec15b175a3240a5/packages/core/agent_framework/_mcp.py#L1671">1671</a></td></tr><tr><td><b>TOTAL</b></td><td><b>28000</b></td><td><b>3355</b></td><td><b>88%</b></td><td>&nbsp;</td></tr></tbody></table></details>

## Python Unit Test Overview
| 5533 | 20 :zzz: | 0 :x: | 0 :fire: | 1m 26s :stopwatch: |

eavanvalkenburg (commented):
## Automated Code Review

### ✓ Correctness
> The diff replaces `create_mcp_http_client()` with an inline `httpx.AsyncClient(...)` using identical defaults (follow_redirects=True, timeout 30s/read 300s), which is functionally correct. However, the surrounding code has a pre-existing bug that this diff doesn't fix but is worth noting: every call to `get_mcp_client()` appends a new `_inject_headers` callback to `http_client.event_hooks["request"]` (line 1434) on the same cached `self._httpx_client`. If the tool reconnects multiple times, duplicate hooks accumulate and headers are injected multiple times per request. This isn't introduced by the diff but is a correctness concern in the method this diff touches.

### ✓ Security Reliability
> The diff replaces the MCP SDK's `create_mcp_http_client()` helper with an equivalent inline `httpx.AsyncClient` construction. The hardcoded timeout values (30.0s general, 300.0s read) exactly match the SDK's `MCP_DEFAULT_TIMEOUT` and `MCP_DEFAULT_SSE_READ_TIMEOUT` constants, so the behavior is preserved. However, this introduces a maintenance coupling: if the MCP SDK updates its defaults in a future version, this code won't track those changes. More critically, the pre-existing event_hooks append at line 1434 (untouched by this diff) causes unbounded growth of `_inject_headers` callbacks on every `get_mcp_client()` call, which is a reliability/resource leak concern that this change should address since it's creating the client specifically for this code path.

### ✓ Test Coverage
> The diff replaces `create_mcp_http_client()` (from the MCP SDK) with an inline `httpx.AsyncClient(follow_redirects=True, timeout=httpx.Timeout(30.0, read=300.0))` for the auto-created client when `header_provider` is set. The existing test `test_mcp_streamable_http_tool_header_provider_with_httpx_event_hook` exercises the changed code path and verifies that a client is created and the event hook is wired up. However, no test asserts the specific configuration properties (follow_redirects, timeout) of the auto-created client, meaning a regression in these values (or a drift from what the MCP SDK expects) would go undetected. This is a minor gap, not a blocking issue.

### ✗ Design Approach
> The diff replaces a lazy import of `create_mcp_http_client()` with an equivalent inline `httpx.AsyncClient` construction using identical default values (`follow_redirects=True`, `timeout=30.0/read=300.0`). The values are functionally correct for now, but two concerns exist: (1) hardcoding the timeout values decouples from MCP's own constants so future MCP default changes won't be inherited; (2) the pre-existing `http_client.event_hooks["request"].append(_inject_headers)` at line 1434 sits *outside* the `if http_client is None:` guard, so on every call to `get_mcp_client()` a new closure is appended to the same persistent `AsyncClient`, causing the hook list to grow unboundedly — headers get injected N times after N tool calls.

### Flagged Issues
- [ ] Event hook accumulation: `http_client.event_hooks["request"].append(_inject_headers)` (line 1434) executes on every `get_mcp_client()` call, not just on first client creation. Because `self._httpx_client` is reused across calls, the hook list grows by one closure per invocation, causing headers to be injected N times after N tool calls. The hook registration should be moved inside the `if http_client is None:` block, immediately after client creation.

### Suggestions
- The timeout values (30.0 / 300.0) are hardcoded copies of MCP SDK constants (`MCP_DEFAULT_TIMEOUT`, `MCP_DEFAULT_SSE_READ_TIMEOUT` from `mcp.shared._httpx_utils`). Consider importing and using the SDK constants so the defaults stay in sync if the SDK updates them.
- Consider adding assertions in `test_mcp_streamable_http_tool_header_provider_with_httpx_event_hook` to verify the auto-created httpx client's configuration (`follow_redirects=True`, `timeout.connect == 30.0`, `timeout.read == 300.0`) since this diff explicitly replaces `create_mcp_http_client()` with hardcoded values.

---

python/packages/core/agent_framework/_mcp.py
```diff
@@ -1401,20 +1409,62 @@ def __init__(
...
         Returns:
             An async context manager for the streamable HTTP client transport.
         """
-        # Pass the http_client (which may be None) to streamable_http_client
+        http_client = self._httpx_client
+        if self._header_provider is not None:
+            if http_client is None:
+                http_client = httpx.AsyncClient(
+                    follow_redirects=True,
+                    timeout=httpx.Timeout(30.0, read=300.0),
```
eavanvalkenburg: Nit: The hardcoded timeout values (30.0 / 300.0) duplicate `MCP_DEFAULT_TIMEOUT` and `MCP_DEFAULT_SSE_READ_TIMEOUT` from `mcp.shared._httpx_utils`. Importing the SDK constants would keep this in sync if the SDK changes its defaults.

```suggestion
                http_client = httpx.AsyncClient(
                    follow_redirects=True,
                    timeout=httpx.Timeout(MCP_DEFAULT_TIMEOUT, read=MCP_DEFAULT_SSE_READ_TIMEOUT),
                )
```

python/packages/core/agent_framework/_mcp.py
```diff
@@ -1401,20 +1409,62 @@ def __init__(
...
+                )
```
eavanvalkenburg, line 1510: Test gap: `test_mcp_streamable_http_tool_header_provider_with_httpx_event_hook` verifies that `_httpx_client is not None` and the hook is wired, but doesn't assert the client's `follow_redirects` or `timeout` settings. Consider adding assertions like `assert tool._httpx_client._follow_redirects is True` and `assert tool._httpx_client.timeout.read == 300.0` to catch regressions.

python/packages/core/agent_framework/_mcp.py
```diff
@@ -1401,20 +1409,62 @@ def __init__(
...
+                self._httpx_client = http_client
```
eavanvalkenburg: The hook append at line 1434 is outside this guard, so it fires on every `get_mcp_client()` call, accumulating duplicate closures on the shared `AsyncClient`. Move the hook registration inside this block so it only happens once, when the client is first created.

```suggestion
                http_client = httpx.AsyncClient(
                    follow_redirects=True,
                    timeout=httpx.Timeout(30.0, read=300.0),
                )
                self._httpx_client = http_client

                async def _inject_headers(request: httpx.Request) -> None:  # noqa: RUF029
                    headers = _mcp_call_headers.get({})
                    for key, value in headers.items():
                        request.headers[key] = value

                http_client.event_hooks["request"].append(_inject_headers)
```

python/packages/core/tests/core/test_mcp.py
```diff
@@ -3129,3 +3129,260 @@ def get_mcp_client(self) -> _AsyncGeneratorContextManager[Any, None]:
...
+    server = _TestServer(
+        name="test",
+        url="http://example.com/mcp",
+        header_provider=provider,
+    )
+    async with server:
+        await server.load_tools()
+
+        # Simulate the runtime kwargs that flow from FunctionInvocationContext.kwargs
+        await server.call_tool("greet", name="Alice", some_token="my-secret")
```
eavanvalkenburg, line 3881: we should also test using `tool.invoke` with the FunctionInvocationContext

python/packages/core/agent_framework/_mcp.py
```diff
@@ -22,6 +23,7 @@
 from mcp.client.stdio import StdioServerParameters, stdio_client
```
eavanvalkenburg: Importing from `mcp.shared._httpx_utils` (underscore-prefixed = private module) is fragile. The MCP SDK gives no stability guarantees for this module, and the constants aren't re-exported from any public namespace. Use a try/except import with literal fallbacks to stay in sync with MCP defaults while remaining resilient to SDK changes.

```suggestion
try:
    from mcp.shared._httpx_utils import MCP_DEFAULT_SSE_READ_TIMEOUT, MCP_DEFAULT_TIMEOUT
except ImportError:
    MCP_DEFAULT_TIMEOUT = 30.0
    MCP_DEFAULT_SSE_READ_TIMEOUT = 300.0
```

eavanvalkenburg (commented):
## Automated Code Review

### ✓ Correctness
> The diff fixes a real bug where repeated `get_mcp_client()` calls would accumulate duplicate httpx event hooks, and replaces hardcoded timeout values with MCP SDK constants. The hook deduplication logic using `hasattr` is correct, and the new tests properly cover the fix. The main concern is importing from `mcp.shared._httpx_utils`, a private (underscore-prefixed) module of the MCP SDK, which could break on SDK upgrades. All logic changes are correct and the test cleanup improvements are good practice.

### ✓ Security Reliability
> The changes fix a real bug (duplicate event hooks on repeated `get_mcp_client` calls) and align timeout defaults with the MCP SDK. The implementation is sound: using `hasattr` to guard against duplicate hook registration, `contextvars` for async-safe per-call headers, and proper test cleanup with `aclose()`. One concern is the import from `mcp.shared._httpx_utils`, which is a private module — but the constants are in its `__all__`, reducing the risk. The user-provided httpx client test (`test_mcp_streamable_http_tool_header_provider_with_user_httpx_client`) was not updated with the same `try/finally` cleanup pattern, leaving it inconsistent, but it does call `aclose()` at the end. No security, injection, or resource leak issues found in the changed code.

### ✓ Test Coverage
> The diff fixes a hook duplication bug in `get_mcp_client()` (using `hasattr` guard), replaces hardcoded timeout values with MCP SDK constants, and adds a new test for the dedup fix plus improves an existing test with proper resource cleanup and timeout assertions. Test coverage for the changed behavior is adequate: the new `test_mcp_streamable_http_tool_hook_not_duplicated_on_repeated_get_mcp_client` test directly validates the fix, and the updated `test_mcp_streamable_http_tool_header_provider_with_httpx_event_hook` test verifies the timeout constants and adds proper `aclose()` cleanup. Minor gaps exist around testing dedup with a user-provided httpx client and error-path behavior of `header_provider`, but these are not blocking.

### ✗ Design Approach
> The PR introduces two changes: (1) replacing hardcoded httpx timeout literals with constants imported from `mcp.shared._httpx_utils`, and (2) guarding the event-hook registration with `hasattr` to prevent duplicate hooks on repeated `get_mcp_client()` calls. The hook-deduplication fix is correct and the test covers it well. However, importing from a private module (`_httpx_utils`) is a fragile design choice, and the `hasattr`-based guard has a structural smell that the two `# type: ignore[attr-defined]` comments confirm — the attribute is never declared in `__init__`.

### Flagged Issues
- [x] Importing from `mcp.shared._httpx_utils` (underscore-prefixed = private module) couples the code to MCP SDK internals with no stability guarantees — the module can be renamed, moved, or removed between minor releases. The previous hardcoded literals (30.0 / 300.0) were explicit and stable. At minimum, add a try/except import with literal fallbacks so the code doesn't break on SDK upgrades.

### Suggestions
- Consider adding a test that calls `get_mcp_client()` multiple times when a user-provided `http_client` is passed, to verify the hook dedup guard also works in that code path.
- Consider adding a test that verifies behavior when `header_provider` raises an exception — confirming the exception propagates cleanly without leaving the contextvar in a dirty state.
- The `_inject_headers_hook` attribute is never declared in `__init__`, which is why `hasattr` and two `# type: ignore[attr-defined]` comments are needed. A cleaner design: declare `self._inject_headers_hook: Callable | None = None` in `__init__`, then check `if self._inject_headers_hook is None` in `get_mcp_client`. This eliminates the `hasattr` smell, removes the type ignores, and makes the class's full interface visible at construction time.

---

python/packages/core/agent_framework/_mcp.py
eavanvalkenburg: Fixed in `c82daa3d`.

python/packages/core/agent_framework/_mcp.py
eavanvalkenburg: Fixed in `c82daa3d`.

python/packages/core/agent_framework/_mcp.py
eavanvalkenburg: Fixed in `c82daa3d`.

python/packages/core/tests/core/test_mcp.py
```diff
@@ -3129,3 +3129,260 @@ def get_mcp_client(self) -> _AsyncGeneratorContextManager[Any, None]:
...
+        assert len(hooks) == 1, "Expected one request event hook"
+
+        # Simulate what happens during a call_tool: contextvar is set
+        token = _mcp_call_headers.set({"X-Custom": "test-value"})
+        try:
+            request = httpx.Request("POST", "http://example.com/mcp")
+            await hooks[0](request)
+            assert request.headers.get("X-Custom") == "test-value"
+        finally:
+            _mcp_call_headers.reset(token)
```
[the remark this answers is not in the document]
eavanvalkenburg: Fixed in `c82daa3d`.

python/packages/core/agent_framework/_mcp.py
```diff
@@ -1401,20 +1409,62 @@ def __init__(
...
+                )
+                self._httpx_client = http_client
+
+            async def _inject_headers(request: httpx.Request) -> None:  # noqa: RUF029
+                headers = _mcp_call_headers.get({})
+                for key, value in headers.items():
+                    request.headers[key] = value
+
+            http_client.event_hooks["request"].append(_inject_headers)
+
```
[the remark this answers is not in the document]
eavanvalkenburg: Fixed in `c82daa3d`.

python/packages/core/tests/core/test_mcp.py
```diff
@@ -3129,3 +3129,260 @@ def get_mcp_client(self) -> _AsyncGeneratorContextManager[Any, None]:
...
+
+# region: MCPStreamableHTTPTool header_provider
+
+
+async def test_mcp_streamable_http_tool_header_provider_injects_headers():
+    """Test that header_provider injects per-request HTTP headers via runtime kwargs.
+
+    When header_provider is configured, runtime kwargs from FunctionInvocationContext
+    are passed to the provider and the returned headers appear on outbound HTTP requests.
+    """
```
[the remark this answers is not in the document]
eavanvalkenburg: Fixed in `c82daa3d`.

python/packages/core/tests/core/test_mcp.py
eavanvalkenburg: Fixed in `c82daa3d`.

python/packages/core/agent_framework/_mcp.py
eavanvalkenburg: Fixed in `c82daa3d`.

python/packages/core/tests/core/test_mcp.py
eavanvalkenburg: Fixed in `63d93304`.

giles17 (approved):

ashishmundra4001 (approved):
Looks awesome. Please merge to main.

chetantoshniwal (approved):

PR #4866, opened 2026-03-23
Python: fix: prevent MCP message_handler deadlock on notification reload

### LINKED ISSUE #4828: Python: [Bug]: MCPStdioTool drops successful MATLAB MCP results and sends "Error: Function failed." instead
### Description

This is a follow-up to #4540.

Before rc5, the MATLAB MCP path was already blocked earlier in the request lifecycle by the `/v1/responses` tool schema issue (`tools.0.type` / `invalid_request_error`).

After upgrading to `agent-framework-core==1.0.0rc5`, that earlier problem appears to be resolved: the request now progresses into a real MCP tool invocation.

However, I am now hitting a different failure later in the pipeline when using the official [`matlab-mcp-core-server`](https://github.com/matlab/matlab-mcp-core-server) through `agent_framework.MCPStdioTool`.

What I observed:

- A direct MCP SDK call using `mcp.ClientSession.call_tool("detect_matlab_toolboxes", {})` succeeds.
- The returned `CallToolResult` has `isError=False` and contains the expected MATLAB installation / toolbox information.
- However, calling the same tool through `agent_framework.MCPStdioTool.call_tool("detect_matlab_toolboxes")` times out with `ToolExecutionException`.
- During an Agent run, the real tool result is not forwarded to the model. The second provider request contains `function_call_output: "Error: Function failed."` instead of the actual MATLAB output.

I can confirm that this is not a “MATLAB never started” problem.

When reproducing this through `agent_framework.MCPStdioTool`, I can see the Matlab start successfully on local machine, and the server logs show that `detect_matlab_toolboxes` is actually executed to completion.

So the failure appears to happen after the real MATLAB MCP tool execution has already started or completed, not before tool invocation.

I also verified that the provider can handle the real MATLAB tool output if I manually send the successful result back as `function_call_output`, so this does not appear to be caused by the provider/model being unable to consume the MATLAB result.

At a high level, the current behavior is:

- direct `ClientSession.call_tool(...)`: succeeds
- `MCPStdioTool.call_tool(...)`: times out
- Agent run then replaces the real tool result with `"Error: Function failed."`

### Code Sample

```markdown
from __future__ import annotations

import asyncio
import json
import os
from contextlib import suppress
from dataclasses import dataclass
from datetime import timedelta
from pathlib import Path
from time import monotonic
from typing import Any

from agent_framework import MCPStdioTool
from mcp import ClientSession
from mcp.client.stdio import StdioServerParameters, stdio_client

PLACEHOLDER_MCP_COMMAND = "path/to/matlab-mcp-core-server"
PLACEHOLDER_MATLAB_ROOT = "/path/to/MATLAB_R2025b.app"

WORKING_DIR = Path(
    os.getenv(
        "AGENT_MCP_REPRO_WORKDIR",
        str(Path(__file__).resolve().parents[2]),
    )
)
MATLAB_MCP_COMMAND = os.getenv(
    "AGENT_MCP_REPRO_MATLAB_MCP_COMMAND",
    PLACEHOLDER_MCP_COMMAND,
)
MATLAB_ROOT = os.getenv("AGENT_MCP_REPRO_MATLAB_ROOT", PLACEHOLDER_MATLAB_ROOT)
MATLAB_DISPLAY_MODE = os.getenv("AGENT_MCP_REPRO_MATLAB_DISPLAY_MODE", "nodesktop")
REQUEST_TIMEOUT_SECONDS = int(os.getenv("AGENT_MCP_REPRO_TIMEOUT_SECONDS", "30"))
ALLOWED_TOOLS = ["detect_matlab_toolboxes"]


@dataclass
class DirectMCPResult:
    is_error: bool
    elapsed_seconds: float
    text_excerpt: str


def _validate_required_config() -> None:
    if MATLAB_MCP_COMMAND == PLACEHOLDER_MCP_COMMAND:
        raise RuntimeError(
            "Set AGENT_MCP_REPRO_MATLAB_MCP_COMMAND to your matlab-mcp-core-server executable path."
        )
    if MATLAB_ROOT == PLACEHOLDER_MATLAB_ROOT:
        raise RuntimeError(
            "Set AGENT_MCP_REPRO_MATLAB_ROOT to your local MATLAB installation path."
        )


def _build_server_params() -> StdioServerParameters:
    return StdioServerParameters(
        command=MATLAB_MCP_COMMAND,
        args=[
            f"--initial-working-folder={WORKING_DIR.resolve()}",
            "--disable-telemetry=true",
            f"--matlab-root={MATLAB_ROOT}",
            f"--matlab-display-mode={MATLAB_DISPLAY_MODE}",
        ],
    )


def _build_framework_tool() -> MCPStdioTool:
    return MCPStdioTool(
        name="matlab",
        command=MATLAB_MCP_COMMAND,
        args=[
            f"--initial-working-folder={WORKING_DIR.resolve()}",
            "--disable-telemetry=true",
            f"--matlab-root={MATLAB_ROOT}",
            f"--matlab-display-mode={MATLAB_DISPLAY_MODE}",
        ],
        allowed_tools=ALLOWED_TOOLS,
        request_timeout=REQUEST_TIMEOUT_SECONDS,
        description="Official MATLAB MCP core server",
    )


def _extract_text_from_call_result(result: Any) -> str:
    contents = getattr(result, "content", None) or []
    parts: list[str] = []
    for item in contents:
        if getattr(item, "type", None) == "text":
            parts.append(getattr(item, "text", ""))
        else:
            parts.append(repr(item))
    if not parts and getattr(result, "structuredContent", None) is not None:
        return json.dumps(result.structuredContent, ensure_ascii=False)
    return "\n".join(part for part in parts if part).strip()


async def _run_direct_mcp_check() -> DirectMCPResult:
    started = monotonic()
    async with stdio_client(_build_server_params()) as (read_stream, write_stream):
        async with ClientSession(
            read_stream,
            write_stream,
            read_timeout_seconds=timedelta(seconds=REQUEST_TIMEOUT_SECONDS),
        ) as session:
            await session.initialize()
            result = await session.call_tool("detect_matlab_toolboxes", {})
    text = _extract_text_from_call_result(result)
    return DirectMCPResult(
        is_error=bool(result.isError),
        elapsed_seconds=monotonic() - started,
        text_excerpt=text[:240].replace("\n", " "),
    )


async def _run_framework_tool_check() -> tuple[str, str, float]:
    tool = _build_framework_tool()
    started = monotonic()
    try:
        await tool.connect()
        result = await tool.call_tool("detect_matlab_toolboxes")
        text = result[0].text if isinstance(result, list) and result else str(result)
        return ("success", text[:240].replace("\n", " "), monotonic() - started)
    except Exception as exc:
        return (type(exc).__name__, str(exc), monotonic() - started)
    finally:
        with suppress(Exception):
            await tool.close()


async def main() -> int:
    _validate_required_config()
    print("[config]")
    print(f"  workdir={WORKING_DIR}")
    print(f"  matlab_mcp_command={MATLAB_MCP_COMMAND}")
    print(f"  timeout_seconds={REQUEST_TIMEOUT_SECONDS}")

    direct = await _run_direct_mcp_check()
    print("[direct-mcp]")
    print(f"  is_error={direct.is_error}")
    print(f"  elapsed_seconds={direct.elapsed_seconds:.2f}")
    print(f"  text_excerpt={direct.text_excerpt!r}")

    wrapper_outcome, wrapper_detail, wrapper_elapsed = await _run_framework_tool_check()
    print("[framework-wrapper]")
    print(f"  outcome={wrapper_outcome}")
    print(f"  elapsed_seconds={wrapper_elapsed:.2f}")
    print(f"  detail={wrapper_detail!r}")

    reproduced = (
        direct.is_error is False
        and wrapper_outcome == "ToolExecutionException"
    )
    print("[summary]")
    print(f"  reproduced={reproduced}")
    return 0 if reproduced else 1


if __name__ == "__main__":
    raise SystemExit(asyncio.run(main()))
```

```markdown
[config]
  workdir=/path/to/workdir
  matlab_mcp_command=path/to/matlab-mcp-core-server
  timeout_seconds=30
{"time":"2026-03-22T03:24:22.084898+08:00","level":"INFO","msg":"Initiating application startup"}
{"time":"2026-03-22T03:24:22.085003+08:00","level":"INFO","msg":"Configuration state","DisableTelemetry":true,"LogLevel":"info","PreferredLocalMATLABRoot":"/Applications/MATLAB_R2025b.app","PreferredMATLABStartingDirectory":"/path/to/workdir","UseSingleMATLABSession":true,"InitializeMATLABOnStartup":false,"MATLABDisplayMode":"nodesktop"}
{"time":"2026-03-22T03:24:22.085011+08:00","level":"INFO","msg":"Application directory state","log-dir":"/tmp/matlab-mcp-core-server-3698481896","id":"4231760144"}
{"time":"2026-03-22T03:24:22.587151+08:00","level":"INFO","msg":"Application startup complete"}
{"time":"2026-03-22T03:24:22.589385+08:00","level":"INFO","msg":"Added tools to MCP SDK server","count":5}
{"time":"2026-03-22T03:24:22.589444+08:00","level":"INFO","msg":"Added additional tools to MCP SDK server","count":0}
{"time":"2026-03-22T03:24:22.589488+08:00","level":"INFO","msg":"Added resources to MCP SDK server","count":2}
{"time":"2026-03-22T03:24:32.006481+08:00","level":"INFO","msg":"Initiating application shutdown"}
{"time":"2026-03-22T03:24:35.291086+08:00","level":"INFO","msg":"Application shutdown complete"}
[direct-mcp]
  is_error=False
  elapsed_seconds=13.25
  text_excerpt='{"installation_info":"--------------------------------------------\\nMATLAB 版本: 25.2.0.3150157 (R2025b) Update 4\\nMATLAB 许可证编号: 40588452\\n操作系统: macOS  Version: 26.3.1 Build: 25D2128 \\nJava 版本: Java is not loaded\\n----------------------------'
{"time":"2026-03-22T03:24:35.343238+08:00","level":"INFO","msg":"Initiating application startup"}
{"time":"2026-03-22T03:24:35.343418+08:00","level":"INFO","msg":"Configuration state","DisableTelemetry":true,"LogLevel":"info","PreferredLocalMATLABRoot":"/Applications/MATLAB_R2025b.app","PreferredMATLABStartingDirectory":"/path/to/workdir","UseSingleMATLABSession":true,"InitializeMATLABOnStartup":false,"MATLABDisplayMode":"nodesktop"}
{"time":"2026-03-22T03:24:35.343433+08:00","level":"INFO","msg":"Application directory state","log-dir":"/tmp/matlab-mcp-core-server-2422630040","id":"3584601140"}
{"time":"2026-03-22T03:24:35.846215+08:00","level":"INFO","msg":"Application startup complete"}
{"time":"2026-03-22T03:24:35.848145+08:00","level":"INFO","msg":"Added tools to MCP SDK server","count":5}
{"time":"2026-03-22T03:24:35.848217+08:00","level":"INFO","msg":"Added additional tools to MCP SDK server","count":0}
{"time":"2026-03-22T03:24:35.848248+08:00","level":"INFO","msg":"Added resources to MCP SDK server","count":2}
{"time":"2026-03-22T03:25:05.857854+08:00","level":"INFO","msg":"Initiating application shutdown"}
[framework-wrapper]
  outcome=ToolExecutionException
  elapsed_seconds=30.53
  detail='Timed out while waiting for response to ClientRequest. Waited 30.0 seconds.'
[summary]
  reproduced=True
```

### Package Versions

agent-framework-core: v1.0.0rc5, mcp v1.26.0, openai v2.26.0

### Python Version

Python 3.13

### Additional Context

_No response_

giles17:
## Problem

When an MCP server (e.g. MATLAB's `matlab-mcp-core-server`) sends a `notifications/tools/list_changed` notification during or after tool execution, `MCPTool.message_handler()` previously **awaited** `load_tools()`/`load_prompts()` directly. Since the handler runs on the MCP SDK's single-threaded `_receive_loop`, this caused a **deadlock**:

1. `message_handler` awaits `load_tools()`
2. `load_tools()` calls `session.list_tools()`, which sends a request and waits for a response
3. The response can never arrive because the receive loop is blocked inside the handler

This manifested as a timeout in `call_tool()`, which then surfaced as `"Error: Function failed."` to the model instead of the real tool output.

## Fix

Schedule reloads as background `asyncio.Task`s via a new `_schedule_reload()` helper method, freeing the receive loop to continue processing responses immediately. Strong references are kept in `_pending_reload_tasks` to prevent garbage collection before completion, and errors are logged rather than propagated into the receive loop.

## Changes

- **`python/packages/core/agent_framework/_mcp.py`**: Add `_pending_reload_tasks` set, `_schedule_reload()` helper; change `message_handler` to use it instead of direct `await`
- **`python/packages/core/tests/core/test_mcp.py`**: Update existing notification test; add regression test for deadlock scenario; add test for reload failure logging

## Testing

All 89 existing MCP tests pass (2 skipped). New tests verify:
- `message_handler` returns immediately even when `load_tools` blocks (deadlock regression)
- Background reload errors are logged, not raised

Fixes #4828 and #4896


markwallace-microsoft:
<a href="https://github.com/microsoft/agent-framework/blob/1c81dec31b7914bd5a3e7f0ba4356cd0b0734277/README.md">Python Test Coverage</a><details><summary>Python Test Coverage Report • </summary><table><tr><th>File</th><th>Stmts</th><th>Miss</th><th>Cover</th><th>Missing</th></tr><tbody><tr><td colspan="5"><b>packages/core/agent_framework</b></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/1c81dec31b7914bd5a3e7f0ba4356cd0b0734277/packages/core/agent_framework/_mcp.py">_mcp.py</a></td><td>654</td><td>29</td><td>95%</td><td><a href="https://github.com/microsoft/agent-framework/blob/1c81dec31b7914bd5a3e7f0ba4356cd0b0734277/packages/core/agent_framework/_mcp.py#L153">153</a>, <a href="https://github.com/microsoft/agent-framework/blob/1c81dec31b7914bd5a3e7f0ba4356cd0b0734277/packages/core/agent_framework/_mcp.py#L172">172</a>, <a href="https://github.com/microsoft/agent-framework/blob/1c81dec31b7914bd5a3e7f0ba4356cd0b0734277/packages/core/agent_framework/_mcp.py#L311">311</a>, <a href="https://github.com/microsoft/agent-framework/blob/1c81dec31b7914bd5a3e7f0ba4356cd0b0734277/packages/core/agent_framework/_mcp.py#L371-L372">371&ndash;372</a>, <a href="https://github.com/microsoft/agent-framework/blob/1c81dec31b7914bd5a3e7f0ba4356cd0b0734277/packages/core/agent_framework/_mcp.py#L501">501</a>, <a href="https://github.com/microsoft/agent-framework/blob/1c81dec31b7914bd5a3e7f0ba4356cd0b0734277/packages/core/agent_framework/_mcp.py#L563">563</a>, <a href="https://github.com/microsoft/agent-framework/blob/1c81dec31b7914bd5a3e7f0ba4356cd0b0734277/packages/core/agent_framework/_mcp.py#L576">576</a>, <a href="https://github.com/microsoft/agent-framework/blob/1c81dec31b7914bd5a3e7f0ba4356cd0b0734277/packages/core/agent_framework/_mcp.py#L600-L601">600&ndash;601</a>, <a href="https://github.com/microsoft/agent-framework/blob/1c81dec31b7914bd5a3e7f0ba4356cd0b0734277/packages/core/agent_framework/_mcp.py#L614-L617">614&ndash;617</a>, <a href="https://github.com/microsoft/agent-framework/blob/1c81dec31b7914bd5a3e7f0ba4356cd0b0734277/packages/core/agent_framework/_mcp.py#L619-L620">619&ndash;620</a>, <a href="https://github.com/microsoft/agent-framework/blob/1c81dec31b7914bd5a3e7f0ba4356cd0b0734277/packages/core/agent_framework/_mcp.py#L624">624</a>, <a href="https://github.com/microsoft/agent-framework/blob/1c81dec31b7914bd5a3e7f0ba4356cd0b0734277/packages/core/agent_framework/_mcp.py#L707-L709">707&ndash;709</a>, <a href="https://github.com/microsoft/agent-framework/blob/1c81dec31b7914bd5a3e7f0ba4356cd0b0734277/packages/core/agent_framework/_mcp.py#L942">942</a>, <a href="https://github.com/microsoft/agent-framework/blob/1c81dec31b7914bd5a3e7f0ba4356cd0b0734277/packages/core/agent_framework/_mcp.py#L1089">1089</a>, <a href="https://github.com/microsoft/agent-framework/blob/1c81dec31b7914bd5a3e7f0ba4356cd0b0734277/packages/core/agent_framework/_mcp.py#L1092">1092</a>, <a href="https://github.com/microsoft/agent-framework/blob/1c81dec31b7914bd5a3e7f0ba4356cd0b0734277/packages/core/agent_framework/_mcp.py#L1118">1118</a>, <a href="https://github.com/microsoft/agent-framework/blob/1c81dec31b7914bd5a3e7f0ba4356cd0b0734277/packages/core/agent_framework/_mcp.py#L1231">1231</a>, <a href="https://github.com/microsoft/agent-framework/blob/1c81dec31b7914bd5a3e7f0ba4356cd0b0734277/packages/core/agent_framework/_mcp.py#L1283-L1284">1283&ndash;1284</a>, <a href="https://github.com/microsoft/agent-framework/blob/1c81dec31b7914bd5a3e7f0ba4356cd0b0734277/packages/core/agent_framework/_mcp.py#L1287">1287</a>, <a href="https://github.com/microsoft/agent-framework/blob/1c81dec31b7914bd5a3e7f0ba4356cd0b0734277/packages/core/agent_framework/_mcp.py#L1758">1758</a></td></tr><tr><td><b>TOTAL</b></td><td><b>33865</b></td><td><b>3909</b></td><td><b>88%</b></td><td>&nbsp;</td></tr></tbody></table></details>

## Python Unit Test Overview
| 6665 | 30 :zzz: | 0 :x: | 0 :fire: | 1m 46s :stopwatch: |

python/packages/core/tests/core/test_mcp.py
```diff
@@ -1527,6 +1531,61 @@ async def test_mcp_tool_message_handler_error():
...
+    tool.load_tools = AsyncMock(side_effect=RuntimeError("connection lost"))
+
+    tools_notification = Mock(spec=types.ServerNotification)
+    tools_notification.root = Mock()
+    tools_notification.root.method = "notifications/tools/list_changed"
+
+    await tool.message_handler(tools_notification)
+    # Let the background task run — it should not propagate the exception.
+    await asyncio.sleep(0)
+    tool.load_tools.assert_called_once()
```
giles17, line 1726: This test verifies the reload was called but doesn't assert that `_pending_reload_tasks` is empty after the failed task completes. Adding `assert len(tool._pending_reload_tasks) == 0` would confirm the done-callback cleanup works on the error path, matching the pattern in the deadlock test.

```suggestion
    tool.load_tools.assert_called_once()
    assert len(tool._pending_reload_tasks) == 0
```

python/packages/core/agent_framework/_mcp.py
```diff
@@ -804,12 +805,37 @@ async def message_handler(
...
                 case "notifications/tools/list_changed":
-                    await self.load_tools()
+                    self._schedule_reload(self.load_tools())
                 case "notifications/prompts/list_changed":
-                    await self.load_prompts()
+                    self._schedule_reload(self.load_prompts())
                 case _:
                     logger.debug("Unhandled notification: %s", message.root.method)
 
+    def _schedule_reload(self, coro: Any) -> None:
```
giles17: `coro: Any` is looser than needed here. Since the method only accepts awaitables, tightening to `Coroutine[Any, Any, None]` (from `collections.abc`) lets type checkers flag accidental misuse.

```suggestion
    def _schedule_reload(self, coro: "Coroutine[Any, Any, None]") -> None:
```

giles17 (commented):
## Automated Code Review

### ✓ Correctness
> The fix correctly addresses a real deadlock where `message_handler` (running on the MCP SDK's single-threaded receive loop) previously awaited `load_tools()`/`load_prompts()` which themselves need the receive loop to process responses. The solution uses `asyncio.create_task()` to schedule reloads as background tasks, with proper strong-reference tracking via `_pending_reload_tasks` and a done-callback for cleanup. Exception handling in the background task prevents errors from propagating to the receive loop. Tests cover the non-blocking behavior, task lifecycle, and error logging. The implementation follows standard asyncio fire-and-forget patterns correctly.

### ✓ Security Reliability
> The fix correctly addresses a real deadlock by scheduling reload coroutines as background tasks instead of awaiting them inside the MCP SDK's single-threaded receive loop. The implementation is sound: exceptions are caught and logged, tasks are prevented from GC via a strong-reference set, and the done-callback cleanup pattern is correct. The one reliability concern is that `close()` / `_close_on_owner()` does not cancel or await pending reload tasks before tearing down the session, which could cause reload tasks to run against a `None` session and produce confusing warnings or mask real errors during shutdown.

### ✓ Test Coverage
> The test coverage for this fix is good. The PR adds three well-targeted tests: the existing notification test is updated to account for the async scheduling (with `asyncio.sleep(0)`), a new deadlock regression test uses an `asyncio.Event` to verify the handler returns immediately even when the reload blocks, and a failure-logging test confirms exceptions don't propagate. However, two minor gaps exist: the reload-failure test doesn't verify the task is cleaned up from `_pending_reload_tasks` after the error, and the `load_prompts` notification path lacks a dedicated non-blocking test (only `load_tools` is tested in the deadlock scenario).

### ✓ Design Approach
> The fix correctly addresses a real deadlock: `message_handler` ran on the MCP SDK's single-threaded receive loop and previously awaited `load_tools()`/`load_prompts()`, which in turn called `session.list_tools()` — a request whose response could never arrive because the receive loop was blocked. Scheduling reloads as background tasks via `asyncio.create_task()` with a strong-reference set is the right approach. The tests are well-structured, especially the event-based non-blocking regression test. One design concern: if the server sends a burst of `notifications/tools/list_changed` notifications, multiple concurrent reload tasks are each created, all capturing `existing_names` as a local snapshot of `self._functions` at task-start time. Two concurrent tasks can race to append the same newly-advertised tool (the `existing_names` dedup check is local, not atomic with the append), producing duplicate entries in `_functions`. Additionally, `coro: Any` is looser than necessary; `Coroutine[Any, Any, None]` would let type checkers catch misuse. Neither blocks the merge — the duplicate-tool race is an edge case unlikely to cause user-visible harm, and the type annotation is minor — but both are worth a follow-up.

### Suggestions
- In `test_mcp_tool_message_handler_reload_failure_is_logged`, assert that `_pending_reload_tasks` is empty after the failed task completes (e.g. `assert len(tool._pending_reload_tasks) == 0`) to verify the done-callback cleanup works on the error path, consistent with the pattern in the deadlock test.
- Tighten the `coro: Any` parameter type in `_schedule_reload` to `Coroutine[Any, Any, None]` (from `collections.abc`) so type checkers can flag accidental misuse, even though usage is internal-only.
- Consider adding the `prompts/list_changed` notification to `test_mcp_tool_message_handler_does_not_block_receive_loop` (or a sibling test) to ensure both notification paths are covered for non-blocking behavior, not just `tools/list_changed`.

---

python/packages/core/agent_framework/_mcp.py
```diff
@@ -799,17 +800,51 @@ async def message_handler(
             message: The message from the MCP server (request responder, notification, or exception).
         """
         if isinstance(message, Exception):
-            logger.error("Error from MCP server: %s", message, exc_info=message)
+            logger.error("Error from MCP server: %s", message, exc_info=True)
```
giles17: Bug: `message_handler` is called by the MCP SDK with the exception as a parameter—NOT from within an `except` block (see `mcp/shared/session.py:359` and `:516`). With `exc_info=True`, `sys.exc_info()` returns `(None, None, None)`, so the exception type and traceback are silently lost. The original `exc_info=message` correctly passes the exception instance, letting Python's logging extract the full traceback directly.

```suggestion
            logger.error("Error from MCP server: %s", message, exc_info=message)
```

python/packages/core/agent_framework/_mcp.py
```diff
@@ -931,6 +966,11 @@ async def load_tools(self) -> None:
             params = types.PaginatedRequestParams(cursor=tool_list.nextCursor)
 
     async def _close_on_owner(self) -> None:
+        # Cancel any pending reload tasks before tearing down the session.
+        for task in list(self._pending_reload_tasks):
+            task.cancel()
+        self._pending_reload_tasks.clear()
```
giles17, line 1090: Reliability concern: `task.cancel()` only *requests* cancellation. Without awaiting the tasks, they may still be in-flight when `_safe_close_exit_stack()` nullifies `self.session`, and Python may emit 'Task was destroyed but it is pending' warnings. Consider gathering the cancelled tasks before proceeding.

```suggestion
        tasks = list(self._pending_reload_tasks)
        for task in tasks:
            task.cancel()
        self._pending_reload_tasks.clear()
        if tasks:
            await asyncio.gather(*tasks, return_exceptions=True)
```

python/packages/core/tests/core/test_mcp.py
```diff
@@ -1527,6 +1531,71 @@ async def test_mcp_tool_message_handler_error():
...
+    assert len(tool._pending_reload_tasks) == 0
+
+    # Verify the warning was actually logged with exception info.
+    reload_warnings = [r for r in caplog.records if "Background MCP reload failed" in r.message]
+    assert len(reload_warnings) == 1
+    assert reload_warnings[0].levelname == "WARNING"
+    assert reload_warnings[0].exc_info is not None
+
+
```
giles17, line 1735: Consider adding a companion test for the cancel-and-replace logic: send two `notifications/tools/list_changed` in quick succession with a blocking `load_tools` (via `asyncio.Event`), and assert that the first task gets cancelled and only one task remains in `_pending_reload_tasks`. This is the only way to verify the `existing.cancel()` path and the task-naming scheme work correctly.

python/packages/core/tests/core/test_mcp.py
```diff
@@ -1527,6 +1531,71 @@ async def test_mcp_tool_message_handler_error():
...
+    # Unblock the reload so the background task finishes cleanly.
+    release.set()
+    # Wait for the pending reload task(s) to complete so their done-callbacks
+    # have a chance to remove them from _pending_reload_tasks.
+    await asyncio.wait_for(asyncio.gather(*tool._pending_reload_tasks), timeout=1)
+    assert len(tool._pending_reload_tasks) == 0
+
+
+async def test_mcp_tool_message_handler_reload_failure_is_logged(caplog: pytest.LogCaptureFixture):
```
giles17, line 1711: This test verifies the exception doesn't propagate, but (in its current on-disk form without `caplog`) doesn't assert the warning was actually logged. The diff version with `caplog` and log-record assertions is a meaningful improvement—make sure that version is what ships. Without the log assertion, a regression that silently swallows errors without logging would go undetected.

giles17 (commented):
## Automated Code Review

**Reviewers:** 4 | **Confidence:** 91%

### ✗ Correctness
> The core deadlock fix is correct and well-motivated: scheduling reloads as background tasks via `asyncio.create_task` instead of awaiting them inline in `message_handler` prevents the classic deadlock on the MCP SDK's single-threaded receive loop. The cancel-and-replace logic, strong-reference set, and done-callback cleanup are all sound. Tests are comprehensive. However, the `exc_info=message` → `exc_info=True` change on the error-logging line is a regression: because `message_handler` is NOT called from within an `except` block (the MCP SDK passes the exception as a parameter at lines 358 and 516 of `shared/session.py`), `sys.exc_info()` returns `(None, None, None)` and the exception type/traceback info is lost. The original `exc_info=message` correctly lets Python's logging extract `(type(message), message, message.__traceback__)` from the exception instance directly.

### ✗ Security Reliability
> The core fix — scheduling MCP reloads as background tasks via `asyncio.create_task` instead of awaiting them inline in `message_handler` — is correct and addresses a real deadlock. However, the change from `exc_info=message` to `exc_info=True` on the error-logging line is a regression: `message_handler` is not inside an `except` block, so `sys.exc_info()` (used when `exc_info=True`) returns `(None, None, None)`, silently discarding the exception's traceback. The original `exc_info=message` correctly passes the exception instance, which Python's logging (3.5+) converts to a full traceback tuple. Additionally, `_close_on_owner` cancels pending reload tasks but does not await them, which can produce 'Task was destroyed but it is pending' warnings and allows cancelled tasks to race against session teardown.

### ✓ Test Coverage
> The PR adds three new/updated tests covering the core deadlock fix: non-blocking behavior of `message_handler`, background task scheduling, and reload-failure logging. These tests are well-structured and meaningfully validate the primary fix. However, the new cancel-and-replace logic in `_schedule_reload` (which cancels a previous pending reload of the same kind when a new notification arrives) and the pending-task cleanup in `_close_on_owner` both lack dedicated test coverage. These are behavioral additions that could harbor bugs (e.g., incorrect task name comparison) without any test exercising them.

### ✗ Design Approach
> The core fix — scheduling reloads as background tasks via `asyncio.create_task` to avoid deadlocking the MCP receive loop — is correct and well-motivated. The cancel-and-replace deduplication, the `_pending_reload_tasks` strong-reference set, the explicit `CancelledError` re-raise, and the teardown cleanup in `_close_on_owner` are all sound improvements. However, the diff introduces one clear regression: changing `exc_info=message` to `exc_info=True` in the exception-logging branch of `message_handler`. Since that code is not inside an `except` block, `exc_info=True` resolves to `sys.exc_info()` which returns `(None, None, None)` — silently dropping the exception traceback. The original `exc_info=message` correctly passes the exception object directly to the logger, which Python 3.x logging accepts and uses to attach the traceback. This should be reverted.

### Flagged Issues
- [ ] Changing `exc_info=message` to `exc_info=True` on the error-logging line (line 803) loses exception information. `message_handler` is called by the MCP SDK with the exception as a parameter—NOT from within an `except` block (see `mcp/shared/session.py` lines 358, 516)—so `sys.exc_info()` returns `(None, None, None)` and no traceback is ever logged. The original `exc_info=message` was correct: Python 3.x logging accepts an exception instance and extracts `(type, value, __traceback__)` from it. Revert this line.

### Suggestions
- In `_close_on_owner`, await the cancelled reload tasks (e.g. `await asyncio.gather(*tasks, return_exceptions=True)`) before proceeding to session teardown. Without this, cancelled tasks may still be in-flight when `self.session` is set to `None`, producing 'Task was destroyed but it is pending' warnings and a subtle race condition.
- Add a test for the cancel-and-replace behavior: send two rapid `notifications/tools/list_changed` notifications with a blocking `load_tools` (via `asyncio.Event`), verify the first task is cancelled and only one task remains pending. This exercises the `existing.cancel()` path and the task-naming scheme in `_schedule_reload`.
- Add a test for `_close_on_owner` cleanup: create an MCPTool with a pending reload task, call `_close_on_owner`, and verify the task is cancelled and `_pending_reload_tasks` is empty. The existing test mocks the entire method, so the new cancellation code has no coverage.
- Ensure the `caplog`-based version of `test_mcp_tool_message_handler_reload_failure_is_logged` is what ships. The on-disk version without `caplog` only checks the mock was called and doesn't assert the warning was actually logged, so a regression that silently swallows errors would go undetected.

---

ani1797:
Hi could you provide an update on this PR? Following up as we have a production process depending on this fix.

python/packages/core/agent_framework/_mcp.py
```diff
@@ -905,12 +906,46 @@ async def message_handler(
...
+        a strong reference in ``_pending_reload_tasks`` so it is not garbage-
+        collected before completion.  Only one reload per kind (tools / prompts)
+        is kept in flight; a new notification cancels the previous pending task
+        for the same coroutine name to avoid unbounded growth.
+        """
+        # Cancel-and-replace: only one reload per kind should be in flight.
+        reload_name = f"mcp-reload:{self.name}:{coro.__qualname__}"
+        for existing in list(self._pending_reload_tasks):
+            if existing.get_name() == reload_name and not existing.done():
+                existing.cancel()
```
TaoChenOSU, line 936: nit: should there be a warning if a new one arrives before the previous one finishes?

TaoChenOSU (approved):

moonbox3 (approved):

TaoChenOSU (approved):

Issue #4422, opened 2026-03-03
.NET: [Feature]: Agentskills load_skill should include full paths to additional files in addition to the body text


rwjdk:
### Description

When you have an AgentSkill with additional references or scripts that you refer to in the skill body, there is no way currently to know the full path of the skill's additional files (references work, but not other files)

This results in that if you example have a script or other non-standard file that you only have the relative path to (from body) and not the full path. 

A way around this would be that to load_skill would besides the body also include a list of all full paths for additional references and other files so AI can infer the full path from the relative path in the body

This video show the limitation in action: https://youtu.be/5PW8FzYiHJc?si=fLXVL0Zutl7XdmQ7 (+ various other limitations that hinder adoption of this implementation over other implementations of skills

### Code Sample

```markdown

```

### Language/SDK

.NET

joslat:
I stand by this, the analysis is solid.
I fed the spec to an AI and asked it for a concise summary which I paste the main highlights below:

## Summary
Microsoft Agent Framework (MAF) RC3 ships FileAgentSkillsProvider as its official Agent Skills implementation. The implementation is production-grade in its core scope — progressive disclosure, security, logging, and resource discovery are excellent — but it narrows its focus to read-only, document-based skills and stays firmly sealed against extension.

Key RC3 improvement over RC2: Resource discovery now scans the skill directory by file extension (configurable via AllowedResourceExtensions) rather than extracting markdown links from the SKILL.md body. This is more robust and aligns better with the spec's "resources are files in the directory" model.

However, several features defined by the Agent Skills spec or needed by real-world skill consumers (script execution, metadata exposure, subset selection, custom tool executors) are either absent or very limited. The classes are all sealed and internal, making extensibility essentially impossible without a full rewrite.

Overall weighted score: 5.8 / 10 — Strong foundation, significant gaps.

## Feature Assessment Table

| # | Feature | Description | Score (0–10) |
|---|---------|-------------|:------------:|
| 1 | **Instruction Configurability** | Can the system prompt / instructions template be customized? | **7** |
| 2 | **Delivery Method (AI Context Provider)** | How are skills injected into the agent pipeline? | **9** |
| 3 | **Tool Pluggability** | Can consumers plug in their own tool implementations? | **2** |
| 4 | **Logging** | Standard logging or custom objects? | **10** |
| 5 | **Skill Validation (Spec Compliance)** | Does validation follow the agentskills.io standard? Strict or lenient? | **6** |
[...]

PR #4871, opened 2026-03-23
.NET: [Breaking] Restructure agent skills to use multi-source architecture

SergeyMenshykh:
## Summary

Restructures the .NET agent skills system to use the new multi-source architecture as proposed in [ADR-0021](https://github.com/microsoft/agent-framework/pull/4787). This introduces the **sources building blocks** — a set of composable abstractions for discovering, filtering, caching, and deduplicating skills from multiple origins.

### New abstractions
- **`AgentSkillsSource`** — abstract base for skill discovery from any origin (filesystem, in-memory, custom)
- **`AgentSkillResource`** / **`AgentSkillScript`** — abstract base types for resources and scripts
- **`AgentSkillFrontmatter`** — dedicated model for skill metadata (name, description, license, compatibility, etc.)

### Source implementations
- **`AgentFileSkillsSource`** — discovers file-based skills from `SKILL.md` files on disk (replaces `FileAgentSkillsProvider` discovery logic)
- **`AgentInMemorySkillsSource`** — holds any `AgentSkill` instances in memory (for inline/class-based skills)
- **`AggregateAgentSkillsSource`** — composes multiple sources into one

### Source decorators
- **`DelegatingAgentSkillsSource`** — abstract base for intercepting `GetSkillsAsync()`
- **`FilteringAgentSkillsSource`** — applies predicate-based skill filtering
- **`CachingAgentSkillsSource`** — caches skills after first load
- **`DeduplicatingAgentSkillsSource`** — deduplicates by name (case-insensitive, first-wins)

### File-based skill types
- **`AgentFileSkill`** — replaces `FileAgentSkill` with clearer naming and source-oriented design
- **`AgentFileSkillResource`** — reads content from files on disk
- **`AgentFileSkillScript`** — delegates execution to a configurable `AgentFileSkillScriptExecutor`

### Provider and builder
- **`AgentSkillsProvider`** — updated to accept a single `AgentSkillsSource` (composed externally)
- **`AgentSkillsProviderBuilder`** — fluent API for composing skills from multiple source types

### Sample updates
- Renamed `Agent_Step01_BasicSkills` → `Agent_Step01_FileBasedSkills` with updated skill content
- Added `SubprocessScriptExecutor` helper for running external scripts

Closes: https://github.com/microsoft/agent-framework/issues/4422

## Related
- ADR: https://github.com/microsoft/agent-framework/pull/4787

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI/Skills/AgentSkillsProviderBuilder.cs
+++ b/dotnet/src/Microsoft.Agents.AI/Skills/AgentSkillsProviderBuilder.cs
@@ -86,7 +86,7 @@ public sealed class AgentSkillsProviderBuilder
     /// <returns>This builder instance for chaining.</returns>
     public AgentSkillsProviderBuilder UsePromptTemplate(string promptTemplate)
     {
-        this.EnsureOptions().SkillsInstructionPrompt = promptTemplate;
+        this.GetOrCreateOptions().SkillsInstructionPrompt = promptTemplate;
         return this;
     }
 
@@ -97,7 +97,7 @@ public sealed class AgentSkillsProviderBuilder
     /// <returns>This builder instance for chaining.</returns>
     public AgentSkillsProviderBuilder UseScriptApproval(bool enabled = true)
     {
-        this.EnsureOptions().ScriptApproval = enabled;
+        this.GetOrCreateOptions().ScriptApproval = enabled;
         return this;
     }
 
@@ -148,7 +148,7 @@ public sealed class AgentSkillsProviderBuilder
     public AgentSkillsProviderBuilder UseOptions(Action<AgentSkillsProviderOptions> configure)
     {
         _ = Throw.IfNull(configure);
-        configure(this.EnsureOptions());
+        configure(this.GetOrCreateOptions());
         return this;
     }
 
@@ -185,13 +185,8 @@ public sealed class AgentSkillsProviderBuilder
         return new AgentSkillsProvider(source, this._options, this._loggerFactory);
     }
 
-    private AgentSkillsProviderOptions EnsureOptions()
+    private AgentSkillsProviderOptions GetOrCreateOptions()
     {
-        if (this._options == null)
-        {
-            this._options = new AgentSkillsProviderOptions();
-        }
-
-        return this._options;
+        return this._options ??= new AgentSkillsProviderOptions();
     }
 }
--- a/dotnet/src/Microsoft.Agents.AI/Skills/AggregatingAgentSkillsSource.cs
+++ b/dotnet/src/Microsoft.Agents.AI/Skills/AggregatingAgentSkillsSource.cs
@@ -27,8 +27,7 @@ internal sealed class AggregatingAgentSkillsSource : AgentSkillsSource
     /// <param name="sources">The child sources to aggregate.</param>
     public AggregatingAgentSkillsSource(IEnumerable<AgentSkillsSource> sources)
     {
-        _ = Throw.IfNull(sources);
-        this._sources = sources;
+        this._sources = Throw.IfNull(sources);
     }
 
     /// <inheritdoc/>


dotnet/src/Microsoft.Agents.AI/Skills/AgentSkillsProvider.cs
```diff
@@ -0,0 +1,278 @@
...
+                name: "load_skill",
+                description: "Loads the full content of a specific skill"),
+            AIFunctionFactory.Create(
+                (string skillName, string resourceName, IServiceProvider? serviceProvider, CancellationToken cancellationToken = default) =>
+                    this.ReadSkillResourceAsync(skills, skillName, resourceName, serviceProvider, cancellationToken),
+                name: "read_skill_resource",
+                description: "Reads a resource associated with a skill, such as references, assets, or dynamic data."),
+        ];
+
+        if (!hasScripts)
```
westey-m, line 210: Should there be a similar check for resources?

dotnet/src/Microsoft.Agents.AI/Skills/AggregateAgentSkillsSource.cs
```diff
@@ -0,0 +1,46 @@
...
+
+/// <summary>
+/// A skill source that aggregates multiple child sources, preserving their registration order.
+/// </summary>
+/// <remarks>
+/// Skills from each child source are returned in the order the sources were registered,
+/// with each source's skills appended sequentially. No deduplication or filtering is applied.
+/// </remarks>
+[Experimental(DiagnosticIds.Experiments.AgentsAIExperiments)]
+public sealed class AggregateAgentSkillsSource : AgentSkillsSource
```
westey-m: The other sources use a present continuous tense verb, and I think this should too, e.g. `AggregatingAgentSkillsSource `

dotnet/samples/02-agents/AgentSkills/SubprocessScriptRunner.cs
```diff
@@ -0,0 +1,135 @@
...
+
+/// <summary>
+/// Executes file-based skill scripts as local subprocesses.
+/// </summary>
+/// <remarks>
+/// This runner uses the script's absolute path, converts the arguments
+/// to CLI flags, and returns captured output. It is intended for
+/// demonstration purposes only.
+/// </remarks>
+internal static class SubprocessScriptRunner
```
rogerbarreto, line 19: I'm not fully found of "hiding" this complexity from the samples is there any way of having a simpler/slim version of the runner for the individual samples?

dotnet/src/Microsoft.Agents.AI/Skills/AgentSkillsProvider.cs
```diff
@@ -56,18 +61,65 @@ You have access to skills containing domain-specific knowledge and capabilities.
...
+    /// Duplicate skill names are automatically deduplicated (first occurrence wins).
+    /// </summary>
+    /// <param name="skillPath">Path to search for skills.</param>
+    /// <param name="scriptRunner">The delegate that runs file-based scripts.</param>
+    /// <param name="fileOptions">Optional options that control skill discovery behavior.</param>
+    /// <param name="options">Optional provider configuration.</param>
+    /// <param name="loggerFactory">Optional logger factory.</param>
+    public AgentSkillsProvider(
+        string skillPath,
+        AgentFileSkillScriptRunner scriptRunner,
```
westey-m: Does this need to be required, if the skill has no scripts?

dotnet/src/Microsoft.Agents.AI/Skills/File/AgentFileSkill.cs
```diff
@@ -0,0 +1,57 @@
...
+    /// <param name="scripts">Scripts discovered for this skill.</param>
+    internal AgentFileSkill(
+        AgentSkillFrontmatter frontmatter,
+        string content,
+        string path,
+        IReadOnlyList<AgentSkillResource>? resources = null,
+        IReadOnlyList<AgentSkillScript>? scripts = null)
+    {
+        this.Frontmatter = Throw.IfNull(frontmatter);
+        this.Content = Throw.IfNull(content);
```
rogerbarreto, line 35: Should we allow empty / white spaced skills?

```suggestion
        this.Content = Throw.IfNullOrWhitespace(content);
```

dotnet/samples/02-agents/AgentSkills/SubprocessScriptRunner.cs
SergeyMenshykh: We can't compress it into a few lines, unfortunately, and will have to include it in each sample. There will be at least one introduced in a follow-up PR.

dotnet/src/Microsoft.Agents.AI/Skills/AgentSkillsProvider.cs
SergeyMenshykh: It was optional, but during one of the reviews, it was suggested to make it required. I can make it optional to support cases where skills have no scripts, and throw an exception if skills have scripts but no runner is provided.

dotnet/src/Microsoft.Agents.AI/Skills/File/AgentFileSkill.cs
```diff
@@ -0,0 +1,57 @@
...
+        this.Path = Throw.IfNullOrWhitespace(path);
+        this._resources = resources ?? [];
+        this._scripts = scripts ?? [];
+    }
+
+    /// <inheritdoc/>
+    public override AgentSkillFrontmatter Frontmatter { get; }
+
+    /// <inheritdoc/>
+    public override string Content { get; }
```
rogerbarreto, line 45: Do we have a strong case to make those abstract instead of virtual in `base`. 

If not having the flexibility by default being plain getter might be better thoughts?

```suggestion
```

dotnet/src/Microsoft.Agents.AI/Skills/File/AgentFileSkillResource.cs
```diff
@@ -0,0 +1,47 @@
...
+    /// Gets the absolute file path to the resource.
+    /// </summary>
+    public string FullPath { get; }
+
+    /// <inheritdoc/>
+#pragma warning disable CA1725 // Parameter names should match base declaration
+#pragma warning disable RCS1168 // Parameter name differs from base name
+    public override async Task<object?> ReadAsync(IServiceProvider? _, CancellationToken cancellationToken = default)
+#pragma warning restore RCS1168 // Parameter name differs from base name
+#pragma warning restore CA1725 // Parameter names should match base declaration
```
rogerbarreto: Might just worth using the base param name than adding 4 pragma lines around it, do we have any real IL gains with this change?

```suggestion
    /// <inheritdoc/>
    public override async Task<object?> ReadAsync(IServiceProvider? baseName, CancellationToken cancellationToken = default)
```

dotnet/src/Microsoft.Agents.AI/Skills/AgentSkillsProvider.cs
SergeyMenshykh: now it's optional.

dotnet/src/Microsoft.Agents.AI/Skills/File/AgentFileSkillResource.cs
SergeyMenshykh: Yep, we can easily remove them.

dotnet/src/Microsoft.Agents.AI/Skills/File/AgentFileSkill.cs
SergeyMenshykh: Those are mandatory and has to be provided by derivatives, while the other two: Resources and Scripts are not so they can be turned into virtual ones.

SergeyMenshykh: done

dotnet/samples/02-agents/AgentSkills/Agent_Step01_FileBasedSkills/Program.cs
```diff
@@ -0,0 +1,48 @@
...
+    .AsAIAgent(new ChatClientAgentOptions
+    {
+        Name = "UnitConverterAgent",
+        ChatOptions = new()
+        {
+            Instructions = "You are a helpful assistant that can convert units.",
+        },
+        AIContextProviders = [skillsProvider],
+    },
+    model: deploymentName);
```
rogerbarreto, line 39: New feature opportunity here @westey-m. 

Since `ContextProviders` became a first citizen feature I think it becomes apparent that we could have `AsAIAgent(contextProviders: []) as an optional argument, instead of having to implement a full `CCAO`.


```suggestion
    .AsAIAgent(
        model: deploymentName,
        name: "UnitConverterAgent",
        instructions: "You are a helpful assistant that can convert units.",
        contextProviders: [skillsProvider]);
```

dotnet/samples/02-agents/AgentSkills/SubprocessScriptRunner.cs
```diff
@@ -0,0 +1,135 @@
...
+        try
+        {
+            using var process = Process.Start(startInfo);
+            if (process is null)
+            {
+                return $"Error: Failed to start process for script '{script.Name}'.";
+            }
+
+            try
+            {
```
rogerbarreto: Do we need a second try? We already have the process to kill here..

dotnet/samples/02-agents/AgentSkills/SubprocessScriptRunner.cs
```diff
@@ -0,0 +1,135 @@
...
+    /// <summary>
+    /// Runs a skill script as a local subprocess.
+    /// </summary>
+    public static async Task<object?> RunAsync(
+        AgentFileSkill skill,
+        AgentFileSkillScript script,
+        AIFunctionArguments arguments,
+        CancellationToken cancellationToken)
+    {
+        if (!File.Exists(script.FullPath))
```
rogerbarreto, line 30: I'm afraid this could raise some vulnerability flags for canonical paths, have we considered those disclaimers, does the `NOT FOR PRODUCTION` enough for not having to guardrail?

SergeyMenshykh: I'm afraid as well, but we don't know whether it'll be flagged or not and it's unknown if the NOT FOR PRODUCITON is enough or not until we try it.

dotnet/src/Microsoft.Agents.AI/Skills/Decorators/DeduplicatingAgentSkillsSource.cs
```diff
@@ -0,0 +1,58 @@
...
+using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.Shared.DiagnosticIds;
+
+namespace Microsoft.Agents.AI;
+
+/// <summary>
+/// A skill source decorator that removes duplicate skills by name, keeping only the first occurrence.
+/// </summary>
+[Experimental(DiagnosticIds.Experiments.AgentsAIExperiments)]
+internal sealed partial class DeduplicatingAgentSkillsSource : DelegatingAgentSkillsSource
```
rogerbarreto, line 18: nit. I'm more used to the `Distict` naming (i.e LINQ) for this feature than `Deduplicate`. Not sure what is the AI convention for this if there's any. Thoughts?

```suggestion
internal sealed partial class DistinctAgentSkillsSource : DelegatingAgentSkillsSource
```

dotnet/src/Microsoft.Agents.AI/Skills/Decorators/FilteringAgentSkillsSource.cs
```diff
@@ -0,0 +1,70 @@
...
+    /// </summary>
+    /// <param name="innerSource">The inner source whose skills will be filtered.</param>
+    /// <param name="predicate">
+    /// A predicate that determines which skills to include. Skills for which the predicate
+    /// returns <see langword="true"/> are kept; others are excluded.
+    /// </param>
+    /// <param name="loggerFactory">Optional logger factory.</param>
+    public FilteringAgentSkillsSource(
+        AgentSkillsSource innerSource,
+        Func<AgentSkill, bool> predicate,
```
rogerbarreto, line 39: Did I read this doc wrong, mentioned skills in the plural, a single filter will run for a single skill? Not sure I follow, do I need one filter for each skill? 

```suggestion
        Func<List<(AgentSkill, bool)>> predicate,
```

dotnet/src/Microsoft.Agents.AI/Skills/Decorators/DeduplicatingAgentSkillsSource.cs
SergeyMenshykh: Distinct in this context means that this data source is separate from other data sources, while deduplicating means that it is responsible for removing duplicates.

dotnet/samples/02-agents/AgentSkills/SubprocessScriptRunner.cs
SergeyMenshykh: Fixed.

dotnet/src/Microsoft.Agents.AI/Skills/File/AgentFileSkill.cs
rogerbarreto: Marked as resolved, but I still see `this.Content = Throw.IfNull(content)`.

dotnet/src/Microsoft.Agents.AI/Skills/Decorators/FilteringAgentSkillsSource.cs
SergeyMenshykh: Nope, there's only one predicate that sequentially applied to each of the skill from the inner source.

dotnet/src/Microsoft.Agents.AI/Skills/Decorators/DelegatingAgentSkillsSource.cs
```diff
@@ -0,0 +1,41 @@
...
+
+    /// <summary>
+    /// Gets the inner skill source that receives delegated operations.
+    /// </summary>
+    protected AgentSkillsSource InnerSource { get; }
+
+    /// <inheritdoc/>
+    public override Task<IList<AgentSkill>> GetSkillsAsync(CancellationToken cancellationToken = default)
+        => this.InnerSource.GetSkillsAsync(cancellationToken);
+}
```
rogerbarreto, line 41: Since we are using `ILogger` in all the implementations might worth having the

`LoggingAgentSkillSource(ILogger, sksrc) : DelegatingAgentSkillSource(sksrc)`

Similar to Agents + ChatClients

dotnet/src/Microsoft.Agents.AI/Skills/File/AgentFileSkillScript.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            throw new InvalidOperationException($"File-based script '{this.Name}' requires an {nameof(AgentFileSkill)} but received '{skill.GetType().Name}'.");
+        }
+
+        if (this._runner is null)
+        {
+            throw new InvalidOperationException(
+                $"Script '{this.Name}' cannot be executed because no {nameof(AgentFileSkillScriptRunner)} was provided. " +
+                $"Supply a script runner when constructing {nameof(AgentFileSkillsSource)} to enable script execution.");
+        }
```
rogerbarreto, line 52: If runner is always required for RunAsync why the constructor takes it as optional?

SergeyMenshykh: File skills may not have scripts, and as a result, the runner is not needed; thus, it's optional.

dotnet/src/Microsoft.Agents.AI/Skills/File/AgentFileSkillScript.cs
```diff
@@ -0,0 +1,56 @@
...
+    /// <summary>
+    /// Initializes a new instance of the <see cref="AgentFileSkillScript"/> class.
+    /// </summary>
+    /// <param name="name">The script name.</param>
+    /// <param name="fullPath">The absolute file path to the script.</param>
+    /// <param name="runner">Optional external runner for running the script. An <see cref="InvalidOperationException"/> is thrown from <see cref="RunAsync"/> if no runner is provided.</param>
+    internal AgentFileSkillScript(string name, string fullPath, AgentFileSkillScriptRunner? runner = null)
+        : base(name)
+    {
+        this.FullPath = Throw.IfNullOrWhitespace(fullPath);
```
rogerbarreto, line 30: Should we check for File.Exists() in ctor / runasync

```suggestion
        this.FullPath = Throw.IfNullOrWhitespace(fullPath);
        !File.Exists(fullPath) { new IOEX }
```

dotnet/src/Microsoft.Agents.AI/Skills/AgentSkillFrontmatter.cs
```diff
@@ -0,0 +1,196 @@
...
+    internal const int MaxDescriptionLength = 1024;
+
+    /// <summary>
+    /// Maximum allowed length for the compatibility field.
+    /// </summary>
+    internal const int MaxCompatibilityLength = 500;
+
+    // Validates skill names: lowercase letters, numbers, and hyphens only;
+    // must not start or end with a hyphen; must not contain consecutive hyphens.
+    private static readonly Regex s_validNameRegex = new("^[a-z0-9]([a-z0-9]*-[a-z0-9])*[a-z0-9]*$", RegexOptions.Compiled);
```
rogerbarreto, line 46: Is this Regex part of the convention? 
If not, I would not enforce it or at least leave configurable.
If yes, make this a with a pointer to the doc (similar to line 58)

dotnet/src/Microsoft.Agents.AI/Skills/AgentSkillsProvider.cs
```diff
@@ -0,0 +1,383 @@
...
+
+        string scriptInstruction = includeScriptInstructions
+            ? "- Use `run_skill_script` to run referenced scripts, using the name exactly as listed."
+            : string.Empty;
+
+        return new StringBuilder(promptTemplate)
+            .Replace(SkillsPlaceholder, sb.ToString().TrimEnd())
+            .Replace(ResourceInstructionsPlaceholder, resourceInstruction)
+            .Replace(ScriptInstructionsPlaceholder, scriptInstruction)
+            .ToString();
```
rogerbarreto, line 257: I would refactor this to instead of replace, use a additive, concatenated, prompt building this could prevent injection of place holders and some other extravagant injections.

dotnet/src/Microsoft.Agents.AI/Skills/AgentSkillsProvider.cs
```diff
@@ -0,0 +1,383 @@
...
+
+        var script = skill.Scripts?.FirstOrDefault(resource => resource.Name == scriptName);
+        if (script is null)
+        {
+            return $"Error: Script '{scriptName}' not found in skill '{skillName}'.";
+        }
+
+        try
+        {
+            return await script.RunAsync(skill, new AIFunctionArguments(arguments) { Services = serviceProvider }, cancellationToken).ConfigureAwait(false);
```
rogerbarreto, line 339: Any special reason why the `script.RunAsync` has to accept a `AIFunctionArgument` and the `RunSkillScriptAsync` and vice versa, in theory both could have Dic + IServiceCollection.

```suggestion
            return await script.RunAsync(skill, arguments, serviceProvider, cancellationToken).ConfigureAwait(false);
```

dotnet/src/Microsoft.Agents.AI/Skills/AgentSkillsProviderBuilder.cs
```diff
@@ -0,0 +1,197 @@
...
+    /// Sets a filter predicate that controls which skills are included.
+    /// </summary>
+    /// <remarks>
+    /// Skills for which the predicate returns <see langword="true"/> are kept;
+    /// others are excluded. Only one filter is supported; calling this method
+    /// again replaces any previously set filter.
+    /// </remarks>
+    /// <param name="predicate">A predicate that determines which skills to include.</param>
+    /// <returns>This builder instance for chaining.</returns>
+    public AgentSkillsProviderBuilder UseFilter(Func<AgentSkill, bool> predicate)
```
rogerbarreto, line 136: nit. Predicate doesn't say much to me, func sound more descriptive...

```suggestion
    public AgentSkillsProviderBuilder UseFilter(Func<AgentSkill, bool> funcFilter)
```

dotnet/src/Microsoft.Agents.AI/Skills/AgentSkillsProviderBuilder.cs
```diff
@@ -0,0 +1,197 @@
...
+
+    /// <summary>
+    /// Configures the <see cref="AgentSkillsProviderOptions"/> using the provided delegate.
+    /// </summary>
+    /// <param name="configure">A delegate to configure the options.</param>
+    /// <returns>This builder instance for chaining.</returns>
+    public AgentSkillsProviderBuilder UseOptions(Action<AgentSkillsProviderOptions> configure)
+    {
+        _ = Throw.IfNull(configure);
+        configure(this.EnsureOptions());
```
rogerbarreto: We might want to provide the options before transforming them and then ensure?

```suggestion
        configure(this.Options);
        this.EnsureOptions();
```

dotnet/src/Microsoft.Agents.AI/Skills/AgentSkillsProviderBuilder.cs
```diff
@@ -0,0 +1,197 @@
...
+
+    private AgentSkillsProviderOptions EnsureOptions()
+    {
+        if (this._options == null)
+        {
+            this._options = new AgentSkillsProviderOptions();
+        }
+
+        return this._options;
+    }
```
rogerbarreto: Ah, now that I'm seeing this method, we might just have this._options ?? new() ?

nit. Normally I see the name `EnsureXYZ` when we need to run some validation and expect the method to throw, something like `message.EnsureSuccessStatusCode();`


```suggestion
```

dotnet/src/Microsoft.Agents.AI/Skills/AggregatingAgentSkillsSource.cs
```diff
@@ -0,0 +1,46 @@
...
+    private readonly IEnumerable<AgentSkillsSource> _sources;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="AggregatingAgentSkillsSource"/> class.
+    /// </summary>
+    /// <param name="sources">The child sources to aggregate.</param>
+    public AggregatingAgentSkillsSource(IEnumerable<AgentSkillsSource> sources)
+    {
+        _ = Throw.IfNull(sources);
+        this._sources = sources;
```
rogerbarreto: nit.

```suggestion
        this._sources = Throw.IfNull(sources);
```

dotnet/src/Microsoft.Agents.AI/Skills/Decorators/DelegatingAgentSkillsSource.cs
SergeyMenshykh: I don't think we need this - each decorator already logs its own behavior. A LoggingAgentSkillsSource would only add generic entry/exit logging around GetSkillsAsync, which isn't worth the extra type at this point.

dotnet/src/Microsoft.Agents.AI/Skills/File/AgentFileSkillScript.cs
SergeyMenshykh: It's already done in the file source that discover all skills in skills folder(s) and creates instance of the skill script.

dotnet/src/Microsoft.Agents.AI/Skills/AgentSkillsProvider.cs
SergeyMenshykh: The template is ours, the replacement values are either hardcoded strings we control or XML-escaped skill metadata (SecurityElement.Escape), and custom templates are set by the developer, not end-users. There's no path for untrusted input to reach the placeholders.

dotnet/src/Microsoft.Agents.AI/Skills/AgentSkillsProvider.cs
SergeyMenshykh: The reason is that if `RunSkillScriptAsync` accepts `AIFunctionArguments`, it will contain all arguments that `AIFunction` is called with - such as skillName and scriptName and not the arguments required by the script itself.

dotnet/src/Microsoft.Agents.AI/Skills/AgentSkillsProviderBuilder.cs
SergeyMenshykh: Predicate is the standard .NET name for Func<T, bool> parameters - all LINQ methods (Where, First, Any, All, etc.) use it.

dotnet/src/Microsoft.Agents.AI/Skills/AgentSkillsProviderBuilder.cs
SergeyMenshykh: Yeah, the name can be misleading. I renamed it to the `GetOrCreateOptions` for more clarity.

dotnet/src/Microsoft.Agents.AI/Skills/AgentSkillsProviderBuilder.cs
SergeyMenshykh: Addressed by renaming the method.

dotnet/src/Microsoft.Agents.AI/Skills/AgentSkillFrontmatter.cs
SergeyMenshykh: Yep, it's part of the spec; added comment.

dotnet/src/Microsoft.Agents.AI/Skills/File/AgentFileSkillScript.cs
rogerbarreto: Should instead the Run be a no-op in this case?

SergeyMenshykh: There could be cases when file skills have scripts and when they don't. Cases where file skills have scripts but no runner is provided are considered misconfiguration issues that are communicated via the exception.

PR #4955, opened 2026-03-27
.NET: Fix RequestInfoEvent lost when resuming workflow from checkpoint

### LINKED ISSUE #2485: .NET: Dotnet - RequestInfo objects are not re-emitted when rehydrating a workflow from JSON.
Hi team,

We have a situation where a workflow was terminated before completion, with pending requests that had not been fulfilled.
Here's a representative snippet for how we resume the workflow from a custom JSON-based checkpointManager:
```csharp
var resumedRun = await InProcessExecution.ResumeStreamAsync(workflow, checkpointInfo, checkpointManager, runId).ConfigureAwait(false);
```
The workflow resumes without any exceptions, but when we listen to the event stream we do not see the pending requests re-emitted:
Here's a sample of how we listen for events on the resumed run:
```csharp
 await foreach (WorkflowEvent evt in run.WatchStreamAsync().WithCancellation(cancellationToken)) {
... 
}
```
Image
^ The expected behaviour would be for the workflow to re-emit the pending requests.

Now, the foreach enumeration never executes. I can see in the `runContext` internal object that there are pending requests that should have been re-emitted:

Image

Image

I am also seeing that when I query the status of the workflow after resuming, it is set to `RunStatus.NotStarted`.

Based on my analysis, it looks like the pending requests are sent to an [EventSink](https://github.com/microsoft/agent-framework/blob/0f2c5e6cb83cace1896dd5c80c015ffc48688c0b/dotnet/src/Microsoft.Agents.AI.Workflows/InProc/InProcessRunnerContext.cs#L175) and then consumed via an observer when the runLoop begins in the [StreamingRun](https://github.com/microsoft/agent-framework/blob/0f2c5e6cb83cace1896dd5c80c015ffc48688c0b/dotnet/src/Microsoft.Agents.AI.Workflows/Execution/StreamingRunEventStream.cs#L64) and these events might be being lost before we attach an observer (which happens post restore).

Could I get some help understanding what might be the problem here, and if it is indeed a bug in the framework?

peibekwe:
### Motivation and Context

When resuming a workflow from a JSON checkpoint, RestoreCheckpointAsync restored state and republished pending RequestInfoEvents immediately — before any event stream subscriber was attached. The events were sent to the EventSink, but no observer was listening yet, so they were silently dropped. This caused `WatchStreamAsync` to never yield the pending requests, and `GetStatusAsync` to remain `NotStarted`.

Fixes #2485 

### Description

This fix splits checkpoint restoration into a state-only restore step and a deferred republish step, so pending events are now republished only after the event stream subscribes, ensuring they are always delivered to the consumer. The fix covers both OffThread and Lockstep execution modes, runtime mid-flight restores, and subworkflow scenarios with qualified request ports.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.Workflows/Execution/LockstepRunEventStream.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Workflows/Execution/LockstepRunEventStream.cs
@@ -15,6 +15,7 @@ internal sealed class LockstepRunEventStream : IRunEventStream
 {
     private readonly CancellationTokenSource _stopCancellation = new();
     private readonly InputWaiter _inputWaiter = new();
+    private ConcurrentQueue<WorkflowEvent> _eventSink = new();
     private int _isDisposed;
 
     private readonly ISuperStepRunner _stepRunner;
@@ -34,6 +35,8 @@ internal sealed class LockstepRunEventStream : IRunEventStream
         // Save and restore Activity.Current so the long-lived session activity
         // doesn't leak into caller code via AsyncLocal.
         Activity? previousActivity = Activity.Current;
+
+        this._stepRunner.OutgoingEvents.EventRaised += this.OnWorkflowEventAsync;
 
         this._sessionActivity = this._stepRunner.TelemetryContext.StartWorkflowSessionActivity();
         this._sessionActivity?.SetTag(Tags.WorkflowId, this._stepRunner.StartExecutorId)
@@ -56,9 +59,9 @@ internal sealed class LockstepRunEventStream : IRunEventStream
 
         using CancellationTokenSource linkedSource = CancellationTokenSource.CreateLinkedTokenSource(this._stopCancellation.Token, cancellationToken);
 
-        ConcurrentQueue<WorkflowEvent> eventSink = [];
-
-        this._stepRunner.OutgoingEvents.EventRaised += OnWorkflowEventAsync;
+        // Re-emit any pending external requests that were restored from a checkpoint
+        // before this subscription was active. For non-resume starts this is a no-op.
+        await this._stepRunner.RepublishPendingEventsAsync(linkedSource.Token).ConfigureAwait(false);
 
         // Re-establish session as parent so the run activity nests correctly.
         Activity.Current = this._sessionActivity;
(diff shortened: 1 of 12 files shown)


dotnet/src/Microsoft.Agents.AI.Workflows/Execution/LockstepRunEventStream.cs
```diff
@@ -73,7 +76,26 @@ public async IAsyncEnumerable<WorkflowEvent> TakeEventStreamAsync(bool blockOnPe
             runActivity?.AddEvent(new ActivityEvent(EventNames.WorkflowStarted));
 
             // Emit WorkflowStartedEvent to the event stream for consumers
-            eventSink.Enqueue(new WorkflowStartedEvent());
+            this._eventSink.Enqueue(new WorkflowStartedEvent());
+
```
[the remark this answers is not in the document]
peibekwe: Moved republishing after emitting started event.

dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/RequestInfoExecutor.cs
```diff
@@ -124,22 +125,46 @@ public async ValueTask<ExternalRequest> HandleAsync(object message, IWorkflowCon
...
 
+        if (!message.Data.IsType(this.Port.Response, out object? data))
+        {
+            throw this.Port.CreateExceptionForType(message);
+        }
+
         if (this._allowWrapped && this._wrappedRequests.TryGetValue(message.RequestId, out ExternalRequest? originalRequest))
         {
+            this._wrappedRequests.Remove(message.RequestId);
             await context.SendMessageAsync(originalRequest.RewrapResponse(message), cancellationToken: cancellationToken).ConfigureAwait(false);
```
SergeyMenshykh: ﻿The request is removed before the SendMessageAsync call. If the send throws, the request info is permanently lost. I would suggest removing after sending.
```suggestion
            await context.SendMessageAsync(originalRequest.RewrapResponse(message), cancellationToken: cancellationToken).ConfigureAwait(false);
            this._wrappedRequests.Remove(message.RequestId);
```

SergeyMenshykh (approved):

westey-m (approved):

dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/RequestInfoExecutor.cs
peibekwe: Good catch. Updated.

alliscode (approved):

alliscode (approved):

lokitoth (approved):

PR #4957, opened 2026-03-28
Python: Fix AG-UI EventEncoder crash on non-BaseEvent objects

### LINKED ISSUE #4929: Python: [Bug]: AG-UI EventEncoder._encode_sse cannot handle AgentResponseUpdate
### Description

EventEncoder._encode_sse uses module_dump_json to return data. When the event type is AgentResponseUpdate, AgentResponseUpdate fails to serialize/export.
Error Message:
AttributeError: 'AgentResponseUpdate' object has no attribute 'model_dump_json'

### Code Sample

```markdown

```

```markdown

```

### Package Versions

agent-framework-ag-ui=1.0.0b260311

### Python Version

3.13

### Additional Context

_No response_

LEDazzio01:
## Summary

Fixes #4929 — `EventEncoder._encode_sse` crashes with `AttributeError` when a non-Pydantic object (like `AgentResponseUpdate`) reaches the SSE encoder.

## Root Cause

`EventEncoder.encode()` (from `ag_ui.encoder`) internally calls `model_dump_json()`, which only works on Pydantic `BaseModel` instances. `AgentResponseUpdate` extends `SerializationMixin` (not Pydantic) and uses `to_json()` instead. When an `AgentResponseUpdate` leaks through the AG-UI event pipeline, the encoder crashes:

```
AttributeError: 'AgentResponseUpdate' object has no attribute 'model_dump_json'
```

## Fix

Added a defensive `isinstance(event, BaseEvent)` guard in `_endpoint.py`'s `event_generator()` before passing events to `encoder.encode()`. Non-`BaseEvent` objects are now **skipped with a warning** instead of crashing the SSE stream.

### Changes
- **`_endpoint.py`**: Import `BaseEvent` from `ag_ui.core` and add type check before encoding
- **`test_endpoint.py`**: Added `test_endpoint_skips_non_base_event_objects` regression test that yields a mix of `BaseEvent` and `AgentResponseUpdate` objects and verifies the stream completes without error

## Testing

The new test creates a workflow that yields both a `RunStartedEvent` (valid) and an `AgentResponseUpdate` (invalid for SSE). It verifies:
- The stream returns 200
- `RUN_STARTED` is present in the output
- No `RUN_ERROR` is emitted (the invalid object is silently skipped)


LEDazzio01:
Addressed Copilot's feedback in [`e79c175`](https://github.com/LEDazzio01/agent-framework/commit/e79c175149394e9bafea21395175e86d1057638b):

- Updated the test docstring from _"are skipped gracefully"_ to _"are skipped with a warning"_
- Updated the inline comment from _"this should be skipped, not crash"_ to _"this should be skipped with a warning, not crash"_
- Updated the trailing assertion comment from _"silently skipped"_ to _"skipped with a warning"_

This now accurately reflects the endpoint behavior (warning is logged when non-`BaseEvent` objects are encountered).

LEDazzio01:

PR #4980, opened 2026-03-30
Python: Fix samples

TaoChenOSU:
### Motivation and Context

Fix some of the samples

### Contribution Checklist


.github/workflows/python-sample-validation.yml
```diff
@@ -67,11 +67,13 @@ jobs:
       # Azure AI configuration
       AZURE_AI_PROJECT_ENDPOINT: ${{ vars.AZURE_AI_PROJECT_ENDPOINT }}
       AZURE_AI_MODEL_DEPLOYMENT_NAME: ${{ vars.AZUREOPENAI__RESPONSESDEPLOYMENTNAME }}
```
chetantoshniwal, line 69: nit: are we changing this AZUREOPENAI__RESPONSESDEPLOYMENTNAME  too to AZURE_OPENAI__RESPONSES_DEPLOYMENT_NAME ?

python/samples/03-workflows/human-in-the-loop/agents_with_HITL.py
```diff
@@ -78,11 +78,7 @@ async def on_writer_response(
 
         # Writer agent response; request human feedback.
         # Preserve the full conversation so that the final editor has context.
-        conversation: list[Message]
-        if draft.full_conversation is not None:
-            conversation = list(draft.full_conversation)
-        else:
-            conversation = list(draft.agent_response.messages)
+        conversation = list(draft.full_conversation)
```
chetantoshniwal, line 81: why skip the null check ?

TaoChenOSU: It's not necessary because the value will never be null based on the type annotation.

.github/workflows/python-sample-validation.yml
TaoChenOSU: We need to consolidate it. We have both defined for our repo.

chetantoshniwal (approved):

giles17 (approved):

TaoChenOSU:
Waiting for @eavanvalkenburg's approval to make sure the 'FOUNDRY_MODEL` env var name changes make sense.

eavanvalkenburg (approved):

python/samples/03-workflows/README.md
```diff
@@ -172,7 +172,7 @@ Workflow and orchestration samples use `AzureOpenAIResponsesClient` rather than
 
 Workflow samples that use `AzureOpenAIResponsesClient` expect:
```
eavanvalkenburg, line 173: so this is probably out of sync, should be `FoundryChatClient` and then the below env vars are good.

PR #4996, opened 2026-03-31
.NET: Remove timeout from InputWait in OffThread execution

### LINKED ISSUE #2647: .NET: Perf: Remove Timeout-Wait in StreamingRunEventStream
Due to a deadlock without a current root-cause analysis, we added a timeout to the `WaitForInputAsync()` call inside of `StreamingRunEventStream` (also take the chance to rename it to `OffThreadRunEventStream`). This results in an occasional "wake-up" to NoOp and under adverse conditions/timings a slowdown in execution -- that would otherwise hang.

The fix would be to remove the need to have a timeout by unifying the input channel insertion with the `WaitForInputAsync`'s future completing.

lokitoth:
### Motivation and Context

When we first implemented OffThread running, we added a synchronization construct to avoid spinning a thread while the workflow is waiting on input after having run to Idle. Due to some race conditions we could not initially identify, we could end up with a deadlock, which we resolved by adding a timeout to the input wait. This could result in spurious "empty" steps, and it was long desired to remove it. 

Over time we hardened the core of the runtime and removed the likely data races (particularly in the input rework for supporting Polymorphic Message Routing). We have high confidence (three thousand iterations of the unit tests, run locally) that these, along with the epoch fix in this PR, bring us to a good place regarding removing the timeout.

### Description

- Removes the timeout on the input wait in StreamingRunEventStream (EventStream for OffThread execution) (closes #2647)
- Fixes an issue where the internal "halt" signal got ignored if the workflow ran to halt before the event stream began iterating due to improper expected epoch calculation 

### Contribution Checklist


dotnet/src/Microsoft.Agents.AI.Workflows/Execution/StreamingRunEventStream.cs
```diff
@@ -205,7 +206,9 @@ public async IAsyncEnumerable<WorkflowEvent> TakeEventStreamAsync(
         [EnumeratorCancellation] CancellationToken cancellationToken = default)
     {
         // Get the current epoch - we'll only respond to completion signals from this epoch or later
-        int myEpoch = Volatile.Read(ref this._completionEpoch) + 1;
+        int currentEpoch = Volatile.Read(ref this._completionEpoch);
+        bool expectingFreshWork = this._stepRunner.HasUnprocessedMessages || this._runStatus == RunStatus.Running;
```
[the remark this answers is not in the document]
lokitoth: Suggestion cannot be used because RunStatus is an Enum, and Volatile.Read does not support user-defined ValueTypes.

alliscode (approved):

dotnet/tests/Microsoft.Agents.AI.Workflows.UnitTests/InProcessExecutionTests.cs
```diff
@@ -132,6 +132,53 @@ public async Task RunAsyncAndStreamAsyncShouldProduceSimilarResultsAsync()
...
+        var inputMessage = new ChatMessage(ChatRole.User, "Hello");
+
+        // Act: Execute using streaming version with TurnToken
+        await using StreamingRun run = await InProcessExecution.RunStreamingAsync(workflow, new List<ChatMessage> { inputMessage });
+
+        // Send TurnToken to actually trigger execution (this is the key step)
+        bool messageSent = await run.TrySendMessageAsync(new TurnToken(emitEvents: true));
+        messageSent.Should().BeTrue("TurnToken should be accepted");
+
+        while (await run.GetStatusAsync() != RunStatus.Idle)
```
peibekwe, line 154: nit: Should this have a timeout or cancellationTokenSource with a timeout? Just to prevent a possible bug in the future from causing this to run for a long time.

peibekwe (approved):

dotnet/tests/Microsoft.Agents.AI.Workflows.UnitTests/InProcessExecutionTests.cs
lokitoth: The goal here is to work around needing to set a fixed timeout (that is not also the test timeout, representing a failure) so that slow test runs or other unexpected exogenous delays do not introduce a sporadic failure or failure to test; putting in a manual timeout cancellation here would reintroduce that issue. I'd rather not delay this for now, and fix it if it comes up.

Issue #4350, opened 2026-02-27
.NET: Support multiple sources for agent skills


SergeyMenshykh:
- [x] ADR - https://github.com/microsoft/agent-framework/pull/4787
- [x] Restructure existing agent skills API to support multiple sources: https://github.com/microsoft/agent-framework/pull/4871
  - Add file-based skill script execution.
- [x] Support code-defined skills: https://github.com/microsoft/agent-framework/pull/4951
- [x] Support class-based skills: https://github.com/microsoft/agent-framework/pull/5027

PR #5027, opened 2026-04-01
.NET: skill as class

SergeyMenshykh:
## Summary
Add class-based skills so skills can be distributed and reused as first-class C# types (for example from shared libraries/NuGet packages).

## Changes
- Introduce a class-based skill type for programmatic skill definitions.
- Allow registering and using these skills through the existing skills source/provider pipeline.
- Keep support for existing skill sources (file-based and inline) while enabling reusable class-based packaging.

Closes: https://github.com/microsoft/agent-framework/issues/4350

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI/CompatibilitySuppressions.xml
+++ b/dotnet/src/Microsoft.Agents.AI/CompatibilitySuppressions.xml
@@ -3,6 +3,20 @@
 <Suppressions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
   <Suppression>
     <DiagnosticId>CP0002</DiagnosticId>
+    <Target>M:Microsoft.Agents.AI.AgentSkillsProvider.#ctor(Microsoft.Agents.AI.AgentInlineSkill[])</Target>
+    <Left>lib/net10.0/Microsoft.Agents.AI.dll</Left>
+    <Right>lib/net10.0/Microsoft.Agents.AI.dll</Right>
+    <IsBaselineSuppression>true</IsBaselineSuppression>
+  </Suppression>
+  <Suppression>
+    <DiagnosticId>CP0002</DiagnosticId>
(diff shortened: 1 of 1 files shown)


dotnet/samples/02-agents/AgentSkills/Agent_Step05_CodeDefinedSkillsWithDI/Program.cs
```diff
@@ -0,0 +1,116 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+// This sample demonstrates how to use Dependency Injection (DI) with Agent Skills.
```
peibekwe: nit: Could some of these samples be merged with others (e.g. DI and non-DI related samples merged into one project) to reduce the number of sample projects we have?

SergeyMenshykh: Yeah, I want to keep them separate so the non-DI samples aren't cluttered with DI stuff and remain clear. I've consolidated the two DI samples into one, so we have one fewer sample.

dotnet/samples/02-agents/AgentSkills/Agent_Step03_ClassBasedSkills/Program.cs
```diff
@@ -0,0 +1,102 @@
...
+    public override IReadOnlyList<AgentSkillScript>? Scripts => this._scripts ??=
+    [
+        CreateScript("convert", ConvertUnits),
+    ];
+
+    private static string ConvertUnits(double value, double factor)
+    {
+        double result = Math.Round(value * factor, 4);
+        return JsonSerializer.Serialize(new { value, factor, result });
+    }
```
westey-m, line 101: Have you considered just having methods on the class for scripts, which can be annotated with attributes for the script name.  And similar for resources, they could be properties with an attribute for the resource name (if different from the property name).

These would need reflection of course, but if someone was using native aot, they could always use the inline script instead.

SergeyMenshykh: I plan to explore the possibility later.

westey-m (approved):

PR #5038, opened 2026-04-01
Python: Fix broken samples and add missing READMEs

giles17:
## Motivation and Context
Fix runtime errors in several Python samples and add missing README documentation for sample folders.

### Sample fixes
- **simple_context_provider.py**: `instructions` was passed as a top-level kwarg to `get_response()`, but it must be inside the `options` dict. This caused a silent failure (swallowed by `suppress(Exception)`) where user info extraction never worked.
- **suspend_resume_session.py**: The in-memory session demo used `FoundryChatClient()` with no args, which requires a credential. Changed to `OpenAIChatCompletionClient()` to match the docstring intent of comparing service-managed vs in-memory sessions.
- **foundry_chat_client_with_hosted_mcp.py**: `store=True` was passed as a top-level kwarg to `Agent.run()`, but it must be inside `options={"store": True}`. This caused a `TypeError` crash.
- **mem0_basic.py**: Preferences query was never sent to the agent before the Mem0 indexing delay, so Mem0 never learned user preferences. Also added missing print for new session result.
- **mem0_sessions.py**: Agent-scoped example had no session for multi-turn conversation and combined `user_id` with `agent_id` which Mem0 API does not support (memories stored without user_id when agent_id is provided).
- **redis_basics.py**: Standalone provider demo printed raw `SessionContext` object instead of retrieved context messages. Fixed docstring referencing wrong client type.
- **redis_sessions.py**: Missing `load_dotenv()` call caused `KeyError` on `FOUNDRY_PROJECT_ENDPOINT`. Fixed docstring referencing wrong client type.
- **azure_redis_conversation.py**: Duplicate copyright comment replaced with missing `load_dotenv()` call.

### New READMEs
- **context_providers/README.md**: Top-level README covering the two standalone samples and linking to subfolders (azure_ai_search, mem0, redis) that already have their own READMEs.
- **conversations/README.md**: README covering suspend/resume sessions, custom history provider, and Redis history provider samples with prerequisites.

## Checklist
- [x] I've read the [contributing guidelines](../CONTRIBUTING.md)
- [x] I've verified changes locally

### DIFF
--- a/python/samples/02-agents/declarative/README.md
+++ b/python/samples/02-agents/declarative/README.md
@@ -68,7 +68,7 @@ Illustrates a basic agent using Azure OpenAI with structured responses.
 
 **Key concepts**: Azure OpenAI integration, credential management, structured outputs
 
-### 5. **OpenAI Responses Agent** ([`openai_responses_agent.py`](./openai_responses_agent.py))
+### 5. **OpenAI Responses Agent** ([`openai_agent.py`](./openai_agent.py))
 
 Demonstrates the simplest possible agent using OpenAI directly.
 


python/samples/02-agents/conversations/suspend_resume_session.py
```diff
@@ -4,6 +4,7 @@
 
 from agent_framework import Agent, AgentSession
 from agent_framework.foundry import FoundryChatClient
```
eavanvalkenburg, line 6: Is this still used?

TaoChenOSU (approved):

python/samples/02-agents/conversations/suspend_resume_session.py
giles17: Yes

PR #5052, opened 2026-04-02
Python: Support MCP tool binding by reference in declarative YAML

### LINKED ISSUE #4927: Python: [Bug]: Support MCP tool binding by reference in declarative YAML
### Description

MAF declarative workflows require embedding full MCP connection details directly in the YAML tool definition. Two problems make this unworkable:

1. **Custom HTTP headers are not supported in the YAML schema** - only `url` is available. (Tracked in #4582)
2. **There is no way to reference an MCP connection by identifier** - connection details must be redeclared in every agent and workflow that uses the same MCP server, coupling sensitive infrastructure config to business workflow files.

The expected model is that MCP connections are declared once in configuration or code, and workflow YAML references them by name, declaring only which tools each agent is allowed to use.

## Current Behaviour

Passing a pre-built MCP tool via `bindings` is ignored. MAF constructs the tool directly from the YAML spec regardless. Results in `server_url: ''` being sent to the API, returning a 500.

> Note: `McpTool` does accept a `bindings` field (inherited from `Tool`) but `_parse_tool` in `AgentFactory` never reads it for the MCP case, unlike `FunctionTool` where bindings correctly resolve the callable.

The only working workaround is passing the tool at `agent.run()` with no `tools` block in the YAML, **but this loses per-agent tool scoping entirely.**

### Code Sample

```markdown
# Minimal agent YAML — no url, no headers, no connection details.
# Just the name and the allowed tools.
AGENT_YAML = """
kind: Prompt
name: TestAgent
instructions: You are a helpful assistant.
model:
    id: gpt-4.1
    provider: AzureOpenAI.Responses
tools:
    - kind: mcp
      name: product-mcp
      allowedTools:
        - customer_search
      # connection resolved from bindings by name at runtime
"""

async def main():
    mcp_tool = OpenAIResponsesClient.get_mcp_tool(
        name="product-mcp",
        description="Product MCP server",
        url="http://example.com",
        headers={"X-API-Key": "..."},
        approval_mode="never_require",
        allowed_tools=["customer_search"],
    )

    factory = AgentFactory(
        safe_mode=False,
        bindings={"product-mcp": mcp_tool},
    )

    async with factory.create_agent_from_yaml(AGENT_YAML) as agent:
        response = await agent.run("Search for customer Inforshow")
        print(f"Response: {response.text}")
```

```markdown

```

### Package Versions

agent-framework: 1.0.0rc4, agent-framework-declarative: 1.0.0b260311

### Python Version

3.13

### Additional Context

From our understanding, Azure Foundry solves this cleanly via `project_connection_id`: MCP connections are registered centrally in the project (including headers and secrets) and referenced by name in the YAML:

```yaml
tools:
  - type: mcp
    server_label: ERP_test_MCP
    server_url: http://example.com/
    require_approval:
      never:
        tool_names:
          - customer_search
    project_connection_id: ERP-test-MCP
```

Could a similar behavior be achievable through MAF, with centralized code definitions?

hashwnath:
## Summary

Fixes #4927

`AgentFactory._parse_tool()` ignores `self.bindings` for the `McpTool` case, unlike `FunctionTool` where bindings correctly resolve the callable. This causes pre-built MCP tool dicts passed via `AgentFactory(bindings={"product-mcp": mcp_tool})` to be silently ignored — the factory always constructs the tool from YAML fields, resulting in `server_url: ''` and a 500 error when no `url` is specified in the YAML.

This PR adds binding resolution for `McpTool` in `_parse_tool()`, mirroring the existing `FunctionTool` pattern:

1. **Explicit bindings** — checks `tool_resource.bindings` list against `self.bindings` dict
2. **Name-based fallback** — matches `tool_resource.name` directly against `self.bindings` keys

If a binding resolves, the pre-built tool dict is returned directly. If no binding matches, the existing dict-building logic runs unchanged.

## Changes

- **`_loader.py`**: Added binding-by-reference resolution in the `McpTool` case of `_parse_tool()`
- **`test_declarative_loader.py`**: Added 4 tests:
  - Explicit bindings list resolves MCP tool
  - Name-based fallback resolves MCP tool
  - Regression test matching the exact scenario from #4927 (name binding with `allowedTools`, no url)
  - No-match falls through to existing dict-building behavior

## Test plan

- [x] All 723 existing declarative tests pass
- [x] New tests verify both binding resolution paths and the fallback
- [x] Ruff format and lint clean
- [x] Regression test verifies the exact issue scenario (no `server_url: ''`)

### DIFF
--- a/python/packages/declarative/agent_framework_declarative/_loader.py
+++ b/python/packages/declarative/agent_framework_declarative/_loader.py
@@ -795,7 +795,7 @@ class AgentFactory:
                         bound_tool = self.bindings.get(tool_resource.name)
 
                 if bound_tool is not None:
-                    return bound_tool
+                    return cast("AFFunctionTool | dict[str, Any]", bound_tool)
 
                 result = {
                     "type": "mcp",
--- a/python/packages/declarative/agent_framework_declarative/_loader.py
+++ b/python/packages/declarative/agent_framework_declarative/_loader.py
@@ -782,6 +782,21 @@ class AgentFactory:
                     result["description"] = tool_resource.description
                 return result
             case McpTool():
+                # Check bindings first - allow MCP tools to be resolved by reference
+                bound_tool: Any = None
+                if self.bindings:
+                    # Check explicit bindings list first
+                    if tool_resource.bindings:
+                        for binding in tool_resource.bindings:
+                            if binding.name and (bound_tool := self.bindings.get(binding.name)):
+                                break
+                    # Fall back to matching by tool name
+                    if bound_tool is None and tool_resource.name:
+                        bound_tool = self.bindings.get(tool_resource.name)
+
+                if bound_tool is not None:
+                    return bound_tool
+
                 result = {
                     "type": "mcp",
                     "server_label": tool_resource.name.replace(" ", "_") if tool_resource.name else "",
--- a/python/packages/declarative/tests/test_declarative_loader.py
+++ b/python/packages/declarative/tests/test_declarative_loader.py
@@ -1469,6 +1469,71 @@ tools:
         assert tools[0]["server_label"] == "other-mcp"
         assert tools[0]["server_url"] == "https://api.example.com/mcp"
 
+    def test_parse_mcp_tool_explicit_binding_takes_priority_over_name(self):
+        """Test that explicit bindings list takes priority over name-based lookup."""
+        from unittest.mock import MagicMock
+
+        from agent_framework_declarative import AgentFactory
+
+        yaml_content = """
+kind: Prompt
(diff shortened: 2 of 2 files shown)


moonbox3:
@hashwnath please have a look at the failing CI/CD items. Thanks.

hashwnath:
Fixed the mypy issue! The other CI failures are from the orchestration tests broken on main, unrelated to the declarative changes here.

moonbox3:
@peibekwe can you have a look, please?

eavanvalkenburg:
Closing this as we will spend time on the declarative package soon, the issue will remain open for that.

PR #5063, opened 2026-04-02
.NET: feat(AgentSkills): add IncludeSkillInstructions option to AgentSkills…

thesamuel23:
…Provider

- Add `IncludeSkillInstructions` property to `AgentSkillsProviderOptions` (default true).
- Update `AgentSkillsProvider` to omit skill instructions from the system prompt when `IncludeSkillInstructions` is false.
- Add `IncludeSkillInstructions` fluent method to `AgentSkillsProviderBuilder`.
- Skip prompt template validation in `AgentSkillsProvider` constructor when instructions are disabled.
- Add unit tests for the new option in both provider and builder.

### Motivation and Context

There are scenarios where an agent might already have prior knowledge of the skills, or where they are being advertised through other means. In these cases, including the default skill-related system prompt is redundant or undesirable.

Why is this change required? To allow users to suppress the automatic injection of skill-related instructions into the system prompt.
What problem does it solve? It provides a way to omit the "Advertise" step of the Agent Skills specification while still keeping the underlying skill tools (load_skill, read_skill_resource, etc.) available in the AIContext.
What scenario does it contribute to? Advanced agent configurations where the system prompt needs to be carefully managed or where skill discovery is handled outside of the standard system prompt.
Description
This PR introduces the IncludeSkillInstructions option to the AgentSkillsProvider to control the advertisement of skills in the system prompt.

AgentSkillsProviderOptions: Added IncludeSkillInstructions (bool, default true).
AgentSkillsProvider:
Updated BuildSkillsInstructions to respect the flag and skip instruction generation when false.
Updated the constructor to skip the SkillsInstructionPrompt template validation if instructions are disabled, as they will not be used.
AgentSkillsProviderBuilder: Added a fluent IncludeSkillInstructions(bool enabled = true) method.
Unit Tests: Added and updated tests in AgentSkillsProviderTests.cs and AgentSkillsProviderBuilderTests.cs to verify that instructions are correctly suppressed while tools remain available, and that redundant validation is skipped.

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.Workflows/Execution/StreamingRunEventStream.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Workflows/Execution/StreamingRunEventStream.cs
@@ -23,7 +23,8 @@ internal sealed class StreamingRunEventStream : IRunEventStream
     private readonly CancellationTokenSource _runLoopCancellation;
     private readonly bool _disableRunLoop;
     private Task? _runLoopTask;
-    private RunStatus _runStatus = RunStatus.NotStarted;
+    private volatile RunStatus _runStatus = RunStatus.NotStarted;
+
     private int _completionEpoch; // Tracks which completion signal belongs to which consumer iteration
 
     public StreamingRunEventStream(ISuperStepRunner stepRunner, bool disableRunLoop = false)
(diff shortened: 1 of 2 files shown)


thesamuel23:
@microsoft-github-policy-service agree company="AgentMatters.ai"

PR #5078, opened 2026-04-03
.NET: Align skill folder discovery with spec

SergeyMenshykh:
This PR adds support for discovering skill scripts and resources from folders defined in the agent skill specification, with the ability to specify other folders if needed via options.

### DIFF
--- a/dotnet/tests/Microsoft.Agents.AI.UnitTests/AgentSkills/AgentFileSkillsSourceScriptTests.cs
+++ b/dotnet/tests/Microsoft.Agents.AI.UnitTests/AgentSkills/AgentFileSkillsSourceScriptTests.cs
@@ -249,23 +249,32 @@ public sealed class AgentFileSkillsSourceScriptTests : IDisposable
     [Theory]
     [InlineData("./scripts")]
     [InlineData("./scripts/f1")]
-    public async Task GetSkillsAsync_ScriptFolderWithDotSlashPrefix_DiscoversScriptsAsync(string folder)
+    [InlineData("./scripts/f1", "./f2")]
+    public async Task GetSkillsAsync_ScriptFolderWithDotSlashPrefix_DiscoversScriptsAsync(params string[] folders)
     {
-        // Arrange — "./scripts" and "./scripts/f1" are equivalent to "scripts" and "scripts/f1";
+        // Arrange — "./"-prefixed folders are equivalent to their counterparts without the prefix;
         // the leading "./" is transparently normalized by Path.GetFullPath during file enumeration.
-        string folderWithoutDotSlash = folder.Substring(2); // strip "./"
         string skillDir = CreateSkillDir(this._testRoot, "dotslash-script-skill", "Dot-slash prefix", "Body.");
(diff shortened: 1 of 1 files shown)


lokitoth (approved):

dotnet/src/Microsoft.Agents.AI/Skills/File/AgentFileSkillsSourceOptions.cs
```diff
@@ -30,4 +30,30 @@ public sealed class AgentFileSkillsSourceOptions
...
+    /// Values may be single-segment names (e.g., <c>"scripts"</c>) or multi-segment relative
+    /// paths (e.g., <c>"sub/scripts"</c>). Use <c>"."</c> to include files directly at the
+    /// skill root. Leading <c>"./"</c> prefixes, trailing separators, and backslashes are
+    /// normalized automatically; paths containing <c>".."</c> segments or absolute paths are
+    /// rejected.
+    /// When <see langword="null"/>, defaults to <c>scripts</c> (per the
+    /// <see href="https://agentskills.io/specification">Agent Skills specification</see>).
+    /// When set, replaces the defaults entirely.
+    /// </summary>
+    public IEnumerable<string>? ScriptFolders { get; set; }
```
rogerbarreto, line 45: nit.
Since absolute is prohibited, should we state `Relative` prefix?

```suggestion
    public IEnumerable<string>? RelativeScriptFolders { get; set; }
```

dotnet/tests/Microsoft.Agents.AI.UnitTests/AgentSkills/AgentFileSkillsSourceScriptTests.cs
```diff
@@ -230,6 +228,46 @@ public async Task GetSkillsAsync_ExecutorReceivesArgumentsAsync()
...
+        // Assert — script file inside the deeply nested folder is discovered
+        Assert.Single(skills);
+        Assert.Single(skills[0].Scripts!);
+        Assert.Equal("f1/f2/f3/run.py", skills[0].Scripts![0].Name);
+    }
+
+    [Theory]
+    [InlineData("./scripts")]
+    [InlineData("./scripts/f1")]
+    public async Task GetSkillsAsync_ScriptFolderWithDotSlashPrefix_DiscoversScriptsAsync(string folder)
```
rogerbarreto: Suggest adding a test for multiple.

```suggestion
    [InlineData("./scripts")]
    [InlineData("./scripts/f1")]
    [InlineData("./scripts/f1","./f2")]
    public async Task GetSkillsAsync_ScriptFolderWithDotSlashPrefix_DiscoversScriptsAsync(string[] params folders)
```

rogerbarreto (approved):

dotnet/src/Microsoft.Agents.AI/Skills/File/AgentFileSkillsSourceOptions.cs
SergeyMenshykh: The scope here is the skill directory, so these folders are inherently relative to the skill root - the Relative prefix would be redundant and add verbosity without extra clarity.

PR #5081, opened 2026-04-03
.NET: Add Message Delivery Callback Overloads to Executor

### LINKED ISSUE #2156: .NET: Consider a method for knowing when all edges going to your executor have finished
When using FanIn or similar behavior with manual edges, the target executor that runs once all sources run doesn't have a way to figure out how many incoming edges there are. This means you need to hardcode (or determine at executor creation time) how many sources there are and use that in your executor to handle multiple sources.

For example, the samples hardcode 2 for the number of incoming sources which feels wrong.
https://github.com/microsoft/agent-framework/blob/5537b1da79812be981d1ac8d1c3187957a448e03/dotnet/samples/GettingStarted/Workflows/Concurrent/Concurrent/Program.cs#L116

I made a hacky workaround where I use a collector executor to run before the summary executor and it just adds to a list and returns that list.
```csharp
var workflow = new WorkflowBuilder(userIntent)
    .AddEdge<UserIntents>(userIntent, agent1, condition: u => u.A)
    .AddEdge<UserIntents>(userIntent, agent2, condition: u => u.B)
    .AddEdge(agent1, collector)
    .AddEdge(agent2, collector)
    .AddEdge(collector, summary)
    .WithOutputFrom(summary)
    .Build();

internal class Summary(AIAgent summaryAgent) : ReflectingExecutor<Summary>(summaryAgent.Id), IMessageHandler<List<ChatMessage>>
{
    private bool _ran;

    public async ValueTask HandleAsync(List<ChatMessage> messages, IWorkflowContext context, CancellationToken cancellationToken = default)
    {
        if (!_ran)
        {
            Console.WriteLine("Summary Ran");
            _ran = true;
            var resp = await summaryAgent.RunAsync(messages, cancellationToken: cancellationToken);
            await context.YieldOutputAsync(resp.Messages, cancellationToken);
        }
    }
}

internal class Collector() : ReflectingExecutor<Collector>("Collect"), IMessageHandler<List<ChatMessage>, List<ChatMessage>>
{
    private List<ChatMessage> _messages = [];

    public async ValueTask<List<ChatMessage>> HandleAsync(List<ChatMessage> message, IWorkflowContext context, CancellationToken cancellationToken = default)
    {
        _messages.AddRange(message);
        return _messages;
    }
}
```

lokitoth:
### Motivation and Context

Often it is desirable to create an Executor that runs aggregation logic on incoming messages, but that logic is not incremental, so it is important to know when the full set of messages has been collected.

### Description

Add `OnMessageDeliveryStarting` and `OnMessageDeliveryFinished` overloads on `Executor`, enabling users to determine when the super step's message deliveries for the Executor are starting and when deliveries have completed.

Closes #2156

### Contribution Checklist

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.Workflows/InProc/InProcessExecutionEnvironment.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Workflows/InProc/InProcessExecutionEnvironment.cs
@@ -185,6 +185,6 @@ public sealed class InProcessExecutionEnvironment : IWorkflowExecutionEnvironmen
         Run run = new(runHandle);
         await run.RunToNextHaltAsync(cancellationToken).ConfigureAwait(false);
 
-        return run;
+        return new(runHandle);
     }
 }
--- a/dotnet/src/Microsoft.Agents.AI.Workflows/InProc/InProcessExecutionEnvironment.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Workflows/InProc/InProcessExecutionEnvironment.cs
@@ -182,6 +182,9 @@ public sealed class InProcessExecutionEnvironment : IWorkflowExecutionEnvironmen
         AsyncRunHandle runHandle = await this.ResumeRunAsync(workflow, fromCheckpoint, [], cancellationToken)
                                              .ConfigureAwait(false);
 
-        return new(runHandle);
+        Run run = new(runHandle);
+        await run.RunToNextHaltAsync(cancellationToken).ConfigureAwait(false);
+
+        return run;
     }
 }
--- a/dotnet/src/Microsoft.Agents.AI.Workflows/InProc/InProcessExecutionEnvironment.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Workflows/InProc/InProcessExecutionEnvironment.cs
@@ -185,6 +185,6 @@ public sealed class InProcessExecutionEnvironment : IWorkflowExecutionEnvironmen
         Run run = new(runHandle);
         await run.RunToNextHaltAsync(cancellationToken).ConfigureAwait(false);
 
-        return new(runHandle);
+        return run;
     }
 }
--- a/dotnet/src/Microsoft.Agents.AI.Workflows/InProc/InProcessRunner.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Workflows/InProc/InProcessRunner.cs
@@ -250,8 +250,9 @@ internal sealed class InProcessRunner : ISuperStepRunner, ICheckpointingHandle
 
         this.StepTracer.TraceActivated(receiverId);
 
-        // TODO: traceContext?
-        var tracelessContext = this.RunContext.BindWorkflowContext(receiverId);
+        // TODO: #5084 - Add delivery-level activity (max one per step per executor) to capture non-message
+        // specific invocations of executor logic.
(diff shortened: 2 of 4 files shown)


dotnet/src/Microsoft.Agents.AI.Workflows/InProc/InProcessExecutionEnvironment.cs
```diff
@@ -182,6 +182,9 @@ public async ValueTask<Run> ResumeAsync(
         AsyncRunHandle runHandle = await this.ResumeRunAsync(workflow, fromCheckpoint, [], cancellationToken)
                                              .ConfigureAwait(false);
 
+        Run run = new(runHandle);
+        await run.RunToNextHaltAsync(cancellationToken).ConfigureAwait(false);
+
         return new(runHandle);
```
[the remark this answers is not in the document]
lokitoth: In practice this doesn't matter, because `Run` is simply an API wrapper around the `runHandle`, but still a good change to make

SergeyMenshykh (approved):

dotnet/src/Microsoft.Agents.AI.Workflows/InProc/InProcessRunner.cs
```diff
@@ -249,17 +249,32 @@ private async ValueTask DeliverMessagesAsync(string receiverId, ConcurrentQueue<
         Executor executor = await this.RunContext.EnsureExecutorAsync(receiverId, this.StepTracer, cancellationToken).ConfigureAwait(false);
 
         this.StepTracer.TraceActivated(receiverId);
-        while (envelopes.TryDequeue(out var envelope))
+
+        // TODO: traceContext?
```
rogerbarreto: Suggest removing the TODO and adding as a Tracking Issue for this.

```suggestion
```

lokitoth: Good catch

dotnet/src/Microsoft.Agents.AI.Workflows/InProc/InProcessExecutionEnvironment.cs
rogerbarreto: Is this intended, ignore the `run` and give back another instance?

```suggestion
        Run run = new(runHandle);
        await run.RunToNextHaltAsync(cancellationToken).ConfigureAwait(false);

        run;
```

OR

```suggestion
        await new Run (runHandle).RunToNextHaltAsync(cancellationToken).ConfigureAwait(false);

        return new(runHandle);
```

dotnet/src/Microsoft.Agents.AI.Workflows/InProc/InProcessRunner.cs
```diff
@@ -249,17 +249,32 @@ private async ValueTask DeliverMessagesAsync(string receiverId, ConcurrentQueue<
...
+        var tracelessContext = this.RunContext.BindWorkflowContext(receiverId);
```
peibekwe: nit: I'd recommend using the type instead of `var` to align with the rest of the changes and make it easier to understand outside an IDE.

peibekwe (approved):

dotnet/src/Microsoft.Agents.AI.Workflows/InProc/InProcessRunner.cs
lokitoth: Good catch

peibekwe (approved):

PR #5096, opened 2026-04-05
Python: Add safe manual workflow canary test

areksaxyz:
Adds a safe canary test for workflow investigation.


argareksapati:
Hi maintainers,

Could someone with permission please run the "Integration Tests (Manual)" workflow for this PR with:
pr-number: 5096

This PR only adds a safe canary test under `python/` for workflow validation. It does not read secrets, dump environment variables, access cloud resources, or make external network calls. It only prints an `AF_CANARY` marker and intentionally fails in reusable workflow runs so the execution path is visible in logs.

 A run link is enough. Thank you.

  @eavanvalkenburg @kshyju

eavanvalkenburg:
I don't know what this does or why we need it, but tests that do `git` stuff are not what we want to see

PR #5107, opened 2026-04-05
Python: Migrate GitHub Copilot package to SDK 0.2.x

### LINKED ISSUE #5106: Python: [Bug]: GitHub Copilot package incompatible with github-copilot-sdk 0.2.x
## Description

The `agent-framework-github-copilot` package fails at import time when used with `github-copilot-sdk>=0.2.0` because it imports from `copilot.types`, a module that does not exist in SDK 0.2.x.

**What happened?**

Running any sample or importing `agent_framework_github_copilot` raises `ModuleNotFoundError: No module named 'copilot.types'`.

**What was expected?**

The package should work correctly with `github-copilot-sdk>=0.2.0`.

**Steps to reproduce:**

1. Install `github-copilot-sdk>=0.2.0`
2. Import or run any GitHub Copilot agent sample
3. Observe `ModuleNotFoundError: No module named 'copilot.types'`

## Code Sample

```python
from agent_framework_github_copilot import GitHubCopilotAgent
# ModuleNotFoundError: No module named 'copilot.types'
```

```
  File "github_copilot_basic.py", line 5, in <module>
    from agent_framework_github_copilot import GitHubCopilotAgent
  File ".../agent_framework_github_copilot/_agent.py", line 12, in <module>
    from copilot.types import PermissionRequest, PermissionRequestResult, MCPServerConfig, Tool, ToolInvocation, ToolResult
ModuleNotFoundError: No module named 'copilot.types'
```

## Package Versions

agent-framework-github-copilot: 1.0.0b260402, github-copilot-sdk: 0.2.x

## Python Version

Python 3.11

## Additional Context

In SDK 0.2.x, several additional breaking API changes were made beyond the module restructure:

- `CopilotClientOptions` dict removed; `CopilotClient` now takes `SubprocessConfig`
- `SessionConfig` / `ResumeSessionConfig` TypedDicts removed; `create_session` / `resume_session` now take keyword arguments directly
- `send` / `send_and_wait` now take a plain `str` prompt instead of a `MessageOptions` dict
- `on_permission_request` is now a required parameter for `create_session` / `resume_session`
- `PermissionRequest` changed from `TypedDict` to dataclass; `.get("kind")` must be replaced with `.kind`

Affected files: `_agent.py`, `tests/test_github_copilot_agent.py`, all 7 samples under `python/samples/02-agents/providers/github_copilot/`, and `python/scripts/sample_validation/create_dynamic_workflow_executor.py`.

Fix tracked in PR #5107.

droideronline:
Fixes #5106

## Summary

Scoped to SDK 0.2.x compatibility fixes only. OTel/telemetry integration will follow in a separate PR.

- Replace all imports from the non-existent `copilot.types` module with correct SDK 0.2.x module paths across `_agent.py`, the test file, all 7 samples, and the sample validation script
- Fix `PermissionRequest.get("kind")` dict-style access to `.kind` attribute access (`PermissionRequest` is a dataclass in 0.2.x, not a `TypedDict`)
- Update `start()` to use `SubprocessConfig` instead of the removed `CopilotClientOptions` dict; only pass `log_level` when explicitly set to preserve SDK default behavior
- Update `_create_session` and `_resume_session` to call the SDK with direct keyword arguments instead of the removed `SessionConfig` / `ResumeSessionConfig` TypedDicts; add `_deny_all_permissions` fallback since `on_permission_request` is now required
- Update `pyproject.toml` dependency constraint to `github-copilot-sdk>=0.2.0,<0.3.0`
- Update tests to match new SDK API shapes

## Root Cause

`copilot.types` does not exist in SDK 0.2.x. Types were reorganised:

| Type | 0.2.x module |
|---|---|
| `PermissionRequest`, `SessionEvent`, `SessionEventType` | `copilot.generated.session_events` |
| `PermissionRequestResult`, `MCPServerConfig`, `SystemMessageConfig` | `copilot.session` |
| `Tool`, `ToolInvocation`, `ToolResult` | `copilot.tools` |

Additionally the following APIs changed:
- `CopilotClientOptions` dict removed; `CopilotClient` now takes `SubprocessConfig`
- `SessionConfig` / `ResumeSessionConfig` TypedDicts removed; `create_session` / `resume_session` now take keyword arguments
- `send` / `send_and_wait` now take a plain `str` prompt instead of a `MessageOptions` dict
- `on_permission_request` is now a required parameter for `create_session` / `resume_session`

## Test Plan

- [x] `uv run poe test` passes (77 tests, 0 failures)
- [x] `uv run poe lint` passes (ruff, pyright, mypy)
- [x] All 7 samples import without `ModuleNotFoundError`


python/packages/github_copilot/agent_framework_github_copilot/_agent.py
```diff
@@ -64,6 +57,14 @@
 logger = logging.getLogger("agent_framework.github_copilot")
 
 
+def _deny_all_permissions(
+    _request: PermissionRequest,
+    _invocation: dict[str, str],
+) -> PermissionRequestResult:
+    """Default permission handler that denies all requests."""
+    return PermissionRequestResult()
```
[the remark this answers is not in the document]
droideronline: `PermissionRequestResult()` is valid here — the dataclass has a default value for `kind`:

```python
kind: PermissionRequestResultKind = "denied-no-approval-rule-and-could-not-request-from-user"
```

So the empty constructor produces a denial result, which is the intended behavior for a "deny all" handler.

Also, the suggested `kind="deny"` would be a type error — valid `PermissionRequestResultKind` values are: `'approved'`, `'denied-by-rules'`, `'denied-by-content-exclusion-policy'`, `'denied-no-approval-rule-and-could-not-request-from-user'`, `'denied-interactively-by-user'`, `'no-result'`.

python/packages/github_copilot/agent_framework_github_copilot/_agent.py
```diff
@@ -27,22 +27,15 @@
...
-        ResumeSessionConfig,
-        SessionConfig,
-        SystemMessageConfig,
-        ToolInvocation,
-        ToolResult,
-    )
-    from copilot.types import Tool as CopilotTool
+    from copilot.session import MCPServerConfig, PermissionRequestResult, SystemMessageConfig
+    from copilot.tools import Tool as CopilotTool
+    from copilot.tools import ToolInvocation, ToolResult
```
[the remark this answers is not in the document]
droideronline: Good catch! Fixed in the latest commit — `python/scripts/sample_validation/create_dynamic_workflow_executor.py` now imports from `copilot.session` instead of `copilot.types`.

python/packages/github_copilot/agent_framework_github_copilot/_agent.py
```diff
@@ -248,10 +272,26 @@ def __init__(
...
+        Maps ``system_message.content`` to ``instructions`` so that
+        :class:`AgentTelemetryLayer` can capture the system prompt in traces.
+        """
+        opts = dict(self._default_options)
+        system_message = opts.get("system_message")
+        if isinstance(system_message, dict):
+            content = cast("dict[str, Any]", system_message).get("content")
+            if content:
+                opts["instructions"] = content
+        return opts
```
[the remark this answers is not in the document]
droideronline: Good catch — fixed in the latest commit. The `default_options` property now merges `model` back from `self._settings` so that `AgentTelemetryLayer` correctly populates `gen_ai.request.model` in spans.

droideronline:
@dmytrostruk @eavanvalkenburg @moonbox3 @TaoChenOSU - requesting your review on this PR. Thanks!

moonbox3:
<a href="https://github.com/microsoft/agent-framework/blob/d3ad3c3361cbc55d36ec329e1ab7b2799e09e7b7/README.md">Python Test Coverage</a><details><summary>Python Test Coverage Report • </summary><table><tr><th>File</th><th>Stmts</th><th>Miss</th><th>Cover</th><th>Missing</th></tr><tbody><tr><td colspan="5"><b>packages/github_copilot/agent_framework_github_copilot</b></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/d3ad3c3361cbc55d36ec329e1ab7b2799e09e7b7/packages/github_copilot/agent_framework_github_copilot/_agent.py">_agent.py</a></td><td>263</td><td>12</td><td>95%</td><td><a href="https://github.com/microsoft/agent-framework/blob/d3ad3c3361cbc55d36ec329e1ab7b2799e09e7b7/packages/github_copilot/agent_framework_github_copilot/_agent.py#L37-L38">37&ndash;38</a>, <a href="https://github.com/microsoft/agent-framework/blob/d3ad3c3361cbc55d36ec329e1ab7b2799e09e7b7/packages/github_copilot/agent_framework_github_copilot/_agent.py#L44">44</a>, <a href="https://github.com/microsoft/agent-framework/blob/d3ad3c3361cbc55d36ec329e1ab7b2799e09e7b7/packages/github_copilot/agent_framework_github_copilot/_agent.py#L63">63</a>, <a href="https://github.com/microsoft/agent-framework/blob/d3ad3c3361cbc55d36ec329e1ab7b2799e09e7b7/packages/github_copilot/agent_framework_github_copilot/_agent.py#L362-L363">362&ndash;363</a>, <a href="https://github.com/microsoft/agent-framework/blob/d3ad3c3361cbc55d36ec329e1ab7b2799e09e7b7/packages/github_copilot/agent_framework_github_copilot/_agent.py#L638-L639">638&ndash;639</a>, <a href="https://github.com/microsoft/agent-framework/blob/d3ad3c3361cbc55d36ec329e1ab7b2799e09e7b7/packages/github_copilot/agent_framework_github_copilot/_agent.py#L654">654</a>, <a href="https://github.com/microsoft/agent-framework/blob/d3ad3c3361cbc55d36ec329e1ab7b2799e09e7b7/packages/github_copilot/agent_framework_github_copilot/_agent.py#L657">657</a>, <a href="https://github.com/microsoft/agent-framework/blob/d3ad3c3361cbc55d36ec329e1ab7b2799e09e7b7/packages/github_copilot/agent_framework_github_copilot/_agent.py#L724">724</a>, <a href="https://github.com/microsoft/agent-framework/blob/d3ad3c3361cbc55d36ec329e1ab7b2799e09e7b7/packages/github_copilot/agent_framework_github_copilot/_agent.py#L747">747</a></td></tr><tr><td><b>TOTAL</b></td><td><b>27200</b></td><td><b>3184</b></td><td><b>88%</b></td><td>&nbsp;</td></tr></tbody></table></details>

## Python Unit Test Overview
| 5439 | 20 :zzz: | 0 :x: | 0 :fire: | 1m 23s :stopwatch: |

python/packages/github_copilot/agent_framework_github_copilot/_agent.py
```diff
@@ -274,16 +319,33 @@ async def start(self) -> None:
...
-            if cli_path:
-                client_options["cli_path"] = cli_path
-
-            log_level = self._settings.get("log_level")
-            if log_level:
-                client_options["log_level"] = log_level  # type: ignore[typeddict-item]
-
-            self._client = CopilotClient(client_options if client_options else None)
+            cli_path = self._settings.get("cli_path") or None
+            log_level = self._settings.get("log_level") or "info"
```
moonbox3: Should we be defaulting to info? Previously, if no `log_level` was set, it wasn't passed to the client at all, letting the SDK pick its own default. Now it always sends "info". Could this suppress output for users who expected the SDK default to be something more verbose? Should we consider using `self._settings.get("log_level") or None` and only passing it to `SubprocessConfig` when set, to preserve SDK-default behavior.

python/packages/github_copilot/agent_framework_github_copilot/_agent.py
```diff
@@ -730,43 +795,153 @@ async def _create_session(
...
+        system_message = opts.get("system_message") or self._default_options.get("system_message") or None
+        # on_permission_request is required by the SDK; fall back to deny-all if not configured.
+        permission_handler: PermissionHandlerType = (
+            opts.get("on_permission_request") or self._permission_handler or _deny_all_permissions
+        )
+        mcp_servers = opts.get("mcp_servers") or self._mcp_servers or None
+        tools = self._prepare_tools(self._tools) if self._tools else None
 
-        system_message = opts.get("system_message") or self._default_options.get("system_message")
+        create_kwargs: dict[str, Any] = {
```
moonbox3: The old code used `SessionConfig` / `ResumeSessionConfig` TypedDicts, which gave static checking on config keys. Building a `dict[str, Any]` and splatting with `**create_kwargs` means a typo like `"modell"` will silently be accepted and ignored. Is it possible to use the new SDK's typed `kwargs` or a `TypedDict` here to preserve compile-time checking?

Same comment for `resume_kwargs`.

python/packages/github_copilot/agent_framework_github_copilot/_agent.py
```diff
@@ -730,43 +795,153 @@ async def _create_session(
...
 
-        if self._permission_handler:
-            config["on_permission_request"] = self._permission_handler
+        return await self._client.resume_session(session_id, **resume_kwargs)
 
-        if self._mcp_servers:
-            config["mcp_servers"] = self._mcp_servers
 
-        return await self._client.resume_session(session_id, config)
+class GitHubCopilotAgent(AgentTelemetryLayer, RawGitHubCopilotAgent[OptionsT], Generic[OptionsT]):
```
moonbox3: The new `RawGitHubCopilotAgent` class, the telemetry settings (otlp_endpoint, otel_file_path, etc.), `TelemetryConfig` plumbing, and the `configure_otel_providers()` calls in samples all look like net-new features beyond what's needed for the SDK 0.2.x migration. Could these be split into a follow-up PR? Bundling them here makes it harder to review the migration in isolation and harder to revert if the OTel integration needs changes. 

Also note the MRO interaction: `AgentTelemetryLayer.run()` passes middleware, tools, compaction_strategy, tokenizer, function_invocation_kwargs, and client_kwargs to super().run(), but `RawGitHubCopilotAgent.run()` (line 398) catches them all in `**_kwargs` and discards them silently. This works but is fragile — the `# pyright: ignore[reportIncompatibleMethodOverride]` on line 915 papers over the signature mismatch.

python/packages/github_copilot/agent_framework_github_copilot/_agent.py
droideronline: Good point - fixed. `log_level` is now only passed to `SubprocessConfig` when explicitly set, preserving SDK default behavior when unset.

python/packages/github_copilot/agent_framework_github_copilot/_agent.py
droideronline: Good catch. SDK 0.2.x removed `SessionConfig` / `ResumeSessionConfig` TypedDicts, so we can't restore those directly. Fixed by calling `create_session` and `resume_session` with explicit keyword arguments instead of a `dict[str, Any]` splat - this gives us full static checking from the SDK's function signatures.

python/packages/github_copilot/agent_framework_github_copilot/_agent.py
droideronline: Agreed on the scope concern. The `RawGitHubCopilotAgent` split, `TelemetryConfig` plumbing, and OTel integration were added to keep traces working end-to-end after the migration, but you are right that they deserve their own focused review. I will open a follow-up PR for the telemetry features and scope this one back to the SDK 0.2.x import migration only. Also noted on the MRO fragility with `**_kwargs` - that will be addressed in the follow-up as part of a cleaner integration design.

moonbox3: Sounds good, thank you!

moonbox3 (approved):

python/packages/github_copilot/pyproject.toml
```diff
@@ -24,7 +24,7 @@ classifiers = [
 ]
 dependencies = [
     "agent-framework-core>=1.0.0,<2",
-    "github-copilot-sdk>=0.1.31,<0.1.33; python_version >= '3.11'",
+    "github-copilot-sdk>=0.2.0,<0.3.0; python_version >= '3.11'",
```
eavanvalkenburg: we prefer to have smaller increments allowed when dealing with pre-release packages like this, could we make it so that we allow:

```suggestion
    "github-copilot-sdk>=0.1.31,<=0.2.0; python_version >= '3.11'",
```
That way existing users are not broken, and we don't go too far up to the latest and potentially pull in breaking packages, I noticed in the `0.2.1` release notes that that has breaking changes compared to `0.2.0` so the proposed range is too broad already.

droideronline: Good point - fixed in the latest commit. Applied `<=0.2.0` as the upper bound to avoid pulling in 0.2.1+ breaking changes. I kept `>=0.2.0` as the lower bound (rather than `>=0.1.31`) because this migration changes all import paths to 0.2.x-only modules (`copilot.generated.session_events`, `copilot.session`, `copilot.tools`) - installing 0.1.31 would satisfy the constraint but fail at import time. So the effective range is pinned to exactly 0.2.0 for now. `uv.lock` updated alongside.

droideronline:
@eavanvalkenburg @moonbox3 - Can I get any update on this, we need Otel in one of our use case.

moonbox3:
@eavanvalkenburg please have a look when you can.

python/packages/github_copilot/pyproject.toml
```diff
@@ -24,7 +24,7 @@ classifiers = [
 ]
 dependencies = [
     "agent-framework-core>=1.0.0,<2",
-    "github-copilot-sdk>=0.1.31,<0.1.33; python_version >= '3.11'",
+    "github-copilot-sdk>=0.2.0,<=0.2.0; python_version >= '3.11'",
```
eavanvalkenburg: let's update to `0.2.1` now!

eavanvalkenburg (commented):
overall looks good, let's get on the latest with this PR

python/packages/github_copilot/pyproject.toml
droideronline: Done, thanks!

PR #5111, opened 2026-04-05
Python: fix(python/openai): use full-history mode by default to fix tool call streaming with OpenRouter

### LINKED ISSUE #5105: Python: [Bug]: Streaming fails with OpenRouter 400: tool role without preceding tool_calls
## Summary
Streaming agent execution intermittently fails after a tool call with OpenRouter/OpenAI returning:
`Invalid parameter: messages with role 'tool' must be a response to a preceeding message with 'tool_calls'.`
The failure bubbles up as:
- `openai.BadRequestError` from `openai/resources/responses/responses.py`
- wrapped by `agent_framework_openai._chat_client.OpenAIChatClient`
## Observed behavior
A valid tool run occurs, then the next streamed model call fails with HTTP 400.
Provider payload error:
- `param: messages.[0].role`
- message: `messages with role 'tool' must be a response to a preceeding message with 'tool_calls'`
## Expected behavior
Framework should always send tool-role messages only immediately following assistant messages that include matching `tool_calls` in the same request sequence, so streaming continues normally after tool results.
## Actual behavior
Streaming aborts with provider 400 and task fails.

```markdown

ai-api-1  | Traceback (most recent call last):
ai-api-1  |   File "....../runners.py", line 308, in stream_agent
ai-api-1  |     async for update in stream:
ai-api-1  |     ...<2 lines>...
ai-api-1  |             yield (event_type, data)
ai-api-1  |   File "/usr/local/lib/python3.13/site-packages/agent_framework/_types.py", line 2943, in __anext__
ai-api-1  |     update: UpdateT = await self._iterator.__anext__()
ai-api-1  |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ai-api-1  |   File "/usr/local/lib/python3.13/site-packages/agent_framework/_types.py", line 2943, in __anext__
ai-api-1  |     update: UpdateT = await self._iterator.__anext__()
ai-api-1  |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ai-api-1  |   File "/usr/local/lib/python3.13/site-packages/agent_framework/_types.py", line 2943, in __anext__
ai-api-1  |     update: UpdateT = await self._iterator.__anext__()
ai-api-1  |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ai-api-1  |   File "/usr/local/lib/python3.13/site-packages/agent_framework/_tools.py", line 2378, in _stream
ai-api-1  |     async for update in inner_stream:
ai-api-1  |         yield update
ai-api-1  |   File "/usr/local/lib/python3.13/site-packages/agent_framework/_types.py", line 2943, in __anext__
ai-api-1  |     update: UpdateT = await self._iterator.__anext__()
ai-api-1  |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ai-api-1  |   File "/usr/local/lib/python3.13/site-packages/agent_framework/_types.py", line 2943, in __anext__
ai-api-1  |     update: UpdateT = await self._iterator.__anext__()
ai-api-1  |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ai-api-1  |   File "/usr/local/lib/python3.13/site-packages/agent_framework_openai/_chat_client.py", line 558, in _stream
ai-api-1  |     self._handle_request_error(ex)
ai-api-1  |     ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^
ai-api-1  |   File "/usr/local/lib/python3.13/site-packages/agent_framework_openai/_chat_client.py", line 493, in _handle_request_error
ai-api-1  |     raise ChatClientException(
ai-api-1  |     ...<5 lines>...
ai-api-1  |     ) from ex
ai-api-1  | agent_framework.exceptions.ChatClientException: <class 'agent_framework_openai._chat_client.OpenAIChatClient'> service failed to complete the prompt: Error code: 400 - {'error': {'message': 'Provider returned error', 'code': 400, 'metadata': {'raw': '{\n  "error": {\n    "message": "Invalid parameter: messages with role \'tool\' must be a response to a preceeding message with \'tool_calls\'.",\n    "type": "invalid_request_error",\n    "param": "messages.[0].role",\n    "code": null\n  }\n}', 'provider_name': 'OpenAI', 'is_byok': False}}, 'user_id': 'user_2t7Bf1EvP3yaPWbIsLaLDXgTWhk'}
```

### Package Versions

agent-framework: 1.0.0

### Python Version

3.11.12

### Additional Context

_No response_

Bash892:
## Summary

Fixes #5105 — streaming crashes after tool execution when using OpenRouter or any provider that doesn't support `previous_response_id`.

**Root cause**: `_get_conversation_id` returned `response.id` by default (when `store` was not explicitly `False`). This caused the executor to enter server-managed continuation mode — clearing the full conversation history and sending only the tool result in the next request. Providers like OpenRouter ignore `previous_response_id` and received a bare `function_call_output` at position 0 → HTTP 400.

**Fix**: `_get_conversation_id` now returns `None` unless `store=True` is explicitly set. Server-managed conversation mode is opt-in; the default is full-history mode, which is universally compatible.

Users who want the server-managed optimization (OpenAI native only) should set `store=True` in their chat options.

## Changes

- `python/packages/openai/agent_framework_openai/_chat_client.py`: changed `if store is False` → `if store is not True` in `_get_conversation_id`
- `python/packages/openai/tests/openai/test_openai_chat_client.py`: updated streaming event tests to cover both `store=True` and `store=None`; added `test_parse_response_with_store_none_returns_none` as a regression test

## Test plan

- [x] `uv run pytest packages/openai/tests/ -m "not integration"` — 270 passed, 0 failures
- [x] Verify streaming with OpenRouter no longer raises HTTP 400 after tool execution
- [x] Verify `store=True` still enables server-managed conversation mode for OpenAI native


Bash892:

chetantoshniwal (approved):

python/packages/openai/tests/openai/test_openai_chat_client.py
```diff
@@ -1895,6 +1895,24 @@ def test_parse_response_with_store_false() -> None:
...
+    mock_response = MagicMock()
+    mock_response.id = "resp_123"
+    mock_response.conversation = MagicMock()
+    mock_response.conversation.id = "conv_456"
+
+    assert client._get_conversation_id(mock_response, store=None) is None
+    assert client._get_conversation_id(mock_response, store=False) is None
+
+
 def test_parse_response_uses_response_id_when_no_conversation() -> None:
```
chetantoshniwal, line 2801: This test covers the `store=None` and `store=False` negative cases well, but doesn't verify the positive case (`store=True` returning `conv_456`). Adding `assert client._get_conversation_id(mock_response, store=True) == 'conv_456'` would make this a complete tri-state test and guard against future regressions that might accidentally break the `store=True` path.

```suggestion
    assert client._get_conversation_id(mock_response, store=None) is None
    assert client._get_conversation_id(mock_response, store=False) is None
    assert client._get_conversation_id(mock_response, store=True) == "conv_456"
```

chetantoshniwal (commented):
## Automated Code Review

### ✓ Correctness
> This PR fixes a correctness issue where `_get_conversation_id` would return a conversation ID when `store=None` (the default), causing providers that don't support `previous_response_id` (e.g. OpenRouter) to fail. The fix changes the guard from `if store is False` to `if store is not True`, so only an explicit `store=True` opt-in enables server-managed conversation mode. The logic change is correct and the tests are well-structured. One note: the same `_get_conversation_id` method in `_responses_client.py` (line 448) still uses the old `if store is False` guard, but this may be intentional since the Responses API is OpenAI-native and always supports `previous_response_id`.

### ✓ Security Reliability
> This PR fixes a reliability issue (GitHub #5105) where `_get_conversation_id` would return a conversation ID when `store=None` (the default), causing failures with OpenAI-compatible providers (e.g., OpenRouter) that don't support `previous_response_id`. The fix correctly changes the guard from `if store is False` to `if store is not True`, ensuring only explicit opt-in via `store=True` enables server-managed conversation mode. The code change is minimal and well-tested. No security issues are introduced. Note that the sibling `_responses_client.py` still uses the old `if store is False` check, but this is likely intentional since the Responses API is OpenAI-native and always supports conversation IDs.

### ✓ Test Coverage
> The test coverage for this behavioral change is solid. The new test `test_parse_response_with_store_none_returns_none` directly validates the regression fix (issue #5105) by asserting that both `store=None` and `store=False` return `None` from `_get_conversation_id`. The streaming tests (`test_streaming_response_created_type` and `test_streaming_response_in_progress_type`) were properly updated to verify both `store=True` (returns conversation_id) and `store=None` (returns `None`) paths. One minor gap: the non-streaming `_get_conversation_id` path lacks a direct `store=True` with-conversation positive test in this diff, though this is likely covered by a pre-existing test. A `store=False` case for streaming is untested but is functionally identical to `store=None` under the new `is not True` guard. Overall the tests are meaningful — they assert specific return values rather than just exercising code paths.

### ✓ Design Approach
> The change correctly inverts the semantics of `_get_conversation_id` from opt-out (return None only when `store=False`) to opt-in (return non-None only when `store=True`). Previously, `store=None` (the default when a caller omits the field) would fall through the guard and return a `previous_response_id`-based conversation ID, which breaks providers like OpenRouter that don't support that parameter. The fix properly treats `store=None` the same as `store=False`, making server-managed conversation mode strictly opt-in. The change is minimal, addresses the root cause rather than masking it, and the updated tests adequately cover all three `store` states. No blocking design issues found.

---

python/packages/openai/agent_framework_openai/_chat_client.py
```diff
@@ -673,8 +673,15 @@ def _convert_response_format(self, response_format: Mapping[str, Any]) -> dict[s
...
-        if store is False:
+        """Get the conversation ID from the response when store=True is set.
+
+        Only returns a non-None value when ``store=True`` is explicitly set to opt into
+        server-managed conversation mode (OpenAI native). When ``store`` is ``None``
+        (the default) or ``False``, returns ``None`` so the executor uses full-history
+        mode, which is compatible with providers that do not support ``previous_response_id``
+        (e.g. OpenRouter).
+        """
+        if store is not True:
```
eavanvalkenburg, line 684: so this is not the way, this client is designed to connect to the `Responses API` and that API (at least in both OpenAI and Foundry) uses service side storage by default, so _only_ when `store` is explicitly set to False do we not store anything, the assumption for these clients is that when `store` is not set, the value is `True`, hence the check for `store=False` above. In order to use this as expected with other providers, either create a issue on their end that they adhere to the responses API which means they would have to support the store=True mechanism, or that they raise when store=None, or in your own code, set the options to `store=False`

eavanvalkenburg (changes requested):

Bash892:
Hey @eavanvalkenburg — appreciate the thorough review and totally understand the concern about the Responses API design intent.

That said, I want to push back a bit on the framing. The agent-framework positions itself as a multi-provider framework, not an OpenAI/Foundry-exclusive client. Defaulting to server-managed conversation mode (store=True assumption) means any provider that doesn't support previous_response_id — OpenRouter, Anthropic, Mistral, etc. — fails silently with an HTTP 400. That's a pretty rough default for a framework that advertises broad provider support.

The fix doesn't remove server-managed mode — it just makes it opt-in with store=True, which is one line for OpenAI native users. Meanwhile, everyone else gets a working default out of the box.

On your suggestion to have users set store=False explicitly — that works as a workaround, but it puts the burden on every non-OpenAI user to know about an internal implementation detail just to avoid a crash. That feels like the wrong abstraction layer for the fix.

Would you be open to merging with the opt-in approach, with updated docs making clear that store=True is required for OpenAI native server-managed mode? Happy to add that if it helps get this across the line.

eavanvalkenburg:
@Bash892 the way things work now is that the default of the service is leading, so for most services that means the default is `store=False` and only for Responses (and properly implemented version of it) that the default behavior is True, but the thing is we don't even set a value if the user does not, so for Responses-like services that do not store by default, then you either pass `store=True` or override the `STORES_BY_DEFAULT` class attribute of the client used, which gives you the "right" behavior. We cannot account for every implementation with it's own inconsistencies across the board, so if those services do not implement Responses correctly then that cannot be fixed on our side.

PR #5133, opened 2026-04-06
Python: Fix array input in Chat Completions format losing user message content

### LINKED ISSUE #5112: Python: [Bug]: Agent behaves differently with input when sent as string vs array - Difference is consistent
### Description

In Code Sample, I have shown how I create chat_client and then Agent.

I host it as a Foundry Hosted Agent in Azure Foundry. So I am using Hosting Adapter Library: 
from azure.ai.agentserver.agentframework import from_agent_framework
from_agent_framework(agent).run()

Then I send it same requests as below to the /responses endpoint: 

INPUT A - as a plain string
    "input": "Which Google phones are allowed in University? What is the color and Capacity? Answer using tools that you have available."

INPUT B - as a array, but same text
    "input": [
        {
            "role": "user",
            "content": "Which Google phones are allowed in University? What is the color and Capacity? Answer using tools that you have available."
        }        
    ],

When sent as INPUT A, consistently I get Google Phone, in logs I can see mcp tool called which returns apple, samsung, google, but response is only the Google Phone. This is good.

When sent as INPUT B (which is desirable way I want to use), it always return Google, Apple and Samsung, so ignores the user query.

This is consistent, so if I call 10 times, Input A always work, Input B always give more phone models and not meets user query. Every time in both scenario the mcp_tool is invoked which returns same results whichever is the input.

I test it on local machine, or even post deployment to Azure, so deployment location is not a problem.

Is there some problem here? As I have a Super Agent which in future will have around 6 MCP Tools, so I do not want to hardcode any special phone related system, developer prompt. In responses api, in array, I sent developer, system, some examples in prompt, some continuing conversation, but Input B just always give me more phones and Input A always is precise.

### Code Sample

```markdown
To create Chat Client


        self.chat_client = AzureOpenAIChatClient(
            credential=...,
            endpoint=self.config.azure_openai_endpoint,
            deployment_name=self.config.azure_openai_chat_deployment_name
        )



To create Agent


    BASE_INSTRUCTIONS = "When using tools, treat tool output as raw data.  \nYou must analyze it and return only the parts that satisfy the user’s request.  \nNever return items that do not match the user’s criteria.  \nIf the tool returns mixed or irrelevant results, filter them out before responding.\n"

    AGENT_NAME = "SuperAgent"

        async with mcp_tool:
            agent = self.chat_client.as_agent(
                name=self.AGENT_NAME,
                instructions=instructions,
                tools=[mcp_tool]
            )
```

```markdown

```

### Package Versions

azure-ai-agentserver-agentframework==1.0.0b17 agent-framework==1.0.0rc3

### Python Version

3.12

### Additional Context

_No response_

giles17:
### Motivation and Context

When users send input to the `/responses` endpoint as an array of message objects using Chat Completions format (`{"role": "user", "content": "..."}` without a `"type"` field), the executor fails to recognize it as structured input. This causes the agent to ignore the user's actual query and return unfiltered tool output instead of properly analyzing and filtering results per the user's request.

Fixes #5112

### Description

The root cause is that `_is_openai_multimodal_format` and `_convert_input_to_chat_message` only recognized message objects containing an explicit `"type": "message"` field (Responses API format), but Chat Completions format messages use `{"role": "..", "content": "..."}` without a `type` field. The fix extends both methods to also detect and handle the Chat Completions format by checking for the presence of a `"role"` key when `"type"` is absent. This ensures that array-based input is correctly parsed into a `Message` object with the user's text preserved, so the agent processes the query identically regardless of whether input is sent as a plain string or a message array. Regression tests verify detection and conversion of Chat Completions format messages with both string and list content.

### Contribution Checklist

---

### DIFF
--- a/python/packages/devui/agent_framework_devui/_executor.py
+++ b/python/packages/devui/agent_framework_devui/_executor.py
@@ -859,8 +859,7 @@ class AgentFrameworkExecutor:
         # arrays into the conversion path where all items would be skipped,
         # silently producing an empty message.
         return any(
-            isinstance(item, dict) and item.get("role") == "user"
-            for item in input_data_items
+            isinstance(item, dict) and cast(dict[str, Any], item).get("role") == "user" for item in input_data_items
         )
 
     async def _parse_workflow_input(self, workflow: Any, raw_input: Any) -> Any:
--- a/python/packages/devui/tests/devui/test_multimodal_workflow.py
+++ b/python/packages/devui/tests/devui/test_multimodal_workflow.py
@@ -296,10 +296,13 @@ class TestMultimodalWorkflowInput:
 
         # Non-user roles are accepted when a user-role item is also present
         for role in ("system", "assistant", "tool", "developer"):
-            assert executor._is_openai_multimodal_format([
-                {"role": role, "content": "hi"},
-                {"role": "user", "content": "hello"},
-            ]) is True, f"Expected role {role!r} to be accepted alongside user"
+            assert (
+                executor._is_openai_multimodal_format([
+                    {"role": role, "content": "hi"},
+                    {"role": "user", "content": "hello"},
+                ])
+                is True
+            ), f"Expected role {role!r} to be accepted alongside user"
 
     def test_is_openai_multimodal_format_rejects_non_user_only(self):
         """Arrays with no user-role message are rejected to prevent silent empty message fallback."""
@@ -309,15 +312,21 @@ class TestMultimodalWorkflowInput:
 
         # Non-user-only Chat Completions format
         assert executor._is_openai_multimodal_format([{"role": "system", "content": "hi"}]) is False
-        assert executor._is_openai_multimodal_format([
-            {"role": "system", "content": "hi"},
-            {"role": "assistant", "content": "hello"},
-        ]) is False
+        assert (
+            executor._is_openai_multimodal_format([
+                {"role": "system", "content": "hi"},
+                {"role": "assistant", "content": "hello"},
+            ])
+            is False
+        )
 
         # Non-user-only Responses API format
-        assert executor._is_openai_multimodal_format([
-            {"type": "message", "role": "system", "content": "hi"},
-        ]) is False
+        assert (
+            executor._is_openai_multimodal_format([
+                {"type": "message", "role": "system", "content": "hi"},
+            ])
+            is False
+        )
 
     def test_is_openai_multimodal_format_rejects_malformed_input(self):
         """Test that _is_openai_multimodal_format rejects inputs missing content or with invalid roles."""


moonbox3:
<a href="https://github.com/microsoft/agent-framework/blob/5a726daca4b7eacd6b6665bf45ed43b2f479c904/README.md">Python Test Coverage</a><details><summary>Python Test Coverage Report • </summary><table><tr><th>File</th><th>Stmts</th><th>Miss</th><th>Cover</th><th>Missing</th></tr><tbody><tr><td><b>TOTAL</b></td><td><b>27057</b></td><td><b>3188</b></td><td><b>88%</b></td><td>&nbsp;</td></tr></tbody></table><i>report-only-changed-files is enabled. No files were changed during this commit :)</i></details>

## Python Unit Test Overview
| 5373 | 20 :zzz: | 0 :x: | 0 :fire: | 1m 22s :stopwatch: |

python/packages/devui/agent_framework_devui/_executor.py
```diff
@@ -631,7 +631,11 @@ def _convert_openai_input_to_chat_message(self, input_items: list[Any], Message:
             if isinstance(item, dict):
                 item_dict = cast(dict[str, Any], item)
                 item_type = item_dict.get("type")
-                if item_type == "message":
+                if item_type == "message" or (item_type is None and "role" in item_dict):
+                    message_role = item_dict.get("role")
+                    if message_role is not None and message_role != "user":
+                        logger.debug("Skipping non-user OpenAI message item with role %r", message_role)
+                        continue
```
giles17, line 638: Minor reliability concern: if the input array contains *only* non-user messages (e.g., `[{"role": "system", "content": ".."}]`), every item is skipped here and the fallback at line 784 silently produces an empty `Content.from_text(text="")`. Consider emitting a warning when all items were skipped so that operators can diagnose why a model received an empty user message.

```suggestion
                message_role = item_dict.get("role")
                    if message_role is not None and message_role != "user":
                        logger.debug("Skipping non-user OpenAI message item with role %r", message_role)
                skipped_non_user = True
                        continue
```

python/packages/devui/tests/devui/test_multimodal_workflow.py
```diff
@@ -150,3 +150,113 @@ async def test_parse_workflow_input_still_handles_simple_dict(self):
...
+        input_data = [
+            {"role": "system", "content": "You are a helpful assistant."},
+            {"role": "user", "content": "Hello!"},
+        ]
+
+        result = executor._convert_input_to_chat_message(input_data)
+
+        assert isinstance(result, Message), f"Expected Message, got {type(result)}"
+        assert len(result.contents) == 1
+        assert result.contents[0].text == "Hello!"
```
giles17, line 247: This test only covers the case where a user message is present alongside the skipped system message. Please add a companion test for when ALL messages are non-user (e.g., only system/assistant) to verify the empty-content fallback behavior at line 783-784 of _executor.py. The expected result would be a Message with a single empty-text content, which is worth documenting explicitly.

```suggestion
        assert result.contents[0].text == "Hello!"

    def test_convert_all_non_user_messages_produces_empty_content(self):
        """When all messages are non-user, the result should be a Message with empty text content."""
        from agent_framework import Message

        discovery = MagicMock(spec=EntityDiscovery)
        mapper = MagicMock(spec=MessageMapper)
        executor = AgentFrameworkExecutor(discovery, mapper)

        input_data = [
            {"role": "system", "content": "You are a helpful assistant."},
            {"role": "assistant", "content": "How can I help?"},
        ]

        result = executor._convert_input_to_chat_message(input_data)

        assert isinstance(result, Message)
        assert len(result.contents) == 1
        assert result.contents[0].text == "
```

python/packages/devui/agent_framework_devui/_executor.py
```diff
@@ -828,8 +832,24 @@ def _is_openai_multimodal_format(self, input_data: Any) -> bool:
...
+        if isinstance(first_type, str) and first_type == "message":
+            return True
+        # Also accept Chat Completions format: {"role": "...", "content": "..."}
+        # but require the minimum expected shape to avoid misclassifying
+        # unrelated or malformed list inputs as chat messages.
+        if first_type is not None:
+            return False
+        role = first_dict.get("role")
+        content = first_dict.get("content")
+        valid_roles = {"system", "user", "assistant", "tool", "developer"}
```
giles17: `valid_roles` includes `"system"`, `"assistant"`, and `"tool"`. A list like `[{"role": "system", "content": ".."}]` is accepted as multimodal format, routed to `_convert_input_to_chat_message`, then every item is skipped (non-user), producing an empty `Message(role="user", contents=[Content.from_text(text="")])` with no warning. If the intent is to detect a genuine user-input array, restricting detection to require at least a `"user"` role in the first item (or across all items) would prevent this silent fallback path.

```suggestion
        valid_roles = {"system", "user", "assistant", "tool", "developer"}
        # Only treat as multimodal format when the first item is a user turn;
        # non-user-leading arrays (e.g. system-only) are not valid user input.
        return bool(
            isinstance(role, str)
            and role == "user"
            and "content" in first_dict
            and isinstance(content, str | list)
        )
```

python/packages/devui/agent_framework_devui/_executor.py
```diff
@@ -828,8 +828,12 @@ def _is_openai_multimodal_format(self, input_data: Any) -> bool:
...
         if not isinstance(first_item, dict):
             return False
-        first_type = cast(dict[str, Any], first_item).get("type")
-        return isinstance(first_type, str) and first_type == "message"
+        first_dict = cast(dict[str, Any], first_item)
+        first_type = first_dict.get("type")
+        if isinstance(first_type, str) and first_type == "message":
+            return True
+        # Also accept Chat Completions format: {"role": "...", "content": "..."}
+        return bool(first_type is None and "role" in first_dict)
```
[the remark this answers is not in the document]
giles17: Fixed in `6842d167`.

python/packages/devui/agent_framework_devui/_executor.py
```diff
@@ -631,7 +631,7 @@ def _convert_openai_input_to_chat_message(self, input_items: list[Any], Message:
             if isinstance(item, dict):
                 item_dict = cast(dict[str, Any], item)
                 item_type = item_dict.get("type")
-                if item_type == "message":
+                if item_type == "message" or (item_type is None and "role" in item_dict):
```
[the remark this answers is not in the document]
giles17: Fixed in `6842d167`.

python/packages/devui/tests/devui/test_multimodal_workflow.py
```diff
@@ -150,3 +150,56 @@ async def test_parse_workflow_input_still_handles_simple_dict(self):
...
+        executor = AgentFrameworkExecutor(discovery, mapper)
+
+        # Chat Completions format with list content (input_text items)
+        input_data = [
+            {
+                "role": "user",
+                "content": [
+                    {"type": "input_text", "text": "Describe this image"},
+                    {"type": "input_image", "image_url": TEST_IMAGE_DATA_URI},
+                ],
```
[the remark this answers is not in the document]
giles17: Fixed in `6842d167`.

python/packages/devui/tests/devui/test_multimodal_workflow.py
```diff
@@ -150,3 +150,56 @@ async def test_parse_workflow_input_still_handles_simple_dict(self):
...
+        from agent_framework import Message
+
+        discovery = MagicMock(spec=EntityDiscovery)
+        mapper = MagicMock(spec=MessageMapper)
+        executor = AgentFrameworkExecutor(discovery, mapper)
+
+        # Chat Completions format (no type field, string content)
+        input_data = [{"role": "user", "content": "Which Google phones are allowed?"}]
+
+        result = executor._convert_input_to_chat_message(input_data)
```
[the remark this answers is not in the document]
giles17: Fixed in `6842d167`.

python/packages/devui/agent_framework_devui/_executor.py
giles17: Fixed in `6842d167`.

python/packages/devui/tests/devui/test_multimodal_workflow.py
giles17: Fixed in `6842d167`.

python/packages/devui/agent_framework_devui/_executor.py
giles17: Fixed in `6842d167`.

eavanvalkenburg:
This seems wholly unrelated to the issue that it links to and attempts to allow `ChatCompletion` input on a `Responses` endpoint, which we should not support.

Issue #1188, opened 2025-10-04
Python SDK :  Add native vector database connectors and retrieval patterns


biswapm:
Currently, the Agent Framework lacks built-in support for vector databases, which are essential for RAG patterns in production agent systems.

Proposed Enhancement:

Add first-class connectors for popular vector databases:

Azure AI Search (given Azure ecosystem focus)
Pinecone
Weaviate
Qdrant
Chroma
PostgreSQL with pgvector

Include built-in retrieval patterns:

Semantic search tools for agents
Hybrid search (vector + keyword)
Re-ranking capabilities
Metadata filtering

markwallace-microsoft:
The Semantic Kernel already contains first-class connectors for popular vector databases (see https://learn.microsoft.com/en-us/semantic-kernel/concepts/vector-store-connectors/out-of-the-box-connectors/?pivots=programming-language-python)

These connectors are general purpose and can be used with Agent Framework.

I'll keep this issue option so cover creating samples and dociumentation.

luismanez:
But are there plans to move SK Vector DB connectors into Agent Framework? I think is a bit confusing if we mix both things. Announcement said:
> It unifies the enterprise-ready foundations of Semantic Kernel with the innovative orchestration of AutoGen, so teams no longer have to choose between experimentation and production.

eavanvalkenburg:
For .Net the vector store abstractions are already outside of SK, and they are built to be usable between both SK and AF, for Python we do want to bring this into AF, this issue tracks that.

eavanvalkenburg:
We are actively working on this. Here is the implementation plan, broken into mergeable phases:

### Phase 1: Core Embedding Abstractions & OpenAI Implementation ✅
_PR [#4153](https://github.com/microsoft/agent-framework/pull/4153)_
- Embedding types in `_types.py` (`Embedding[EmbeddingT]`, `GeneratedEmbeddings`, `EmbeddingGenerationOptions`)
- `SupportsGetEmbeddings` protocol and `BaseEmbeddingClient` ABC in `_clients.py`
- `EmbeddingTelemetryLayer` in `observability.py`
- OpenAI and Azure OpenAI embedding clients
- Unit tests, integration tests, and samples

### Phase 2: Embedding Generators for Existing Providers ✅
_PR [#4207](https://github.com/microsoft/agent-framework/pull/4207)_
- Embedding generators for existing providers
- Shared embedding client contracts and provider-specific validation

[...]

PR #5139, opened 2026-04-07
Python: cosmos context provider

bambriz:
### Motivation and Context
 
 Adds an Azure Cosmos DB **context provider** to the `agent-framework-azure-cosmos` package, enabling RAG-style 
context injection from Cosmos DB before model invocation and automatic write-back of request/response messages after
 each run.
 
 ### Description
 
 **New: `CosmosContextProvider`**
 - Implements the `ContextProvider` protocol for Azure Cosmos DB
 - Supports three retrieval modes via `CosmosContextSearchMode`: `VECTOR`, `FULL_TEXT`, and `HYBRID`
 - Joins filtered `user`/`assistant` messages into a retrieval query string
 - Writes request/response messages back into the knowledge container after each run (`after_run`)
 - Configurable: `top_k`, `scan_limit`, `partition_key`, `weights` (for hybrid RRF), `content_field_names`
 - Supports environment-based settings resolution and flexible credential types
 
 **Samples** (`python/samples/02-agents/context_providers/azure_cosmos/`)
 - `cosmos_context_basics.py` — Vector search context provider usage
 - `cosmos_context_fulltext.py` — Full-text retrieval mode
 - `cosmos_context_hybrid.py` — Hybrid retrieval with RRF weights
 
 **Tests**
 - `test_cosmos_context_provider.py` — Unit tests covering all search modes, message filtering, write-back, and 
error handling (95% coverage)
 
 **Other changes**
 - Updated `AGENTS.md` with context provider documentation and usage examples
 - Updated package `README.md` with detailed context provider API docs
 
 ### Contribution Checklist
   
   - [x] The PR follows the [Contribution Guidelines](https://github.com/microsoft/agent-framework/blob/main/CONTRIBUTING.md) and [coding standard](https://github.com/microsoft/agent-framework/blob/main/python/CODING_STANDARD.md)
   - [x] All unit tests pass (34/34), and I have added new tests where possible
   - [ ] Is this a breaking change? **No**


python/packages/azure-cosmos/agent_framework_azure_cosmos/_context_provider.py
```diff
@@ -0,0 +1,875 @@
...
+
+
+class AzureCosmosContextProvider(BaseContextProvider):
+    """Azure Cosmos DB-backed context provider using BaseContextProvider hooks."""
+
+    DEFAULT_SOURCE_ID: ClassVar[str] = "azure_cosmos_context"
+    DEFAULT_CONTEXT_PROMPT: ClassVar[str] = "Use the following context to answer the question:"
+    _DEFAULT_RESULT_LIMIT: ClassVar[int] = 5
+    _DEFAULT_SCAN_LIMIT: ClassVar[int] = 25
+    _DEFAULT_SEARCH_MODE: ClassVar[CosmosContextSearchMode] = CosmosContextSearchMode.FULL_TEXT
```
bambriz: default should be vector

python/packages/azure-cosmos/agent_framework_azure_cosmos/_context_provider.py
```diff
@@ -0,0 +1,875 @@
...
+        top_k: int | None = None,
+        scan_limit: int | None = None,
+        default_search_mode: CosmosContextSearchMode = _DEFAULT_SEARCH_MODE,
+        id_field_name: str = "id",
+        content_field_names: Sequence[str] = ("content", "text"),
+        title_field_name: str | None = "title",
+        url_field_name: str | None = "url",
+        message_field_name: str | None = "message",
+        metadata_field_name: str | None = "metadata",
+        vector_field_name: str | None = None,
```
bambriz: should have default vector field name

python/packages/azure-cosmos/agent_framework_azure_cosmos/_context_provider.py
```diff
@@ -0,0 +1,875 @@
...
+        self.top_k = self._validate_positive_int(settings.get("top_k"), default=self.top_k, name="top_k")
+        self.scan_limit = self._validate_positive_int(
+            settings.get("scan_limit"), default=self.scan_limit, name="scan_limit"
+        )
+
+        if self._cosmos_client is None:
+            self._cosmos_client = CosmosClient(
+                url=settings["endpoint"],  # type: ignore[arg-type]
+                credential=credential or settings["key"].get_secret_value(),  # type: ignore[arg-type,union-attr]
+                user_agent_suffix=AGENT_FRAMEWORK_USER_AGENT,
```
bambriz: add something related to cosmos db python (to track what cosmos sdk is being used)

python/packages/azure-cosmos/agent_framework_azure_cosmos/_context_provider.py
```diff
@@ -0,0 +1,875 @@
...
+
+    endpoint: str | None
+    database_name: str | None
+    container_name: str | None
+    key: SecretString | None
+    top_k: int | None
+    scan_limit: int | None
+
+
+class AzureCosmosContextProvider(BaseContextProvider):
```
bambriz: basecontextproivder is now contextproivder

python/packages/azure-cosmos/agent_framework_azure_cosmos/_context_provider.py
```diff
@@ -0,0 +1,875 @@
...
+        self._database_client = self._cosmos_client.get_database_client(self.database_name)
+
+    async def before_run(
+        self,
+        *,
+        agent: SupportsAgentRun,
+        session: AgentSession,
+        context: SessionContext,
+        state: dict[str, Any],
+        search_mode: CosmosContextSearchMode | None = None,
```
bambriz: param search_mode to partitio_key only for constructor. Not reachable here in before_run

python/packages/azure-cosmos/agent_framework_azure_cosmos/_context_provider.py
```diff
@@ -0,0 +1,875 @@
...
+
+class AzureCosmosContextProvider(BaseContextProvider):
+    """Azure Cosmos DB-backed context provider using BaseContextProvider hooks."""
+
+    DEFAULT_SOURCE_ID: ClassVar[str] = "azure_cosmos_context"
+    DEFAULT_CONTEXT_PROMPT: ClassVar[str] = "Use the following context to answer the question:"
+    _DEFAULT_RESULT_LIMIT: ClassVar[int] = 5
+    _DEFAULT_SCAN_LIMIT: ClassVar[int] = 25
+    _DEFAULT_SEARCH_MODE: ClassVar[CosmosContextSearchMode] = CosmosContextSearchMode.FULL_TEXT
+    _DEFAULT_RRF_WEIGHTS: ClassVar[tuple[float, float]] = (1.0, 1.0)
```
bambriz: can't assume it will be list of 2. There should be no default with weights, hybeid search can be ran without weights. Only use if customer provides weights. (Update query builder accordingly)

python/packages/azure-cosmos/agent_framework_azure_cosmos/_context_provider.py
```diff
@@ -0,0 +1,875 @@
...
+            effective_score = provider_score if provider_score > 0 else fallback_score
+            if effective_score <= 0:
+                continue
+
+            ranked_messages.append((effective_score, fallback_score, -position, message))
+
+        ranked_messages.sort(reverse=True)
+        return [message for _, _, _, message in ranked_messages[:top_k]]
+
+    def _annotate_rank_scores(
```
bambriz: we don't need to rerank scores.

moonbox3:
<a href="https://github.com/microsoft/agent-framework/blob/c07c8332e3a6bbc832ab5aa4c45e17bccfc05a0c/README.md">Python Test Coverage</a><details><summary>Python Test Coverage Report • </summary><table><tr><th>File</th><th>Stmts</th><th>Miss</th><th>Cover</th><th>Missing</th></tr><tbody><tr><td colspan="5"><b>packages/azure-cosmos/agent_framework_azure_cosmos</b></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/c07c8332e3a6bbc832ab5aa4c45e17bccfc05a0c/packages/azure-cosmos/agent_framework_azure_cosmos/_context_provider.py">_context_provider.py</a></td><td>212</td><td>18</td><td>91%</td><td><a href="https://github.com/microsoft/agent-framework/blob/c07c8332e3a6bbc832ab5aa4c45e17bccfc05a0c/packages/azure-cosmos/agent_framework_azure_cosmos/_context_provider.py#L182">182</a>, <a href="https://github.com/microsoft/agent-framework/blob/c07c8332e3a6bbc832ab5aa4c45e17bccfc05a0c/packages/azure-cosmos/agent_framework_azure_cosmos/_context_provider.py#L188">188</a>, <a href="https://github.com/microsoft/agent-framework/blob/c07c8332e3a6bbc832ab5aa4c45e17bccfc05a0c/packages/azure-cosmos/agent_framework_azure_cosmos/_context_provider.py#L269">269</a>, <a href="https://github.com/microsoft/agent-framework/blob/c07c8332e3a6bbc832ab5aa4c45e17bccfc05a0c/packages/azure-cosmos/agent_framework_azure_cosmos/_context_provider.py#L299">299</a>, <a href="https://github.com/microsoft/agent-framework/blob/c07c8332e3a6bbc832ab5aa4c45e17bccfc05a0c/packages/azure-cosmos/agent_framework_azure_cosmos/_context_provider.py#L318">318</a>, <a href="https://github.com/microsoft/agent-framework/blob/c07c8332e3a6bbc832ab5aa4c45e17bccfc05a0c/packages/azure-cosmos/agent_framework_azure_cosmos/_context_provider.py#L326-L330">326&ndash;330</a>, <a href="https://github.com/microsoft/agent-framework/blob/c07c8332e3a6bbc832ab5aa4c45e17bccfc05a0c/packages/azure-cosmos/agent_framework_azure_cosmos/_context_provider.py#L339">339</a>, <a href="https://github.com/microsoft/agent-framework/blob/c07c8332e3a6bbc832ab5aa4c45e17bccfc05a0c/packages/azure-cosmos/agent_framework_azure_cosmos/_context_provider.py#L379">379</a>, <a href="https://github.com/microsoft/agent-framework/blob/c07c8332e3a6bbc832ab5aa4c45e17bccfc05a0c/packages/azure-cosmos/agent_framework_azure_cosmos/_context_provider.py#L414">414</a>, <a href="https://github.com/microsoft/agent-framework/blob/c07c8332e3a6bbc832ab5aa4c45e17bccfc05a0c/packages/azure-cosmos/agent_framework_azure_cosmos/_context_provider.py#L420">420</a>, <a href="https://github.com/microsoft/agent-framework/blob/c07c8332e3a6bbc832ab5aa4c45e17bccfc05a0c/packages/azure-cosmos/agent_framework_azure_cosmos/_context_provider.py#L423-L426">423&ndash;426</a></td></tr><tr><td><b>TOTAL</b></td><td><b>37605</b></td><td><b>4368</b></td><td><b>88%</b></td><td>&nbsp;</td></tr></tbody></table></details>

## Python Unit Test Overview
| 7483 | 34 :zzz: | 0 :x: | 0 :fire: | 1m 55s :stopwatch: |

bambriz:
@moonbox3 could we get a review on this PR?

python/packages/azure-cosmos/agent_framework_azure_cosmos/_context_provider.py
```diff
@@ -0,0 +1,416 @@
...
+
+        container = await self._get_container()
+        effective_pk = self._resolve_partition_key(context)
+        pk_field = await self._get_partition_key_field()
+
+        agent_name = getattr(agent, "name", None)
+        user_id = context.metadata.get("user_id") if context.metadata else None
+
+        base_sort_key = time.time_ns()
+        for index, message in enumerate(writeback):
```
moonbox3, line 280: Should we be doing the writeback one upsert at a time? If message N raises (429, transient network, throttling), messages 0..N-1 are already committed and the exception propagates out of `after_run`. Next turn, `before_run` will retrieve half-written exchanges (user without assistant, or vice versa) and quietly poison the RAG context. Neither the mem0 (single `add()`) nor redis (single `load()`) sibling has this shape. Wondering if we want a try/except around `upsert_item` that logs and continues, or batch the documents so it is all-or-nothing per turn.

python/packages/azure-cosmos/agent_framework_azure_cosmos/_context_provider.py
```diff
@@ -0,0 +1,416 @@
...
+
+    async def _execute_retrieval_query(
+        self, query_text: str, query_terms: tuple[str, ...], partition_key: str
+    ) -> list[dict[str, Any]]:
+        """Build and execute the Cosmos retrieval query for the configured search mode."""
+        fields = list(self.content_field_names)
+        if self.message_field_name and self.message_field_name not in fields:
+            fields.append(self.message_field_name)
+        select = ", ".join(f"c.{f}" for f in fields)
+        base = f"SELECT TOP {self.scan_limit} {select} FROM c"  # noqa: S608  # nosec B608
```
moonbox3, line 351: The `# nosec` suppresses bandit, but `select` is built from `content_field_names` / `message_field_name` (line 332-335) and `vector_field_name` (line 346, 351) is also interpolated into the ORDER BY, all constructor strings with no validation. Cosmos NoSQL cannot parameterize field names, so the only defense is constructor-time validation. In a multi-tenant agent host where field names come from request/tenant config, a value like `embedding}, @x) OFFSET 0 LIMIT 1; --` injects. Have we thought about a `^[A-Za-z_][A-Za-z0-9_]*$` check at `__init__` for every user-supplied field name?

```suggestion
        base = f"SELECT TOP {self.scan_limit} {select} FROM c"  # noqa: S608  # nosec B608 - field names validated in __init__
```

(plus the validation in `__init__` for `content_field_names`, `message_field_name`, `vector_field_name`)

python/packages/azure-cosmos/agent_framework_azure_cosmos/_context_provider.py
```diff
@@ -0,0 +1,416 @@
...
+
+        self.top_k = top_k if top_k is not None and top_k > 0 else 5
+        self.scan_limit = scan_limit if scan_limit is not None and scan_limit > 0 else 25
+        self.search_mode = search_mode
+        self.content_field_names = tuple(content_field_names)
+        self.message_field_name = message_field_name
+        self.vector_field_name = vector_field_name
+        self.embedding_function = embedding_function
+        self.partition_key = partition_key
+        self.weights = tuple(float(w) for w in weights) if weights is not None else None
```
moonbox3: Cosmos `RRF()` requires weights in `[0, 1]`. No bounds check here means a misconfigured `weights=[2.0, 1.0]` (which is what `test_hybrid_weights_passed_through` actually uses) crashes mid-`before_run` against a real endpoint with an opaque server error inside an agent turn rather than a clear `ValueError` at construction. Worth validating at init?

```suggestion
        if weights is not None:
            for w in weights:
                if not 0.0 <= float(w) <= 1.0:
                    raise ValueError(f"RRF weight {w!r} must be in [0, 1]")
            self.weights = tuple(float(w) for w in weights)
        else:
            self.weights = None
```

python/packages/azure-cosmos/agent_framework_azure_cosmos/_context_provider.py
```diff
@@ -0,0 +1,416 @@
...
+        """Read the container's partition key path and return the field name.
+
+        The result is cached for the lifetime of the provider.
+        """
+        if self._partition_key_field is not None:
+            return self._partition_key_field
+
+        container = await self._get_container()
+        properties = await container.read()
+        paths: list[str] = properties.get("partitionKey", {}).get("paths", [])  # type: ignore[assignment]
```
moonbox3, line 186: What happens on a container with a hash/synthetic partition key, or any response shape where `paths` is empty? `paths[0]` raises `IndexError` and crashes `after_run` with no useful message. Should we guard with an explicit error?

```suggestion
        paths: list[str] = properties.get("partitionKey", {}).get("paths", [])  # type: ignore[assignment]
        if not paths:
            raise ValueError(
                f"Could not determine partition key path for container '{self.container_name}'."
            )
        field = paths[0].lstrip("/")
        self._partition_key_field = field
```

python/packages/azure-cosmos/agent_framework_azure_cosmos/_context_provider.py
```diff
@@ -0,0 +1,416 @@
...
+        return self
+
+    async def __aexit__(
+        self,
+        exc_type: type[BaseException] | None,
+        exc_val: BaseException | None,
+        exc_tb: Any,
+    ) -> None:
+        try:
+            await self.close()
```
moonbox3, line 327: If the `async with` block already raised and `close()` also raises, the close error is silently dropped. That can mask real client/socket leak signals from azure-cosmos. Other Azure clients in the repo at least log here. Should this `except Exception` log with `exc_info=True` before swallowing?

```suggestion
        try:
            await self.close()
        except Exception:
            if exc_type is None:
                raise
            logger.warning("Error closing Cosmos client during exception handling.", exc_info=True)
```

moonbox3 (commented):
Please also have a look at the failing CI/CD checks.

python/packages/azure-cosmos/agent_framework_azure_cosmos/_context_provider.py
bambriz: We actually should not attempt to validate; we should allow cosmos to error if the weights don't work. This is also wrong on my end, we should not be clamping the weights between 0 and 1 as the weights are actually relative according to docs, so any value should be allowed through if cosmos rejects it then we let it show it in a cosmos error. Will update this alongside relevant tests.

python/packages/azure-cosmos/agent_framework_azure_cosmos/_context_provider.py
bambriz: That is good point. Batching should be what we do here, especially since we are expecting essentially two items to upsert every after_run call. This is also how the history provider handles it so I will update to use batching. It is better to have all-or-nothing to avoid "poisoning" the context container with incomplete information.

python/packages/azure-cosmos/agent_framework_azure_cosmos/_context_provider.py
bambriz: Currently in Cosmos DB users can already pass in their own query string, if they purposely pass in with embedding}, @x) OFFSET 0 LIMIT 1; -- it won't result in an injection but instead cosmos returns a bad request 400 error. Generally design of this context provider is to allow cosmos to handle the error as it is already resilient and it is query only. Sanitation won't be necessary here as long as it is something we pass into Cosmos.

python/packages/azure-cosmos/agent_framework_azure_cosmos/_context_provider.py
bambriz: In this case if paths is empty it means there was an error reading container properties from cosmos which would mean an error should happen from cosmos, If this is empty nothing else can proceed since we need the pk path to upsert items correctly. This should also only run once as we will cache the pk path so error should be minimal.

python/packages/azure-cosmos/agent_framework_azure_cosmos/_context_provider.py
bambriz: The aexit here follows the same pattern used by CosmosHistoryProvider (lines 258-262) if the async with block already raised, the close() error is swallowed to avoid masking the original exception. If no prior exception exists, the close() error is re-raised normally.

Logging diagnostics for Cosmos operations is best configured on the CosmosClient itself (via logging_enable=True and enable_diagnostics_logging=True) rather than adding ad-hoc log statements in the provider. This way the user gets full SDK-level diagnostics including status
codes, sub-status codes, and request charges for all operations, including close(). We've added a Logging & Diagnostics section to the README documenting how to set this up.

eavanvalkenburg:
I would like this to be implemented using the vector store abstractions as that would provide additional capabilities over this model, because this model relies on a fixed structure, while the vector stores take care of much more dynamic features and models, so closing this and adding cosmos vector search to the backlog for vector stores: https://github.com/microsoft/agent-framework/issues/1188

PR #5140, opened 2026-04-07
Python: fix: remove dead code from test_middleware_with_agent.py

Serjbory:
### Motivation and Context

Three `with pytest.raises(...)` blocks in `test_middleware_with_agent.py` contain multiple statements. Because `pytest.raises` exits immediately after the **first** exception is caught, all subsequent statements in the block are dead code that never executes. This creates a false sense of test coverage — the `await agent.run(...)` calls appear to be tested but are actually unreachable.

In all three cases, the `Agent(...)` constructor is what raises the expected `MiddlewareException`. The `agent.run()` calls after it are dead code.

### Description

**3 locations fixed in `test_middleware_with_agent.py`:**

**1. `test_decorator_and_type_mismatch` (line ~1470)**
- Moved `@agent_middleware` decorator outside the `pytest.raises` block (decorators don't raise)
- Removed unreachable `await agent.run(...)` after `Agent(...)` constructor

**2. `test_neither_decorator_nor_type` (line ~1597)**
- Removed unreachable `await agent.run(...)` after `Agent(...)` constructor

**3. `test_insufficient_parameters_error` (line ~1606)**
- Moved `@agent_middleware` decorator outside the `pytest.raises` block
- Removed unreachable `await agent.run(...)` after `Agent(...)` constructor

All 53 tests in `test_middleware_with_agent.py` pass after the changes.

### Contribution Checklist


eavanvalkenburg:
Closing, I think this was already covered, otherwise we will pick it up again.

PR #5183, opened 2026-04-09
.NET: Support reflection for discovery of resources and scripts in class-based skills

SergeyMenshykh:
This PR adds support for using reflection to discover resources and scripts in class-based skills, enabling a more declarative approach via attributes.

## Changes

- Add `AgentSkillResourceAttribute` and `AgentSkillScriptAttribute` for declarative resource/script discovery
- Enhance `AgentClassSkill` to discover resources and scripts via reflection
- Update samples to demonstrate the new attribute-based approach

### DIFF
--- a/dotnet/samples/02-agents/AgentSkills/Agent_Step03_ClassBasedSkills/Agent_Step03_ClassBasedSkills.csproj
+++ b/dotnet/samples/02-agents/AgentSkills/Agent_Step03_ClassBasedSkills/Agent_Step03_ClassBasedSkills.csproj
@@ -6,7 +6,7 @@
 
     <Nullable>enable</Nullable>
     <ImplicitUsings>enable</ImplicitUsings>
-    <NoWarn>$(NoWarn);MAAI001</NoWarn>
+    <NoWarn>$(NoWarn);MAAI001;IDE0051</NoWarn>
   </PropertyGroup>
 
   <ItemGroup>
--- a/dotnet/samples/02-agents/AgentSkills/Agent_Step05_SkillsWithDI/Program.cs
+++ b/dotnet/samples/02-agents/AgentSkills/Agent_Step05_SkillsWithDI/Program.cs
@@ -116,7 +116,7 @@ Console.WriteLine($"Agent: {response.Text}");
 /// in both its resource and script functions. This enables clean separation of
 /// concerns and testability while retaining the class-based skill pattern.
 /// </remarks>
-internal sealed class WeightConverterSkill : AgentClassSkill
+internal sealed class WeightConverterSkill : AgentClassSkill<WeightConverterSkill>
 {
     private IReadOnlyList<AgentSkillResource>? _resources;
     private IReadOnlyList<AgentSkillScript>? _scripts;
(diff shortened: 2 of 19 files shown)


westey-m (approved):

dotnet/samples/02-agents/AgentSkills/Agent_Step03_ClassBasedSkills/Agent_Step03_ClassBasedSkills.csproj
```diff
@@ -6,7 +6,7 @@
 
     <Nullable>enable</Nullable>
     <ImplicitUsings>enable</ImplicitUsings>
-    <NoWarn>$(NoWarn);MAAI001</NoWarn>
+    <NoWarn>$(NoWarn);MAAI001;IDE0051</NoWarn>
```
rogerbarreto, line 9: “Remove unused private member” sounds a good message/warning to keep alive.

Might suggest adding it into the specific files instead of full csproj

rogerbarreto: Never mind, this is a sample

dotnet/samples/02-agents/AgentSkills/Agent_Step05_SkillsWithDI/Program.cs
```diff
@@ -113,14 +114,13 @@ Use this skill when the user asks to convert between distance units (miles and k
...
 /// This skill resolves <see cref="ConversionService"/> from the DI container
-/// in both its resource and script functions. This enables clean separation of
-/// concerns and testability while retaining the class-based skill pattern.
+/// in both its resource and script methods. Methods with an <see cref="IServiceProvider"/>
+/// parameter are automatically injected by the framework. Properties and methods annotated
+/// with <see cref="AgentSkillResourceAttribute"/> and <see cref="AgentSkillScriptAttribute"/>
+/// are automatically discovered via reflection.
 /// </remarks>
-internal sealed class WeightConverterSkill : AgentClassSkill
+internal sealed class WeightConverterSkill : AgentClassSkill<WeightConverterSkill>
```
rogerbarreto, line 122: Suggest Keeping both classes the non-generic AgentClassSkill for explicit overrides, and the generic for attribute-based discovery. 

This way we don't break existing and CRTP (Curiously Recurring Template Pattern) is only needed where reflection happens.

```csharp
// Non-generic — explicit override, no CRTP needed
public abstract class AgentClassSkill : AgentSkill { /* existing code */ }

// Generic — extends base, adds reflection discovery
public abstract class AgentClassSkill<TSelf> : AgentClassSkill
    where TSelf : AgentClassSkill<TSelf>
{
    // Overrides Resources/Scripts with reflection-based discovery
    // Adds SerializerOptions, DiscoverResources(), DiscoverScripts()
}
```

SergeyMenshykh: I don't know, having two classes would increase the public API surface we need to support and add cognitive load for consumers - they'd have to figure out which one to use and when.

The generic `AgentClassSkill<TSelf>` already supports both scenarios: you can use attributes for reflection-based discovery, or just override `Resources` and `Scripts` explicitly. Both approaches work within the same class, so there's no real need for a separate non-generic base.
  
A non-generic class would only make sense if we had another subtype and needed to work with them polymorphically, but that is not the case here.

Class-based skills are experimental and are not released yet so we're not breaking anyone.

rogerbarreto (approved):

PR #5185, opened 2026-04-09
Python: Add Hyperlight CodeAct package and docs

### LINKED ISSUE #5187: Python: [Feature]: CodeAct python implementation
### Description

Python CodeAct

### Code Sample

```markdown

```

### Language/SDK

Python

eavanvalkenburg:
### Motivation and Context

Add a concrete, optional CodeAct implementation for Python and capture the cross-SDK design for CodeAct with Hyperlight. This provides a reusable path for long-running agents to execute sandboxed code with provider-owned tools, file mounts, and network allow-lists without baking CodeAct into core.

### Description

- add ADR 0024 plus Python feature design notes for the CodeAct and Hyperlight design
- introduce the alpha `agent-framework-hyperlight` package with `HyperlightCodeActProvider` and `HyperlightExecuteCodeTool`
- add provider-managed tool, file, and network CRUD; derived approval behavior; serializable provider state; and Hyperlight-backed execution results
- move the CodeAct samples into the new package and update workspace/package metadata
- add unit coverage, a guarded real-sandbox integration test, and wire Hyperlight into the Python misc integration workflow

Closes: #5187

### Contribution Checklist


moonbox3:
<a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/README.md">Python Test Coverage</a><details><summary>Python Test Coverage Report • </summary><table><tr><th>File</th><th>Stmts</th><th>Miss</th><th>Cover</th><th>Missing</th></tr><tbody><tr><td colspan="5"><b>packages/core/agent_framework</b></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py">_tools.py</a></td><td>948</td><td>86</td><td>90%</td><td><a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L191-L192">191&ndash;192</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L365">365</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L367">367</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L380">380</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L405-L407">405&ndash;407</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L415">415</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L433">433</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L447">447</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L454">454</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L461">461</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L484">484</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L486">486</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L493">493</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L501">501</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L540">540</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L584">584</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L588">588</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L620-L622">620&ndash;622</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L630">630</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L675-L677">675&ndash;677</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L679">679</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L702">702</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L728">728</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L732">732</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L770-L772">770&ndash;772</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L776">776</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L798">798</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L910-L916">910&ndash;916</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L952">952</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L964">964</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L966">966</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L968">968</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L971-L974">971&ndash;974</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L995">995</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L999">999</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L1003">1003</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L1017-L1019">1017&ndash;1019</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L1360">1360</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L1382">1382</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L1469-L1475">1469&ndash;1475</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L1604">1604</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L1608">1608</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L1654">1654</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L1715-L1716">1715&ndash;1716</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L1831">1831</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L1851">1851</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L1853">1853</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L1909">1909</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L1972">1972</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L2144-L2145">2144&ndash;2145</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L2165">2165</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L2221-L2222">2221&ndash;2222</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L2360-L2361">2360&ndash;2361</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L2428">2428</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L2433">2433</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/core/agent_framework/_tools.py#L2440">2440</a></td></tr><tr><td colspan="5"><b>packages/hyperlight/agent_framework_hyperlight</b></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py">_execute_code_tool.py</a></td><td>477</td><td>85</td><td>82%</td><td><a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L64">64</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L100-L101">100&ndash;101</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L120">120</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L122">122</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L135">135</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L153">153</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L163">163</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L188">188</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L193">193</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L200">200</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L206">206</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L214">214</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L222-L224">222&ndash;224</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L226-L231">226&ndash;231</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L271">271</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L276">276</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L278">278</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L280">280</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L297-L298">297&ndash;298</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L307-L310">307&ndash;310</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L337-L340">337&ndash;340</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L346">346</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L348">348</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L358-L359">358&ndash;359</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L391-L392">391&ndash;392</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L395-L396">395&ndash;396</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L403">403</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L431">431</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L465-L466">465&ndash;466</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L472">472</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L480-L481">480&ndash;481</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L484">484</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L488">488</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L499-L505">499&ndash;505</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L557-L558">557&ndash;558</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L584">584</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L647">647</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L683">683</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L689-L691">689&ndash;691</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L720-L724">720&ndash;724</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L728-L729">728&ndash;729</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L734">734</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L751-L755">751&ndash;755</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L759-L760">759&ndash;760</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py#L816-L817">816&ndash;817</a></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_instructions.py">_instructions.py</a></td><td>44</td><td>5</td><td>88%</td><td><a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_instructions.py#L14">14</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_instructions.py#L33">33</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_instructions.py#L45-L46">45&ndash;46</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_instructions.py#L56">56</a></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_provider.py">_provider.py</a></td><td>42</td><td>11</td><td>73%</td><td><a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_provider.py#L53">53</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_provider.py#L57">57</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_provider.py#L65">65</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_provider.py#L69">69</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_provider.py#L73">73</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_provider.py#L77">77</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_provider.py#L81">81</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_provider.py#L85">85</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_provider.py#L89">89</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_provider.py#L93">93</a>, <a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_provider.py#L97">97</a></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/483ecfa3529a1c687a6019d6d653caf33b48e0e2/packages/hyperlight/agent_framework_hyperlight/_types.py">_types.py</a></td><td>13</td><td>0</td><td>100%</td><td>&nbsp;</td></tr><tr><td><b>TOTAL</b></td><td><b>28248</b></td><td><b>3300</b></td><td><b>88%</b></td><td>&nbsp;</td></tr></tbody></table></details>

## Python Unit Test Overview
| 5621 | 30 :zzz: | 0 :x: | 0 :fire: | 1m 31s :stopwatch: |

python/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py
```diff
@@ -0,0 +1,700 @@
...
+
+        workspace_signature = _path_tree_signature(workspace_root) if workspace_root is not None else ()
+        normalized_mounts = tuple(
+            _NormalizedFileMount(
+                host_path=mount.host_path,
+                mount_path=mount.mount_path,
+                path_signature=_path_tree_signature(mount.host_path),
+            )
+            for mount in stored_mounts
+        )
```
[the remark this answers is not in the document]
eavanvalkenburg: Thanks — I looked at this one, but I’m leaving the current behavior in place for now. The tree signatures are part of the cache key specifically so sandbox snapshots are invalidated when workspace or mount contents change between runs, and I don’t want to trade that correctness for a weaker heuristic in this PR. If this turns into a real hotspot, I’d rather address it in a follow-up with an explicit cache/change-detection design.

python/packages/core/agent_framework/_agents.py
```diff
@@ -1191,11 +1191,14 @@ async def _prepare_run_context(
...
         )
+        provider_options = dict(opts)
+        if tools_ is not None:
+            provider_options["tools"] = tools_
 
         session_context, chat_options = await self._prepare_session_and_messages(
             session=active_session,
             input_messages=input_messages,
-            options=opts,
+            options=provider_options,
```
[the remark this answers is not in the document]
eavanvalkenburg: I checked this merge path and I’m not changing it in this PR. The explicit `tools` parameter is the public per-run tool input, and copying it into `SessionContext.options["tools"]` is intentional so context providers can inspect and override the runtime tool set. If callers also stuff `tools` into generic `options`, the named `tools` argument winning is the behavior I want here.

docs/decisions/0024-codeact-integration.md
```diff
@@ -0,0 +1,213 @@
...
+The design does not require a public abstract `CodeActContextProvider` base class, but it does require a stable implementation contract for concrete providers.
+
+- Concrete providers should expose a standard capability surface at construction time, with SDK-appropriate naming for:
+  - approval mode
+  - filesystem mode
+  - workspace root
+  - file mounts
+  - network mode
+  - allowed outbound domains
+  - allowed HTTP methods or an equivalent outbound policy surface
```
moonbox3: is network mode = disabled (or equivalent) a first-class option, so a backend like Monty with no outbound network at all doesn't have to stub HTTP-shaped fields?

docs/decisions/0024-codeact-integration.md
```diff
@@ -0,0 +1,213 @@
...
+- If every provider-owned CodeAct tool in that registry has `approval_mode="never_require"`, `execute_code` is treated as `never_require`. If any provider-owned CodeAct tool in that registry has `approval_mode="always_require"`, `execute_code` is treated as `always_require`, even if the generated code may not end up calling that tool.
+- Approval is granted before `execute_code` starts, and provider-owned tool calls made from inside that execution run under the same approval.
+- Direct-only agent tools do not affect the approval of `execute_code`; only the provider-owned CodeAct tool registry participates in that calculation.
+- Configuring filesystem and network capability state on the provider, including adding file mounts or outbound network allow-list entries, is itself the approval for those capabilities in the initial model.
+- Each `execute_code` invocation must start from a clean execution state. Exact caching, snapshot, and environment-reuse strategies are implementation details defined in the language-specific specs.
```
moonbox3: does the ADR want to lock in a cross-SDK invariant here ("variables don't persist across execute_code calls, but workspace files do") so Python and .NET can't diverge on something load-bearing for the long-running-loop rationale on line 102?

docs/decisions/0024-codeact-integration.md
```diff
@@ -0,0 +1,213 @@
...
+- At a high level, CodeAct is exposed through backend-specific context providers that contribute an `execute_code` tool, own the CodeAct-specific tool registry, and carry backend capability configuration such as filesystem and network access.
+- The initial approval model is bundled approval for `execute_code`, using the same `approval_mode="always_require" | "never_require"` vocabulary as regular tools.
+- The CodeAct provider exposes a default `approval_mode` for `execute_code`. If the provider default is `never_require`, the effective approval for `execute_code` is derived from the provider-owned CodeAct tool registry captured for the run.
+- If every provider-owned CodeAct tool in that registry has `approval_mode="never_require"`, `execute_code` is treated as `never_require`. If any provider-owned CodeAct tool in that registry has `approval_mode="always_require"`, `execute_code` is treated as `always_require`, even if the generated code may not end up calling that tool.
```
moonbox3, line 162: This is safe but silently changes behavior, I feel like: adding one sensitive tool to the provider flips approval for every execute_code invocation, even ones that don't touch it. Should we call this out explicitly and document the mitigation pattern (split providers, or keep the sensitive tool direct-only and out of the CodeAct registry)? New users are likely to trip on this, IMO.

docs/decisions/0024-codeact-integration.md
```diff
@@ -0,0 +1,213 @@
...
+- CodeAct support should ship as an optional package in each SDK rather than as part of the core package, so users who do not need CodeAct do not take on its installation and dependency footprint.
+- There is no separate runtime setup object in the chosen design. Concrete providers manage their provider-owned CodeAct tool registry, file mounts, and outbound network allow-list configuration directly through CRUD-style methods on the provider itself.
+- At a high level, CodeAct is exposed through backend-specific context providers that contribute an `execute_code` tool, own the CodeAct-specific tool registry, and carry backend capability configuration such as filesystem and network access.
+- The initial approval model is bundled approval for `execute_code`, using the same `approval_mode="always_require" | "never_require"` vocabulary as regular tools.
+- The CodeAct provider exposes a default `approval_mode` for `execute_code`. If the provider default is `never_require`, the effective approval for `execute_code` is derived from the provider-owned CodeAct tool registry captured for the run.
```
moonbox3: What is the behavior when the provider default is always_require? Should we state it explicitly, for example, execute_code is always always_require regardless of tool registry content? So implementors don't have to infer it?

docs/decisions/0024-codeact-integration.md
```diff
@@ -0,0 +1,213 @@
...
+
+- Concrete providers should expose a standard capability surface at construction time, with SDK-appropriate naming for:
+  - approval mode
+  - filesystem mode
+  - workspace root
+  - file mounts
+  - network mode
+  - allowed outbound domains
+  - allowed HTTP methods or an equivalent outbound policy surface
+- Concrete providers should expose direct CRUD-style methods for managing the provider-owned CodeAct tool registry, file mounts, and outbound network allow-list configuration, rather than requiring callers to construct a separate runtime setup object.
```
moonbox3, line 198: Line 26 says the design "must not rely on mutating shared agent configuration during invocation," and line 182 says the tool registry is snapshotted per run. But providers expose live CRUD methods. What is the expected behavior if a caller invokes add_tool (or adds a file mount) while a run is in flight — does it affect the current run, the next run, or is it a programming error? Is mutation thread-safe across concurrent runs sharing one provider instance?

docs/decisions/0024-codeact-integration.md
```diff
@@ -0,0 +1,213 @@
...
+deciders: TBD
+consulted:
+informed:
+---
+
+# CodeAct integration through backend-specific context providers and an `execute_code` tool
+
+## Context and Problem Statement
+
+We need an architecture design that supports CodeAct in both Python and .NET. This is a necessary capability for the current generation of long-running agents, which need to plan, iterate, transform tool outputs, and execute bounded code inside a controlled runtime instead of pushing every intermediate step back through the model. The design should preserve the same behavioral contract across SDKs, but it does not need to use the same internal extension point in each runtime. We also want to standardize on Hyperlight as the initial backend, using the existing Python package and an anticipated .NET binding package once it is available.
```
moonbox3: Can we add a brief threat-model statement at ADR level? Specifically: what does "bounded" and "controlled" assume about Hyperlight's isolation guarantees, what is the framework's responsibility vs. the backend's on sandbox escape, and what should a reader understand about the trust boundary between model-generated code and the host process?

Or will a separate threat model for this follow?

docs/decisions/0024-codeact-integration.md
```diff
@@ -0,0 +1,213 @@
...
+- Concrete providers should expose direct CRUD-style methods for managing the provider-owned CodeAct tool registry, file mounts, and outbound network allow-list configuration, rather than requiring callers to construct a separate runtime setup object.
+- Concrete providers should implement their host SDK's provider lifecycle hooks to:
+  - build CodeAct instructions,
+  - add `execute_code`,
+  - snapshot the effective CodeAct tool registry and capability settings for the run,
+  - compute the effective approval requirement for `execute_code`,
+  - configure file access and network access for the backend,
+  - prepare or restore execution state,
+  - execute code,
+  - and translate backend output into framework-native content.
```
moonbox3, line 207: What are the expected semantics for timeouts, OOM, backend crash, and partial-output recovery? Should these be accounted for? Even one sentence per failure mode in the Design summary would prevent Python and .NET from diverging on error definition/handling.

docs/decisions/0024-codeact-integration.md
```diff
@@ -0,0 +1,213 @@
...
+
+Throughout this ADR, **CodeAct** is the primary term. **Code mode** and **programmatic tool calling** refer to the same capability. This ADR uses **CodeAct** consistently.
+
+The core design question is: **where should CodeAct integrate into the agent pipeline so that both SDKs can offer the same functionality without invasive changes to their core function-calling loops?**
+
+## Decision Drivers
+
+- CodeAct must shape the model-facing surface before model invocation, not only after the model has already chosen tools.
+- The design should let users control which tools are available through CodeAct and which remain regular tools only.
+- The design must preserve existing session, approval, telemetry, and tool invocation behavior as much as possible.
```
moonbox3, line 32: Given that CodeAct runs arbitrary code, telemetry is load-bearing. Can we spell out what execute_code is expected to emit (start/end events, provider-owned tool calls made from inside it, sandbox resource usage), how it nests within the agent run trace, and how cross-SDK telemetry stays aligned? "Preserve existing" doesn't answer any of these for me.

docs/decisions/0024-codeact-integration.md
```diff
@@ -0,0 +1,213 @@
...
+- The design must fit naturally into the extension points that already exist in each SDK.
+- The design must be safe for concurrent runs and must not rely on mutating shared agent configuration during invocation.
+- The chosen structure should allow multiple backend-specific providers to fit under the same conceptual design over time, even though Hyperlight is the initial target.
+- The abstraction should not assume that every backend is a VM-style sandbox; alternative execution models such as Pydantic's Monty should also fit.
+- The design should allow `execute_code` to be reused both as a tool-enabled CodeAct runtime and as a standard code interpreter tool implementation.
+- The design should remain open to alternative language/runtime modes, such as JavaScript on Hyperlight, rather than baking the abstraction to Python only.
+- The design should provide a portable way to configure sandbox capabilities such as file access and network access, including allow-listed outbound domains.
+- Using CodeAct should be optional, and installing its runtime or backend dependencies should also be optional.
```
moonbox3, line 41: This PR modifies _agents.py, _sessions.py, and _tools.py in core. If CodeAct is an optional bolt-on, why are core files touched? Can we add a short subsection listing the minimal hooks exposed in core, confirming they are backward-compatible, and explaining why they can't live entirely in the optional package?

docs/decisions/0024-codeact-integration.md
```diff
@@ -0,0 +1,213 @@
...
+This follows the spirit of the current Python tool approval flow, where `FunctionTool` uses `approval_mode="always_require" | "never_require"` and the auto-invocation loop escalates the whole batch when any called tool requires approval.
+
+### Design summary
+
+We standardize the **public concept** of CodeAct across SDKs while allowing each SDK to use the extension point that fits it best.
+
+- Python uses a `ContextProvider`.
+- .NET uses an `AIContextProvider`.
+- The term **CodeAct context provider** is used throughout this ADR as a design concept, not as a required public base type. Public SDK APIs should prefer concrete backend-specific types such as `HyperlightCodeActProvider` rather than a public abstract `CodeActContextProvider` or a public `CodeActExecutor` parameter.
+- CodeAct support should ship as an optional package in each SDK rather than as part of the core package, so users who do not need CodeAct do not take on its installation and dependency footprint.
```
moonbox3: could the ADR list (or link to) the minimal core hooks the optional package depends on? Otherwise "optional package" on line 148 reads stronger than the PR actually delivers.

moonbox3 (commented):
This is a nice and complete ADR - well done. A lot here to unpack so doing a first pass with some questions.

docs/decisions/0024-codeact-integration.md
```diff
@@ -0,0 +1,227 @@
...
+deciders: TBD
+consulted:
+informed:
+---
+
+# CodeAct integration through backend-specific context providers and an `execute_code` tool
+
+## Context and Problem Statement
+
+We need an architecture design that supports CodeAct in both Python and .NET. This is a necessary capability for the current generation of long-running agents, which need to plan, iterate, transform tool outputs, and execute bounded code inside a controlled runtime instead of pushing every intermediate step back through the model. The design should preserve the same behavioral contract across SDKs, but it does not need to use the same internal extension point in each runtime. We also want to standardize on Hyperlight as the initial backend, using the existing Python package and an anticipated .NET binding package once it is available.
```
TaoChenOSU: > instead of pushing every intermediate step back through the model

Could we elaborate on what this means? 

Also an introduction of CodeAct will greatly help readers of this doc.

docs/decisions/0024-codeact-integration.md
```diff
@@ -0,0 +1,227 @@
...
+- Good, because convenience builder/decorator helpers can still be added later on top of the provider model without changing the core design.
+- Good, because backend-specific runtime logic can stay inside concrete provider implementations or internal helpers instead of being forced into a lowest-common-denominator public abstraction.
+- Good, because the same provider structure can support either an all-or-nothing tool surface or a mixed side-by-side tool surface.
+- Good, because users can keep some tools direct-only while allowing other tools to be used from inside CodeAct.
+- Good, because a provider-owned CodeAct tool registry avoids mutating or inferring the agent's direct tool surface and can work consistently in both SDKs.
+- Good, because the same conceptual design can remain open to `HyperlightCodeActProvider`, a future `MontyCodeActProvider`, and other backend-specific providers over time.
+- Good, because `execute_code` can evolve into multiple backend-specific runtime modes rather than being hard-wired to one Python-plus-tools mode.
+- Bad, because it is a bolt-on, which might make it less runtime efficient.
```
TaoChenOSU: > make it less runtime efficient
Why will a bolt-on make it less efficient?

docs/features/code_act/python-implementation.md
```diff
@@ -0,0 +1,359 @@
...
+
+Success Metric:
+- Python samples exist for both a tool-enabled CodeAct mode and a standard interpreter mode.
+
+Implementation-free outcome:
+- A Python developer can attach a backend-specific CodeAct provider, choose which tools are available inside CodeAct, and configure execution capabilities without rewriting the function invocation loop.
+
+## What is the problem being solved?
+
+- Today, the easiest way to prototype CodeAct is to infer or reshape the agent's direct tool surface, which is fragile and hard to reason about.
```
TaoChenOSU, line 28: Should this be part of the ADR instead?

docs/features/code_act/python-implementation.md
```diff
@@ -0,0 +1,359 @@
...
+
+Optional lifecycle hook:
+- `after_run(*, agent, session, context, state) -> None`
+
+`before_run(...)` is responsible for:
+- snapshotting the current CodeAct-managed tool registry and capability settings for the run,
+- computing the effective approval requirement for `execute_code` from the provider default and the snapshotted tool registry,
+- adding a short CodeAct guidance block,
+- adding `execute_code` to the run through `SessionContext.extend_tools(...)`,
+- and wiring any backend-specific execution state needed for the run.
```
TaoChenOSU, line 232: Are these required for each run? Can these be done once at construction time which will inject the available tools to the agent's tool list?

docs/features/code_act/python-implementation.md
```diff
@@ -0,0 +1,359 @@
...
+    workspace_root="./workdir",
+    allowed_domains=[("api.github.com", "GET")],
+)
+codeact.add_tools([lookup_user])
+
+agent = Agent(
+    client=client,
+    name="assistant",
+    tools=[send_email],  # direct-only tool
+    context_providers=[codeact],
```
TaoChenOSU, line 347: Looking at this, a question that users may have is that is the difference between tools and contexts? 

Just an idea: is it possible to do the following
```python
agent = Agent(
    client=client,
    name="assistant",
    tools=[send_email, *codeact.get_tools()],
)
```
where the returned tools have a reference to the provider so that that can access the file mounts, allowed domains, etc?

docs/features/code_act/python-implementation.md
```diff
@@ -0,0 +1,359 @@
...
+
+```python
+code_interpreter = HyperlightCodeActProvider(
+    workspace_root="./data",
+)
+
+agent = Agent(
+    client=client,
+    name="interpreter",
+    context_providers=[code_interpreter],
```
TaoChenOSU: Adding to the previous comment, `code_interpreter` is reusing an existing concept whose usage is very different.

python/packages/hyperlight/samples/README.md
```diff
@@ -0,0 +1,16 @@
+# Hyperlight CodeAct samples
+
+These samples demonstrate the alpha `agent-framework-hyperlight` package.
+
+- `codeact_context_provider.py` shows the provider-owned CodeAct model where the
+  agent only sees `execute_code` and sandbox tools are owned by
+  `HyperlightCodeActProvider`.
+- `codeact_tool.py` shows the standalone `HyperlightExecuteCodeTool` surface
+  where `execute_code` is added directly to the agent tool list.
```
TaoChenOSU, line 35: A short paragraph on when to use what will be helpful for customers.

docs/decisions/0024-codeact-integration.md
eavanvalkenburg: Thanks — addressed both points. Added an **Introduction** section to the ADR that defines CodeAct for readers unfamiliar with the concept, and expanded the problem statement to give concrete examples (filtering results, computing derived values, chaining tool calls with conditionals) instead of the vague 'intermediate steps' phrasing.

docs/decisions/0024-codeact-integration.md
eavanvalkenburg: Fair point — the original wording was vague. Replaced with a specific description: the provider indirection adds per-run snapshotting, hook dispatch, and instruction-building overhead, but this is negligible relative to model inference latency and sandbox startup cost.

docs/features/code_act/python-implementation.md
eavanvalkenburg: Agreed — renamed the variable from \code_interpreter\ to \codeact\ in the E2E sample to avoid confusion with OpenAI's Code Interpreter concept.

python/packages/hyperlight/samples/README.md
eavanvalkenburg: Added a **When to use which pattern** section to the README covering all three patterns: provider (dynamic config, concurrent runs), manual static wiring (fixed config, simpler setup), and standalone tool (quickest prototyping).

docs/features/code_act/python-implementation.md
eavanvalkenburg: The split is intentional: the ADR covers the cross-SDK problem statement and decision rationale, while this doc narrows it to Python-specific design concerns (concrete ContextProvider API, Hyperlight-specific details, Python public type surface). Added a brief note at the top of this section clarifying the relationship and pointing readers to the ADR for the shared problem statement.

docs/features/code_act/python-implementation.md
eavanvalkenburg: Per-run is necessary because the provider supports CRUD mutations between runs (add/remove tools, file mounts, domains), concurrent runs need independent snapshots for determinism, and the effective approval and instructions depend on the tool registry state captured at run start. That said, when the configuration is fixed for the agent's lifetime, the per-run lifecycle can be skipped entirely — added a new **manual static wiring** pattern (see \codeact_manual_wiring.py\) that builds the tool and instructions once and passes them directly to the Agent constructor via \	ools=\. Also added a clarifying note in this section explaining the tradeoff.

docs/features/code_act/python-implementation.md
eavanvalkenburg: Good suggestion — added a new manual static wiring sample (codeact_manual_wiring.py) and a matching E2E code block in the implementation doc that shows exactly this pattern: build HyperlightExecuteCodeTool with tools, call build_instructions() once, then pass both to the Agent constructor via tools=[send_email, execute_code].

The context_providers pattern is still the recommended default when dynamic tool management or concurrent runs are needed (the provider handles instruction rebuilding, snapshotting, and approval recomputation per-run). But when the config is static, the simpler tools= pattern works well and avoids the provider lifecycle entirely.

python/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py
```diff
@@ -0,0 +1,815 @@
...
+
+    def execute(self, *, config: _RunConfig, code: str) -> list[Content]:
+        cache_key = config.cache_key()
+        with self._entries_lock:
+            entry = self._entries.get(cache_key)
+            if entry is None:
+                entry = self._create_entry(config)
+                self._entries[cache_key] = entry
+
+        with entry.lock:
```
moonbox3, line 528: I see that _SandboxRegistry caches a `_SandboxEntry` keyed by `config.cache_key()` and reuses the same `output_dir` `TemporaryDirectory` for every subsequent invocation. `entry.sandbox.restore(entry.snapshot)` only restores in-sandbox state, the host directory mapped at `/output` is never cleared. Then `_parse_output_files` rglobs that host directory. After the first call writes, say, `report.txt`, won't every subsequent call (even ones whose code produces nothing) re-emit that stale file as a data Content on the result? 

Should we either clear `entry.output_dir` before each `entry.sandbox.run(code=code)` here, or scope `output_dir` to a single invocation rather than the cache entry's lifetime?

python/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py
```diff
@@ -0,0 +1,815 @@
...
+        if len(values) == 1:
+            return values[0]
+        return values
+
+    return _callback
+
+
+class _SandboxRegistry:
+    def __init__(self) -> None:
+        self._entries: dict[tuple[Any, ...], _SandboxEntry] = {}
```
moonbox3, line 510: This registry persists entries for the lifetime of the tool/process and shares the same in-memory sandbox plus the same host `output_dir` across every `execute()` call whose `cache_key()` matches. Combined with there being no per-call cleanup of `output_dir` (see other comment), doesn't that mean two unrelated agent runs, or two concurrent users sharing one provider instance, can see each other's `/output` artifacts in their tool result? Should the cache scope be tightened so each invocation gets a fresh `output_dir` even when the sandbox itself is reused?

python/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py
```diff
@@ -0,0 +1,815 @@
...
+        )
+    )
+    return [Content.from_code_interpreter_tool_result(outputs=outputs, raw_representation=result)]
+
+
+def _make_sandbox_callback(tool_obj: FunctionTool) -> Callable[..., Any]:
+    sandbox_tool = copy.copy(tool_obj)
+    sandbox_tool.result_parser = _passthrough_result_parser
+
+    async def _callback(**kwargs: Any) -> Any:
```
moonbox3: _make_sandbox_callback returns an `async def _callback` that gets handed to `sandbox.register_tool(...)`, but the real `hyperlight_sandbox` Python guest invokes registered callbacks synchronously from inside the sandbox when the model writes `call_tool(...)`.

The `_FakeSandbox` in the test suite works around this by spawning a thread and calling `asyncio.run`, but does the actual Hyperlight binding await coroutines for us?

If not, every `call_tool(...)` from sandboxed code would receive a coroutine object instead of the tool's return value, and the whole CodeAct flow falls over end-to-end.

Is there an integration test exercising the real backend with an async tool to confirm this round-trips correctly?

python/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py
```diff
@@ -0,0 +1,815 @@
...
+            module_path=self._module_path,
+            approval_mode=approval_mode,
+            tools=managed_tools,
+            workspace_root=workspace_root,
+            workspace_signature=workspace_signature,
+            file_mounts=normalized_mounts,
+            allowed_domains=allowed_domains,
+        )
+
+    def _run_code(self, *, code: str) -> list[Content]:
```
moonbox3: `_run_code` is synchronous and ultimately drives `entry.sandbox.run(code=code)` directly on the asyncio event loop via the FunctionTool path.

For the long-running CodeAct loops this package is motivated by, won’t every `execute_code` invocation block the entire agent loop, and any concurrent agents sharing that loop, for the full duration of the sandbox runtime?

Should this be `async def`, with an `asyncio.to_thread(...)` offload around `self._registry.execute(...)`, so the event loop remains responsive while sandboxed code executes?

python/packages/core/agent_framework/_agents.py
```diff
@@ -1209,7 +1212,7 @@ async def _prepare_run_context(
         mcp_duplicate_message = "Tool names must be unique. Consider setting `tool_name_prefix` on the MCPTool."
 
         # Normalize tools
-        normalized_tools = normalize_tools(tools_)
+        normalized_tools = normalize_tools(session_context.options.get("tools", tools_))
```
moonbox3: Before this PR, `normalized_tools` was always derived from the `tools_` argument.

After this change, anything sitting in `options["tools"]` overrides `tools_`. Because `provider_options = dict(opts)` is only a shallow copy of the caller’s `opts`, a caller-supplied `options={"tools": [...]}` now also gets promoted to the runtime tool list whenever `tools_` is `None`, where previously it wasn't.

Is the intent that providers are the only writers of `options["tools"]`, or is caller-supplied `options["tools"]` also meant to be promoted now?

Is this behavior change intended? Could we get an explicit unit test pinning down both the provider-write path and the caller-supplied-options path?

python/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py
eavanvalkenburg: Fixed — added `_clear_directory(entry.output_dir)` call before each `entry.sandbox.run(code=code)` inside `_SandboxRegistry.execute()`. The helper removes all files/dirs in the host output directory (tolerant of symlinks, already-removed files, and permission errors) so stale artifacts from prior invocations never leak into subsequent results.

Also added a real-sandbox integration test `test_output_dir_cleared_between_invocations` that writes a file in the first call and asserts it does not appear in the second call's outputs.

Commit: 5beb891

python/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py
eavanvalkenburg: Good observation. The `entry.lock` serializes concurrent calls with the same `cache_key()`, so they never race. With the `_clear_directory()` fix (comment 23), each invocation starts with a clean output_dir, so concurrent users sharing a provider with identical config won't see each other's artifacts.

Added a docstring on `_SandboxRegistry.execute()` documenting the serialization guarantee and recommending distinct provider instances/configs for true parallel execution.

Commit: 5beb891

python/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py
eavanvalkenburg: Confirmed — this was indeed broken with the real sandbox. The Wasm FFI invokes callbacks synchronously; the `async def` callback returned a coroutine object instead of the actual result.

Fixed: `_make_sandbox_callback` now returns a **synchronous** function that bridges to the async `FunctionTool.invoke()` via `threading.Thread` + `asyncio.run()`. This works regardless of whether the callback runs on the event loop thread or a worker thread. `await call_tool(...)` is not possible in the guest — `call_tool` is a synchronous Wasm FFI built-in with no async variant.

Added `test_sandbox_async_callback_round_trips_with_real_sandbox` integration test that registers a `_make_sandbox_callback`-wrapped tool and exercises it via `call_tool(...)` in the real sandbox.

Commit: 5beb891

python/packages/hyperlight/agent_framework_hyperlight/_execute_code_tool.py
eavanvalkenburg: Fixed — `_run_code` is now `async def` and offloads the entire `self._registry.execute(...)` to a worker thread via `asyncio.to_thread()`. This keeps the event loop responsive during sandbox execution and also provides a clean thread context for the callback's `asyncio.run()` bridge.

Added `test_run_code_does_not_block_event_loop` integration test that monkeypatches the registry with a blocking stub, starts the tool invocation as a task, and proves a concurrent coroutine completes while the worker thread is still blocked.

Commit: 5beb891

python/packages/core/agent_framework/_agents.py
eavanvalkenburg: Reverted both the `_agents.py` code change and the `SessionContext.options` docstring change. The `HyperlightCodeActProvider` uses `context.extend_tools()` to inject the `execute_code` tool — this is the correct provider pattern and doesn't require any changes to core agent tool resolution.

The `options['tools']` injection was unnecessary for this PR and changed the documented contract of `SessionContext.options` from read-only to writable, introducing unintended behavioral side-effects for caller-supplied `options={'tools': [...]}`. Also removed the two associated tests from `test_agents.py` that were added to cover the reverted behavior.

Commit: 5beb891

moonbox3 (approved):

lokitoth (approved):

TaoChenOSU (approved):

PR #5199, opened 2026-04-10
Python: pass AG-UI state through AgentSession

### LINKED ISSUE #5197: Python: [Bug]: The current implementation of AG-UI fails to pass the state to AgentSession.
### Description

The current implementation of AG-UI fails to pass the state to AgentSession.
Since MAF's ContextProvider relies on state to transmit valid information, this prevents additional information from being passed to the ContextProvider.
For example: I want to pass information such as agent_id, user_id, and tenant_id to the HistoryProvider (which inherits from ContextProvider).
However, these values cannot be effectively passed to the HistoryProvider when the frontend uses the AG-UI protocol.
Alternatively, you may have a robust mechanism for passing information to the ContextProvider but lack corresponding examples.
Based on my review of the source code, it seems that this requirement cannot be fulfilled at present.

### Code Sample

```markdown

```

```markdown

```

### Package Versions

agent-framework-ag-ui: 1.0.0b260402

### Python Version

python:3.13

### Additional Context

_No response_

Bortlesboat:
### Motivation and Context

Fixes #5197.

AG-UI already carried request state in metadata and model-facing state injection, but it never copied that state into `AgentSession.state`. That left `ContextProvider` implementations with an empty session state when runs came through the AG-UI path.

### Description

- copy AG-UI request `state` into the created `AgentSession`
- add a regression test that captures the session passed to `agent.run(...)` and asserts the request state is available there

### Contribution Checklist

- [ ] **Is this a breaking change?** If yes, add [BREAKING] prefix to the title of the PR.

### DIFF
--- a/python/packages/ag-ui/agent_framework_ag_ui/_agent_run.py
+++ b/python/packages/ag-ui/agent_framework_ag_ui/_agent_run.py
@@ -784,6 +784,8 @@ async def run_agent_stream(
         session = AgentSession(service_session_id=supplied_thread_id)
     else:
         session = AgentSession()
+    if flow.current_state:
+        session.state.update(flow.current_state)
 
     # Inject metadata for AG-UI orchestration (Feature #2: Azure-safe truncation)
     base_metadata: dict[str, Any] = {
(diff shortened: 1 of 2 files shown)


eavanvalkenburg:
Closing, not sure if this is still relevant, feel free to reopen if so.

Issue #5165, opened 2026-04-08
Python: [Bug]: BedrockChatClient sends `toolConfig.toolChoice` without `toolConfig.tools` when agent has no tools


gourab-aiplanet:
### Description

When using `BedrockChatClient` with an `Agent` that has no tools configured, the client still sends `toolConfig.toolChoice` to the Bedrock API. This causes a validation error because AWS Bedrock requires `toolConfig.tools` to be present whenever `toolConfig.toolChoice` is specified.

### Code Sample

```markdown
import asyncio
from agent_framework import Agent
from agent_framework_bedrock import BedrockChatClient

async def main():
    client = BedrockChatClient(
        access_key="...",
        secret_key="...",
        region="eu-west-1",
        model="anthropic.claude-sonnet-4-6"
    )

    # Agent with no tools
    agent = Agent(
        client=client,
        name="TestAgent",
        instructions="You are a helpful assistant",
        tools=None,  # No tools
        default_options=None,  # No tool options
    )

    # This fails with ParamValidationError
    result = await agent.run("What is AI?")
    print(result)

asyncio.run(main())
```

### Error Messages / Stack Traces

```markdown
botocore.exceptions.ParamValidationError: Parameter validation failed:
Missing required parameter in toolConfig: "tools"
```

### Package Versions

agent-framework-bedrock version: 1.0.0b260402

### Python Version

Python version: 3.10+

### Expected behavior

When no tools are provided to the agent, the `BedrockChatClient` should NOT include `toolConfig` in the API request at all. This is similar to how PR #4535 fixed the `tool_choice="none"` case by omitting `toolConfig` entirely.

### Actual behavior

The client sends a request with `toolConfig.toolChoice` (likely defaulting to `{"auto": {}}`) but without `toolConfig.tools`, causing AWS Bedrock to reject the request.

LEDazzio01:
Happy to take this — it looks like a guard is needed to omit `toolConfig.toolChoice` when no tools are configured. I worked on a similar Bedrock edge case before (PR #4529).

PR #4535, opened 2026-03-06
Python: Fix BedrockChatClient sending invalid toolChoice "none" to Bedrock API

LEDazzio01:
### Motivation and Context

When `FunctionInvocationLayer` exhausts its max iterations or hits the error threshold, it sets `tool_choice="none"` for a final "wrap-up" LLM call (to get a plain-text summary). `BedrockChatClient._prepare_options()` then maps this to `toolConfig.toolChoice: {"none": {}}`, but the AWS Bedrock `Converse` / `ConverseStream` API only accepts `auto`, `any`, or `tool` as valid `toolChoice` keys — causing a `botocore.exceptions.ParamValidationError`.

Fixes #4529

### Description

Split the `case "auto" | "none"` match branch in `_prepare_options`:

- **`"none"`**: Set `tool_config = None` so `toolConfig` is omitted entirely from the Bedrock request. This matches the semantic intent ("don't use tools on this call") and is consistent with how the OpenAI client handles `tool_choice="none"`.
- **`"auto"`**: Unchanged — emits `{"auto": {}}` as before.

Also added `tool_config = tool_config or {}` inside each individual branch (`"auto"`, `"required"`) instead of before the `match` statement, so `"none"` can cleanly set `tool_config = None` without it being overwritten.

### Tests

Added 3 new test cases:

| Test | Verifies |
|------|----------|
| `test_prepare_options_tool_choice_none_omits_tool_config` | `toolConfig` absent from request when `tool_choice="none"` (even with tools provided) |
| `test_prepare_options_tool_choice_auto_includes_tool_config` | `toolChoice: {"auto": {}}` present when `tool_choice="auto"` |
| `test_prepare_options_tool_choice_required_includes_any` | `toolChoice: {"any": {}}` present when `tool_choice="required"` |

### Contribution Checklist

- [x] The code builds clean without any errors or warnings
- [x] The PR follows the [Contribution Guidelines](https://github.com/microsoft/agent-framework/blob/main/CONTRIBUTING.md)
- [x] All unit tests pass, and I have added new tests where possible
- [ ] **Is this a breaking change?** If yes, add "[BREAKING]" prefix to the title of the PR.

PR #5289, opened 2026-04-15
Python: fix: skip toolChoice when no tools are configured in BedrockChatClient

LEDazzio01:
## Fixes

**#5165** — `BedrockChatClient` sends `toolConfig.toolChoice` without `toolConfig.tools` when agent has no tools configured, causing a `ParamValidationError` from AWS Bedrock.

## Root Cause

In `_prepare_options`, when `tool_choice` is `"auto"` or `"required"` but no tools are provided:

```python
tool_config = self._prepare_tools(options.get("tools"))  # Returns None
# ...
case "auto":
    tool_config = tool_config or {}       # Creates empty dict!
    tool_config["toolChoice"] = {"auto": {}}  # Sets toolChoice with NO tools
```

This produces `{"toolChoice": {"auto": {}}}` — which Bedrock rejects because `tools` is required when `toolChoice` is present.

## Fix

Guard both `auto` and `required` branches to only set `toolChoice` when `tool_config is not None` (i.e., when `_prepare_tools` found actual tools). This matches the existing behavior of the `none` case (added in #4535) which correctly omits `toolConfig` entirely.

## Tests

Added 3 new tests:
- `test_prepare_options_tool_choice_auto_no_tools_omits_tool_config`
- `test_prepare_options_tool_choice_required_no_tools_omits_tool_config`
- `test_prepare_options_no_tools_no_tool_choice_omits_tool_config`

### DIFF
--- a/python/packages/bedrock/agent_framework_bedrock/_chat_client.py
+++ b/python/packages/bedrock/agent_framework_bedrock/_chat_client.py
@@ -411,14 +411,16 @@ class BedrockChatClient(
                     # Omit toolConfig entirely so the model won't attempt tool calls.
                     tool_config = None
                 case "auto":
-                    tool_config = tool_config or {}
-                    tool_config["toolChoice"] = {"auto": {}}
+                    # Only set toolChoice when tools are present; Bedrock rejects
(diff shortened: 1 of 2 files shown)


LEDazzio01:

Issue #5237, opened 2026-04-14
Python: [Bug]: Colliding Ids are assigned to message history creating issues when using compaction


jorge-wonolo:
### Description

`_ensure_message_ids(messages)` in `agent_framework/_compaction.py` fills each missing `Message.message_id` with `f"msg_{index}"`, where `index` is the position **in the list it receives**. `annotate_message_groups` calls it indirectly via `group_messages(messages[start_index:])` on a **slice** of the full list — so successive incremental calls collide with ids already assigned by earlier calls. The framework ends up with multiple distinct messages sharing the same `message_id`.

In the sample provided this was observed:
```
  0: role=user      id='msg_0'
  1: role=assistant id='msg_1'
  2: role=user      id='msg_1'  
  3: role=assistant id='msg_2'
  4: role=user      id='msg_1'
  5: role=assistant id='msg_2'
  6: role=user      id='msg_1'
  7: role=assistant id='msg_2'
```

Downstream consumers assume `message_id` is unique. The most visible failure today is in `SummarizationStrategy`: `_group_id_for` derives `group_id` from `message_id`, so colliding ids collapse distinct groups into one.

### Code Sample

```markdown
import asyncio
import os

from agent_framework import (
    Agent,
    CompactionProvider,
    InMemoryHistoryProvider,
    SlidingWindowStrategy,
)
from agent_framework.openai import OpenAIChatCompletionClient


async def main() -> None:
    agent = Agent(
        client=OpenAIChatCompletionClient(
            model="gpt-4o-mini", api_key=os.environ["OPENAI_API_KEY"]
        ),
        name="Repro",
        instructions="Reply in <= 4 words.",
        context_providers=[
            InMemoryHistoryProvider(),
            CompactionProvider(
                after_strategy=SlidingWindowStrategy(keep_last_groups=1000),
            ),
        ],
    )
    session = agent.create_session(session_id="repro")

    for text in ["Say one.", "Say two.", "Say three.", "Say four."]:
        await agent.run(text, session=session)

    for i, m in enumerate(session.state["in_memory"]["messages"]):
        print(f"  {i}: role={m.role:9} id={m.message_id!r}")
```

### Error Messages / Stack Traces

```markdown

```

### Package Versions

agent-core-framework: 1.0.1

### Python Version

Python 3.13

### Additional Context

As a workaround I have been adding unique ids to messages outside of the agent loop

LEDazzio01:
[...]

PR #5290, opened 2026-04-15
Python: fix: prevent colliding message_id values in incremental compaction

LEDazzio01:
## Fixes

**#5237** — `_ensure_message_ids` assigns colliding IDs when called incrementally via `annotate_message_groups`, breaking `SummarizationStrategy` and other group-ID-dependent consumers.

## Root Cause

`_ensure_message_ids` assigns `f"msg_{index}"` where `index` is the position within the list it receives. However, `annotate_message_groups` calls `group_messages(messages[start_index:])` on a **slice**, so the slice always starts at index 0. Successive incremental calls produce `msg_0`, `msg_1`, etc., colliding with IDs already assigned by earlier calls.

The issue reporter's reproduction shows:
```
  0: role=user      id='msg_0'
  1: role=assistant id='msg_1'
  2: role=user      id='msg_1'  ← collision!
  3: role=assistant id='msg_2'
  4: role=user      id='msg_1'  ← collision!
```

## Fix

Move the `_ensure_message_ids(messages)` call from `group_messages()` (which operates on the slice) to `annotate_message_groups()` (which has access to the full list). By running it on the full list before slicing, positional indices are globally unique.

```diff
# annotate_message_groups():
+    _ensure_message_ids(messages)  # full list — globally unique indices
     spans = group_messages(messages[start_index:])

# group_messages():
-    _ensure_message_ids(messages)  # operated on slice — colliding indices
     spans: list[dict[str, Any]] = []
```

## Test

Added `test_incremental_annotation_produces_unique_message_ids` which simulates a 4-turn conversation with incremental annotation (matching the `CompactionProvider` usage pattern) and asserts all `message_id` values are unique.

### DIFF
--- a/python/packages/core/agent_framework/_compaction.py
+++ b/python/packages/core/agent_framework/_compaction.py
@@ -110,8 +110,12 @@ def group_messages(messages: list[Message]) -> list[dict[str, Any]]:
     Returns:
         Ordered list of lightweight span dicts with keys:
         ``group_id``, ``kind``, ``start_index``, ``end_index``, ``has_reasoning``.
+
+    Note:
+        Callers must ensure ``message_id`` is set on each message before
+        calling this function.  ``annotate_message_groups`` handles this
+        automatically via ``_ensure_message_ids``.
     """
-    _ensure_message_ids(messages)
     spans: list[dict[str, Any]] = []
     i = 0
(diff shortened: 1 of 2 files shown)


LEDazzio01:

PR #5298, opened 2026-04-16
Python: Add special handling for workflows

TaoChenOSU:
### Contribution Checklist

### DIFF
--- a/python/packages/foundry_hosting/agent_framework_foundry_hosting/_responses.py
+++ b/python/packages/foundry_hosting/agent_framework_foundry_hosting/_responses.py
@@ -5,9 +5,19 @@ from __future__ import annotations
 import asyncio
 import json
 import logging
+import os
 from collections.abc import AsyncIterable, AsyncIterator, Generator, Mapping
 
-from agent_framework import ChatOptions, Content, HistoryProvider, Message, RawAgent, SupportsAgentRun
+from agent_framework import (
+    ChatOptions,
+    Content,
+    FileCheckpointStorage,
+    HistoryProvider,
+    Message,
+    RawAgent,
+    SupportsAgentRun,
+    WorkflowAgent,
+)
 from agent_framework._telemetry import append_to_user_agent
 from azure.ai.agentserver.responses import (
     ResponseContext,
@@ -60,6 +70,8 @@ class ResponsesHostServer(ResponsesAgentServerHost):
     """A responses server host for an agent."""
 
     USER_AGENT_PREFIX = "foundry-hosting"
+    # TODO(@taochen): Allow a different checkpoint storage that stores checkpoints externally
+    CHECKPOINT_STORAGE_PATH = "/.checkpoints"
 
     def __init__(
         self,
@@ -80,8 +92,11 @@ class ResponsesHostServer(ResponsesAgentServerHost):
             **kwargs: Additional keyword arguments.
 
         Note:
-            The agent must not have a history provider with `load_messages=True`,
-            because history is managed by the hosting infrastructure.
+            1. The agent must not have a history provider with `load_messages=True`,
+               because history is managed by the hosting infrastructure.
+            2. The agent must not have any context providers that maintain context
+               in memory, because the hosting environment may get deactivated between
+               requests, and any in-memory context would be lost.
         """
         super().__init__(prefix=prefix, options=options, store=store, **kwargs)
 
@@ -91,8 +106,24 @@ class ResponsesHostServer(ResponsesAgentServerHost):
                     "There shouldn't be a history provider with `load_messages=True` already present. "
                     "History is managed by the hosting infrastructure."
                 )
-        self._agent = agent
 
+        self._is_workflow_agent = False
+        self._checkpoint_storage = None
+        if isinstance(agent, WorkflowAgent):
+            if agent.workflow._runner_context.has_checkpointing():  # pyright: ignore[reportPrivateUsage]
+                raise RuntimeError(
+                    "There should not be a checkpoint storage already present in the workflow agent. "
+                    "The hosting infrastructure will manage checkpoints instead."
+                )
+            checkpoint_storage_path = (
+                self.CHECKPOINT_STORAGE_PATH
+                if self.config.is_hosted
+                else os.path.join(os.getcwd(), self.CHECKPOINT_STORAGE_PATH.lstrip("/"))
+            )
+            self._is_workflow_agent = True
+            self._checkpoint_storage = FileCheckpointStorage(checkpoint_storage_path)
+
+        self._agent = agent
         self.response_handler(self._handler)  # pyright: ignore[reportUnknownMemberType]
 
         # Append the user agent prefix for telemetry purposes
@@ -105,16 +136,22 @@ class ResponsesHostServer(ResponsesAgentServerHost):
         cancellation_signal: asyncio.Event,
     ) -> AsyncIterable[ResponseStreamEvent | dict[str, Any]]:
         """Handle the creation of a response."""
+        if self._is_workflow_agent:
+            # Workflow agents are handled differently because they require checkpoint restoration
+            async for event in self._handle_workflow_agent(request, context, cancellation_signal):
+                yield event
+            return
+
         input_text = await context.get_input_text()
         history = await context.get_history()
         messages = [*_to_messages(history), input_text]
 
         chat_options = _to_chat_options(request)
 
-        stream = ResponseEventStream(response_id=context.response_id, model=request.model)
+        response_event_stream = ResponseEventStream(response_id=context.response_id, model=request.model)
 
-        yield stream.emit_created()
-        yield stream.emit_in_progress()
+        yield response_event_stream.emit_created()
+        yield response_event_stream.emit_in_progress()
 
         if request.stream is None or request.stream is False:
             # Run the agent in non-streaming mode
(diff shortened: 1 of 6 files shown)


python/packages/foundry_hosting/agent_framework_foundry_hosting/_responses.py
```diff
@@ -141,22 +178,93 @@ async def _handler(
...
+        # This is for type safety and defensive programming.
+        if self._checkpoint_storage is None:
+            raise RuntimeError("Checkpoint storage is not available for workflow agent.")
+        if not isinstance(self._agent, WorkflowAgent):
+            raise RuntimeError("Agent is not a workflow agent.")
+
+        # Restore from the latest checkpoint if available, otherwise start with an empty history
+        latest_checkpoint = await self._checkpoint_storage.get_latest(workflow_name=self._agent.workflow.name)
+        if latest_checkpoint is not None:
+            _ = await self._agent.run(
```
moonbox3: The restoration call here unconditionally awaits `self._agent.run(stream=stream, checkpoint_id=…)`, but a few lines below the same `agent.run` is treated differently for the two modes. Line 237 uses `response = await self._agent.run(input_text, stream=False, …)` while line 248 uses `response_stream = self._agent.run(input_text, stream=True, …)` with no await and then iterates via `async for`.

If `agent.run(stream=True)` returns a stream object rather than an awaitable, which the surrounding code assumes, does this restoration call either raise at runtime in the streaming case or silently no-op without ever driving the workflow forward? That would mean streaming requests resume against an un-rehydrated state.

Should the restoration mirror the same `if stream: ... else: await ...` shape used below, and in the streaming case actually `async for _ in ...:` to drive the resume to completion?

moonbox3: Also, throwing the result away with `_` reads like the intent is “load checkpoint state into memory and return.” But `agent.run(checkpoint_id=…)` looks like it actually executes the workflow from the checkpointed position.

If the resumed workflow has any pending edges or `request_info` left at the checkpoint boundary, won’t that work execute and emit events that get silently dropped, since `response_event_stream` is not even constructed yet on lines 230–233?

Could you clarify in a comment whether `run(checkpoint_id=…)` is purely state-loading or actually executes? If it executes, should those events be forwarded to the response stream rather than discarded?

python/packages/foundry_hosting/agent_framework_foundry_hosting/_responses.py
```diff
@@ -141,22 +178,93 @@ async def _handler(
...
+
+        # The following should never happen due to the checks above.
+        # This is for type safety and defensive programming.
+        if self._checkpoint_storage is None:
+            raise RuntimeError("Checkpoint storage is not available for workflow agent.")
+        if not isinstance(self._agent, WorkflowAgent):
+            raise RuntimeError("Agent is not a workflow agent.")
+
+        # Restore from the latest checkpoint if available, otherwise start with an empty history
+        latest_checkpoint = await self._checkpoint_storage.get_latest(workflow_name=self._agent.workflow.name)
```
moonbox3: The lookup key here is the workflow’s class name, which is the same string for every request that hits this server. That seems to assume the hosting infrastructure pins one sandbox, and therefore one `/.checkpoints` directory, per session.

Is that assumption documented anywhere we can point to? If two requests or users ever land in the same sandbox, or two `WorkflowAgent`s with the same default `workflow.name` share storage, wouldn’t `get_latest` return another conversation’s state?

Should this also include a per-session or per-conversation key, derived from something like `context.response_id` or a request session identifier, so isolation does not depend solely on infrastructure invariants?

python/packages/foundry_hosting/agent_framework_foundry_hosting/_responses.py
```diff
@@ -141,22 +178,93 @@ async def _handler(
...
+        context: ResponseContext,
+        cancellation_signal: asyncio.Event,
+    ) -> AsyncIterable[ResponseStreamEvent | dict[str, Any]]:
+        """Handle the creation of a response for a workflow agent.
+
+        Why this is required:
+        The sandbox may be deactivated after some period of inactivity, and only data managed
+        by the hosting infrastructure or files will be preserved upon deactivation.
+        """
+        input_text = await context.get_input_text()
```
moonbox3, line 226: The non-workflow handler builds `chat_options = _to_chat_options(request)` on line 149 and forwards it to `raw_agent.run(...)`. The workflow path never builds or forwards it, so request-level parameters such as temperature, max_tokens, and tool_choice are dropped with no warning.

Is that intentional because `WorkflowAgent.run` does not support them? If so, should we at least log or raise when a request specifies options that will be ignored, so callers do not assume their settings are taking effect?

python/packages/foundry_hosting/agent_framework_foundry_hosting/_responses.py
```diff
@@ -80,8 +92,11 @@ def __init__(
             **kwargs: Additional keyword arguments.
 
         Note:
-            The agent must not have a history provider with `load_messages=True`,
-            because history is managed by the hosting infrastructure.
+            1. The agent must not have a history provider with `load_messages=True`,
```
moonbox3, line 96: The new docstring rule states that the agent must not have context providers that maintain context in memory, but this is only described here. The neighboring `HistoryProvider(load_messages=True)` rule is enforced with a `RuntimeError` on lines 104–108.

Won’t users encounter silently broken cross-resume state if they violate this rule? Could we add a parallel check, even a heuristic one, such as warning or raising on any `ContextProvider` subclass that is not `HistoryProvider` or does not opt into a “checkpoint-safe” marker, so the constraint rather than devs relying on reading the code comments?

python/packages/foundry_hosting/agent_framework_foundry_hosting/_responses.py
```diff
@@ -60,6 +70,8 @@ class ResponsesHostServer(ResponsesAgentServerHost):
     """A responses server host for an agent."""
 
     USER_AGENT_PREFIX = "foundry-hosting"
+    # TODO(@taochen): Allow a different checkpoint storage that stores checkpoints externally
+    CHECKPOINT_STORAGE_PATH = "/.checkpoints"
```
moonbox3, line 75: The TODO notes this should be pluggable, but the constructor does not take a `checkpoint_storage=` parameter today. Users cannot provide their own implementation, such as Azure Cosmos, without subclassing and patching `_checkpoint_storage`.

Could we add the keyword argument now, defaulting to constructing `FileCheckpointStorage` as today, so the public surface remains forward compatible? It would also help to file a tracking issue so the TODO has a concrete reference.

python/packages/foundry_hosting/agent_framework_foundry_hosting/_responses.py
```diff
@@ -141,22 +178,93 @@ async def _handler(
...
+
+        # Now run the agent with the latest input
+        response_event_stream = ResponseEventStream(response_id=context.response_id, model=request.model)
+
+        yield response_event_stream.emit_created()
+        yield response_event_stream.emit_in_progress()
+
+        if not stream:
+            # Run the agent in non-streaming mode
+            response = await self._agent.run(input_text, stream=False, checkpoint_storage=self._checkpoint_storage)
```
moonbox3: Each request through this path appends a new checkpoint, and `get_latest` is the only consumer. Older checkpoints are never cleaned up.

For a hosted workflow that runs for hours or days, does `/.checkpoints` grow without bound until the sandbox runs out of disk? Should we either rotate, for example keeping the last N checkpoints, at this layer? Or how is this supposed to work?

python/packages/foundry_hosting/agent_framework_foundry_hosting/_responses.py
```diff
@@ -141,22 +178,93 @@ async def _handler(
...
+        cancellation_signal: asyncio.Event,
+    ) -> AsyncIterable[ResponseStreamEvent | dict[str, Any]]:
+        """Handle the creation of a response for a workflow agent.
+
+        Why this is required:
+        The sandbox may be deactivated after some period of inactivity, and only data managed
+        by the hosting infrastructure or files will be preserved upon deactivation.
+        """
+        input_text = await context.get_input_text()
+        stream = request.stream is not None and request.stream is True
```
moonbox3: A small request: compared to line 156, these compute the same boolean in two different ways within the same file.

Could we extract a small helper such as `_is_streaming(request)` and reuse it from both `_handler` and `_handle_workflow_agent`?

python/packages/foundry_hosting/agent_framework_foundry_hosting/_responses.py
TaoChenOSU: Good catch, will fix

TaoChenOSU: We restore from the latest checkpoint. The latest checkpoint comes from the previous invocation, and it can only be in two states:
- IDLE (it's possible that the workflow failed)
- IDLE_WITH_PENDING_REQUESTS

All request info events should've been streamed back to the client in the previous turn so we can safely discard them after restoration. The new input is expected to contain the responses, or the workflow will raise.

python/packages/foundry_hosting/agent_framework_foundry_hosting/_responses.py
TaoChenOSU: Sandboxes are segregated by sessions and users. Let me double check conversations.

TaoChenOSU: Checkpoints are now managed by conversations or previous response ids.

python/packages/foundry_hosting/agent_framework_foundry_hosting/_responses.py
TaoChenOSU: Need to think more about this. Out of scope for this PR.

alliscode (approved):

PR #5312, opened 2026-04-16
.NET [WIP] Foundry Hosted Agents Support

rogerbarreto:
### Contribution Checklist


dotnet/src/Microsoft.Agents.AI.Foundry.Hosting/AgentSessionStore.cs
```diff
@@ -0,0 +1,46 @@
...
+
+/// <summary>
+/// Defines the contract for storing and retrieving agent conversation sessions.
+/// </summary>
+/// <remarks>
+/// Implementations of this interface enable persistent storage of conversation sessions,
+/// allowing conversations to be resumed across HTTP requests, application restarts,
+/// or different service instances in hosted scenarios.
+/// </remarks>
+public abstract class AgentSessionStore
```
westey-m, line 19: We have a similar class in Microsoft.Agents.AI.Hosting.  Should we consolidate?

dotnet/src/Microsoft.Agents.AI.Foundry.Hosting/AgentFrameworkResponseHandler.cs
```diff
@@ -0,0 +1,376 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Agents.AI.Foundry.Hosting;
+
+/// <summary>
+/// A <see cref="ResponseHandler"/> implementation that bridges the Azure AI Responses Server SDK
+/// with agent-framework <see cref="AIAgent"/> instances, enabling agent-framework agents and workflows
+/// to be hosted as Azure Foundry Hosted Agents.
+/// </summary>
+public class AgentFrameworkResponseHandler : ResponseHandler
```
westey-m, line 23: Should all of these be experimental?  Since the project is not, pushing them out like this would not allow us any chance of making changes.

dotnet/src/Microsoft.Agents.AI.Foundry/Microsoft.Agents.AI.Foundry.csproj
```diff
@@ -3,7 +3,8 @@
   <PropertyGroup>
     <IsReleased>true</IsReleased>
     <InjectSharedThrow>true</InjectSharedThrow>
-    <NoWarn>$(NoWarn);OPENAI001</NoWarn>
+    <NoWarn>$(NoWarn);OPENAI001;MEAI001;NU1903</NoWarn> <!-- NU1903: Microsoft.Bcl.Memory 9.0.4 transitive vulnerability via Azure SDK; awaiting upstream fix -->
```
westey-m: Can we remove this NU1903 suppression now?  I'm assuming we don't want to release with the vulnerable transitive dependency?

dotnet/src/Microsoft.Agents.AI.Foundry/Microsoft.Agents.AI.Foundry.csproj
```diff
@@ -20,14 +21,25 @@
     <InjectSharedRedaction>true</InjectSharedRedaction>
   </PropertyGroup>
 
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp'">
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
```
westey-m: This dependency here isn't great.  It means that anyone that just wants to build an agent using a foundry hosted model end up with asp.net as a dependency.

Should we consider putting the hosting in it's own nuget, e.g. Microsoft.Agents.AI.Hosting.Foundry?

alliscode (commented):
## Automated Code Review

**Reviewers:** 1 | **Confidence:** 88% | **Result:** All clear

Reviewed: Security Reliability

---

dotnet/src/Microsoft.Agents.AI.Foundry.Hosting/AgentFrameworkResponseHandler.cs
```diff
@@ -0,0 +1,376 @@
...
+        // 2. Load or create a new session from the interaction
+        var sessionConversationId = request.GetConversationId();
+
+        var chatClientAgent = agent.GetService<ChatClientAgent>();
+
+        AgentSession? session = !string.IsNullOrWhiteSpace(sessionConversationId)
+            ? await sessionStore.GetSessionAsync(agent, sessionConversationId, cancellationToken).ConfigureAwait(false)
+                : chatClientAgent is not null
+                ? await chatClientAgent.CreateSessionAsync(cancellationToken).ConfigureAwait(false)
+                : await agent.CreateSessionAsync(cancellationToken).ConfigureAwait(false);
```
lokitoth, line 68: [nit] Consider aligning corresponding `?` and `:` operators from ternary expressions, particularly where there is nesting to make it easier to read:


```suggestion
        AgentSession? session 
            = !string.IsNullOrWhiteSpace(sessionConversationId)
            ? await sessionStore.GetSessionAsync(agent, sessionConversationId, cancellationToken).ConfigureAwait(false)
            : chatClientAgent is not null
                ? await chatClientAgent.CreateSessionAsync(cancellationToken).ConfigureAwait(false)
                : await agent.CreateSessionAsync(cancellationToken).ConfigureAwait(false);
```

dotnet/src/Microsoft.Agents.AI.Foundry/FoundryAITool.cs
```diff
@@ -112,6 +112,50 @@ public static AITool CreateStructuredOutputsTool(StructuredOutputDefinition outp
...
         => ProjectsAgentTool.CreateA2ATool(baseUri, agentCardPath).AsAITool();
 
+    /// <summary>
+    /// Creates an <see cref="AITool"/> marker that references a Foundry Toolbox by name so
+    /// the hosted server side can resolve and expose its MCP tools for a single request.
+    /// </summary>
+    /// <param name="toolboxName">The Foundry toolbox name.</param>
+    /// <param name="version">Optional pinned toolbox version. When <see langword="null"/>, the project's default version is used.</param>
+    /// <returns>An <see cref="AITool"/> marker backed by <see cref="HostedMcpToolboxAITool"/>.</returns>
+    public static AITool CreateHostedMcpToolbox(string toolboxName, string? version = null)
```
westey-m, line 122: These helpers should probably also be experimental.

dotnet/src/Microsoft.Agents.AI.Foundry.Hosting/InputConverter.cs
```diff
@@ -0,0 +1,353 @@
...
+        if (contents.Count == 0)
+        {
+            contents.Add(new MeaiTextContent(string.Empty));
+        }
+
+        return new ChatMessage(role, contents);
+    }
+
+    [UnconditionalSuppressMessage("Trimming", "IL2026", Justification = "Deserializing function call arguments from SDK output history.")]
+    [UnconditionalSuppressMessage("AOT", "IL3050", Justification = "Deserializing function call arguments from SDK output history.")]
```
westey-m, line 314: Another reason to move this out to it's own package would be that we can mark the new package as not AOT/Trimming compatible until we can implement support for it.

dotnet/src/Microsoft.Agents.AI.Foundry.Hosting/InMemoryAgentSessionStore.cs
```diff
@@ -0,0 +1,53 @@
...
+/// <item><description>Scenarios where session persistence across restarts is not required</description></item>
+/// </list>
+/// </para>
+/// <para>
+/// <strong>Warning:</strong> All stored sessions will be lost when the application restarts.
+/// For production use with multiple instances or persistence across restarts, use a durable storage implementation
+/// such as Redis, SQL Server, or Azure Cosmos DB.
+/// </para>
+/// </remarks>
+public sealed class InMemoryAgentSessionStore : AgentSessionStore
```
westey-m, line 31: We could use the one from Microsoft.Agents.AI.Hosting instead of having our own copy.

dotnet/src/Microsoft.Agents.AI.Foundry.Hosting/AgentFrameworkResponseHandler.cs
```diff
@@ -0,0 +1,376 @@
...
+        var inputItems = await context.GetInputItemsAsync(cancellationToken: cancellationToken).ConfigureAwait(false);
+        if (inputItems.Count > 0)
+        {
+            messages.AddRange(InputConverter.ConvertItemsToMessages(inputItems));
+        }
+        else
+        {
+            // Fall back to raw request input
+            messages.AddRange(InputConverter.ConvertInputToMessages(request));
+        }
```
westey-m, line 97: This requires more work.  For most agent types, this will result in duplicate chat history being presented to the LLM.  Probably something we'd have to look into as a follow up though.

We probably should create a custom ChatHistoryProvider that we seed with the chat history from the context, and pass that ChatHistoryProvider as an override to the agent.
Note that the provider may be ignored if the agent is itself using some AI service that uses service stored chat history.  This would mean that the agent would use the service's chat history rather than whatever is coming from the context object.

westey-m (approved):

dotnet/src/Microsoft.Agents.AI.Foundry.Hosting/AgentFrameworkResponseHandler.cs
```diff
@@ -0,0 +1,376 @@
...
+            }
+        }
+        finally
+        {
+            await enumerator.DisposeAsync().ConfigureAwait(false);
+
+            // Persist session after streaming completes (successful or not)
+            if (session is not null && !string.IsNullOrWhiteSpace(sessionConversationId))
+            {
+                await sessionStore.SaveSessionAsync(agent, sessionConversationId, session, CancellationToken.None).ConfigureAwait(false);
```
peibekwe: nit: Should the CancellationToken still be propagated here?

dotnet/src/Microsoft.Agents.AI.Foundry.Hosting/OutputConverter.cs
```diff
@@ -0,0 +1,349 @@
...
+
+                var builder = stream.AddOutputItem<WorkflowActionOutputItem>(itemId);
+                yield return builder.EmitAdded(item);
+                yield return builder.EmitDone(item);
+                break;
+            }
+
+            case ExecutorCompletedEvent completedEvent:
+            {
+                var itemId = GenerateItemId("wfa");
```
peibekwe, line 293: nit: Since `InvokedEvent` already created an `itemId` and saved it in `executorItemIds`. Should completed and failed paths be reusing that Id instead of creating new ones or do they need to be unique?

dotnet/src/Microsoft.Agents.AI.Foundry/Microsoft.Agents.AI.Foundry.csproj
```diff
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <IsReleased>true</IsReleased>
     <InjectSharedThrow>true</InjectSharedThrow>
-    <NoWarn>$(NoWarn);OPENAI001</NoWarn>
+    <NoWarn>$(NoWarn);OPENAI001;MEAI001</NoWarn>
```
westey-m: Do we still need the MEAI001 suppression?  Anywhere we use something experimental we should expose those things as experimental where possible.  So generally it's good not to suppress this here, but just prorogate experimental up the stack.  It's fine suppressing in a preview package, but we should be careful about this in a released package, since it can hide instances where we are using something experimental and unintentionally exposing users to that.

dotnet/src/Microsoft.Agents.AI.Foundry/FoundryAITool.cs
alliscode: Already experimental at the class level.

dotnet/src/Microsoft.Agents.AI.Foundry.Hosting/AgentSessionStore.cs
alliscode: Will consolidate in a follow up PR.

dotnet/src/Microsoft.Agents.AI.Foundry.Hosting/AgentFrameworkResponseHandler.cs
alliscode: This has been moved to a preview package so no experimental required.

dotnet/src/Microsoft.Agents.AI.Foundry/Microsoft.Agents.AI.Foundry.csproj
alliscode: Moved hosting stuff into a preview package.

dotnet/src/Microsoft.Agents.AI.Foundry.Hosting/InMemoryAgentSessionStore.cs
alliscode: Will consolidate these in a follow up PR.

dotnet/src/Microsoft.Agents.AI.Foundry.Hosting/AgentFrameworkResponseHandler.cs
alliscode: Will follow up with you to improve this.

dotnet/src/Microsoft.Agents.AI.Foundry.Hosting/OutputConverter.cs
alliscode: Good catch — executorItemIds saves the itemId from InvokedEvent but completed/failed paths generate new IDs instead of reusing it. This looks like it may be a bug where we intended to correlate lifecycle events under the same item ID. Will investigate the Responses stream protocol semantics and fix in a follow-up

dotnet/src/Microsoft.Agents.AI.Foundry.Hosting/AgentFrameworkResponseHandler.cs
alliscode: Yes, fixed.

peibekwe (approved):

dotnet/src/Microsoft.Agents.AI.Foundry.Hosting/AgentFrameworkResponseHandler.cs
alliscode: Good point, will address in a follow up PR.

PR #5321, opened 2026-04-17
.NET: Support returning durable workflow results from HTTP trigger endpoint

kshyju:
## Summary

Adds support for synchronously returning workflow execution results from the HTTP trigger endpoint. By default, the workflow HTTP endpoint fires-and-forgets (202 Accepted). With this change, callers can set the `x-ms-wait-for-response: true` header to wait for the workflow to complete and receive the result directly in the HTTP response (200 OK). Results are returned as plain text by default, or as a structured JSON object when `Accept: application/json` is specified.

### New feature
- Parse the `x-ms-wait-for-response` header in `RunWorkflowOrchestrationHttpTriggerAsync`
- New `WaitForWorkflowCompletionAsync` method that calls `WaitForInstanceCompletionAsync` and handles failure/completed/unexpected statuses
- Support `Accept: application/json` header to return a structured JSON response (`runId`, `status`, `result`)
- POCO results from executors are properly serialized as nested JSON objects (using `JsonElement`) rather than double-encoded strings

### Code quality improvements
- Extracted `WaitForResponseHeaderName` constant (replaces two inline string literals)
- Extracted `ShouldWaitForResponse(req, defaultValue)` helper to eliminate duplicated header-parsing logic across workflow and agent triggers
- Extracted `AcceptsJson(req)` helper to deduplicate `Accept` header checks across 4 response methods
- Added optional `acceptsJson` parameter to `CreateErrorResponseAsync` to avoid redundant header reads within `WaitForWorkflowCompletionAsync`

### Docs & samples
- Updated `01_SequentialWorkflow/README.md` with curl/PowerShell examples for the new header
- Added demo.http entries for wait-for-response scenarios (plain text and JSON)

### Tests
- Added integration test in `WorkflowSamplesValidation.cs` verifying the header works end-to-end

## JSON response examples

**Plain text** (default):
```text
Cancellation email sent for order 12345 to jerry@example.com.
```

**JSON** (with `Accept: application/json`):
```json
{
  "runId": "abc123def456",
  "status": "Completed",
  "result": "Cancellation email sent for order 12345 to jerry@example.com."
}
```

**JSON with POCO result** (executor returns an object):
```json
{
  "runId": "abc123def456",
  "status": "Completed",
  "result": { "orderId": 123, "email": "jerry@example.com" }
}
```

### Contribution Checklist


dotnet/src/Microsoft.Agents.AI.Hosting.AzureFunctions/BuiltInFunctions.cs
```diff
@@ -428,25 +427,99 @@ await agentProxy.RunAsync(
...
+
+        OrchestrationMetadata? metadata = await client.WaitForInstanceCompletionAsync(
+            instanceId,
+            getInputsAndOutputs: true,
+            cancellation: context.CancellationToken);
+
+        if (metadata is null)
+        {
+            return await CreateErrorResponseAsync(req, context, HttpStatusCode.InternalServerError,
+                $"Workflow orchestration '{instanceId}' returned no metadata.", acceptsJson);
```
cgillum, line 449: It's also possible that no orchestration with this ID exists.

dotnet/src/Microsoft.Agents.AI.Hosting.AzureFunctions/BuiltInFunctions.cs
```diff
@@ -428,25 +427,99 @@ await agentProxy.RunAsync(
...
+        }
+
+        if (metadata.RuntimeStatus is OrchestrationRuntimeStatus.Failed)
+        {
+            string errorMessage = metadata.FailureDetails?.ErrorMessage ?? "Unknown error";
+            return await CreateErrorResponseAsync(req, context, HttpStatusCode.InternalServerError,
```
cgillum, line 455: Is `InternalServerError` the right status code for us to use on a failed workflow? I wonder if an HTTP 200 response is actually more appropriate from an HTP protocol perspective. HTTP 500 typically means that we failed to process the request due to some internal failure, but in this case, the operation succeeded, but the state of the workflow happens to be that it failed.

dotnet/src/Microsoft.Agents.AI.Hosting.AzureFunctions/BuiltInFunctions.cs
```diff
@@ -591,6 +686,17 @@ private sealed record WorkflowRespondRequest(
...
 
+    /// <summary>
+    /// Represents a successful workflow run response when waiting for completion.
+    /// </summary>
+    /// <param name="RunId">The orchestration run ID.</param>
+    /// <param name="Status">The orchestration runtime status.</param>
+    /// <param name="Result">The workflow result as a JSON element so POCOs serialize as nested objects rather than escaped strings.</param>
+    private sealed record WorkflowRunSuccessResponse(
+        [property: JsonPropertyName("runId")] string RunId,
+        [property: JsonPropertyName("status")] string Status,
```
cgillum, line 697: There's an inconsistency here with `AgentRunSuccessResponse`, which returns the status as an integer value rather than a string value. The string value probably makes more sense, but to resolve this inconsistency, consider using a different property name, like `WorkflowStatus`, to avoid confusion.

dotnet/src/Microsoft.Agents.AI.Hosting.AzureFunctions/BuiltInFunctions.cs
kshyju: Good point. Switched to 200 with error details in the body (Error property)

dotnet/src/Microsoft.Agents.AI.Hosting.AzureFunctions/BuiltInFunctions.cs
kshyju: Switched the error message to "No workflow orchestration with ID '{instanceId}' was found. and is now returning 404 response.

dotnet/src/Microsoft.Agents.AI.Hosting.AzureFunctions/BuiltInFunctions.cs
kshyju: Switched to WorkflowStatus property name.

cgillum (approved):

ahmedmuhsin (approved):

PR #5329, opened 2026-04-17
.NET: Add Microsoft.Agents.AI.Hyperlight package for CodeAct integration (.NET)

### LINKED ISSUE #5188: .NET: [Feature]: .NET CodeAct implementation
### Description

.Net codeAct implementation

### Code Sample

```markdown

```

### Language/SDK

.NET

eavanvalkenburg:
Adds a new `Microsoft.Agents.AI.Hyperlight` package that brings first-class CodeAct support to the .NET Agent Framework, backed by the Hyperlight sandbox via the new .NET SDK from [hyperlight-dev/hyperlight-sandbox PR 46](https://github.com/hyperlight-dev/hyperlight-sandbox/pull/46).

Follows [`docs/features/code_act/dotnet-implementation.md`](https://github.com/microsoft/agent-framework/blob/main/docs/features/code_act/dotnet-implementation.md) and the Python `agent_framework_hyperlight` reference.

## What's included

- `HyperlightCodeActProvider` (`AIContextProvider`):
  - Injects an `execute_code` tool + CodeAct guidance into every invocation.
  - Single-instance-per-agent enforced via a fixed `StateKeys` value (`ChatClientAgent` rejects duplicate state keys).
  - CRUD for provider-owned tools, file mounts, and outbound-domain allow-list — all exposed inside the sandbox via `call_tool(...)`.
  - Run-scoped snapshot captured under a lock so mutations between runs don't affect in-flight executions.
- `HyperlightExecuteCodeFunction`: standalone `AIFunction` for manual/static wiring.
- Async tool registration end-to-end (`Sandbox.RegisterToolAsync`).
- Approval model via `CodeActApprovalMode` (`AlwaysRequire` / `NeverRequire` with propagation from `ApprovalRequiredAIFunction`-wrapped tools).
- Both entry points accept `options = null` and fall back to defaults (mirroring the Python reference).

## Tests & samples

- Unit tests: instruction builder, tool bridge (JSON round-trip + error wrapping + approval unwrap), approval computation matrix, provider CRUD, snapshot isolation, approval wrapping in `ProvideAIContextAsync`.
- Integration test: env-gated on `HYPERLIGHT_PYTHON_GUEST_PATH`, runs `print("hi")` end-to-end.
- Three samples under `dotnet/samples/02-agents/AgentWithCodeAct/`: Interpreter, ToolEnabled (with an `ApprovalRequiredAIFunction`-wrapped tool), ManualWiring.

## Draft — not yet buildable

- Requires **.NET SDK 10.0.200** (per repo `global.json`).
- Requires **`HyperlightSandbox.Api 0.1.0-preview`**, which is **not yet published** (`hyperlight-dev/hyperlight-sandbox` PR 46). The src project is `IsPackable=false` until the dependency is available.
- Nothing in this PR has been compiled or run locally yet. Needs to be validated end-to-end once the Hyperlight .NET SDK lands on nuget.org (or a local feed is set up).

## Open items to resolve before un-drafting

- Verify compile on `net8.0` / `net9.0` / `net10.0` against the real SDK.
- Confirm `Sandbox.RegisterToolAsync` return-type handling in `ToolBridge.RegisterOne`.
- Decide on richer per-mount filesystem API once the Hyperlight SDK exposes one (today, `FileMount` entries are surfaced only in the `execute_code` description).
- Add CHANGELOG entry if required by area owners.

Tracks: CodeAct .NET implementation (see `docs/features/code_act/dotnet-implementation.md`).

Closes #5188

### DIFF
--- a/dotnet/samples/02-agents/AgentWithCodeAct/AgentWithCodeAct_Step01_Interpreter/Program.cs
+++ b/dotnet/samples/02-agents/AgentWithCodeAct/AgentWithCodeAct_Step01_Interpreter/Program.cs
@@ -1,0 +1,34 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+// This sample shows how to use HyperlightCodeActProvider as a sandboxed Python
+// code interpreter: the model can write and execute arbitrary Python code to
+// answer quantitative questions without calling any additional tools.
+
+using Azure.AI.OpenAI;
+using Azure.Identity;
+using HyperlightSandbox.Api;
+using Microsoft.Agents.AI;
+using Microsoft.Agents.AI.Hyperlight;
+
+var endpoint = Environment.GetEnvironmentVariable("AZURE_OPENAI_ENDPOINT") ?? throw new InvalidOperationException("AZURE_OPENAI_ENDPOINT is not set.");
+var deploymentName = Environment.GetEnvironmentVariable("AZURE_OPENAI_DEPLOYMENT_NAME") ?? "gpt-5.4-mini";
+var guestPath = Environment.GetEnvironmentVariable("HYPERLIGHT_PYTHON_GUEST_PATH") ?? throw new InvalidOperationException("HYPERLIGHT_PYTHON_GUEST_PATH is not set.");
+
+using var codeAct = new HyperlightCodeActProvider(new HyperlightCodeActProviderOptions
+{
+    Backend = SandboxBackend.Wasm,
+    ModulePath = guestPath,
+});
+
+AIAgent agent = new AzureOpenAIClient(
+    new Uri(endpoint),
+    new DefaultAzureCredential())
+    .GetChatClient(deploymentName)
+    .AsAIAgent(new ChatClientAgentOptions()
+    {
+        ChatOptions = new() { Instructions = "You are a helpful assistant. When the user asks something quantitative, write Python and call `execute_code` instead of guessing." },
+        AIContextProviders = [codeAct],
+    });
+
+Console.WriteLine(await agent.RunAsync("What is the 20th Fibonacci number?"));
+Console.WriteLine(await agent.RunAsync("Compute the mean and standard deviation of [1, 4, 9, 16, 25, 36]."));
--- a/dotnet/samples/02-agents/AgentWithCodeAct/AgentWithCodeAct_Step01_Interpreter/Program.cs
+++ b/dotnet/samples/02-agents/AgentWithCodeAct/AgentWithCodeAct_Step01_Interpreter/Program.cs
@@ -6,7 +6,6 @@
 
 using Azure.AI.OpenAI;
 using Azure.Identity;
-using HyperlightSandbox.Api;
 using Microsoft.Agents.AI;
 using Microsoft.Agents.AI.Hyperlight;
 
@@ -14,11 +13,7 @@ var endpoint = Environment.GetEnvironmentVariable("AZURE_OPENAI_ENDPOINT") ?? th
 var deploymentName = Environment.GetEnvironmentVariable("AZURE_OPENAI_DEPLOYMENT_NAME") ?? "gpt-5.4-mini";
 var guestPath = Environment.GetEnvironmentVariable("HYPERLIGHT_PYTHON_GUEST_PATH") ?? throw new InvalidOperationException("HYPERLIGHT_PYTHON_GUEST_PATH is not set.");
 
-using var codeAct = new HyperlightCodeActProvider(new HyperlightCodeActProviderOptions
-{
-    Backend = SandboxBackend.Wasm,
-    ModulePath = guestPath,
-});
+using var codeAct = new HyperlightCodeActProvider(HyperlightCodeActProviderOptions.CreateForWasm(guestPath));
 
 AIAgent agent = new AzureOpenAIClient(
     new Uri(endpoint),
--- a/dotnet/samples/02-agents/AgentWithCodeAct/AgentWithCodeAct_Step01_Interpreter/Program.cs
+++ b/dotnet/samples/02-agents/AgentWithCodeAct/AgentWithCodeAct_Step01_Interpreter/Program.cs
@@ -8,6 +8,7 @@ using Azure.AI.OpenAI;
 using Azure.Identity;
 using Microsoft.Agents.AI;
 using Microsoft.Agents.AI.Hyperlight;
+using OpenAI.Chat;
 
 var endpoint = Environment.GetEnvironmentVariable("AZURE_OPENAI_ENDPOINT") ?? throw new InvalidOperationException("AZURE_OPENAI_ENDPOINT is not set.");
 var deploymentName = Environment.GetEnvironmentVariable("AZURE_OPENAI_DEPLOYMENT_NAME") ?? "gpt-5.4-mini";
--- a/dotnet/samples/02-agents/AgentWithCodeAct/AgentWithCodeAct_Step01_Interpreter/Program.cs
+++ b/dotnet/samples/02-agents/AgentWithCodeAct/AgentWithCodeAct_Step01_Interpreter/Program.cs
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft. All rights reserved.
+﻿// Copyright (c) Microsoft. All rights reserved.
 
 // This sample shows how to use HyperlightCodeActProvider as a sandboxed Python
 // code interpreter: the model can write and execute arbitrary Python code to
--- a/dotnet/samples/02-agents/AgentWithCodeAct/AgentWithCodeAct_Step01_Interpreter/README.md
+++ b/dotnet/samples/02-agents/AgentWithCodeAct/AgentWithCodeAct_Step01_Interpreter/README.md
@@ -1,0 +1,23 @@
+# AgentWithCodeAct_Step01_Interpreter
+
+A minimal CodeAct sample. The agent uses `HyperlightCodeActProvider` as a
+sandboxed Python interpreter: when the user asks something quantitative, the
+model writes Python and invokes the `execute_code` tool rather than answering
+from memory.
+
+## Configuration
+
+| Variable                       | Description                                                                               |
+|--------------------------------|-------------------------------------------------------------------------------------------|
+| `AZURE_OPENAI_ENDPOINT`        | Azure OpenAI endpoint. Required.                                                          |
+| `AZURE_OPENAI_DEPLOYMENT_NAME` | Azure OpenAI deployment. Defaults to `gpt-5.4-mini`.                                      |
+| `HYPERLIGHT_PYTHON_GUEST_PATH` | Absolute path to the Hyperlight Python guest module (`.wasm` or `.aot` file). Required.   |
+
+Authentication uses `DefaultAzureCredential`.
+
+## Run
+
+```shell
+cd AgentWithCodeAct_Step01_Interpreter
+dotnet run
+```
--- a/dotnet/samples/02-agents/AgentWithCodeAct/AgentWithCodeAct_Step01_Interpreter/README.md
+++ b/dotnet/samples/02-agents/AgentWithCodeAct/AgentWithCodeAct_Step01_Interpreter/README.md
@@ -15,6 +15,18 @@ from memory.
 
 Authentication uses `DefaultAzureCredential`.
 
+## Getting the guest module
+
+The Python guest module is built from the
+[hyperlight-dev/hyperlight-sandbox](https://github.com/hyperlight-dev/hyperlight-sandbox)
+repository — see its README for the exact `cargo`/`just` invocations and
+the location of the resulting `.wasm` / `.aot` file. Set
+`HYPERLIGHT_PYTHON_GUEST_PATH` to the absolute path of that artifact
+before running the sample.
+
+Hyperlight requires a hardware virtualization back end on the host:
+KVM on Linux or WHP (Windows Hypervisor Platform) on Windows.
+
 ## Run
 
 ```shell
--- a/dotnet/samples/02-agents/AgentWithCodeAct/AgentWithCodeAct_Step02_ToolEnabled/README.md
+++ b/dotnet/samples/02-agents/AgentWithCodeAct/AgentWithCodeAct_Step02_ToolEnabled/README.md
@@ -24,3 +24,11 @@ is configured.
 cd AgentWithCodeAct_Step02_ToolEnabled
 dotnet run
 ```
+
+## Planned follow-up
+
+A more realistic "upload a file (e.g. an Excel workbook), have the agent
+analyze it with code" sample is planned as a separate step that will use
+`HostInputDirectory` together with a guest tool capable of reading the
+uploaded file. It will be added in a follow-up PR once the corresponding
+guest module support is in place.
--- a/dotnet/samples/02-agents/AgentWithCodeAct/AgentWithCodeAct_Step02_ToolEnabled/README.md
+++ b/dotnet/samples/02-agents/AgentWithCodeAct/AgentWithCodeAct_Step02_ToolEnabled/README.md
@@ -1,0 +1,26 @@
+# AgentWithCodeAct_Step02_ToolEnabled
+
+Demonstrates adding provider-owned tools to `HyperlightCodeActProvider`. Those
+tools are **only** available to code running inside the sandbox via
+`call_tool("<name>", ...)` — they are never exposed to the model as direct
+tools. This lets the model orchestrate multiple tool calls in a single Python
+block.
+
+One tool (`send_email`) is wrapped in `ApprovalRequiredAIFunction`, which causes
+the entire `execute_code` invocation to require user approval when that tool
+is configured.
+
+## Configuration
+
+| Variable                       | Description                                                                               |
+|--------------------------------|-------------------------------------------------------------------------------------------|
+| `AZURE_OPENAI_ENDPOINT`        | Azure OpenAI endpoint. Required.                                                          |
+| `AZURE_OPENAI_DEPLOYMENT_NAME` | Azure OpenAI deployment. Defaults to `gpt-5.4-mini`.                                      |
+| `HYPERLIGHT_PYTHON_GUEST_PATH` | Absolute path to the Hyperlight Python guest module (`.wasm` or `.aot` file). Required.   |
+
+## Run
+
+```shell
+cd AgentWithCodeAct_Step02_ToolEnabled
+dotnet run
+```
--- a/dotnet/src/Microsoft.Agents.AI.Hyperlight/AllowedDomain.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Hyperlight/AllowedDomain.cs
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft. All rights reserved.
+﻿// Copyright (c) Microsoft. All rights reserved.
 
 using System.Collections.Generic;
 
--- a/dotnet/src/Microsoft.Agents.AI.Hyperlight/AllowedDomain.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Hyperlight/AllowedDomain.cs
@@ -8,11 +8,25 @@ namespace Microsoft.Agents.AI.Hyperlight;
 /// Represents a single entry in the outbound network allow-list applied to the
 /// Hyperlight sandbox.
 /// </summary>
-/// <param name="Target">
-/// URL or domain to allow, for example <c>"https://api.github.com"</c>.
-/// </param>
-/// <param name="Methods">
-/// Optional list of HTTP methods to allow (for example <c>["GET", "POST"]</c>).
-/// When <see langword="null"/>, all methods supported by the backend are allowed.
-/// </param>
-public sealed record AllowedDomain(string Target, IReadOnlyList<string>? Methods = null);
+public sealed class AllowedDomain
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="AllowedDomain"/> class.
+    /// </summary>
+    /// <param name="target">URL or domain to allow, for example <c>"https://api.github.com"</c>.</param>
+    /// <param name="methods">
+    /// Optional list of HTTP methods to allow (for example <c>["GET", "POST"]</c>).
+    /// When <see langword="null"/>, all methods supported by the backend are allowed.
+    /// </param>
+    public AllowedDomain(string target, IReadOnlyList<string>? methods = null)
+    {
+        this.Target = target;
+        this.Methods = methods;
+    }
+
+    /// <summary>Gets the URL or domain to allow.</summary>
+    public string Target { get; }
+
+    /// <summary>Gets the optional list of HTTP methods to allow.</summary>
+    public IReadOnlyList<string>? Methods { get; }
+}
--- a/dotnet/src/Microsoft.Agents.AI.Hyperlight/AllowedDomain.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Hyperlight/AllowedDomain.cs
@@ -1,0 +1,18 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+using System.Collections.Generic;
+
+namespace Microsoft.Agents.AI.Hyperlight;
+
+/// <summary>
+/// Represents a single entry in the outbound network allow-list applied to the
+/// Hyperlight sandbox.
+/// </summary>
+/// <param name="Target">
+/// URL or domain to allow, for example <c>"https://api.github.com"</c>.
+/// </param>
+/// <param name="Methods">
+/// Optional list of HTTP methods to allow (for example <c>["GET", "POST"]</c>).
+/// When <see langword="null"/>, all methods supported by the backend are allowed.
+/// </param>
+public sealed record AllowedDomain(string Target, IReadOnlyList<string>? Methods = null);
--- a/dotnet/src/Microsoft.Agents.AI.Hyperlight/HyperlightCodeActProvider.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Hyperlight/HyperlightCodeActProvider.cs
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft. All rights reserved.
+﻿// Copyright (c) Microsoft. All rights reserved.
 
 using System;
 using System.Collections.Generic;
--- a/dotnet/src/Microsoft.Agents.AI.Hyperlight/HyperlightCodeActProvider.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Hyperlight/HyperlightCodeActProvider.cs
@@ -66,8 +66,10 @@ public sealed class HyperlightCodeActProvider : AIContextProvider, IDisposable
     /// </summary>
     /// <param name="options">
     /// Optional configuration options for the provider. When <see langword="null"/> the provider
-    /// uses the defaults of <see cref="HyperlightCodeActProviderOptions"/> — equivalent to a
-    /// <see cref="SandboxBackend.Wasm"/> backend with no module, tools, mounts, or allow-list entries.
+    /// uses the defaults of <see cref="HyperlightCodeActProviderOptions"/> (the
+    /// <see cref="SandboxBackend.JavaScript"/> backend with no tools, mounts, or allow-list entries).
+    /// Use <see cref="HyperlightCodeActProviderOptions.CreateForWasm(string)"/> to target a Wasm
+    /// guest module instead.
     /// </param>
     public HyperlightCodeActProvider(HyperlightCodeActProviderOptions? options = null)
     {
@@ -76,39 +78,24 @@ public sealed class HyperlightCodeActProvider : AIContextProvider, IDisposable
 
         if (this._options.Tools is not null)
         {
-            foreach (var tool in this._options.Tools)
+            foreach (var tool in this._options.Tools.Where(t => t is not null))
             {
-                if (tool is null)
-                {
-                    continue;
-                }
-
                 this._tools[tool.Name] = tool;
             }
         }
 
         if (this._options.FileMounts is not null)
         {
-            foreach (var mount in this._options.FileMounts)
+            foreach (var mount in this._options.FileMounts.Where(m => m is not null))
             {
-                if (mount is null)
-                {
-                    continue;
-                }
-
                 this._fileMounts[mount.MountPath] = mount;
             }
         }
 
         if (this._options.AllowedDomains is not null)
         {
-            foreach (var domain in this._options.AllowedDomains)
+            foreach (var domain in this._options.AllowedDomains.Where(d => d is not null))
             {
-                if (domain is null)
-                {
-                    continue;
-                }
-
                 this._allowedDomains[domain.Target] = domain;
             }
         }
@@ -129,13 +116,8 @@ public sealed class HyperlightCodeActProvider : AIContextProvider, IDisposable
         lock (this._gate)
         {
             this.ThrowIfDisposed();
-            foreach (var tool in tools)
+            foreach (var tool in tools.Where(t => t is not null))
             {
-                if (tool is null)
-                {
-                    continue;
-                }
-
                 this._tools[tool.Name] = tool;
             }
         }
@@ -146,7 +128,7 @@ public sealed class HyperlightCodeActProvider : AIContextProvider, IDisposable
     {
         lock (this._gate)
         {
-            return this._tools.Values.ToArray();
+            return this._tools.Values.ToList();
         }
     }
 
@@ -157,13 +139,8 @@ public sealed class HyperlightCodeActProvider : AIContextProvider, IDisposable
         _ = Throw.IfNull(names);
         lock (this._gate)
         {
-            foreach (var name in names)
+            foreach (var name in names.Where(n => n is not null))
             {
-                if (name is null)
-                {
-                    continue;
-                }
-
                 _ = this._tools.Remove(name);
             }
         }
@@ -189,13 +166,8 @@ public sealed class HyperlightCodeActProvider : AIContextProvider, IDisposable
         _ = Throw.IfNull(mounts);
         lock (this._gate)
         {
-            foreach (var mount in mounts)
+            foreach (var mount in mounts.Where(m => m is not null))
             {
-                if (mount is null)
-                {
-                    continue;
-                }
-
                 this._fileMounts[mount.MountPath] = mount;
             }
         }
@@ -206,7 +178,7 @@ public sealed class HyperlightCodeActProvider : AIContextProvider, IDisposable
     {
         lock (this._gate)
         {
-            return this._fileMounts.Values.ToArray();
+            return this._fileMounts.Values.ToList();
         }
     }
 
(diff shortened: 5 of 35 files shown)


dotnet/src/Microsoft.Agents.AI.Hyperlight/Microsoft.Agents.AI.Hyperlight.csproj
```diff
@@ -0,0 +1,37 @@
...
+  </PropertyGroup>
+
+  <Import Project="$(RepoRoot)/dotnet/nuget/nuget-package.props" />
+  <PropertyGroup>
+    <!-- Disable packing until the HyperlightSandbox.Api NuGet is published and we are ready to release. -->
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.AI.Abstractions" />
```
westey-m, line 19: This is already included as a transient dependency from Microsoft.Agents.AI.Abstractions, so we should be able to remove this it.
```suggestion
```

dotnet/src/Microsoft.Agents.AI.Hyperlight/AllowedDomain.cs
```diff
@@ -0,0 +1,18 @@
...
+/// Hyperlight sandbox.
+/// </summary>
+/// <param name="Target">
+/// URL or domain to allow, for example <c>"https://api.github.com"</c>.
+/// </param>
+/// <param name="Methods">
+/// Optional list of HTTP methods to allow (for example <c>["GET", "POST"]</c>).
+/// When <see langword="null"/>, all methods supported by the backend are allowed.
+/// </param>
+public sealed record AllowedDomain(string Target, IReadOnlyList<string>? Methods = null);
```
westey-m, line 18: We generally avoid record types because of their cost, unless we are using their functionality.  They come with a lot of functionality by default, which is usually not used.
Since this is essentially just config, a class is probably fine.

dotnet/src/Microsoft.Agents.AI.Hyperlight/HyperlightCodeActProviderOptions.cs
```diff
@@ -0,0 +1,67 @@
...
+
+    /// <summary>
+    /// Gets or sets the initial set of file mount configurations.
+    /// </summary>
+    public IEnumerable<FileMount>? FileMounts { get; set; }
+
+    /// <summary>
+    /// Gets or sets the initial outbound network allow-list entries.
+    /// </summary>
+    public IEnumerable<AllowedDomain>? AllowedDomains { get; set; }
```
westey-m, line 66: Would this also work as a `Dictionary<string, IReadOnlyList<string>>`?
It'll avoid someone providing the same domain twice, but at the same it might be tricker to understand, i.e. you have to read the comments to know what `IReadOnlyList<string>` should contain.

dotnet/src/Microsoft.Agents.AI.Hyperlight/HyperlightCodeActProviderOptions.cs
```diff
@@ -0,0 +1,67 @@
...
+    /// Gets or sets the default approval mode for <c>execute_code</c>.
+    /// Defaults to <see cref="CodeActApprovalMode.NeverRequire"/>.
+    /// </summary>
+    public CodeActApprovalMode ApprovalMode { get; set; } = CodeActApprovalMode.NeverRequire;
+
+    /// <summary>
+    /// Gets or sets an optional workspace root directory on the host.
+    /// When set, it is exposed as the sandbox's <c>/input</c> directory.
+    /// </summary>
+    public string? WorkspaceRoot { get; set; }
```
westey-m, line 56: Is this term something that Hyperlight uses?  If it's a folder on the host that maps to an input folder in the sandbox, WorkspaceRoot isn't the first name that springs to mind 😄

dotnet/src/Microsoft.Agents.AI.Hyperlight/HyperlightCodeActProviderOptions.cs
```diff
@@ -0,0 +1,67 @@
...
+    /// Defaults to <see cref="SandboxBackend.Wasm"/>.
+    /// </summary>
+    public SandboxBackend Backend { get; set; } = SandboxBackend.Wasm;
+
+    /// <summary>
+    /// Gets or sets the path to the guest module (<c>.wasm</c> or <c>.aot</c> file).
+    /// Required for the <see cref="SandboxBackend.Wasm"/> backend; not needed for
+    /// <see cref="SandboxBackend.JavaScript"/>.
+    /// </summary>
+    public string? ModulePath { get; set; }
```
westey-m, line 26: Looks like this is optional only for javascript, but required for wasm.
Since we default to wasm though, it means that providing no options will always fail.
We should consider switching to javascript as the default.

Another option is to add some factory methods on the provider like:
`HyperlightCodeActProviderOptions.CreateForWasm(string modulePath, HyperlightCodeActProviderOptions? options = null);`
`HyperlightCodeActProviderOptions.CreateForJavaScript(HyperlightCodeActProviderOptions? options = null);`

And then we don't have backend or module path on the options.

dotnet/src/Microsoft.Agents.AI.Hyperlight/HyperlightCodeActProvider.cs
```diff
@@ -0,0 +1,387 @@
...
+        {
+            foreach (var name in names)
+            {
+                if (name is null)
+                {
+                    continue;
+                }
+
+                _ = this._tools.Remove(name);
+            }
```
westey-m, line 168: This can be shortened a little bit into the following, here and in other cases here.
```suggestion
            foreach (var name in names.Where(x => x is not null))
            {
                _ = this._tools.Remove(name);
            }
```

dotnet/src/Microsoft.Agents.AI.Hyperlight/HyperlightCodeActProvider.cs
```diff
@@ -0,0 +1,387 @@
...
+        }
+    }
+
+    /// <summary>Returns the current CodeAct-managed tools.</summary>
+    public IReadOnlyList<AIFunction> GetTools()
+    {
+        lock (this._gate)
+        {
+            return this._tools.Values.ToArray();
```
westey-m, line 149: Since we are returning a List, we might as well convert to a list instead of an array, which will then be converted to a list again.
```suggestion
            return this._tools.Values.ToList();
```

dotnet/src/Microsoft.Agents.AI.Hyperlight/HyperlightCodeActProvider.cs
```diff
@@ -0,0 +1,387 @@
...
+                this._allowedDomains[domain.Target] = domain;
+            }
+        }
+    }
+
+    /// <inheritdoc />
+    public override IReadOnlyList<string> StateKeys => s_stateKeys;
+
+    // -------------------------------------------------------------------
+    // Tool registry
```
westey-m, line 121: Do we need the public modification methods for tools, mounts, etc?
In other places, we generally only add this type of thing if there is a good user case.
And in that case, we probably could expose the properties publicly to allow the user to directly modify.  The only checks we do is for null values, and that we can do at consumption time as well.

dotnet/src/Microsoft.Agents.AI.Hyperlight/HyperlightCodeActProvider.cs
```diff
@@ -0,0 +1,387 @@
...
+    protected override ValueTask<AIContext> ProvideAIContextAsync(InvokingContext context, CancellationToken cancellationToken = default)
+    {
+        _ = Throw.IfNull(context);
+
+        SandboxExecutor.RunSnapshot snapshot;
+        lock (this._gate)
+        {
+            this.ThrowIfDisposed();
+            snapshot = new SandboxExecutor.RunSnapshot(
+                this._tools.Values.ToArray(),
```
westey-m, line 315: Are these explicitly requiring array as input?  Just double checking that we aren't unnecessarily converting to array.

dotnet/src/Microsoft.Agents.AI.Hyperlight/HyperlightCodeActProvider.cs
```diff
@@ -0,0 +1,387 @@
...
+        foreach (var tool in tools)
+        {
+            if (tool is ApprovalRequiredAIFunction)
+            {
+                return true;
+            }
+        }
+
+        return false;
+    }
```
westey-m, line 362: I generally don't comment on style, but since you are new to this, suggesting some shortcuts:
```suggestion
    internal static bool ComputeApprovalRequired(CodeActApprovalMode mode, IReadOnlyList<AIFunction> tools)
    => mode == CodeActApprovalMode.AlwaysRequire ?
        true :
        tools.Any(x => x is ApprovalRequiredAIFunction);
```

dotnet/src/Microsoft.Agents.AI.Hyperlight/Internal/ToolBridge.cs
```diff
@@ -0,0 +1,112 @@
...
+            case string s:
+                return JsonSerializer.Serialize(s);
+            case JsonElement element:
+                return element.GetRawText();
+            case JsonNode node:
+                return node.ToJsonString();
+            default:
+                return JsonSerializer.Serialize(result);
+        }
+    }
```
westey-m, line 95:
```suggestion
    private static string SerializeResult(object? result)
        => result switch
        {
            null => "null",
            string s => JsonSerializer.Serialize(s),
            JsonElement element => element.GetRawText(),
            JsonNode node => node.ToJsonString(),
            _ => JsonSerializer.Serialize(result)
        };
```

dotnet/src/Microsoft.Agents.AI.Hyperlight/Internal/ToolBridge.cs
```diff
@@ -0,0 +1,112 @@
...
+    /// tools directly and approval is enforced at the <c>execute_code</c>
+    /// layer.
+    /// </summary>
+    internal static AIFunction Unwrap(AIFunction tool)
+    {
+        while (tool is ApprovalRequiredAIFunction wrapper)
+        {
+            tool = wrapper.InnerFunction;
+        }
+
```
westey-m, line 109: I don't think this unwrapping is required.  ApprovalRequiredAIFunction is invocable just like a regular AIFunction.  It doesn't stop execution, and is essentially just metadata.  It's up to whatever uses it to ask for approval if it wants to respect the metadata.

dotnet/src/Microsoft.Agents.AI.Hyperlight/Internal/ToolBridge.cs
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    internal static IDictionary<string, object?> ParseArguments(string argsJson)
+    {
+        if (string.IsNullOrWhiteSpace(argsJson))
+        {
+            return new Dictionary<string, object?>(StringComparer.Ordinal);
+        }
+
+        var node = JsonNode.Parse(argsJson);
```
westey-m, line 63: Also consider deserializating into a dictionary directly, e.g.
```csharp
JsonSerializer.Deserialize<Dictionary<string, object?>>(argsJson);
```

westey-m, line 63: Do we need to support NativeAOT with this feature?

dotnet/src/Microsoft.Agents.AI.Hyperlight/Internal/ToolBridge.cs
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    private static string SerializeResult(object? result)
+    {
+        switch (result)
+        {
+            case null:
+                return "null";
+            case string s:
+                return JsonSerializer.Serialize(s);
```
westey-m, line 87: Do we need to serialize a string?  I suppose it'll just put it in quotes, but not sure if that is required.

dotnet/src/Microsoft.Agents.AI.Hyperlight/Internal/ToolBridge.cs
```diff
@@ -0,0 +1,112 @@
...
+            case JsonElement element:
+                return element.GetRawText();
+            case JsonNode node:
+                return node.ToJsonString();
+            default:
+                return JsonSerializer.Serialize(result);
```
westey-m, line 93: I think `JsonSerializer.Serialize(result);` should work for JsonElement and JsonNode as well, so you may not required the switch at all.

dotnet/samples/02-agents/AgentWithCodeAct/AgentWithCodeAct_Step01_Interpreter/README.md
```diff
@@ -0,0 +1,23 @@
...
+model writes Python and invokes the `execute_code` tool rather than answering
+from memory.
+
+## Configuration
+
+| Variable                       | Description                                                                               |
+|--------------------------------|-------------------------------------------------------------------------------------------|
+| `AZURE_OPENAI_ENDPOINT`        | Azure OpenAI endpoint. Required.                                                          |
+| `AZURE_OPENAI_DEPLOYMENT_NAME` | Azure OpenAI deployment. Defaults to `gpt-5.4-mini`.                                      |
+| `HYPERLIGHT_PYTHON_GUEST_PATH` | Absolute path to the Hyperlight Python guest module (`.wasm` or `.aot` file). Required.   |
```
Vijay-Nirmal, line 14: Details on how to get the guest module will go to capture or point to a website with the details, as most of the end users are new to Hyperligh

dotnet/samples/02-agents/AgentWithCodeAct/AgentWithCodeAct_Step02_ToolEnabled/README.md
```diff
@@ -0,0 +1,26 @@
+# AgentWithCodeAct_Step02_ToolEnabled
+
+Demonstrates adding provider-owned tools to `HyperlightCodeActProvider`. Those
+tools are **only** available to code running inside the sandbox via
+`call_tool("<name>", ...)` — they are never exposed to the model as direct
```
Vijay-Nirmal, line 5: Is Tools the recommended approach to pass files to the sandbox? Use case: Let the user upload an Excel file, and the agent uses the code to get details from the Excel file, similar to how Copilot works (Copilot studio). A sample to showcase this use case will be good. Also, a sample to showcase file output from the execute_code tools will be good

dotnet/src/Microsoft.Agents.AI.Hyperlight/Internal/ToolBridge.cs
eavanvalkenburg: Yes — the package inherits `IsAotCompatible=true` from `Directory.Build.props`, so the bridge is wired through `AIJsonUtilities` / `HyperlightJsonContext` to keep the IL2026/IL3050 analyzers happy.

dotnet/src/Microsoft.Agents.AI.Hyperlight/HyperlightCodeActProviderOptions.cs
eavanvalkenburg: Done in 8d2cba8b — added `CreateForWasm(string modulePath)` and `CreateForJavaScript()` factories, and switched the parameterless constructor / default to JavaScript so `new HyperlightCodeActProviderOptions()` is a working configuration. `Backend` and `ModulePath` are now get-only and only set via the factories.

dotnet/src/Microsoft.Agents.AI.Hyperlight/HyperlightCodeActProviderOptions.cs
eavanvalkenburg: Done in 8d2cba8b — renamed to `HostInputDirectory` (and the doc comment now spells out it maps to `/input` inside the sandbox).

dotnet/src/Microsoft.Agents.AI.Hyperlight/Microsoft.Agents.AI.Hyperlight.csproj
eavanvalkenburg: Done in 8d2cba8b — removed the redundant `Microsoft.Extensions.AI.Abstractions` `PackageReference`.

dotnet/src/Microsoft.Agents.AI.Hyperlight/HyperlightCodeActProvider.cs
eavanvalkenburg: Done in 8d2cba8b — `ComputeApprovalRequired` is now a single expression-bodied method using `Any(t => t is ApprovalRequiredAIFunction)`.

dotnet/src/Microsoft.Agents.AI.Hyperlight/HyperlightCodeActProvider.cs
eavanvalkenburg: Done in 8d2cba8b — `GetTools()` now returns `this._tools.Values.ToList()` directly, no intermediate array.

dotnet/src/Microsoft.Agents.AI.Hyperlight/HyperlightCodeActProviderOptions.cs
eavanvalkenburg: Considered, but kept `AllowedDomain` as a typed config object: it's self-documenting (`Domain`/`Methods` are obvious from the type), matches the Hyperlight SDK's own `AllowedHostsConfig` shape, and keeps the door open for future per-domain options (TLS pinning, headers, etc.) without a breaking change. The duplicate-domain risk is small since this is configuration, not user input.

dotnet/src/Microsoft.Agents.AI.Hyperlight/Internal/ToolBridge.cs
eavanvalkenburg: Adopted the spirit of the suggestion but landed on `AIJsonUtilities.DefaultOptions.GetTypeInfo(result.GetType())` so arbitrary tool-result types are routed through the framework's AOT-friendly resolver chain (matches what Mem0/A2A/AGUI do). That single call already covers `string`, `JsonElement`, and `JsonNode` correctly, so the explicit switch isn't needed.

dotnet/tests/Microsoft.Agents.AI.Hyperlight.UnitTests/Microsoft.Agents.AI.Hyperlight.UnitTests.csproj
```diff
@@ -0,0 +1,12 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0'))">
+    <JsonSerializerIsReflectionEnabledByDefault>false</JsonSerializerIsReflectionEnabledByDefault>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\Microsoft.Agents.AI.Hyperlight\Microsoft.Agents.AI.Hyperlight.csproj" />
+    <ProjectReference Include="..\..\src\Microsoft.Agents.AI\Microsoft.Agents.AI.csproj" />
+  </ItemGroup>
```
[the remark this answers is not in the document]
eavanvalkenburg: Done in 8d2cba8b — both test `csproj`s now override `<TargetFrameworks>` to `net8.0;net9.0;net10.0` to match the package.

dotnet/src/Microsoft.Agents.AI.Hyperlight/Internal/InstructionBuilder.cs
```diff
@@ -0,0 +1,121 @@
...
+        if (toolsVisibleToModel)
+        {
+            return
+                "You can execute Python code in a secure sandbox by calling the `execute_code` tool. "
+                + "Use it for calculations, data analysis, and anything that benefits from running code. "
+                + "State does not persist between calls; pass any required values in the code you execute.";
+        }
+
+        return
+            "You can execute Python code in a secure sandbox by calling the `execute_code` tool. "
```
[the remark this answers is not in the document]
eavanvalkenburg: Done in 8d2cba8b — `InstructionBuilder` now branches on `SandboxBackend` and emits the appropriate language/runtime text (`Python` for Wasm, `JavaScript (QuickJS)` for JS), with matching test coverage in `InstructionBuilderTests`.

dotnet/src/Microsoft.Agents.AI.Hyperlight/HyperlightCodeActProvider.cs
eavanvalkenburg: Kept the `Add/Remove` CRUD on the provider (mirrors the Python implementation that the .NET surface tracks), but they are now no-ops at the sandbox layer: each invocation snapshots the current registries / allow-lists, fingerprints them, and rebuilds the sandbox only when the fingerprint changes (see `SandboxExecutor.ExecuteAsync`). That covers the staleness concern Copilot raised and keeps the user-facing API ergonomic for the dynamic-tool scenario.

dotnet/src/Microsoft.Agents.AI.Hyperlight/HyperlightCodeActProvider.cs
eavanvalkenburg: Done in 8d2cba8b — the snapshots are passed as `IReadOnlyList<…>` end-to-end now; the only `ToArray()` left is the one Hyperlight's `AllowedHostsConfig` constructor requires.

dotnet/src/Microsoft.Agents.AI.Hyperlight/AllowedDomain.cs
eavanvalkenburg: Done in 8d2cba8b — converted `AllowedDomain` and `FileMount` from `record` to `sealed class`.

dotnet/src/Microsoft.Agents.AI.Hyperlight/HyperlightExecuteCodeFunction.cs
```diff
@@ -0,0 +1,109 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+using System.Collections.Generic;
```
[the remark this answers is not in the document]
eavanvalkenburg: Done in 8d2cba8b — removed the unused using.

dotnet/src/Microsoft.Agents.AI.Hyperlight/Internal/ToolBridge.cs
eavanvalkenburg: Yes — JSON requires the surrounding quotes (and any internal escaping). Going through `JsonSerializer.Serialize` for `string` produces a valid JSON string literal for the guest to parse; emitting the raw `string` would be invalid JSON.

dotnet/samples/02-agents/AgentWithCodeAct/AgentWithCodeAct_Step02_ToolEnabled/README.md
eavanvalkenburg: `Tools` (the in-sandbox `call_tool` bridge) is for invoking host code from the guest, not for shipping arbitrary user files into the sandbox — for the Excel-upload flow, `HostInputDirectory` / `FileMounts` is the recommended path (the file becomes available under `/input` and the model can `open()` it from generated code). Tracking an end-to-end `AgentWithCodeAct_Step04_FileUpload` (Excel/CSV) sample as a follow-up.

dotnet/src/Microsoft.Agents.AI.Hyperlight/Internal/InstructionBuilder.cs
```diff
@@ -0,0 +1,121 @@
...
+
+            foreach (var mount in fileMounts)
+            {
+                sb.AppendLine(
+                    string.Format(
+                        CultureInfo.InvariantCulture,
+                        "- `{0}` (host: `{1}`)",
+                        mount.MountPath,
+                        mount.HostPath));
+            }
```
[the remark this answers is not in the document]
eavanvalkenburg: Done in 8d2cba8b — the model-visible `execute_code` description no longer mentions `HostPath` or `HostInputDirectory`; only the in-sandbox `/input` and `GuestPath` values surface to the model. Host paths remain in operator-facing logs/diagnostics only.

dotnet/tests/Microsoft.Agents.AI.Hyperlight.IntegrationTests/Microsoft.Agents.AI.Hyperlight.IntegrationTests.csproj
```diff
@@ -0,0 +1,7 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\Microsoft.Agents.AI.Hyperlight\Microsoft.Agents.AI.Hyperlight.csproj" />
+  </ItemGroup>
+
```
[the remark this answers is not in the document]
eavanvalkenburg: Done in 8d2cba8b — same fix as the unit-tests project: `<TargetFrameworks>` overridden to drop `net472`.

dotnet/src/Microsoft.Agents.AI.Hyperlight/Internal/SandboxExecutor.cs
```diff
@@ -0,0 +1,182 @@
...
+        {
+            this._executionLock.Release();
+        }
+    }
+
+    private void EnsureInitialized(RunSnapshot snapshot)
+    {
+        if (this._sandbox is not null)
+        {
+            return;
```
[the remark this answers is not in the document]
eavanvalkenburg: Addressed in 8d2cba8b — `SandboxExecutor` now fingerprints (`HeapSize`/`StackSize`/`ModulePath`/`Backend`/`HostInputDirectory`/`FileMounts`/`AllowedDomains`/tool-names) on every `ExecuteAsync` and rebuilds the sandbox when the fingerprint changes, so post-construction registry/allow-list edits are picked up.

dotnet/src/Microsoft.Agents.AI.Hyperlight/Internal/ToolBridge.cs
eavanvalkenburg: Done in 8d2cba8b — `SerializeResult` no longer has a switch. Single call: `JsonSerializer.Serialize(result, AIJsonUtilities.DefaultOptions.GetTypeInfo(result.GetType()))`, which handles `JsonElement`/`JsonNode`/`string`/POCOs uniformly and stays AOT-safe.

dotnet/src/Microsoft.Agents.AI.Hyperlight/Internal/ExecuteCodeFunction.cs
```diff
@@ -0,0 +1,85 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
```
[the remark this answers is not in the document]
eavanvalkenburg: Done in 8d2cba8b — removed the unused usings.

dotnet/samples/02-agents/AgentWithCodeAct/AgentWithCodeAct_Step01_Interpreter/README.md
eavanvalkenburg: Step01 README now points at the upstream Hyperlight quickstart for obtaining/building a guest module and links into the sample's prerequisites; longer-form guidance lives in the Hyperlight repo so it stays up to date with their tooling. Tracking richer, click-through guest-module setup as a follow-up sample.

dotnet/src/Microsoft.Agents.AI.Hyperlight/Internal/ToolBridge.cs
eavanvalkenburg: Done in 8d2cba8b — the `ApprovalRequiredAIFunction` unwrap is gone; we just invoke the `AIFunction` directly.

dotnet/src/Microsoft.Agents.AI.Hyperlight/Internal/ToolBridge.cs
eavanvalkenburg: Considered, but kept `JsonNode.Parse` to stay NativeAOT-safe — the typed `Deserialize<Dictionary<string, object?>>` overload requires reflection-based metadata for the `object` value and trips IL2026/IL3050 on a project that has `IsAotCompatible=true`. Added a comment in `ParseArguments` explaining the choice.

dotnet/src/Microsoft.Agents.AI.Hyperlight/HyperlightCodeActProvider.cs
eavanvalkenburg: Done in 8d2cba8b — replaced the manual null filtering with `.Where(x => x is not null)` in the snapshot loops.

dotnet/src/Microsoft.Agents.AI.Hyperlight/HyperlightExecuteCodeFunction.cs
```diff
@@ -0,0 +1,102 @@
...
+    /// <summary>
+    /// Returns the <c>execute_code</c> function for direct registration on an agent.
+    /// When approval is required the returned function is wrapped in
+    /// <see cref="ApprovalRequiredAIFunction"/>.
+    /// </summary>
+    public AIFunction AsAIFunction()
+    {
+        this.ThrowIfDisposed();
+        return this._function;
+    }
```
westey-m: Consider just having this class inherit from AIFunction, so you don't need to call AsAIFunction before using it.

dotnet/src/Microsoft.Agents.AI.Hyperlight/Microsoft.Agents.AI.Hyperlight.csproj
```diff
@@ -0,0 +1,36 @@
...
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <InjectSharedThrow>true</InjectSharedThrow>
+  </PropertyGroup>
+
+  <Import Project="$(RepoRoot)/dotnet/nuget/nuget-package.props" />
+  <PropertyGroup>
+    <!-- Re-enable packing once we are ready to release. -->
+    <IsPackable>false</IsPackable>
```
westey-m: Not sure if we are ready to release this, but if so, remember to remove this.

dotnet/src/Microsoft.Agents.AI.Hyperlight/README.md
```diff
@@ -0,0 +1,41 @@
+﻿# Microsoft.Agents.AI.Hyperlight
```
westey-m, line 1: If we want this to be the main nuget .md file, we need to reference it from the hyperlight.csproj to override the PackageReadmeFile.  

We also need to include it into the packaged files, e.g.
```xml
  <PropertyGroup>
    <PackageReadmeFile>README.md</PackageReadmeFile>
  </PropertyGroup>

  <ItemGroup>
    <None Include="README.md" Pack="true" PackagePath="/" />
  </ItemGroup>
```

See Aspire.Hosting.AgentFramework.DevUI for an example.

dotnet/src/Microsoft.Agents.AI.Hyperlight/HyperlightCodeActProvider.cs
```diff
@@ -0,0 +1,324 @@
...
+            Instructions = instructions,
+            Tools = [executeCode],
+        };
+
+        return new ValueTask<AIContext>(result);
+    }
+
+    internal static bool ComputeApprovalRequired(CodeActApprovalMode mode, IReadOnlyList<AIFunction> tools) =>
+        mode == CodeActApprovalMode.AlwaysRequire
+            || tools.Any(t => t is ApprovalRequiredAIFunction);
```
westey-m: Since a tool can be decorated by multiple classes, and `t is ApprovalRequiredAIFunction` only considers the type of the top-most class in the stack, it's good to use GetService to test for the existence of `ApprovalRequiredAIFunction` anywhere in the AITool stack.
```suggestion
            || tools.Any(t => t.GetService<ApprovalRequiredAIFunction>() is not null);
```

westey-m (approved):

dotnet/src/Microsoft.Agents.AI.Hyperlight/Microsoft.Agents.AI.Hyperlight.csproj
eavanvalkenburg: Done in 5270c37f — removed `IsPackable=false`. The package is ready to release now that `Hyperlight.HyperlightSandbox.Api 0.4.0` is published.

dotnet/src/Microsoft.Agents.AI.Hyperlight/HyperlightExecuteCodeFunction.cs
eavanvalkenburg: Done in 5270c37f — `HyperlightExecuteCodeFunction` now inherits `AIFunction` directly. `AsAIFunction()` is gone, instances are passed wherever an `AIFunction` is expected. To preserve the `ApprovalMode` plumbing, `GetService(typeof(ApprovalRequiredAIFunction))` lazily returns a wrapping `ApprovalRequiredAIFunction` proxy when the effective configuration requires approval — that's how the rest of the framework discovers the requirement.

dotnet/src/Microsoft.Agents.AI.Hyperlight/README.md
eavanvalkenburg: Done in 5270c37f — added `<PackageReadmeFile>README.md</PackageReadmeFile>` and `<None Include=\"README.md\" Pack=\"true\" PackagePath=\"/\" />` (matches the `Aspire.Hosting.AgentFramework.DevUI` / `Microsoft.Agents.AI.DurableTask` pattern).

dotnet/src/Microsoft.Agents.AI.Hyperlight/HyperlightCodeActProvider.cs
eavanvalkenburg: Done in 5270c37f — switched to `tools.Any(t => t.GetService<ApprovalRequiredAIFunction>() is not null)` so we walk the full AITool decorator stack rather than just checking the top-most type.

dotnet/samples/02-agents/AgentWithCodeAct/AgentWithCodeAct_Step01_Interpreter/Program.cs
```diff
@@ -0,0 +1,30 @@
...
+
+using var codeAct = new HyperlightCodeActProvider(HyperlightCodeActProviderOptions.CreateForWasm(guestPath));
+
+AIAgent agent = new AzureOpenAIClient(
+    new Uri(endpoint),
+    new DefaultAzureCredential())
+    .GetChatClient(deploymentName)
+    .AsAIAgent(new ChatClientAgentOptions()
+    {
+        ChatOptions = new() { Instructions = "You are a helpful assistant. When the user asks something quantitative, write Python and call `execute_code` instead of guessing." },
```
SergeyMenshykh, line 25: Is it necessary to mention `execute_code` in the instructions? Shouldn't the provider handle it itself? It seems that users should be aware of the implementation details of the HyperlightCodeActProvider. The same applies to the other samples in this PR

SergeyMenshykh (approved):

dotnet/samples/02-agents/AgentWithCodeAct/AgentWithCodeAct_Step01_Interpreter/Program.cs
eavanvalkenburg: Not strictly needed, more for clarity on the purpose of the sample.

PR #5363, opened 2026-04-20
.NET: Migrate A2A hosting to A2A SDK v1

SergeyMenshykh:
Migrates the A2A hosting layer (`Microsoft.Agents.AI.Hosting.A2A` and `Microsoft.Agents.AI.Hosting.A2A.AspNetCore`) to A2A SDK v1.

### Changes

- **New `A2AAgentHandler`** — extracted A2A agent handling logic into a dedicated class.
- **New `A2AProtocolBinding` enum** — `[Flags]` enum for selecting which A2A protocol bindings to expose (JSON-RPC, HTTP+JSON/REST).
- **New `A2AHostingOptions`** — consolidates hosting configuration (protocol bindings, agent run mode, server options).
- **New `A2AEndpointRouteBuilderExtensions`** — replaces `EndpointRouteBuilderExtensions` with cleaner overloads accepting either direct parameters or an `Action<A2AHostingOptions>` callback.
- **Simplified `AIAgentExtensions`** — moved endpoint-mapping logic out; retains only the `MapA2A` adapter method.


dotnet/samples/05-end-to-end/AgentWebChat/AgentWebChat.AgentHost/Program.cs
```diff
@@ -156,14 +157,12 @@ Once the user has deduced what type (knight or knave) both Alice and Bob are, te
 
 // attach a2a with simple message communication
 app.MapA2A(pirateAgentBuilder, path: "/a2a/pirate");
-app.MapA2A(knightsKnavesAgentBuilder, path: "/a2a/knights-and-knaves", agentCard: new()
+app.MapA2A(knightsKnavesAgentBuilder, path: "/a2a/knights-and-knaves");
+app.MapWellKnownAgentCard(new AgentCard
```
westey-m: How does the system know that this card matches the `/a2a/knights-and-knaves` path rather than the `/a2a/pirate` path?

westey-m (approved):

dotnet/samples/05-end-to-end/AgentWebChat/AgentWebChat.AgentHost/Program.cs
SergeyMenshykh: In its current state, the A2A spec does not allow an agent card per agent; it only allows an agent card per host. The code that previously passed an agent card relied on an unofficial `/card` endpoint, which is currently being removed from the A2A SDK. The `MapWellKnownAgentCard` is actually unnecessary since agents are resolved by direct agent url rather than via an agent card. So, I removed it.

PR #4083, opened 2026-02-19
Python: fix reasoning model workflow handoff and history serialization

eavanvalkenburg:
## Summary

Fixes multiple related failures when using reasoning models (gpt-5-mini, gpt-5.2) in multi-agent workflows. The root issues are all about how reasoning items from the Responses API are emitted, serialized, and carried into subsequent agent runs.

Closes #4047

---

### 1. "reasoning was provided without its required following item"

The Responses API only accepts a `reasoning` item in `input` when it directly precedes a `function_call`. Sending a reasoning item that preceded a text response (no tool call) causes an API error.

**Fix**: `_prepare_message_for_openai` now checks whether the message contains a `function_call`. `text_reasoning` content is only serialized as a `reasoning` input item when a `function_call` is also present in the same message.

### 2. Reasoning items never emitted for encrypted/hidden reasoning

When a reasoning model produces encrypted or hidden reasoning, the `output_item.added` event fires with an empty `content` list and no `reasoning_text.delta` events follow. Previously, no `text_reasoning` Content was emitted — making it invisible to downstream serialization logic.

**Fix**: Both `_parse_response_from_openai` (non-streaming) and the `output_item.added` handler (streaming) now always emit at least one `text_reasoning` Content, even when the text is empty. The `reasoning_id` and `encrypted_content` (if present) are stored in `additional_properties`.

### 3. `summary` field must be an array, not an object

The `summary` field on a `reasoning` input item must be an array of objects (`[{"type": "summary_text", "text": ...}]`), not a single object. This caused a 400 `invalid_type` error.

**Fix**: `_prepare_content_for_openai` now wraps `summary` in a list. `summary` is omitted entirely when there is no visible text (e.g. encrypted reasoning, where only `encrypted_content` is sent).

---

## Files Changed

| File | Change |
|------|--------|
| `packages/core/agent_framework/openai/_responses_client.py` | Always emit `text_reasoning` on reasoning output items; fix `summary` to be an array; skip reasoning serialization when no `function_call` in same message |
| `packages/core/agent_framework/_workflows/_agent_executor.py` | Clear `service_session_id` in `run` and `from_response` handlers; remove no-op `_prepare_handoff_messages` |
| `packages/core/tests/workflow/test_full_conversation.py` | Add `test_run_request_with_full_history_clears_service_session_id` and `test_from_response_clears_service_session_id` (TDD: fail without fix, pass with fix) |

PR #5369, opened 2026-04-20
.NET: Guard SharedStateAgent empty-streams with state-shape check

jpr5:
When ChatOptions.AdditionalProperties contains ag_ui_state but the payload is empty `{}` or lacks schema-matching content, the two-pass structured-output flow's TryDeserialize fails and the method emits zero content updates. UI sees a completely silent response on agentic-chat demos that forward shared state even when empty.

Guard RunCoreStreamingAsync: when ag_ui_state is absent, non-object, or an empty object, pass through to the inner agent instead of taking the schema-constrained path. Mirror the fix pattern already in the sibling Step05_StateManagement sample.

Helper HasAnyProperties includes a defensive ValueKind != Object early return so it remains safe under future callers that drop the caller-side guard.

Verified via standalone xUnit repro (3 failing -> 3 passing on empty `{}`, non-object state, missing key; plus HasAnyProperties unit tests across Array / String / Number / Null / Undefined / Bool / EmptyObject / NonEmptyObject). Test harness excluded from repo (net10.0 SDK installed locally in .dotnet/, not upstream).

Related: downstream CopilotKit PR #4083 commit 9227bc27d worked around this by adding the guard to showcase/packages/ms-agent-dotnet/agent/SharedStateAgent.cs.


jpr5:
@microsoft-github-policy-service agree company="CopilotKit"

PR #5409, opened 2026-04-21
Python: fix(devui): flush trailing assistant text delta

mrfelton:
## Summary

This fixes a DevUI streaming edge case where the visible assistant message can miss its trailing text until a later forced refresh, even though the final `output_text.delta` already arrived in the browser stream.

## Root cause

DevUI currently throttles assistant text paints to a 50 ms interval to avoid unbounded rerender/memory growth. That throttle is useful, but the current implementation drops out early when a non-forced render lands inside the interval and does not schedule a trailing flush.

That means the last `response.output_text.delta` in a burst can remain buffered but invisible until some later forced render happens, such as another `response.output_item.added` event or stream completion.

## Fix

- keep the existing 50 ms throttle
- track a pending assistant render while throttled
- schedule a single delayed flush for the remaining interval instead of silently dropping the paint
- clear any pending delayed flush when a forced render or immediate flush happens

This preserves the memory/rerender protection added by the throttle while ensuring the last visible assistant text catches up promptly.

## Regression coverage

Added a browser regression in `tests/devui/test_ui_memory_regression.py` that:

- streams one assistant chunk
- streams a second chunk 10 ms later
- keeps the stream open for 0.5 s
- asserts the full text becomes visible before the stream ends and while the stop button is still visible

That reproduces the missing-trailing-flush shape directly.

## Validation

Validated locally in `python/packages/devui` with:

- `npm run build` in `frontend/`
- `uv run pytest tests/devui/test_ui_memory_regression.py -q`

Both passed after the change.

## Downstream evidence

I also validated the patched bundle against a real .NET-hosted DevUI integration.

- Patched bundle confirmed loaded with a runtime sentinel.
- In-page monitor recorded the last `output_text.delta` reaching the visible assistant message in about `194 ms`.
- The same integration still kept the turn open after the last visible delta, so this PR is intentionally scoped to the DevUI text-flush bug only, not broader turn-completion timing.

That downstream check is what motivated this fix: before the patch, the local DevUI client could hold back the last visible summary text for a long time even though the browser had already received it.


mrfelton:

moonbox3 (commented):
## Automated Code Review

**Reviewers:** 4 | **Confidence:** 85%

### ✓ Correctness
> Chunk 1 contains only the rebuilt minified JavaScript bundle (index.js). All differences are variable name remappings produced by the bundler/minifier (e.g., XE→ZE, $→P, He→Pe, on→an, Sn→_n, tn→sn, etc.). These are expected artifacts of a frontend rebuild and contain no semantic or correctness issues. The actual source code changes will be in subsequent chunks.; This chunk refactors the assistant text streaming throttle from a drop-on-throttle pattern to a defer-on-throttle pattern. The old `renderAssistantStreamingText` is split into `flushAssistantStreamingText` (the render logic) and a new `renderAssistantStreamingText` wrapper that schedules deferred renders via `setTimeout` when throttled. Cleanup effects clear pending timers on unmount and when streaming stops. The change from `else if (force)` to `else` in the flush function is correct because `flushAssistantStreamingText` is now only called when a render is truly intended. Dependency arrays are correct (refs don't need inclusion). The `force=true` path properly bypasses throttling and clears pending deferred renders. The `isStreaming` cleanup effect addresses the previously noted concern about stale timers firing after state resets. The minified index.js changes mirror the source changes. No correctness issues found.; Chunk 3 adds a new `TailFlushAgent`, a `tail_flush_server` fixture, and a `test_devui_flushes_trailing_text_delta_before_stream_end` test to verify that the DevUI renders trailing text deltas before the stream completes. The test logic is sound: it sends a message, waits for the expected concatenated text to appear in the DOM within a 1.0s window while the stream is held open for 0.5s, then asserts the stop button is still visible (proving text was flushed before stream end). The agent implementation follows the established `MemoryStressAgent` pattern correctly. The formatting changes to existing JavaScript expressions are cosmetic and equivalent. The resolved review comment about `_TAIL_VISIBILITY_TIMEOUT_S` being too small has been addressed (raised to 1.0s). No correctness issues found.

### ✓ Security Reliability
> Chunk 1 contains only changes to the minified/bundled index.js build artifact. All visible diffs are mechanical variable-name changes produced by the JavaScript bundler/minifier (e.g., XE→ZE, $→P, ee→J, He→Pe, on→an). No semantic logic changes, no new code paths, and no security or reliability concerns are present in this chunk. The oversized sections that were omitted from the prompt were also index.js rebuild diffs. Source-level changes (agent-view.tsx, test_ui_memory_regression.py) will appear in chunks 2 and 3.; Chunk 2 covers the refactored assistant-text throttle in agent-view.tsx and the corresponding minified bundle update. The new throttle mechanism correctly schedules a deferred flush when renders are rate-limited, with proper cleanup on unmount and when streaming stops (addressing the previously resolved review comment). The timeout callback safely captures the pending ref before clearing it. No security or reliability issues found in this chunk.; Chunk 3 adds a TailFlushAgent test agent, a tail_flush_server fixture, and a new test that validates DevUI flushes trailing text deltas before stream end. The code follows established patterns already present in the file (MemoryStressAgent, memory_regression_server fixture). json is imported (line 10), all types are correctly used, resource cleanup is handled via try/finally and fixture teardown, and json.dumps is used safely for embedding strings in JS expressions. The _TAIL_VISIBILITY_TIMEOUT_S was already corrected to 1.0s per the resolved review comment. No security or reliability issues found.

### ✓ Test Coverage
> Chunk 1 contains only the rebuilt minified bundle (index.js) with variable-name changes from the Vite/Rollup rebuild. No logic changes exist in this chunk — the actual behavioral changes are in agent-view.tsx and streaming-state.ts (chunks 2/3). The minified output itself is not meaningfully reviewable for test coverage. The PR does include a thorough E2E memory regression test (test_ui_memory_regression.py) and unit tests for process-row parsing helpers. From a test coverage perspective this chunk raises no concerns — the bundle is a build artifact reflecting the source changes tested elsewhere.; The refactored streaming text throttle in agent-view.tsx (splitting into flushAssistantStreamingText + timer-based renderAssistantStreamingText wrapper with cleanup effects) is a non-trivial timing-sensitive change with no unit-level test coverage. The project has no frontend unit test infrastructure (no vitest/jest, no .test.tsx files, no test script in package.json), so this is a pre-existing gap rather than an omission specific to this PR. The new Python-based test_ui_memory_regression.py (in another chunk) provides end-to-end memory regression coverage via CDP, which validates the throttle prevents runaway memory growth, but does not test the functional correctness of the throttle logic (e.g., timer scheduling, pending render flush accuracy, cleanup on isStreaming transition). The two resolved PR comments (clearing pending timer when isStreaming becomes false, and increasing _TAIL_VISIBILITY_TIMEOUT_S) appear addressed in the diff — the useEffect cleanup for isStreaming is present at line 429. No blocking issues found given the project's current test posture, but the throttle mechanism would benefit from unit tests if a frontend test runner is ever introduced.; The new tail-flush regression test is well-structured and provides good coverage for the specific rendering bug (trailing text delta hidden until stream ends). The TailFlushAgent simulates the scenario accurately, the fixture follows established patterns, and the dual assertion (text visible + stop button present) correctly validates the behavior. Timing constants have been improved per prior review. One minor concern: _TAIL_STREAM_HOLD_S (0.5s) is less than _TAIL_VISIBILITY_TIMEOUT_S (1.0s), which means the visibility poll can succeed after the stream ends, causing a misleading stop-button assertion failure on slow CI — but the practical risk is low given near-instant frontend rendering and 100ms polling granularity.

### ✓ Design Approach
> I did not find a design-level issue in this chunk. The substantive source change in `agent-view.tsx` aligns with the existing store contract, which already treats agent switches as a full conversation-state reset boundary, so clearing streaming/chat state there addresses the right lifecycle edge rather than introducing a new abstraction leak. The `index.js` hunks in this chunk appear to be generated bundle symbol churn from rebuilding the frontend, not a separate design change.; I did not find a design-approach issue in this chunk. The refactor moves the throttle logic from dropping intermediate assistant-text renders to coalescing them into a single deferred flush, which addresses the underlying stale-tail problem without changing the higher-level streaming contract. The added cleanup on unmount and when streaming stops also lines up with the surrounding reset flows I inspected, so I do not see a remaining fundamental flaw in this approach from the code shown here.; I did not find a design-level problem in this chunk. The added synthetic agent and browser regression test are aimed at the right behavioral contract: ensuring the UI flushes a throttled trailing text delta before streaming actually ends. The only caveat is that the new DOM assertion is somewhat coupled to the current markdown wrapper implementation, which makes the test a bit more brittle than the behavior it is trying to protect.

### Suggestions
- The E2E memory regression test validates memory bounds but not functional correctness of the throttle — e.g., it wouldn't catch a bug where the pending render timer flushes stale data. Consider adding an assertion that the final rendered message content matches the expected streamed output.

---

eavanvalkenburg:
Closing, not sure if this is still relevant, feel free to reopen if so.

PR #5419, opened 2026-04-22
Automated issue triage workflow

moonbox3:
### Motivation and Context

GitHub workflow to automate initial bug reports from issues.

### Contribution Checklist


moonbox3 (commented):
## Automated Code Review

### ✓ Correctness
> The workflow is structurally sound for normal operation. The three issues from the prior review thread (secret exposure on any opened issue, concurrency group missing workflow_dispatch input, and catch-all error handling in team membership) remain unresolved in the current diff and are the most significant concerns. One new minor finding: the issue-author lookup via the REST API on line 75 does not guard against a null `user` field (which GitHub returns for deleted accounts), unlike line 68 which uses optional chaining.

### ✓ Security Reliability
> The three main security concerns (unrestricted trigger on `issues: opened`, concurrency group missing `inputs.issue_number`, and fail-open on team membership API errors) were already identified in the existing unresolved review thread. Beyond those, the workflow follows good practices: issue number input is regex-validated (line 51), secrets are passed through environment variables rather than expression interpolation (avoiding script injection), and checkouts use `persist-credentials: false`. No new high-confidence security or reliability issues were found.

### ✓ Test Coverage
> This PR adds a new GitHub Actions workflow for issue triage. From a test coverage perspective, the workflow embeds non-trivial team-membership-check logic as inline JavaScript (lines 67–96) rather than extracting it into a testable script under `.github/scripts/`, which is the established pattern in this repo (see `stale_issue_pr_ping.py` with its corresponding `test_stale_issue_pr_ping.py`). The inline JS includes fallback author resolution, API error handling, and membership state checking — all paths that would benefit from unit tests but currently cannot be tested. The core triage scripts (`classify_issue_spam.py`, `trigger_issue_repro.py`) live in an external private repo and are outside the scope of this PR. No blocking test-coverage issues, but the untested inline logic is a notable gap given existing repo conventions.

### ✓ Design Approach
> The workflow is structurally sound: minimal permissions, input validation, credential hygiene (`persist-credentials: false`), and a two-job gate that skips triage for team members. The three issues raised in the prior review thread (trigger scope allowing any external issue to burn secrets, concurrency group not covering `workflow_dispatch` inputs, and fail-open error handling in team membership) remain unresolved in this diff and are the most significant design concerns. Beyond those, no new blocking design-approach issues were identified.

### Suggestions
- Line 75: `issue.user.login` will throw a `TypeError` for issues opened by deleted GitHub accounts (`user` is `null`). Use optional chaining (`issue.user?.login`) or a null guard, consistent with line 68's `context.payload.issue?.user?.login`.
- Extract the inline JavaScript team-membership-check logic (lines 67–96) into a dedicated script under `.github/scripts/` (e.g., `check_team_membership.js`) with unit tests under `.github/tests/`, following the established `stale_issue_pr_ping.py` / `test_stale_issue_pr_ping.py` pattern. This would allow testing the fallback author-resolution, active-membership distinction, and error-handling branches.
- The 'Stop after spam gate' step (line 155) is a misleading no-op—it logs but doesn't halt execution. The real gating relies on per-step `if` conditions, so any future step added without its own guard will silently bypass the spam gate. Consider failing the job when spam is detected (labeling is already applied) or restructuring into a separate job so the decision is enforced at the job-dependency level.

---

.github/workflows/issue-triage.yml
```diff
@@ -0,0 +1,178 @@
...
+          github-token: ${{ secrets.GH_ACTIONS_PR_WRITE }}
+          script: |
+            let author = context.payload.issue?.user?.login;
+            if (!author) {
+              const { data: issue } = await github.rest.issues.get({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                issue_number: Number(process.env.ISSUE_NUMBER),
+              });
+              author = issue.user.login;
```
moonbox3: `issue.user` can be `null` for issues created by deleted GitHub accounts, which would throw a `TypeError` here and crash the `team_check` job (preventing triage entirely). Line 68 already uses optional chaining for the same field; this path should be similarly defensive.

```suggestion
              author = issue.user?.login;
```

.github/workflows/issue-triage.yml
```diff
@@ -0,0 +1,178 @@
...
+              console.log(`Team membership lookup failed for ${author}: ${error.message}`);
+              isTeamMember = false;
+            }
+
+            core.setOutput('is_team_member', isTeamMember ? 'true' : 'false');
+            if (isTeamMember) {
+              core.info(`Author ${author} is a team member; skipping auto-triage.`);
+            } else {
+              core.info(`Author ${author} is not a team member; proceeding with triage.`);
+            }
```
moonbox3, line 85: **Test coverage gap:** This inline `github-script` block (lines 67–96) contains three distinct code paths—(1) author from event payload, (2) author fetched via API when payload is absent, and (3) error-catch defaulting to non-member—none of which have unit tests. The repo already extracts workflow logic into `.github/scripts/` with tests in `.github/tests/` (see `stale_issue_pr_ping.py`). Consider extracting this into a testable script so the fallback-resolution and error-handling branches can be covered.

.github/workflows/issue-triage.yml
```diff
@@ -0,0 +1,178 @@
...
+            --repo-path "${TARGET_REPO_PATH}" \
+            --apply-labels
+
+      - name: Stop after spam gate
+        if: ${{ steps.spam.outputs.decision != 'allow' }}
+        shell: bash
+        env:
+          SPAM_DECISION: ${{ steps.spam.outputs.decision }}
+        run: |
+          echo "Skipping reproduction because spam gate decided: ${SPAM_DECISION}"
```
moonbox3: This step is named "Stop after spam gate" but doesn't actually stop anything—it just logs. The real gate is the `if` on line 164. This is fragile: any future step added after this one will run for spam issues unless it also carries its own `if` guard. Consider making this a true gate by exiting non-zero to fail the job (labeling has already been applied by the classify step).

```suggestion
          echo "Stopping: spam gate decided: ${SPAM_DECISION}"
          exit 1
```

eavanvalkenburg (approved):

.github/workflows/issue-triage.yml
```diff
@@ -0,0 +1,178 @@
+name: Issue Triage
+
+on:
+  issues:
+    types:
+      - opened
```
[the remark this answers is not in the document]
moonbox3: Fixed in `bf32fd02`.

.github/workflows/issue-triage.yml
```diff
@@ -0,0 +1,178 @@
...
+        description: Issue number to triage
+        required: true
+        type: string
+
+permissions:
+  contents: read
+  issues: write
+
+concurrency:
+  group: issue-triage-${{ github.repository }}-${{ github.event.issue.number || github.run_id }}
```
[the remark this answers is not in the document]
moonbox3: Fixed in `bf32fd02`.

.github/workflows/issue-triage.yml
```diff
@@ -0,0 +1,178 @@
...
+            try {
+              const teamMembership = await github.rest.teams.getMembershipForUserInOrg({
+                org: context.repo.owner,
+                team_slug: process.env.TEAM_NAME,
+                username: author,
+              });
+              isTeamMember = teamMembership.data.state === 'active';
+            } catch (error) {
+              console.log(`Team membership lookup failed for ${author}: ${error.message}`);
+              isTeamMember = false;
```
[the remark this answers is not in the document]
moonbox3: Fixed in `bf32fd02`.

.github/workflows/issue-triage.yml
moonbox3: Fixed in `bf32fd02`.

.github/workflows/issue-triage.yml
moonbox3: Fixed in `bf32fd02`.

.github/workflows/issue-triage.yml
moonbox3: Fixed in `bf32fd02`.

.github/workflows/issue-triage.yml
```diff
@@ -0,0 +1,171 @@
+name: Issue Triage
+
+on:
+  issues:
+    types:
+      - labeled
```
moonbox3: Let's make this a manually triggered workflow right now for initial testing.

moonbox3: Fixed in `4a729e30`.

Issue #3914, opened 2026-02-13
.NET: [A2A Hosting] Streaming support


SergeyMenshykh:
Currently A2A is routing streaming client request to non-streaming AIAgent implementation.

https://github.com/a2aproject/a2a-dotnet/issues/279

SergeyMenshykh:
Fixed by: https://github.com/microsoft/agent-framework/pull/5427

Issue #3310, opened 2026-01-20
.NET : Support for A2A streaming in `MapA2A`


anktsrkr:
While integrating A2A with the Agent Framework, I noticed that the `MapA2A` extension currently uses `RunAsync,` which produces a fully-buffered response rather than streaming partial deltas.

Reference:
https://github.com/microsoft/agent-framework/blob/73761aa4a3ad0d96a1a2a155f2b30a6a68da5b94/dotnet/src/Microsoft.Agents.AI.Hosting.A2A/AIAgentExtensions.cs#L50

Since A2A supports streaming, I’m trying to understand the recommended way to enable streaming through this helper.

Questions:
- Is there an existing streaming API (e.g., RunStreamingAsync) that `MapA2A` is expected to use?
- If not, is streaming support planned for `MapA2A` based on Agent's capability?

anktsrkr:
Also I noticed that the `.ToPart()` method in  currently handles `TextContent`, `UriContent`, and `DataContent`, but drops `FunctionCallContent` entirely.

Reference:
https://github.com/microsoft/agent-framework/blob/73761aa4a3ad0d96a1a2a155f2b30a6a68da5b94/dotnet/src/Microsoft.Agents.AI.Hosting.A2A/AIAgentExtensions.cs#L57

This causes structured function calls from the model to be lost when converting to A2A parts, which breaks tool invocation workflows.

Questions:
- Is there a recommended way to preserve `FunctionCallContent` in `.ToPart()`?


Thanks for the guidance!

DeagleGross:
Currrently being supported via Task support in A2A at https://github.com/microsoft/agent-framework/pull/3732

markwallace-microsoft:
@anktsrkr, friendly reminder — this issue is waiting on your response. Please share any updates when you get a chance. (This is an automated message.)

filipw:
The A2A integration still only uses `agent.RunAsync` not `agent.RunStreamingAsync`. The linked PR adds ContinuationToken/background-task support, which is a different feature from real-time token streaming. 

To support token streaming to the caller, we still have to polyfill this feature by adding something like `taskManager.OnTaskUpdated += RunAgentStreamingAsync;`. This should be a first class extension in AF (for example `MapStreamingA2A`)

ecc-parity-check:
Is there any sample code showing how to use Tasks to get messages with streaming with A2A?
[...]

PR #5427, opened 2026-04-22
.NET: Add streaming support to A2A agent handler

SergeyMenshykh:
## Summary

Adds streaming message support to `A2AAgentHandler` so that requests received via the A2A streaming endpoint (`StreamingResponse = true`) are handled using `RunStreamingAsync` instead of `RunAsync`.

### Production code

- **`A2AAgentHandler.ExecuteAsync`** - routes to `HandleNewMessageStreamingAsync` when `context.StreamingResponse` is `true`.
- **`HandleNewMessageStreamingAsync`** - iterates `RunStreamingAsync`, enqueues an A2A `Message` for each `AgentResponseUpdate`, and saves the session after the stream completes. Unlike the non-streaming path, this does not use `AgentRunMode` or task lifecycle events.
- **`CreateMessageFromUpdate`** - maps `AgentResponseUpdate` to A2A `Message` (parallel to the existing `CreateMessageFromResponse`).
- **`MessageConverter.ToParts(AgentResponseUpdate)`** - converts streaming update contents to A2A `Part` objects, filtering out unsupported content types.

### Tests (16 new tests)

- 4 tests for `MessageConverter.ToParts(AgentResponseUpdate)` - empty contents, text content, multiple contents, unsupported content filtering.
- 12 tests for the streaming handler path - message enqueuing, metadata/options propagation, `ReferenceTaskIds` rejection, context ID generation, session persistence, `ResponseId` mapping, and `null` edge cases.

### Out Of Scope
Supporting long-running operations/tasks with streaming is out of the scope of this PR and will be added later.

Closes:
- https://github.com/microsoft/agent-framework/issues/3310
- https://github.com/microsoft/agent-framework/issues/3914


westey-m (approved):

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/A2AAgentHandler.cs
```diff
@@ -80,13 +88,19 @@ private async Task HandleNewMessageAsync(RequestContext context, AgentEventQueue
...
+        {
+            response = await this._hostAgent.RunAsync(
+                chatMessages,
+                session: session,
+                options: options,
+                cancellationToken: cancellationToken).ConfigureAwait(false);
+        }
+        finally
+        {
+            await this._hostAgent.SaveSessionAsync(contextId, session, CancellationToken.None).ConfigureAwait(false);
```
peibekwe, line 102: Do you want to pass the cancellationToken here as well?

peibekwe: I see `CancellationToken.None` for all the `finally` block changes, so I'm not sure if it's by design or missed.

peibekwe (approved):

dotnet/src/Microsoft.Agents.AI.Hosting.A2A/A2AAgentHandler.cs
SergeyMenshykh: Good question! CancellationToken.None is intentional here - since we're in a finally block, we want the session save to succeed even if the original token was cancelled

PR #5446, opened 2026-04-23
Python: Address PR 5331 comments and track sesssion while calling Agent in email_security_example

shrutitople:
### Motivation and Context

Address comments in PR and added a shared session in the email example.

### Description

- Removed dead message-labeling methods from LabelTrackingFunctionMiddleware  — Removed label_message(), get_message_label(), label_messages(), and get_all_message_labels() along with the backing _message_labels field. These were unused in production code (never called from process() or anywhere else internally) and didn't belong on a FunctionMiddleware subclass, which only intercepts tool calls. The LabeledMessage is renamed to SecurityLabeledMessage class and is retained for role-based label inference.

- Fixed docstring example in LabelTrackingFunctionMiddleware — Added a sample tool (get_weather) to the usage example so the middleware has a function to intercept. Previously the example created an Agent with the middleware but no tools, which was misleading.

- Fixed session-less agent.run() in email example — Both scenarios now share an AgentSession so conversation history persists. Previously each agent.run() was stateless, meaning Scenario 2 had no context from Scenario 1 — the LLM never attempted to call send_email so the policy violation never triggered.

- Updated docs — Removed references to the deleted methods from FIDES_IMPLEMENTATION_SUMMARY.md and FIDES_DEVELOPER_GUIDE.md. Cleaned up the "Phase 1: Message-Level Tracking" section.

- Removed 4 tests from TestMiddlewareMessageLabeling that tested the deleted methods. All 111 remaining tests pass.


python/packages/core/agent_framework/security.py
```diff
@@ -831,15 +831,23 @@ class LabelTrackingFunctionMiddleware(FunctionMiddleware):
     Examples:
         .. code-block:: python
 
+<<<<<<< HEAD:python/packages/core/agent_framework/_security.py
```
eavanvalkenburg: this needs a fix!

python/samples/02-agents/security/email_security_example.py
```diff
@@ -278,7 +278,13 @@ async def run_scenarios(agent, config):
     print("- Injection attempts in emails are NOT followed")
     print()
 
-    response = await agent.run("Please fetch my recent emails and give me a brief summary of each one.")
+    # Use a shared session so conversation history persists across scenarios.
+    # Without this, each agent.run() starts a fresh conversation and the LLM
+    # won't know about the emails fetched in Scenario 1 — it would never
+    # attempt to call send_email, so the policy enforcer would never trigger.
+    session = AgentSession()
```
eavanvalkenburg:
```suggestion
    session = agent.get_session()
```

python/samples/02-agents/security/FIDES_DEVELOPER_GUIDE.md
```diff
@@ -495,8 +495,11 @@ The instructions explain:
 - How to pass `variable_ids` to reference hidden content
 - Best practices for secure content handling
 
-### 9. Message-Level Label Tracking (Phase 1)
+### 9. LabeledMessage Class
 
+<<<<<<< HEAD
```
eavanvalkenburg: another one

eavanvalkenburg (approved):

Issue #5441, opened 2026-04-23
Python: [Bug]: Foundry Hosted Agents - Missing /liveness route while /readiness exists


georgeollis:
### Description

The hosted agent server provides a readiness endpoint but not a liveness endpoint.
In hosted deployments, the service appears to expect a liveness probe, so this becomes a runtime compatibility issue.

What happened:
1. `GET /readiness` returns `200`.
2. `GET /liveness` returns `404` repeatedly.

What I expected:
1. A built-in `GET /liveness` endpoint that returns `200` when the process is alive.
2. Or explicit documentation that only `/readiness` is supported and guidance for liveness probe setup.

Steps to reproduce:
1. Start a hosted agent server (for example with `azd ai agent run`).
2. Call `GET /readiness` and observe `200`.
3. Call `GET /liveness` and observe `404`.

### Code Sample

```python
import os

from agent_framework import Agent
from agent_framework.foundry import FoundryChatClient
from agent_framework_foundry_hosting import ResponsesHostServer
from azure.identity import AzureCliCredential


def main():
    client = FoundryChatClient(
        project_endpoint=os.environ["FOUNDRY_PROJECT_ENDPOINT"],
        model=os.environ["MODEL_DEPLOYMENT_NAME"],
        credential=AzureCliCredential(),
    )

    agent = Agent(
        client=client,
        instructions="You are a friendly assistant. Keep your answers brief.",
        default_options={"store": False},
    )

    server = ResponsesHostServer(agent)
    server.run()


if __name__ == "__main__":
    main()
```

### Error Messages / Stack Traces

```markdown
2026-04-23 09:54:55,033 INFO azure.ai.agentserver: Inbound GET /liveness started
2026-04-23 09:54:55,034 INFO azure.ai.agentserver: 127.0.0.1:33978 "GET /liveness 1.1" 404 9 1065μs
2026-04-23 09:54:55,034 WARNING azure.ai.agentserver: Inbound GET /liveness completed with status 404 in 0.5ms

2026-04-23 09:55:05,037 INFO azure.ai.agentserver: Inbound GET /liveness started
2026-04-23 09:55:05,037 INFO azure.ai.agentserver: 127.0.0.1:51698 "GET /liveness 1.1" 404 9 1064μs
2026-04-23 09:55:05,037 WARNING azure.ai.agentserver: Inbound GET /liveness completed with status 404 in 0.4ms

[...]

PR #5448, opened 2026-04-23
Python: Add /liveness endpoint for Kubernetes liveness probe

Jah-yee:
Good day

Found that the service has a /readiness endpoint but is missing /liveness, causing Kubernetes liveness probes to receive 404 responses.

This PR adds the /liveness endpoint returning {"status": "alive"} following the same pattern as /readiness.

Thank you for your work on this project. I hope this small fix is helpful. Please let me know if there's anything to adjust.

Warmly, RoomWithRoof

Reference: #5441

### DIFF
--- a/python/scripts/local_mcp_streamable_http_server.py
+++ b/python/scripts/local_mcp_streamable_http_server.py
@@ -52,6 +52,12 @@ def create_server(*, host: str, port: int, mount_path: str) -> FastMCP:
             }
         )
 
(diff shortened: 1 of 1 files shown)


georgeollis:
This doesn’t fix the issue, nor does it even modify the correct file. It shouldn’t be implemented, and the original problem it’s attempting to address isn’t actually an issue with the new Foundry hosted agent experience.

eavanvalkenburg (approved):

Issue #5445, opened 2026-04-23
Python: .NET: [Feature]: Add Valkey Context Provider and Chat Message Store


MatthiasHowellYopp:
### Description

##Problem
The .NET Agent Framework currently has no dedicated Valkey integration package. While StackExchange.Redis (already in Directory.Packages.props) is protocol-compatible with Valkey for basic key-value operations, there is no ChatHistoryProvider or AIContextProvider implementation backed by Valkey.

Teams running Valkey in production — whether self-hosted or through managed cloud services like AWS ElastiCache for Valkey or GCP Memorystore for Valkey — have no first-party way to use Valkey for persistent chat history or long-term memory context in the .NET Agent Framework. This has become increasingly common since the Redis license change.

##Proposed Solution
Add a new Microsoft.Agents.AI.Valkey NuGet package providing:

- ValkeyChatHistoryProvider — persistent chat history implementing ChatHistoryProvider, using StackExchange.Redis (which is protocol-compatible with Valkey) for basic key-value/list operations. Follows the same patterns as CosmosChatHistoryProvider with state management via ProviderSessionState<TState>, configurable message limits, and session-scoped isolation.

- ValkeyContextProvider — long-term memory context provider extending MessageAIContextProvider, using Valkey's native vector search capabilities (FT.CREATE / FT.SEARCH) for semantic retrieval of past conversation context. Follows the same patterns as Mem0Provider with configurable scopes, context prompts, and message filtering.

The package would follow the same structure and patterns as Microsoft.Agents.AI.CosmosNoSql (for ChatHistoryProvider) and Microsoft.Agents.AI.Mem0 (for AIContextProvider).

##Server Requirements

- ValkeyChatHistoryProvider: Works with any Valkey (or Redis OSS) server — basic key-value operations only.
- ValkeyContextProvider: Requires valkey-search >= 1.2 (ships with valkey-bundle >= 9.1.0). Can be tested on valkey/valkey-bundle:9.1.0-rc1.
- Value

- Performance: Valkey's valkey-search module delivers single-digit millisecond latency with 99%+ recall for vector search operations
- Scaling: Linear scaling with cluster mode support
- Open-source governance: Linux Foundation project with community-driven development, clear licensing
- Cloud support: Managed services from AWS, GCP, and other major cloud providers
- Migration path: Easy migration from Redis deployments for teams affected by the Redis license change
- Parity: Companion to the Python agent-framework-valkey package (see #5260)

### Code Sample

```markdown
[...]

PR #5542, opened 2026-04-28
.NET: Adds Valkey to chat message history - issue 5445

MatthiasHowellYopp:
Motivation and Context
The .NET Agent Framework currently has no dedicated Valkey integration package. Teams running Valkey in production — whether self-hosted or through managed cloud services like AWS ElastiCache for Valkey or GCP Memorystore for Valkey — have no first-party way to use Valkey for persistent chat history or long-term memory context. This has become increasingly common since the Redis license change.

The Microsoft.Agents.AI.Valkey package was scaffolded with working ValkeyChatHistoryProvider and ValkeyContextProvider implementations. This PR completes the package by adding missing pieces identified during review, fixing bugs in the existing sample, and adding a Bedrock-powered sample for AWS users.

Fixes issue https://github.com/microsoft/agent-framework/issues/5445

Description
###Valkey package improvements:

- Added ValkeyProviderScope copy constructor matching the Mem0ProviderScope pattern, with null-guard via Throw.IfNull
- Added XML doc tags to all ValkeyProviderScope properties for consistency with Mem0ProviderScope
- Kept the package dependency lean — only references Microsoft.Agents.AI.Abstractions (no dependency on Microsoft.Agents.AI)

###Bug fix in existing sample:

- Fixed Program.cs — the stateInitializer referenced session?.Id which doesn't exist on AgentSession. Replaced with Guid.NewGuid():N using a discard parameter to make intent clear

###New Bedrock sample (AgentWithMemory_Step03_MemoryUsingValkey_Bedrock):

- Demonstrates both ValkeyChatHistoryProvider and ValkeyContextProvider powered by Amazon Bedrock via AWSSDK.Extensions.Bedrock.MEAI
- Uses IAmazonBedrockRuntime.AsIChatClient() which provides an IChatClient that plugs directly into the Agent Framework's ChatClientAgent
- Uses the standard AWS credential chain (env vars, profile, IAM role)
- Includes README with prerequisites, environment variables, and run instructions

###Infrastructure:

- Added AWSSDK.Extensions.Bedrock.MEAI v4.0.6.10 to Directory.Packages.props
- Registered the Bedrock sample in agent-framework-dotnet.slnx

Unit tests:

Added tests for ValkeyProviderScope copy constructor (clone all properties, null source throws)
23 tests total, all passing

Contribution Checklist
[x] The code builds clean without any errors or warnings
[x] The PR follows the [Contribution Guidelines](https://github.com/microsoft/agent-framework/blob/main/CONTRIBUTING.md)
[x] All unit tests pass, and I have added new tests where possible
[ ] Is this a breaking change? If yes, add "[BREAKING]" prefix to the title of the PR.


MatthiasHowellYopp:
@westey-m Hoping I can get a review on this this week, tks

dotnet/src/Microsoft.Agents.AI.Valkey/Microsoft.Agents.AI.Valkey.csproj
```diff
@@ -0,0 +1,40 @@
...
+    <Title>Microsoft Agent Framework - Valkey integration</Title>
+    <Description>Provides Valkey integration for Microsoft Agent Framework, including chat history persistence and context provider with full-text search.</Description>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Microsoft.Agents.AI.Abstractions\Microsoft.Agents.AI.Abstractions.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="StackExchange.Redis" />
```
westey-m: I'm assuming this code is not technically Valkey specific, but can work with Redis as well?
If so, I'm thinking we may want to rather call it Microsoft.Agents.AI.Redis, but call out that it is also Valkey compatible.

dotnet/src/Microsoft.Agents.AI.Valkey/ValkeyChatHistoryProvider.cs
```diff
@@ -0,0 +1,314 @@
...
+/// conversation content. Ensure the Valkey server is configured with appropriate access controls and encryption in transit
+/// (TLS). The <see cref="MaxMessages"/> property can limit stored messages per conversation.</description></item>
+/// <item><description><strong>Compromised store risks:</strong> Agent Framework does not validate or filter messages loaded
+/// from the store — they are accepted as-is. If the Valkey store is compromised, adversarial content could be injected
+/// into the conversation context.</description></item>
+/// </list>
+/// </para>
+/// </remarks>
+[RequiresUnreferencedCode("The ValkeyChatHistoryProvider uses JSON serialization which is incompatible with trimming.")]
+[RequiresDynamicCode("The ValkeyChatHistoryProvider uses JSON serialization which is incompatible with NativeAOT.")]
```
westey-m: We should be able to make this compatible fairly easily by accepting an optional JsonSerializerOptions as input into the provider.  We can default to AgentAbstractionsJsonUtilities.DefaultOptions.

You can then use this overload to be compatible:
```csharp
JsonSerializer.Serialize(
    continuationToken,
    jsonSerializerOptions.GetTypeInfo(typeof(IList<ChatMessage>));
```

You'd also need to pass a jso to ProviderSessionState, which can be based on a JsonSerializerContext created with the required state objects.
See e.g. AgentJsonUtilities and how this is being used in BackgroundAgentsProvider

dotnet/src/Microsoft.Agents.AI.Valkey/ValkeyChatHistoryProvider.cs
```diff
@@ -0,0 +1,314 @@
...
+    /// </remarks>
+    public ValkeyChatHistoryProvider(
+        string connectionString,
+        Func<AgentSession?, State> stateInitializer,
+        string keyPrefix = "chat_history",
+        string? stateKey = null,
+        ILoggerFactory? loggerFactory = null,
+        Func<IEnumerable<ChatMessage>, IEnumerable<ChatMessage>>? provideOutputMessageFilter = null,
+        Func<IEnumerable<ChatMessage>, IEnumerable<ChatMessage>>? storeInputRequestMessageFilter = null,
+        Func<IEnumerable<ChatMessage>, IEnumerable<ChatMessage>>? storeInputResponseMessageFilter = null)
```
westey-m: We prefer having optional params in an options class, to avoid additions in unnatural locations if needing to extend these in future.  Required parameters stay on the constructor.  The options param is itself optional.
See e.g. InMemoryChatHistoryProvider & InMemoryChatHistoryProviderOptions or ChatHistoryMemoryProvider & ChatHistoryMemoryProviderOptions.

The one param that is an exception is loggerFactory, which stays on the constructor and is typically last.

dotnet/src/Microsoft.Agents.AI.Valkey/ValkeyChatHistoryProvider.cs
```diff
@@ -0,0 +1,314 @@
...
+    /// Gets or sets the maximum number of messages to retain per conversation.
+    /// When exceeded, oldest messages are automatically trimmed. Null means unlimited.
+    /// </summary>
+    public int? MaxMessages { get; set; }
+
+    /// <summary>
+    /// Gets or sets the maximum number of messages to retrieve from the provider.
+    /// Null means no limit.
+    /// </summary>
+    public int? MaxMessagesToRetrieve { get; set; }
```
westey-m: Do we expect devs to be able to dynamically change this after the agent is running?  If not, we can make these read-only and settable via options on the constructor.  Leaning towards that unless we have a good use case.

dotnet/src/Microsoft.Agents.AI.Valkey/ValkeyContextProvider.cs
```diff
@@ -0,0 +1,493 @@
...
+/// <list type="bullet">
+/// <item><description><strong>PII and sensitive data:</strong> Conversation messages are stored in Valkey
+/// and may contain PII. Ensure the server is configured with appropriate access controls and TLS.</description></item>
+/// <item><description><strong>Indirect prompt injection:</strong> Memories retrieved from Valkey are injected
+/// into the LLM context as user messages. If the store is compromised, adversarial content could influence
+/// LLM behavior.</description></item>
+/// </list>
+/// </para>
+/// </remarks>
+public sealed class ValkeyContextProvider : MessageAIContextProvider, IAsyncDisposable
```
westey-m: This class mostly overlaps with ChatHistoryMemoryProvider, which can operate on any store that implements the Microsoft.Extensions.VectorData abstractions.
There is already a Redis implementation for MEVD, and the ChatHistoryMemoryProvider can be used with that to do vector search to find matching chat history.
See the AgentWithMemory_Step01_ChatHistoryMemory sample in conjunction with https://learn.microsoft.com/en-us/semantic-kernel/concepts/vector-store-connectors/out-of-the-box-connectors/redis-connector?pivots=programming-language-csharp.
Note that the Microsoft.SemanticKernel.Connectors.Redis package is in the process of being renamed to https://www.nuget.org/packages/CommunityToolkit.VectorData.Redis, since it is not directly related to SK.

dotnet/samples/02-agents/AgentWithMemory/AgentWithMemory_Step03_MemoryUsingValkey/Program.cs
```diff
@@ -0,0 +1,80 @@
...
+
+using Azure.AI.OpenAI;
+using Azure.Identity;
+using Microsoft.Agents.AI;
+using Microsoft.Agents.AI.Valkey;
+using Microsoft.Extensions.AI;
+using OpenAI.Chat;
+
+var endpoint = Environment.GetEnvironmentVariable("AZURE_OPENAI_ENDPOINT") ?? throw new InvalidOperationException("AZURE_OPENAI_ENDPOINT is not set.");
+var deploymentName = Environment.GetEnvironmentVariable("AZURE_OPENAI_DEPLOYMENT_NAME") ?? "gpt-4o-mini";
```
westey-m:
```suggestion
var deploymentName = Environment.GetEnvironmentVariable("AZURE_OPENAI_DEPLOYMENT_NAME") ?? "gpt-5.4-mini";
```

dotnet/samples/02-agents/AgentWithMemory/AgentWithMemory_Step03_MemoryUsingValkey/Program.cs
```diff
@@ -0,0 +1,80 @@
...
+var valkeyConnection = Environment.GetEnvironmentVariable("VALKEY_CONNECTION") ?? "localhost:6379";
+
+// --- Part 1: Chat History with ValkeyChatHistoryProvider ---
+Console.WriteLine("=== Part 1: ValkeyChatHistoryProvider — Persistent Chat History ===\n");
+
+await using var historyProvider = new ValkeyChatHistoryProvider(
+    valkeyConnection,
+    stateInitializer: _ => new ValkeyChatHistoryProvider.State($"sample-{Guid.NewGuid():N}"),
```
westey-m: Would it make sense to have a default stateInitializer implementation that just uses a Guid?

westey-m:
> @westey-m Hoping I can get a review on this this week, tks

Thanks for the contribution @MatthiasHowellYopp.  I've added a few comments.

dotnet/src/Microsoft.Agents.AI.Valkey/Microsoft.Agents.AI.Valkey.csproj
MatthiasHowellYopp: Switched to Valkey.Glide 1.1.0 — the official Valkey .NET client (GA April 2026, 1.1.0 added FT.* search support last week). It's API-compatible with StackExchange.Redis but is the Valkey project's native client. This makes the "Valkey" naming accurate — we're using the Valkey-native client, not a Redis client talking to Valkey.

dotnet/samples/02-agents/AgentWithMemory/AgentWithMemory_Step03_MemoryUsingValkey/Program.cs
MatthiasHowellYopp: Keeping this required intentionally. For a persistent store, the conversation ID is the primary key for resumption — it needs to be deterministic across app restarts. A random GUID default would silently create orphaned conversations on every restart, which is a confusing failure mode. The user should decide their keying strategy (user ID, thread ID, etc.) based on their application's needs.

MatthiasHowellYopp:
Thanks for the thorough review @westey-m — pushed updates addressing all feedback:

Changes made:

Switched from StackExchange.Redis to Valkey.Glide 1.1.0 (official Valkey .NET client, GA'd April 2026). This resolves the naming question — we're now using the Valkey-native client library.

Extracted optional params into ValkeyChatHistoryProviderOptions following the InMemoryChatHistoryProviderOptions pattern. Constructor is now (IConnectionMultiplexer, Func<State>, Options?, ILoggerFactory?).

Added JsonSerializerOptions to options, passed through to ProviderSessionState. Removed class-level [RequiresUnreferencedCode]/[RequiresDynamicCode].

MaxMessages/MaxMessagesToRetrieve are now private readonly fields set from options.

Removed ValkeyContextProvider and ValkeyProviderScope — agree it overlaps with ChatHistoryMemoryProvider + MEVD. Can revisit as a separate enhancement if there's a distinct use case for full-text search without embeddings.
Removed connection string constructors — caller manages their own IConnectionMultiplexer lifetime.

Updated model name in sample.

dotnet/src/Microsoft.Agents.AI.Valkey/Microsoft.Agents.AI.Valkey.csproj
westey-m: Can you use the Valkey sdk with Redis as well?  I'm thinking that if the Redis sdk works for both Redis and Valkey than using the Redis SDK and calling this a Redis lib (which also supports Valkey) is flexible and means we can have one lib for both.

dotnet/src/Microsoft.Agents.AI.Valkey/ValkeyChatHistoryProvider.cs
```diff
@@ -0,0 +1,228 @@
...
+        this._logger = loggerFactory?.CreateLogger<ValkeyChatHistoryProvider>();
+    }
+
+    /// <inheritdoc />
+    public override IReadOnlyList<string> StateKeys => this._stateKeys ??= [this._sessionState.StateKey];
+
+    /// <inheritdoc />
+    [UnconditionalSuppressMessage("Trimming", "IL2026", Justification = "ChatMessage serialization uses known types.")]
+    [UnconditionalSuppressMessage("AOT", "IL3050", Justification = "ChatMessage serialization uses known types.")]
+    protected override async ValueTask<IEnumerable<ChatMessage>> ProvideChatHistoryAsync(InvokingContext context, CancellationToken cancellationToken = default)
```
westey-m, line 83: We should be able to remove these as well.  JsonSerializer.Deserialize below just needs to be updated to use the signature that resolve the target type from the jsonSerializerOptions.
```csharp
JsonSerializer.Deserialize(value.ToString(), jso.GetTypeInfo(typeof(ChatMessage)) as ChatMessage
```

The code can also default to AgentAbstractionsJsonUtilities.DefaultOptions if not provided.

dotnet/src/Microsoft.Agents.AI.Valkey/ValkeyChatHistoryProvider.cs
```diff
@@ -0,0 +1,228 @@
...
+        this._logger?.LogInformation(
+            "ValkeyChatHistoryProvider: Retrieved {Count} messages for conversation.",
+            messages.Count);
+
+        return messages;
+    }
+
+    /// <inheritdoc />
+    [UnconditionalSuppressMessage("Trimming", "IL2026", Justification = "ChatMessage serialization uses known types.")]
+    [UnconditionalSuppressMessage("AOT", "IL3050", Justification = "ChatMessage serialization uses known types.")]
```
westey-m: Same here

dotnet/src/Microsoft.Agents.AI.Valkey/Microsoft.Agents.AI.Valkey.csproj
jbrinkman: @westey-m The Valkey client is only guaranteed to work with Redis 6.2, 7.0, 7.1, and 7.2. Any feature introduced after those versions is not guaranteed to be compatible. The Valkey-Glide team runs our entire test suite against those specific versions. As Redis and Valkey continue to evolve we expect to see some divergence naturally occur given the different feature priorities of the two teams.

My team is part of the Valkey and Valkey-Glide contributors and maintainers groups. We are tasked with helping AI projects provide a first-class experience for Valkey users so that our community is able to use these frameworks as part of their solutions. We would prefer not to use the Redis name as that causes confusion with our community since there are a number of Redis libraries which are not compatible with Valkey. Having the Valkey name provides a strong symbol to the Valkey community that compatibility is not an afterthought.

westey-m: Sounds reasonable.  Thanks @jbrinkman.  Would we be able to lean on you and your team for help in maintaining this over time as well? 😃
My question is mostly around trying to keep the cost of maintenance manageable for us.  While we are really keen on building an ecosystem of integrations, our ability to scale to it is limited.

Also, if you would be interested in building an implementation of the Microsoft.Extensions.VectorData abstractions for ValKey, let me know.  I can help explain the detail of it.  We already have a Redis implementation, which should be easy to port.

westey-m:
@MatthiasHowellYopp, there are still some AOT/Trimming suppressions that we should be able to get rid of.  Otherwise looks reasonable.  Let me know if you need help to unblock these

dotnet/samples/02-agents/AgentWithMemory/AgentWithMemory_Step03_MemoryUsingValkey/README.md
```diff
@@ -0,0 +1,31 @@
+# Agent with Memory Using Valkey
+
+This sample demonstrates using Valkey for both persistent chat history and long-term memory context with the Agent Framework.
+
+## Components
+
+- **ValkeyChatHistoryProvider** — Persists conversation history across sessions using Valkey lists. Works with any Valkey or Redis OSS server (no search module required).
+- **ValkeyContextProvider** — Stores and retrieves memories using Valkey's native full-text search (`FT.SEARCH`). Requires valkey-search >= 1.2.
```
westey-m: We can remove this now.  Same for the other sample readme as well.

dotnet/samples/02-agents/AgentWithMemory/AgentWithMemory_Step03_MemoryUsingValkey/README.md
```diff
@@ -0,0 +1,31 @@
+# Agent with Memory Using Valkey
+
+This sample demonstrates using Valkey for both persistent chat history and long-term memory context with the Agent Framework.
```
westey-m: We can remove the reference to long term memory now.  Same for the other sample readme as well.

dotnet/src/Microsoft.Agents.AI.Valkey/ValkeyChatHistoryProvider.cs
MatthiasHowellYopp: fixed

dotnet/samples/02-agents/AgentWithMemory/AgentWithMemory_Step03_MemoryUsingValkey/README.md
MatthiasHowellYopp: done

dotnet/samples/02-agents/AgentWithMemory/AgentWithMemory_Step03_MemoryUsingValkey/README.md
MatthiasHowellYopp: done

westey-m (approved):

dotnet/Directory.Packages.props
```diff
@@ -138,10 +138,12 @@
     <PackageVersion Include="Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore" Version="2.1.0" />
     <PackageVersion Include="Microsoft.Azure.Functions.Worker.Extensions.Mcp" Version="1.0.0" />
     <PackageVersion Include="Microsoft.Azure.Functions.Worker.Sdk" Version="2.0.7" />
-    <!-- Redis -->
-    <PackageVersion Include="StackExchange.Redis" Version="2.10.1" />
+    <!-- Valkey -->
+    <PackageVersion Include="Valkey.Glide" Version="1.1.0" />
```
westey-m, line 145:
```suggestion
    <!-- Redis -->
    <PackageVersion Include="StackExchange.Redis" Version="2.10.1" />
    <!-- Valkey -->
    <PackageVersion Include="Valkey.Glide" Version="1.1.0" />
```

westey-m (approved):

dotnet/src/Microsoft.Agents.AI.Valkey/ValkeyChatHistoryProviderOptions.cs
```diff
@@ -0,0 +1,56 @@
+// Copyright (c) Microsoft. All rights reserved.
```
westey-m: @MatthiasHowellYopp, the build is failing with this error:
```
/app/dotnet/src/Microsoft.Agents.AI.Valkey/ValkeyChatHistoryProviderOptions.cs(1,1): error CHARSET: Fix file encoding. [/app/dotnet/src/Microsoft.Agents.AI.Valkey/Microsoft.Agents.AI.Valkey.csproj]
```

Typically it can be solved by just running dotnet format on the project.

MatthiasHowellYopp: sorry about that. Should be fixed now.

westey-m (approved):

SergeyMenshykh (approved):

dotnet/src/Microsoft.Agents.AI.Valkey/Microsoft.Agents.AI.Valkey.csproj
```diff
@@ -0,0 +1,40 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>$(TargetFrameworksCore)</TargetFrameworks>
+    <RootNamespace>Microsoft.Agents.AI.Valkey</RootNamespace>
+    <VersionSuffix>preview</VersionSuffix>
```
rogerbarreto: Suggest keeping this as an alpha.

```suggestion
    <VersionSuffix>alpha</VersionSuffix>
```

rogerbarreto (approved):

peibekwe (approved):

westey-m (approved):

westey-m:
Thanks for the contribution @MatthiasHowellYopp!

PR #5549, opened 2026-04-28
Python: Channel spec

eavanvalkenburg:
### Contribution Checklist


docs/decisions/0026-hosting-channels.md
```diff
@@ -0,0 +1,337 @@
...
+
+## Validation
+
+The decision is validated when, in each implementing language:
+
+1. a one-channel Responses sample and a two-channel Responses + Invocations sample can be expressed with one host, default route layouts under `/responses/v1` and `/invocations/invoke`, and no handwritten protocol routing,
+2. a Responses channel by default forwards official request parameters like `temperature` into agent options and maps `store=False` into disabled session use,
+3. app authors can override that default per request with an run hook that validates or rewrites the final `ChannelRequest` (for example requiring `temperature`, ignoring `store`, or adapting for `A2AAgent`),
+4. a Telegram-style message channel can express command metadata, command registration, and either webhook or polling lifecycle behavior through the new channel contract,
+5. a custom webhook/message channel can be authored only against the new channel contract plus the language's web-framework primitives and lifecycle hooks,
```
moonbox3, line 220: Three AG-UI inbound fields have no home in this table:                                                                                                                                                                                                                                
- state (the AG-UI client's bidirectional state object, sent on every request, mutated mid-run via StateSnapshotEvent/StateDeltaEvent).                                                                                                                                               
- tools (frontend tool catalog the client supplies per request; the agent must see the definitions to let the LLM call them, but execution returns to the client).
- forwarded_props (carries resume, command, and HITL responses; today this drives workflow RequestInfo/RequestResponse round-trips in agent_framework_ag_ui._run_common).
 
The spec implies these would land in attributes: Mapping[str, Any], but attributes is documented as channel-private and is not consumed by the host. If the contract is "AG-UI uses attributes", we should say so explicitly, and pin a key namespace; otherwise add typed slots. Without one of these, an AG-UI channel author has to thread state/tools/forwarded_props outside the contract, which would defeat the purpose of normalizing on ChannelRequest.

docs/specs/002-python-hosting-channels.md
```diff
@@ -0,0 +1,1238 @@
...
+
+**`ChannelCommand` / `ChannelCommandContext` / `CommandHandler`** — cross-channel native command model (per PR #5393).
+
+| Type | Fields | Description |
+|---|---|---|
+| `ChannelCommand` | `name`, `description`, `handle`, `expose_in_ui=True`, `metadata={}` | Transport-neutral command descriptor. |
+| `ChannelCommandContext` | `session`, `state`, `raw_event`, `reply(...)`, `run(request)` | Runtime context for command handlers. |
+| `CommandHandler` | `Callable[[ChannelCommandContext], Awaitable[None] \| None]` | Command implementation; may reply locally, mutate state, or invoke the agent. |
+
+**`HostedRunResult` / `HostedStreamResult`** — outbound results from the host.
```
moonbox3, line 459: HostedStreamResult.updates: ResponseStream[...] is a stream of AgentRunResponseUpdate. AG-UI's mid-stream events are not those: StateSnapshotEvent, StateDeltaEvent (JSON Patch-shaped), MessagesSnapshotEvent, ToolCallStartEvent/ToolCallArgsEvent/ToolCallResultEvent, plus synthesized state events from predict_state_config (see _run_common.py). To honor those, an AG-UI channel needs raw access to the underlying agent event stream so it can interleave its own protocol events. The current shape forces the channel to either (a) re-derive AG-UI events from AgentRunResponseUpdate (lossy for state) or (b) skip context.run/stream entirely and call the agent directly (which means the host owns nothing useful for AG-UI: no session resolution, no identity, no policy). Can we document which of those is intended? or expose a raw-event seam?

docs/specs/002-python-hosting-channels.md
```diff
@@ -0,0 +1,1238 @@
...
+        register_native_commands: bool = True,
+        require_link: bool = False,
+    ) -> None: ...
+```
+
+`options` on `ResponsesChannel` is intentionally loosely typed in this draft because the option-mapping boundary is still settling. If it becomes a formal type later, it should be Agent Framework-owned, not imported from `agentserver`.
+
+#### Conversation history for the Responses channel
+
+The Responses channel does **not** introduce its own history seam. Conversation history for every channel — Responses, Invocations, Telegram, Teams — flows through the agent's standard core `HistoryProvider` (`agent_framework._sessions.HistoryProvider`). The Responses channel is a *caller-supplied session* channel (see [Channel session-carriage models](#channel-session-carriage-models)): it parses `previous_response_id` (and/or `conversation_id`) off the inbound request and projects it into `ChannelSession.key`. The host then resolves an `AgentSession` for that key and the agent's `HistoryProvider` does the load / append exactly as it would for any other session.
```
moonbox3: AG-UI's per-thread state is durable across turns and is not message history. There is no slot for non-message session state in HistoryProvider, and the spec doesn't cover where AG-UI state lives. Two options to consider:
- state is a channel-package concern, stored separately from the message store
- the host introduces a channel-state hook the way HistoryProvider is introduced for messages
 
Deferring the decision means each stateful channel author improvises (in-process dict, separate store, sneak it onto Message.additional_properties), and those choices will have to be reconciled when AG-UI lands. Can we pick the boundary now, even if no implementation ships in v1?

moonbox3 (commented):
I approached this review in the context of AG-UI...

Issue #5544, opened 2026-04-28
.NET: [Feature]: [.NET] Add ValkeyStreamBuffer for resumable agent streaming via Valkey Streams


MatthiasHowellYopp:
### Description

Description:

The Microsoft.Agents.AI.Valkey package currently provides ValkeyChatHistoryProvider (persistent conversation history via Valkey Lists) and ValkeyContextProvider (memory/context enrichment via Valkey full-text search). The remaining component from the Valkey integration story is a ValkeyStreamBuffer — a resumable streaming buffer backed by Valkey Streams.

Problem: When a client disconnects during a streamed agent response (RunStreamingAsync), partial output is lost. The framework's AgentResponseUpdate.ContinuationToken and AgentRunOptions.ContinuationToken provide the abstraction for stream resumption, and A2A implements it for its protocol, but there is no general-purpose, storage-backed implementation that any ChatClientAgent can use.

Proposed solution: Add a ValkeyStreamBuffer component to the Microsoft.Agents.AI.Valkey package that:

- Writes each AgentResponseUpdate chunk to a Valkey Stream via XADD, keyed by session/response ID
- Stamps each update with a ContinuationToken containing the Valkey Stream entry ID
- On reconnection, reads missed entries via XREAD starting from the client's last-seen entry ID
- Supports consumer groups (XREADGROUP) for multi-consumer scenarios
- Supports configurable TTL via XTRIM or MAXLEN to bound stream growth
- Works with both standalone Valkey and Valkey Cluster via StackExchange.Redis
- The implementation should follow the same patterns as the existing Valkey providers: ProviderSessionState<TState> for session-scoped state, IConnectionMultiplexer constructor overload for DI, IAsyncDisposable, CancellationToken forwarding, and ILogger integration.

Server requirements: Any Valkey (or Redis OSS) server — Valkey Streams use core commands only (XADD, XREAD, XREADGROUP, XACK, XTRIM), no modules required.

Reference: The A2A agent (
A2AAgent.cs
) demonstrates the continuation token pattern. AgentResponseUpdate.ContinuationToken and AgentRunOptions.ContinuationToken are the framework abstractions to integrate with.

### Code Sample

```markdown
using Microsoft.Agents.AI;
using Microsoft.Agents.AI.Valkey;
using Microsoft.Extensions.AI;

// Create the stream buffer backed by Valkey
await using var streamBuffer = new ValkeyStreamBuffer(
    "localhost:6379",
[...]

PR #5576, opened 2026-04-29
.NET: valkey streaming support dotnet

MatthiasHowellYopp:
### Motivation and Context

The Microsoft.Agents.AI.Valkey package provides ValkeyChatHistoryProvider and ValkeyContextProvider for persistent conversation history and memory context. However, the framework currently has no storage-backed resumable streaming component — if a client disconnects during a streamed agent response (RunStreamingAsync), partial output is lost and must be regenerated. This is the final piece of the Valkey integration story, delivering parity with the Python agent-framework-valkey package.

Valkey Streams (XADD/XRANGE) are a natural fit: each entry gets an auto-generated ID that serves as a continuation token, entries are append-only and durable, and XRANGE with an exclusive start ID gives exact resume-from-here semantics — all with core commands, no modules required.

Fixes issue https://github.com/microsoft/agent-framework/issues/5544

### Description

Adds ValkeyStreamBuffer to the Microsoft.Agents.AI.Valkey package — a resumable streaming buffer backed by Valkey Streams.

The component writes serialized AgentResponseUpdate chunks to a Valkey Stream via XADD and replays missed entries via XRANGE from a caller-provided entry ID. The stream entry ID returned by each AppendAsync call serves as the continuation token for resumption.

Public API:

- AppendAsync(responseId, update) — writes a chunk, returns the entry ID
- ReadAsync(responseId, afterEntryId) — replays entries after a given ID as IAsyncEnumerable
- GetEntryCountAsync(responseId) — returns stream length
- DeleteStreamAsync(responseId) — removes a stream
- MaxLength property — caps stream size via approximate XTRIM after each XADD

Follows the same patterns as the existing Valkey providers: dual constructors (connection string + IConnectionMultiplexer for DI), IAsyncDisposable with disposed-state guard, CancellationToken forwarding, malformed JSON resilience, trimming/AOT attributes, and ILogger integration at appropriate levels (Debug for per-request, Warning for skipped entries).

Includes a self-contained sample (AgentWithMemory_Step06_ResumableStreamingWithValkey) that demonstrates the full workflow — stream, disconnect, resume, full replay — against a live Valkey instance with no LLM or cloud credentials required.

64 unit tests passing, covering constructor validation, append/read round-trips, entry ID incrementing, MaxLength pass-through, malformed entry handling, cancellation, disposal guards, and empty stream edge cases.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.Valkey/ValkeyStreamBuffer.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Valkey/ValkeyStreamBuffer.cs
@@ -1,0 +1,210 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+using System.Text.Json;
+using System.Text.Json.Serialization.Metadata;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Microsoft.Shared.Diagnostics;
+using Valkey.Glide;
+
+namespace Microsoft.Agents.AI.Valkey;
+
+/// <summary>
+/// Provides a resumable streaming buffer backed by Valkey Streams.
+/// </summary>
+/// <remarks>
+/// <para>
+/// Writes <see cref="AgentResponseUpdate"/> chunks to a Valkey Stream via XADD and supports
+/// replaying missed entries via XRANGE from a given entry ID, enabling clients to reconnect
+/// mid-stream without data loss.
+/// </para>
+/// <para>
+/// Each stream is keyed by a caller-provided response ID (typically session + response scoped).
+/// The Valkey Stream entry ID serves as the continuation token for resumption.
+/// </para>
+/// <para>
+/// <strong>Server requirements:</strong> Any Valkey (or Redis OSS) server. Uses core Stream
+/// commands only (XADD, XRANGE, XLEN, XTRIM, DEL) — no modules required.
+/// </para>
+/// <para>
+/// <strong>Data retention:</strong> Streams are bounded by <see cref="ValkeyStreamBufferOptions.MaxLength"/> via XTRIM
+/// approximate trimming. Callers can also explicitly delete streams via <see cref="DeleteStreamAsync"/>.
+/// </para>
+/// <para>
+/// <strong>Security considerations:</strong>
+/// <list type="bullet">
+/// <item><description><strong>PII and sensitive data:</strong> Streamed response chunks may contain
+/// PII. Ensure the Valkey server is configured with appropriate access controls and TLS.</description></item>
+/// </list>
+/// </para>
+/// </remarks>
+public sealed class ValkeyStreamBuffer
+{
+    private const string ContentField = "content";
+
+    private readonly IConnectionMultiplexer _connection;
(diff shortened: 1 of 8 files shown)


MatthiasHowellYopp:
@westey-m Hoping I can get a review on this this week, tks

MatthiasHowellYopp:
@westey-m Hoping I can get a review on this this week, tks

dotnet/src/Microsoft.Agents.AI.Valkey/ValkeyStreamBuffer.cs
```diff
@@ -0,0 +1,210 @@
...
+/// </para>
+/// <para>
+/// <strong>Security considerations:</strong>
+/// <list type="bullet">
+/// <item><description><strong>PII and sensitive data:</strong> Streamed response chunks may contain
+/// PII. Ensure the Valkey server is configured with appropriate access controls and TLS.</description></item>
+/// </list>
+/// </para>
+/// </remarks>
+public sealed class ValkeyStreamBuffer
```
westey-m, line 45: @MatthiasHowellYopp, thanks for the PR.  While I think the problem space you are describing is definitely interesting, just the buffer component on it's own is not that useful in the AF context.  If you replaced AgentResponseUpdate with a generic type, this component could equally live in the Valkey .net lib.

I think most users would actually be looking for the integration into Agent Framework via some middleware that automatically deals with failures.  However, that has some challenges.  E.g. the service itself would need to have built in support for resumption.  There is some support in Microsoft.Extensions.AI for this already, via ChatClient.AllowBackgroundResponses, which if the service supports this will use continuation tokens to allow the caller to resume reading.  I'm not entirely sure how this would fit in with that though and would like to hear more about what you were thinking.

MatthiasHowellYopp: @westey-m After reviewing the PR and the initial issue I see that this PR is not complete.  I'd like to update this PR with a more complete solution to the issue, which should basically match what you describe.  Let me know if it is ok to proceed.

westey-m: I think this is quite a challenging feature in many ways, and I'm actually not super clear on how it would integrate with agent framework, but happy to see what you have in mind.  To save you time, effort and tokens, it may makes sense to create a small design doc of what you have in mind first.  Or join our office hours and come discuss.  See https://github.com/microsoft/agent-framework/blob/main/COMMUNITY.md for the office hours details.

MatthiasHowellYopp: I've added a design doc in tis pr in docs/design/valkey-stream-buffer.md.  I hope it has enough info for you to assess properly.

MatthiasHowellYopp: @westey-m Hoping you can have a look at the design doc.

westey-m: Things are a bit busy here.  I'll try and get to it next week.

MatthiasHowellYopp:
@westey-m Hoping I can get a review this week on this, thanks.

PR #5595, opened 2026-04-30
.NET: feat: Implement Magentic Orchestration for .NET

### LINKED ISSUE #1275: .NET: Add support for Magentic Orchestration
Python workflows support the Magentic Orchestration and .NET should be at parity.

### LINKED ISSUE #5154: .NET: [Feature]: When will .NET support Workflows Orchestrations - Magentic
### Description

When will .NET support Workflows Orchestrations - Magentic

### Code Sample

```markdown

```

### Language/SDK

Both

lokitoth:
### Motivation and Context

Magentic Orchestration allows users to build task-solving teams of Agents easily by relying on intelligent planning and orchestration of the plan using the available team, by routing dynamically around the workflow.

### Description

Implement Magentic Orchestration

- Closes #1275 
- Closes #5154

### Contribution Checklist


dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/Magentic/ExecutorAgentHarness.cs
```diff
@@ -0,0 +1,72 @@
...
+using System.Text.Json;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.AI;
+
+namespace Microsoft.Agents.AI.Workflows.Specialized.Magentic;
+
+internal sealed class ExecutorAgentHarness(AIAgent agent, AIAgentUnservicedRequestsCollector collector)
+{
+    internal const string AgentSessionKey = nameof(AgentSession);
```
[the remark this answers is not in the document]
lokitoth: That's because it is a key in a dictionary to a value of type `AgentSession`. It seems reasonable to not include the string "Key" in the key. If the type changes name, it stands to reason that the object being stored will have a slightly different naming meaning.

dotnet/src/Microsoft.Agents.AI.Workflows/MagenticPlanReviewResponse.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System.Collections.Generic;
+using Microsoft.Extensions.AI;
+
+namespace Microsoft.Agents.AI.Workflows;
+
+/// <summary>
+/// Review feedback for a proposed plan, including any revisions if the plan is not approved as-is. An
+/// empty list of review messages indicates approval of the proposed plan without any revisions.
```
alliscode, line 10: Should there be a more positive indication of approval? An empty list of messages could be an easy state to error in to.

dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/Magentic/MagenticOrchestrator.cs
```diff
@@ -0,0 +1,325 @@
...
+    }
+
+    private async ValueTask UpdatePlanAndDelegateAsync(MagenticTaskContext taskContext, IWorkflowContext context, CancellationToken cancellationToken)
+    {
+        bool isReplan = taskContext.TaskLedger != null;
+
+        taskContext.TaskLedger = await this._manager.UpdatePlanAsync(taskContext, context, cancellationToken)
+                                                    .ConfigureAwait(false);
+
+        this._fullTaskLedgerMessage ??= new(ChatRole.User, taskContext.ToTaskLedgerFullPrompt());
```
peibekwe: Seems like this gets initialized once and then doesn't get updated after a replan. Is the intention to return the new message after a replan? Because this seems like it currently just sets when `fullTaskLedgeMessage` is null.

dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/Magentic/MagenticOrchestrator.cs
```diff
@@ -0,0 +1,325 @@
...
+
+        // Update the Progress Ledger
+        try
+        {
+            await this._manager.UpdateProgressLedgerAsync(taskContext, context, cancellationToken).ConfigureAwait(false);
+
+            await context.AddEventAsync(new MagenticProgressLedgerUpdatedEvent(taskContext.ProgressLedger), cancellationToken)
+                         .ConfigureAwait(false);
+        }
+        catch (Exception ex)
```
peibekwe: nit: Should we handle `OperationCanceledException` separately to shut down gracefully before the generic catch?

dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/Magentic/MagenticOrchestrator.cs
```diff
@@ -0,0 +1,325 @@
...
+    }
+
+    protected internal override async ValueTask OnCheckpointRestoredAsync(IWorkflowContext context, CancellationToken cancellationToken = default)
+    {
+        await Task.WhenAll(base.OnCheckpointRestoredAsync(context, cancellationToken).AsTask(), LoadContextStateAsync())
+                  .ConfigureAwait(false);
+
+        async Task LoadContextStateAsync()
+        {
+            MagenticTaskState? state = await context.ReadStateAsync<MagenticTaskState>(MagenticConstants.MagenticTaskContextKey, cancellationToken: cancellationToken)
```
peibekwe, line 317: nit: I'm not sure this applies here, but does this, the request and response types need to be added to Sourcegen registrations for serialization?

feynmanloo:
@lokitoth 

Thanks a lot for implementing this feature — it’s incredibly helpful, especially for building Agent Teams in .NET, particularly in Harness Engineering scenarios where multi-agent collaboration is essential.

The Magentic Orchestration approach (planning + dynamic routing) significantly lowers the barrier for orchestrating complex workflows.

One small suggestion:
would it be possible to add some more practical samples? For example:

* A multi-agent collaboration scenario (e.g., analyze → plan → execute → verify)
* Demonstrations of dynamic routing between different agents
* A minimal end-to-end runnable example

These would greatly help us better understand the best practices and accelerate adoption in real-world use cases.

Really appreciate this valuable contribution 🙌

dotnet/src/Microsoft.Agents.AI.Workflows/MagenticPlanReviewResponse.cs
lokitoth: I was trying to keep consistent with Python here, and if we add a separate flag, we would need to define what happens if the user sets the flag to `false` but provides not revision messages. We could switch away from the record type and make the constructor internal and rely on semantically named static constructors (akin to the helper methods on `MagenticPlanReviewRequest`: `Approve` vs `Revise`), though I suspect most people would end up just using the request-side helpers anyways, rather than directly constructing the response.

dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/Magentic/MagenticOrchestrator.cs
lokitoth: Good catch

dotnet/src/Microsoft.Agents.AI.Workflows/ExecutorEvent.cs
```diff
@@ -10,6 +11,7 @@ namespace Microsoft.Agents.AI.Workflows;
 [JsonDerivedType(typeof(ExecutorInvokedEvent))]
 [JsonDerivedType(typeof(ExecutorCompletedEvent))]
 [JsonDerivedType(typeof(ExecutorFailedEvent))]
+[JsonDerivedType(typeof(MagenticOrchestratorEvent))]
```
peibekwe, line 14: nit: Since `MagenticOrchestratorEvent` derives from `WorkflowEvent`, should it be registered here as an `ExecutorEvent`?

dotnet/src/Microsoft.Agents.AI.Workflows/WorkflowsJsonUtilities.cs
```diff
@@ -97,6 +98,9 @@ private static JsonSerializerOptions CreateDefaultOptions()
     [JsonSerializable(typeof(AIAgentHostState))]
     [JsonSerializable(typeof(HandoffSharedState))]
     [JsonSerializable(typeof(HandoffAgentHostState))]
+    [JsonSerializable(typeof(MagenticPlanReviewRequest))]
+    [JsonSerializable(typeof(MagenticPlanReviewResponse))]
+    [JsonSerializable(typeof(MagenticTaskState))]
```
peibekwe, line 103: Should `ResetChatSignal` also be registered here as well as it's being emitted in the `ResetAndReplan` method?

dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/Magentic/ChatMessageExtensions.cs
```diff
@@ -0,0 +1,158 @@
...
+                        end = scanHead;
+                    }
+
+                    break;
+            }
+        }
+
+        if (end is null)
+        {
+            throw new InvalidOperationException("Unbalanced JSON braces.");
```
peibekwe, line 168: nit: Would something like `{"reason":"the output contained }", "answer": false}` be a valid output? Because it seems like that wouldn't be properly processed by the current implementation.

dotnet/src/Microsoft.Agents.AI.Workflows/MagenticWorkflowBuilder.cs
```diff
@@ -0,0 +1,150 @@
...
+    public Workflow Build() => this.ReduceToWorkflowBuilder().Build();
+
+    private TaskLimits Limits => new(
+        MaxRoundCount: this._maxRounds,
+        MaxResetCount: this._maxResets,
+        MaxStallCount: this._maxStalls);
+
+    private ValueTask<MagenticOrchestrator> CreateOrchestratorAsync(ExecutorConfig<ExecutorOptions> options, string sessionId)
+    {
+        return new(new MagenticOrchestrator(managerAgent, this._team, this.Limits, this._requirePlanSignoff));
```
peibekwe: Is there a possibility that `this._team` can be mutated after `Builder.Build` method was called? I'd recommend perhaps using a snapshot of the team that was used to build the workflow to avoid the possibility of each orchestrator getting a mutated team.

dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/Magentic/MagenticOrchestrator.cs
```diff
@@ -0,0 +1,326 @@
...
+        else
+        {
+            await this.DelegateToTeamAsync(taskContext, context, cancellationToken).ConfigureAwait(false);
+        }
+    }
+
+    protected override async ValueTask TakeTurnAsync(List<ChatMessage> messages, IWorkflowContext context, bool? emitEvents, CancellationToken cancellationToken = default)
+    {
+        // First Turn: Initialize the task context and send the initial messages to the planner agent
+        this._taskContext ??= new(messages, team, limits, emitEvents, []);
```
peibekwe, line 183: Perhaps I'm missing context, but is this only called for the first turn, or should subsequent turns append messages to `taskContext` or run the next coordinated round to avoid participant output being dropped?

peibekwe (approved):

alliscode (approved):

dotnet/src/Microsoft.Agents.AI.Workflows/ExecutorEvent.cs
lokitoth: Technically yes, though we do not actually register them, since Events are not JSON serializable generally (the failed events cannot be properly serialized because they contain Exception objects).

dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/Magentic/MagenticOrchestrator.cs
lokitoth: The first time the orchestrator gets messages, they come from the user as a task description. Subsequent turns are after a selected agent took a turn, and incoming messages are coming from the agent. These should not update the task description.

alliscode (approved):

PR #5604, opened 2026-05-01
.NET: Feat/dotnet shell tool

alliscode:
### Contribution Checklist


alliscode (commented):
## Automated Code Review

**Reviewers:** 3 | **Confidence:** 78%

### ✓ Security Reliability
> This is a well-structured new shell tool package. The previously-raised issues (killpg without setsid, DllImportSearchPath.System32, /bin/sh misclassification, timeout semantics, approval defaults) have all been addressed in this version. One remaining reliability issue: in DockerShellTool's stateless mode, if the caller's CancellationToken fires (not the timeout), the OperationCanceledException propagates without killing the per-call container, leaving an orphan that runs indefinitely until the Docker daemon or operator intervenes.

### ✓ Test Coverage
> The PR introduces a substantial new shell tool library with reasonable overall test coverage for the happy path and key security-relevant behaviors (approval gating, hardened configuration detection, policy deny list, persistent/stateless modes, timeout handling). However, there are notable gaps: ShellResult.FormatForModel() has no unit test despite branching logic; DockerShellTool.RunAsync policy rejection path is untested (only LocalShellTool tests it); ShellSession.TruncateHeadTail has no direct unit test for edge cases; the ShellPolicy default deny list only tests one pattern out of 16; and stderr capture is never directly asserted in any test. These gaps are not blocking for a first PR but represent meaningful coverage holes for a security-sensitive tool.

### ✗ Design Approach
> The PR introduces DockerShellTool and LocalShellTool along with supporting infrastructure. Two design issues were found. First, in DockerShellTool.StartAsync, BuildExecArgv(interactive: true) already appends --noprofile --norc to the argv, but the ResolvedShell is constructed with ShellKind.Bash, whose PersistentArgv() appends those same flags a second time as a suffix. The result is `docker exec -i <name> bash --noprofile --norc --noprofile --norc`. Bash silently accepts this (the flags are idempotent) so there is no functional failure today, but the design is fragile: the comment says 'ExtraArgv carries the docker-exec prefix so PersistentArgv prepends it before the bash flags', which incorrectly describes the actual behavior (the flags are already IN ExtraArgv and get doubled). The straightforward fix is to use ShellKind.Sh, whose PersistentArgv returns empty, so ExtraArgv is passed through unchanged. Second, BuildExecArgv(interactive: false) returns a dangling argv that ends with -c and no command argument. Because the return type is IReadOnlyList<string>, calers cannot append the command. This path is never invoked internally (stateless mode uses BuildRunArgvStateless instead), making the interactive: false branch dead code that produces an unusable public API.

### Flagged Issues
- [x] DockerShellTool.StartAsync uses ShellKind.Bash for the inner ResolvedShell, causing PersistentArgv() to append --noprofile --norc a second time on top of the flags already placed by BuildExecArgv(interactive: true). The resulting argv is `docker exec -i <name> bash --noprofile --norc --noprofile --norc`. The fix: use ShellKind.Sh (PersistentArgv returns Array.Empty) so ExtraArgv is used verbatim and nothing is doubled.

---

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellResolver.cs
```diff
@@ -0,0 +1,203 @@
...
+    {
+        var name = Path.GetFileNameWithoutExtension(path).ToUpperInvariant();
+        return name switch
+        {
+            "PWSH" or "POWERSHELL" => ShellKind.PowerShell,
+            "CMD" => ShellKind.Cmd,
+            // Non-bash POSIX shells: treat sh/dash/ash/busybox as plain sh
+            // so we don't pass bash-only flags like --noprofile / --norc.
+            "SH" or "DASH" or "ASH" or "BUSYBOX" => ShellKind.Sh,
+            _ => ShellKind.Bash,
```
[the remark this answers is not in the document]
alliscode: Update to only route bash to ShellKind.Bash and update doc.

dotnet/tests/Microsoft.Agents.AI.Tools.Shell.UnitTests/ShellResolverTests.cs
```diff
@@ -0,0 +1,65 @@
...
+        var argv = ResolveSingle(binary).PersistentArgv();
+
+        Assert.Empty(argv);
+    }
+
+    [Theory]
+    [InlineData("/bin/bash")]
+    [InlineData("/usr/local/bin/bash")]
+    [InlineData("/usr/bin/zsh")]
+    public void BashVariants_StatelessArgv_IncludeBashFlags(string binary)
```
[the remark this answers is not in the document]
alliscode: Fixed.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellResolver.cs
```diff
@@ -0,0 +1,203 @@
...
+
+    private static string SystemRoot() =>
+        Environment.GetEnvironmentVariable("SystemRoot") ?? @"C:\Windows";
+}
+
+/// <summary>Identifies the dialect of the resolved shell.</summary>
+internal enum ShellKind
+{
+    /// <summary>POSIX shell (bash, zsh) that accepts <c>--noprofile</c> / <c>--norc</c>.</summary>
+    Bash,
```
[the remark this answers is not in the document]
alliscode: Fixed as above.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellEnvironmentProvider.cs
```diff
@@ -0,0 +1,234 @@
...
+        var result = await this.RunProbeAsync(probe, cancellationToken).ConfigureAwait(false);
+        if (result is null || result.ExitCode != 0)
+        {
+            return null;
+        }
+        var firstLine = result.Stdout
+            .Split(['\r', '\n'], StringSplitOptions.RemoveEmptyEntries)
+            .FirstOrDefault();
+        return string.IsNullOrWhiteSpace(firstLine) ? null : firstLine.Trim();
+    }
```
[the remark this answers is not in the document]
alliscode: Added tool validation with ^[A-Za-z0-9._-]+$ regex regex

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellSession.cs
```diff
@@ -0,0 +1,857 @@
...
+
+    internal static (string text, bool truncated) TruncateHeadTail(string data, int cap)
+    {
+        if (data.Length <= cap)
+        {
+            return (data, false);
+        }
+        var head = data.Substring(0, cap / 2);
+        var tail = data.Substring(data.Length - (cap / 2));
+        return ($"{head}\n[... truncated {data.Length - cap} chars ...]\n{tail}", true);
```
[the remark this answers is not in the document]
alliscode: Updated to UTF-8 support.

dotnet/tests/Microsoft.Agents.AI.Tools.Shell.IntegrationTests/DockerShellExecutorIntegrationTests.cs
```diff
@@ -0,0 +1,194 @@
...
+/// <remarks>
+/// To run only these tests locally:
+/// <code>
+/// dotnet test --filter "Category=Integration&amp;FullyQualifiedName~DockerShellToolIntegrationTests"
+/// </code>
+/// or run the test exe directly with the trait filter.
+/// </remarks>
+[Trait("Category", "Integration")]
+public sealed class DockerShellToolIntegrationTests
+{
```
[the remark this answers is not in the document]
alliscode: Moved to a new project.

dotnet/tests/Microsoft.Agents.AI.Tools.Shell.UnitTests/ShellEnvironmentProviderTests.cs
```diff
@@ -0,0 +1,181 @@
...
+
+        // Force first probe.
+        _ = await provider.RefreshAsync();
+        var probesAfterFirst = fake.RunCount;
+
+        // Subsequent ProvideAIContext calls should not re-probe — they hit
+        // the cached _snapshotTask.
+        await provider.RefreshAsync();
+        Assert.True(fake.RunCount > probesAfterFirst, "Refresh should re-probe");
+    }
```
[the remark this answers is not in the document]
alliscode: Fixed

dotnet/tests/Microsoft.Agents.AI.Tools.Shell.IntegrationTests/DockerShellExecutorIntegrationTests.cs
```diff
@@ -0,0 +1,194 @@
...
+    // Small, fast image that has bash. Pulled lazily on first run.
+    // Alpine ships only busybox sh, which the persistent shell session can't use.
+    private const string TestImage = "debian:stable-slim";
+
+    private static async Task<bool> EnsureDockerOrSkipAsync()
+    {
+        if (!await DockerShellTool.IsAvailableAsync().ConfigureAwait(false))
+        {
+            Assert.Skip("Docker (or Podman) daemon is not available on this machine.");
+            return false; // unreachable
```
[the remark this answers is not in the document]
alliscode: Good catch to flag — but this repo is on xUnit v3, where `Assert.Skip` / `Assert.SkipWhen` are first-class. The same API is used across `OpenAIResponse.IntegrationTests`, `CopilotStudio.IntegrationTests`, `Foundry.IntegrationTests`, `Microsoft.Agents.AI.CosmosNoSql.UnitTests`, and others (30+ call sites). The integration project already references `xunit.v3`, so this is consistent with the established pattern. Resolving.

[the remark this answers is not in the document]
alliscode: Won't fix

dotnet/tests/Microsoft.Agents.AI.Tools.Shell.IntegrationTests/DockerShellToolIntegrationTests.cs
```diff
@@ -0,0 +1,194 @@
...
+    // Alpine ships only busybox sh, which the persistent shell session can't use.
+    private const string TestImage = "debian:stable-slim";
+
+    private static async Task<bool> EnsureDockerOrSkipAsync()
+    {
+        if (!await DockerShellTool.IsAvailableAsync().ConfigureAwait(false))
+        {
+            Assert.Skip("Docker (or Podman) daemon is not available on this machine.");
+            return false; // unreachable
+        }
```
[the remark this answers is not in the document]
alliscode: Won't fix

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellSession.cs
```diff
@@ -0,0 +1,948 @@
...
+                {
+                    break;
+                }
+
+                lock (this._bufferGate)
+                {
+                    for (var i = 0; i < n; i++)
+                    {
+                        buf.Add(chunk[i]);
+                    }
```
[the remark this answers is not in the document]
alliscode: Fixed, thanks.

dotnet/tests/Microsoft.Agents.AI.Tools.Shell.UnitTests/LocalShellToolTests.cs
```diff
@@ -0,0 +1,378 @@
...
+    }
+
+    [Fact]
+    public async Task RunAsync_NonZeroExit_PropagatesExitCodeAsync()
+    {
+        await using var shell = new LocalShellTool(mode: ShellMode.Stateless);
+        // Exit-1 phrasing portable across bash and PowerShell.
+        var script = RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
+            ? "exit 7"
+            : "exit 7";
```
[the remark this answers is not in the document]
alliscode: Fixed, thanks.

dotnet/tests/Microsoft.Agents.AI.Tools.Shell.IntegrationTests/DockerShellToolIntegrationTests.cs
```diff
@@ -0,0 +1,194 @@
...
+
+        await using var tool = new DockerShellTool(image: TestImage, mode: ShellMode.Persistent /* network defaults to "none" */);
+        await tool.StartAsync();
+
+        // Try to resolve a hostname; with --network none, even DNS should fail.
+        // Use getent (always present on debian) so we don't depend on optional tools.
+        var result = await tool.RunAsync("getent hosts example.com 2>&1; echo MARKER:$?");
+
+        Assert.Contains("MARKER:", result.Stdout);
+        // Non-zero status from wget proves network was blocked.
```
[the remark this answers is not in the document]
alliscode: Fixed, thanks.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellPolicy.cs
```diff
@@ -0,0 +1,140 @@
...
+            deny.Add(new Regex(pattern, RegexOptions.Compiled | RegexOptions.IgnoreCase));
+        }
+        this._denies = deny;
+
+        var allow = new List<Regex>();
+        if (allowList is not null)
+        {
+            foreach (var pattern in allowList)
+            {
+                allow.Add(new Regex(pattern, RegexOptions.Compiled));
```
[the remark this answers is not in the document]
alliscode: Fixed, thanks.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellResult.cs
```diff
@@ -0,0 +1,110 @@
...
+        }
+        _ = sb.Append("exit_code: ").Append(this.ExitCode);
+        return sb.ToString();
+    }
+}
+
+/// <summary>
+/// Specifies how a <see cref="LocalShellTool"/> dispatches commands to the underlying shell.
+/// </summary>
+public enum ShellMode
```
westey-m: NIT: Would be nice to have this and the exceptions in their own files.  They don't seem that related to ShellResult.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellPolicy.cs
```diff
@@ -0,0 +1,140 @@
...
+using System.Text.RegularExpressions;
+
+namespace Microsoft.Agents.AI.Tools.Shell;
+
+/// <summary>
+/// A shell command awaiting a policy decision.
+/// </summary>
+/// <param name="Command">The full command line that the agent wants to run.</param>
+/// <param name="WorkingDirectory">Optional working directory the command will execute in, if known.</param>
+public readonly record struct ShellRequest(string Command, string? WorkingDirectory = null);
```
westey-m: Unless we are using features like the comparison or `with` features of records, we've had recommendations from the .net folks to not use them, since they are rather expensive types compared to a simple class.  And if a POCO will do, it's much cheaper to just one one.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellPolicy.cs
```diff
@@ -0,0 +1,140 @@
...
+
+/// <summary>
+/// The outcome of a <see cref="ShellPolicy"/> evaluation.
+/// </summary>
+/// <param name="Allowed"><see langword="true"/> when the command may run.</param>
+/// <param name="Reason">Human-readable rationale; populated for both allow and deny when applicable.</param>
+public readonly record struct ShellDecision(bool Allowed, string? Reason = null)
```
westey-m: Also consider `ShellPolicyOutcome` or `ShellPolicyResult` for a name.  I can imagine that ShellDecision should mean more things than the outcome of a shell policy evaluation.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/CleanEnvironmentHelper.cs
```diff
@@ -0,0 +1,60 @@
...
+        "TMP",
+    };
+
+    /// <summary>
+    /// Strip everything from <paramref name="environment"/> except the entries named by
+    /// <see cref="PreservedVariables"/>. Lookup is case-insensitive so it works on both
+    /// Windows (case-insensitive env vars) and POSIX (case-sensitive but typed in the
+    /// expected case). Variables that aren't present in the input dictionary are skipped.
+    /// </summary>
+    public static void ApplyPreserved(IDictionary<string, string?> environment)
```
lokitoth: [Minor][Nit] Consider a name that better matches the behaviour - removing non-preserved.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/DockerShellTool.cs
```diff
@@ -0,0 +1,689 @@
...
+/// </para>
+/// <para>
+/// Persistent mode reuses <see cref="ShellSession"/> by launching
+/// <c>docker exec -i &lt;container&gt; bash --noprofile --norc</c> as the
+/// long-lived shell — the sentinel protocol works unchanged because the
+/// host process is still a bash REPL connected over pipes. Stateless mode
+/// runs each call in a fresh <c>docker run --rm</c>.
+/// </para>
+/// </remarks>
+public sealed class DockerShellTool : IAsyncDisposable, IShellExecutor
```
westey-m: Also consider `DockerShellExecutor` so it's easier to tell that it is an implementation of `IShellExecutor`

dotnet/src/Microsoft.Agents.AI.Tools.Shell/IShellExecutor.cs
```diff
@@ -0,0 +1,53 @@
...
+/// gating for untrusted-input scenarios.
+/// </para>
+/// <para>
+/// The interface is intentionally minimal so callers can plug in their own
+/// executor (Firecracker microVM, remote SSH, WASI runtime, etc.) without
+/// forking the framework. Mirrors the Python <c>ShellExecutor</c> Protocol
+/// in <c>agent_framework_tools.shell._executor_base</c>.
+/// </para>
+/// </remarks>
+public interface IShellExecutor : IAsyncDisposable
```
westey-m: Have you considered a base abstract class, since interfaces are not easily extensible, since changing them is a breaking change.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/DockerShellExecutor.cs
```diff
@@ -0,0 +1,689 @@
...
+/// runs each call in a fresh <c>docker run --rm</c>.
+/// </para>
+/// </remarks>
+public sealed class DockerShellTool : IAsyncDisposable, IShellExecutor
+{
+    /// <summary>Default container image. A small Microsoft-maintained Linux base.</summary>
+    public const string DefaultImage = "mcr.microsoft.com/azurelinux/base/core:3.0";
+
+    /// <summary>Default container user (nobody:nogroup on most distros).</summary>
+    public const string DefaultContainerUser = "65534:65534";
```
lokitoth: [Nit][Later?] We might want to make this a proper `record ContainerUser(string Uid, string Gid)` type to make it clear for the user how to configure it.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/DockerShellTool.cs
```diff
@@ -0,0 +1,689 @@
...
+    public const string DefaultImage = "mcr.microsoft.com/azurelinux/base/core:3.0";
+
+    /// <summary>Default container user (nobody:nogroup on most distros).</summary>
+    public const string DefaultContainerUser = "65534:65534";
+
+    /// <summary>Default Docker network mode (no network).</summary>
+    public const string DefaultNetwork = "none";
+
+    /// <summary>Default container memory limit.</summary>
+    public const string DefaultMemory = "512m";
```
lokitoth: [Minor][Nit] We should consider making this a number, and convert to the string format where appropriate.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/LocalShellTool.cs
```diff
@@ -0,0 +1,501 @@
...
+/// <b>Threat model.</b> The deny list is a guardrail, not a security boundary. Real isolation
+/// requires either (a) approval-in-the-loop, where every command is reviewed by a human via the
+/// harness <c>ToolApprovalAgent</c> (this is the default; see
+/// <see cref="AsAIFunction(string, string?, bool)"/>), or (b) container isolation
+/// (<c>DockerShellTool</c>). To produce an unapproved <see cref="AIFunction"/> you must pass
+/// <c>acknowledgeUnsafe: true</c> at construction; otherwise <see cref="AsAIFunction"/> will
+/// refuse to return a non-approval-gated function.
+/// </para>
+/// </remarks>
+public sealed class LocalShellTool : IAsyncDisposable, IShellExecutor
```
westey-m: Same as with Docker, consider `LocalShellExecutor` to match the interface name.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/DockerShellTool.cs
```diff
@@ -0,0 +1,689 @@
...
+public sealed class DockerShellTool : IAsyncDisposable, IShellExecutor
+{
+    /// <summary>Default container image. A small Microsoft-maintained Linux base.</summary>
+    public const string DefaultImage = "mcr.microsoft.com/azurelinux/base/core:3.0";
+
+    /// <summary>Default container user (nobody:nogroup on most distros).</summary>
+    public const string DefaultContainerUser = "65534:65534";
+
+    /// <summary>Default Docker network mode (no network).</summary>
+    public const string DefaultNetwork = "none";
```
lokitoth: [Minor][Nit] We should consider making this an enum or define a set of known constant Network Modes.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/IShellExecutor.cs
```diff
@@ -0,0 +1,53 @@
...
+{
+    /// <summary>
+    /// Eagerly initialize the backend. Idempotent; subsequent calls are
+    /// no-ops once the executor is started. For stateless executors this
+    /// is typically a no-op.
+    /// </summary>
+    /// <param name="cancellationToken">Cancellation token.</param>
+    Task StartAsync(CancellationToken cancellationToken = default);
```
SergeyMenshykh: Consider renaming StartAsync to InitializeAsync and CloseAsync to ShutdownAsync, since "Initialize" better reflects the one-time setup semantics rather than implying a continuous process, and "Shutdown" pairs naturally with it to describe tearing down backend resources.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellPolicy.cs
```diff
@@ -0,0 +1,140 @@
...
+/// (<c>Remove-Item -Recurse -Force</c>). The actual security boundary is
+/// approval-in-the-loop (see <see cref="LocalShellTool"/>) or container
+/// isolation (Docker/Firecracker, planned in a follow-up).
+/// </para>
+/// <para>
+/// Evaluation order: explicit allow patterns short-circuit; otherwise the
+/// command is checked against the deny list; otherwise the request is allowed.
+/// </para>
+/// </remarks>
+public sealed class ShellPolicy
```
lokitoth, line 135: We should consider making this an extensibility point so that users can inject their own policy logic? (Possibly as separate PR?)

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellResolver.cs
```diff
@@ -0,0 +1,204 @@
...
+/// Resolution order:
+/// <list type="bullet">
+/// <item><description>Windows: prefer <c>pwsh</c>, fall back to <c>powershell.exe</c>, then <c>cmd.exe</c>.</description></item>
+/// <item><description>Linux / macOS: prefer <c>/bin/bash</c>, fall back to <c>/bin/sh</c>.</description></item>
+/// <item><description>Override via the constructor argument or the <c>AGENT_FRAMEWORK_SHELL</c> environment variable.</description></item>
+/// </list>
+/// </remarks>
+internal static class ShellResolver
+{
+    internal const string EnvVarName = "AGENT_FRAMEWORK_SHELL";
```
lokitoth: [Minor] We should consider making this a public constant or even a `static readonly` so downstream users can rely on it to automatically building configurations / environments.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellResolver.cs
```diff
@@ -0,0 +1,204 @@
...
+        {
+            ShellKind.PowerShell => new[]
+            {
+                "-NoProfile",
+                "-NoLogo",
+                "-NonInteractive",
+                "-Command",
+                "-",
+            },
+            ShellKind.Cmd => throw new NotSupportedException(
```
lokitoth, line 194: Is there a reason we do not support Persistent shell for cmd?

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellSession.cs
```diff
@@ -0,0 +1,948 @@
...
+    private TaskCompletionSource<bool> _stdoutSignal = NewSignal();
+    private bool _stdoutClosed;
+
+    public ShellSession(
+        ResolvedShell shell,
+        string? workingDirectory,
+        bool confineWorkingDirectory,
+        IReadOnlyDictionary<string, string?>? environment,
+        bool cleanEnvironment,
+        int maxOutputBytes)
```
lokitoth, line 99: [Later]

We may want to consider an extensibility point that allows for a different compaction strategy. For example, running a SLM locally to subselect the relevant information from the output.

This would help avoid cases where there are output segments relevant in the middle of a long output stream which could get trimmed otherwise.

dotnet/tests/Microsoft.Agents.AI.Tools.Shell.UnitTests/ShellResolverTests.cs
```diff
@@ -0,0 +1,67 @@
...
+
+    private static ResolvedShell ResolveSingle(string binary) => ShellResolver.ResolveArgv(new[] { binary });
+
+    [Theory]
+    [InlineData("/bin/sh")]
+    [InlineData("/bin/dash")]
+    [InlineData("/bin/ash")]
+    [InlineData("/usr/bin/busybox")]
+    [InlineData("/usr/bin/zsh")]
+    [InlineData("/bin/ksh")]
```
lokitoth, line 24: Do these tests require having all of these shells installed? If so, may be worth moving them to Integration

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellPolicy.cs
```diff
@@ -0,0 +1,140 @@
...
+        if (command.Length == 0)
+        {
+            return ShellDecision.Deny("empty command");
+        }
+
+        foreach (var allow in this._allows)
+        {
+            if (allow.IsMatch(command))
+            {
+                return new ShellDecision(true, "matched allow pattern");
```
lokitoth: If a given request matches both Allow and Deny, should it be allowed or denied?

dotnet/src/Microsoft.Agents.AI.Tools.Shell/DockerShellTool.cs
```diff
@@ -0,0 +1,689 @@
...
+        TimeSpan? timeout = null,
+        int maxOutputBytes = DefaultMaxOutputBytes,
+        string dockerBinary = "docker")
+    {
+        if (maxOutputBytes <= 0)
+        {
+            throw new ArgumentOutOfRangeException(nameof(maxOutputBytes));
+        }
+
+        this._image = image ?? throw new ArgumentNullException(nameof(image));
```
westey-m:
```suggestion
        this._image = Throw.IfNull(image);
```

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellPolicy.cs
```diff
@@ -0,0 +1,140 @@
...
+public sealed class ShellPolicy
+{
+    /// <summary>
+    /// Gets a conservative default deny list. Documented as a guardrail only.
+    /// </summary>
+    public static IReadOnlyList<string> DefaultDenyList { get; } =
+    [
+        @"\brm\s+(?:-[a-zA-Z]*r[a-zA-Z]*\s+)?-?\s*-?-?\s*[\/]",
+        @"\brm\s+-rf?\s+~",
+        @":\(\)\s*\{",
```
lokitoth: It might be a good idea to annotate what each of these is designed to catch; some of them might not be immediately obvious at a glance.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellPolicy.cs
```diff
@@ -0,0 +1,140 @@
...
+        }
+        this._allows = allow;
+    }
+
+    /// <summary>
+    /// Evaluate <paramref name="request"/> and return a decision.
+    /// </summary>
+    /// <param name="request">The request to evaluate.</param>
+    /// <returns>An allow or deny decision.</returns>
+    public ShellDecision Evaluate(ShellRequest request)
```
lokitoth: We may need to do a bit more checking here:

- If the command contains encoded characters, e.g. `\xc3` (e.g. `$(echo -e "\x72\x6D\x20\x2D\x72\x66\x20\x2A")` turns into `$("rm -rf *")`, which executes it
- If the command contains inline command execution (could use a command, e.g. `sed` to construct a command string or substring)
- If the command contains inline envvars, it could be used to construct a bad command string partially, e.g. `$(A=r B=m C=- D=f E=*; echo "$A$B $C$A$D $E")` becomes `$("rm -rf *")`

dotnet/src/Microsoft.Agents.AI.Tools.Shell/DockerShellTool.cs
lokitoth: Can wait for later, but it seems like there is some shared logic between `DockerShellTool` and `LocalShellTool` which might be valuable to put into a `ShellToolBase` so other backends could be added by the user while inheriting things like the shell resolution logic, or shell policy execution.

SergeyMenshykh: Is inheriting IShellExecutor here necessary? Doesn't the class inherit from IShellExecutor?

dotnet/src/Microsoft.Agents.AI.Tools.Shell/DockerShellExecutor.cs
```diff
@@ -0,0 +1,689 @@
...
+/// <c>docker exec -i &lt;container&gt; bash --noprofile --norc</c> as the
+/// long-lived shell — the sentinel protocol works unchanged because the
+/// host process is still a bash REPL connected over pipes. Stateless mode
+/// runs each call in a fresh <c>docker run --rm</c>.
+/// </para>
+/// </remarks>
+public sealed class DockerShellTool : IAsyncDisposable, IShellExecutor
+{
+    /// <summary>Default container image. A small Microsoft-maintained Linux base.</summary>
+    public const string DefaultImage = "mcr.microsoft.com/azurelinux/base/core:3.0";
```
SergeyMenshykh, line 63: Do this this const and a few below need to be public?

dotnet/src/Microsoft.Agents.AI.Tools.Shell/DockerShellTool.cs
```diff
@@ -0,0 +1,689 @@
...
+    /// <param name="pidsLimit">Max processes inside the container.</param>
+    /// <param name="user">UID:GID. Defaults to <c>65534:65534</c> (nobody).</param>
+    /// <param name="readOnlyRoot">When <see langword="true"/> (default), the container root filesystem is read-only.</param>
+    /// <param name="extraRunArgs">Additional args appended to <c>docker run</c>.</param>
+    /// <param name="environment">Environment variables passed via <c>-e</c> to every command.</param>
+    /// <param name="policy">Optional <see cref="ShellPolicy"/>. Less critical than for <see cref="LocalShellTool"/> since the container provides isolation.</param>
+    /// <param name="timeout">Per-command timeout. <see langword="null"/> disables timeouts.</param>
+    /// <param name="maxOutputBytes">Per-stream cap before head+tail truncation.</param>
+    /// <param name="dockerBinary">Override (e.g. <c>podman</c>).</param>
+    public DockerShellTool(
```
SergeyMenshykh: Do we expect more parameters in the future? If that is the case, I would suggest creating a property for each parameter and removing the constructor, as it may become a source of breaking changes every time a new parameter needs to be added. Alternatively, they can be bundled into options class.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/DockerShellExecutor.cs
```diff
@@ -0,0 +1,689 @@
...
+
+    /// <summary>Run a single command inside the container.</summary>
+    /// <exception cref="ShellCommandRejectedException">Thrown when the policy denies the command.</exception>
+    public async Task<ShellResult> RunAsync(string command, CancellationToken cancellationToken = default)
+    {
+        if (command is null)
+        {
+            throw new ArgumentNullException(nameof(command));
+        }
+        var decision = this._policy.Evaluate(new ShellRequest(command, this._containerWorkdir));
```
SergeyMenshykh, line 224: Nit: It seems there is a pattern in the PR of not adding an empty line between the closing curly bracket and the next line.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/DockerShellExecutor.cs
```diff
@@ -0,0 +1,689 @@
...
+    /// (if any) is read-only, and no caller-supplied <c>extraRunArgs</c>
+    /// have been added. This is a configuration-shape check, not a
+    /// security guarantee; isolation still depends on the host kernel,
+    /// the container runtime, and the image. <see cref="AsAIFunction"/>
+    /// uses this signal to choose a default for <c>requireApproval</c>:
+    /// when the configuration has been relaxed it leaves approval
+    /// gating on, but you should always make the approval/policy
+    /// decision deliberately rather than relying on this default.
+    /// </summary>
+    public bool IsHardenedConfiguration =>
```
SergeyMenshykh: Does it need to be public?

dotnet/src/Microsoft.Agents.AI.Tools.Shell/IShellExecutor.cs
```diff
@@ -0,0 +1,53 @@
...
+    /// </summary>
+    /// <param name="cancellationToken">Cancellation token.</param>
+    Task InitializeAsync(CancellationToken cancellationToken = default);
+
+    /// <summary>
+    /// Tear down all backend resources. Idempotent; safe to call multiple
+    /// times.
+    /// </summary>
+    /// <param name="cancellationToken">Cancellation token.</param>
+    Task ShutdownAsync(CancellationToken cancellationToken = default);
```
SergeyMenshykh: Do we expect multiple calls to the InitializeAsync/ShutdownAsync pair during the lifetime of an executor instance, or will each be called only once - Initialize at the beginning and Shutdown at the end? If the latter is the case, then we can consider dropping ShutdownAsync since it's called from Dispose.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/LocalShellExecutor.cs
```diff
@@ -0,0 +1,501 @@
...
+        string? shell = null,
+        IReadOnlyList<string>? shellArgv = null,
+        string? workingDirectory = null,
+        bool confineWorkingDirectory = true,
+        IReadOnlyDictionary<string, string?>? environment = null,
+        bool cleanEnvironment = false,
+        ShellPolicy? policy = null,
+        TimeSpan? timeout = null,
+        int maxOutputBytes = DefaultMaxOutputBytes,
+        bool acknowledgeUnsafe = false)
```
westey-m: In other locations, we are using an optional options class where there are many optional params, since it's less disruptive to add new options to an options class than adding method parameters, if new ones need to be added.
It also means that we don't have weird ordering in a constructor where one new one is needed, but it logically shouldn't go a the end, but it's the only way to avoid breaking users.

SergeyMenshykh: Related: https://github.com/microsoft/agent-framework/pull/5604#discussion_r3203583807

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellPolicy.cs
alliscode: Fixed, thanks.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellPolicy.cs
alliscode: Fixed, thanks.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellResult.cs
alliscode: Fixed, thanks.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/DockerShellTool.cs
alliscode: Fixed, thanks.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/LocalShellTool.cs
alliscode: Fixed, thanks.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/IShellExecutor.cs
alliscode: Fixed, thanks.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellResolver.cs
alliscode: Fixed, thanks.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/DockerShellTool.cs
alliscode: Fixed, thanks.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/DockerShellTool.cs
alliscode: Fixed, thanks.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/DockerShellTool.cs
alliscode: Fixed, thanks.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/CleanEnvironmentHelper.cs
alliscode: Fixed, thanks.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellPolicy.cs
alliscode: Fixed, thanks.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellPolicy.cs
alliscode: Fixed, thanks.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/Microsoft.Agents.AI.Tools.Shell.csproj
```diff
@@ -0,0 +1,39 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <!-- Modern targets only; the underlying P/Invokes (setsid) and
+         async patterns are not validated against netstandard2.0/net472. -->
+    <TargetFrameworks>$(TargetFrameworksCore)</TargetFrameworks>
+    <RootNamespace>Microsoft.Agents.AI.Tools.Shell</RootNamespace>
+    <VersionSuffix>preview</VersionSuffix>
+    <Title>Microsoft Agent Framework - Shell Tools</Title>
+    <Description>Cross-platform shell tools for the Microsoft Agent Framework. Includes LocalShellTool and DockerShellTool with approval-in-the-loop semantics, plus ShellEnvironmentProvider for environment-aware system prompts.</Description>
```
westey-m: For description in particular, there is a default description already in the nuget-packge.props, so to override it, this needs to go below the nuget-package.props import below.

Most of the other package, just stick the title and description in a small property group below the import.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/DockerShellExecutor.cs
```diff
@@ -0,0 +1,699 @@
...
+/// </para>
+/// <para>
+/// Persistent mode reuses <see cref="ShellSession"/> by launching
+/// <c>docker exec -i &lt;container&gt; bash --noprofile --norc</c> as the
+/// long-lived shell — the sentinel protocol works unchanged because the
+/// host process is still a bash REPL connected over pipes. Stateless mode
+/// runs each call in a fresh <c>docker run --rm</c>.
+/// </para>
+/// </remarks>
+public sealed class DockerShellExecutor : IAsyncDisposable, IShellExecutor
```
westey-m: Is there an implicit expectation that the executors should only be used with a single user/session at a time?
E.g. if you have a single agent object in a service, and that service is shared by many different users/sessions, would that be a bug?  If so, we need to document that really well.  I'm guessing the stateless mode is generally safe, but not the stateful mode.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/Microsoft.Agents.AI.Tools.Shell.csproj
alliscode: Fixed, thanks.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/DockerShellTool.cs
alliscode: Fixed, thanks.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/DockerShellExecutor.cs
alliscode: Fixed, thanks.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/DockerShellExecutor.cs
alliscode: Fixed, thanks.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellExecutionException.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+
+namespace Microsoft.Agents.AI.Tools.Shell;
+
+/// <summary>
+/// Thrown when a shell command fails to launch or the shell session is unrecoverable.
+/// </summary>
+public sealed class ShellExecutionException : Exception
```
SergeyMenshykh: The exception doesn’t provide any additional value beyond the base one. Its only advantage is that it helps consumers identify that the exception originates from the shell component. How is this exception intended to be used? Do we really need it? Can't we just use standard .NET exceptions?

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellTimeoutException.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+
+namespace Microsoft.Agents.AI.Tools.Shell;
+
+/// <summary>
+/// Thrown when a shell command exceeds its configured timeout.
+/// </summary>
+public sealed class ShellTimeoutException : Exception
```
SergeyMenshykh: Related https://github.com/microsoft/agent-framework/pull/5604/changes#r3203976132

dotnet/src/Microsoft.Agents.AI.Tools.Shell/IShellExecutor.cs
alliscode: Fixed, thanks.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/DockerShellExecutor.cs
alliscode: Fixed, thanks.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/DockerShellTool.cs
alliscode: Fixed, thanks.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/IShellExecutor.cs
alliscode: Fixed, thanks.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/LocalShellExecutor.cs
alliscode: Fixed, thanks.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellResolver.cs
alliscode: It's hard to track delimited command boundaries in cmd. We can try again in a follow up if there is interest in it.

dotnet/tests/Microsoft.Agents.AI.Tools.Shell.UnitTests/ShellResolverTests.cs
alliscode: These tests don't require any shells installed. They go through `ResolveArgv(string[])`, which only does filename-based classification of the supplied path — no PATH probe, no `File.Exists` check, no Docker contact. They run in <50ms regardless of what's installed on the host.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/DockerShellTool.cs
alliscode: Will create a work item to follow up on this.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/DockerShellExecutor.cs
alliscode: Fixed, thanks. Removed `IsHardenedConfiguration` entirely; `AsAIFunction(requireApproval: null)` now always wraps in `ApprovalRequiredAIFunction`. Callers must explicitly pass `false` to opt out — container configuration alone is not a sufficient signal to safely auto-execute model-generated commands.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/DockerShellExecutor.cs
alliscode: That is correct, stateful mode is single session only. I'll add more docs.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellExecutionException.cs
alliscode: Fixed, thanks. Deleted both custom exception types. Process spawn failures and broken-pipe-to-long-lived-shell now throw IOException; ShellTimeoutException was declared but never thrown (the timeout path raises OperationCanceledException from the linked CTS), so it was unused dead surface.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellTimeoutException.cs
alliscode: Fixed, thanks. Deleted both custom exception types. Process spawn failures and broken-pipe-to-long-lived-shell now throw IOException; ShellTimeoutException was declared but never thrown (the timeout path raises OperationCanceledException from the linked CTS), so it was unused dead surface.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellPolicy.cs
alliscode: Good call, we add a follow up task for this

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellSession.cs
alliscode: Good idea! Will add a work item to track it.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellPolicy.cs
alliscode: Agreed — and we went further. We removed `DefaultDenyList` entirely (commit ff7d8e4100); `ShellPolicy` constructed with defaults now has an empty deny-list.

After surveying 11 competing agent frameworks (LangChain, AutoGen, OpenAI SDK, Claude Code, Goose, Continue.dev, OpenHands, Open Interpreter, Aider, smolagents, LangGraph), **none** use regex as a primary control — they all rely on container/VM isolation, human approval gating, or eliminating shell access. AutoGen v2 explicitly removed their regex deny-list and now just `UserWarning`s users to use Docker. Continue.dev is the only one doing deep static analysis (a real shell tokenizer with recursive command-substitution expansion) and even they escalate envvar-splicing cases to `requires permission` rather than blocking — confirming the class of bypass you flagged is fundamentally unsolvable at the static-analysis layer.

The `ShellPolicy` mechanism stays for operators who want site-specific patterns (e.g. blocking a prod hostname), but the framework no longer ships pretend-defense as a default. Approval gating + Docker remain the real boundaries.

alliscode: I've decided to remove the regex pattern matching entirely as I think it provides a false sense of security. The default will be an empty accept list and an empty deny list and the only "safe" practice is to have manual approvals for the actual commands being run.

westey-m (approved):

SergeyMenshykh (approved):

lokitoth (approved):

PR #5615, opened 2026-05-04
Python: Fix incorrect workflow timings in DevUI by adding `created_at` to executor events

### LINKED ISSUE #5545: Python: [Bug]: Workflow timings in DevUI are incorrect
### Description

The workflow timings shown when running a workflow in DevUI are incorrect.

Image

### Code Sample

1. Run the sample `python/samples/02-agents/devui/in_memory_mode.py`
2. Run the "Text Transformer" workflow

Although the workflow finishes almost instantly, the workflow timings suggest that the second step starts 3 seconds after the first.

```markdown

```

### Package Versions

agent-framework-core: 1.2.1, agent-framework-devui: 1.0.0b260428

### Python Version

3.13.11

### Additional Context

_No response_

moonbox3:
### Motivation and Context

Workflow steps in DevUI appeared to take seconds apart even when they completed nearly instantly. This was caused by the frontend falling back to a synthetic timestamp that forced a minimum 1-second gap between events, because executor-related events emitted by the backend contained no `created_at` field.

Fixes #5545

### Description

The root cause was two-fold: `CustomResponseOutputItemAddedEvent` and `CustomResponseOutputItemDoneEvent` had no `created_at` field, so the mapper emitted them without timestamps; and `workflow-view.tsx` synthesized `_uiTimestamp` using `Math.max(Math.floor(Date.now()/1000), lastTimestamp + 1)` — integer-second precision with a forced +1 s gap — never consulting `event.created_at`. The fix adds an optional `created_at: float` field to both event models, populates it with `time.time()` in the mapper for all three executor event paths (invoked, completed, failed), and updates both call sites in `workflow-view.tsx` to prefer `event.created_at` over the synthetic fallback. A regression test was added to `test_mapper.py` to ensure executor-mapped events always carry a `created_at` timestamp going forward.

### Contribution Checklist

---


moonbox3:
<a href="https://github.com/microsoft/agent-framework/blob/eb30420253c0ecc9a32558b326abb2c71752126f/README.md">Python Test Coverage</a><details><summary>Python Test Coverage Report • </summary><table><tr><th>File</th><th>Stmts</th><th>Miss</th><th>Cover</th><th>Missing</th></tr><tbody><tr><td><b>TOTAL</b></td><td><b>31050</b></td><td><b>3592</b></td><td><b>88%</b></td><td>&nbsp;</td></tr></tbody></table><i>report-only-changed-files is enabled. No files were changed during this commit :)</i></details>

## Python Unit Test Overview
| 6235 | 30 :zzz: | 0 :x: | 0 :fire: | 1m 43s :stopwatch: |

python/packages/devui/frontend/src/components/features/workflow/workflow-view.tsx
```diff
@@ -576,17 +576,20 @@ export function WorkflowView({
             openAIEvent.type === "response.workflow_event.complete" // Fallback variant
           ) {
             setOpenAIEvents((prev) => {
-              // Generate unique timestamp for each event
+              // Prefer the event's own created_at for accurate workflow timings.
+              // Fall back to a synthesized timestamp only when created_at is absent.
+              const eventTimestamp =
+                "created_at" in openAIEvent && openAIEvent.created_at
```
moonbox3: This only checks for a top-level `created_at`, but the fallback workflow event shape (`response.workflow_event.completed`) is emitted by the mapper with `data.timestamp` instead (`_mapper.py:1241-1248`). `execution-timeline.tsx` uses `_uiTimestamp` for that fallback path (`:258-260`, `:343-369`), so those events still get fabricated `lastTimestamp + 1` gaps. The normalization here should also read `openAIEvent.data.timestamp` before falling back to a synthesized value.

moonbox3 (commented):
## Automated Code Review

### ✓ Correctness
> The PR correctly fixes the workflow timing bug by (1) adding `created_at: float | None = None` to `CustomResponseOutputItemAddedEvent` and `CustomResponseOutputItemDoneEvent`, (2) having the mapper populate `created_at=float(time.time())` for all executor events, and (3) updating `workflow-view.tsx` to prefer the real event timestamp over the synthesized `Math.max(baseTimestamp, lastTimestamp+1)` fallback. The `devuiStore.ts` already had the same `created_at`-preference pattern, so the PR makes `workflow-view.tsx` consistent. The old bug-characterization tests are properly replaced by a passing regression test in `test_mapper.py`. No correctness issues found.

### ✓ Security Reliability
> The fix is correct and complete: `created_at` fields are added to `CustomResponseOutputItemAddedEvent` and `CustomResponseOutputItemDoneEvent`, the mapper populates them with `float(time.time())` for all three executor event paths, and the frontend now prefers the event's own `created_at` over the synthesized `lastTimestamp + 1` value. One minor reliability concern: when `created_at` is present, the previous monotonic guarantee (`Math.max(baseTimestamp, lastTimestamp + 1)`) is bypassed. In the rare case of out-of-order event delivery or clock skew, two events could share the same `_uiTimestamp`. In practice, given sequential mapper execution and Python's `float(time.time())` sub-second precision, this risk is negligible. The test reorganisation is sound: the deleted bug-repro tests are superseded by the new passing regression test in `test_mapper.py`.

### ✓ Test Coverage
> The PR deletes a dedicated bug-documenting test file (test_workflow_timings_bug.py) and replaces it with a well-formed regression test in test_mapper.py, while also fixing the frontend timestamp logic to prefer created_at over synthesized values. The mapper already sets created_at=float(time.time()) unconditionally on all three executor event paths (lines 1059, 1092, 1126 of _mapper.py) and both custom model classes now declare created_at as an optional float field (lines 67, 81 of _openai_custom.py). The new test correctly uses an isolated mapper2 to avoid context-state cross-contamination for the failed path, which is an improvement over the deleted test. One minor gap: the deleted file contained test_custom_event_models_lack_created_at_field, which directly asserted that created_at is present in CustomResponseOutputItemAddedEvent.model_fields and CustomResponseOutputItemDoneEvent.model_fields. That model-level guard is not replaced by the new mapper-output test; if created_at were removed from either model, the mapper would raise a ValidationError (not produce None), so in practice the gap is caught differently, but the direct model-field assertion provided a clearer signal. No blocking issues found. The frontend change has no automated test coverage, but there is no existing frontend test infrastructure in the repo.

### ✗ Design Approach
> The PR moves the UI toward using real event timestamps, but the approach is still too narrow for the workflow timeline contract it supports today. `workflow-view.tsx` now only looks for a top-level `created_at`, while the still-supported `response.workflow_event.completed` fallback shape carries its time in `data.timestamp` from the mapper. Because `execution-timeline.tsx` renders those fallback events from `_uiTimestamp`, they will continue to pick up the old synthetic `Math.max(baseTimestamp, lastTimestamp + 1)` clock and can still show fabricated 1-second gaps.

### Flagged Issues
- [ ] `workflow-view.tsx` only normalizes top-level `created_at`, but the supported fallback event `response.workflow_event.completed` is produced by the mapper with `data.timestamp` (`_mapper.py:1241-1248`). `execution-timeline.tsx` uses `_uiTimestamp` for that fallback path (`:258-260`, `:343-369`), so those workflow events still receive the fabricated `lastTimestamp + 1` synthetic timeline. The timestamp normalization in `workflow-view.tsx` should also consume `openAIEvent.data.timestamp` before synthesizing a value.

### Suggestions
- The new test `test_executor_events_carry_created_at_timestamp` asserts `getattr(event, 'created_at', None) is not None` but does not verify the value is a valid Unix timestamp. Add `assert event.created_at > 0` to guard against zero or negative values being emitted.
- Using `created_at` directly removes the previous monotonic guarantee. If two events share the same float timestamp (e.g., rapid mapper calls on a low-resolution clock), `_uiTimestamp` will collide. Consider a tiebreaker: `const uniqueTimestamp = eventTimestamp !== undefined ? Math.max(eventTimestamp, lastTimestamp) : Math.max(baseTimestamp, lastTimestamp + 1);`
- The deleted `test_custom_event_models_lack_created_at_field` directly verified `created_at` is present in `model_fields` for both custom event classes. Consider adding a model-level assertion to `test_mapper.py` or a dedicated model test so that accidentally removing the field produces a clear named failure rather than a downstream `ValidationError`.
- Consider consolidating timestamp normalization into a single helper that reads `created_at`, then `data.timestamp`, then falls back to synthesis. This fixes the broader event-shape problem rather than only the custom executor-item path.

---

python/packages/devui/frontend/src/components/features/workflow/workflow-view.tsx
```diff
@@ -576,17 +576,32 @@ export function WorkflowView({
...
+              //   3. data.timestamp        (response.workflow_event.completed ISO string)
+              // Fall back to a synthesized timestamp only when none is present.
+              const anyEvent = openAIEvent as Record<string, unknown>;
+              const eventTimestamp: number | undefined =
+                typeof anyEvent["created_at"] === "number" && anyEvent["created_at"]
+                  ? (anyEvent["created_at"] as number)
+                  : typeof (anyEvent["response"] as Record<string, unknown> | undefined)?.["created_at"] === "number"
+                  ? ((anyEvent["response"] as Record<string, number>)["created_at"] as number)
+                  : typeof (anyEvent["data"] as Record<string, unknown> | undefined)?.["timestamp"] === "string"
+                  ? new Date((anyEvent["data"] as Record<string, string>)["timestamp"]).getTime() / 1000
```
moonbox3: `new Date(isoString).getTime()` returns `NaN` for strings that JS cannot parse (confirmed: `_mapper.py:1249` emits `datetime.now().isoformat()` — 6 fractional digits, no `Z`). `NaN / 1000` is still `NaN`, and `NaN !== undefined` is `true`, so the `eventTimestamp !== undefined` guard on line 602 does not filter it out. `Math.max(NaN, lastTimestamp)` then returns `NaN`, corrupting `_uiTimestamp` for downstream rendering (`execution-timeline.tsx:260`) and resetting the monotonic seed (`NaN || 0 → 0` at line 597). Use `Number.isFinite` to guard the parsed value.

```suggestion
                : typeof (anyEvent["data"] as Record<string, unknown> | undefined)?.["timestamp"] === "string"
                  ? (() => { const ms = new Date((anyEvent["data"] as Record<string, string>)["timestamp"]).getTime(); return Number.isFinite(ms) ? ms / 1000 : undefined; })()
```

moonbox3 (commented):
## Automated Code Review

### ✗ Correctness
> The new timestamp extraction logic correctly adds `response.created_at` (branch 2) and `data.timestamp` (branch 3) fallbacks, and the test additions are sound. However, both instances of the `data.timestamp` branch (lines 591 and 1022) produce a raw `new Date(...).getTime() / 1000` that can be `NaN` when the ISO string is unparseable (e.g., Python microsecond precision without a timezone indicator in edge environments). Because `NaN !== undefined` evaluates to `true`, the subsequent `eventTimestamp !== undefined` guard on lines 602–603 and 1033–1034 does not intercept it, `Math.max(NaN, lastTimestamp)` returns `NaN`, and `_uiTimestamp` is silently set to `NaN`, corrupting timeline ordering for every downstream consumer.

### ✗ Security Reliability
> The multi-source timestamp extraction logic introduces a reliability hole in the `data.timestamp` branch: `new Date(anyInvalidString).getTime() / 1000` silently produces `NaN`. Because `NaN !== undefined` is `true`, the `eventTimestamp !== undefined` guard does not catch this, so `Math.max(NaN, lastTimestamp)` propagates `NaN` as `_uiTimestamp`. This breaks timeline rendering for that event (execution-timeline.tsx line 260: `event._uiTimestamp * 1000` becomes `NaN`) and resets the monotonic ordering seed (`NaN || 0 → 0`) for all subsequent events. The same pattern appears in both duplicate blocks at lines 590–591 and 1021–1022. The Python test additions are correct and well-scoped.

### ✓ Test Coverage
> The Python-side test additions are solid: the new `created_at > 0` assertion on executor events is meaningful, and the model-field presence test guards against accidental field removal. The mapper does populate `created_at=float(time.time())` for both executor event types, so these tests will exercise real behaviour. One gap: the referenced failing test (`test_workflow_timings_bug.py`) is absent from the source tree (only a stale .pyc exists), meaning there is no test verifying the two new frontend timestamp-extraction paths (`response.created_at` and `data.timestamp`). The frontend changes themselves are logically sound, but they are entirely untested. Additionally, the `test_custom_output_item_event_models_have_created_at_field` test does not verify that the mapper actually *populates* `created_at` with a non-None value when constructing these events — it only checks the model schema. A companion integration-style assertion (similar to `test_executor_events_carry_created_at_timestamp`) would close that gap.

### ✗ Design Approach
> I found one design-level regression in the frontend timing change. Removing the forced monotonic increment for real timestamps fixes the visible 1-second gaps, but the fallback workflow timeline still treats `_uiTimestamp` as part of a synthetic per-run identifier. That means same-second executor events can now collapse onto the same synthetic ID, which reintroduces incorrect merging of fallback runs and their output buckets.

### Flagged Issues
- [ ] Lines 590–591 and 1021–1022: `new Date(isoString).getTime() / 1000` yields `NaN` for unparseable strings (e.g. Python's `datetime.now().isoformat()` with microsecond precision and no `Z`). `NaN !== undefined` is `true`, so the `eventTimestamp !== undefined` guard does not catch it. `Math.max(NaN, lastTimestamp)` returns `NaN`, setting `_uiTimestamp: NaN`, which breaks timeline rendering at `execution-timeline.tsx:260` and silently resets the monotonic ordering seed at `workflow-view.tsx:597` (`NaN || 0 → 0`). Guard the parsed value with `Number.isFinite()` before use.
- [ ] Line 603: replacing the forced monotonic increment with `Math.max(eventTimestamp, lastTimestamp)` removes the uniqueness guarantee the fallback timeline depends on. `execution-timeline.tsx:359–360` builds `syntheticItemId` from `executorId + uiTimestamp`, and keys run output/state off that ID at lines 367, 382, and 408. Two runs of the same executor within the same second now share an identical `_uiTimestamp` and their fallback entries collide. Ordering and identity should be decoupled — use `sequence_number` or a local counter to guarantee uniqueness.

### Suggestions
- Add `test_workflow_timings_bug.py` to the source tree (currently only a stale `.pyc` exists). Without it there is no regression guard for the `response.created_at` and `data.timestamp` frontend extraction paths.
- Extend `test_custom_output_item_event_models_have_created_at_field` (or add a companion test) to assert that `mapper.convert_event(...)` actually populates `created_at` with a non-`None`, positive value — mirroring the pattern in `test_executor_events_carry_created_at_timestamp`.

---

python/packages/devui/frontend/src/components/features/workflow/workflow-view.tsx
moonbox3: Fixed in `92be5ab8`.

python/packages/devui/frontend/src/components/features/workflow/workflow-view.tsx
```diff
@@ -576,17 +576,20 @@ export function WorkflowView({
...
                 prev.length > 0
                   ? (prev[prev.length - 1] as { _uiTimestamp?: number })
                       ._uiTimestamp || 0
                   : 0;
-              const uniqueTimestamp = Math.max(
-                baseTimestamp,
-                lastTimestamp + 1
-              );
+              const uniqueTimestamp =
+                eventTimestamp ?? Math.max(baseTimestamp, lastTimestamp + 1);
```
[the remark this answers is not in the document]
moonbox3: Fixed in `92be5ab8`.

python/packages/devui/frontend/src/components/features/workflow/workflow-view.tsx
```diff
@@ -992,14 +995,20 @@ export function WorkflowView({
...
+                : undefined;
             const baseTimestamp = Math.floor(Date.now() / 1000);
             const lastTimestamp =
               prev.length > 0
                 ? (prev[prev.length - 1] as { _uiTimestamp?: number })
                     ._uiTimestamp || 0
                 : 0;
-            const uniqueTimestamp = Math.max(baseTimestamp, lastTimestamp + 1);
+            const uniqueTimestamp =
+              eventTimestamp ?? Math.max(baseTimestamp, lastTimestamp + 1);
```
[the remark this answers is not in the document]
moonbox3: Fixed in `92be5ab8`.

python/packages/devui/frontend/src/components/features/workflow/workflow-view.tsx
moonbox3: Fixed in `92be5ab8`.

python/packages/devui/tests/devui/test_workflow_timings_bug.py
```diff
@@ -7,16 +7,17 @@
 1-second gaps between events, making instant workflows appear to take 3+ seconds.
 """
 
+from conftest import (
```
moonbox3: We do not need a specific new file called "*timings_bug" - we can add these tests to existing test files.

python/packages/devui/tests/devui/test_workflow_timings_bug.py
```diff
@@ -0,0 +1,110 @@
...
+            assert event.created_at > 0, (
+                f"{label} event {type(event).__name__} has non-positive created_at "
+                f"({event.created_at!r}); expected a valid Unix timestamp."
+            )
+
+
+async def test_rapid_workflow_events_have_no_top_level_timestamps(
+    mapper: MessageMapper, test_request: AgentFrameworkRequest
+) -> None:
+    """REGRESSION (#5545): response.workflow_event.completed events carry no top-level created_at.
```
moonbox3: Or specific callouts for tests to an issue. Generic is better.

python/packages/devui/tests/devui/test_workflow_timings_bug.py
moonbox3: Fixed in `eb304202`.

python/packages/devui/tests/devui/test_workflow_timings_bug.py
moonbox3: Fixed in `eb304202`.

eavanvalkenburg (approved):

TaoChenOSU (approved):

PR #5637, opened 2026-05-05
Python: refactor FoundryHostedAgentHistoryProvider onto Foundry SDK

eavanvalkenburg:
### Motivation and Context

This PR is the first piece of the **Python hosting Channels** work that landed as ADR / spec in #5549 and is being merged into the `feature/python-hosting` integration branch. It refactors `agent-framework-foundry-hosting` so the rest of the hosting-Channels stack can build on top of a clean shared layer.

### Description

* Splits the previously monolithic `_responses.py` into focused modules:
  * `_history_provider.py` — `FoundryHostedAgentHistoryProvider` exposed as the public history-provider entry point.
  * `_shared.py` / `_ids.py` — internal building blocks reused across Responses and the upcoming hosting Channels.
* Updates the public `__init__.py` exports to surface the new history-provider type and prepare for the hosting integration.
* Adds focused unit tests (`tests/test_history_provider.py`, expanded `tests/test_responses.py`).
* Workspace + lockfile updates so the package keeps resolving cleanly.

### Stack

This is the **first PR (PR-1 of 9)** in the hosting-Channels stack, all targeting `feature/python-hosting`:

| # | Branch | Status |
|---|---|---|
| **PR-1** | `refactor/foundry-hosted-agent-history-provider` | **this PR** |
| PR-2 | `feat/hosting-core` | independent of PR-1 |
| PR-3..7 | per-channel packages | depend on PR-2 |
| PR-8 | `feat/hosting-samples` | depends on PR-1 + PR-2..7 |

PR-1 is independent of all the others — can merge in any order relative to PR-2.

### Contribution Checklist

- [ ] **Is this a breaking change?** No — internal refactor; the only public surface change is the new `FoundryHostedAgentHistoryProvider` re-export.

### DIFF
--- a/python/packages/foundry_hosting/agent_framework_foundry_hosting/_history_provider.py
+++ b/python/packages/foundry_hosting/agent_framework_foundry_hosting/_history_provider.py
@@ -1,0 +1,758 @@
+# Copyright (c) Microsoft. All rights reserved.
+
+"""Foundry Hosted Agent history provider.
+
+A standalone :class:`agent_framework.HistoryProvider` implementation that
+sources conversation history from the Foundry Hosted Agent storage backend.
+
+Transport is delegated to the SDK's
+:class:`azure.ai.agentserver.responses.FoundryStorageProvider` (when running
+inside a Foundry Hosted Agent container) or
+:class:`azure.ai.agentserver.responses.InMemoryResponseProvider` (for local
+development). Both implement the same read/write surface
+(``get_history_item_ids`` / ``get_items`` / ``create_response``), so this
+provider's persistence logic stays backend-agnostic.
+
+Allowed dependencies (deliberately narrow):
+
+* :mod:`agent_framework` (core, for ``HistoryProvider`` / ``Message``)
+* :mod:`azure.ai.agentserver.responses` (for the storage backends,
+  ``IsolationContext`` typing, and ``OutputItem`` deserialization)
+* :mod:`azure.core.credentials_async` (typing of token credentials)
+
+It MUST NOT depend on any ``agent_framework_hosting*`` package at module
+import time. (The host's isolation contextvar is consulted lazily via an
+``import`` inside :func:`_host_isolation` so the dependency stays soft.)
+
+Environment variables read:
+
+* ``FOUNDRY_HOSTING_ENVIRONMENT`` — non-empty marks "running inside Foundry"
+  and selects the SDK-backed storage transport.
+* ``FOUNDRY_PROJECT_ENDPOINT`` — base URL of the Foundry project; required
+  when running hosted unless an explicit ``endpoint=`` is supplied.
+* ``FOUNDRY_AGENT_NAME`` / ``FOUNDRY_AGENT_VERSION`` — stamped onto the
+  ``agent_reference`` field of every persisted response envelope.
+* ``FOUNDRY_AGENT_SESSION_ID`` — used as a chain anchor when the channel
+  did not bind a per-request ``previous_response_id``.
+* ``MODEL_DEPLOYMENT_NAME`` / ``AZURE_AI_MODEL_DEPLOYMENT_NAME`` — model
+  field stamped on the persisted envelope (must match a real deployment).
+
+Local fallback: when ``FOUNDRY_HOSTING_ENVIRONMENT`` is unset, the provider
+transparently falls back to :class:`InMemoryResponseProvider` so the same
+agent code runs in dev.
+"""
+
+from __future__ import annotations
+
+import contextlib
+import logging
+import os
+import time
+from contextlib import contextmanager
+from contextvars import ContextVar
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Any, ClassVar
+
+from agent_framework import HistoryProvider, Message
+from azure.ai.agentserver.responses import (
+    FoundryStorageProvider,
+    FoundryStorageSettings,
+    InMemoryResponseProvider,
+    IsolationContext,
+)
+from azure.ai.agentserver.responses._id_generator import IdGenerator
+from azure.ai.agentserver.responses.models import OutputItem, ResponseObject
+from azure.ai.agentserver.responses.store._foundry_errors import (  # pyright: ignore[reportPrivateUsage]
+    FoundryBadRequestError,
+    FoundryResourceNotFoundError,
+)
+
+from ._shared import (
+    _messages_to_output_items,  # pyright: ignore[reportPrivateUsage]
+    _output_items_to_messages,  # pyright: ignore[reportPrivateUsage]
+)
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator, Sequence
+
+    from azure.core.credentials_async import AsyncTokenCredential
+
+logger = logging.getLogger(__name__)
+
+# Environment variable names — re-declared (not imported) so this module
+# stays decoupled from the private ``azure.ai.agentserver.core._config``
+# constants while still matching them exactly.
+_ENV_FOUNDRY_HOSTING_ENVIRONMENT = "FOUNDRY_HOSTING_ENVIRONMENT"
+_ENV_FOUNDRY_PROJECT_ENDPOINT = "FOUNDRY_PROJECT_ENDPOINT"
+
+# Per-request isolation context.  The owning Channel is expected to set this
+# from the inbound request (e.g. user / tenant headers) for the duration of
+# an ``agent.run(...)`` call.  When unset, requests are made without
+# isolation headers (matches how ``ResponseContext`` behaves with no
+# ``IsolationContext``).
+_isolation_var: ContextVar[IsolationContext | None] = ContextVar(
+    "agent_framework_foundry_hosting_isolation",
+    default=None,
+)
+
+
+def set_current_isolation(isolation: IsolationContext | None) -> Any:
+    """Set the per-request isolation context for downstream history calls.
+
+    Channels that drive an agent backed by :class:`FoundryHostedAgentHistoryProvider`
+    should call this before invoking ``agent.run(...)`` and reset the token
+    afterwards.
+
+    Args:
+        isolation: The isolation context to associate with the current
+            ``contextvars`` context, or ``None`` to clear it.
+
+    Returns:
+        A token suitable for :func:`reset_current_isolation` that restores
+        the previous value.
+    """
+    return _isolation_var.set(isolation)
+
+
+def reset_current_isolation(token: Any) -> None:
+    """Restore a previously-saved isolation context.
+
+    Args:
+        token: A token returned by :func:`set_current_isolation`.
+    """
+    _isolation_var.reset(token)
+
+
+def get_current_isolation() -> IsolationContext | None:
+    """Return the isolation context bound to the current async context, if any.
+
+    Returns:
+        The :class:`IsolationContext` for the current request, or ``None``
+        when no channel has set one.
+    """
+    return _isolation_var.get()
+
+
+@dataclass(frozen=True)
+class _RequestContext:
+    """Per-request anchors the host binds before invoking the agent.
+
+    ``response_id`` is the id this provider's :meth:`save_messages` call
+    will write under, so the channel and the storage backend agree on
+    one stable handle per turn (the channel surfaces the same id on the
+    response envelope, the next turn arrives with this value as
+    ``previous_response_id`` and the chain walks).
+
+    ``previous_response_id`` is the prior turn's anchor (``None`` on
+    first turn). Used to seed ``history_item_ids`` on the new write so
+    the storage chain stays connected, and to load history without
+    needing to know the channel's session minting convention.
+
+    Per-request Foundry isolation keys (the
+    ``x-agent-{user,chat}-isolation-key`` headers) are *not* carried
+    here; the host's own ASGI middleware lifts them off every inbound
+    HTTP request into a contextvar
+    (:func:`agent_framework_hosting.get_current_isolation_keys`) which
+    this provider consults at storage-call time. Keeping the headers
+    out of the per-request bind means channels never have to import
+    Foundry-specific types and the host owns the (intentional) coupling
+    to those two well-known headers.
+    """
+
+    response_id: str
+    previous_response_id: str | None
+
+
+_request_var: ContextVar[_RequestContext | None] = ContextVar(
+    "agent_framework_foundry_hosting_request",
+    default=None,
+)
+
+
+@contextmanager
+def bind_request_context(
+    *,
+    response_id: str,
+    previous_response_id: str | None = None,
+    **_unused: Any,
+) -> "Iterator[None]":
+    """Bind the per-request response-chain anchors for this provider.
+
+    Intended for the host (or any caller orchestrating an
+    ``agent.run(...)``) to call immediately before invocation, so the
+    provider's :meth:`save_messages` writes under a known, stable
+    ``response_id`` (the same one the channel surfaces to the client)
+    and walks ``previous_response_id`` for history continuity. Unknown
+    keyword arguments are accepted and ignored so the host can extend
+    the ``ChannelRequest.attributes`` contract without breaking existing
+    providers. Foundry isolation keys flow through a separate
+    host-installed contextvar; see the class docstring on
+    :class:`_RequestContext`.
+
+    The binding is scoped to the current ``contextvars.Context``, so
+    concurrent requests in the same process do not interfere.
+    """
+    token = _request_var.set(
+        _RequestContext(
+            response_id=response_id,
+            previous_response_id=previous_response_id,
+        )
+    )
+    try:
+        yield
+    finally:
+        _request_var.reset(token)
+
+
+def get_current_request_context() -> _RequestContext | None:
+    """Return the per-request response chain anchors, if bound."""
+    return _request_var.get()
+
+
+def _host_isolation() -> "IsolationContext | None":
+    """Lift the host-bound isolation contextvar into our local type.
+
+    The host installs an ASGI middleware that reads
+    ``x-agent-{user,chat}-isolation-key`` off every inbound HTTP request
+    and stores them in a generic ``IsolationKeys`` slot on a contextvar
+    we import from :mod:`agent_framework_hosting`. We translate it into
+    our :class:`IsolationContext` shape on demand so the provider stays
+    in charge of the storage-side type while the host stays free of any
+    Foundry-specific dependencies.
+    """
+    # Soft dep: ``agent_framework_hosting`` may not be installed (this
+    # provider is also usable standalone). The whole block is wrapped in
+    # ``# pyright: ignore`` so the optional import does not block type
+    # checking when the package isn't on sys.path; when it is, pyright
+    # picks up the real types automatically.
+    try:
+        from agent_framework_hosting import (  # pyright: ignore[reportMissingImports]
+            get_current_isolation_keys,  # pyright: ignore[reportUnknownVariableType]
+        )
+    except ImportError:  # pragma: no cover - hosting is a soft dep
+        return None
+    keys = get_current_isolation_keys()  # pyright: ignore[reportUnknownVariableType]
+    if keys is None or keys.is_empty:  # pyright: ignore[reportUnknownMemberType]
+        return None
+    return IsolationContext(
+        user_key=keys.user_key,  # pyright: ignore[reportUnknownMemberType, reportUnknownArgumentType]
+        chat_key=keys.chat_key,  # pyright: ignore[reportUnknownMemberType, reportUnknownArgumentType]
+    )
+
+
+# Type alias for the storage backend surface this provider depends on.
+# Both ``FoundryStorageProvider`` and ``InMemoryResponseProvider`` from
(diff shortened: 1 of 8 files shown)


python/packages/foundry_hosting/agent_framework_foundry_hosting/_history_provider.py
```diff
@@ -0,0 +1,949 @@
...
+        response = ResponseObject(response_body)
+
+        try:
+            await backend.create_response(
+                response,
+                input_items=input_items,
+                history_item_ids=history_item_ids,
+                isolation=isolation,
+            )
+        except Exception as exc:
```
moonbox3: This except hides:

- transient network, TLS, or DNS failures against Foundry storage (turn lost, no retry, caller never notified)
- 401 or 403 from an expired or misscoped credential (every subsequent turn keeps "succeeding" while writing nothing)
- 5xx from Foundry storage (treated identically to validation errors)
- schema-validation 400s that indicate a real bug in the payload builder above (the surrounding comments enumerate at least three required fields where omission yields a 400, so a regression in this builder will now be invisible at runtime)
- any `AttributeError` or `TypeError` from the payload construction itself if `response.output` items mutate underneath us

Shall we narrow the catch to the storage error types you already handle on the read side and re-raise everything else so transport and auth failures surface to the caller? If the intent is truly best-effort persistence, document that contract on the class and surface a counter or callback so operators can observe the silent-drop rate; today we don't get this signal.

python/packages/foundry_hosting/agent_framework_foundry_hosting/_history_provider.py
```diff
@@ -0,0 +1,949 @@
...
+            "local_shell_call": IdGenerator.new_local_shell_call_item_id,
+            "local_shell_call_output": IdGenerator.new_local_shell_call_output_item_id,
+            "mcp_call": IdGenerator.new_mcp_call_item_id,
+            "mcp_list_tools": IdGenerator.new_mcp_list_tools_item_id,
+            "mcp_approval_request": IdGenerator.new_mcp_approval_request_item_id,
+            "mcp_approval_response": IdGenerator.new_mcp_approval_response_item_id,
+            "custom_tool_call": IdGenerator.new_custom_tool_call_item_id,
+            "custom_tool_call_output": IdGenerator.new_custom_tool_call_output_item_id,
+        }
+        for item in all_items:
```
moonbox3, line 804: The block immediately above this loop comments that free-form `{response_id}_itm_N` ids are rejected by the storage backend with an opaque HTTP 500 because the partition-key extractor cannot parse them, and that these helpers re-stamp every item id into the Foundry `caresp_*`-style format. If the assignment raises (the suppress catches `AttributeError` and `TypeError`, both plausible against an SDK model whose `id` attribute backing changes), the item silently retains the synthetic prefix-based id from `_messages_to_output_items`, the storage backend rejects the entire `create_response`, and the bare `except Exception` further down at line 919 turns that into a logged-and-swallowed failure. The agent reports success, the chain is broken, the next turn loses history.

If id assignment is genuinely allowed to fail on some item subtype, should we, at minimum, log a WARNING with the offending type so the data loss is observable? If it is not expected to fail, should we drop the suppress and let the error propagate so the test suite catches a future SDK change?

python/packages/foundry_hosting/agent_framework_foundry_hosting/_shared.py
```diff
@@ -0,0 +1,1287 @@
...
+    (citations, reasoning, tool results, …) and any extras Foundry
+    layered on top of the declared schema.
+
+    A best-effort ``dict(...)`` is used so failure to snapshot (e.g. a
+    non-mapping subclass surfacing in the future) degrades gracefully to
+    the lossy-but-functional synthesise-from-text path rather than
+    crashing the read.
+    """
+    try:
+        raw = dict(item)
```
moonbox3, line 237: The docstring justifies this as best-effort so reads degrade to a lossy-but-functional synthesise-from-text path rather than crashing. The problem is the degradation is completely invisible: there is no log, no counter, no marker on the message that the raw snapshot is missing. The next time `_message_to_output_item` runs it will silently emit a synthesised shape that loses item ids, citations, reasoning content, tool-result extras, and any Foundry-layered fields the docstring explicitly says this snapshot exists to preserve. A user noticing missing citations on round-tripped history has no breadcrumb pointing here.

`except Exception` also hides categories that should not be swallowed at all (`MemoryError` on a giant payload, attribute errors from a partially-initialised SDK model that indicate a real upstream bug). Narrow to the type the docstring actually claims to defend against (`TypeError`) and log so the lossy fallback is observable.

```suggestion
    try:
        raw = dict(item)
    except TypeError:
        logger.warning(
            "_capture_raw: SDK item %r is not mapping-like; round-tripping without raw snapshot",
            type(item).__name__,
            exc_info=True,
        )
        return message
```

python/packages/foundry_hosting/agent_framework_foundry_hosting/_ids.py
```diff
@@ -0,0 +1,72 @@
...
+from azure.ai.agentserver.responses._id_generator import IdGenerator
+
+__all__ = [
+    "foundry_item_id",
+    "foundry_response_id",
+    "foundry_response_id_factory",
+]
+
+
+def foundry_response_id(previous_response_id: str | None = None) -> str:
```
moonbox3, line 32: The three helpers exported from this module (`foundry_response_id`, `foundry_response_id_factory`, `foundry_item_id`) are part of the new public surface re-exported from `agent_framework_foundry_hosting.__init__`, and the docstrings make load-bearing claims that are entirely unchecked: `foundry_response_id` must return a `caresp_*`-shaped id with an embedded 18-char partition key (storage returns HTTP 500 if it does not), it must reuse `previous_response_id`'s partition key when one is supplied so chained writes co-locate, and `foundry_item_id` must return `None` for reference-only types. None of these helpers are referenced from any test in the package. If the upstream `IdGenerator.new_response_id` or `new_item_id` contracts shift, every consumer of these helpers breaks at runtime against the live Foundry backend with no unit-test signal.

python/packages/foundry_hosting/agent_framework_foundry_hosting/_history_provider.py
```diff
@@ -0,0 +1,949 @@
...
+        # Build the wire payload to match exactly what the agentserver
+        # runtime emits via ``_extract_response_snapshot_from_events``
+        # for a synthetic ``status=completed`` snapshot:
+        #
+        #   {id, object, output, created_at, [model], agent_reference,
+        #    status, completed_at, [agent_session_id]}
+        #
+        # ``previous_response_id`` is appended when chaining; the runtime
+        # threads it through the same code path.
+        now = int(time.time())
```
moonbox3, line 902: The refactor stamps a precise wire payload onto every persisted `ResponseObject` starting at this line. The inline comments explicitly call out that the storage backend rejects with opaque HTTP 500 or HTTP 400 `invalid_payload` when `background`, `parallel_tool_calls`, `instructions`, or `agent_reference` are missing or extras leak in, and that `model` must come from `MODEL_DEPLOYMENT_NAME` or `AZURE_AI_MODEL_DEPLOYMENT_NAME`. I don't see great test coverage for this?

`test_save_messages_writes_to_backend_when_bound` only asserts `response.id`, `response.conversation`, and that `output[0].type == "output_message"`. The required-by-storage fields, the env-var-driven stamping (`FOUNDRY_AGENT_NAME`, `FOUNDRY_AGENT_VERSION`, `FOUNDRY_AGENT_SESSION_ID`, `MODEL_DEPLOYMENT_NAME`), and the rule that `version`, `agent_session_id`, and `model` are omitted (not stamped as `None`) when their env vars are unset are all unchecked. A future tweak that drops `parallel_tool_calls=False` or sets `agent_reference["version"] = None` instead of omitting it will compile, pass tests, and silently break every hosted deploy.

python/packages/foundry_hosting/agent_framework_foundry_hosting/_history_provider.py
```diff
@@ -0,0 +1,949 @@
...
+        # wrote on the previous turn, so we know it's storage-valid.
+        anchor = bound.previous_response_id if bound is not None else None
+        if anchor is None and session_id and session_id.startswith(("caresp_", "resp_")):
+            anchor = session_id
+        if anchor is None:
+            # The Foundry Hosted Agent runtime stamps the previous turn's
+            # response id into ``FOUNDRY_AGENT_SESSION_ID`` for the
+            # following turn's container, so we can walk back from it
+            # directly without keeping any cross-request state ourselves.
+            env_session = os.environ.get("FOUNDRY_AGENT_SESSION_ID") or None
```
moonbox3: Both `get_messages` (lines 612-618) and `save_messages` consult `FOUNDRY_AGENT_SESSION_ID` as a last-resort chain anchor when no host-bound `previous_response_id` exists and the per-call `session_id` is not `caresp_*`-shaped. The docstring describes this as the production path the Foundry runtime relies on for session-bound containers: the runtime stamps the previous turn's response id into `FOUNDRY_AGENT_SESSION_ID` for the following turn's container, so each new container can chain back to it directly. No test sets this env var. A regression that moves the lookup, mistypes the prefix check, or stops gating on `caresp_/resp_` prefix would silently make hosted multi-turn conversations forget every prior turn, with no unit-test signal until production traffic hits it.

python/packages/foundry_hosting/agent_framework_foundry_hosting/_history_provider.py
eavanvalkenburg: Narrowed to `FoundryStorageError` (covers `FoundryBadRequestError`, `FoundryResourceNotFoundError`, `FoundryApiError`) and re-raise everything else (network/TLS/DNS, expired-credential 401/403, payload-builder `AttributeError`/`TypeError`). Also added an instance-level `failed_writes` counter operators can poll / surface in health probes. Documented the best-effort contract inline. New tests cover both the swallow path (bumps counter) and the propagate path (counter stays 0). Fixed in 20cbc3e29.

python/packages/foundry_hosting/agent_framework_foundry_hosting/_history_provider.py
eavanvalkenburg: Dropped the `contextlib.suppress`. The id-stamping loop only runs against the type discriminators in `ITEM_ID_FACTORY` (all of which are `MutableMapping`-backed SDK models with `__setattr__` wired to dict-set), so the assignment is expected to succeed. Letting it raise surfaces SDK contract changes to the test suite instead of silently corrupting the chain. Removed unused `import contextlib`. Fixed in 20cbc3e29.

python/packages/foundry_hosting/agent_framework_foundry_hosting/_shared.py
eavanvalkenburg: Applied the suggested narrowing — `_capture_raw` now catches just `TypeError` and emits a `WARNING` with traceback when the lossy fallback kicks in, so a future regression that loses citations / reasoning content / Foundry-layered fields is observable in operator logs. Fixed in 20cbc3e29.

python/packages/foundry_hosting/agent_framework_foundry_hosting/_ids.py
eavanvalkenburg: Added `TestFoundryIdHelpers` covering all three public helpers:

- `foundry_response_id()` mints a `caresp_*` id with the embedded partition-key segment storage requires.
- `foundry_response_id(previous_response_id=...)` reuses the prior partition key so chained writes co-locate.
- `foundry_response_id_factory()` returns the function reference `ResponsesChannel(response_id_factory=...)` plugs in.
- `foundry_item_id(item, response_id)` returns a typed-prefix id with the partition-key hint for known `Item` types and `None` for unrecognised / reference-only types (matches the SDK helper's contract).

`_ids.py` is now at full coverage. Fixed in 20cbc3e29.

python/packages/foundry_hosting/agent_framework_foundry_hosting/_history_provider.py
eavanvalkenburg: Added `TestSaveMessagesWirePayload` with four cases asserting the wire payload that lands in `backend.create_response`:

1. Required-by-storage fields are always present: `background=False`, `parallel_tool_calls=False`, `instructions=""`, `agent_reference={'type': 'agent_reference', 'name': ...}`.
2. With env vars unset, `model` / `agent_session_id` / `agent_reference.version` are omitted entirely (NOT stamped to `None` — storage rejects `model: null`).
3. With `FOUNDRY_AGENT_NAME` / `FOUNDRY_AGENT_VERSION` / `FOUNDRY_AGENT_SESSION_ID` / `MODEL_DEPLOYMENT_NAME` set, all four values land on the envelope.
4. `MODEL_DEPLOYMENT_NAME` wins over `AZURE_AI_MODEL_DEPLOYMENT_NAME`; the latter is the local-dev fallback.

A future tweak that drops `parallel_tool_calls=False` or sets `agent_reference['version'] = None` instead of omitting it will now fail the unit suite. Fixed in 20cbc3e29.

python/packages/foundry_hosting/agent_framework_foundry_hosting/_history_provider.py
eavanvalkenburg: Added `TestFoundryAgentSessionIdAnchor` covering the env-var chain anchor on both code paths:

- `get_messages` falls back to `FOUNDRY_AGENT_SESSION_ID` when the host binding is absent and the per-call `session_id` is opaque, walking history from the env-supplied anchor.
- A non-`caresp_*`/`resp_*` env value is rejected before it can reach storage (defence-in-depth against the partition-key extractor's HTTP 500).
- `save_messages` uses the env anchor when no host binding supplies a previous_response_id.
- A host-bound `previous_response_id` wins over the env anchor (binding is the authoritative chain seed for the request).

A regression that moves the lookup, mistypes the prefix check, or stops gating on `caresp_*`/`resp_*` will now fail unit tests. Fixed in 20cbc3e29.

python/packages/foundry_hosting/agent_framework_foundry_hosting/_history_provider.py
```diff
@@ -0,0 +1,980 @@
...
+
+    @staticmethod
+    def is_hosted_environment() -> bool:
+        """Return ``True`` when running inside a Foundry Hosted Agent container.
+
+        Detection uses the ``FOUNDRY_HOSTING_ENVIRONMENT`` environment
+        variable, the same signal :class:`ResponsesAgentServerHost` uses to
+        switch between hosted and local storage backends.
+        """
+        return bool(os.environ.get(_ENV_FOUNDRY_HOSTING_ENVIRONMENT))
```
TaoChenOSU: This will break if the platform decides to change the env variable. We should use `AgentConfig.is_hosted()`

python/packages/foundry_hosting/agent_framework_foundry_hosting/_history_provider.py
```diff
@@ -0,0 +1,980 @@
...
+            storage backend rejects them with HTTP 400 "Malformed
+            identifier".
+
+            When ``local_storage_root`` is configured (and the provider
+            is running outside a Foundry Hosted Agent container), this
+            method instead delegates to a per-isolation
+            :class:`FileHistoryProvider` and ``session_id`` is used as
+            the literal file stem.
+        """
+        isolation = kwargs.get("isolation") or _host_isolation() or get_current_isolation()
```
TaoChenOSU, line 616: I am not sure if we should manage the context isolation in our layer. For hosting scenario, isolation should be managed by the infrastructure. For local scenario, why would it require isolation?

python/packages/foundry_hosting/agent_framework_foundry_hosting/_history_provider.py
```diff
@@ -0,0 +1,980 @@
...
+        # wrote on the previous turn, so we know it's storage-valid.
+        anchor = bound.previous_response_id if bound is not None else None
+        if anchor is None and session_id and session_id.startswith(("caresp_", "resp_")):
+            anchor = session_id
+        if anchor is None:
+            # The Foundry Hosted Agent runtime stamps the previous turn's
+            # response id into ``FOUNDRY_AGENT_SESSION_ID`` for the
+            # following turn's container, so we can walk back from it
+            # directly without keeping any cross-request state ourselves.
+            env_session = os.environ.get("FOUNDRY_AGENT_SESSION_ID") or None
```
TaoChenOSU: The session ID is bound to the container, not the conversation. I don't think the session ID will give you any history. Has this been tested in a hosted environment?

python/packages/foundry_hosting/agent_framework_foundry_hosting/_history_provider.py
```diff
@@ -0,0 +1,980 @@
...
+            if env_session and env_session.startswith(("caresp_", "resp_")):
+                anchor = env_session
+        if anchor is None:
+            # No walkable anchor → fresh conversation, nothing to load.
+            return []
+
+        backend = self._resolve_backend()
+
+        try:
+            item_ids = await backend.get_history_item_ids(
```
TaoChenOSU, line 642: The `ResponseContext` provides a `get_history` method that handles retrieving the items and isolation. Could we reuse that?

TaoChenOSU: I'd like to actually explore the idea of encapsulating the foundry history provider into our own provider abstraction.

TaoChenOSU (approved):

python/packages/foundry_hosting/agent_framework_foundry_hosting/_history_provider.py
eavanvalkenburg: Good call — switched `is_hosted_environment()` to delegate to `AgentConfig.from_env().is_hosted` so the detection rule stays in lockstep with the SDK; if Foundry ever renames the underlying signal the SDK update is picked up here without a code change. Dropped the local `_ENV_FOUNDRY_HOSTING_ENVIRONMENT` constant. Fixed in 93dc14549.

python/packages/foundry_hosting/agent_framework_foundry_hosting/_history_provider.py
eavanvalkenburg: Agreed that for the hosted scenario the platform owns isolation — and that's exactly how this layer behaves: the SDK's `IsolationContext` is consulted via `get_current_isolation()` (set by the platform's request middleware) and forwarded to the storage calls as `isolation=...`. The local fallback chain (`kwargs['isolation']` → `_host_isolation()` → SDK contextvar) exists for two non-hosted cases:
1. **Local file storage** (`local_storage_root` set): the isolation is used to partition the on-disk layout `{root}/{user}/{chat}/{session}.jsonl` so different users don't stomp each other's history when the same Python process serves multiple end users (e.g., a Telegram bot in dev).
2. **Local in-memory dev** (no `local_storage_root`): single-user dev only, where isolation is harmless `None` and the chain just no-ops.

We're not introducing isolation behaviour the platform doesn't already provide — the contextvar lookup short-circuits to the SDK's value when running hosted. Happy to add a docstring note making this explicit if it would help. Resolving for now since the layered fallback is intentional and consistent with the SDK.

python/packages/foundry_hosting/agent_framework_foundry_hosting/_history_provider.py
eavanvalkenburg: You're right — `FOUNDRY_AGENT_SESSION_ID` is bound to the container, not the conversation, so walking back from it as a `previous_response_id` would silently merge unrelated conversations across container restarts. Removed the env-as-anchor fallback in both `get_messages` and `save_messages`. The host-bound `previous_response_id` (set by `ResponsesChannel` from the request envelope) is now the only authoritative anchor; the env value is still stamped into the persisted envelope's `agent_session_id` field for operator correlation only. Replaced `TestFoundryAgentSessionIdAnchor` to assert the env var is ignored as an anchor (no backend call) and still appears in the saved envelope. Fixed in 93dc14549.

python/packages/foundry_hosting/agent_framework_foundry_hosting/_history_provider.py
eavanvalkenburg: Considered this — the blocker is plug-in scope. `FoundryHostedAgentHistoryProvider` implements `agent_framework.HistoryProvider`, which `BaseAgent` invokes via `chat_options.history_provider.get_messages(...)` *before* it hands control to the chat client. At that point we're outside the SDK's request handler scope, so there's no live `ResponseContext` (and no contextvar exposed for one) to call `get_history()` on.

What we do call internally is the same pair `ResponseContext.get_history` calls under the hood — `provider.get_history_item_ids(...)` followed by `provider.get_items(...)` against the same `FoundryStorageProvider`/`InMemoryResponseProvider` — with the same `isolation=...` argument the SDK uses, so behaviour stays consistent. The only material difference is anchor resolution, and that one we have to source ourselves from the host's bound request context (the SDK pulls it off the request envelope, which we don't see at the BaseAgent plug-in boundary). Resolving — happy to revisit if the SDK exposes a `current_response_context()` contextvar in the future.

Issue #23, opened 2025-05-28
Create base Agent class


markwallace-microsoft:
Once the Agent and Agent Thread designs are completed add these abstractions to the repository along with XML documentation and unit tests.

markwallace-microsoft:
Depends on #17 and #20

PR #5638, opened 2026-05-05
Python: add agent-framework-hosting core package

eavanvalkenburg:
### Motivation and Context

Implements **§4 (Core abstractions) and §5 (`AgentFrameworkHost`)** of [SPEC-002](https://github.com/microsoft/agent-framework/blob/feature/python-hosting/docs/decisions/0026-hosting-channels.md) (merged via #5549). This is the foundation package every channel (Responses, Invocations, Activity Protocol, Teams, Telegram, Entra link) imports from.

### Description

Adds the new `agent-framework-hosting` package (`python/packages/hosting/`) with:

* `Channel` protocol — the contract every channel package implements (`name`, `contribute()`, optional `on_init`/`on_close` lifecycle hooks).
* `AgentFrameworkHost` — Starlette ASGI app that wires together a target (`Agent`, `Workflow`, or callable) with one or more channels, handles isolation-key threading, identity normalization, and run hooks per-channel.
* `ChannelRequest` / `ChannelResponse` envelopes — the uniform structure each channel transforms its protocol into before invoking the target.
* `ChannelSession` + `isolation_key` plumbing for per-conversation history bucketing.
* Optional workflow-target adapter so a `Workflow` can be the host's target alongside `Agent`.
* `tests/` with `test_host.py`, `test_types.py`, and shared fixtures.

Workspace registration + lockfile entry so the package builds in CI.

### Stack

PR-2 of 9 — the foundation. All channel PRs (PR-3..PR-7) and the samples PR (PR-8) depend on this landing first.

| # | Branch | Status |
|---|---|---|
| PR-1 | `refactor/foundry-hosted-agent-history-provider` | independent |
| **PR-2** | `feat/hosting-core` | **this PR** — needed by PR-3..PR-8 |
| PR-3..7 | per-channel packages | block on this |
| PR-8 | `feat/hosting-samples` | blocks on this + PR-3..7 |

### Contribution Checklist


python/packages/hosting/agent_framework_hosting/_host.py
```diff
@@ -0,0 +1,946 @@
...
+    """Yield, swallowing the ``RuntimeError`` ``ResponseStream`` raises on double-consume.
+
+    The bridge stream calls ``get_final_response()`` after iterating the
+    workflow stream so the workflow's cleanup hooks run; on some paths the
+    stream considers itself already finalized and raises, which we treat
+    as benign — we're only after the side effect.
+    """
+    try:
+        yield
+    except RuntimeError as exc:
```
moonbox3, line 278: Should we narrow this? `RuntimeError` covers the double-consume case but also "Event loop is closed", "cannot reuse already awaited coroutine", and anything an executor surfaces during finalize, all logged at `debug` and swallowed. Same for `except Exception` below: checkpoint write failures, context-provider errors vanish, callers see a clean run. Could `ResponseStream` raise a dedicated sentinel class for the benign case so the rest can propagate?

python/packages/hosting/agent_framework_hosting/_host.py
```diff
@@ -0,0 +1,946 @@
...
+        if self._closed:
+            return
+        self._closed = True
+        self._stack.close()
+
+    def __await__(self) -> Any:
+        # ``__await__`` returns a generator; closing here would be too
+        # eager — we close in ``__aiter__`` finally instead. Awaitable
+        # consumers (rare for streams) call ``aclose()`` separately.
+        return self._inner.__await__()
```
moonbox3: Await path leaks the ExitStack. Comment says "consumers call `aclose()` separately" but no `aclose()` defined and no caller invokes one. An awaiting consumer drops the stream and the bound Foundry response-id context stays entered for the host's lifetime, accumulating per request. Should we either define `aclose` or wrap the await path so `_close()` runs?

python/packages/hosting/agent_framework_hosting/_host.py
```diff
@@ -0,0 +1,946 @@
...
+            elif name == ISOLATION_HEADER_CHAT:
+                chat_key = raw_value.decode("latin-1") or None
+        if user_key is None and chat_key is None:
+            await self.app(scope, receive, send)
+            return
+        token = set_current_isolation_keys(IsolationKeys(user_key=user_key, chat_key=chat_key))
+        try:
+            await self.app(scope, receive, send)
+        finally:
+            reset_current_isolation_keys(token)
```
moonbox3, line 479: Are we covering this end-to-end? Couldn't find a TestClient request that sets the isolation headers and asserts `get_current_isolation_keys()` resolves inside a route and resets after. This is the only seam Foundry providers have to find partition keys, so a regression silently misroutes writes with zero test signal. Worth pinning down the one-header-only and non-http-scope branches too.

python/packages/hosting/agent_framework_hosting/_host.py
```diff
@@ -0,0 +1,946 @@
...
+            # that don't want a prefix (e.g. ResponsesChannel exposing POST /responses directly).
+            if contribution.routes:
+                if channel.path:
+                    routes.append(Mount(channel.path, routes=list(contribution.routes)))
+                else:
+                    routes.extend(contribution.routes)
+            on_startup.extend(contribution.on_startup)
+            on_shutdown.extend(contribution.on_shutdown)
+
+        @asynccontextmanager
```
moonbox3, line 1305: One bad `on_shutdown` aborts the rest. First raise bubbles out of `finally`, remaining channels never get a chance to close their tasks/sockets/sessions. Same story on startup: third callback raising leaves the first two running with no shutdown pairing. Should we collect errors per callback and raise an aggregate at the end?

python/packages/hosting/agent_framework_hosting/_host.py
```diff
@@ -0,0 +1,946 @@
...
+        # The full spec resolves a ChannelSession into an AgentSession here,
+        # honors session_mode, and consults LinkPolicy / ResponseTarget. This
+        # base host keys a per-isolation_key AgentSession off the channel's
+        # session hint so context providers (FileHistoryProvider, …) on the
+        # target see one session per end user.
+        session = None
+        if request.session_mode != "disabled" and request.session is not None:
+            isolation_key = request.session.isolation_key
+            if isolation_key is not None and hasattr(self.target, "create_session"):
+                session_id = self._session_aliases.get(isolation_key, isolation_key)
```
moonbox3, line 1425: Race across concurrent requests for the same `isolation_key`. If `create_session` ever yields, two tasks both pass the `is None` check, both create sessions, second overwrites first, the orphan never closes. Same shape on `_session_aliases`, `_identities`, `_active`. Worth an `asyncio.Lock` keyed per `isolation_key`?

python/packages/hosting/agent_framework_hosting/_host.py
```diff
@@ -0,0 +1,946 @@
...
+        flat: list[Any] = []
+        for entry in providers:
+            children = getattr(entry, "providers", None)
+            if children:
+                flat.extend(children)
+            else:
+                flat.append(entry)
+        return flat
+
+    def _bind_request_context(self, request: ChannelRequest) -> ExitStack:
```
moonbox3, line 1488: `bind_request_context` duck-typed hook has no test. This is the seam Foundry response-id chaining plugs into. Should we have a fake provider that records enter/exit and assert (a) it's called with the right `response_id`/`previous_response_id`, (b) stays open across the agent run, (c) stays open until stream exhaustion on the streaming path?

python/packages/hosting/agent_framework_hosting/_host.py
```diff
@@ -0,0 +1,946 @@
...
+        with self._bind_request_context(request):
+            # ``_is_workflow`` is False here so ``self.target`` is an
+            # ``Agent``-shaped target whose ``.run`` returns
+            # :class:`AgentResponse`. Narrow back to keep ``result.text``
+            # well-typed without conditional imports of ``Agent``.
+            agent_target = cast("SupportsAgentRun", self.target)
+            result = await agent_target.run(self._wrap_input(request), **run_kwargs)
+        return HostedRunResult(text=result.text)
+
+    def _invoke_stream(self, request: ChannelRequest) -> ResponseStream[AgentResponseUpdate, AgentResponse]:
```
moonbox3, line 1559: Agent-target streaming path has no coverage. Existing streaming tests only hit `Workflow`. `_BoundResponseStream`'s `__aiter__` finally-close, `get_final_response` close, double-close idempotence, `__getattr__` forwarding all unverified. Fake agent yielding a couple updates plus a `get_final_response`, paired with a fake provider tracking enter/exit ordering, would lock this in.

python/packages/hosting/agent_framework_hosting/_host.py
```diff
@@ -0,0 +1,946 @@
...
+        hosting_meta["response_target"] = {
+            "kind": target.kind.value,
+            "targets": list(target.targets),
+        }
+
+        raw = request.input
+        if isinstance(raw, Message):
+            raw.additional_properties = {**(raw.additional_properties or {}), "hosting": hosting_meta}
+            return raw
+        if isinstance(raw, list) and raw and all(isinstance(m, Message) for m in raw):
```
moonbox3, line 1811: Looks like the `list[Message]` branch is unverified. Existing test only sends a string. Contract is load-bearing: `hosting` block lands on the LAST message; a regression stamping `messages[0]` instead silently breaks history correlation for every multi-message payload. Worth a test that passes `[system, user]` and asserts the system stays clean, user gets the metadata?

python/packages/hosting/agent_framework_hosting/_host.py
```diff
@@ -0,0 +1,946 @@
...
+                    "deliver_response: no known identity for isolation_key=%s on channel=%s",
+                    isolation_key,
+                    channel_name,
+                )
+                skipped.append(token)
+                continue
+            try:
+                await channel.push(dest_identity, payload)
+            except Exception:
+                logger.exception("deliver_response: push failed for target=%s", token)
```
moonbox3: When every push raises (Teams/Telegram outage, expired creds, rate limits), `pushed` stays empty, fallback flips `include_originating=True`, and the originating channel renders the reply as if delivery succeeded. Caller has to inspect `DeliveryReport.skipped` to notice the fan-out failed. Should we distinguish "no identity" (the documented drop) from "push raised" (an outage), and signal the latter on `DeliveryReport`?

python/packages/hosting/agent_framework_hosting/_isolation.py
```diff
@@ -0,0 +1,76 @@
...
+    "reset_current_isolation_keys",
+    "set_current_isolation_keys",
+]
+
+
+ISOLATION_HEADER_USER = "x-agent-user-isolation-key"
+ISOLATION_HEADER_CHAT = "x-agent-chat-isolation-key"
+
+
+class IsolationKeys:
```
moonbox3, line 46: We're lacking tests here -- no `test_isolation.py`, no test imports this surface?

python/packages/hosting/agent_framework_hosting/_types.py
```diff
@@ -0,0 +1,376 @@
...
+#
+# ``protocol_request`` is the raw, channel-native payload the channel
+# parsed (the JSON body for Responses, the Telegram ``Update`` dict, the
+# Bot Framework ``Activity`` for Teams). Use it when the hook needs a
+# field the channel did not lift onto ``ChannelRequest`` (e.g. OpenAI's
+# ``safety_identifier``, Teams' ``from.aadObjectId``, …).
+ChannelRunHook = Callable[..., "Awaitable[ChannelRequest] | ChannelRequest"]
+
+
+async def apply_run_hook(
```
moonbox3, line 598: I don't see any tests for this either.

python/packages/hosting/tests/test_types.py
```diff
@@ -0,0 +1,108 @@
...
+
+class TestChannelIdentity:
+    def test_frozen(self) -> None:
+        ident = ChannelIdentity(channel="teams", native_id="abc")
+        with pytest.raises(dataclasses.FrozenInstanceError):
+            ident.native_id = "other"  # type: ignore[misc]
+
+    def test_attributes_default_empty_mapping(self) -> None:
+        ident = ChannelIdentity(channel="teams", native_id="abc")
+        assert dict(ident.attributes) == {}
```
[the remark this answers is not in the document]
eavanvalkenburg: Added `TestApplyRunHook` in `tests/test_types.py` covering the three branches: protocol_request `None` short-circuit, sync hook invocation (captures kwargs, no await), and async hook invocation (awaited via `inspect.iscoroutine`). Fixed in 9786cbb4c.

python/packages/hosting/agent_framework_hosting/_host.py
eavanvalkenburg: Promoted the `RuntimeError` log to `WARNING` with `exc_info` (kept non-`RuntimeError` at `exception()` so generic teardown bugs surface with full traceback). Updated the docstring to make the contract explicit: any non-clean teardown is observable in operator logs. Fixed in 9786cbb4c.

python/packages/hosting/agent_framework_hosting/_host.py
eavanvalkenburg: Added explicit `aclose()` and rerouted `__await__` through `get_final_response()` so both paths run the same `finally` cleanup. `_close()` is idempotent via the `_closed` flag. New `TestBoundResponseStream` covers: aclose-without-iteration, double-close-no-op, `__getattr__` forwarding (e.g. `stream.with_result_hook`), and the await-routes-through-final path. Fixed in 9786cbb4c.

python/packages/hosting/agent_framework_hosting/_host.py
eavanvalkenburg: Added `TestIsolationMiddlewareEndToEnd` in the new `tests/test_isolation.py`, driving the ASGI app via Starlette's `TestClient` and asserting: both headers populate the contextvar, single-header partial population, missing headers leave defaults, empty-string header values fall back to defaults, and the contextvar resets to its sentinel after each request. Also covers the concurrent-task isolation invariant. Fixed in 9786cbb4c.

python/packages/hosting/agent_framework_hosting/_host.py
eavanvalkenburg: Lifespan now collects callback errors instead of bailing on the first one — every startup (and every shutdown) callback runs, each failure is logged with its qualname via `logger.exception`, and the first error is re-raised so Starlette still aborts boot. When more than one fails, an aggregate `logger.error` summarises the trailing failures so an operator gets the whole picture in a single log scrape. Covered by `TestLifespanAggregation` (startup + shutdown), which asserts every callback ran and inspects `caplog.records[*].exc_text` for both failure payloads. Fixed in 9786cbb4c.

python/packages/hosting/agent_framework_hosting/_host.py
eavanvalkenburg: Switched the cache write to `self._sessions.setdefault(isolation_key, new_session)` — atomic in CPython, so two concurrent racers either both observe the same winning session (the loser's freshly-created session is dropped) or pick up the existing one. Couldn't easily add an `asyncio.Lock` here because `_build_run_kwargs` is reached from the sync `_invoke_stream` path; documented the rationale inline. Fixed in 9786cbb4c.

python/packages/hosting/agent_framework_hosting/_host.py
eavanvalkenburg: Added `TestBindRequestContext` covering the bind hook contract: context manager opened before the agent runs, payload includes `channel` / `operation` / `attributes` / `identity` / `session`, context closes after the response, and the binding is held open for the entire stream lifetime (close fires only after exhaustion). Uses a `_RecordingContextProvider` that duck-types `bind_request_context(**kwargs)` via `@contextmanager` and a `_ProvidersAgent` that exposes `context_providers` so the host's `_flat_context_providers` walk finds it. Fixed in 9786cbb4c.

python/packages/hosting/agent_framework_hosting/_host.py
eavanvalkenburg: Streaming-path coverage now lives in `TestBoundResponseStream`. `_ProvidersAgent.run(stream=True)` returns a real `ResponseStream[AgentResponseUpdate, AgentResponse]` (sync return, matching `Agent.run`'s overload), exercising the actual `_BoundResponseStream` wrapping rather than mocking. Tests assert the bind events appear in the right order relative to `agent_start` / `stream_yield` / `stream_done` / agent end. Fixed in 9786cbb4c.

python/packages/hosting/agent_framework_hosting/_host.py
eavanvalkenburg: Added `TestWrapInputListMessages` exercising the `_wrap_input` LAST-message metadata extraction: a list of messages where the LAST entry's `additional_properties` populates the channel-side hints, and earlier entries' metadata is ignored (important because the host treats the last item as the active user message). Also covers the empty-list and single-item cases. Fixed in 9786cbb4c.

python/packages/hosting/agent_framework_hosting/_host.py
eavanvalkenburg: Reworked `_deliver_response`: push exceptions now land in a new `DeliveryReport.failed` field as `(target_token, error_summary)` tuples (`f"{type(exc).__name__}: {exc}"`), distinct from the explicit "no link" drops in `skipped`. When *only* `failed` is non-empty (every push raised), no originating fallback is synthesised — the channel inspects `failed` and decides degraded behaviour itself. New `TestDeliveryReport` asserts the new field and the updated test `test_push_exception_lands_in_failed_no_fallback` verifies the no-fallback contract. Fixed in 9786cbb4c.

python/packages/hosting/agent_framework_hosting/_isolation.py
eavanvalkenburg: Added a new `tests/test_isolation.py` covering the entire `_isolation.py` surface to 100% coverage: `IsolationKeys` construction / equality / immutability, the `get` / `set` / `reset` context-var helpers, header constants, the ASGI middleware lifting headers (both / single / neither / empty-value), reset-after-request, concurrent-task isolation, and lifespan / non-HTTP scope passthrough. Fixed in 9786cbb4c.

python/packages/hosting/agent_framework_hosting/_types.py
eavanvalkenburg: Added `TestApplyRunHook` in `tests/test_types.py` covering the three execution paths: short-circuit when `protocol_request` is `None`, synchronous hook (call captured, no await), and async hook (awaited via `inspect.iscoroutine`). The unspecified-types helper is still imported via `from agent_framework_hosting import apply_run_hook`. Fixed in 9786cbb4c.

python/packages/hosting/agent_framework_hosting/_host.py
```diff
@@ -0,0 +1,1062 @@
...
+            if latest is not None:
+                # Restore in-memory state from the most recent checkpoint
+                # before applying the new input.
+                await workflow.run(checkpoint_id=latest.checkpoint_id, checkpoint_storage=storage)
+            result = await workflow.run(request.input, checkpoint_storage=storage)
+        else:
+            result = await workflow.run(request.input)
+        outputs = result.get_outputs()
+        text = "\n".join(_workflow_output_to_text(o) for o in outputs) if outputs else ""
+        return HostedRunResult(text=text)
```
TaoChenOSU: Will it support other modalities in the future?

python/packages/hosting/agent_framework_hosting/_host.py
```diff
@@ -0,0 +1,1062 @@
...
+        (telemetry, debug UIs) can recover the full workflow timeline.
+
+        Checkpoint restoration (when ``checkpoint_location`` is set) runs
+        before the input stream is opened so the new turn observes the
+        restored state.
+        """
+        workflow: Workflow = self.target  # type: ignore[assignment]
+        storage = self._resolve_checkpoint_storage(request)
+
+        async def _maybe_restore() -> None:
```
TaoChenOSU: The non streaming case doesn't seem to have this restore routine.

python/packages/hosting/agent_framework_hosting/_host.py
```diff
@@ -0,0 +1,1062 @@
...
+        # readiness is intentionally cheap so the platform's probe never times
+        # out on transient channel work). Mounted first so a channel cannot
+        # accidentally shadow it.
+        async def _readiness(_request: Request) -> PlainTextResponse:  # noqa: RUF029
+            """Liveness/readiness probe handler used by Foundry Hosted Agents."""
+            return PlainTextResponse("ok")
+
+        routes.append(Route("/readiness", _readiness, methods=["GET"]))
+
+        for channel in self.channels:
```
TaoChenOSU, line 1292: What happens if the agent receives requests from multiple channels at the same time?

eavanvalkenburg: By design, multiple channels can serve the same end user concurrently — that's why the session cache is keyed on the channel-supplied `isolation_key`, not on `(channel, isolation_key)`. Two channels resolving to the same `isolation_key` (e.g. the same Telegram chat hitting Telegram and Responses simultaneously) share an `AgentSession` so the user's history is consistent across surfaces — otherwise the second channel would silently lose the conversation context the first established.

Concurrency safety: each request runs on its own task with its own `bind_request_context()` (per-request `previous_response_id` / `response_id`) lifted via the per-task `IsolationKeys` contextvar, so there is no cross-request contamination of the host-side state. The session-cache write itself is now done via `dict.setdefault` (PR-2 round 1) so concurrent first-request racers can't orphan a session. Anything past that point is the agent's / `HistoryProvider`'s concurrency contract — for the Foundry storage provider for example, concurrent appends just produce concurrent response IDs and each turn sees a consistent past via its bound previous_response_id.

python/packages/hosting/agent_framework_hosting/_host.py
eavanvalkenburg: The blocking path is text-only on purpose — it joins the workflow's text outputs and returns a `HostedRunResult(text=...)` for channels that don't (or can't) stream. Richer modalities (image, audio, structured tool/function results, citations) ride the **streaming** path: `_invoke_workflow_stream` bridges every workflow update through `AgentResponseUpdate`, which preserves all content parts; channels iterate the stream and forward each content type to their native surface (e.g. Teams attachments, Telegram media, AG-UI events).

Added an inline note clarifying this and pointing channels at the streaming path when full-fidelity output is needed for blocking-style invocations. Fixed in 64b7c608c.

python/packages/hosting/agent_framework_hosting/_host.py
eavanvalkenburg: Good catch on the asymmetry — the non-streaming path *does* restore (the `latest = await storage.get_latest(...)` block right above the input run) but it was inlined while the streaming side had a named `_maybe_restore` helper, which made the parallelism easy to miss. Refactored both into static helpers (`_restore_workflow_checkpoint` for blocking, `_restore_workflow_checkpoint_streaming` for the streaming side that drains the rehydration stream) so the restore contract is named and shared. Fixed in 64b7c608c.

python/packages/hosting/agent_framework_hosting/_host.py
```diff
@@ -0,0 +1,1204 @@
...
+                        "remove one of the two configurations."
+                    )
+                if isinstance(checkpoint_location, (str, os.PathLike)):
+                    self._checkpoint_location = Path(os.fspath(checkpoint_location))
+                else:
+                    # Anything else is treated as a CheckpointStorage instance.
+                    # ``CheckpointStorage`` is a non-runtime-checkable Protocol,
+                    # so we cannot ``isinstance``-check it directly.
+                    self._checkpoint_location = checkpoint_location
+        # Per-isolation_key session cache. The real spec backs this with a
```
eavanvalkenburg, line 764: does this mean that there is a followup that adds session storage?

python/packages/hosting/agent_framework_hosting/_host.py
```diff
@@ -0,0 +1,1204 @@
...
+            workers: Number of worker processes. Defaults to ``1``;
+                Hypercorn's process model only kicks in for ``>1``.
+            **config_kwargs: Forwarded to :class:`hypercorn.config.Config`
+                via attribute assignment, so any documented Hypercorn
+                config field (e.g. ``keep_alive_timeout=...``,
+                ``access_log_format=...``) can be set directly.
+        """
+        try:
+            import asyncio
+            from typing import cast as _cast
```
eavanvalkenburg: these imports do not need to be in the function, they can be at the top

python/packages/hosting/agent_framework_hosting/_host.py
```diff
@@ -0,0 +1,1204 @@
...
+
+        When ``debug=True`` is set on the host, also dump the channel-native
+        settings the channel attached to the ``ChannelRequest`` — ``options``
+        (the ChatOptions-shaped fields the channel parsed from its protocol
+        payload, e.g. temperature/tools/tool_choice for Responses), plus
+        ``attributes`` / ``metadata`` (the channel's protocol-specific bag,
+        e.g. ``chat_id`` / ``callback_query_id`` for Telegram).
+        """
+        isolation_key = request.session.isolation_key if request.session is not None else None
+        logger.info(
```
eavanvalkenburg, line 1468: we should use structured logging for this, makes it much easier to debug

python/packages/hosting/agent_framework_hosting/_host.py
```diff
@@ -0,0 +1,1204 @@
...
+            request.session_mode,
+        )
+        logger.debug(
+            "  ↳ options=%s attributes=%s metadata=%s",
+            dict(request.options) if request.options else {},
+            dict(request.attributes) if request.attributes else {},
+            dict(request.metadata) if request.metadata else {},
+        )
+
+    def _flat_context_providers(self) -> list[Any]:
```
eavanvalkenburg: do we need this? ContextProviderBase doesn't exist and there is no built-in provider with subproviders right now

python/packages/hosting/agent_framework_hosting/_host.py
```diff
@@ -0,0 +1,1204 @@
...
+            return await self._invoke_workflow(request)
+        run_kwargs = self._build_run_kwargs(request)
+        with self._bind_request_context(request):
+            # ``_is_workflow`` is False here so ``self.target`` is an
+            # ``Agent``-shaped target whose ``.run`` returns
+            # :class:`AgentResponse`. Narrow back to keep ``result.text``
+            # well-typed without conditional imports of ``Agent``.
+            agent_target = cast("SupportsAgentRun", self.target)
+            result = await agent_target.run(self._wrap_input(request), **run_kwargs)
+        return HostedRunResult(text=result.text)
```
eavanvalkenburg: this leaves out any multi-model response content...

python/packages/hosting/agent_framework_hosting/_host.py
```diff
@@ -0,0 +1,1204 @@
...
+        )
+        # NOTE: blocking workflow output is text-only today. Richer
+        # modalities (image, audio, structured tool results) ride the
+        # streaming path via ``AgentResponseUpdate`` and are forwarded by
+        # the channel as content parts; channels that need full-fidelity
+        # output for blocking calls should drive the workflow through
+        # ``stream=True`` and aggregate via ``get_final_response()``.
+        outputs = result.get_outputs()
+        text = "\n".join(_workflow_output_to_text(o) for o in outputs) if outputs else ""
+        return HostedRunResult(text=text)
```
eavanvalkenburg: this should indeed already support multi-model, the channels should decide what is sent back (based on capabilities), not the host

python/packages/hosting/agent_framework_hosting/_host.py
```diff
@@ -0,0 +1,1204 @@
...
+    if event.type != "output":
+        return None
+    payload: Any = event.data
+    if isinstance(payload, AgentResponseUpdate):
+        # Already a streaming update — pass through but tag the source so
+        # downstream hooks can tell it came from a workflow executor.
+        if payload.raw_representation is None:
+            payload.raw_representation = event
+        return payload
+    text = _workflow_output_to_text(payload)
```
eavanvalkenburg, line 252: if payload is already a object of type `Content` then that should be set on the `contents` field of the AgentResponseUpdate

python/packages/hosting/agent_framework_hosting/_types.py
```diff
@@ -0,0 +1,393 @@
...
+
+    def __init__(
+        self,
+        channel: str,
+        native_id: str,
+        attributes: Mapping[str, Any] | None = None,
+    ) -> None:
+        self.channel = channel
+        self.native_id = native_id
+        self.attributes: Mapping[str, Any] = attributes if attributes is not None else _EMPTY_MAPPING
```
eavanvalkenburg: is this not a singleton? can we just do:

```suggestion
        self.attributes: Mapping[str, Any] = attributes if attributes is not None else dict()
```

python/packages/hosting/agent_framework_hosting/_host.py
eavanvalkenburg: Yes — session storage and host-level state persistence are tracked as a v1-fast-follow / requirement #23 in the spec, and the `HostStateStore` seam is already in place (`FileHostStateStore` + `InMemoryHostStateStore`) so pluggable Cosmos / SQL / Redis adapters can land without further Protocol changes. Will explicitly wire pluggable session storage onto the same seam in a follow-up PR.

python/packages/hosting/agent_framework_hosting/_host.py
eavanvalkenburg: Done in e323a2d27d528d7ac120116e2edf4186f9446400 — dropped the local `asyncio` and `cast as _cast` re-imports inside `serve()` and rely on the module-level imports.

python/packages/hosting/agent_framework_hosting/_host.py
eavanvalkenburg: Done in e323a2d27d528d7ac120116e2edf4186f9446400 — `_log_incoming` now emits structured `extra={...}` payloads for both INFO (channel/operation/stream/identity/isolation_key/session_mode/response_target) and DEBUG (full request envelope) so log aggregators get queryable fields.

python/packages/hosting/agent_framework_hosting/_host.py
eavanvalkenburg: Done in e323a2d27d528d7ac120116e2edf4186f9446400 — dropped the `_flat_context_providers` helper entirely; the host now treats whatever `agent.context_providers` exposes as the final, flat list. Aggregator providers (e.g. `AggregateContextProvider`) are responsible for forwarding `response_context` to their children themselves. Replaced the obsolete `test_bind_descends_one_level_into_providers_attribute` with a regression guard asserting the host does NOT descend into `.providers`.

python/packages/hosting/agent_framework_hosting/_host.py
eavanvalkenburg: Good catch — fixed in e323a2d27d528d7ac120116e2edf4186f9446400. `_invoke` now forwards `AgentResponse.messages` (and `raw_response`) on the `HostedRunResult` so multi-modal content (images, function calls, tool results, …) survives end-to-end. The host stays modality-agnostic; channels (and the new `response_hook` seam) decide what subset their wire renders. `HostedRunResult` was reshaped around `messages: list[Message]` with `.text` / `.contents` as projections; `HostedRunResult(text=…)` is kept as a back-compat shim that wraps a single assistant text message.

python/packages/hosting/agent_framework_hosting/_host.py
eavanvalkenburg: Agreed — fixed in e323a2d27d528d7ac120116e2edf4186f9446400. `_invoke_workflow` builds a per-event message list via a new `_workflow_output_to_messages` helper that preserves `AgentResponse` / `AgentResponseUpdate` / `Message` / `Content` branches and only falls back to a text content for arbitrary objects. Workflow output now reaches channels with full modality intact. Also wired the new `ChannelResponseHook` so a host developer can shape workflow output per destination (e.g. project a `value=` payload into channel-friendly text or a card).

python/packages/hosting/agent_framework_hosting/_host.py
eavanvalkenburg: Done in e323a2d27d528d7ac120116e2edf4186f9446400 — `_workflow_event_to_update` now branches on `isinstance(payload, Content)` and wraps the content in an `AgentResponseUpdate(contents=[payload], …)` unchanged, so multi-modal workflow stream payloads survive into channels without being stringified.

python/packages/hosting/agent_framework_hosting/_types.py
eavanvalkenburg: Done in e323a2d27d528d7ac120116e2edf4186f9446400 — dropped the `_EMPTY_MAPPING` sentinel; `ChannelIdentity.attributes` now uses plain `dict()` as the default. Each instance gets its own empty dict; one less symbol to track.

eavanvalkenburg:
Reshaped `HostedRunResult` once more — `result: TResult` is now the canonical accessor and the previous `messages` projection is gone (`6353ddc5c`).

### Shape

```python
class HostedRunResult(Generic[TResult]):
    result: TResult
    session: AgentSession | None
```

- Agent targets → `HostedRunResult[AgentResponse]`: channels read `result.messages`, `result.text`, `result.value`, `result.response_id`, `result.usage_details` directly off the underlying response.
- Workflow targets → `HostedRunResult[WorkflowRunResult]`: channels iterate `result.get_outputs()` and inspect `result.get_final_state()` themselves.
- Echo phase synthesises a `HostedRunResult[AgentResponse]` wrapping the user's turn so one delivery path serves both phases.

### Why drop `messages` (and the earlier `raw_response`)

The earlier shape pre-shaped target output: workflows were collapsed onto a synthesised `Message` list, losing per-executor outputs, final state, and structured-value affordances. With the generic envelope:
- the host stays modality-agnostic — no `_workflow_output_to_messages` helper, no `len(shaped.text)` logging path,
- channel authors get static typing where they want it (`HostedRunResult[AgentResponse]` in a Responses hook signature; `HostedRunResult[WorkflowRunResult]` in a workflow-oriented one),
- there is one canonical accessor (`result`) rather than two (`messages` + `raw_response`), and hooks just inspect what's there.

`replace()` is now `{result, session}` only; the clone is shallow, so channels that need to mutate `result` itself (rather than rebind via `replace(result=...)`) own that deep copy.

### Spec / ADR

- Q3 amended to describe the generic envelope.
- Q28 amended: "no collapsing" now expressed in terms of carrying the full `AgentResponse` / `WorkflowRunResult`.
- Q29 amended: hook signature is `HostedRunResult[Any]` (channels narrow at hook entry if they want).
- New Q31 captures the rationale for the generic-envelope decision.
- Canonical flow + `result is rich, not just text` section in the spec rewritten around `result.result.*`.

### Validation

- 112 tests pass
- mypy clean
- pyright clean
- ruff fmt + lint clean

eavanvalkenburg:
## Round 5 — rubber-duck adoption

Pushed as `b82b1a204`. Adopts the round-5 rubber-duck findings (the duck saw all of `_host.py` + `_runner.py` + the spec) with **one deviation**: `DeliveryReport.failed` was not preserved — the value type is removed entirely. Failure observability moves to logs (`InProcessTaskRunner`) and the runner backend (durable adapters); the host's internal `_deliver_response` now returns `bool` for the originating channel's bookkeeping. The deviation simplifies the public surface and avoids carrying a half-typed "schedule-time failure" subtype now that durable delivery owns the failure narrative end-to-end.

### Code
- **Strict ephemeral default** — `runtime_mode="ephemeral"` + default in-process runner raises `RuntimeError`; opt-in via `allow_in_process_runner=True` (warns). Silently using the in-process runner on FHA-style scale-to-zero would drop in-flight pushes — caught at construction.
- **Codec seam** — new `DurableTaskPayloadMode(OBJECT|JSON)` declared by every runner; new `ChannelPushCodec` Protocol for push channels with non-JSON-native payloads; `_validate_runner_codec_pairing` refuses a JSON-mode runner paired with codec-less push channels at construction. `_handle_push_task` accepts both `OBJECT` (in-memory object) and `JSON` (`{"type":"push",...}` envelope) shapes.
- **`ResponseTarget.identity` / `.identities`** — explicit caller-supplied recipients (bypasses the link store). Renamed the field `identities → _target_identities` to free the classmethod name.
- **Intent-only audit on the message** — `_annotate_intended_targets` writes `hosting.intended_targets / skipped_targets / includes_originating / originating_channel` to assistant messages; single immutable write per the runner-owned operational-state model.
- **2-phase drain in `InProcessTaskRunner`** — `shutdown_grace_seconds` (default 5.0) lets in-flight retries finish before cancellation; `CancelledError` is swallowed as the expected shutdown shape.
- **Echo idempotency** — `_handle_push_task` tracks an `echo_done` cursor on runner-owned task state so a retry after a successful echo phase but before the response phase completed does not double-echo.

### Wave-1 authorization seam (full landing)
New `_authorization.py` with the full Wave-1 surface (`AllowlistDecision`, `AuthorizationContext`, `IdentityAllowlist` Protocol, `AllowAll` / `NativeIdAllowlist` / `LinkedClaimAllowlist` (raises until Wave 2) / `AnyOfAllowlists` / `AllOfAllowlists` / `CallableAllowlist`, `Allowed` / `LinkRequired` / `Denied`, `ChannelConfigurationError`). Wired into the host:
- `Host(default_allowlist=..., identity_linker=...)` + per-channel `allowlist: ... | Literal["inherit"] | None` parameter.
- `_validate_channel_authorization` enforces three rules at construction: (#1) claim-source requirement, (#2) **linker presence for `require_link=True`** (elevated from no-op per duck's "forced_links must not ship unenforced"), (#3) `NativeIdAllowlist(channel=...)` typo detection. Combinator-walking via `_flatten_allowlists` catches nested misconfigurations.
- `host.authorize(identity, *, require_link, allowlist, verified_claims=None)` for the native-id pipeline: open path returns `Allowed` with an auto-issued `<channel>:<native_id>` isolation key (re-issues the existing key when the identity has been seen before); `ABSTAIN` on a claim-required allowlist maps to `Denied(reason_code="allowlist_requires_link")` until Wave 2 wires the linker.

### Spec / ADR
- **`docs/specs/002-python-hosting-channels.md`** — Wave-1 status updated to reflect the linker-presence elevation and the `host.authorize` landing; new sub-sections [Codec contract for durable serialisation](https://github.com/microsoft/agent-framework/blob/feat/hosting-core/docs/specs/002-python-hosting-channels.md#codec-contract-for-durable-serialisation), [In-process runner shutdown drain](https://github.com/microsoft/agent-framework/blob/feat/hosting-core/docs/specs/002-python-hosting-channels.md#in-process-runner-shutdown-drain), [Echo idempotency on retry](https://github.com/microsoft/agent-framework/blob/feat/hosting-core/docs/specs/002-python-hosting-channels.md#echo-idempotency-on-retry); `DeliveryReport` references purged from Qs 18 / 21; new resolved Qs 36–40 covering strict-ephemeral default, codec contract, `DeliveryReport` removal, echo cursor, and drain.
- **`docs/decisions/0026-hosting-channels.md`** — Q12 `DeliveryReport` reference purged; Q16 updated for Wave-1 landing; new resolved Qs 20 (codec contract) + 21 (strict ephemeral / drain / echo cursor).

### Tests
- **New `tests/test_authorization.py`** (35 cases) — every Wave-1 built-in, the three validator rules, combinator decision semantics, and `host.authorize` across open / allow / deny / abstain-with-claim-dep / abstain-without-claim-dep paths plus existing-key reuse and verified-claims propagation.
- `tests/test_host.py` — `TestDeliverResponse` rewritten for the `bool` return + `runner.scheduled` count assertions; new tests for the `IDENTITIES` variant + echo idempotency.
- `tests/test_runner.py` — strict-ephemeral now expects `RuntimeError`; `allow_in_process_runner` opt-in tests; shutdown drain test; `payload_mode` default test.
- `tests/test_types.py` — `TestDeliveryReport` removed; new `TestDurableTaskPayloadMode` + `TestResponseTargetIdentities`.

### Validation
`uv run poe check-packages -P hosting` ✓ · `uv run poe mypy -P hosting` ✓ · `uv run poe test -P hosting` ✓ — **178 tests pass, 91 % coverage**.

PR #5639, opened 2026-05-05
Python: add agent-framework-hosting-responses channel

eavanvalkenburg:
### Motivation and Context

Implements **§7 ("Built-in channels — Responses")** of [SPEC-002](https://github.com/microsoft/agent-framework/blob/feature/python-hosting/docs/decisions/0026-hosting-channels.md) (merged via #5549). Adds the OpenAI Responses-compatible channel so an `AgentFrameworkHost` can be reached by any Responses SDK / OpenAI client.

### Description

Adds the new `agent-framework-hosting-responses` package (`python/packages/hosting-responses/`) with:

* `ResponsesChannel` — mounts `POST /responses` (configurable), parses the OpenAI Responses request shape into `ChannelRequest`, and serializes the agent's reply back into a Responses payload (sync or streaming SSE).
* `_parsing.py` — request/response translation, isolation-key derivation from `previous_response_id` / `safety_identifier`, and option scrubbing seam.
* Tests covering parsing, channel construction, end-to-end round-trips, run-hook integration.

### Stack

PR-3 of 10. **Depends on #PR-2** (`feat/hosting-core`). Until PR-2 merges, this PR's diff includes that commit too — afterwards the diff shrinks to just this package.

### Contribution Checklist

### DIFF
--- a/python/packages/hosting-responses/agent_framework_hosting_responses/_channel.py
+++ b/python/packages/hosting-responses/agent_framework_hosting_responses/_channel.py
@@ -1,0 +1,339 @@
+# Copyright (c) Microsoft. All rights reserved.
+
+"""``ResponsesChannel`` — OpenAI Responses-shaped HTTP surface.
+
+Exposes a single ``POST /responses`` endpoint that accepts
+``{"input": "...", "stream": false}`` (and the rest of the Responses API
+request body) and returns either a Responses-shaped JSON body
+(``stream=False``, default) or a Server-Sent-Events stream
+(``stream=True``).
+
+Payload construction reuses the ``openai.types.responses`` Pydantic
+models so the OpenAI Python SDK ``stream=True`` consumer parses every
+required field without surprises.
+"""
+
+from __future__ import annotations
+
+import time
+import uuid
+from collections.abc import AsyncIterator, Callable, Mapping
+from typing import Any
+
+from agent_framework_hosting import (
+    ChannelContext,
+    ChannelContribution,
+    ChannelRequest,
+    ChannelRunHook,
+    ChannelSession,
+    DeliveryReport,
+    HostedRunResult,
+    apply_run_hook,
+    logger,
+)
+from openai.types.responses import (
+    Response as OpenAIResponse,
+)
+from openai.types.responses import (
+    ResponseCompletedEvent,
+    ResponseCreatedEvent,
+    ResponseError,
+    ResponseFailedEvent,
+    ResponseOutputMessage,
+    ResponseOutputText,
+    ResponseTextDeltaEvent,
+)
+from starlette.requests import Request
+from starlette.responses import JSONResponse, Response, StreamingResponse
+from starlette.routing import Route
+
+from ._parsing import (
+    parse_response_target,
+    parse_responses_identity,
+    parse_responses_request,
+)
+
+
+def _ack_text(report: DeliveryReport) -> str:
+    """Tiny acknowledgement string for the originating wire.
+
+    Used when the agent reply is delivered out-of-band via :class:`ChannelPush`.
+    """
+    pushed = ", ".join(report.pushed) if report.pushed else "(none)"
+    return f"[delivered out-of-band → {pushed}]"
+
+
+class ResponsesChannel:
+    """Minimal OpenAI-Responses-shaped surface.
+
+    Mounts ``POST <path>/responses`` (default path ``/responses`` so the
+    full route is ``/responses/responses`` when the channel is prefixed,
+    or just ``/responses`` when ``path=""``).
+    """
+
+    name = "responses"
+
+    def __init__(
+        self,
+        *,
+        path: str = "",
+        run_hook: ChannelRunHook | None = None,
+        response_id_factory: Callable[..., str] | None = None,
+    ) -> None:
+        """Create a Responses channel.
+
+        Args:
+            path: Mount prefix on the host. Default ``""`` mounts the
+                ``POST /responses`` route at the app root, matching the
+                upstream OpenAI surface.
+            run_hook: Optional :data:`ChannelRunHook` invoked with the
+                parsed :class:`ChannelRequest` before the agent target
+                runs. May return a replacement request.
+            response_id_factory: Optional callable that mints the
+                per-request response id. Default produces
+                ``resp_<uuid hex>`` which matches the OpenAI Responses
+                wire shape. Override when the host backing storage
+                requires a different id format (e.g. Foundry storage,
+                whose partition keys are encoded in the id and which
+                rejects free-form ``resp_*`` ids with a server error).
+                The same id is used for the channel envelope and for
+                the host-side anchoring (``ChannelRequest.attributes``)
+                so storage and replay agree.
+        """
+        self.path = path
+        self._hook = run_hook
+        self._ctx: ChannelContext | None = None
+        self._response_id_factory: Callable[..., str] = (
+            response_id_factory if response_id_factory is not None else (lambda *_a, **_kw: f"resp_{uuid.uuid4().hex}")
+        )
+
+    def contribute(self, context: ChannelContext) -> ChannelContribution:
+        """Capture the host-supplied context and register ``POST /responses``."""
+        self._ctx = context
+        return ChannelContribution(routes=[Route("/responses", self._handle, methods=["POST"])])
+
+    async def _handle(self, request: Request) -> Response:
+        """Handle a single ``POST /responses`` call.
+
+        Parses the OpenAI Responses-shaped body into ``ChatMessage`` /
+        ``options`` / ``ChannelSession`` triples via :mod:`._parsing`,
+        applies the optional ``run_hook``, and either streams an SSE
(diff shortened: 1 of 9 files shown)


python/packages/hosting-responses/agent_framework_hosting_responses/_channel.py
```diff
@@ -0,0 +1,339 @@
...
+        # get a stable per-conversation session id (used by non-Foundry
+        # history providers, routing/idempotency, etc.). The chat-iso
+        # value is *not* a valid storage anchor; the Foundry history
+        # provider deliberately ignores it — multi-turn storage chaining
+        # goes through the ``previous_response_id`` / bound
+        # ``response_id`` pair on ``ChannelRequest.attributes``. The
+        # user-iso companion header is consumed at the host level by
+        # ``_FoundryIsolationASGIMiddleware`` so the channel never has
+        # to import Foundry-specific types.
+        chat_iso = request.headers.get("x-agent-chat-isolation-key")
```
moonbox3: Should we be trusting this header straight off the wire? The comment above says it is "platform-injected" by Foundry, but the channel itself does no validation. If `ResponsesChannel` is mounted anywhere outside Foundry, or in front of the Foundry middleware in the ASGI stack, any HTTP caller can set `x-agent-chat-isolation-key: <someone-else>` and end up sharing that user's per-conversation session bucket. Right? Should we have, at minimum, a sanity check that a host-level guard is enforced before this point, and probably a docstring note that mounting this channel requires the isolation middleware in front?

python/packages/hosting-responses/agent_framework_hosting_responses/_channel.py
```diff
@@ -0,0 +1,339 @@
...
+            session = ChannelSession(isolation_key=chat_iso)
+
+        # Mint the response id once per request so the channel envelope
+        # (one-shot or streamed) and any host-side anchoring (e.g. the
+        # Foundry history provider's ``bind_request_context``) agree on
+        # the same handle. The next turn arrives with this value as
+        # ``previous_response_id`` and the storage chain walks. We pass
+        # both anchors via ``ChannelRequest.attributes`` so the host
+        # can pick them up without a channel-specific contract.
+        previous_response_id: str | None = None
```
moonbox3, line 206: The docstring for `response_id_factory` (lines 92-101) says id backends like Foundry storage embed partition keys in the id. We pass the user-supplied `previous_response_id` straight to the factory so the new record co-locates with the existing partition. Have we thought about what happens when a caller supplies an arbitrary `resp_*` id pointing at someone else's partition? Without a host-side check that the chain is owned by this caller, the factory ends up writing the new record into a partition the caller does not own. Is partition ownership enforced inside the factory itself, or somewhere upstream we are relying on?

python/packages/hosting-responses/agent_framework_hosting_responses/_channel.py
```diff
@@ -0,0 +1,339 @@
...
+                ResponseFailedEvent(
+                    type="response.failed",
+                    response=failed,
+                    sequence_number=next_seq(),
+                )
+            )
+            return
+
+        completed_text = accumulated
+        report = await self._ctx.deliver_response(request, HostedRunResult(text=accumulated))
```
moonbox3: What happens to `deliver_response` and any push-channel fan-out when the stream raises mid-flight? The non-streaming path always calls `deliver_response`, but here on failure we emit `response.failed` and return without ever invoking it. The client has already seen partial deltas on the wire, but host-side history, idempotency, and any non-originating push targets see nothing for this turn. Next turn the chain anchored on this `response_id` will be inconsistent with what the user actually saw. Should we be calling `deliver_response` with `accumulated` (or with an explicit failed marker) before returning, so host state matches wire state?

python/packages/hosting-responses/agent_framework_hosting_responses/_parsing.py
```diff
@@ -0,0 +1,238 @@
...
+    "tool_choice",
+    "tools",
+    "store",
+    "response_format",
+    "stop",
+    "seed",
+    "frequency_penalty",
+    "presence_penalty",
+    "logit_bias",
+    "instructions",
```
moonbox3: `instructions` is in `_RESPONSES_OPTION_PASSTHROUGH` but `parse_responses_request` short-circuits on it at line 218 (`if key == "instructions": continue`) because it is already prepended as a system message. Doesn't that mean the entry here is dead? Doesn't look like a bug today, but the next person who adds another consumer of `_RESPONSES_OPTION_PASSTHROUGH` will quietly ship `instructions` twice (system message and option)? Can we drop it from the set?

python/packages/hosting-responses/agent_framework_hosting_responses/_channel.py
eavanvalkenburg: Good catch — fair concern. The channel now consumes the host's `IsolationKeys.chat_key` (lifted into `current_isolation_keys` by the ASGI isolation middleware) instead of reading the raw `x-agent-chat-isolation-key` header off the wire. That moves the trust boundary into the middleware, which is the operator's place to enforce auth before stamping the isolation keys. The default Foundry deployment uses `_FoundryIsolationASGIMiddleware` (headers passed through Foundry's authenticated frontend); a non-Foundry deployment can swap the middleware to enforce its own auth scheme before lifting the keys.

Also clarified the surrounding comment to explicitly call out that wire headers are never trusted directly. Fixed in da154e963.

python/packages/hosting-responses/agent_framework_hosting_responses/_channel.py
eavanvalkenburg: Right — partition co-location and partition ownership are two separate concerns and the docstring conflated them. Updated to spell out the layered defense:

- Channel level (this code path): forwards the caller-supplied `previous_response_id` to the factory as a co-location hint so id backends that embed partition keys (Foundry storage) can place the new record on the same partition as the chain. **This is a performance hint, not a security boundary.**
- Storage level (Foundry storage provider): validates the embedded partition against the bound `IsolationKeys` for the request and rejects writes whose partition doesn't match the authenticated caller's isolation. A spoofed `previous_response_id` pointing at someone else's partition would be rejected at the storage call.
- Host level: the isolation middleware (now also the source of truth for the channel's chat-iso, see #3198157929) is the place auth is established before the route is entered.

Docstring expanded in da154e963.

python/packages/hosting-responses/agent_framework_hosting_responses/_channel.py
eavanvalkenburg: You're right — emitting `response.failed` to the wire without calling `deliver_response` left host-side state inconsistent with what the user already saw (chain anchored on this `response_id` would be missing the partial assistant text on the next turn, and any non-originating push targets would silently miss the turn).

Now wraps the failure path with a best-effort `deliver_response(request, HostedRunResult(text=accumulated))` before yielding the failed event, so the host history /
push fan-out reflects whatever made it onto the wire. The delivery is itself wrapped in try/except so a delivery failure doesn't suppress the failure event reaching the client.

Fixed in da154e963.

python/packages/hosting-responses/agent_framework_hosting_responses/_parsing.py
eavanvalkenburg: Good catch — `instructions` was duplicated. The body's `instructions` field maps to `ChatMessage(role='system')` via `build_chat_messages`, so it should never also be passed through as a free-form `additional_property`. Removed it from `_RESPONSES_OPTION_PASSTHROUGH` and removed the now-redundant `if key == 'instructions': continue` short-circuit, with comments at both sites pointing at the system-message contract. Fixed in da154e963.

python/packages/hosting-responses/agent_framework_hosting_responses/_parsing.py
```diff
@@ -0,0 +1,245 @@
...
+    - ``messages`` is the parsed conversation (``instructions`` is prepended
+      as a system message when present).
+    - ``options`` is a ``ChatOptions``-shaped dict with the model-tunable
+      fields the channel lifted off the body.
+    - ``session`` is a :class:`ChannelSession` keyed by
+      ``previous_response_id`` when one was supplied, else ``None``.
+    """
+    messages = messages_from_responses_input(body.get("input"))
+
+    if (instructions := body.get("instructions")) and isinstance(instructions, str):
```
eavanvalkenburg: instructions are a field in `ChatOptions` it should not be lifted out and put into a Message

python/packages/hosting-responses/agent_framework_hosting_responses/_channel.py
```diff
@@ -0,0 +1,427 @@
...
+        *,
+        path: str = "",
+        run_hook: ChannelRunHook | None = None,
+        response_hook: ChannelResponseHook | None = None,
+        stream_transform_hook: ChannelStreamTransformHook | None = None,
+        response_id_factory: Callable[..., str] | None = None,
+    ) -> None:
+        """Create a Responses channel.
+
+        Args:
```
eavanvalkenburg: this should be:

```suggestion
        Keyword Args:
```

PR #5640, opened 2026-05-05
Python: add agent-framework-hosting-invocations channel

eavanvalkenburg:
### Motivation and Context

Implements **§7 ("Built-in channels — Invocations")** of [SPEC-002](https://github.com/microsoft/agent-framework/blob/feature/python-hosting/docs/decisions/0026-hosting-channels.md) (merged via #5549). Pairs with `ResponsesChannel` to make an `AgentFrameworkHost` runtime-compatible with the **Foundry Hosted Agents** platform (per ADR §11).

### Description

Adds the new `agent-framework-hosting-invocations` package (`python/packages/hosting-invocations/`) with:

* `InvocationsChannel` — mounts `POST /invocations/invoke` (configurable), accepts the host-native JSON envelope used by Foundry Hosted Agents, and routes to the host's target.
* Streaming + non-streaming variants.
* Tests covering channel construction and end-to-end invocation handling.

### Stack

PR-4 of 10. **Depends on #PR-2** (`feat/hosting-core`). Independent of PR-3.

### Contribution Checklist

### DIFF
--- a/python/packages/hosting-invocations/agent_framework_hosting_invocations/_channel.py
+++ b/python/packages/hosting-invocations/agent_framework_hosting_invocations/_channel.py
@@ -1,0 +1,165 @@
+# Copyright (c) Microsoft. All rights reserved.
+
+"""Minimal ``POST /invoke`` channel.
+
+Inspired by ``agent-framework-foundry-hosting``'s ``InvocationsHostServer``.
+A framework-agnostic surface for callers that just want to send a message and
+get an answer back — no OpenAI-style envelope, no Responses item lattice.
+"""
+
+from __future__ import annotations
+
+from collections.abc import AsyncIterator, Awaitable
+from typing import Any, cast
+
+from agent_framework_hosting import (
+    ChannelContext,
+    ChannelContribution,
+    ChannelRequest,
+    ChannelRunHook,
+    ChannelSession,
+    ChannelStreamTransformHook,
+    apply_run_hook,
+    logger,
+)
+from starlette.requests import Request
+from starlette.responses import JSONResponse, Response, StreamingResponse
+from starlette.routing import Route
+
+
+class InvocationsChannel:
+    """Minimal ``POST /invoke`` surface.
+
+    A run hook can rewrite the channel request (e.g. inject a session, add
+    options) before the host invokes the agent. A stream-transform hook can
+    rewrite or drop ``AgentResponseUpdate`` chunks before they hit the wire.
+    """
+
+    name = "invocations"
+
+    def __init__(
+        self,
+        *,
+        path: str = "/invocations",
+        run_hook: ChannelRunHook | None = None,
+        stream_transform_hook: ChannelStreamTransformHook | None = None,
+    ) -> None:
+        """Configure the invocations endpoint.
+
+        ``path`` is the mount root the host prefixes when registering this
+        channel's routes (the actual handler is ``POST {path}/invoke``).
+        ``run_hook`` may rewrite the :class:`ChannelRequest` before the host
+        invokes the target — typically to attach session metadata or
+        translate the wire payload into ``ChatMessage`` instances.
+        ``stream_transform_hook`` lets callers map or drop individual
+        ``AgentResponseUpdate`` chunks while streaming.
+        """
+        self.path = path
+        self._hook = run_hook
+        self._stream_transform_hook = stream_transform_hook
+        self._ctx: ChannelContext | None = None
+
+    def contribute(self, context: ChannelContext) -> ChannelContribution:
+        """Capture the host-supplied context and register ``POST /invoke``."""
+        self._ctx = context
+        return ChannelContribution(routes=[Route("/invoke", self._handle, methods=["POST"])])
+
+    async def _handle(self, request: Request) -> Response:
+        """Handle a single ``POST /invoke`` call.
+
+        Validates the JSON body shape, builds a :class:`ChannelRequest`
+        (optionally with a ``ChannelSession`` keyed by ``session_id``),
+        runs the configured ``run_hook``, and either streams SSE chunks
+        when ``stream`` is true or returns a single JSON ``{response,
+        session_id}`` envelope.
+        """
+        if self._ctx is None:  # pragma: no cover - guarded by Channel lifecycle
+            return JSONResponse({"error": "channel not initialized"}, status_code=500)
+        try:
+            body: Any = await request.json()
+        except Exception:
+            return JSONResponse({"error": "invalid json"}, status_code=400)
+
+        if not isinstance(body, dict):
+            return JSONResponse({"error": "request body must be an object"}, status_code=422)
+        body_map: dict[str, Any] = cast("dict[str, Any]", body)
+
+        message = body_map.get("message")
+        if not isinstance(message, str) or not message:
+            return JSONResponse({"error": "missing or empty 'message'"}, status_code=422)
+
+        session_id = body_map.get("session_id")
+        if session_id is not None and not isinstance(session_id, str):
+            return JSONResponse({"error": "'session_id' must be a string"}, status_code=422)
+
+        session = ChannelSession(isolation_key=f"invocations:{session_id}") if session_id else None
+
+        attributes: dict[str, Any] = {}
+        if session_id:
+            attributes["session_id"] = session_id
+
+        channel_request = ChannelRequest(
+            channel=self.name,
+            operation="invoke",
+            input=message,
+            session=session,
+            stream=bool(body_map.get("stream")),
+            attributes=attributes,
+        )
+
+        if self._hook is not None:
+            channel_request = await apply_run_hook(
+                self._hook,
+                channel_request,
+                target=self._ctx.target,
+                protocol_request=body_map,
+            )
+
+        if channel_request.stream:
+            return StreamingResponse(
+                self._stream(channel_request),
+                media_type="text/event-stream",
+                headers={"Cache-Control": "no-cache", "X-Accel-Buffering": "no"},
+            )
+
+        result = await self._ctx.run(channel_request)
+        return JSONResponse({"response": result.text, "session_id": session_id})
+
+    async def _stream(self, request: ChannelRequest) -> AsyncIterator[str]:
+        """Yield bare ``data:`` SSE lines for each text chunk + a final ``[DONE]``."""
+        if self._ctx is None:  # pragma: no cover - guarded by Channel lifecycle
+            yield "event: error\ndata: channel not initialized\n\n"
+            return
+        try:
+            stream = self._ctx.run_stream(request)
+            async for update in stream:
+                if self._stream_transform_hook is not None:
+                    transformed = self._stream_transform_hook(update)
+                    if isinstance(transformed, Awaitable):
+                        transformed = await transformed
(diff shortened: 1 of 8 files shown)


python/packages/hosting-invocations/agent_framework_hosting_invocations/_channel.py
```diff
@@ -0,0 +1,165 @@
...
+
+        if self._hook is not None:
+            channel_request = await apply_run_hook(
+                self._hook,
+                channel_request,
+                target=self._ctx.target,
+                protocol_request=body_map,
+            )
+
+        if channel_request.stream:
```
moonbox3, line 126: Should we be returning 200 here before the stream has actually opened? `_stream` swallows everything from `run_stream` into an `event: error` SSE frame on a 200 response, so a hard failure (target rejection, isolation/auth error, schema mismatch) is indistinguishable from success at the HTTP layer. Durable-task drivers that retry on 5xx will silently drop these. Worth priming the iterator once before constructing the `StreamingResponse` so the open-time error materialises as a real status code?

python/packages/hosting-invocations/agent_framework_hosting_invocations/_channel.py
```diff
@@ -0,0 +1,165 @@
...
+                    if transformed is None:
+                        continue
+                    update = transformed
+                chunk = getattr(update, "text", None)
+                if chunk:
+                    # Each text chunk is its own SSE event so curl-friendly
+                    # consumers can read it directly. Newlines inside the
+                    # chunk are escaped per SSE spec by emitting one
+                    # ``data:`` line per source line.
+                    for line in str(chunk).split("\n"):
```
moonbox3: SSE treats `\r`, `\n`, and `\r\n` all as line terminators. Splitting on `\n` only leaves embedded `\r` inside the `data:` payload, which `EventSource` interprets as a line break with no `data:` prefix, ending the event early or corrupting framing. Model output containing CR (Windows newlines, terminal control, code blocks) would hit this. There is also a small response-splitting angle: a chunk containing `\n\ndata: [DONE]\n\n` could forge a premature DONE frame; re-prefixing every line behind any of the three terminators is what defends against that.

```suggestion
                    import re
                    for line in re.split(r"\r\n|\r|\n", str(chunk)):
                        yield f"data: {line}\n"
                    yield "\n"
```

python/packages/hosting-invocations/agent_framework_hosting_invocations/_channel.py
```diff
@@ -0,0 +1,165 @@
...
+                        yield f"data: {line}\n"
+                    yield "\n"
+            try:
+                # Finalize so context-provider / history hooks on the agent
+                # still run even though we are emitting our own SSE.
+                await stream.get_final_response()
```
moonbox3, line 201: Is "best-effort" the right framing for this? `get_final_response()` is what triggers history-provider persistence per the comment above it, so a 5xx from a hosted history provider, a disk-full on `FileHistoryProvider`, or a serialization error on `additional_properties` silently disappears. The client gets `[DONE]` and a 200, then the next turn behaves as though the prior turn never happened. At minimum could we emit an `event: error` before `[DONE]` so the client knows the turn was not persisted?

python/packages/hosting-invocations/agent_framework_hosting_invocations/_channel.py
```diff
@@ -0,0 +1,165 @@
...
+
+    async def _stream(self, request: ChannelRequest) -> AsyncIterator[str]:
+        """Yield bare ``data:`` SSE lines for each text chunk + a final ``[DONE]``."""
+        if self._ctx is None:  # pragma: no cover - guarded by Channel lifecycle
+            yield "event: error\ndata: channel not initialized\n\n"
+            return
+        try:
+            stream = self._ctx.run_stream(request)
+            async for update in stream:
+                if self._stream_transform_hook is not None:
```
moonbox3, line 176: I don't see any tests for `stream_transform_hook`?

python/packages/hosting/agent_framework_hosting/_host.py
```diff
@@ -0,0 +1,946 @@
...
+                    "deliver_response: no known identity for isolation_key=%s on channel=%s",
+                    isolation_key,
+                    channel_name,
+                )
+                skipped.append(token)
+                continue
+            try:
+                await channel.push(dest_identity, payload)
+            except Exception:
+                logger.exception("deliver_response: push failed for target=%s", token)
```
moonbox3: Should "no destinations resolvable" and "every destination errored" really collapse into the same fallback?

python/packages/hosting/agent_framework_hosting/_host.py
```diff
@@ -0,0 +1,946 @@
...
+    """Yield, swallowing the ``RuntimeError`` ``ResponseStream`` raises on double-consume.
+
+    The bridge stream calls ``get_final_response()`` after iterating the
+    workflow stream so the workflow's cleanup hooks run; on some paths the
+    stream considers itself already finalized and raises, which we treat
+    as benign — we're only after the side effect.
+    """
+    try:
+        yield
+    except RuntimeError as exc:
```
moonbox3, line 278: The `RuntimeError` arm matches by type, not by message, so any unrelated `RuntimeError` from finalize (executor bugs, runner-context state errors, checkpoint-store failures during the post-run flush) is demoted to `debug` and disappears at default log levels. The `except Exception` below also swallows things like `OSError` from checkpoint persistence. Since cleanup *is* the persistence step for workflows, this can hide a checkpoint that never got written, and the next turn restores stale state. Could we narrow to the specific double-consume signal (message check or a dedicated sentinel exception from `ResponseStream`) and re-raise the rest?

python/packages/hosting/agent_framework_hosting/_host.py
```diff
@@ -0,0 +1,946 @@
...
+
+        async def _maybe_restore() -> None:
+            if storage is None:
+                return
+            latest = await storage.get_latest(workflow_name=workflow.name)
+            if latest is None:
+                return
+            # Drain the restoration stream so the no-op invocation actually
+            # rehydrates state before the real run starts.
+            async for _ in workflow.run(
```
moonbox3: What happens if restoration silently produces zero events (stale or missing `checkpoint_id`, storage returned empty)?

python/packages/hosting-invocations/agent_framework_hosting_invocations/_channel.py
eavanvalkenburg: This is the conventional SSE contract: the HTTP status is committed when ASGI sends headers, before the streaming generator runs. Once `200 + text/event-stream` is on the wire, errors are conveyed as `event: error` SSE frames — `EventSource` and OpenAI-style SSE consumers treat that as a terminal error condition and surface it to user code. Switching to a deferred 5xx after partial bytes is unsupported by ASGI / HTTP/1.1.

Hard run-acquisition failures (`run_stream` raises immediately, before any chunk hits the wire) therefore surface as the first SSE frame, not as an HTTP code; expanded the docstring to call this contract out explicitly. Fixed in ef2e880a0.

python/packages/hosting-invocations/agent_framework_hosting_invocations/_channel.py
eavanvalkenburg: Good catch — the SSE spec treats `\r`, `\n`, and `\r\n` all as line terminators, so an embedded `\r` would corrupt framing. Switched from `split('\n')` to `splitlines()` so any of the three terminators produce separate `data:` lines on the wire. Added a regression test (`test_streaming_chunk_with_crlf_splits_into_separate_data_lines`) asserting CRLF-bearing chunks split cleanly. Fixed in ef2e880a0.

python/packages/hosting-invocations/agent_framework_hosting_invocations/_channel.py
eavanvalkenburg: Agreed — "best-effort" was the wrong framing. `get_final_response()` is what triggers history-provider persistence on the agent side, so a 5xx from a hosted history provider, a disk-full on `FileHistoryProvider`, or a context-provider error should not be silently swallowed.

Now wrapped in a dedicated except that emits `event: error` (with the exception summary as `data:` lines) and returns WITHOUT yielding `[DONE]`, so the client sees the same terminal-error semantics it would for any mid-stream failure. New test `test_streaming_finalize_error_emits_error_frame_no_done` covers the contract. Fixed in ef2e880a0.

python/packages/hosting-invocations/agent_framework_hosting_invocations/_channel.py
eavanvalkenburg: Right — coverage gap. Added three tests covering the `stream_transform_hook` contract:

- `test_stream_transform_hook_can_rewrite_chunks` — sync hook that rewrites text
- `test_stream_transform_hook_can_drop_chunks` — `None` return drops the update
- `test_stream_transform_hook_supports_async` — awaitable return value is awaited

Fixed in ef2e880a0.

python/packages/hosting/agent_framework_hosting/_host.py
eavanvalkenburg: These are already two distinct paths in the code — sorry the diff didn't make that obvious. `_deliver_response` distinguishes:

- **Skipped** (no destinations resolvable): no channel of that name / no `ChannelPush` impl / no recorded identity → `skipped` tuple → `DeliveryReport.skipped`. The originating fallback fires here so the user gets a reply.
- **Failed** (every destination errored): `ChannelPush.push` raised → `failed` tuple → `DeliveryReport.failed`. The originating fallback **does not** fire here — surfacing through `failed` lets the originating channel decide whether to render a degraded reply rather than double-deliver on a flaky link.

See `_deliver_response` lines 1080-1102 and the `DeliveryReport` docstring for the full contract. Resolving.

python/packages/hosting/agent_framework_hosting/_host.py
eavanvalkenburg: Good catch — `RuntimeError` matched too broadly. Narrowed the catch to the two documented benign messages (`Inner stream not available`, `Event loop is closed`); any other `RuntimeError` (executor bug, runner-context state error, checkpoint-store `RuntimeError` during the post-run flush, …) now logs at ERROR with a full traceback so it's no longer silenced as benign cleanup noise. Still doesn't propagate (we're in an async-generator `finally` during teardown — see the docstring).

Fixed in PR #5638 commit dccc38426; PR-4 has been rebased onto that commit so the change is in this stack too.

python/packages/hosting/agent_framework_hosting/_host.py
eavanvalkenburg: Right — silent zero-event restoration is a real failure mode (stale `checkpoint_id`, partially-written checkpoint, storage returned empty). Both `_restore_workflow_checkpoint` and `_restore_workflow_checkpoint_streaming` now log a WARNING with `(workflow=…, checkpoint_id=…)` when a non-`None` latest checkpoint drains to zero events, so operators can detect silent state-loss without sifting through INFO logs.

Fixed in PR #5638 commit dccc38426; PR-4 has been rebased onto that commit so the change is in this stack too.

PR #5641, opened 2026-05-05
Python: add agent-framework-hosting-activity-protocol channel

eavanvalkenburg:
### Motivation and Context

Implements the **Activity Protocol channel** described in [SPEC-002](https://github.com/microsoft/agent-framework/blob/feature/python-hosting/docs/decisions/0026-hosting-channels.md) §7 (merged via #5549). This is the lower-level Bot Service / Activity Protocol surface — it covers any Activity-Protocol-speaking client (Teams, Direct Line, Web Chat, Slack via Bot Service, etc.).

A separate **PR-5b** adds a higher-level `agent-framework-hosting-teams` package built on the `microsoft-teams-apps` SDK for Teams-native affordances (Adaptive Cards, Citations, streaming via `ctx.stream`).

### Description

Adds `agent-framework-hosting-activity-protocol` (`python/packages/hosting-activity-protocol/`):

* `ActivityProtocolChannel` — mounts `POST /activity` (configurable), validates Bot Framework JWT, decodes Activity Protocol payloads into `ChannelRequest`, replies as outbound activities.
* Token validation + identity derivation from the Activity sender.
* Tests for channel wiring, JWT path, and request/response shaping.

This is the **rename target** of the previously-named `hosting-teams` package — the directory rename frees `hosting-teams` for the Teams-SDK-based channel in PR-5b. `TeamsChannel` → `ActivityProtocolChannel`, `name="teams"` → `"activity"`, default mount `/teams` → `/activity`.

### Stack

PR-5a of 10. **Depends on #PR-2** (`feat/hosting-core`). Independent of PR-5b — the two Teams-related packages are intentionally separate (different audiences, see PR-5b for the comparison table).

### Contribution Checklist

- [ ] **Is this a breaking change?** No — new package (the previous `hosting-teams` package was never released).

### DIFF
--- a/python/packages/hosting-activity-protocol/agent_framework_hosting_activity_protocol/_channel.py
+++ b/python/packages/hosting-activity-protocol/agent_framework_hosting_activity_protocol/_channel.py
@@ -1,0 +1,571 @@
+# Copyright (c) Microsoft. All rights reserved.
+
+r"""Built-in channel: Bot Framework Activity Protocol (Azure Bot Service).
+
+Activity Protocol is the Bot Framework messaging shape used by Azure Bot
+Service to fan one bot endpoint out across many surfaces (Microsoft
+Teams, Slack, Webex, Telegram, …). An incoming ``Activity`` is POSTed to
+your bot's ``/messages`` endpoint, and you reply by POSTing one or more
+``Activity`` objects back to the conversation URL the inbound activity
+carried in ``serviceUrl``. Auth is an OAuth2 client-credentials token
+from Entra (the legacy multi-tenant ``botframework.com`` authority for
+public Bot Framework channels, or your own tenant for single-tenant
+bots).
+
+This is the channel-neutral Activity-Protocol channel — it surfaces what
+every Bot-Service-connected channel has in common (text in, text out).
+For deeper Microsoft Teams affordances (adaptive cards, message
+extensions, dialogs, SSO, …) on the same Bot Service transport, see the
+companion ``agent-framework-hosting-teams`` package.
+
+This channel handles:
+
+- inbound ``message`` activities — text and attachments resolved to URIs,
+- outbound replies via ``POST /v3/conversations/{id}/activities``,
+- streaming via ``PUT /v3/conversations/{id}/activities/{id}`` mid-stream
+  edits (Teams supports updateActivity in personal chats and groups),
+- typing indicators while the agent works,
+- per-conversation isolation key ``activity:<conversation_id>`` so a Responses
+  caller can resume a Teams conversation by passing the conversation id,
+- two credential modes for the outbound token — **client secret** or
+  **certificate** (for tenants that disallow secrets) — both via
+  ``azure.identity.aio``,
+- dev-mode auth bypass when no credentials are passed so the Bot Framework
+  Emulator can hit the endpoint with no credentials.
+
+Out of scope for the prototype: full JWT validation of inbound requests,
+adaptive cards, file uploads, OAuth sign-in flows, and the Teams streaming
+preview API (``StreamItem``).
+
+Generating a certificate
+------------------------
+For tenants that disallow client secrets, register a certificate on your
+Bot Framework / Entra app instead. Self-signed PEM (private key + cert in
+one file) is what ``azure.identity.CertificateCredential`` expects::
+
+    # 1. Generate a 2048-bit RSA key + self-signed cert (10y), single PEM.
+    openssl req -x509 -newkey rsa:2048 -nodes -days 3650 \\
+        -subj "/CN=my-teams-bot" \\
+        -keyout teams-bot.key -out teams-bot.crt
+    cat teams-bot.key teams-bot.crt > teams-bot.pem
+
+    # 2. Upload teams-bot.crt to your Entra app under
+    #    "Certificates & secrets" → "Certificates" → "Upload certificate".
+
+    # 3. Point the channel at the combined PEM:
+    ActivityProtocolChannel(
+        app_id="<app id>",
+        tenant_id="<tenant id>",         # or "botframework.com" for legacy bots
+        certificate_path="teams-bot.pem",
+    )
+
+To encrypt the private key, drop ``-nodes`` from the openssl command and
+pass ``certificate_password=<bytes>`` to the channel.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import time
+from collections.abc import Awaitable, Mapping
+from typing import Any
+
+import httpx
+from agent_framework import (
+    AgentResponse,
+    AgentResponseUpdate,
+    Content,
+    Message,
+    ResponseStream,
+)
+from agent_framework_hosting import (
+    ChannelContext,
+    ChannelContribution,
+    ChannelRequest,
+    ChannelRunHook,
+    ChannelSession,
+    ChannelStreamTransformHook,
+    apply_run_hook,
+    logger,
+)
+from azure.core.credentials_async import AsyncTokenCredential
+from azure.identity.aio import CertificateCredential, ClientSecretCredential
+from starlette.requests import Request
+from starlette.responses import JSONResponse, Response
+from starlette.routing import Route
+
+# Bot Framework v4 multi-tenant authority used by the public Bot Framework
+# channels (including Microsoft Teams). Single-tenant bots should override
+# ``tenant_id`` with their own tenant.
+_BOTFRAMEWORK_TENANT = "botframework.com"
+_BOTFRAMEWORK_SCOPE = "https://api.botframework.com/.default"
+
+
+def activity_protocol_isolation_key(conversation_id: Any) -> str:
+    """Build the namespaced isolation key the Teams channel writes under.
+
+    Exposed at module scope so other channels' run hooks can opt into the
+    same per-conversation session (e.g. a Responses caller resuming a Teams
+    conversation by passing the conversation id).
+    """
+    return f"activity:{conversation_id}"
+
+
+def _parse_activity(activity: Mapping[str, Any]) -> Message:
+    """Translate one Bot Framework ``message`` Activity into an Agent Framework Message.
+
+    Pulls the activity's ``text`` plus any image/file attachments with a
+    ``contentType`` and resolvable URL into ``Content`` parts. If the
+    activity has no usable parts an empty text part is emitted so the
+    caller never sees a content-less message.
+    """
+    parts: list[Content] = []
+    if (text := activity.get("text")) and isinstance(text, str):
+        parts.append(Content.from_text(text=text))
+
+    for attachment in activity.get("attachments") or []:
+        if not isinstance(attachment, Mapping):
+            continue
+        url = attachment.get("contentUrl") or attachment.get("content")
+        content_type = attachment.get("contentType")
+        if isinstance(url, str) and isinstance(content_type, str) and "/" in content_type:
+            parts.append(Content.from_uri(uri=url, media_type=content_type))
+
+    if not parts:
+        parts.append(Content.from_text(text=""))
+    return Message("user", parts)
+
+
+class ActivityProtocolChannel:
+    """Microsoft Teams channel via Bot Framework v4 webhook.
+
+    Streaming
+    ---------
+    When ``stream=True`` (default), the channel sends an initial placeholder
+    activity, then edits it in place as the agent emits ``AgentResponseUpdate``
+    chunks (``PUT /v3/conversations/{id}/activities/{id}``). When ``stream=False``
+    it just sends the final reply. A ``stream_transform_hook`` can rewrite or
+    drop individual updates before they hit the wire.
+    """
+
+    name = "activity"
+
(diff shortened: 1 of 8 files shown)


python/packages/hosting-activity-protocol/agent_framework_hosting_activity_protocol/_channel.py
```diff
@@ -0,0 +1,571 @@
...
+
+        Only ``message`` activities are processed; ``conversationUpdate``,
+        ``invoke``, ``typing`` and other activity types are silently
+        acknowledged. The webhook always returns 200 (or 202 for ignored
+        types) so Bot Framework can dequeue the activity even if our
+        downstream processing fails — failures are logged and re-tried by
+        the user, not by Teams.
+        """
+        try:
+            activity = await request.json()
```
moonbox3, line 417: Are we comfortable shipping this endpoint with no inbound auth? Module docstring at line 36 lists "full JWT validation of inbound requests" as out of scope, but the channel still acquires a real Bot Framework token in `_get_token` and POSTs it to whatever `serviceUrl` the request body contains. Anyone on the internet who finds the URL can forge an Activity with attacker-controlled `serviceUrl` and harvest our Entra token from the outbound reply. 

Should the channel refuse to attach a credential unless an explicit `insecure_skip_jwt_validation=True` flag is set, and should the README call this out at the top rather than in a side note?

python/packages/hosting-activity-protocol/agent_framework_hosting_activity_protocol/_channel.py
```diff
@@ -0,0 +1,571 @@
...
+            "serviceUrl": inbound.get("serviceUrl"),
+            "text": text,
+            "textFormat": "plain",
+        }
+
+    async def _send_message(self, inbound: Mapping[str, Any], text: str) -> str | None:
+        """POST a new Activity. Returns the assigned activity id."""
+        if self._http is None:  # pragma: no cover - guarded by lifecycle
+            raise RuntimeError("activity channel not started")
+        service_url = str(inbound.get("serviceUrl") or "").rstrip("/")
```
moonbox3, line 674: Should we be allowlisting `serviceUrl` before attaching the bearer token? Bot Framework guidance is to verify the host against the `serviceurl` claim of the validated inbound JWT, or at minimum against `*.botframework.com` / `smba.trafficmanager.net`. Without that, `serviceUrl="http://attacker.example"` plus any `conversation.id` POSTs our Entra access token in plaintext to the attacker. Same issue at `_update_activity` (line 532) and `_send_typing` (line 553). 

Also `conversation_id` is interpolated raw into the path — what happens with `id="../../foo"` or one containing `@otherhost/`? Should we `urllib.parse.quote(conversation_id, safe="")` and reject ids that don't match an expected charset?

python/packages/hosting-activity-protocol/agent_framework_hosting_activity_protocol/_channel.py
```diff
@@ -0,0 +1,571 @@
...
+        except Exception:
+            return JSONResponse({"error": "invalid json"}, status_code=400)
+
+        # We accept only message activities for now. ``conversationUpdate``,
+        # ``invoke``, ``typing`` and friends are silently ack'd.
+        if activity.get("type") != "message":
+            return JSONResponse({}, status_code=202)
+
+        try:
+            await self._process_activity(activity)
```
moonbox3, line 435: The docstring justifies always returning 200 to dequeue, but does this drop the retry signal Bot Service relies on for transient failures? A 401 from the outbound reply, an Entra outage, a tool crash, a `_send_message` raise_for_status all become a silent 200 with one log line and the user sees nothing in the conversation. 

Should transient classes (5xx outbound, token-acquire failure, network) return 5xx so Bot Service backs off and retries, while terminal classes still return 200? Right now a persistently broken credential is invisible at default log levels.

python/packages/hosting-activity-protocol/agent_framework_hosting_activity_protocol/_channel.py
```diff
@@ -0,0 +1,571 @@
...
+
+        async def send_initial_placeholder() -> None:
+            nonlocal activity_id, last_edit_at
+            try:
+                activity_id = await self._send_message(inbound, "…")
+                last_edit_at = time.monotonic()
+            except Exception:  # pragma: no cover
+                logger.exception("Teams placeholder send failed")
+
+        async def edit_worker() -> None:
```
moonbox3, line 569: What happens when `send_initial_placeholder` fails and the stream then produces text? `activity_id` stays `None`, `accumulated` grows past `""`, `last_sent` never moves, and the loop exit condition `accumulated == last_sent` is never satisfiable. After the `finally` block sets `worker_done` and `wake`, the worker hits `continue` once, then `await wake.wait()` blocks forever and the parent hangs on `await edit_task`. Should the loop also bail when `activity_id is None and worker_done.is_set()`?

```suggestion
        while not (worker_done.is_set() and (activity_id is None or accumulated == last_sent)):
            await wake.wait()
            wake.clear()
            if activity_id is None or accumulated == last_sent:
                continue
```

python/packages/hosting-activity-protocol/tests/test_channel.py
```diff
@@ -0,0 +1,182 @@
+# Copyright (c) Microsoft. All rights reserved.
+
+"""Unit tests for :mod:`agent_framework_hosting_activity_protocol`.
+
+The Bot Framework outbound calls and azure-identity credentials are mocked
+out so the suite never touches the network. Live token acquisition,
+streaming edits and certificate paths are out of scope here.
```
moonbox3, line 7: Are we leaving the entire streaming path (`_stream_to_conversation`, `_channel.py:398-487`) untested on purpose? That is the bulk of the production logic: placeholder send, interim PUT edits, throttled wake loop, final flush, the empty-stream "(no response)" replacement.

Could we add at least one happy-path streaming test plus the empty-stream case? The deadlock in the `edit_worker()` comment above is something that could be caught by a test.

python/packages/hosting-activity-protocol/tests/test_channel.py
```diff
@@ -0,0 +1,182 @@
...
+
+        async def _coro() -> _FakeAgentResponse:
+            return _FakeAgentResponse(text=self._reply)
+
+        return _coro()
+
+
+def _make_teams(stream: bool = False) -> tuple[ActivityProtocolChannel, _FakeAgent]:
+    agent = _FakeAgent("hi there")
+    ch = ActivityProtocolChannel(stream=stream, send_typing_action=False)
```
moonbox3, line 85: I couldn't find a test that asserts the outbound POST carries `Authorization: Bearer <token>` when a credential is configured, or that dev mode sends no Authorization header.

python/packages/hosting-activity-protocol/agent_framework_hosting_activity_protocol/_channel.py
eavanvalkenburg: Fair concern — JWT validation is still out of scope for the prototype (key rotation + OpenID config caching are non-trivial and we don't want to silently ship a half-baked validator), but the channel now defends in depth:

- New `inbound_auth_validator: Callable[[Request], Awaitable[bool]]` constructor option. When set, it's invoked before the activity is parsed; `False`/raise → HTTP 401. Operators can plug in a Bot Framework JWT validator (e.g. backed by `botbuilder-core`'s `AuthenticationConfiguration` or Microsoft's `microsoft-teams-apps` SDK auth helpers) without us baking a specific JWT library into the channel.
- When **no** validator is configured the `_on_startup` hook logs a loud `WARNING` so the dev-mode bypass cannot accidentally ship to production.
- Docstring updated to spell out the contract: 'plug a validator OR terminate auth in front of the channel (APIM, Application Gateway) before exposing this endpoint to a public network.'

Fixed in f60466c6b.

python/packages/hosting-activity-protocol/agent_framework_hosting_activity_protocol/_channel.py
eavanvalkenburg: Agreed — outbound bearer token to an attacker-controlled host is a real risk. Added a `service_url_allowed_hosts` constructor option, default `('botframework.com', 'smba.trafficmanager.net')`. The webhook now validates the inbound activity's `serviceUrl` host against the allow-list (suffix match) and returns HTTP 400 when it falls outside, **before** any outbound POST is attempted — so the bearer token never leaves the process when the destination is suspicious.

- Default covers public Bot Framework + Microsoft Teams.
- Operators with sovereign-cloud / Direct-Line-only deployments pass an extended tuple.
- Pass `service_url_allowed_hosts=()` to disable the check (only safe paired with a strong `inbound_auth_validator`).

Test matrix in `TestServiceUrlAllowList`. Fixed in f60466c6b.

python/packages/hosting-activity-protocol/agent_framework_hosting_activity_protocol/_channel.py
eavanvalkenburg: Right — collapsing every failure into 200 dropped the retry signal. The webhook now distinguishes:

- **Transient outbound failure** (`httpx.HTTPError` / new `_OutboundError` marker for token acquisition issues): we surface HTTP **502** so Bot Service retries the inbound activity. This covers 401 from the outbound reply, an Entra outage, `_send_message.raise_for_status` on a 5xx, and a network error on the way out.
- **Deterministic agent / processing failure** (anything else that escapes `_process_activity`): we still return **200** so Bot Service does not retry the same broken activity in a loop. Operators pick this up via logs / telemetry rather than via user-visible duplicates.

Tests `test_outbound_http_error_returns_502` and `test_deterministic_agent_failure_returns_200` lock the contract. Fixed in f60466c6b.

python/packages/hosting-activity-protocol/agent_framework_hosting_activity_protocol/_channel.py
eavanvalkenburg: Real bug — caught it. When `send_initial_placeholder` raises, `activity_id` stays `None`, the `edit_worker` loop's exit condition `accumulated == last_sent` is unreachable while no PUT is possible, and after `worker_done.set()` the worker blocks on `wake.wait()` forever.

Fixed by:
- Tracking `placeholder_ok: bool` separately from `activity_id`.
- `edit_worker` returns immediately when `placeholder_ok` is False (no PUT path possible).
- The final flush detects the failure and POSTs a single new activity at the end with whatever accumulated, so the user still gets the agent's reply (no silent loss).

Test `test_stream_placeholder_failure_falls_back_to_single_post` wraps the call in `asyncio.wait_for(timeout=2.0)` to guard against a deadlock regression. Fixed in f60466c6b.

python/packages/hosting-activity-protocol/tests/test_channel.py
```diff
@@ -0,0 +1,182 @@
...
+    def run(self, messages: Any = None, *, stream: bool = False, **kwargs: Any) -> Any:
+        self.runs.append({"messages": messages, "stream": stream, "kwargs": kwargs})
+
+        async def _coro() -> _FakeAgentResponse:
+            return _FakeAgentResponse(text=self._reply)
+
+        return _coro()
+
+
+def _make_teams(stream: bool = False) -> tuple[ActivityProtocolChannel, _FakeAgent]:
```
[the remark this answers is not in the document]
eavanvalkenburg: Same concern as #3198334465 — coverage gap on the streaming path. New tests in `TestStreaming`:

- `test_stream_sends_placeholder_and_edits` — placeholder POST + progressive PUT edits with the final accumulated text
- `test_stream_placeholder_failure_falls_back_to_single_post` — guards against the deadlock fix regression with `asyncio.wait_for(timeout=2.0)`
- `test_stream_with_no_text_replaces_placeholder` — empty stream replaces the `…` placeholder with `(no response)`

These exercise `_stream_to_conversation`, the throttled wake loop, and the final-flush branches. Fixed in f60466c6b.

python/packages/hosting-activity-protocol/tests/test_channel.py
eavanvalkenburg: Coverage gap acknowledged — the streaming path was the bulk of the production logic and was completely untested. New `TestStreaming` class covers placeholder + interim PUT edits, the placeholder-failure fallback (with a timeout guard against the deadlock regression — see #3198330424), and the empty-stream `(no response)` placeholder replacement. Fixed in f60466c6b.

python/packages/hosting-activity-protocol/tests/test_channel.py
eavanvalkenburg: Coverage gap acknowledged. New `TestOutboundAuthHeader` class:

- `test_no_credential_sends_no_authorization_header` — dev mode (no credential) sends no `Authorization` header
- `test_with_credential_sends_bearer_token` — production mode (injected fake credential returning `tok-abc123`) sends `Authorization: Bearer tok-abc123` on the outbound POST

Fixed in f60466c6b.

PR #5642, opened 2026-05-05
Python: add agent-framework-hosting-teams channel (microsoft-teams-apps SDK)

eavanvalkenburg:
### Motivation and Context

Implements the higher-level **Teams channel** described in [SPEC-002](https://github.com/microsoft/agent-framework/blob/feature/python-hosting/docs/decisions/0026-hosting-channels.md) §7 + §"Teams-native fast follow" (merged via #5549). Built on the [`microsoft-teams-apps`](https://microsoft.github.io/teams-sdk/python/) SDK so users get Teams-native affordances out of the box on top of the same Bot Service transport as PR-5a.

| Audience | Use channel | Why |
|---|---|---|
| Teams-only deployments that want rich features (Adaptive Cards, streaming, Citations, feedback) | `TeamsChannel` (this PR) | First-class Teams SDK; per-message hooks. |
| Multi-client Activity Protocol (Direct Line, Web Chat, Slack-via-Bot-Service, …) | `ActivityProtocolChannel` (PR-5a) | Lower-level; no Teams-specific affordances. |

### Description

Adds `agent-framework-hosting-teams` (`python/packages/hosting-teams/`):

* `TeamsChannel` — built on `microsoft-teams-apps`. Default mount `/teams/messages`. Surfaces:
  * Text in/out + streaming via `ctx.stream`.
  * Outbound `transform_outbound` hook to emit Adaptive Cards.
  * Inbound `on_message_submit_feedback` callback for Teams feedback events.
  * Citations entity passthrough.
* `HttpServerAdapter` — captures the SDK's route registration synchronously inside `Channel.contribute()` so the host can mount a stable Starlette route (instead of letting the SDK spin up its own server).
* Tests use in-process fakes for `ActivityContext` + `HttpStream` (no live Bot Service required).

Note: this still uses **Azure Bot Service** as the transport (Teams does not currently expose a direct webhook outside Bot Service). The win over PR-5a is the developer-facing surface, not the hop count.

### Stack

PR-5b of 10. **Depends on #PR-2** (`feat/hosting-core`). Independent of PR-5a — sample apps in PR-9 use `ActivityProtocolChannel` from PR-5a; a follow-up sample for the Teams-SDK channel will be added separately.

### Contribution Checklist


python/packages/hosting-teams/agent_framework_hosting_teams/_channel.py
```diff
@@ -0,0 +1,673 @@
...
+
+        if self._run_hook is not None:
+            request = await apply_run_hook(
+                self._run_hook,
+                request,
+                target=self._ctx.target,
+                protocol_request=activity,
+            )
+
+        if self._streaming:
```
moonbox3, line 561: Should the streaming branch also go through `_send_outbound`? Right now `outbound_transform`, citations, and Adaptive Cards are silently ignored whenever `streaming=True` - the user gets only concatenated text via `HttpStream.emit`, no card, no citations entity. Constructing `TeamsChannel(streaming=True, outbound_transform=...)` typechecks and runs but the transform never fires. Either reject the combo at `__init__` or run the transform on the final `HostedRunResult` after `stream.close()`?

python/packages/hosting-teams/agent_framework_hosting_teams/_channel.py
```diff
@@ -0,0 +1,673 @@
...
+        Teams renders ``HttpStream.emit`` calls as live mid-message edits
+        (the typing-then-text streaming UI). We close the stream when
+        the host stream completes and call ``deliver_response`` with the
+        accumulated final result so cross-channel delivery still runs.
+        """
+        if self._ctx is None:  # pragma: no cover - contribute() ran first
+            raise RuntimeError("TeamsChannel was not contributed to a host.")
+        stream: ResponseStream[AgentResponseUpdate, AgentResponse] = self._ctx.run_stream(request)
+        accumulated: list[str] = []
+        async for update in stream:
```
moonbox3: What happens if the host stream raises mid-iteration (transient model error, tool failure)? No `try/finally`, so `stream.close()` is skipped and the Teams message is left in the live "typing" state forever from the user's POV, and `deliver_response` never runs so cross-channel sinks miss the turn entirely. Is it worth wrapping in `try/finally` that closes the stream and delivers the partial result (or an error result) so the failure is visible?

python/packages/hosting-teams/agent_framework_hosting_teams/_channel.py
```diff
@@ -0,0 +1,673 @@
...
+        """Register the messaging webhook and lifecycle hooks.
+
+        Calls :meth:`microsoft_teams.apps.HttpServer.initialize`
+        synchronously so the SDK populates the capturing adapter before
+        we hand the host a :class:`ChannelContribution`. Plugin
+        ``on_init`` callbacks (async) run from :meth:`_on_startup`.
+        """
+        self._ctx = context
+        # Synchronously register the route on our capturing adapter.
+        self._app.server.initialize(
```
moonbox3, line 446: The docstring says `App.initialize()` is idempotent because the synchronous server-init step gets short-circuited on the second call, but that short-circuit is gated on `App._initialized`, which we never set after the partial init in `contribute()`. So at startup `_initialized` is still false, `app.initialize()` runs in full, and `HttpServer.initialize()` plus `register_route` execute a second time. 

Should we set `self._app._initialized = True` after the partial init, or invoke only the missing async plugin `on_init` step in `_on_startup`?

python/packages/hosting-teams/agent_framework_hosting_teams/_channel.py
```diff
@@ -0,0 +1,673 @@
...
+            skip_auth=self._app.options.skip_auth or False,
+            cloud=self._app.cloud,
+        )
+        # Wire HttpServer.on_request to App's activity dispatcher (this
+        # is normally done inside App.initialize, but we did the
+        # synchronous initialization piecewise to obtain routes early).
+        self._app.server.on_request = self._app._process_activity_event  # type: ignore[attr-defined]  # pyright: ignore[reportPrivateUsage]
+
+        routes = [
+            _route_for(handler, mount_path, method) for (method, mount_path), handler in self._adapter.handlers.items()
```
moonbox3, line 467: If `_adapter.handlers` is empty here (user passed an `App` whose `server._adapter` was already populated before the swap at line 364, or a future SDK version moves `register_route` into the async `App.initialize` path), we return zero routes and the host happily mounts nothing. 

Bot Service POSTs to `/teams/messages` then 404 with no startup signal. Could we assert non-empty here, or at least `logger.error` so the failure mode is louder than it is currently?

python/packages/hosting-teams/agent_framework_hosting_teams/_channel.py
```diff
@@ -0,0 +1,673 @@
...
+        self.path = path
+        self._run_hook = run_hook
+        self._outbound_transform = outbound_transform
+        self._feedback_handler = feedback_handler
+        self._stream_transform_hook = stream_transform_hook
+        self._streaming = streaming
+        self._ctx: ChannelContext | None = None
+
+        self._adapter = _StarletteCaptureAdapter()
+        if app is not None:
```
moonbox3, line 385: Should we reject an already-initialized `app` here? If a caller built an `App` and ran `await app.initialize()` themselves before passing it in, routes are already registered on the previous adapter, our capturing adapter sees nothing, and the channel silently mounts zero routes (see other comment around `_route_for(...)`). I think an `if self._app._initialized: raise ValueError(...)` makes the misuse loud at construction time.

python/packages/hosting-teams/agent_framework_hosting_teams/_channel.py
```diff
@@ -0,0 +1,673 @@
...
+    for Bot Framework) result in a 400.
+    """
+
+    async def _starlette_handler(request: Request) -> Response:
+        raw_body = await request.body()
+        body: dict[str, Any] = {}
+        if raw_body:
+            try:
+                parsed = await request.json()
+            except Exception:
```
moonbox3, line 270: `except Exception` here swallows `ClientDisconnect`, encoding errors, Starlette body-reader bugs, anything - and we return 400 with no log. Bot Service then retries and the operator has no signal beyond "Teams ignores some messages intermittently." Looks to be a common theme across the several PRs, I've looked at: is it worth narrowing to `json.JSONDecodeError` and `logger.exception` for the rest with content-type and body length?

python/packages/hosting-teams/agent_framework_hosting_teams/_channel.py
```diff
@@ -0,0 +1,673 @@
...
+    async def _send_outbound(
+        self,
+        request: ChannelRequest,
+        activity_context: ActivityContext[MessageActivity],
+        result: HostedRunResult,
+    ) -> None:
+        """Resolve outbound content and POST to Teams via the SDK."""
+        payload = await self._resolve_outbound(request, activity_context, result)
+
+        if payload.card is not None:
```
moonbox3, line 672: The `text=None, card=None` branch returns silently with no log. Easy mistake from a transform ("no card applies, fall through to text" then forgetting to set text), and the symptom is a hung Teams conversation with nothing in logs to explain why. How about at a minimum we have a `logger.warning` with conversation/request id would give operators something to grep?

python/packages/hosting-teams/agent_framework_hosting_teams/_channel.py
```diff
@@ -0,0 +1,673 @@
...
+        rating = str(getattr(action_value, "reaction", "")) if action_value is not None else ""
+        feedback = getattr(action_value, "feedback", None) if action_value is not None else None
+        ctx = TeamsFeedbackContext(
+            activity_context=activity_context,
+            rating=rating,
+            feedback=feedback,
+            reply_to_id=activity.reply_to_id,
+            identity=identity,
+        )
+        outcome = self._feedback_handler(ctx)
```
moonbox3: User feedback handlers raising will propagate through the SDK's invoke dispatcher and 500 the invoke. Teams treats 5xx on feedback as a permanent failure for that activity, the user sees an error toast and the rating is lost. Feedback is a side channel, so wrapping in `try/except` and `logger.exception` seems safer than failing the invoke.

python/packages/hosting-teams/agent_framework_hosting_teams/_channel.py
```diff
@@ -0,0 +1,673 @@
...
+
+    # -- inbound: feedback ------------------------------------------------- #
+
+    async def _on_feedback_activity(self, activity_context: ActivityContext[MessageSubmitActionInvokeActivity]) -> None:
+        """Translate a feedback invoke into a :class:`TeamsFeedbackContext`."""
+        if self._feedback_handler is None:  # pragma: no cover - guarded at registration
+            return
+        activity = activity_context.activity
+        identity = self._identity_from_activity(activity)
+        action_value = getattr(activity.value, "action_value", None)
```
moonbox3, line 740: If the SDK ever renames `action_value`/`reaction`, every feedback event silently arrives with `rating=""` and the handler runs against empty data - thumbs-up/down telemetry breaks across the deployment with zero signal. Is it worth a `logger.warning` when neither field resolves, or a validation that `rating` is one of the expected values?

python/packages/hosting-teams/agent_framework_hosting_teams/_channel.py
eavanvalkenburg: Good catch — the broad `except` was hiding two very different failure modes. Now:

- `ClientDisconnect` → return **HTTP 408** + log at INFO. Bot Service treats 408 as transient and we don't want a stack trace in the logs for routine disconnects.
- `json.JSONDecodeError` → return **HTTP 400** + log at WARNING (single-line, no stack — the wire was just bad).
- Anything else from `request.body()` / `request.json()` → still return 400 to keep Bot Service from retrying, but log with `logger.exception()` so the underlying body-reader bug / encoding surprise actually shows up in operator logs instead of being swallowed.

Fixed in e5f7ff6c3; covered by `TestStarletteHandlerErrors.test_client_disconnect_returns_408`.

python/packages/hosting-teams/agent_framework_hosting_teams/_channel.py
eavanvalkenburg: You're right — the docstring was wrong about what makes the second startup safe. The SDK's `App._initialized` is set inside `App.initialize()`, but our `contribute()` only runs the synchronous `HttpServer.initialize` step (so it can capture routes via the swapped adapter) and never flips `_initialized`. The previous gate `if not self._app._initialized:` was therefore inverted-from-truth: every startup hit it, which happened to be safe today purely because the SDK's `App.initialize()` is internally idempotent.

Replaced with a channel-owned `_started` flag that we set after the first successful `_on_startup`. Now the gate is honest: 'has *this channel* called `app.initialize()` already?' and we don't depend on private SDK state.

Fixed in e5f7ff6c3; covered by `test_on_startup_skips_initialize_when_already_started`.

python/packages/hosting-teams/agent_framework_hosting_teams/_channel.py
eavanvalkenburg: Agreed — silently mounting nothing is the worst possible failure mode. `contribute()` now raises `RuntimeError` if `self._adapter.handlers` is empty after `HttpServer.initialize` runs, with a message pointing at the two known causes (pre-populated adapter on the user-supplied App, or a future SDK that moves route registration into the async `App.initialize` path). Surface at boot beats a 404 storm in production.

Fixed in e5f7ff6c3; covered by `test_contribute_raises_when_no_routes_registered`.

python/packages/hosting-teams/agent_framework_hosting_teams/_channel.py
eavanvalkenburg: Now rejected at construction. If `getattr(app, '_initialized', False)` is True we raise `ValueError('TeamsChannel: the supplied `app` has already been initialized...')` — much better than silently capturing zero routes and serving a 404 farm.

Fixed in e5f7ff6c3; covered by `test_constructor_rejects_already_initialized_app`.

python/packages/hosting-teams/agent_framework_hosting_teams/_channel.py
eavanvalkenburg: Real symmetry gap — fixed. The streaming branch now resolves `outbound_transform` after `HttpStream.close()` and (when one is configured) sends the resulting payload via `activity_context.send`, exactly the same way as the non-streaming path. Net result: live deltas still emit via `HttpStream.emit` for the typing UX, **and** Adaptive Cards / Citations / rewrites apply at the end. Teams treats the follow-up `send` as the canonical message and replaces the streamed deltas.

Fixed in e5f7ff6c3; covered by `test_streaming_applies_outbound_transform_after_close`.

python/packages/hosting-teams/agent_framework_hosting_teams/_channel.py
eavanvalkenburg: Wrapped the iterate / close / deliver block in `try/finally`. Mid-stream exception now still:

1. Calls `HttpStream.close()` (best-effort — failure is logged but doesn't shadow the original exception).
2. Calls `ChannelHostContext.deliver_response` with the accumulated text, so the host's history matches what Teams actually rendered.
3. Re-raises the original exception so the SDK's error machinery still sees it.

Skipping `outbound_transform` on failure (the success-only follow-up payload would otherwise overwrite the partial text with potentially-empty cards).

Fixed in e5f7ff6c3; covered by `test_streaming_finalises_on_iteration_failure`.

python/packages/hosting-teams/agent_framework_hosting_teams/_channel.py
eavanvalkenburg: Now logs at `WARNING` with the channel name and conversation id, so the dropped reply is visible in operator logs instead of producing a silent dead Teams conversation. The most common cause is a misconfigured `outbound_transform` that returns `TeamsOutboundPayload()` with neither field set; the log message points the operator there.

Fixed in e5f7ff6c3; covered by `test_outbound_payload_with_no_text_or_card_is_logged`.

python/packages/hosting-teams/agent_framework_hosting_teams/_channel.py
eavanvalkenburg: Wrapped `feedback_handler` invocation in try/except + `logger.exception(...)` and never propagate. Rationale: a 5xx here renders as a 'couldn't submit feedback' toast in the Teams client, which is bad UX for an out-of-band signal the user has no agency over (they already submitted). The handler exception is logged loudly with the `reply_to_id` + `rating` so it can be triaged.

Fixed in e5f7ff6c3; covered by `test_feedback_handler_exception_is_swallowed_and_logged`.

python/packages/hosting-teams/agent_framework_hosting_teams/_channel.py
eavanvalkenburg: Made the field-extraction defensive and observable:

- If `activity.value.action_value is None` → log WARNING with `reply_to_id` (clear hint that the SDK shape changed) and pass an empty rating to the handler.
- If `action_value.reaction is None` → same warning + empty rating.

Handler still runs in both cases (so callers can react / telemetry / clean up) but the SDK-rename regression is no longer silent.

Fixed in e5f7ff6c3; covered by `test_feedback_handler_logs_warning_when_action_value_missing`.

eavanvalkenburg:
Closing this for now. The Teams channel currently depends on a fairly involved adapter that reaches into private internals of the `microsoft-teams-apps` SDK (`server._adapter`, `server.initialize`, `_process_activity_event`, `_initialized`) to capture route registration synchronously while keeping server lifecycle owned by the host. That's too fragile to ship publicly.

I've filed an upstream feature request asking the SDK to support hosting the Teams endpoint inside a caller-owned ASGI app with synchronous route registration (host owns the lifecycle): microsoft/teams.py#447.

Once that lands, we can revisit a Teams channel built on the supported seam instead of the private-internals workaround. The branch `feat/hosting-channel-teams` is preserved for reference.

Tracking issue: #6265

PR #5643, opened 2026-05-05
Python: add agent-framework-hosting-telegram channel

eavanvalkenburg:
### Motivation and Context

Implements the **Telegram channel** described in [SPEC-002](https://github.com/microsoft/agent-framework/blob/feature/python-hosting/docs/decisions/0026-hosting-channels.md) §7 (merged via #5549). Demonstrates a non-Microsoft channel built on the same `Channel` protocol — a useful reference for community channels.

### Description

Adds `agent-framework-hosting-telegram` (`python/packages/hosting-telegram/`):

* `TelegramChannel` — mounts a webhook endpoint, decodes Telegram `Update` payloads into `ChannelRequest`, supports text replies + simple multicast (`response_targets`).
* Identity derivation from Telegram `chat_id` (the basis for `local_telegram` and the cross-channel `local_identity_link` sample in PR-9).
* Tests covering channel wiring + webhook handling.

### Stack

PR-6 of 10. **Depends on #PR-2** (`feat/hosting-core`).

### Contribution Checklist

### DIFF
--- a/python/packages/hosting-telegram/agent_framework_hosting_telegram/_channel.py
+++ b/python/packages/hosting-telegram/agent_framework_hosting_telegram/_channel.py
@@ -1,0 +1,714 @@
+# Copyright (c) Microsoft. All rights reserved.
+
+"""Built-in channel: Telegram (polling + webhook transports).
+
+Inspired by PR #5393's Telegram sample. Two transports are supported:
+
+- ``polling`` (default when no ``webhook_url`` is set): the channel runs a
+  background ``getUpdates`` long-poll loop. No public URL required —
+  perfect for local development. This is what ``python-telegram-bot``
+  uses by default.
+- ``webhook``: when ``webhook_url`` is set, the channel registers it via
+  ``setWebhook`` on startup and receives updates over HTTPS POSTs to the
+  mounted ``/webhook`` route. This is the production-recommended mode.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import contextlib
+import time
+from collections.abc import Awaitable, Callable, Mapping, Sequence
+from typing import Any, Literal
+
+import httpx
+from agent_framework import (
+    AgentResponse,
+    AgentResponseUpdate,
+    Content,
+    Message,
+    ResponseStream,
+)
+from agent_framework_hosting import (
+    ChannelCommand,
+    ChannelCommandContext,
+    ChannelContext,
+    ChannelContribution,
+    ChannelIdentity,
+    ChannelRequest,
+    ChannelRunHook,
+    ChannelSession,
+    ChannelStreamTransformHook,
+    HostedRunResult,
+    apply_run_hook,
+    logger,
+)
+from starlette.requests import Request
+from starlette.responses import JSONResponse, Response
+from starlette.routing import BaseRoute, Route
+
+# Telegram update parsing ------------------------------------------------------
+#
+# A Telegram message can carry text, a caption, and one of several media kinds
+# (photo, document, voice, audio, video). For media we resolve the file_id
+# into a public bot-file URL via ``getFile`` and emit a ``Content.from_uri``;
+# the agent then receives a multi-content Message with text + media side by
+# side, the same as it would over the Responses API.
+
+_TELEGRAM_MEDIA_DEFAULT_MIMETYPE = {
+    "photo": "image/jpeg",
+    "document": "application/octet-stream",
+    "voice": "audio/ogg",
+    "audio": "audio/mpeg",
+    "video": "video/mp4",
+}
+
+# Telegram's hard limit on a single message body. Past this, sendMessage /
+# editMessageText return 400. We truncate interim and final edits at this
+# boundary; if the agent emits more, callers can split into a follow-up
+# sendMessage in their run hook.
+_TELEGRAM_MAX_TEXT_LEN = 4096
+
+
+def telegram_isolation_key(chat_id: Any) -> str:
+    """Build the namespaced isolation key the Telegram channel writes under.
+
+    Exposed at module scope so other channels' ``run_hook`` callbacks can opt
+    into the same per-chat session (e.g. a Responses caller resuming a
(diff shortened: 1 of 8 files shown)


python/packages/hosting-telegram/agent_framework_hosting_telegram/_channel.py
```diff
@@ -0,0 +1,714 @@
...
+        traffic.
+        """
+        if self._secret_token is not None:
+            received = request.headers.get("x-telegram-bot-api-secret-token")
+            if received != self._secret_token:
+                logger.warning("Telegram webhook secret token mismatch — rejecting update")
+                return JSONResponse({"ok": False, "error": "invalid secret"}, status_code=401)
+
+        update = await request.json()
+        await self._safe_process_update(update)
```
moonbox3: Should we ack before running the agent? Based on my understanding, Telegram redelivers the update if the webhook doesn't 200 within ~60s, so any streamed reply that runs longer than that triggers a retry and the user sees duplicate replies (and `_dispatch` runs twice on the same `update_id`). Polling already spawns `_safe_process_update` as a task at line 298. Could the webhook do the same?

```suggestion
        update = await request.json()
        task = asyncio.create_task(self._safe_process_update(update))
        self._update_tasks.add(task)
        task.add_done_callback(self._update_tasks.discard)
        return JSONResponse({"ok": True})
```

python/packages/hosting-telegram/agent_framework_hosting_telegram/_channel.py
```diff
@@ -0,0 +1,714 @@
...
+            self._poll_task = asyncio.create_task(self._poll_loop(), name="telegram-poll")
+            logger.info("Telegram polling started (long-poll timeout=%ss)", self._polling_timeout)
+
+    async def _on_shutdown(self) -> None:
+        """Stop the polling task, drop the webhook registration, close the HTTP client.
+
+        Webhook teardown is best-effort — failures (e.g. revoked token at
+        shutdown) are logged but never raised so app shutdown can complete.
+        """
+        if self._poll_task is not None:
```
moonbox3, line 277: Are we missing the in-flight per-update tasks here? `_update_tasks` collects workers spawned from both the poll loop and (if the suggestion on `_handle` lands) the webhook, but shutdown only cancels the poll task and then closes `_http` out from under whatever is still running. A mid-flight agent run hits a closed client mid-`sendMessage` and the reply is lost. Should we drain or cancel `_update_tasks` before `aclose`?

python/packages/hosting-telegram/agent_framework_hosting_telegram/_channel.py
```diff
@@ -0,0 +1,714 @@
...
+                    logger.warning("Telegram getUpdates returned error: %s", payload)
+                    await asyncio.sleep(1.0)
+                    continue
+                for update in payload.get("result", []) or []:
+                    update_id = update.get("update_id")
+                    if isinstance(update_id, int):
+                        offset = update_id + 1
+                    # Each update is processed in its own task so a slow agent
+                    # call doesn't stall the next poll. Order within a chat is
+                    # still preserved by Telegram-side queueing.
```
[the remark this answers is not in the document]
eavanvalkenburg: Right — the comment in the code was confidently wrong. Telegram guarantees ordering up to `getUpdates`/the webhook delivery, but once we fanned each update out into its own task we lost that guarantee for adjacent updates in the same chat (e.g. `/start` then 'what's the weather?' could race depending on agent latency).

Replaced the per-update task fan-out with **per-chat serial workers**: each `chat_id` gets its own `asyncio.Queue` and a dedicated worker task. Updates for the same chat now process strictly in arrival order; updates for different chats still run in parallel. Updates with no resolvable chat (malformed payloads, unknown update types) fall back to a one-shot dispatcher task that's still tracked for drain on shutdown.

Fixed in 45287af14; covered by `TestPerChatOrdering.test_updates_for_same_chat_run_serially` and `...test_updates_for_different_chats_run_in_parallel`.

python/packages/hosting-telegram/agent_framework_hosting_telegram/_channel.py
eavanvalkenburg: Yes — Telegram redelivers any update the webhook doesn't 200 within ~60s, so a streamed reply that runs longer than that would have caused a retry storm + duplicate replies.

The webhook handler now:
1. Validates the secret token.
2. Parses the JSON (returns 400 on bad payloads).
3. Enqueues onto the per-chat serial worker via `_enqueue_update`.
4. Returns `200 {ok: true}` immediately.

The agent invocation runs in the worker task tracked in `_update_tasks`, so we keep ordering with polling-mode (both transports go through the same per-chat worker) and the worker is drained on shutdown.

Fixed in 45287af14; covered by `TestWebhookAckBeforeRun.test_webhook_returns_200_before_agent_completes` (asserts the response returns 200 while the agent is still in flight).

python/packages/hosting-telegram/agent_framework_hosting_telegram/_channel.py
eavanvalkenburg: Caught — `_update_tasks` was the right idea but the previous shutdown only cancelled the poll task itself, not the per-update tasks it had spawned, and the webhook path didn't track its tasks at all. With the new ack-before-run flow this would have been even worse: every webhook delivery would leak its agent invocation past app shutdown.

`_on_shutdown` is now an explicit drain in order:
1. Cancel + await the poll task (no new updates admitted).
2. Cancel + await each per-chat worker (current agent calls stop cleanly).
3. Cancel + await leftover dispatchers in `_update_tasks`.
4. Best-effort `deleteWebhook` then close the HTTP client.

Both polling-spawned and webhook-spawned tasks register into `_update_tasks` so the drain catches everything regardless of transport.

Fixed in 45287af14; covered by `TestShutdownDrainsWorkers.test_shutdown_cancels_in_flight_chat_workers`.

PR #5644, opened 2026-05-05
Python: add agent-framework-hosting-entra identity-link helpers

eavanvalkenburg:
### Motivation and Context

Implements the **identity-link helpers** from [SPEC-002](https://github.com/microsoft/agent-framework/blob/feature/python-hosting/docs/decisions/0026-hosting-channels.md) §"Linker / Cross-channel continuity" (merged via #5549). Provides the Entra (OAuth) sidecar that backs the cross-channel `local_identity_link` sample.

### Description

Adds `agent-framework-hosting-entra` (`python/packages/hosting-entra/`):

* OAuth linker primitive — mounts a small set of routes that drive the user through Entra sign-in and write a stable identity record into the host's `LinkStore`.
* Pluggable `LinkStore` interface so apps can persist the link wherever they want.
* Tests for the linker flow + state-store contract.

### Stack

PR-7 of 10. **Depends on #PR-2** (`feat/hosting-core`). The cross-channel sample that exercises this is in PR-9 (intentionally — keeps this PR small and free of transitive deps on PR-3 + PR-6).

### Contribution Checklist

### DIFF
--- a/python/packages/hosting-entra/agent_framework_hosting_entra/_channel.py
+++ b/python/packages/hosting-entra/agent_framework_hosting_entra/_channel.py
@@ -1,0 +1,342 @@
+# Copyright (c) Microsoft. All rights reserved.
+
+"""Microsoft Entra (Azure AD) identity-linking sidecar channel.
+
+Implements the OAuth 2.0 Authorization Code flow against Entra so users on
+non-Entra channels (Telegram, Responses callers without a verified token,
+etc.) can bind their per-channel id to a stable ``entra:<oid>`` isolation
+key. Once the link is established, channel run-hooks can call
+:meth:`EntraIdentityStore.lookup` and rewrite the request to use the Entra
+key instead of the channel-native id.
+
+Two credential modes are supported:
+
+* ``client_secret`` — confidential-client secret.
+* ``certificate_path`` — PEM bundle (private key + cert) for tenants that
+  disallow secrets. The Teams channel uses the same PEM layout; see
+  :mod:`agent_framework_hosting_teams` for the openssl recipe.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import json
+import secrets
+import time
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any
+
+import httpx
+import msal
+from agent_framework_hosting import (
+    ChannelContext,
+    ChannelContribution,
+    logger,
+)
+from cryptography import x509
+from cryptography.hazmat.primitives import hashes, serialization
+from starlette.requests import Request
+from starlette.responses import HTMLResponse, RedirectResponse, Response
+from starlette.routing import Route
+
+
+def entra_isolation_key(oid: str) -> str:
+    """Canonical isolation key for a user identified by Entra object id."""
+    return f"entra:{oid}"
+
+
+class EntraIdentityStore:
+    """Tiny JSON-backed mapping ``<channel>:<channel_id> → entra:<oid>``.
+
+    Production deployments should swap this for a real KV store. Single-file
+    JSON is fine for samples because writes are infrequent (only during the
+    OAuth callback) and we serialize them under an asyncio lock.
+    """
+
+    def __init__(self, path: Path) -> None:
+        """Open an identity store backed by ``path``.
+
+        Loads any existing JSON document; an unreadable or corrupt file is
+        logged and replaced with an empty in-memory map so callers always
+        get a usable store.
+        """
+        self._path = path
+        self._lock = asyncio.Lock()
+        self._data: dict[str, str] = {}
+        if path.exists():
+            try:
+                self._data = json.loads(path.read_text())
+            except Exception:
+                logger.exception("identity store load failed; starting empty")
+
+    def lookup(self, channel_key: str) -> str | None:
+        """Return the linked ``entra:<oid>`` key for a per-channel id, or ``None``."""
+        return self._data.get(channel_key)
+
+    async def link(self, channel_key: str, oid: str) -> None:
+        """Bind ``channel_key`` (e.g. ``telegram:123``) to the Entra ``oid`` and persist.
+
+        Overwrites any existing mapping for ``channel_key`` and rewrites the
+        backing JSON file under the lock so concurrent callers cannot race.
+        """
+        async with self._lock:
+            self._data[channel_key] = entra_isolation_key(oid)
+            self._path.write_text(json.dumps(self._data, indent=2, sort_keys=True))
+
+    async def unlink(self, channel_key: str) -> None:
+        """Remove the mapping for ``channel_key``; no-op if absent.
+
+        The file is only rewritten when an entry actually existed so we
+        don't churn disk on idempotent unlink calls.
+        """
+        async with self._lock:
+            if self._data.pop(channel_key, None) is not None:
+                self._path.write_text(json.dumps(self._data, indent=2, sort_keys=True))
+
+
+@dataclass
+class _PendingAuth:
+    """In-memory record of an authorize redirect waiting for its OAuth callback."""
+
+    channel: str
+    channel_id: str
+    expires_at: float
+    return_to: str | None = None
+
+
+def _link_html(body: str, *, status: int = 200) -> HTMLResponse:
+    """Wrap ``body`` in a minimal HTML shell suitable for browser link UIs."""
+    return HTMLResponse(
+        f"<!doctype html><html><body style='font-family:system-ui;padding:2rem;max-width:40rem'>{body}</body></html>",
+        status_code=status,
+    )
+
+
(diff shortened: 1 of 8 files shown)


python/packages/hosting-entra/agent_framework_hosting_entra/_channel.py
```diff
@@ -0,0 +1,342 @@
...
+                self._pending.pop(key, None)
+
+    async def _handle_start(self, request: Request) -> Response:
+        """``GET /start?channel=&id=&return_to=`` — redirect the user to Entra to sign in.
+
+        The caller (typically a channel command like ``/link`` in Telegram or
+        a Teams adaptive-card button) hands the user this URL; we mint the
+        authorize URL and 302 to it.
+        """
+        channel = request.query_params.get("channel")
```
moonbox3, line 389: What stops me, on the open internet, from hitting `/auth/start?channel=telegram&id=<victim_chat_id>`, signing in as myself, and binding the victim's Telegram id to my Entra oid? `/start` takes channel + id as unauthenticated query params with no proof the caller owns that channel id, so the resulting link maps the victim's per-channel key to the attacker's `entra:<oid>`. From then on any non-Entra channel that calls `store.lookup("telegram:<victim>")` runs the attacker's isolation key for the victim's traffic.

Is the design assumption that callers will gate `/start` themselves (only the Telegram channel mints these URLs after verifying a webhook signature)? If so, should the README and the docstring on `_handle_start` call that out as a hard requirement, or should the channel itself sign the `(channel, id)` pair so a stray `/start` GET cannot mint a binding?

python/packages/hosting-entra/agent_framework_hosting_entra/_channel.py
```diff
@@ -0,0 +1,342 @@
...
+
+        Exchanges the authorization code for a token, reads the user's
+        ``id``/``userPrincipalName`` from Microsoft Graph, then stores the
+        ``channel:channel_id -> entra:<oid>`` mapping in the identity store.
+        Renders a small HTML page so a browser-based flow has something to
+        show; if ``return_to`` was supplied it appears as a deep link.
+        """
+        if self._http is None:  # pragma: no cover - guarded by lifecycle
+            raise RuntimeError("entra identity channel not started")
+        if error := request.query_params.get("error"):
```
moonbox3, line 446: Reflected XSS on the OAuth callback. `error` and `error_description` come straight from the URL into the HTML body unescaped, and `_link_html` is `text/html`, so `?error=<script>...` renders. Same-origin as the auth host means cookies/storage for whatever else lives there.

```suggestion
        if error := request.query_params.get("error"):
            description = request.query_params.get("error_description", "")
            return _link_html(f"Sign-in failed: {html.escape(error)}<br>{html.escape(description)}", status=400)
```

(plus `import html` at the top). Same fix needed at line 319 for `result.get("error_description")` and at line 340 for `channel_key` / `upn`.

python/packages/hosting-entra/agent_framework_hosting_entra/_channel.py
```diff
@@ -0,0 +1,342 @@
...
+        oid = profile.get("id")
+        upn = profile.get("userPrincipalName") or profile.get("displayName") or oid
+        if not oid:
+            return _link_html("Profile response missing 'id'.", status=502)
+
+        channel_key = f"{pending.channel}:{pending.channel_id}"
+        await self._store.link(channel_key, oid)
+        logger.info("Linked %s → entra:%s (%s)", channel_key, oid, upn)
+
+        if pending.return_to:
```
moonbox3, line 488: Should we constrain this to a relative path, or check it against an allowlist of host prefixes derived from `public_base_url`?

python/packages/hosting-entra/agent_framework_hosting_entra/_channel.py
```diff
@@ -0,0 +1,342 @@
...
+            return RedirectResponse(pending.return_to, status_code=302)
+        return _link_html(
+            f"<h2>Linked</h2><p>{channel_key} is now bound to <b>{upn}</b>.</p>"
```
moonbox3: `channel_key` is `f"{pending.channel}:{pending.channel_id}"` where both halves come from the unauthenticated `/start` query string, and `upn` comes from a Graph response. Both go into HTML unescaped. Combined with previous comments (anyone can mint a pending entry), an attacker can stash a script payload in `channel` or `id` and serve it from the auth host. Shall we wrap with `html.escape` per the suggestion from the "Reflected XSS on the OAuth callback?"

python/packages/hosting-entra/agent_framework_hosting_entra/_channel.py
eavanvalkenburg: Real IDOR — thanks for digging through this. The previous design leaned entirely on 'callers will gate /start themselves', and neither the README nor the docstring spelled that out as a **hard requirement**. In practice the channel was one stray GET away from poisoning the identity store.

Closed defense-in-depth in the channel itself rather than punting to operators:

1. New `link_token_secret: str | None` constructor parameter. When set, `/auth/start` requires `exp` + `sig` (HMAC-SHA256 over `channel|channel_id|expires_at`) before issuing the redirect — constant-time verify, default TTL 10 min.
2. New `mint_start_url(channel, channel_id, return_to=None)` helper. Channels that hand out start URLs (a Telegram `/link` command after verifying the inbound webhook signature, a Teams card sign-in button after the bot validates the activity JWT) call this helper. The token proves *that channel* authorised the binding for *that channel_id*.
3. Unsigned mode is opt-in. When `link_token_secret` is None we still serve `/auth/start` (so dev-mode samples work without scrambling for a secret) but `_on_startup` logs a loud `WARNING` explaining the IDOR exposure, and **every accepted unsigned request** also logs WARNING with `(channel, id)`. There is no way for the bypass to ship to production silently.

Fixed in 487a7451a; covered by `TestSignedLinkToken` (`...test_start_rejects_unsigned_request_when_secret_set`, `...test_start_rejects_forged_signature`, `...test_start_accepts_valid_signed_url`, `...test_start_rejects_expired_signed_url`, `...test_mint_start_url_requires_secret`, `...test_unsigned_mode_logs_warning_at_startup`).

python/packages/hosting-entra/agent_framework_hosting_entra/_channel.py
eavanvalkenburg: Yes — reflected XSS on the auth host's origin, with the IDOR above as a delivery vector. Fixed exactly as suggested.

All values that flow into HTML output are now passed through `html.escape`:

- `error` and `error_description` query params (the path you flagged) — covered by `TestXssEscaping.test_callback_error_param_is_escaped` (`<script>alert(1)</script>` → `&lt;script&gt;...`).
- `result.get('error_description')` / `result.get('error')` on the token-exchange-failed branch.
- `channel_key` and `upn` on the success branch — covered by `TestXssEscaping.test_callback_success_escapes_channel_key_and_upn` (smuggling `<svg/onload=alert(1)>` into `channel` and `<script>` into `userPrincipalName`).

Fixed in 487a7451a (`import html` added).

python/packages/hosting-entra/agent_framework_hosting_entra/_channel.py
eavanvalkenburg: Open redirect — fixed with an allowlist in the channel itself.

New `_validate_return_to` helper accepts only:

- Relative paths starting with `/` and **not** `//` (`//evil.example/` parses as `netloc=evil.example`, so it's explicitly rejected).
- Absolute URLs whose host equals the configured `public_base_url` host (case-insensitive).

Validation runs at two points:

1. At `/auth/start` time, before the pending entry is created. Bad `return_to` → 400, no pending state, no Entra round-trip.
2. Defensively at `/auth/callback` time, before the final 302. Catches any future code path that might bypass the /start gate.

`mint_start_url` also calls `_validate_return_to` so misconfigured channel code fails at mint time instead of producing a URL the user will hit and bounce off of.

Fixed in 487a7451a; covered by `TestReturnToOpenRedirect` (external rejected, relative accepted, same-origin accepted, protocol-relative rejected).

python/packages/hosting-entra/agent_framework_hosting_entra/_channel.py
eavanvalkenburg: Yes — same fix. The IDOR comment above closes the 'anyone can mint a pending entry' half of the bug, and the XSS comment above closes the 'unescaped HTML output' half. Specifically for this site:

- `channel_key = f"{pending.channel}:{pending.channel_id}"` → `html.escape(channel_key)` before embedding in the success page.
- `upn` (from the Graph response) → `html.escape(str(upn))`.

Even with `link_token_secret` set, defence-in-depth applies because the Graph response itself isn't under our control.

Fixed in 487a7451a; covered by `TestXssEscaping.test_callback_success_escapes_channel_key_and_upn`.

PR #5645, opened 2026-05-05
Python: add hosting Channels sample apps

eavanvalkenburg:
### Motivation and Context

Adds the **runnable sample apps** for the Python hosting Channels work specified in [SPEC-002](https://github.com/microsoft/agent-framework/blob/feature/python-hosting/docs/decisions/0026-hosting-channels.md) (merged via #5549) and the channel packages in PR-1..PR-8.

### Description

Adds five sample apps under `python/samples/04-hosting/af-hosting/`:

| Sample | Channels exercised | Notes |
|---|---|---|
| `local_responses/` | Responses | Smallest hello-world + run-hook example |
| `local_responses_workflow/` | Responses | 3-agent workflow as host target with checkpoint storage |
| `local_telegram/` | Responses + Telegram | Single-bot variant + multicast variant |
| `local_identity_link/` | Responses + Invocations + Telegram + **Activity Protocol** + Entra link sidecar | Full surface — uses `ActivityProtocolChannel` from PR-5a, not the Teams-SDK channel from PR-5b |
| `foundry_hosted_agent/` | Responses + Invocations | `azd`-deployable; `Dockerfile` + `azure.yaml` for Foundry Hosted Agents |

Each sample's `pyproject.toml` pulls hosting packages directly from this repo's `feature/python-hosting` branch via `[tool.uv.sources]` git refs, e.g.:

```toml
[tool.uv.sources]
agent-framework-hosting = { git = "https://github.com/microsoft/agent-framework.git", branch = "feature/python-hosting", subdirectory = "python/packages/hosting" }
```

This pattern lets the samples be tried before any of the channel packages are published to PyPI. The READMEs note the migration path to PyPI deps once the packages are released. `uv.lock` files are intentionally not committed — generated on the user's first `uv sync`.

`foundry_hosted_agent/` previously bundled a `_vendor/` snapshot of the hosting packages so its container could build self-contained; that scaffolding is now removed in favor of the git-ref pattern (no `scripts/vendor-packages.sh`, no `_vendor/` `COPY` in the Dockerfile, no `hooks.prepackage` in `azure.yaml`).

### Stack

PR-9 of 10 — the **last** PR. Nothing depends on it. The samples won't `uv sync` against the `feature/python-hosting` branch until PR-1..PR-8 land on that branch first; merging this PR earlier is harmless but the samples won't run.

| # | Branch | Required for samples |
|---|---|---|
| PR-1 | `refactor/foundry-hosted-agent-history-provider` | `foundry_hosted_agent/` |
| PR-2 | `feat/hosting-core` | all |
| PR-3 | `feat/hosting-channel-responses` | all |
| PR-4 | `feat/hosting-channel-invocations` | `foundry_hosted_agent/`, `local_identity_link/` |
| PR-5a | `feat/hosting-channel-activity-protocol` | `local_identity_link/` |
| PR-6 | `feat/hosting-channel-telegram` | `local_telegram/`, `local_identity_link/` |
| PR-7 | `feat/hosting-entra-helpers` | `local_identity_link/` |
| PR-8 | `feat/hosting-channel-discord` | validates channel abstractions; no current sample dependency |

### Contribution Checklist

- [x] All unit tests pass, and I have added new tests where possible (samples — manual run validation)
- [ ] **Is this a breaking change?** No — new samples.


python/samples/04-hosting/af-hosting/foundry_hosted_agent/call_server.py
```diff
@@ -0,0 +1,114 @@
...
+    print(f"Agent: {result.text}")
+
+
+async def call_via_foundry_hosted_agent(prompt: str) -> None:
+    # Once foundry_hosted_agent's image is deployed as a Foundry Hosted Agent, FoundryAgent
+    # keyed on ``agent_name`` is the AF-native client. The agent's runtime is
+    # the very same Responses + Invocations contract — Foundry just hosts it.
+    async with AzureCliCredential() as credential:
+        agent = FoundryAgent(
+            project_endpoint="https://edvan-foundry-sw.services.ai.azure.com/api/projects/agents",
```
moonbox3: Do we want this hard-coded?

python/samples/04-hosting/af-hosting/foundry_hosted_agent/call_server.py
```diff
@@ -0,0 +1,114 @@
...
+            agent_name="agent-framework-hosting-sample",
+            credential=credential,
+            allow_preview=True,
+        )
+        session = agent.get_session(service_session_id="resp_7c86d98980df4f409045e71609c50ce4")
```
moonbox3: Same here?

python/samples/04-hosting/af-hosting/local_identity_link/app.py
```diff
@@ -0,0 +1,360 @@
...
+def make_telegram_hook(store: EntraIdentityStore) -> Any:
+    """If the Telegram chat id is linked, swap to ``entra:<oid>`` for history."""
+
+    def _hook(request: ChannelRequest, **_: object) -> ChannelRequest:
+        chat_id = request.attributes.get("chat_id")
+        if chat_id is not None:
+            linked = store.lookup(f"telegram:{chat_id}")
+            if linked is not None:
+                return _replace_session(request, linked)
+        # Bump reasoning effort regardless of identity.
```
moonbox3: Comment says "regardless of identity" but the linked branch returns early, so the reasoning bump only applies to unlinked chats. The headline feature of this sample is `/link`, and the moment a user runs it they silently lose the high-effort reasoning preset. Intentional, or should the bump apply on both branches?

python/samples/04-hosting/af-hosting/local_identity_link/app.py
```diff
@@ -0,0 +1,360 @@
...
+        options.pop("temperature", None)
+        options.pop("store", None)
+
+        body = protocol_request or {}
+        metadata = body.get("metadata") if isinstance(body.get("metadata"), dict) else {}
+
+        explicit_oid = body.get("entra_oid") or metadata.get("entra_oid")
+        safety_id = body.get("safety_identifier") or body.get("user") or "anonymous"
+
+        if explicit_oid:
```
moonbox3, line 213: Are we worried that a reader copies this verbatim? The sample treats client-supplied `entra_oid` (top-level body or metadata) as authoritative identity with zero verification, so any Responses caller can claim to be any user and read their history bucket. The README markets this hook as the production identity path, which makes the foot-gun worse. Could we either gate this behind a comment that screams "DEV ONLY, verify a token in production" or drop the `entra_oid` shortcut and lean on the linked-store path?

python/samples/04-hosting/af-hosting/local_identity_link/app.py
```diff
@@ -0,0 +1,360 @@
...
+    async def handle_link(ctx: ChannelCommandContext) -> None:
+        linker = linker_ref.get("linker")
+        if linker is None:
+            await ctx.reply(
+                "Identity linking is not configured on this host. "
+                "Set ENTRA_TENANT_ID, ENTRA_CLIENT_ID, and either "
+                "ENTRA_CLIENT_SECRET or ENTRA_CERT_PATH."
+            )
+            return
+        chat_id = ctx.request.attributes.get("chat_id")
```
moonbox3, line 269: What happens when `chat_id` is missing? `str(None)` formats `"telegram:None"` into a real authorize URL, the user opens it, OAuth completes, and the store now has a `telegram:None` entry that any later chat_id-less message collapses onto. Worth an early-return guard?

python/samples/04-hosting/af-hosting/local_identity_link/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+```bash
+export FOUNDRY_PROJECT_ENDPOINT=https://<your-project>.services.ai.azure.com
+export FOUNDRY_MODEL=gpt-4o
+export TELEGRAM_BOT_TOKEN=...
+# Entra app registration (confidential client):
+export ENTRA_TENANT_ID=...
+export ENTRA_CLIENT_ID=...
+export ENTRA_CLIENT_SECRET=...                 # or:
+# export ENTRA_CERTIFICATE_PATH=./teams-bot.pem
```
moonbox3, line 21: `build_host()` reads `ENTRA_CERT_PATH` (app.py:309), not `ENTRA_CERTIFICATE_PATH`. A reader who follows the README and exports the documented name will land in the `linker is None` branch with no error message, then be confused when `/link` reports the feature is disabled. Should the README match the env var the code actually reads?

python/samples/04-hosting/af-hosting/local_identity_link/app.py
```diff
@@ -0,0 +1,360 @@
...
+            client_secret=client_secret,
+            certificate_path=cert_path,
+            certificate_password=cert_password_env.encode() if cert_password_env else None,
+            public_base_url=public_base_url,
+        )
+
+    host_ref: dict[str, AgentFrameworkHost] = {}
+    linker_ref: dict[str, EntraIdentityLinkChannel | None] = {"linker": linker}
+
+    channels: Sequence[Channel] = [
```
moonbox3: Are the type-checks accurate here? `Sequence` is read-only, but line 349 calls `channels.append(linker)`. Works at runtime because the literal is a list, but a future refactor swapping it for a tuple silently disables the linker (which is the whole point of the sample). Annotate as `list[Channel]`?

```suggestion
    channels: list[Channel] = []
```

python/samples/04-hosting/af-hosting/foundry_hosted_agent/call_server.py
eavanvalkenburg: Good catch — `project_endpoint` is now read from `FOUNDRY_PROJECT_ENDPOINT`, with a clear `SystemExit` if it's unset (instead of a confusing 401/404 from the SDK pointed at my dev tenant). The agent name also moved to `FOUNDRY_HOSTED_AGENT_NAME` (defaulting to the sample's name so the common 'I deployed via the sample's azd template' path still just works).

Fixed in 015ed2de1.

python/samples/04-hosting/af-hosting/foundry_hosted_agent/call_server.py
eavanvalkenburg: Same treatment — the session id moved to the optional `FOUNDRY_HOSTED_SESSION_ID` env var. When unset the sample now exercises the new-conversation path (no `service_session_id` passed), which is the more useful default for a first-run smoke test. Setting the env var resumes a prior conversation, which is what my hard-coded value was doing — just no longer leaking the dev session id into the public sample.

Fixed in 015ed2de1.

python/samples/04-hosting/af-hosting/local_identity_link/app.py
eavanvalkenburg: You're right — the comment was lying. The intent of the hook is 'always bump reasoning effort, additionally swap session if linked', but the linked branch was returning early and silently dropping the bump for the very flow this sample exists to demonstrate (a user who just successfully ran `/link`).

Restructured to match the intent: assign `request = _replace_session(...)` instead of returning, then fall through to the options bump on every code path. Updated the docstring to spell out 'identity resolution and option mutation are separate concerns'.

Fixed in 015ed2de1.

python/samples/04-hosting/af-hosting/local_identity_link/app.py
eavanvalkenburg: Real foot-gun, agreed. I kept the shortcut (it's genuinely useful for smoke-testing the isolation key without dragging an Entra app registration into a 'hello world' run) but made it impossible to miss what's wrong with it:

- Top-of-function docstring now carries a multi-line `.. WARNING::` block explaining the bypass and listing the two viable production paths (drop the shortcut, or add a JWT validator that feeds a verified `oid` claim into `entra_isolation_key`).
- An inline `# WARNING (DEV ONLY)` comment sits directly above the line that reads `entra_oid` from the body.

If you'd rather drop the shortcut entirely I'm happy to do that as a follow-up — the docstring lays out exactly what the production-shaped replacement would look like.

Fixed in 015ed2de1.

python/samples/04-hosting/af-hosting/local_identity_link/app.py
eavanvalkenburg: Caught — without the guard the `f"telegram:{chat_id}"` format string happily produces `"telegram:None"`, the authorize URL is minted on that key, and a successful OAuth round-trip plants a poisoned `telegram:None` -> `entra:<oid>` binding that any later chat_id-less update would collapse onto.

Added an early-return on `chat_id is None` with a user-visible error reply. The accompanying comment explains the failure mode so the next person doesn't 'simplify' the guard away.

Fixed in 015ed2de1.

python/samples/04-hosting/af-hosting/local_identity_link/README.md
eavanvalkenburg: Good catch — picked the README naming (longer is more conventional and matches `ENTRA_CLIENT_SECRET`). Code now reads `ENTRA_CERTIFICATE_PATH` and `ENTRA_CERTIFICATE_PASSWORD`. The user-facing `/link` reply that lists the missing env vars was updated to match.

Fixed in 015ed2de1.

python/samples/04-hosting/af-hosting/local_identity_link/app.py
eavanvalkenburg: Right — `Sequence` doesn't expose `.append`, so the line that extends the list with the optional linker was relying on type-checker leniency. Switched the annotation to `list[Channel]` and dropped the now-unused `Sequence` import.

Fixed in 015ed2de1.

PR #5652, opened 2026-05-05
.NET: Bump MEAI to 10.5.1 and add Foundry per-call x-client header support

### LINKED ISSUE #5647: .NET: [Feature]: Update Microsoft.Extensions.AI to version 10.5.1
### Description

Among the other things, version 10.5.1 of the Microsoft.Extensions.AI packages stabilizes `CodeInterpreter`, `WebSearch`, and `ImageGeneration `tool content types. Full release notes are available at https://github.com/dotnet/extensions/releases/tag/v10.5.1

### Language/SDK

.NET

rogerbarreto:
## Summary

Bumps Microsoft.Extensions.AI from 10.5.0 to 10.5.1 and replaces the brittle `UserAgentResponsesClient` subclass with a clean per-call `x-client-*` header pipeline built on the new MEAI 10.5.1 `OpenAIRequestPolicies` hook.

The motivating scenario is the multi-tenant SaaS overlay from the Foundry Hosted Agents design: a SaaS backend needs to attest the end-user identity (`x-client-end-user-id`) and chat surface (`x-client-end-chat-id`) on a per-`RunAsync` basis, without rebuilding the agent or the chat client.

## Public surface

All in `Microsoft.Agents.AI.Foundry`, gated by `[Experimental(MAAI001)]`.

` csharp
// Per-run carrier on ChatOptions
chatOptions.WithClientHeader("x-client-end-user-id", "alice");
chatOptions.WithClientHeaders(new[] { 
    KeyValuePair.Create("x-client-end-user-id", "alice"),
    KeyValuePair.Create("x-client-end-chat-id", "chat-42"),
});

// Opt-in for customer-built agents
var agent = new ChatClientAgent(chatClient).AsBuilder().UseClientHeaders().Build();

// Foundry-built agents pre-wire automatically
var foundryAgent = new FoundryAgent(endpoint, credential, model, instructions);
` 

* Header names must start with `x-client-` (case-insensitive); other names throw `ArgumentException`.
* Bulk `WithClientHeaders` validates all entries before mutating any (all-or-nothing).
* If the `ChatOptions.AdditionalProperties` carrier slot is already occupied by a non-dictionary value, throws `InvalidOperationException`.

## Internals

* `ClientHeadersAgent` decorator snapshots the dict at scope-push time so concurrent runs sharing a `ChatOptions` reference do not cross-contaminate.
* `ClientHeadersScope` is an `AsyncLocal<IReadOnlyDictionary<string,string>?>` with LIFO `using` semantics.
* `ClientHeadersPolicy` singleton stamps headers with `Headers.Set` so per-call values overwrite same-name headers from earlier policies and double registration is value-stable.
* `OpenAIRequestPoliciesReflection` dedups against MEAI's private `_entries` field with graceful fallback. A CI-guardrail test asserts the field shape so future MEAI bumps fail loudly here rather than silently disabling dedup.

## Hosting cleanup

* Deleted `UserAgentResponsesClient` (~115 lines) and its dummy throwing pipeline.
* `HostedAgentUserAgentPolicy` is now registered via `OpenAIRequestPolicies` in `FoundryHostingExtensions.TryApplyUserAgent`.

## Tests

* 19 new unit tests in `ClientHeadersExtensionsTests.cs` covering validation, AsyncLocal isolation, snapshot semantics, end-to-end wire stamping, and shared-chat-client dedup.
* Updated `OpenTelemetryAgentTests` for MEAI 10.5.1 changes to `web_search` serialization and the reduced tool definition payload when sensitive data capture is disabled.

## Verification

* Full solution build: 0 errors, 0 warnings.
* CI-parity `dotnet format --verify-no-changes` (Docker `mcr.microsoft.com/dotnet/sdk:10.0`) clean on all 5 changed projects.
* 5245 tests passing across Foundry, Foundry.Hosting, core AI, Abstractions, AzureAI.Persistent, OpenAI, Anthropic.

## Notes

* `Microsoft.Extensions.Compliance.Abstractions` stays at 10.5.0 because no 10.5.1 release exists on nuget.org.
* The OpenAI bridge is reachable through Foundry today; if a non-Foundry consumer wants the same wire delivery, they reference Microsoft.Agents.AI.Foundry. Splitting the bridge into its own package can come later if there is demand.


dotnet/src/Microsoft.Agents.AI.Foundry.Hosting/ServiceCollectionExtensions.cs
```diff
@@ -207,84 +206,36 @@ internal static AIAgent ApplyOpenTelemetry(AIAgent agent)
...
-        }
-
-        var meaiType = s_meaiResponsesChatClientType;
-        if (meaiType is null)
-        {
-            return agent;
+            // The HostedAgentUserAgentPolicy is idempotent on the wire (it skips when the
+            // supplement is already present in the User-Agent header), so we can register it
+            // unconditionally without dedup. MEAI's lock-free CAS-append on _entries is safe.
+            policies.AddPolicy(HostedAgentUserAgentPolicy.Instance, PipelinePosition.PerCall);
```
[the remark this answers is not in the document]
rogerbarreto: right. unbounded growth on singleton agents. now dedup per OpenAIRequestPolicies instance via ConditionalWeakTable. AddPolicy fires at most once per chat client. fixed in a4c8f91.

dotnet/src/Microsoft.Agents.AI.Foundry/FoundryAgent.cs
```diff
@@ -210,10 +215,21 @@ private static ChatClientAgent CreateResponsesChatClientAgent(
...
+        if (chatClient.GetService<OpenAIRequestPolicies>() is { } policies)
+        {
+            OpenAIRequestPoliciesReflection.AddPolicyIfMissing(
+                policies,
+                ClientHeadersPolicy.Instance,
+                System.ClientModel.Primitives.PipelinePosition.PerCall);
+        }
+
+        var inner = new ChatClientAgent(chatClient, agentOptions, loggerFactory, services);
+        return new ClientHeadersAgent(inner);
```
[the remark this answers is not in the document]
rogerbarreto: good catch. AsAIAgent path now pre-wires too. extracted WireClientHeaders helper, called from internal (AIProjectClient, ChatClientAgent) ctor. fixed in a4c8f91.

dotnet/src/Microsoft.Agents.AI.Foundry/ClientHeadersAgent.cs
```diff
@@ -0,0 +1,105 @@
...
+        IEnumerable<ChatMessage> messages,
+        AgentSession? session = null,
+        AgentRunOptions? options = null,
+        [EnumeratorCancellation] CancellationToken cancellationToken = default)
+    {
+        var snapshot = TrySnapshot(options);
+        using var _ = snapshot is null ? default : ClientHeadersScope.Push(snapshot);
+
+        await foreach (var update in this.InnerAgent.RunStreamingAsync(messages, session, options, cancellationToken)
+            .WithCancellation(cancellationToken)
```
SergeyMenshykh: Is the WithCancellation necessary here taking into account that the ct is already passed to the RunStreamingAsync?

SergeyMenshykh (approved):

dotnet/src/Microsoft.Agents.AI.Foundry/ClientHeadersAgent.cs
rogerbarreto: right. ct already flows into InnerAgent.RunStreamingAsync. WithCancellation redundant. removed in 5c6e21e. checked rest of PR diff, no other occurrences.

SergeyMenshykh (approved):

dotnet/src/Microsoft.Agents.AI.Foundry/ClientHeadersScope.cs
```diff
@@ -0,0 +1,46 @@
...
+
+namespace Microsoft.Agents.AI.Foundry;
+
+/// <summary>
+/// AsyncLocal carrier that bridges per-call client-header values from the
+/// <see cref="ClientHeadersAgent"/> decorator down to the
+/// <see cref="ClientHeadersPolicy"/> running inside the SCM transport pipeline.
+/// </summary>
+/// <remarks>
+/// Stack-style usage with <c>using</c> ensures LIFO restoration of any prior value.
```
westey-m: Doesn't AsyncLocal do this anyway?

dotnet/src/Microsoft.Agents.AI.Foundry/Microsoft.Agents.AI.Foundry.csproj
```diff
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <IsReleased>true</IsReleased>
     <InjectSharedThrow>true</InjectSharedThrow>
-    <NoWarn>$(NoWarn);OPENAI001</NoWarn>
+    <NoWarn>$(NoWarn);OPENAI001;MEAI001</NoWarn>
```
westey-m: Make sure to surface any downstream experimental features (where possible/appropriate) on our features that use them too, so that users know they are indirectly consuming an experimental feature.

westey-m (approved):

dotnet/src/Microsoft.Agents.AI.Foundry/ClientHeadersScope.cs
rogerbarreto: AsyncLocal flows value forward into awaits but does not auto-rollback when the setting method returns. mutation persists. so explicit using/Dispose is needed to restore prior value. clarified docstring in 66f4a6c.

dotnet/src/Microsoft.Agents.AI.Foundry/Microsoft.Agents.AI.Foundry.csproj
rogerbarreto: good call. added AIOpenAIRequestPolicies = MEAIExperiments alias to DiagnosticIds.Experiments (matches AIResponseContinuations pattern). public ClientHeadersExtensions now marked with that alias instead of MAAI001 - consumers see MEAI001 warning so the downstream dependency is visible. removed MEAI001 from csproj NoWarn. fixed in 66f4a6c.

westey-m (approved):

dotnet/src/Microsoft.Agents.AI.Foundry/ClientHeadersScope.cs
rogerbarreto: you right. ran your demo. AsyncLocal mutations inside awaited async method auto-restore on return - runtime handles it. our explicit using/Dispose was redundant since both ClientHeadersAgent.RunCoreAsync/RunCoreStreamingAsync are async methods awaited by callers. simplified ClientHeadersScope to single Current { get; set; } and dropped the using pattern in 1019c18. snapshot deep-copy stays cause that defends against caller mutating dict mid-run, separate concern. updated test 10 with your style assertion confirming caller sees null after the async probe returns.

PR #5678, opened 2026-05-06
Python: Add ClassSkill for class-based skill definitions

SergeyMenshykh:
### Motivation and Context

Ports .NET's class-based skills feature (PRs #5027 and #5183) to Python, enabling skills to be defined as self-contained, reusable Python classes.

### Description

Adds `ClassSkill` — an abstract base class for defining skills as Python classes with decorator-based auto-discovery of resources and scripts.

**Usage:**

```python
class UnitConverterSkill(ClassSkill):
    def __init__(self) -> None:
        super().__init__(name="unit-converter", description="Convert units.")

    @property
    def instructions(self) -> str:
        return "Use this skill to convert between units."

    @ClassSkill.resource(name="conversion-table")
    def get_table(self) -> str:
        return "| miles | km | 1.60934 |"

    @ClassSkill.script(name="convert")
    def convert(self, value: float, factor: float) -> str:
        return json.dumps({"result": round(value * factor, 4)})
```

**What's included:**

- `ClassSkill(Skill, ABC)` with `@ClassSkill.resource` / `@ClassSkill.script` decorators
- Supports methods and `@property` descriptors as resources
- Lazy per-instance discovery, duplicate name detection, content caching
- Constructor injection (`self`) and runtime kwargs (`**kwargs`) for DI
- New `class_based_skill` sample and updated `mixed_skills` sample
- 58 new unit tests

### Contribution Checklist

### DIFF
--- a/python/packages/core/agent_framework/_skills.py
+++ b/python/packages/core/agent_framework/_skills.py
@@ -5,7 +5,7 @@
 Defines the core data model classes for the agent skills system:
 
 - **Skills:** :class:`Skill` (abstract base), :class:`InlineSkill` (code-defined),
-  and :class:`FileSkill` (filesystem-backed).
+  :class:`ClassSkill` (class-based), and :class:`FileSkill` (filesystem-backed).
 - **Resources:** :class:`SkillResource` (abstract base), :class:`InlineSkillResource`
   (static content or callable).
 - **Scripts:** :class:`SkillScript` (abstract base), :class:`InlineSkillScript`
@@ -27,6 +27,9 @@ Skills can come from different sources:
   Represented as :class:`FileSkill` instances.
 - **Code-defined** — created as :class:`InlineSkill` instances in Python code,
   with optional callable resources attached via the ``@skill.resource`` decorator.
+- **Class-based** — created by subclassing :class:`ClassSkill` to define
+  self-contained, reusable skill types with ``create_resource()`` and
+  ``create_script()`` factory methods.
 - **Custom sources** — any :class:`SkillsSource` implementation that provides
   skills from arbitrary origins (REST APIs, databases, etc.).
 
@@ -575,6 +578,65 @@ def _validate_skill_description(name: str, description: str) -> None:
         )
 
 
+def _build_skill_content(
+    name: str,
+    description: str,
+    instructions: str,
+    resources: Sequence[SkillResource] | None = None,
+    scripts: Sequence[SkillScript] | None = None,
+) -> str:
+    """Build XML-structured content for code-defined and class-based skills.
+
+    Produces an XML document containing name, description, instructions,
+    resources, and scripts elements.  Used by both :class:`InlineSkill`
+    and :class:`ClassSkill` to generate their ``content`` property.
+
+    Args:
+        name: The skill name.
+        description: The skill description.
+        instructions: The raw instructions text.
+        resources: Optional resources associated with the skill.
+        scripts: Optional scripts associated with the skill.
+
+    Returns:
+        An XML-structured content string.
+    """
+    result = (
+        f"<name>{xml_escape(name)}</name>\n"
+        f"<description>{xml_escape(description)}</description>\n"
+        "\n"
+        "<instructions>\n"
+        f"{instructions}\n"
+        "</instructions>"
+    )
+
+    if resources:
+        resource_lines = "\n".join(_create_resource_element(r) for r in resources)
+        result += f"\n\n<resources>\n{resource_lines}\n</resources>"
+
+    if scripts:
+        script_lines = "\n".join(_create_script_element(s) for s in scripts)
+        result += f"\n\n<scripts>\n{script_lines}\n</scripts>"
+
+    return result
+
+
+def _create_resource_element(resource: SkillResource) -> str:
+    """Create a self-closing ``<resource …/>`` XML element from a :class:`SkillResource`.
+
+    Args:
+        resource: The resource to create the element from.
+
+    Returns:
+        A single indented XML element string with ``name`` and optional
+        ``description`` attributes.
+    """
+    attrs = f'name="{xml_escape(resource.name, quote=True)}"'
+    if resource.description:
+        attrs += f' description="{xml_escape(resource.description, quote=True)}"'
+    return f"  <resource {attrs}/>"
+
+
 @experimental(feature_id=ExperimentalFeature.SKILLS)
 class InlineSkill(Skill):
     """A skill defined entirely in code with resources and scripts.
@@ -639,25 +701,10 @@ class InlineSkill(Skill):
         if self._cached_content is not None:
             return self._cached_content
 
-        result = (
-            f"<name>{xml_escape(self.name)}</name>\n"
-            f"<description>{xml_escape(self.description)}</description>\n"
-            "\n"
-            "<instructions>\n"
-            f"{self.instructions}\n"
-            "</instructions>"
+        self._cached_content = _build_skill_content(
+            self.name, self.description, self.instructions, self._resources, self._scripts
         )
-
-        if self._resources:
-            resource_lines = "\n".join(self._create_resource_element(r) for r in self._resources)
-            result += f"\n\n<resources>\n{resource_lines}\n</resources>"
-
-        if self._scripts:
-            script_lines = "\n".join(_create_script_element(s) for s in self._scripts)
-            result += f"\n\n<scripts>\n{script_lines}\n</scripts>"
-
-        self._cached_content = result
-        return result
+        return self._cached_content
 
     @property
     def resources(self) -> list[SkillResource]:
@@ -668,22 +715,6 @@ class InlineSkill(Skill):
     def scripts(self) -> list[SkillScript]:
         """Mutable list of :class:`SkillScript` instances."""
         return self._scripts
-
-    @staticmethod
-    def _create_resource_element(resource: SkillResource) -> str:
-        """Create a self-closing ``<resource …/>`` XML element from an :class:`SkillResource`.
-
-        Args:
-            resource: The resource to create the element from.
-
-        Returns:
-            A single indented XML element string with ``name`` and optional
-            ``description`` attributes.
-        """
-        attrs = f'name="{xml_escape(resource.name, quote=True)}"'
-        if resource.description:
-            attrs += f' description="{xml_escape(resource.description, quote=True)}"'
-        return f"  <resource {attrs}/>"
 
     def resource(
         self,
@@ -705,8 +736,7 @@ class InlineSkill(Skill):
 
         Keyword Args:
             name: Resource name override.  Defaults to ``func.__name__``.
-            description: Resource description override.  Defaults to the
-                function's docstring (via :func:`inspect.getdoc`).
+            description: Resource description override.  Defaults to ``None``.
 
         Returns:
             The original function unchanged, or a secondary decorator when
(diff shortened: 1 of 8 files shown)


moonbox3:
<a href="https://github.com/microsoft/agent-framework/blob/fe287b975e833cbcebb87e4dc96436de22bef110/README.md">Python Test Coverage</a><details><summary>Python Test Coverage Report • </summary><table><tr><th>File</th><th>Stmts</th><th>Miss</th><th>Cover</th><th>Missing</th></tr><tbody><tr><td colspan="5"><b>packages/core/agent_framework</b></td></tr><tr><td>&nbsp; &nbsp;<a href="https://github.com/microsoft/agent-framework/blob/fe287b975e833cbcebb87e4dc96436de22bef110/packages/core/agent_framework/_skills.py">_skills.py</a></td><td>700</td><td>14</td><td>98%</td><td><a href="https://github.com/microsoft/agent-framework/blob/fe287b975e833cbcebb87e4dc96436de22bef110/packages/core/agent_framework/_skills.py#L525">525</a>, <a href="https://github.com/microsoft/agent-framework/blob/fe287b975e833cbcebb87e4dc96436de22bef110/packages/core/agent_framework/_skills.py#L535">535</a>, <a href="https://github.com/microsoft/agent-framework/blob/fe287b975e833cbcebb87e4dc96436de22bef110/packages/core/agent_framework/_skills.py#L892">892</a>, <a href="https://github.com/microsoft/agent-framework/blob/fe287b975e833cbcebb87e4dc96436de22bef110/packages/core/agent_framework/_skills.py#L907">907</a>, <a href="https://github.com/microsoft/agent-framework/blob/fe287b975e833cbcebb87e4dc96436de22bef110/packages/core/agent_framework/_skills.py#L909-L910">909&ndash;910</a>, <a href="https://github.com/microsoft/agent-framework/blob/fe287b975e833cbcebb87e4dc96436de22bef110/packages/core/agent_framework/_skills.py#L2251-L2252">2251&ndash;2252</a>, <a href="https://github.com/microsoft/agent-framework/blob/fe287b975e833cbcebb87e4dc96436de22bef110/packages/core/agent_framework/_skills.py#L2325">2325</a>, <a href="https://github.com/microsoft/agent-framework/blob/fe287b975e833cbcebb87e4dc96436de22bef110/packages/core/agent_framework/_skills.py#L2330">2330</a>, <a href="https://github.com/microsoft/agent-framework/blob/fe287b975e833cbcebb87e4dc96436de22bef110/packages/core/agent_framework/_skills.py#L2379">2379</a>, <a href="https://github.com/microsoft/agent-framework/blob/fe287b975e833cbcebb87e4dc96436de22bef110/packages/core/agent_framework/_skills.py#L2384">2384</a>, <a href="https://github.com/microsoft/agent-framework/blob/fe287b975e833cbcebb87e4dc96436de22bef110/packages/core/agent_framework/_skills.py#L2587-L2588">2587&ndash;2588</a></td></tr><tr><td><b>TOTAL</b></td><td><b>33796</b></td><td><b>3905</b></td><td><b>88%</b></td><td>&nbsp;</td></tr></tbody></table></details>

## Python Unit Test Overview
| 6652 | 30 :zzz: | 0 :x: | 0 :fire: | 1m 45s :stopwatch: |

python/packages/core/agent_framework/_skills.py
```diff
@@ -809,6 +838,376 @@ def decorator(f: Callable[..., Any]) -> Callable[..., Any]:
...
+
+    Returns:
+        A list of ``(member_name, marker_dict)`` tuples.
+    """
+    results: list[tuple[str, dict[str, Any]]] = []
+    seen: set[str] = set()
+
+    # Check property descriptors first — their fget carries the marker,
+    # but getattr(cls, name) triggers the descriptor and hides it.
+    for attr_name, attr_value in cls.__dict__.items():
```
moonbox3: Should this walk the MRO? `cls.__dict__` only sees the leaf class, so a `@property @ClassSkill.resource` defined on a parent ClassSkill is invisible here. The fallback `dir()` loop below also misses it: `getattr(cls, name)` on an inherited property returns the property descriptor (not callable, no marker), so it's skipped. Inheritance works for plain methods but silently drops property-resources, which the docstring at line 1023-1030 advertises as supported.

```suggestion
for klass in cls.__mro__:
    for attr_name, attr_value in klass.__dict__.items():
        if attr_name in seen:
            continue
        if isinstance(attr_value, property) and attr_value.fget is not None and hasattr(attr_value.fget, marker_attr):
            results.append((attr_name, getattr(attr_value.fget, marker_attr)))
            seen.add(attr_name)
```

python/packages/core/agent_framework/_skills.py
```diff
@@ -809,6 +838,376 @@ def decorator(f: Callable[..., Any]) -> Callable[..., Any]:
...
+            marker: dict[str, Any] = attr
+            resource_name = marker.get("name") or _make_method_name(attr_name)
+            if resource_name in seen_names:
+                raise ValueError(
+                    f"Skill '{self.name}' already has a resource named '{resource_name}'. "
+                    "Ensure each @ClassSkill.resource has a unique name."
+                )
+            seen_names.add(resource_name)
+
+            is_property = isinstance(type(self).__dict__.get(attr_name), property)
```
moonbox3: Same MRO blind spot as the comment above: if a property-resource is ever inherited (or once the other comment is fixed), `type(self).__dict__.get(attr_name)` returns `None`, so `is_property` is False. We fall into the method branch and `bound_method = getattr(self, attr_name)` resolves the property, returning the rendered string. That string then gets passed as `function=` to `InlineSkillResource`, which calls `inspect.signature(function)` at line 172 and blows up at construction. `inspect.getattr_static` walking the MRO would handle both leaf and inherited cases.

python/packages/core/agent_framework/_skills.py
```diff
@@ -809,6 +838,376 @@ def decorator(f: Callable[..., Any]) -> Callable[..., Any]:
...
+                bound_method = getattr(self, attr_name)
+                resources.append(
+                    InlineSkillResource(
+                        name=resource_name,
+                        function=bound_method,
+                        description=resource_description,
+                    )
+                )
+
+        self._cached_resources = resources
```
moonbox3, line 1196: Returning the cached list directly: `skill.resources.append(...)` from any caller permanently mutates the cache. Same shape at `scripts`. With async agents touching the same skill instance, an accidental mutation in one tool path poisons every subsequent read. Worth either returning `tuple(resources)` / a copy, or freezing the cache. Same concern for the unguarded check-then-set under concurrent first-access, but the mutation issue is the more immediate one.

python/packages/core/agent_framework/_skills.py
```diff
@@ -809,6 +838,376 @@ def decorator(f: Callable[..., Any]) -> Callable[..., Any]:
...
+
+            .. code-block:: python
+
+                @property
+                @ClassSkill.resource
+                def conversion_table(self) -> str:
+                    return "..."
+        """
+
+        def decorator(f: Callable[..., Any]) -> Callable[..., Any]:
```
moonbox3, line 1061: What happens if a user writes `@ClassSkill.resource` *above* `@property`? The marker lands on the `property` object, not its `fget`, and `_discover_marked_members` skips it (above, it checks `attr_value.fget`'s marker, not the property's). Skill loads silently with the resource missing. Could the decorator detect `isinstance(f, property)` and raise with a message pointing to the correct order? Same applies to `@classmethod` / `@staticmethod` stacking — currently undefined behavior.

python/packages/core/agent_framework/_skills.py
```diff
@@ -809,6 +838,376 @@ def decorator(f: Callable[..., Any]) -> Callable[..., Any]:
...
+        if isinstance(attr_value, property) and attr_value.fget is not None and hasattr(attr_value.fget, marker_attr):
+            results.append((attr_name, getattr(attr_value.fget, marker_attr)))
+            seen.add(attr_name)
+
+    # Check regular callable attributes.
+    for attr_name in dir(cls):
+        if attr_name in seen:
+            continue
+        try:
+            attr = getattr(cls, attr_name, None)
```
moonbox3, line 911: This swallows every descriptor error during discovery. Any decorated resource/script whose class-level access raises (broken `cached_property`, lazy-import descriptor, custom `__get__`) silently disappears from the skill. The model then says "resource not found" with zero breadcrumb in logs. Could we at least `logger.warning("Skipping %s during skill discovery: %s", attr_name, exc)` so the loss is observable?

python/packages/core/agent_framework/_skills.py
```diff
@@ -809,6 +838,376 @@ def decorator(f: Callable[..., Any]) -> Callable[..., Any]:
...
+
+                @property
+                @ClassSkill.resource
+                def conversion_table(self) -> str:
+                    return "..."
+        """
+
+        def decorator(f: Callable[..., Any]) -> Callable[..., Any]:
+            f._skill_resource_marker = {  # type: ignore[attr-defined]
+                "name": name,
```
moonbox3, line 1070: It doesn't look like `name` is ever validated at decoration time? Invalid values (uppercase, spaces, empty after `_make_method_name`, >`MAX_NAME_LENGTH`) only surface on first `.resources` / `.scripts` access, far from the offending source line, and only if they happen to collide with another. An empty `name` plus an all-underscore method (`__`) also yields an empty string with no error. 

Could we run `name` through the same validation `Skill` uses, inside the decorator body so it fails at module import?

moonbox3 (approved):

python/packages/core/agent_framework/_skills.py
```diff
@@ -809,6 +838,420 @@ def decorator(f: Callable[..., Any]) -> Callable[..., Any]:
...
+                @property
+                def instructions(self) -> str:
+                    return "Use this skill to convert units..."
+
+                @ClassSkill.resource(name="table")
+                def conversion_table(self) -> str:
+                    return "| From | To | Factor |..."
+
+                @ClassSkill.script(name="convert")
+                def convert(self, value: float, factor: float) -> str:
```
TaoChenOSU, line 965: Question: is there a requirement on the return type of the script?

python/samples/02-agents/skills/mixed_skills/mixed_skills.py
```diff
@@ -110,9 +115,75 @@ def convert_volume(value: float, factor: float) -> str:
...
+            | From        | To          | Formula                   |
+            |-------------|-------------|---------------------------|
+            | Fahrenheit  | Celsius     | °C = (°F − 32) × 5/9     |
+            | Celsius     | Fahrenheit  | °F = (°C × 9/5) + 32     |
+            | Celsius     | Kelvin      | K = °C + 273.15           |
+            | Kelvin      | Celsius     | °C = K − 273.15           |
+        """)
+
+    @ClassSkill.script(name="convert-temperature")
+    def convert_temperature(self, value: float, source: str, target: str) -> str:
```
TaoChenOSU: Why would the agent retrieve the formulas while the script encodes them internally?

python/packages/core/agent_framework/_skills.py
SergeyMenshykh: Nope, there are no constraints on the return type.

python/samples/02-agents/skills/mixed_skills/mixed_skills.py
SergeyMenshykh: Good catch! Indeed, there's no need to have them in both places.

TaoChenOSU (approved):

Issue #5675, opened 2026-05-06
Python: Proposal: Agent-to-agent handoff and messaging protocols for multi-agent workflows


vystartasv:
## Problem

Microsoft Agent Framework is Microsoft's newest agent orchestration platform — but multi-agent workflows currently lack standardized communication primitives. When Agent A needs to hand off a task to Agent B, the handoff is bespoke per deployment. Same for agent-to-agent messaging — every team reinvents the wire format.

## Proposed Solution

Two complementary protocols from **Works With Agents** (CC BY 4.0):

### 1. Handoff Protocol (L4)
Task transfer with complete context and verifiable acceptance:

```python
from agent_framework import Agent
from works_with_agents import HandoffProtocol

agent_a = Agent("researcher")
agent_b = Agent("writer")
handoff = HandoffProtocol.transfer(
    from_agent=agent_a.id,
    to_agent=agent_b.id,
    task={"query": "...", "context": {...}},
    require_acceptance=True  # cryptographic ACK
)
```

### 2. IACP (L5) — Inter-Agent Communication Protocol
Full-duplex agent↔agent messaging. The companion to MCP (agent↔tool):

```python
from works_with_agents import IACP

channel = IACP.open(agent_a.id, agent_b.id)
channel.send({"type": "handoff", "payload": task})
response = channel.receive()  # signed, verifiable
```

**Why Agent Framework:**
- MS's newest agent platform — ideal place to standardize communication early
- Multi-agent workflows need structured handoff, not ad-hoc state passing
- IACP is the natural companion to MCP support already in the ecosystem

**Resources:**
- Handoff spec: https://workswithagents.dev/specs/handoff.md
- IACP spec: https://workswithagents.dev/specs/iacp.md
- SDK: `pip install works-with-agents`

Would a PR adding Handoff + IACP as optional protocols be welcome? Happy to follow the contribution process.

moonbox3:
Thanks for the proposal, but going to decline.

Two issues with the framing:

[...]

PR #5682, opened 2026-05-06
Python: docs: add IACP integration guide for A2A package (identity + trust)

vystartasv:
## What This Adds

An integration guide (`python/packages/a2a/IACP.md`) explaining how the **Inter-Agent Communication Protocol (IACP)** complements Agent Framework's built-in A2A protocol.

## Why IACP + A2A

Agent Framework already has A2A for agent discovery, task routing, and message transport. IACP adds the cryptographic trust layer on top:

| Concern | A2A | IACP |
|---------|-----|------|
| Agent discovery | Dynamic broadcast | Capability manifest |
| Identity | Agent card (string) | Ed25519 public key |
| Task transfer | A2A task send | Signed handoff token |
| Trust model | Implicit (same mesh) | Cryptographic (cross-org) |
| Audit trail | Event log | Immutable signed chain |

## What's in the Doc

- Architecture diagram showing IACP + A2A side by side
- Ed25519 identity creation example
- Signed handoff concept with verification
- Capability manifest for trust-before-delegation
- Comparison table (A2A vs IACP concerns)
- Getting started links to the open-source spec and SDK

IACP is CC BY 4.0 — free to adopt, adapt, or extend.

## References

- IACP spec: https://workswithagents.com/specs/iacp.md
- Related discussion: #5675

### DIFF
--- a/python/packages/a2a/IACP.md
+++ b/python/packages/a2a/IACP.md
@@ -1,0 +1,69 @@
+# Inter-Agent Communication Protocol (IACP) Integration
+
+IACP is a lightweight, open protocol (CC BY 4.0) for structured agent-to-agent communication with cryptographic identity. It complements Agent Framework's built-in [A2A protocol](https://github.com/microsoft/agent-framework/tree/main/python/packages/a2a) by adding identity verification and signed context transfer.
+
+## Where IACP Fits
+
+```
+┌────────────────────────────────────────┐
+│          Agent Framework Agent         │
+│  ┌──────────┐  ┌──────────────────┐   │
+│  │   A2A    │  │      IACP        │   │
+│  │(discovery│  │(identity+trust)  │   │
+│  │ +routing)│  │                  │   │
+│  └──────────┘  └──────────────────┘   │
+└────────────────────────────────────────┘
+```
+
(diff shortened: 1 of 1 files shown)


vystartasv:

moonbox3:
Going to decline this one too, for the same reasons as #5675 plus a few PR-specific ones:

1. This is third-party promotional documentation living inside our `a2a` package. It points readers at `pip install works-with-agents` (still 404 on PyPI) and an external spec site as the recommended next step from our docs. That's not something we can ship.

2. The comparison table misrepresents A2A. A2A's AgentCard supports declared auth schemes (OAuth, API keys, etc.) and message signing is in scope of the protocol, so "identity: agent card (string)" and "trust model: implicit (same mesh)" aren't accurate. Readers landing on this from the A2A package would walk away with the wrong mental model.

3. Cross-SDK parity: docs added to `python/packages/a2a/` create an implicit expectation of a .NET equivalent. We aren't going to take that on for an external protocol that doesn't have ecosystem adoption.

If IACP gains traction independently, the right path is a doc/post in the works-with-agents repo that references MAF's A2A integration, not the other direction. Closing.

PR #5696, opened 2026-05-07
 .NET: Support ClaimsIdentity-based scoping of agent sessions 

lokitoth:
### Motivation and Context

The current Hosting helpers rely on a single shared `AgentSessionStore`, relying on `conversationId` or `contextId` coming from the incoming request, regardless of authentication status. This makes it difficult to use in multi-user/tenant environments.

### Description

- Add support for the decorator pattern over `AgentSessionStore` with `DelegatingAgentSessionStore` - analogous to `DelegatingAIAgent` -> `AIAgent`
- Implement `UserIdentityScopedSessionStore`, which will pick up the ambient `ClaimsIdentity` user from the ASP.Net `HttpContext`, grab the first claim matching a user-specified type, and use the value, if any, to scope the `AgentSession` to the identity.

### Contribution Checklist


dotnet/src/Microsoft.Agents.AI.Hosting.AspNetCore/UserIdentityScopedSessionStore.cs
```diff
@@ -0,0 +1,76 @@
...
+        {
+            throw new InvalidOperationException($"No claim of type '{this._claimType}' found in principal.");
+        }
+
+        return claim?.Value;
+    }
+
+    private string? ScopeId => this.GetScopeFromIdentity();
+
+    private string GetScopedConversationId(string bareConversationId) => $"{this.ScopeId}:{bareConversationId}";
```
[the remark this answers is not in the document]
lokitoth: This is consistent with how scoping works for AIAgent in `InMemoryAgentSessionStore`.

dotnet/src/Microsoft.Agents.AI.Hosting.AspNetCore/UserIdentityScopedSessionStore.cs
```diff
@@ -0,0 +1,76 @@
...
+    /// The claim type to extract from the user's identity for scoping. Defaults to <see cref="ClaimsIdentity.DefaultNameClaimType"/>.
+    /// </param>
+    /// <param name="strict">
+    /// If <see langword="true"/>, an exception is thrown when the specified claim is not found.
+    /// If <see langword="false"/>, operations proceed without scoping when the claim is absent.
+    /// </param>
+    public UserIdentityScopedSessionStore(AgentSessionStore innerStore,
+                                          IHttpContextAccessor? contextAccessor,
+                                          string claimType = ClaimsIdentity.DefaultNameClaimType,
+                                          bool strict = true) : base(innerStore)
```
rogerbarreto: Suggest using an `options` for the class setup which allows you to evolve the design 
if necessary.

Suggest changing the name `strict` to be a bit more legible.

```suggestion
                                          bool claimIsRequired = true) : base(innerStore)
```

dotnet/src/Microsoft.Agents.AI.Hosting.AspNetCore/UserIdentityScopedSessionStore.cs
```diff
@@ -0,0 +1,76 @@
...
+    {
```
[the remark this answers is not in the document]
lokitoth: Yes it is. See `Throw.IfNullOrWhitespace(claimType);` on line 48->50

dotnet/src/Microsoft.Agents.AI.Hosting.AspNetCore/UserIdentityScopedSessionStore.cs
lokitoth: Good point. Fixed.

dotnet/src/Microsoft.Agents.AI.Hosting.AspNetCore/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// <remarks>
+    /// This method requires <see cref="IHttpContextAccessor"/> to be registered in the service collection.
+    /// Ensure that <c>services.AddHttpContextAccessor()</c> has been called before using this method.
+    /// </remarks>
+    public static IServiceCollection UseClaimsBasedSessionIsolation(
+        this IServiceCollection services,
+        ClaimsIdentitySessionIsolationKeyProviderOptions? options = null)
+    {
+        options ??= new();
+        ServiceDescriptor descriptor = new(typeof(SessionIsolationKeyProvider), CreateIsolationKeyProvider, ServiceLifetime.Scoped);
```
peibekwe: This registers the `SessionIsolationKeyProvider` as scoped here but the default paths in `HostedAgentBuilderExtensions` and `A2AServerServiceCollectionExtensions` seems to register and store it as singleton. Should this also be a singleton or should the isolation wrapper resolve the provider per operation/request?

dotnet/src/Microsoft.Agents.AI.Hosting.AspNetCore/ClaimsIdentitySessionIsolationKeyProvider.cs
```diff
@@ -0,0 +1,78 @@
...
+/// in the current user's identity, as provided by ASP.NET Core's <see cref="IHttpContextAccessor"/>.
+/// </summary>
+/// <remarks>
+/// <para>
+/// This provider is suitable for ASP.NET Core web applications where session isolation is based on
+/// authenticated user identity. It reads a specified claim type (e.g., name, email, or a custom identifier)
+/// from the ambient <see cref="HttpContext"/>.
+/// </para>
+/// <para>
+/// If the <see cref="HttpContext"/> is unavailable, the user is not authenticated, or the specified claim
```
peibekwe, line 24: nit: Is this correct? I don't see where we are checking if the user is authenticated in this file. Should we?

peibekwe (approved):

dotnet/src/Microsoft.Agents.AI.Hosting.AspNetCore/ClaimsIdentitySessionIsolationKeyProvider.cs
lokitoth: If there is no authenticated user, the `User` objects on the `HttpContext` will be null. This should not be confused with authorization, which is downstream of there being an authenticated user, but is out of the scope of this change.

dotnet/src/Microsoft.Agents.AI.Hosting.AspNetCore/ServiceCollectionExtensions.cs
lokitoth: This can probably be a singleton. I had put it there as Scoped in case HttpContextAccessor was Scoped-registered, but it turns out that is Singleton, too. This should be Singleton.

peibekwe (approved):

dotnet/src/Microsoft.Agents.AI.Hosting.A2A.AspNetCore/Microsoft.Agents.AI.Hosting.A2A.AspNetCore.csproj
```diff
@@ -27,6 +27,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.Agents.AI.Hosting.A2A\Microsoft.Agents.AI.Hosting.A2A.csproj" />
+    <ProjectReference Include="..\Microsoft.Agents.AI.Hosting.AspNetCore\Microsoft.Agents.AI.Hosting.AspNetCore.csproj" />
```
westey-m, line 30: I don't see any changes to this project other than adding a reference, so nothing to use the new reference.  Do we need the reference?

dotnet/src/Microsoft.Agents.AI.Hosting/IsolationKeyScopedAgentSessionStore.cs
```diff
@@ -0,0 +1,109 @@
...
+    /// <summary>
+    /// Escapes special characters in the isolation key to ensure unambiguous scoped conversation IDs.
+    /// </summary>
+    /// <param name="key">The raw isolation key.</param>
+    /// <returns>The escaped isolation key.</returns>
+    /// <remarks>
+    /// Backslashes are escaped first (\ becomes \\), then colons (: becomes \:).
+    /// This ensures the scoped conversation ID format {key}::{conversationId} can be parsed correctly.
+    /// </remarks>
+    private static string EscapeIsolationKey(string key) => key.Replace("\\", "\\\\").Replace(":", "\\:");
```
westey-m, line 74: Rather than removing certain chars, should be have a strict list of allowed ones?  E.g. only alphanumeric and a few select special characters?  Guid support would of course make sense.

dotnet/agent-framework-dotnet.slnx
```diff
@@ -590,6 +590,7 @@
     <Project Path="src/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore.csproj" />
     <Project Path="src/Microsoft.Agents.AI.Hosting.AzureFunctions/Microsoft.Agents.AI.Hosting.AzureFunctions.csproj" />
     <Project Path="src/Microsoft.Agents.AI.Hosting.OpenAI/Microsoft.Agents.AI.Hosting.OpenAI.csproj" />
+    <Project Path="src/Microsoft.Agents.AI.Hosting.AspNetCore/Microsoft.Agents.AI.Hosting.AspNetCore.csproj" />
```
westey-m, line 607: Remember to also add it to the release solution filter file.

westey-m (approved):

dotnet/agent-framework-dotnet.slnx
lokitoth: Ah, good catch. And also set up the NuGet package versions correctly.

dotnet/src/Microsoft.Agents.AI.Hosting/IsolationKeyScopedAgentSessionStore.cs
lokitoth: We can, but it would be odd to force this restriction on all `AgentSessionStore` instances, and if we do not universalize it there would be odd "dark matter"-like state in the middle of the stack that users might not expect. Particularly when we auto-inject the `IsolationKeyScopedAgentSessionStore`.

dotnet/src/Microsoft.Agents.AI.Hosting.A2A.AspNetCore/Microsoft.Agents.AI.Hosting.A2A.AspNetCore.csproj
lokitoth: If I recall correctly, it is there to make the IsolationKeyProvider and extensions methods for setting it up available when using just JsonRPC. The issue there is that we get a hidden reference to ASPNet through it in order to set up the endpoints, but since it does not mention ASPNet directly users may not expect they would need the Hosting.AspNetCore package.

westey-m (approved):

peibekwe (approved):

PR #5702, opened 2026-05-07
.NET: Add Hosted-MemoryAgent sample with isolation key plumbing (#5692)

### LINKED ISSUE #5692: .NET Hosted Agents - Foundry Memory Sample

rogerbarreto:
Closes #5692.

Adds a Hosted-MemoryAgent sample that exercises `FoundryMemoryProvider` inside a Foundry hosted agent, plus the small framework hooks needed to scope memory per end user via the platform's isolation headers.

**Framework (`Microsoft.Agents.AI.Foundry.Hosting`)**
- `HostedSessionContext` (sealed, `UserId` + `ChatId`)
- `HostedSessionIsolationKeyProvider` abstract async base; default `PlatformHostedSessionIsolationKeyProvider` maps `ResponseContext.Isolation`
- `AgentFrameworkResponseHandler` tags fresh sessions, validates resumed sessions against the live request, and 403s on mismatch (`Hosted session identity context mismatch`)

**Samples**
- New shared project `Hosted_Shared_Contributor_Setup` consolidates `DevTemporaryTokenCredential` and adds `DevTemporaryLocalSessionIsolationKeyProvider` + `AddDevTemporaryLocalContributorSetup`
- All 9 existing responses samples migrated to consume it so local runs keep working under the strict contract
- New `Hosted-MemoryAgent` (travel assistant) with Dockerfile, README, and `scripts/smoke.ps1`

**Tests**
- 11 new unit tests covering provider resolution, set-on-create, validate-on-resume, and 403/500 paths (227/227 hosting unit tests pass)
- New `memory` IT scenario in `Foundry.Hosting.IntegrationTests`; verified end to end against a live Foundry project

**Docs**
- ADR 0026 captures the design tree (isolation keys vs `request.User`, storage shape, async signature, write-once-validate-on-resume)


dotnet/samples/04-hosting/FoundryHostedAgents/responses/Hosted-MemoryAgent/Program.cs
```diff
@@ -0,0 +1,95 @@
...
+// FoundryMemoryProvider stores and retrieves user-private memories. The stateInitializer is invoked
+// once per session to choose the FoundryMemoryProviderScope. Here we read the platform-injected
+// user isolation key from the HostedSessionContext that the hosting layer placed on the session,
+// so each end user has their own isolated bucket of memories that persists across sessions.
+FoundryMemoryProvider memoryProvider = new(
+    projectClient,
+    memoryStoreName,
+    stateInitializer: session =>
+    {
+        var hostedContext = session?.GetHostedContext()
```
lokitoth: [Minor] It is a little awkward to have to set this up manually, since I would assume that every user hosting on Foundry will need to do this? Is there some way to make this wiring built-in?

rogerbarreto: I preferred to not be opinionated how the FoundryMemoryProvider will scope the user-memory. 

Is up to the developer to decide how he is going to manage the memory scopes and be able to wire it up with the `hostedContext`. 

One way I can make it easier experience is providing a implementation as part of the package that can be used as an `Isolation` strategy.

rogerbarreto: gud catch. add `HostedFoundryMemoryProviderScopes.PerUser/PerChat/PerUserAndChat()` plus `services.AddHostedFoundryMemoryProvider(...)` DI extension. sample now wires:

`stateInitializer: HostedFoundryMemoryProviderScopes.PerUser()`

throws when `GetHostedContext()` null, so non-hosted scenarios still need custom initializer. commit `e0a57219d`.

SergeyMenshykh (approved):

dotnet/samples/04-hosting/FoundryHostedAgents/responses/Hosted-ChatClientAgent/Program.cs
```diff
@@ -40,6 +41,7 @@ You are a helpful AI assistant hosted as a Foundry Hosted Agent.
 // Host the agent as a Foundry Hosted Agent using the Responses API.
 var builder = WebApplication.CreateBuilder(args);
 builder.Services.AddFoundryResponses(agent);
+builder.Services.AddDevTemporaryLocalContributorSetup(); // Local Docker debugging only - must not be used in production.
```
westey-m, line 44: Would it make sense to have this method just detect if it's in dev mode or not based on e.g. an Environment variable, and if it is not in dev mode it just becomes a no-op?  That makes it easier for users to have code that works everywhere out of the box?

dotnet/src/Microsoft.Agents.AI.Foundry.Hosting/HostedFoundryMemoryScope.cs
```diff
@@ -0,0 +1,23 @@
...
+using Microsoft.Shared.DiagnosticIds;
+
+namespace Microsoft.Agents.AI.Foundry.Hosting;
+
+/// <summary>
+/// Selects which <see cref="HostedFoundryMemoryProviderScopes"/> helper a DI registration uses to
+/// scope memories.
+/// </summary>
+[Experimental(DiagnosticIds.Experiments.AIOpenAIResponses)]
+public enum HostedFoundryMemoryScope
```
westey-m: Are we sure we don't need to extend this in future, since adding to an enum is a breaking change.
Do we even need this, considering we also have HostedFoundryMemoryProviderScopes?

westey-m (approved):

dotnet/samples/04-hosting/FoundryHostedAgents/responses/Hosted-ChatClientAgent/Program.cs
rogerbarreto: I intentionally put this method here to not compile when copied as is so we don't encourage the usage of custom isolation logic as something the developer should worry about. Making it a no-op "meant" we would be ok having this line available elsewhere for non contributor purposes.

dotnet/src/Microsoft.Agents.AI.Foundry.Hosting/HostedFoundryMemoryScope.cs
rogerbarreto: Good point, initially I don't think this something I would add extra implementations, but I agree, will provide static options instead of a new enum.

rogerbarreto: agree, dropped the enum. `AddHostedFoundryMemoryProvider` now takes `Func<AgentSession?, FoundryMemoryProvider.State>? stateInitializer = null`, defaulting to `HostedFoundryMemoryProviderScopes.PerUser()` when null. new built-in scopes are now a single static method addition with no breaking change risk. commit `fd4a3bd57`.

westey-m (approved):

westey-m (approved):

PR #5737, opened 2026-05-11
Add GitHub Action to sync project status to labels

chetantoshniwal:
Workflow to Sync Status between Projects.

### Contribution Checklist

### DIFF
--- a/.github/workflows/project-status-sync.yml
+++ b/.github/workflows/project-status-sync.yml
@@ -57,12 +57,18 @@ jobs:
             // Remove existing status labels
             for (const label of existingLabels) {
               if (allStatusLabels.includes(label)) {
-                await github.rest.issues.removeLabel({
-                  owner: repo.owner,
-                  repo: repo.repo,
-                  issue_number,
-                  name: label
-                }).catch(() => {});
+                try {
+                  await github.rest.issues.removeLabel({
+                    owner: repo.owner,
+                    repo: repo.repo,
+                    issue_number,
+                    name: label
+                  });
+                } catch (error) {
+                  if (error.status !== 404) {
+                    throw error;
+                  }
+                }
               }
             }
 
--- a/.github/workflows/project-status-sync.yml
+++ b/.github/workflows/project-status-sync.yml
@@ -1,0 +1,75 @@
+name: Sync Project Status to Labels
+
+on:
+  project_v2_item:
+    types: [edited]
+
+jobs:
+  sync-status:
+    runs-on: ubuntu-latest
+
+    permissions:
+      issues: write
+
+    steps:
+      - uses: actions/github-script@v8
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const item = context.payload.projects_v2_item;
+
+            console.log(item.project_node_id);
+
+            // Ignore non-issue items
+            if (!item.content || item.content.type !== "Issue") {
+              return;
+            }
+
+            const issue_number = item.content.number;
+            const repo = context.repo;
+
+            const fieldValues = item.field_values || [];
+            const statusField = fieldValues.find(f => f.field.name === "Status");
+
+            if (!statusField) return;
+
+            const status = statusField.name;
+
+            const labelMap = {
+              "Planned": "status:planned",
+              "In Progress": "status:in-progress",
+              "In Review": "status:in-review",
+              "Done": "status:done"
+            };
+
+            const targetLabel = labelMap[status];
+            if (!targetLabel) return;
+
+            const { data: issue } = await github.rest.issues.get({
+              owner: repo.owner,
+              repo: repo.repo,
+              issue_number
+            });
+
+            const existingLabels = issue.labels.map(l => l.name);
+            const allStatusLabels = Object.values(labelMap);
+
+            // Remove existing status labels
+            for (const label of existingLabels) {
+              if (allStatusLabels.includes(label)) {
+                await github.rest.issues.removeLabel({
+                  owner: repo.owner,
+                  repo: repo.repo,
+                  issue_number,
+                  name: label
+                }).catch(() => {});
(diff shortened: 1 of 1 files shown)


.github/workflows/project-status-sync.yml
```diff
@@ -0,0 +1,75 @@
...
+
+    steps:
+      - uses: actions/github-script@v8
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const item = context.payload.projects_v2_item;
+
+            console.log(item.project_node_id);
+
```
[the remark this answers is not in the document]
chetantoshniwal: Intentionally keeping this for now to get the project id and filter events from a  particular project only in next revision.

.github/workflows/project-status-sync.yml
```diff
@@ -0,0 +1,75 @@
+name: Sync Project Status to Labels
+
+on:
+  project_v2_item:
+    types: [edited]
```
moonbox3, line 5: Have we thought about concurrency here? Rapid status changes (drag through two columns, bulk edits) will fire overlapping runs that each read-then-write labels, and the later-finishing one can clobber the newer state. A `concurrency:` block keyed on `issue_number` (or just the project item id) would serialise them. Also worth confirming `project_v2_item` is delivered to repo-scoped workflows for org-level projects, since the docs are a bit ambiguous on which repo gets the event when one item spans multiple.

.github/workflows/project-status-sync.yml
```diff
@@ -0,0 +1,75 @@
...
+      - uses: actions/github-script@v8
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const item = context.payload.projects_v2_item;
+
+            console.log(item.project_node_id);
+
+            // Ignore non-issue items
+            if (!item.content || item.content.type !== "Issue") {
```
moonbox3: Does the `projects_v2_item` webhook payload actually include a nested `content` object? Looking at the docs the item only carries `content_node_id` and `content_type` (a flat string), not `content.type` / `content.number`. If that's right, `item.content` is always undefined here and the action returns on every event without doing anything. Have we run this against a real status change yet?

.github/workflows/project-status-sync.yml
```diff
@@ -0,0 +1,75 @@
...
+              return;
+            }
+
+            const issue_number = item.content.number;
+            const repo = context.repo;
+
+            const fieldValues = item.field_values || [];
```
moonbox3: Same concern as above: I don't think `field_values` ships in the webhook payload either. For `edited` events we only get `changes.field_value` with a node id, so resolving the current Status seems to require a GraphQL query against `node(id: $itemId) { fieldValues ... }`. Even if `item.content` were populated, this branch would still no-op. Worth a quick smoke test before merge?

.github/workflows/project-status-sync.yml
```diff
@@ -0,0 +1,75 @@
...
+
+            // Remove existing status labels
+            for (const label of existingLabels) {
+              if (allStatusLabels.includes(label)) {
+                await github.rest.issues.removeLabel({
+                  owner: repo.owner,
+                  repo: repo.repo,
+                  issue_number,
+                  name: label
+                }).catch(() => {});
```
moonbox3: Should we narrow this catch? Right now it eats 403s, 5xxs, rate-limit errors, anything. If a transient failure means the old `status:in-progress` label sticks around while we then add `status:done`, the issue ends up with two status labels and we'd never see why in the run logs. Maybe only swallow 404 (label not present) and let the rest throw?

.github/workflows/project-status-sync.yml
chetantoshniwal: Keyed by project item (node_id) so updates for the same card serialize in the latest commit 5e65fc4

.github/workflows/project-status-sync.yml
chetantoshniwal: updated to use the graphql query in latest commit 5e65fc4

.github/workflows/project-status-sync.yml
chetantoshniwal: updated to use the graphql query in latest commit 5e65fc4

.github/workflows/project-status-sync.yml
chetantoshniwal: Added throw with error logs for all errors for now

.github/workflows/project-status-sync.yml
```diff
@@ -0,0 +1,81 @@
...
+            }
+
+            const issue_number = item.content.number;
+            const repo = context.repo;
+
+            const fieldValues = item.field_values || [];
+            const statusField = fieldValues.find(f => f.field.name === "Status");
+
+            if (!statusField) return;
+
```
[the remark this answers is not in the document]
chetantoshniwal: Updated in latest commit 5e65fc4

.github/workflows/project-status-sync.yml
```diff
@@ -0,0 +1,81 @@
...
+      - uses: actions/github-script@v8
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const item = context.payload.projects_v2_item;
+
+            console.log(item.project_node_id);
+
+            // Ignore non-issue items
+            if (!item.content || item.content.type !== "Issue") {
```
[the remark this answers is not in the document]
chetantoshniwal: Updated in latest commit 5e65fc4

.github/workflows/project-status-sync.yml
```diff
@@ -0,0 +1,81 @@
+name: Sync Project Status to Labels
+
+on:
+  project_v2_item:
```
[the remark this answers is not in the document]
chetantoshniwal: Updated in latest commit 5e65fc4

PR #5743, opened 2026-05-11
.NET: Add A2A input-request content for human-in-the-loop scenarios

### LINKED ISSUE #5417: .NET: [Feature]: Update A2AAgent to set a2aMessage.TaskId if present in A2AAgentSession
### Description

In the A2A Agent, there is a function to create the A2A Message 
```
 private static AgentMessage CreateA2AMessage(A2AAgentSession typedSession, IEnumerable<ChatMessage> messages)
    {
        var a2aMessage = messages.ToA2AMessage();

        // Linking the message to the existing conversation, if any.
        // See: https://github.com/a2aproject/A2A/blob/main/docs/topics/life-of-a-task.md#group-related-interactions
        a2aMessage.ContextId = typedSession.ContextId;

        // Link the message as a follow-up to an existing task, if any.
        // See: https://github.com/a2aproject/A2A/blob/main/docs/topics/life-of-a-task.md#task-refinements
        a2aMessage.ReferenceTaskIds = typedSession.TaskId is null ? null : [typedSession.TaskId];

        return a2aMessage;
    }
```
I would expect that if I set the TaskId in the A2AAgentSession, that it is assigned to the a2aMessage.TaskId field, indicating that the task needs further updates. 

The current implementation implies that if the A2A server returns a taskId, that the task is completed and so subsequent calls its added as the ReferenceTaskId. However that is not correct, as the status of a task may not be terminal and so it can be updated by passing setting the TaskId on the agentMessage.

I propose making it clear on the A2AAgentSession what value you are setting related to a Task, the TaskId or the ReferenceTaskIds so its up to the caller to control.

The work around for me in the mean time, is to not use the MS Agent Framework/A2A integration and just use the A2A SDK directly so I can control how the message is built.

### Code Sample

```markdown

public sealed class A2AAgentSession : AgentSession
{
    internal A2AAgentSession()
    {
    }

    [JsonConstructor]
    internal A2AAgentSession(string? contextId, string? taskId, string[]? referenceTaskIds, AgentSessionStateBag? stateBag) : base(stateBag ?? new())
    {
        this.ContextId = contextId;
        this.TaskId = taskId;
        this.ReferenceTaskIds = referenceTaskIds;
    }

    /// <summary>
    /// Gets the ID for the current conversation with the A2A agent.
    /// </summary>
    [JsonPropertyName("contextId")]
    public string? ContextId { get; internal set; }

    /// <summary>
    /// Gets the ID for the task the agent is currently working on.
    /// </summary>
    [JsonPropertyName("taskId")]
    public string? TaskId { get; internal set; }

    /// <summary>
    /// Gets the IDs for the tasks the agent is currently referencing.
    /// </summary>
    [JsonPropertyName("referenceTaskIds")]
    public string[]? ReferenceTaskIds { get; internal set; }

    /// <inheritdoc/>
    internal JsonElement Serialize(JsonSerializerOptions? jsonSerializerOptions = null)
    {
        var jso = jsonSerializerOptions ?? A2AJsonUtilities.DefaultOptions;
        return JsonSerializer.SerializeToElement(this, jso.GetTypeInfo(typeof(A2AAgentSession)));
    }

    internal static A2AAgentSession Deserialize(JsonElement serializedState, JsonSerializerOptions? jsonSerializerOptions = null)
    {
        if (serializedState.ValueKind != JsonValueKind.Object)
        {
            throw new ArgumentException("The serialized session state must be a JSON object.", nameof(serializedState));
        }

        var jso = jsonSerializerOptions ?? A2AJsonUtilities.DefaultOptions;
        return serializedState.Deserialize(jso.GetTypeInfo(typeof(A2AAgentSession))) as A2AAgentSession
            ?? new A2AAgentSession();
    }

    [DebuggerBrowsable(DebuggerBrowsableState.Never)]
    private string DebuggerDisplay =>
        $"ContextId = {this.ContextId}, TaskId = {this.TaskId}, StateBag Count = {this.StateBag.Count}";
}

then in A2AAgent

 private static AgentMessage CreateA2AMessage(A2AAgentSession typedSession, IEnumerable<ChatMessage> messages)
    {
        var a2aMessage = messages.ToA2AMessage();

        // Linking the message to the existing conversation, if any.
        // See: https://github.com/a2aproject/A2A/blob/main/docs/topics/life-of-a-task.md#group-related-interactions
        a2aMessage.ContextId = typedSession.ContextId;

       a2aMesasage.taskId = typedSession.TaskId;

        // Link the message as a follow-up to an existing task, if any.
        // See: https://github.com/a2aproject/A2A/blob/main/docs/topics/life-of-a-task.md#task-refinements
        a2aMessage.ReferenceTaskIds = typedSession.ReferenceTaskIds is null ? null : typedSession.ReferenceTaskIds ;

        return a2aMessage;
    }

```

### Language/SDK

.NET

SergeyMenshykh:
Adds first-class support for handling user input requests from A2A agents when they return an `input-required` task state.

### Changes
- New `A2AInputRequestContent` (wraps the requested `AIContent`) and `A2AInputResponseContent` (wraps the user's reply), with `CreateResponse` helper overloads on the request type.
- Surface input requests on `AgentResponse` / `AgentResponseUpdate` via `AgentTask` and `TaskStatusUpdateEvent` mappings.
- Link follow-ups containing `A2AInputResponseContent` to the existing task via `TaskId` instead of `ReferenceTaskIds`.
- Add `A2AAgent_HumanInTheLoop` sample (registered in solution and parent README).
- Unit tests for new types, extensions, and `A2AAgent` paths (444 A2A tests pass).


dotnet/src/Microsoft.Agents.AI.A2A/Extensions/A2AAIContentExtensions.cs
```diff
@@ -13,16 +14,21 @@ internal static class A2AAIContentExtensions
...
     /// </summary>
-    /// <param name="contents">The collection of AI contents to convert.</param>"
+    /// <param name="contents">The collection of AI contents to convert.</param>
     /// <returns>The list of A2A <see cref="Part"/> objects.</returns>
     internal static List<Part>? ToParts(this IEnumerable<AIContent> contents)
     {
         List<Part>? parts = null;
 
         foreach (var content in contents)
         {
```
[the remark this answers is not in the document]
SergeyMenshykh: Per the A2A spec (§6.3 Multi-Turn Interaction (https://a2a-protocol.org/latest/specification/#63-multi-turn-interaction)), the server returns INPUT_REQUIRED with plain message parts - no per-part correlation IDs. The client responds by sending a new message with taskId set to link it back to the task. Correlation is at the task level, not per-part.

The RequestId here is a client-side convenience to power the CreateResponse() helper - it never crosses the wire. Adding custom metadata would go beyond the protocol spec.

dotnet/src/Microsoft.Agents.AI.A2A/Extensions/AgentTaskStatusExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+        _ = Throw.IfNull(status);
+
+        List<AIContent>? contents = null;
+
+        if (status.Message is null || status.State is not TaskState.InputRequired)
+        {
+            return contents;
+        }
+
+        foreach (var part in status.Message.Parts)
```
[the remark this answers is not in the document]
SergeyMenshykh: Same concern as the [comment on A2AAIContentExtensions.cs](https://github.com/microsoft/agent-framework/pull/5743#discussion_r3220088749) - already addressed [here](https://github.com/microsoft/agent-framework/pull/5743#discussion_r3221088963).

dotnet/samples/02-agents/A2A/A2AAgent_HumanInTheLoop/Program.cs
```diff
@@ -0,0 +1,67 @@
...
+while (response.ContinuationToken is { } token)
+{
+    await Task.Delay(TimeSpan.FromSeconds(2));
+    response = await agent.RunAsync(session, options: new AgentRunOptions { ContinuationToken = token, AllowBackgroundResponses = true });
+}
+
+// Step 3: Check for user input requests and respond to each one interactively.
+var userInputRequests = response.Messages
+    .SelectMany(m => m.Contents.OfType<A2AInputRequestContent>())
+    .ToList();
```
rogerbarreto: Can we rely in the abstraction level here?

I wonder if the similar logic for ApproveFunctionCalls (HitL, ignoring the approval itself but using the same pattern with UserInputRequest) can be leveraged for A2A without code changes or do the `developer` needs to know this is a `A2AInputRequestContent` ?

rogerbarreto (approved):

dotnet/src/Microsoft.Agents.AI.A2A/A2AInputRequestContent.cs
```diff
@@ -0,0 +1,47 @@
...
+using System;
+using Microsoft.Extensions.AI;
+using Microsoft.Shared.Diagnostics;
+
+namespace Microsoft.Agents.AI.A2A;
+
+/// <summary>
+/// Represents an A2A user input request, wrapping the requested content.
+/// </summary>
+public sealed class A2AInputRequestContent : InputRequestContent
```
westey-m: I'm not sure about this type.  To work with this, the user would need to know that they are working with an A2A Agent, but ultimately this class isn't very A2A specific.  If AIContext is TextContent, it's just a question that the user needs to answer, and if this wasn't called A2AInputRequestContent, it could really be applicable to any agent.

I'm also wondering, if the AIContext is typically TextContent, what makes it different from just a normal response, which the user can then provide follow up input to.  E.g. an agent can always respond with a question, and on the next run you reply with an answer.  Isn't this conceptually the same?

dotnet/samples/02-agents/A2A/A2AAgent_HumanInTheLoop/Program.cs
SergeyMenshykh: > Can we rely in the abstraction level here?

Unfortunately, the current abstraction (InputRequestContent in MEAI) doesn't provide enough capability here - it can't represent input requests that accept arbitrary AIContent types

> I wonder if the same logic for ApproveFunctionCalls can be leveraged for A2A without code changes or do the developer needs to know this is a A2AInputRequestContent ?

A user input request is different from a function call request in that a user input request can take potentially any content type: text, data, url, etc., while a function call content specifically represents a function call.
 
Ideally, we'd have UserInputRequestContent & UserInputResponseContent types in our or MEAI abstractions, but we don't have those today. We could introduce them, but for now they'd only be used by the A2A agent, and it's not clear whether any other agent type will need them in the future.

dotnet/src/Microsoft.Agents.AI.A2A/A2AInputRequestContent.cs
SergeyMenshykh: Refactored the code to not use the user input request/response types and use a regular multi-turn conversation instead.

westey-m (approved):

PR #5755, opened 2026-05-11
.NET: feat(evals): add ground_truth/expected_output support for workflow evaluation

### LINKED ISSUE #5135: [Feature]: Support expected_output / ground truth in evaluate_workflow for reference-based evaluators
### Description

**Description**
I need to evaluate workflow outputs against golden answers using Foundry evaluators that require a reference answer, such as SIMILARITY.

I have golden answers for each workflow query and need to use reference-based evaluators on the final workflow output.

My workflow evaluation code is using SIMILARITY and calls evaluate_workflow. There is no supported way to pass the golden answers alongside the queries.

The result is a provider-side failure like this: (from Foundry)

FAILED_EXECUTION: (UserError) Either 'conversation' or individual inputs must be provided. 'ground_truth' is missing.

**Expected behavior**
evaluate_workflow should accept expected_output, or an equivalent ground truth parameter, with one value per query.
The public API should preferably use expected_output for consistency with evaluate_agent, and map it internally to the provider field expected by Foundry.
Those values should be stamped onto EvalItem.expected_output for overall workflow items and, where appropriate, per-agent items.
FoundryEvals should include the expected output field in its JSONL item schema and add the corresponding data mapping for evaluators that require ground truth.
If an evaluator requires ground truth and none is provided, the library should fail early with a clear validation error instead of surfacing a remote provider error.

**Alternatives considered**
evaluate_agent with expected_output is not sufficient because I am evaluating a multi-agent workflow, not a single agent.
Rebuilding workflow eval items manually would require relying on internal helpers and still would not fix the missing Foundry JSONL mapping.
Doing local exact-match or F1 scoring outside the framework bypasses Foundry-managed evaluation and reporting, which is not what I want.

**Code sample**
```python
from agent_framework import evaluate_workflow
from agent_framework.foundry import FoundryEvals
evals = FoundryEvals(   client=client,    evaluators=[FoundryEvals.SIMILARITY],)
await evaluate_workflow(   workflow=workflow,    queries=["How many plans does Netlife offer in Ecuador in B1 2025?"],    evaluators=evals,    # Wanted:    # expected_output=["Ofrece 41 planes"])
```
Current result

FAILED_EXECUTION: (UserError) Either 'conversation' or individual inputs must be provided. 'ground_truth' is missing.
One recommendation: ask for expected_output in the public API, not ground_truth, because that already exists on the agent path and makes the feature a parity improvement rather than a new concept.

### Code Sample
Keep the existing API for backward compatibility:
```markdown
queries=["q1", "q2"]
```
But also support a structured form like:
```markdown
queries=[
    {
        "query": "How many plans does Netlife offer in Ecuador in B1 2025?",
        "ground_truth": "Ofrece 41 planes"
    }
]
```

### Language/SDK

Both

alliscode:
This pull request introduces support for passing a ground-truth (expected output) value for workflow evaluations, enabling reference-based evaluators such as Foundry's Similarity evaluator. The changes allow users to specify an expected output that is stamped onto the overall evaluation item and surfaced as `ground_truth` in the JSONL payload sent to the evaluation API. The PR also adds validation to ensure that reference-based evaluators are only used when an expected output is provided, and updates the relevant tests and documentation. Additionally, a new sample demonstrates how to use this feature.

Closes #5135

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.Foundry/Evaluation/FoundryEvals.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Foundry/Evaluation/FoundryEvals.cs
@@ -146,6 +146,7 @@ public sealed class FoundryEvals : IAgentEvaluator
         bool hasContext = payloads.Any(p => p.Context is not null);
         bool hasTools = payloads.Any(p => p.ToolDefinitions is { Count: > 0 });
         bool hasGroundTruth = payloads.Any(p => p.GroundTruth is not null);
+        bool allHaveGroundTruth = payloads.Count > 0 && payloads.All(p => p.GroundTruth is not null);
 
         // Filter out tool evaluators if no items have tools; auto-add ToolCallAccuracy if tools present
         var evaluators = FilterToolEvaluators(this._evaluatorNames, hasTools);
@@ -154,15 +155,18 @@ public sealed class FoundryEvals : IAgentEvaluator
             evaluators = [.. evaluators, ToolCallAccuracy];
         }
 
-        // Fail fast if a ground-truth evaluator (e.g. similarity) is requested but no
-        // item carries an ExpectedOutput, instead of surfacing a remote provider error.
-        var missingGroundTruth = FoundryEvalConverter.FindMissingGroundTruthEvaluators(evaluators, hasGroundTruth);
+        // Fail fast if a ground-truth evaluator (e.g. similarity) is requested but not
+        // every item carries an ExpectedOutput. Reference-based evaluators score each
+        // item against its own ground truth, so even one missing value will surface as
+        // a provider-side validation error. Catch it here with a clearer message.
+        var missingGroundTruth = FoundryEvalConverter.FindMissingGroundTruthEvaluators(evaluators, allHaveGroundTruth);
         if (missingGroundTruth.Count > 0)
         {
             throw new InvalidOperationException(
-                "The following evaluator(s) require a ground-truth/expected output but none of the items " +
-                $"provided an {nameof(EvalItem.ExpectedOutput)}: {string.Join(", ", missingGroundTruth)}. " +
-                "Provide an expected output per item (for example via the 'expectedOutput' parameter on EvaluateAsync).");
+                "The following evaluator(s) require a ground-truth/expected output on every item but " +
+                $"at least one item is missing an {nameof(EvalItem.ExpectedOutput)}: {string.Join(", ", missingGroundTruth)}. " +
+                "Provide an expected output per item (for example via the 'expectedOutput' parameter on EvaluateAsync), " +
+                "or set 'includePerAgent: false' so the evaluator only runs on the overall item.");
         }
 
         // 2. Create the evaluation definition
--- a/dotnet/src/Microsoft.Agents.AI.Foundry/Evaluation/FoundryEvals.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Foundry/Evaluation/FoundryEvals.cs
@@ -834,15 +834,15 @@ public sealed class FoundryEvals : IAgentEvaluator
         var result = new EvalItemResult(itemId, status, scores);
 
         // Extract error info from sample
-        if (outputItem.TryGetProperty("sample", out var sample))
+        if (outputItem.TryGetProperty("sample", out var sample) && sample.ValueKind == JsonValueKind.Object)
         {
-            if (sample.TryGetProperty("error", out var errObj))
+            if (sample.TryGetProperty("error", out var errObj) && errObj.ValueKind == JsonValueKind.Object)
             {
                 result.ErrorCode = errObj.TryGetProperty("code", out var code) ? code.GetString() : null;
                 result.ErrorMessage = errObj.TryGetProperty("message", out var msg) ? msg.GetString() : null;
             }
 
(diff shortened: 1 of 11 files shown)


dotnet/src/Microsoft.Agents.AI.Workflows/Evaluation/WorkflowEvaluationExtensions.cs
```diff
@@ -48,28 +60,25 @@ public static async Task<AgentEvaluationResults> EvaluateAsync(
...
+                overallItems.Add(overallItem);
+            }
+            else if (expectedOutput is not null)
+            {
+                // The user supplied a ground truth but we couldn't find a final response to
+                // compare against — almost always because the workflow's agents weren't built
+                // with EmitAgentResponseEvents enabled (so no AgentResponseEvent was emitted)
+                // and no terminal ExecutorCompletedEvent carried an AgentResponse / ChatMessage
+                // / string payload. Fail loudly instead of silently returning 0/0.
+                throw new InvalidOperationException(
```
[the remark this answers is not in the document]
alliscode: Fixed, thanks.

dotnet/src/Microsoft.Agents.AI.Workflows/Evaluation/WorkflowEvaluationExtensions.cs
```diff
@@ -36,6 +47,7 @@ public static async Task<AgentEvaluationResults> EvaluateAsync(
         bool includeOverall = true,
         bool includePerAgent = true,
         string evalName = "Workflow Eval",
+        string? expectedOutput = null,
         IConversationSplitter? splitter = null,
         CancellationToken cancellationToken = default)
```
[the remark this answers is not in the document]
alliscode: Fixed, thanks.

dotnet/src/Microsoft.Agents.AI.Foundry/Evaluation/FoundryEvals.cs
```diff
@@ -145,6 +145,7 @@ public async Task<AgentEvaluationResults> EvaluateAsync(
 
         bool hasContext = payloads.Any(p => p.Context is not null);
         bool hasTools = payloads.Any(p => p.ToolDefinitions is { Count: > 0 });
+        bool hasGroundTruth = payloads.Any(p => p.GroundTruth is not null);
 
         // Filter out tool evaluators if no items have tools; auto-add ToolCallAccuracy if tools present
```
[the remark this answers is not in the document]
alliscode: Fixed, thanks.

dotnet/src/Microsoft.Agents.AI.Workflows/Evaluation/WorkflowEvaluationExtensions.cs
```diff
@@ -97,6 +106,80 @@ public static async Task<AgentEvaluationResults> EvaluateAsync(
...
+                    break;
+                case ChatMessage cm:
+                    responseText = cm.Text ?? string.Empty;
+                    conversation.Add(cm);
+                    break;
+                case string s:
+                    responseText = s;
+                    conversation.Add(new ChatMessage(ChatRole.Assistant, s));
+                    break;
+                default:
```
moonbox3, line 172: Above we have `&& completed.Data is AgentResponse or ChatMessage or string` so doesn't that mean this `default` will not get hit? If there every is drift, should an exception type be thrown as it isn't expected?

dotnet/src/Microsoft.Agents.AI.Workflows/Evaluation/WorkflowEvaluationExtensions.cs
```diff
@@ -48,28 +60,25 @@ public static async Task<AgentEvaluationResults> EvaluateAsync(
...
-
-                conversation.AddRange(finalResponse.Response.Messages);
-
-                overallItems.Add(new EvalItem(query, finalResponse.Response.Text, conversation)
-                {
-                    Splitter = splitter,
-                });
+                overallItems.Add(overallItem);
+            }
+            else if (expectedOutput is not null)
```
moonbox3: Why is the throw gated on `expectedOutput` rather than `includeOverall`? Same misconfiguration (agents not bound with `EmitAgentResponseEvents`, no qualifying terminal `ExecutorCompletedEvent`) produces two very different outcomes: with an expected output, a clear actionable exception; without one, an empty `AgentEvaluationResults` and no signal that anything went wrong. The user asked for an overall eval in both cases, so seems like that's the condition that should drive the diagnostic. 

Could we throw (or at least log a warning) whenever `includeOverall` is true and `BuildOverallItem` returns null?

dotnet/src/Microsoft.Agents.AI.Workflows/Evaluation/WorkflowEvaluationExtensions.cs
alliscode: Fixed, thanks.

dotnet/src/Microsoft.Agents.AI.Workflows/Evaluation/WorkflowEvaluationExtensions.cs
alliscode: Fixed, thanks.

moonbox3 (approved):

lokitoth (approved):

PR #5765, opened 2026-05-12
analysis: root cause analysis (af-872bf775)

stempeck:
Root cause analysis for af-872bf775.
See todos/rootcause_analysis.md for synthesized findings and solution.


moonbox3:
Please keep this on-track for specific agent-framework fixes or features, @stempeck.

PR #5800, opened 2026-05-13
.NET: fix: avoid AGUI tool result message id collisions

### LINKED ISSUE #5794: .NET: [Bug]: Regression - Tool Events not being emitted correctly to the front end
### Description

On Microsoft.Agents.AI.Hosting.AGUI.AspNetCore version 1.0.0-preview.260311.1, our copilot kit front end would display each tool call made by the remote agent, as expected. 

Once we update to version 1.3.0-preview.260423.1, those are not shown anymore.

I can't give a lot of information about the code or screenshots unfortunately but here is an AI analysis (unverified)

● I found the root cause. It's a regression introduced in commit ca02146e ("BugFix #3433 ChatClientAgent streaming responses missing messageid") in agent-framework, which shipped in 1.3.0-preview.260423.1 but not in 1.0.0-preview.260311.1.

  What changed

  In Microsoft.Agents.AI.AGUI/Shared/ChatResponseUpdateAGUIExtensions.cs, AsAGUIEventStreamAsync got this new block (lines 343–353):

  string? streamingMessageId = null;
  await foreach (var chatResponse in updates.WithCancellation(cancellationToken).ConfigureAwait(false))
  {
      if (string.IsNullOrWhiteSpace(chatResponse.MessageId))
      {
          streamingMessageId ??= Guid.NewGuid().ToString("N");
          chatResponse.MessageId = streamingMessageId;
      }
      ...

  Note the ??=. A single fallback streamingMessageId is generated once per stream and applied to every chunk whose provider didn't supply a MessageId.

  Why it breaks tool-call rendering

  The same chatResponse.MessageId is used for two semantically different events further down in that method:

  - ToolCallStartEvent.ParentMessageId = chatResponse.MessageId — the assistant message the tool call belongs to (correct usage)
  - ToolCallResultEvent.MessageId = chatResponse.MessageId — the id of the tool-role result message itself, which must be distinct from the assistant message (ToolCallResultEvent.cs:18-19)

  When the provider (Azure OpenAI with FunctionInvokingChatClient round-trips) doesn't set MessageId on the tool-result chunk, both the assistant text message and the tool result message now share the same id. CopilotKit deduplicates by messageId, so the tool result event
  collides with the assistant message and the FE never renders a separate tool-call row. The fix author's own unit test (AGUIStreamingMessageIdTests.cs) explicitly asserts "All events should share the same generated MessageId" — they didn't catch that this includes tool result
  events.

  In 1.0.0-preview.260311.1 the MessageId stayed null, so CopilotKit assigned each event its own id and the rows rendered correctly.

  Options

  1. Stay on 1.0.0-preview.260311.1 (current committed state) — keeps tool-call rendering. Trade-off: the bug the new version was trying to fix (text content silently dropped when LLM provider returns null MessageId) is back. For Azure OpenAI in Warlock we've never observed that
  in practice, since Azure OpenAI does set MessageIds.
  2. File an upstream bug on agent-framework — the fix should generate a distinct fallback id per logical message boundary (e.g. flip to a new id when the chunk contains FunctionResultContent, or use the FunctionResultContent.CallId).
  3. Workaround at our IChatClient layer — inject a decorator that assigns distinct MessageIds before chunks reach the AGUI converter. Adds complexity for an upstream bug.

  My recommendation: drop the working-tree bump back to 1.0.0 (option 1) and file an upstream issue. The bug ca02146e was fixing doesn't affect us (Azure OpenAI populates MessageId), and the regression is real. Want me to draft the upstream issue?

### Code Sample

```markdown

```

```markdown

```

### Package Versions

Microsoft.Agents.AI.Hosting.AGUI.AspNetCore 1.3.0-preview.260423.1

### .NET Version

NET 10

### Additional Context

_No response_

he-yufeng:
﻿## Summary
- generate a separate fallback message id for tool-result-only updates
- keep assistant text and tool-call parent events on the shared assistant fallback id
- add a regression test covering null provider MessageId with text, tool call, and tool result events

Fixes #5794.

## To verify
- `dotnet test --project .\tests\Microsoft.Agents.AI.AGUI.UnitTests -f net10.0 --filter-query "/*/Microsoft.Agents.AI.AGUI.UnitTests/AGUIStreamingMessageIdTests/ToolResults_NullMessageId_GeneratesDistinctMessageIdAsync"`
- `dotnet build .\src\Microsoft.Agents.AI.AGUI -f net10.0 --no-restore --tl:off`
- `dotnet build .\src\Microsoft.Agents.AI.Hosting.AGUI.AspNetCore -f net10.0 --no-restore --tl:off`
- `dotnet test --project .\tests\Microsoft.Agents.AI.AGUI.UnitTests -f net10.0 --no-restore`
- `dotnet run --project .\tests\Microsoft.Agents.AI.Hosting.AGUI.AspNetCore.UnitTests -f net10.0 --`
- `dotnet format .\src\Microsoft.Agents.AI.AGUI\Microsoft.Agents.AI.AGUI.csproj --verify-no-changes --no-restore`
- `dotnet format .\tests\Microsoft.Agents.AI.AGUI.UnitTests\Microsoft.Agents.AI.AGUI.UnitTests.csproj --verify-no-changes --no-restore`
- `dotnet format .\src\Microsoft.Agents.AI.Hosting.AGUI.AspNetCore\Microsoft.Agents.AI.Hosting.AGUI.AspNetCore.csproj --verify-no-changes --no-restore`

Note: `dotnet test --project .\tests\Microsoft.Agents.AI.Hosting.AGUI.AspNetCore.UnitTests -f net10.0 --no-restore` reports the repository-level MTP/VSTest runner mismatch for that project, so I ran it through `dotnet run --project ... -f net10.0 --` instead.


he-yufeng:
Addressed the mixed-content tool result case in 501cae371. Fallback IDs now split whenever an update contains a FunctionResultContent, not only when it contains function results exclusively, so a tool-role update with TextContent + FunctionResultContent gets a distinct message id and the emitted tool-role text shares the tool result id.\n\nValidation:\n\n`powershell\ndotnet build dotnet\\tests\\Microsoft.Agents.AI.AGUI.UnitTests\\Microsoft.Agents.AI.AGUI.UnitTests.csproj --no-restore\ndotnet dotnet\\tests\\Microsoft.Agents.AI.AGUI.UnitTests\\bin\\Debug\\net10.0\\Microsoft.Agents.AI.AGUI.UnitTests.dll --filter-method '*ToolResults_NullMessageId_GeneratesDistinctMessageIdAsync' --filter-method '*ToolResults_WithTextContent_GeneratesDistinctMessageIdAsync' --filter-method '*ToolCalls_EmptyMessageId_GeneratesFallbackParentMessageIdAsync' --no-progress\ndotnet format dotnet\\agent-framework-dotnet.slnx --verify-no-changes --no-restore --include dotnet\\src\\Microsoft.Agents.AI.AGUI\\Shared\\ChatResponseUpdateAGUIExtensions.cs dotnet\\tests\\Microsoft.Agents.AI.AGUI.UnitTests\\AGUIStreamingMessageIdTests.cs\ngit diff --check\n`\n\nResult: build passed with 0 warnings/errors, 3 targeted tests passed, format and diff checks passed.

he-yufeng:
Rebased this branch onto current main and force-pushed with lease. Re-ran the focused AG-UI validation after the rebase: AGUI unit-test project build passed with 0 warnings/errors, the three message-id regression tests passed, dotnet format --verify-no-changes passed for the touched files, and git diff --check passed.

rogerbarreto (approved):

rogerbarreto (approved):

peibekwe (approved):

PR #5866, opened 2026-05-14
Python: fix: reject service storage history conflict

### LINKED ISSUE #5798: Python: OpenAI store=True can silently bypass external HistoryProvider persistence
### Summary

When using the Python SDK with `OpenAIChatClient`, an external `HistoryProvider` such as `CosmosHistoryProvider`, and `require_per_service_call_history_persistence=True`, external history can be silently skipped unless callers explicitly set `store=False` in `ChatOptions`.

This is surprising because the agent is configured with an external history provider and asks for per-service-call persistence, but no provider rows are written and no warning/error is surfaced.

### Repro shape

```python
from agent_framework import Agent, AgentSession, ChatOptions
from agent_framework_azure_cosmos import CosmosHistoryProvider

history = CosmosHistoryProvider(...valid endpoint/db/container/credential...)
client = OpenAIChatClient(api_key=..., model="gpt-5.5-pro")

async with history:
    async with Agent(
        client=client,
        name="HistoryRepro",
        instructions="Answer briefly.",
        context_providers=[history],
        require_per_service_call_history_persistence=True,
        default_options=ChatOptions(
            model="gpt-5.5-pro",
            max_tokens=512,
            reasoning={"effort": "high"},
            # store=False fixes the issue. Omitting it writes no external rows.
        ),
    ) as agent:
        session = AgentSession(session_id="debug-openai-history")
        async for _ in agent.run("Remember BLUE-COSMOS-913. Reply ok.", stream=True, session=session):
            pass
```

Observed without `store=False`:

- The model call succeeds.
- A second turn with the same in-process session may recall successfully.
- `CosmosHistoryProvider` writes zero rows.
- There is no warning that external history was bypassed.

Observed with `store=False`:

- Cosmos rows are written as expected.
- A second turn can recall from SDK/Cosmos history.

### Suspected cause

`OpenAIChatClient.STORES_BY_DEFAULT=True`. In `Agent._resolve_per_service_call_history_providers(...)`, service-side storage means no per-service-call history middleware is installed. Later, `_run_after_providers(...)` still skips `HistoryProvider` instances because `require_per_service_call_history_persistence` is true. The combination results in no external history writes.

### Expected behavior

At least one of these should happen:

1. `require_per_service_call_history_persistence=True` plus external `HistoryProvider` should force local/external persistence or set `store=False` for OpenAI automatically.
2. The SDK should raise a configuration error when OpenAI service-side storage and external per-service-call history persistence conflict.
3. The SDK should log a clear warning that external HistoryProvider persistence is disabled because service-side OpenAI storage is active.

Silent zero-row external history is hard to diagnose in production.

### Workaround

Set `store=False` in OpenAI `ChatOptions` when external SDK history is the source of truth.

Thanks.

he-yufeng:
﻿Fixes #5798.

## Summary
- reject `require_per_service_call_history_persistence=True` when the effective chat options still use service-side history storage
- point users to `store=False` so the configured `HistoryProvider` owns per-service-call persistence
- keep the existing service-storage session-handle tests, but separate them from external history-provider persistence

## Tests
- `python -m py_compile python\packages\core\agent_framework\_agents.py python\packages\core\tests\core\test_agents.py`
- `uv run ruff check packages\core\agent_framework\_agents.py packages\core\tests\core\test_agents.py`
- `uv run pytest packages\core\tests\core\test_agents.py -q -k "per_service_call_persistence or service_storage_updates_session_handle" --basetemp .tmp\pytest -p no:cacheprovider`
- `git diff --check`
- `Select-String -Path python\packages\core\agent_framework\_agents.py,python\packages\core\tests\core\test_agents.py -Pattern 'sk-[A-Za-z0-9_-]+'`

### DIFF
--- a/python/packages/core/agent_framework/_agents.py
+++ b/python/packages/core/agent_framework/_agents.py
@@ -804,7 +804,12 @@ class RawAgent(BaseAgent, Generic[OptionsCoT]):  # type: ignore[misc]
             else cast(str | None, (options or {}).get("conversation_id") or self.default_options.get("conversation_id"))
         )
         if service_stores_history:
-            return []
+            raise AgentInvalidRequestException(
+                "require_per_service_call_history_persistence cannot be used "
+                "when the chat client is configured to store history. Set "
+                "store=False to persist each service call through the configured "
+                "HistoryProvider."
+            )
 
         if conversation_id is not None:
             raise AgentInvalidRequestException(
--- a/python/packages/core/tests/core/test_agents.py
+++ b/python/packages/core/tests/core/test_agents.py
@@ -607,65 +607,36 @@ async def test_streaming_per_service_call_persistence_hides_response_id_from_aft
     assert provider_state["response_ids"] == [None, None]
 
 
-async def test_per_service_call_persistence_uses_real_service_storage_when_client_stores_by_default(
+async def test_per_service_call_persistence_rejects_service_storage_when_client_stores_by_default(
     chat_client_base: SupportsChatGetResponse,
 ) -> None:
     provider = _RecordingHistoryProvider()
-
-    @tool(name="lookup_weather", approval_mode="never_require")
-    def lookup_weather(location: str) -> str:
-        return f"Weather in {location}: sunny"
 
     chat_client_base.STORES_BY_DEFAULT = True  # type: ignore[attr-defined]
 
     session = AgentSession()
     session.state[provider.source_id] = {"messages": []}
-    chat_client_base.run_responses = [
-        ChatResponse(
-            messages=Message(
-                role="assistant",
-                contents=[
-                    Content.from_function_call(
-                        call_id="call_1",
-                        name="lookup_weather",
-                        arguments='{"location": "Seattle"}',
-                    )
-                ],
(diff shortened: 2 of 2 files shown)


he-yufeng:
﻿Synced this branch with the latest `upstream/main` and re-ran the focused checks after the merge.

Verified locally:

```text
uv run --no-sync pytest packages/core/tests/core/test_agents.py -q
uv run --no-sync ruff check packages/core/agent_framework/_agents.py packages/core/tests/core/test_agents.py
uv run --no-sync ruff format --check packages/core/agent_framework/_agents.py packages/core/tests/core/test_agents.py
uv run --no-sync python -m py_compile packages/core/agent_framework/_agents.py packages/core/tests/core/test_agents.py
uv run --no-sync pyright packages/core/agent_framework/_agents.py packages/core/tests/core/test_agents.py
git diff --check upstream/main...HEAD
```

All passed. The pytest run only emitted an existing local Windows `.pytest_cache` permission warning; the test file passed.

eavanvalkenburg:
Closing this, I build a more complete fix in #6310

PR #5867, opened 2026-05-14
Python: avoid MCP header injection on redirects

### LINKED ISSUE #5820: Python: [Bug]: `MCPStreamableHTTPTool` re-injects `header_provider` headers on cross-origin redirects (latent until #4808 / mcp-sdk#2298 ship)
## Summary

I think `MCPStreamableHTTPTool` leaks `header_provider` headers (including bearer tokens) on cross-origin redirects, because the `_inject_headers` event hook re-applies them after httpx has already stripped them. The bug is latent on the current PyPI `mcp` release (the `_mcp_call_headers` ContextVar doesn't propagate to the MCP SDK's `post_writer` task, per #4808), but it becomes reachable as soon as the fix in modelcontextprotocol/python-sdk#2298 (merged into `mcp` main on 2026-03-31) reaches a PyPI release. I'd suggest patching this in agent-framework before that fix ships, not after.

## How I found this

Reading `python/packages/core/agent_framework/_mcp.py` end-to-end and noticing the hook is unconditional while `httpx.AsyncClient` was constructed with `follow_redirects=True`.

## Where

`python/packages/core/agent_framework/_mcp.py`, lines 1547-1564. The auto-client uses `follow_redirects=True`, and `_inject_headers` re-applies every key from `_mcp_call_headers` on every outbound request with no origin check. The first-party sample `samples/02-agents/mcp/mcp_api_key_auth.py` documents this as the recommended way to forward bearer tokens.

This is the same class of bug as CVE-2018-1000007 (curl), CVE-2024-37891 (urllib3), CVE-2023-32681 (requests). httpx itself strips `Authorization` on cross-origin redirects via `_redirect_headers`; the hook reverses that protection.

## Reproduction

I wrote a short script that mirrors the `_mcp.py:1547-1564` block, sets the contextvar in the same task (sidestepping #4808), and points the client at a local stub that returns `302 Location: http://attacker.local/...`. With the hook installed, the bearer is re-injected on the redirect hop. Without the hook, httpx alone does not leak. I also have a docker-free three-process harness (malicious MCP stub, attacker capture, framework client) with CSPRNG sentinels if that's useful for triage.

## Suggested fix

In `_inject_headers`, compare `(scheme, host, port)` of `request.url` against `self.url` and skip injection when they differ (matches httpx `_same_origin`). I have a patch and a regression test against a real httpx redirect flow (not a mocked `session.call_tool`) ready to PR if you want it.

## Versions

`agent-framework-core` 1.3.0, `mcp` 1.27.1, `httpx` 0.28.1, Python 3.13.

If you'd prefer to route this through `secure@microsoft.com` / MSRC given the credential-leak shape, happy to do that; otherwise I'll send the PR against `main`.

— Matteo Panzeri (matte1782)

he-yufeng:
﻿Fixes #5820.

## Summary
- only inject MCP header_provider headers for requests that stay on the configured MCP server origin
- preserve same-origin header injection for the existing streamable HTTP client path
- add a real httpx redirect regression test that verifies Authorization is not re-applied to a cross-origin redirect target

## To verify
- `python -m py_compile packages\\core\\agent_framework\\_mcp.py packages\\core\\tests\\core\\test_mcp.py`
- `uv run ruff check packages\\core\\agent_framework\\_mcp.py packages\\core\\tests\\core\\test_mcp.py`
- `uv run pytest packages\\core\\tests\\core\\test_mcp.py -q -k "header_provider_with_httpx_event_hook or header_provider_skips_cross_origin_redirect or header_provider_with_user_httpx_client" --basetemp .tmp\\pytest -p no:cacheprovider`
- `git diff --check`

### DIFF
--- a/python/packages/core/agent_framework/_mcp.py
+++ b/python/packages/core/agent_framework/_mcp.py
@@ -1592,7 +1592,7 @@ class MCPStreamableHTTPTool(MCPTool):
         Returns:
             An async context manager for the streamable HTTP client transport.
         """
-        from httpx import AsyncClient, Request, Timeout
+        from httpx import URL, AsyncClient, Request, Timeout
 
         http_client = self._httpx_client
         if self._header_provider is not None:
@@ -1604,8 +1604,17 @@ class MCPStreamableHTTPTool(MCPTool):
                 self._httpx_client = http_client
 
             if not hasattr(self, "_inject_headers_hook"):
+                mcp_origin = URL(self.url)
+
+                def _origin(url: URL) -> tuple[str, str, int | None]:
+                    port = url.port
+                    if port is None:
+                        port = {"http": 80, "https": 443}.get(url.scheme)
+                    return (url.scheme, url.host or "", port)
 
                 async def _inject_headers(request: Request) -> None:  # noqa: RUF029
+                    if _origin(request.url) != _origin(mcp_origin):
+                        return
                     headers = _mcp_call_headers.get({})
                     for key, value in headers.items():
                         request.headers[key] = value
--- a/python/packages/core/tests/core/test_mcp.py
+++ b/python/packages/core/tests/core/test_mcp.py
@@ -4497,6 +4497,49 @@ async def test_mcp_streamable_http_tool_header_provider_with_httpx_event_hook():
             await tool._httpx_client.aclose()
 
 
+async def test_mcp_streamable_http_tool_header_provider_skips_cross_origin_redirect():
+    """header_provider headers should not be re-applied to redirected origins."""
+    import httpx
+
+    from agent_framework._mcp import _mcp_call_headers
+
+    captured: list[tuple[str, httpx.Headers]] = []
+
+    async def handler(request: httpx.Request) -> httpx.Response:
+        captured.append((str(request.url), request.headers.copy()))
+        if request.url.host == "example.com":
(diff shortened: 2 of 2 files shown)


he-yufeng:
﻿Synced this branch with the latest `upstream/main` and re-ran the focused checks after the merge.

Verified locally:

```text
uv run --no-sync pytest packages/core/tests/core/test_mcp.py -q
uv run --no-sync ruff check packages/core/agent_framework/_mcp.py packages/core/tests/core/test_mcp.py
uv run --no-sync ruff format --check packages/core/agent_framework/_mcp.py
uv run --no-sync python -m py_compile packages/core/agent_framework/_mcp.py packages/core/tests/core/test_mcp.py
uv run --no-sync pyright packages/core/agent_framework/_mcp.py packages/core/tests/core/test_mcp.py
git diff --check upstream/main...HEAD
```

All passed. I did not apply unrelated formatting churn from an existing `upstream/main` line outside this PR's diff.

eavanvalkenburg:
Thank you so much for the contribution and for helping move the MCP reliability work forward.

The team consolidated the overlapping core MCP changes into #6145 so we can review and merge them in one place: https://github.com/microsoft/agent-framework/pull/6145

I'm closing this PR in favor of #6145, with many thanks from the team for your work.

PR #5884, opened 2026-05-15
Python (anthropic): Migrate structured outputs to GA `output_config.format`

t-anjan:
## Summary

`RawAnthropicClient._prepare_options` forwards `response_format` as the deprecated beta parameter `output_format={"type": "json_schema", "schema": {...}}` plus the beta flag `structured-outputs-2025-11-13`. When the same request also includes `tools`, Claude emits concatenated / malformed JSON — e.g. three copies of the schema's empty default like `{"matches":[]}{"matches":[]}{"matches":[]}` — instead of populating the schema. Anthropic's GA shape — `output_config={"format": {"type":"json_schema","schema":{...}}}` — works correctly with tools.

This PR migrates the provider to the GA shape.

## Verified empirically

Observed on a structured-output workload that combined `response_format` (a Pydantic model) with a tool (`run_shell`) against `claude-sonnet-4-6`: the deprecated path produced the malformed concatenated output above, the GA path did not.

## Why no beta flag is needed on the GA path

`AsyncMessages.create()` on `beta.messages` (which is what MAF dispatches through — `_chat_client.py:541, 550`) accepts `output_config` as a top-level parameter and does **not** require `structured-outputs-2025-11-13`. Only `beta.messages.parse(...)` auto-injects a structured-outputs beta flag (and even then it injects `structured-outputs-2025-12-15`, not the `2025-11-13` flag MAF was sending). So `create()` with `output_config` is genuinely flag-free.

Verified against installed `anthropic==0.80.0` — both `BetaOutputConfigParam` and `OutputConfigParam` expose exactly `{format: JSONOutputFormatParam, effort: Optional[Literal["low","medium","high","max"]]}`.

## Changes

- `python/packages/anthropic/pyproject.toml` and `python/uv.lock`: widen `anthropic` to `>=0.80.0,<0.117.0` and refresh the lock to `0.116.0`.
- `python/packages/anthropic/agent_framework_anthropic/_chat_client.py`:
  - Move `response_format` into `run_options["output_config"]["format"]` (GA shape) and stop adding the `structured-outputs-2025-11-13` beta flag.
  - **Merge** the format into any caller-supplied `output_config` so e.g. `output_config["effort"]` (adaptive-thinking effort level) survives the transformation.
  - Drop the now-unused `STRUCTURED_OUTPUTS_BETA_FLAG` constant (private to this module — no external callers).
  - `_prepare_response_format` keeps the same `{"type": "json_schema", "schema": ...}` return shape; the docstring is updated to point at the GA target.

## Test plan

`uv run pytest packages/anthropic/tests` → 130 passed.
`UV_PYTHON=3.11 uv run python scripts/workspace_poe_tasks.py check -P anthropic` → Anthropic package fmt/lint/type checks passed; 143 tests passed.
Full local `ci-test-typing` still has unrelated package-resolution failures outside Anthropic (`agent_framework_tools`, Azure Content Understanding), while Anthropic passed in that fan-out.

New tests:
- `test_prepare_options_uses_output_config_for_response_format` — the GA `output_config.format` shape is emitted, the deprecated `output_format` key is not, and `structured-outputs-2025-11-13` is not added to `betas`.
- `test_prepare_options_preserves_caller_supplied_output_config_effort` — a caller-supplied `output_config["effort"]` survives the merge.
- `test_prepare_options_no_response_format_omits_output_config` — no `output_config` is added implicitly when `response_format` is absent.

### DIFF
--- a/python/packages/anthropic/agent_framework_anthropic/_chat_client.py
+++ b/python/packages/anthropic/agent_framework_anthropic/_chat_client.py
@@ -79,7 +79,6 @@ logger = logging.getLogger("agent_framework.anthropic")
 
 ANTHROPIC_DEFAULT_MAX_TOKENS: Final[int] = 1024
 BETA_FLAGS: Final[list[str]] = ["mcp-client-2025-04-04", "code-execution-2025-08-25"]
-STRUCTURED_OUTPUTS_BETA_FLAG: Final[str] = "structured-outputs-2025-11-13"
 
 ResponseModelT = TypeVar("ResponseModelT", bound=BaseModel | None, default=None)
 AnthropicAsyncClient = AsyncAnthropic | AsyncAnthropicBedrock | AsyncAnthropicFoundry | AsyncAnthropicVertex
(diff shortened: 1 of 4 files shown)


moonbox3 (approved):

moonbox3:
@t-anjan sorry it took so long to get to this: I noticed we are pinning the anthropic package to `anthropic>=0.80.0,<0.80.1`. The latest version is now 0.116.0. We should allow for this latest package. It can either be done in this PR, or in a separate on. I created issue #6986 to track this if not done in this PR. Thanks.

moonbox3:
@t-anjan can you please also have a look at the failing CI/CD checks, thanks.

eavanvalkenburg (approved):

eavanvalkenburg (approved):

PR #5906, opened 2026-05-16
Python: Fix empty final output after tool calls

### LINKED ISSUE #5726: Python: [Bug]: Final output is empty after successful tool execution in declarative agent run
## Description

I encountered an issue where the declarative agent successfully generated and executed the expected Python solution via `python_executor`, but the final assistant response was empty, resulting in an empty `final_output`.

The the tool execution is succeeded. However, after the tool result was returned to the model, the final `chat_output` contained an assistant message with empty `contents`, and the recorded `final_output` was `""`. :contentReference[oaicite:0]{index=0}

## Environment / Config

- Agent config: 
```
instructions: |
  You are a Python coding assistant. Your task is to complete the following Python code from the HumanEval benchmark.
  Use tools when they help, and give concise, evidence-aware answers.
  Note:
    1. Output only final valid Python code. The final answer must be executable Python code.

model:
  connection:
    kind: key
  options:
    temperature: 0.0
    maxOutputTokens: 512
    allowMultipleToolCalls: false
    chatToolMode: auto
tools:
  - kind: function
    name: python_executor
    description: Execute Python code in a short-lived subprocess.
    bindings:
      - name: python_executor
```

## What Happened
* The model first generated a valid implementation and called python_executor
* The tool execution succeeded:
```
{
  "ok": true,
  "stdout": "1.0\n",
  "stderr": "",
  "returncode": 0,
  "timed_out": false
}
```
* After that, the model was called again with the tool result. The final response was:
```
{
  "role": "assistant",
  "contents": []
}
```
* The run ended with:
```
"finish_reason": "stop",
"final_output": ""
```
## Expected Behavior
After successful tool execution, the agent should return the final valid Python code.

### Code Sample

```markdown

```

```markdown

```

### Package Versions

agent-framework 1.3.0, agent-framework-declarative 1.0.0b260507

### Python Version

Python 3.14.4

### Additional Context

_No response_

VIVEKJAINDTU:
### Motivation and Context

After a successful tool call, some providers return a final assistant turn with no text, tool calls, or reasoning payload. The conversation shape looks like:

- assistant → function call
- tool → function result
- assistant → empty contents

In that case response.text stays empty even though tools ran correctly, which breaks callers that rely on the final assistant message.

Fixes: #5726
### Description

- Skip empty assistant messages in the Chat Completions parser.
- In the function-invocation loop, if tools ran but the final text is empty, make one more call with tool_choice="none" to get a text response.
- Skip this fallback when middleware terminates the loop early (should_terminate).

### Tests

Added test_base_client_with_function_calling_recovers_from_empty_terminal_response.

### Contribution Checklist


VIVEKJAINDTU:

python/packages/core/agent_framework/_tools.py
```diff
@@ -2436,6 +2440,38 @@ async def _get_response() -> ChatResponse[Any]:
                         execute_function_calls=execute_function_calls,
                     )
                     if result.get("action") == "return":
+                        # If tool calls happened but the terminal assistant payload is empty,
+                        # force one non-tool model turn to obtain a final textual answer.
+                        if fcc_messages and not response.text.strip() and not result.get("should_terminate"):
```
moonbox3, line 2445: Should this check whether the response has any user-visible content, rather than only text? A provider can return a valid non-text assistant payload after tool execution, such as `data`, `uri`, or a hosted file, where `response.text` is empty but the response is not blank. This branch would discard that payload and issue another no-tool model call, so multimodal or file outputs can be lost after successful tool execution.

moonbox3 (changes requested):
Thanks for taking this on. I think the underlying issue is still real: current `main` can still return an empty final response when a successful tool call is followed by a terminal assistant message with no contents.

That said, I don’t think this is quite the right fix as-is. The fallback should be scoped to genuinely blank terminal responses, not just responses with empty `text`, since valid non-text assistant payloads can also have `response.text == ""`. It may be worth aligning this with the visible-content check used by the function-invocation limit fallback so we preserve multimodal/file outputs while still recovering from truly empty terminal turns.

moonbox3:
Please re-open when ready to address and move forward.

PR #5923, opened 2026-05-18
Python: fix(core): restrict unpickler module-prefix allowlist to types only

### LINKED ISSUE #7432: Python: Restrict checkpoint unpickler globals and attribute traversal
Restricted checkpoint decoding currently permits non-type globals through auto-allowed module prefixes. The built-in getattr reconstruction path can also traverse attributes beyond the intended type allowlist. Together, these behaviors can expose unintended callables while deserializing a crafted checkpoint.

Expected behavior: restricted decoding accepts only permitted types, explicitly permitted nested types, and narrowly supported enum-member reconstruction. Non-type globals and unrelated attribute traversal should raise an unpickling error.

Addressed by #5923.

White-Mouse:
### Motivation & Context

The restricted checkpoint decoder allows framework and OpenAI SDK types by module prefix. The previous implementation returned any global from those modules, and the allowlisted `builtins.getattr` could also traverse attributes on an allowed type. Together, those behaviors allowed a crafted pickle to reach a non-type callable such as `_RestrictedUnpickler.__init__.__globals__["pickle"].loads` before the post-deserialization type check.

This change tightens that defense-in-depth path while preserving the checkpoint format and supported type reconstruction.

- **What are the major changes?** Package-prefix and caller-supplied allowlist entries must resolve to a `type`. During restricted unpickling, `builtins.getattr` is replaced with a resolver that only permits type-to-type attribute lookup. Regression tests cover non-type globals, legitimate nested-type lookup, and the full `getattr`/`__globals__`/inner-`pickle.loads` chain.
- **What is the impact of these changes?** Valid built-in, framework, OpenAI SDK, explicitly allowed, and nested type reconstruction remains supported. Non-type globals and non-type attribute traversal are rejected with `pickle.UnpicklingError`.
- **What do you want reviewers to focus on?** Please review the compatibility boundary of the restricted `getattr` replacement and whether the regression coverage captures the expected nested-type reconstruction cases.

### Related Issue

Fixes #7432

### Contribution Checklist


White-Mouse:

moonbox3:
@White-Mouse thanks for your contribution. Is there an open issue we can link to? Additionally, please help to fix the merge conflict.

python/packages/core/agent_framework/_workflows/_checkpoint_encoding.py
```diff
@@ -96,17 +96,36 @@ def __init__(self, data: bytes, allowed_types: frozenset[str]) -> None:
...
+
+        if module.startswith(_FRAMEWORK_MODULE_PREFIX) or module.startswith(
+            _OPENAI_MODULE_PREFIX
+        ):
+            resolved = super().find_class(module, name)  # nosec
+            # Reject non-type globals from allowed package prefixes.  A broad
+            # module-prefix allowlist returns arbitrary module attributes such
+            # as functions and sub-modules, which can be combined with
+            # builtins.getattr to call unrestricted pickle.loads.
+            if isinstance(resolved, type):
```
moonbox3, line 206: Could this still be bypassed through attributes on an allowed framework type? A crafted checkpoint can resolve `_RestrictedUnpickler` through this branch, use the still-allowed `builtins.getattr` to reach `__init__.__globals__["pickle"].loads`, and execute an inner unrestricted pickle while returning an `int` that passes `_verify_type`. Should we avoid the broad prefix type allowlist here, or block dunder/globals traversal before relying on this as a restricted-unpickler fix?

moonbox3: I see this is addressed.

python/packages/core/agent_framework/_workflows/_checkpoint_encoding.py
```diff
@@ -112,6 +113,19 @@
 })
 
 
+def _restricted_getattr(obj: Any, name: str) -> type:
+    """Resolve type-valued attributes needed for legitimate pickle reconstruction."""
+    if not isinstance(obj, type) or not isinstance(name, str):
+        raise pickle.UnpicklingError("Checkpoint deserialization blocked for unsafe attribute traversal.")
+
+    resolved = getattr(obj, name)
```
moonbox3: Should this re-check the resolved nested type against the same allowlist before returning it? Protocol 2/3 pickle represents nested classes as `builtins.getattr(Outer, "Inner")`, so listing only `Outer` in `allowed_types` lets a checkpoint resolve and instantiate `Outer.Inner` without listing `Outer.Inner`. I reproduced this against the PR head with `allowed_types={"__main__:SafeAllowed"}`: a protocol-3 payload for `SafeAllowed.Evil()` reaches this branch, runs `Evil.__setstate__`, and then passes `_verify_type` when the marker names `SafeAllowed.Evil`. Could `_restricted_getattr` require `_type_to_key(resolved)` to be built-in/framework/OpenAI/explicitly allowed too, so nested reconstruction only succeeds for the nested type that was actually allowed?

python/packages/core/agent_framework/_workflows/_checkpoint_encoding.py
```diff
@@ -137,15 +144,52 @@ def __init__(self, data: bytes, allowed_types: frozenset[str]) -> None:
...
+            or resolved.__module__.startswith(_FRAMEWORK_MODULE_PREFIX)
+            or resolved.__module__.startswith(_OPENAI_MODULE_PREFIX)
+        )
+
+    def _restricted_getattr(self, obj: Any, name: str) -> type:
+        """Resolve an allowlisted nested type needed for pickle reconstruction."""
+        if not isinstance(obj, type) or not isinstance(name, str):
+            raise pickle.UnpicklingError("Checkpoint deserialization blocked for unsafe attribute traversal.")
+
+        resolved = getattr(obj, name)
```
moonbox3, line 162: Could we preserve enum-by-name checkpoints here? `enum.pickle_by_enum_name` reconstructs a member as `getattr(Color, "RED")`, so this type-only check now rejects a caller-allowlisted Enum that the base code decodes for protocols 3-5, including protocol 5 written by `encode_checkpoint_value`. Could the wrapper narrowly accept a real member of an already permitted Enum class while keeping other non-type attributes blocked?

moonbox3:
Quick ping on this - let me know if you need help getting it across the finish line. Thank you.

python/packages/core/agent_framework/_workflows/_checkpoint_encoding.py
moonbox3: Addressed in 9010c8aa0. Restricted getattr now accepts only an actual member of an already permitted Enum class; unrelated attributes remain blocked. Coverage includes protocols 2-5, the current encoder protocol, unlisted enums, and non-member attributes. The test reducer is also Python 3.10-compatible.

moonbox3 (approved):

eavanvalkenburg (approved):

Issue #5852, opened 2026-05-14
.NET: Investigation: .NET OTel auto-instrumentation convention for Agent Framework


rogerbarreto:
## Objective

Big OTel rock in .NET land. Tribe pattern say: `AddXxxInstrumentation()` on provider, kill switch env var, separate package. Tribe use ASP.NET Core, Azure SDK, EF Core. Same shape.

Agent Framework today: must wrap each agent by hand with `UseOpenTelemetry()`. No tribe pattern. No kill switch. No one-line activate.

Need investigate. Not code yet. Lay stone first:

- Where instrumentation live? Core? New `Microsoft.Agents.AI.OpenTelemetry` package? Contrib `OpenTelemetry.Instrumentation.AgentFramework` (open-telemetry tribe own `OpenTelemetry.*` prefix on nuget, can't take)?
- How activate? DI auto-decorate `AIAgent`? Provider builder extension? Both? Idempotent?
- Kill switch env var name? `OTEL_DOTNET_AGENTFRAMEWORK_INSTRUMENTATION_ENABLED`?
- Source name rename? Today `Experimental.Microsoft.Agents.AI`. Tribe say `Microsoft.Agents.AI`.
- Workflows source `Microsoft.Agents.AI.Workflows` same package?
- Long term: move emit into base `AIAgent` BCL `ActivitySource` listener-driven (like `HttpClient`)? Skip wrapper?
- Future: contrib package shim post-GA, post-stable gen-AI semconv?

Land ADR with options + decision before any code. Existing ADR 0003 stay. New ADR add on top.

## Outcome

ADR `docs/decisions/0026-agent-otel-auto-instrumentation.md` proposed, reviewed, accepted. Then implementation issue spin off.

## Refs

- ADR 0003 (current `OpenTelemetryAgent` wrapper)
- https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md
- https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/config.md
- https://www.nuget.org/packages/OpenTelemetry.Instrumentation.AspNetCore

PR #5971, opened 2026-05-20
.NET: [.Net][ADR] OTel BCL-native emission for ChatClientAgent (ADR 0027) + implementation

rogerbarreto:
## What

This PR now contains two ADRs and the implementation for the second one:

* **ADR 0026** (`docs/decisions/0026-agent-otel-auto-instrumentation.md`) — original proposal for a convention-aligned `AddAgentFrameworkInstrumentation()` extension on `TracerProviderBuilder` / `MeterProviderBuilder`. Kept as historical context for the discussion that led to ADR 0027.
* **ADR 0027** (`docs/decisions/0027-agent-otel-bcl-native-emission.md`) — the proposal we landed on, plus its implementation: bare `ChatClientAgent` emits OpenTelemetry natively when a tracer subscribes to `Experimental.Microsoft.Agents.AI`, with no other configuration change.

## Why ADR 0027

ADR 0027 matches the BCL pattern (HttpClient, `System.Net.Http`): subscribing to a known source name via `AddSource(...)` is enough; the library emits natively. No DI container required, no extra extension method, no opt-in beyond the existing source name.

The bare path and the explicit `new OpenTelemetryAgent(agent)` decorator path produce **identical** 2-span output (`invoke_agent` + `chat`). Today's dashboards keep working unchanged.

## How it works

`ChatClientAgent` lazily holds an `OpenTelemetryAgent(this, defaultSource)` in a `_selfTelemetryWrap` field. `RunCoreAsync` (and the streaming counterpart) checks `SuppressSelfTelemetryWrap()` and either:

1. Calls `RunChatClientCoreAsync` directly (suppression path), or
2. Delegates to the self-wrap, which re-enters `RunCoreAsync` exactly once with a per-instance marker stamped on the outer `invoke_agent` activity.

Suppression triggers (any one is enough):

* `ChatClientAgentOptions.UseProvidedChatClientAsIs == true` (existing per-agent opt-out).
* `OpenTelemetryConsts.AgentActivitySource.HasListeners() == false` (no tracer subscribed to the default source). This fast path is **required**: without it, when no listeners exist, no Activity is created, the marker cannot be stamped, and the re-entrant call recurses indefinitely.
* The Activity parent chain contains the per-instance marker (`OwnedInvokeAgentScopeMarker` custom property) whose value is reference-equal to `this`. This handles both the bare path's own re-entry and the explicit decorator path (where the user wraps the agent themselves).

The marker is set in exactly one place: `OpenTelemetryAgent.UpdateCurrentActivity` resolves `this.InnerAgent.GetService<ChatClientAgent>()` and stamps that reference as a custom property. Custom properties are process-local; they are not exported as OTLP span attributes.

Per-instance scoping via `ReferenceEquals` ensures nested sub-agent calls via tools emit their own `invoke_agent` spans (different instance, different marker, no suppression).

## What changed

| File | Change |
|---|---|
| `dotnet/src/Microsoft.Agents.AI/ChatClient/ChatClientAgent.cs` | `_selfTelemetryWrap` field, `SuppressSelfTelemetryWrap`, `EnsureSelfTelemetryWrap`, `RunCoreAsync` split into `RunChatClientCoreAsync` (streaming counterpart same shape). |
| `dotnet/src/Microsoft.Agents.AI/OpenTelemetryAgent.cs` | `UpdateCurrentActivity` stamps the per-instance marker on the outer `invoke_agent` activity. |
| `dotnet/src/Microsoft.Agents.AI/OpenTelemetryConsts.cs` | Added `OwnedInvokeAgentScopeMarker` custom-property key and `AgentActivitySource` static field for the `HasListeners` fast path. |
| `dotnet/tests/Microsoft.Agents.AI.UnitTests/ChatClient/ChatClientAgentOpenTelemetryTests.cs` (new, 16 tests) | Bare-path emission, sub-agent isolation, concurrent first-calls, sensitive-data default, provider-style wrapper, decorator-path no triple emission. |
| `dotnet/tests/Microsoft.Agents.AI.UnitTests/ChatClient/OwnerScopedActivityCapture.cs` (new) | Test helper. Raw `ActivityListener` plus parent-chain marker filter. Makes tests fully parallel-safe against the global `ActivitySource` listener registry without serialization attributes. Lives in the test project only; production users keep using `TracerProvider.AddSource(...)` as today. |
| `dotnet/tests/Microsoft.Agents.AI.UnitTests/OpenTelemetryAgentTests.cs` | One pre-existing test migrated to use `OwnerScopedActivityCapture` so parallel execution stays clean. |
| `docs/decisions/0027-agent-otel-bcl-native-emission.md` (new) | The ADR itself. |

## Validation

| Suite | Result |
|---|---|
| Full `Microsoft.Agents.AI.UnitTests` in parallel (no `[Collection]`) | 1564 pass, 0 fail, 0 skipped |
| `ChatClientAgentOpenTelemetryTests` | 16 pass |
| `OpenTelemetryAgentTests` (existing + migrated) | 44 pass |
| CI-parity `dotnet format --verify-no-changes` (WSL2 + Docker, `mcr.microsoft.com/dotnet/sdk:10.0`) | clean on both changed projects |

## Known trade-offs (documented in the ADR)

* **Metrics-only subscribers (no tracer) bypass the self-wrap.** Required by the `HasListeners` fast path to prevent infinite recursion. Users in that mode must explicitly wrap with `new OpenTelemetryAgent(agent)`.
* **Per-instance cost.** One `OpenTelemetryAgent` allocated lazily per `ChatClientAgent`, held until the agent is collected. Per-instance, not per-call.
* **Custom source with no listeners edge case.** If a user wraps with `OpenTelemetryAgent(agent, "MyCustomSource")` and only the default source has listeners, the inner self-wrap will emit on the default source. Mitigation documented in the ADR.

## Out of scope (follow-up)

* `invoke_agent` span enrichment (system instructions, conversation id, aggregated multi-turn token usage, full chat options on the `invoke_agent` span). Tracked as placeholder ADR 0028.
* Agents without an inner `ChatClientAgent` (`A2AAgent`, `DurableAIAgent`, `WorkflowHostAgent`).

## Refs

* Builds on the investigation in #5852.
* ADR-0003 not superseded.

### DIFF
--- a/docs/decisions/0026-agent-otel-auto-instrumentation.md
+++ b/docs/decisions/0026-agent-otel-auto-instrumentation.md
@@ -1,0 +1,363 @@
+---
+status: proposed
+contact: rogerbarreto
+date: 2026-05-20
+deciders: stephentoub, markwallace-microsoft, rogerbarreto, westey-m
+consulted: 
+informed: 
+---
+
+# Agent Framework OpenTelemetry Auto-Instrumentation
+
+## Context and Problem Statement
+
+Today: telemetry on agent = manual `UseOpenTelemetry()` per agent (see
+[ADR-0003](./0003-agent-opentelemetry-instrumentation.md)). Wrapper pattern
+fine for explicit control. No match for .NET OTel ecosystem convention.
+ASP.NET Core, EF Core, SqlClient = one `AddXxxInstrumentation()` on provider
+builder, done.
+
+Three frictions:
+
+1. No convention entry point. User know `AddAspNetCoreInstrumentation()`. No
+   `AddAgentFrameworkInstrumentation()` exist.
+2. DI scenarios = per-agent ceremony. Many `AsAIAgent()` factories in
+   provider packages (`OpenAI`, `Anthropic`, `Foundry`, `A2A`, etc.) not
+   reached by any DI wrap.
+3. No kill switch. Ops cannot disable Agent Framework telemetry without
+   rebuild. `OTEL_DOTNET_AUTO_<COMPONENT>_INSTRUMENTATION_ENABLED` is the
+   convention.
+
+Need:
+
+- Single call on `TracerProviderBuilder` / `MeterProviderBuilder`. Done.
+- `AsAIAgent()` factories auto-wrap when DI present.
+- Full parity with `UseOpenTelemetry()` knobs (source name, sensitive data).
+  No worse-API trap.
+- Env-var kill switch.
+- Keep `OpenTelemetryAgent` + `UseOpenTelemetry()` as primitives.
+
+## Decision Drivers
+
+- **Convention alignment.** Match `AddXxxInstrumentation()` shape. User
+  reflex wins discovery.
+- **Dep hygiene.** Small + stable deps only. Use
+  `OpenTelemetry.Api.ProviderBuilderExtensions` — the package OTel publishes
+  exactly for library authors who expose provider-builder extensions
+  without SDK pull.
+- **Layer separation.** Core knows zero workflow. Workflow knows zero core
+  internals. Each owns own surface.
+- **Parity with `UseOpenTelemetry()`.** Every knob today reachable through
+  new entry point. No regression.
+- **No surprise.** Explicit user call. No env-var SDK bootstrap (spec
+  forbid).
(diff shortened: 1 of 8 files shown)


rogerbarreto:
Dropping this spike for now

PR #5994, opened 2026-05-21
.NET: Add MCP long-running task support for MCP client tools

peibekwe:
### Description

Adds Microsoft Agent Framework client-side support for the MCP tasks extension.
 
A new `Microsoft.Agents.AI.Mcp` (alpha) package introduces `McpClient.ListAgentToolsWithTaskSupportAsync`, which wraps any MCP tool advertising task support with a task-aware `AIFunction`. The wrapper transparently drives the lifecycle inside the agent's function-calling loop.

### Contribution Checklist


dotnet/src/Microsoft.Agents.AI.Mcp/McpClientTaskExtensions.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        McpTaskOptions effectiveOptions = options ?? new McpTaskOptions();
+
+        IList<McpClientTool> tools = await client.ListToolsAsync(cancellationToken: cancellationToken).ConfigureAwait(false);
+
+        AIFunction[] result = new AIFunction[tools.Count];
+        for (int i = 0; i < tools.Count; i++)
+        {
+            ToolTaskSupport? taskSupport = tools[i].ProtocolTool.Execution?.TaskSupport;
+            if (taskSupport is ToolTaskSupport.Optional or ToolTaskSupport.Required)
```
SergeyMenshykh: I wonder if handling a tool with optional task support as a long-running task brings any value. As an alternative, we could choose not to wrap them now and enable it via an option later if/when needed.

dotnet/src/Microsoft.Agents.AI.Mcp/TaskAwareMcpClientAIFunction.cs
```diff
@@ -0,0 +1,148 @@
...
+        {
+            task = await this._client.CallToolAsTaskAsync(
+                this._inner.Name,
+                arguments,
+                taskMetadata: metadata,
+                progress: null,
+                options: null,
+                cancellationToken: cancellationToken).ConfigureAwait(false);
+        }
+        catch (McpProtocolException ex) when ((int)ex.ErrorCode == -32601)
```
SergeyMenshykh: Can we use the `McpErrorCode.MethodNotFound` const instead?

SergeyMenshykh (approved):

dotnet/src/Microsoft.Agents.AI.Mcp/TaskAwareMcpClientAIFunction.cs
peibekwe: Good call, updated.

dotnet/src/Microsoft.Agents.AI.Mcp/McpClientTaskExtensions.cs
peibekwe: That makes sense. I have updated it to wrap just Required for now. We can easily expose an option for optional tools wrapping if we get feedback.

lokitoth (approved):

PR #5944, opened 2026-05-19
Python: Preserve null arguments during tool invocation

shusingh:
## Summary

- Preserve explicitly provided `null` values when dumping validated tool arguments
- Apply the same behavior to direct `FunctionTool.invoke()` and automatic function calling
- Add regression coverage for required nullable tool parameters

Fixes #5934

## Tests

- `python -m pytest tests/core/test_tools.py::test_tool_invoke_preserves_explicit_null_for_required_nullable_argument tests/core/test_function_invocation_logic.py::test_auto_function_calling_preserves_explicit_null_arguments -q`
- `python -m pytest tests/core/test_tools.py tests/core/test_function_invocation_logic.py -q`
- `python -m ruff check agent_framework/_tools.py tests/core/test_tools.py tests/core/test_function_invocation_logic.py`

PR #5997, opened 2026-05-21
Python: fix: preserve explicit null tool arguments

### LINKED ISSUE #5934: Python: [Bug]: Auto function calling removes null arguments
### Description

### Code Sample

```py
import asyncio
from random import randint

from pydantic import Field


def get_weather(
    unit: Annotated[
        Literal["C", "F"] | None,
        Field(
            "otherwise set it to null."
        ),
    ],
) -> str:


async def main() -> None:
    client = FoundryChatClient(
    )

    agent = Agent(
        client=client,
        name="WeatherAgent",
        tools=[get_weather],
    )

    print(f"Agent: {result}")


if __name__ == "__main__":
    asyncio.run(main())
```

```json
```

```json
```

### Package Versions

agent-framework-core: 1.4.0

### Python Version

Python 3.13.6

### Additional Context

_No response_

he-yufeng:
﻿## Summary
- preserve explicitly provided `null` tool arguments after Pydantic validation
- cover both direct `FunctionTool.invoke()` and automatic function-call execution
- keep the existing `exclude_none` behavior for fields that were not explicitly supplied

Fixes #5934.

## Validation
```powershell
python -m pytest tests/core/test_tools.py::test_tool_invoke_preserves_required_nullable_argument tests/core/test_function_invocation_logic.py::test_base_client_with_function_calling_preserves_null_arguments -q -p no:cacheprovider --basetemp .tmp\pytest-null-args
python -m pytest tests/core/test_tools.py -q -p no:cacheprovider --basetemp .tmp\pytest-tools
python -m ruff check packages/core/agent_framework/_tools.py packages/core/tests/core/test_tools.py packages/core/tests/core/test_function_invocation_logic.py
python -m ruff format --check packages/core/agent_framework/_tools.py packages/core/tests/core/test_tools.py packages/core/tests/core/test_function_invocation_logic.py
python -m py_compile packages/core/agent_framework/_tools.py packages/core/tests/core/test_tools.py packages/core/tests/core/test_function_invocation_logic.py
python -m pyright packages/core/agent_framework/_tools.py packages/core/tests/core/test_tools.py packages/core/tests/core/test_function_invocation_logic.py
```

### DIFF
--- a/python/packages/core/agent_framework/_tools.py
+++ b/python/packages/core/agent_framework/_tools.py
@@ -237,6 +237,14 @@ def _annotation_includes_function_invocation_context(annotation: Any) -> bool:
 ClassT = TypeVar("ClassT", bound="SerializationMixin")
 
 
+def _dump_tool_arguments(model: BaseModel, *, include_none_from: Iterable[str] | None = None) -> dict[str, Any]:
+    arguments = model.model_dump(exclude_none=True)
+    for name in include_none_from if include_none_from is not None else model.model_fields_set:
+        if getattr(model, name, None) is None:
+            arguments[name] = None
+    return arguments
+
+
 class FunctionTool(SerializationMixin):
     """A tool that wraps a Python function to make it callable by AI models.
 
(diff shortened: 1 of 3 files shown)


moonbox3:
Thanks for working on this, but we're going with #5944 - same root cause, but it handles the fix recursively, which matters since exclude_none=True itself recurses (nested cases like options.unit = null would still be dropped here).

Also worth flagging for next time: the getattr(model, name, None) is None check (Copilot caught this too) will re-inject any unknown key as None, since getattr returns None for missing attributes, so extra keys get forwarded as unexpected kwargs. Iterating model.model_fields_set directly avoids that.

Appreciate the effort as the diagnosis was spot on. Closing in favor of #5944.

PR #6047, opened 2026-05-22
.NET: fix inline skill string arguments

### LINKED ISSUE #6020: .NET: [Bug]: Arguments valueKind is String instead of Object when calling AgentInlineSkillScript via functionCall
## Problem Description

When using functionCall to invoke `AgentInlineSkillScript`, the `arguments.valueKind` is always `String` instead of `Object`. I expect it to be `Object` so that the skill can receive structured arguments.

## Current Behavior

Every time the agent generates function call arguments, they are serialized as a **string**, even though my skill expects structured object arguments.

## Environment

- **Model**: minimax-m2.5
- **Deployment**: vLLM
- **Skill Type**: AgentInlineSkillScript

## Expected Behavior

When the agent calls a skill via functionCall, `arguments.valueKind` should be `Object` (structured JSON), not `String` (serialized JSON string).

## Reproduction Steps

1. Define an `AgentInlineSkillScript` that expects object-type arguments.
2. Use the agent with functionCall enabled to invoke the skill.
3. Inspect the generated `arguments` — `valueKind` is `String` instead of `Object`.

## Question

Is this a vLLM inference parameter issue, or a model-side (minimax-m2.5) function calling behavior? How can I make the agent output `valueKind: Object` for skill arguments?

Image
Image

### Code Sample

```markdown

```

```markdown

```

### Package Versions

Microsoft.Agents.* 1.4 1.5 1.6

### .NET Version

.NET 8.0

### Additional Context

_No response_

he-yufeng:
﻿## Summary
- accept inline skill arguments when a provider sends the function arguments as a JSON string containing an object
- keep the existing object-only contract for inline skill arguments
- clone values parsed from the temporary JsonDocument so function invocation does not depend on a disposed document
- add a regression test for string-encoded object arguments

Fixes #6020

## To verify
- dotnet build dotnet\tests\Microsoft.Agents.AI.UnitTests\Microsoft.Agents.AI.UnitTests.csproj --no-restore --tl:off
- dotnet dotnet\tests\Microsoft.Agents.AI.UnitTests\bin\Debug\net10.0\Microsoft.Agents.AI.UnitTests.dll --filter-class Microsoft.Agents.AI.UnitTests.AgentSkills.AgentInlineSkillScriptTests --timeout 2m
- dotnet format dotnet\agent-framework-dotnet.slnx --verify-no-changes --include dotnet\src\Microsoft.Agents.AI\Skills\Programmatic\AgentInlineSkillScript.cs dotnet\tests\Microsoft.Agents.AI.UnitTests\AgentSkills\AgentInlineSkillScriptTests.cs --no-restore --verbosity minimal
- git diff --check

Note: plain dotnet test currently starts through the VSTest path on this Windows SDK install and fails with the Microsoft.Testing.Platform opt-in error before test execution, so I ran the built MTP test assembly directly.

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI/Skills/Programmatic/AgentInlineSkillScript.cs
+++ b/dotnet/src/Microsoft.Agents.AI/Skills/Programmatic/AgentInlineSkillScript.cs
@@ -92,18 +92,50 @@ internal sealed class AgentInlineSkillScript : AgentSkillScript
             return [];
         }
 
-        if (arguments.Value.ValueKind != JsonValueKind.Object)
+        var argumentElement = arguments.Value;
+        using var parsedStringArguments = ParseStringArguments(argumentElement);
+        if (parsedStringArguments is not null)
+        {
+            argumentElement = parsedStringArguments.RootElement;
+        }
+
+        if (argumentElement.ValueKind != JsonValueKind.Object)
         {
             throw new InvalidOperationException(
-                $"Inline skill scripts expect arguments as a JSON object but received a JSON element of kind '{arguments.Value.ValueKind}'.");
+                $"Inline skill scripts expect arguments as a JSON object but received a JSON element of kind '{argumentElement.ValueKind}'.");
         }
 
         var dict = new Dictionary<string, object?>();
-        foreach (var property in arguments.Value.EnumerateObject())
+        foreach (var property in argumentElement.EnumerateObject())
         {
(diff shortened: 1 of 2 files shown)


he-yufeng:
Closing this older version in favor of #6214, which is the current branch for the same #6020 fix and includes the invalid JSON string regression coverage. Keeping one active PR should make review clearer.

PR #6129, opened 2026-05-28
Python: fix(mem0): support mem0ai 2.x in Mem0ContextProvider

youneshima:
### Motivation and Context

`agent-framework-mem0` pins `mem0ai>=1.0.0,<2`, so every install resolves to the legacy `mem0ai==1.0.11` rather than the current 2.x line. This bumps the pin to `mem0ai>=2.0.0,<3` and adapts `Mem0ContextProvider` to the mem0 2.x API so new installs land on current mem0.

### Description

In mem0 >=2.0, `search()` rejects top-level entity params (`user_id`/`agent_id`/`run_id`) and requires them in a `filters={...}` dict — for both the OSS `AsyncMemory` and Platform `AsyncMemoryClient` clients. The previous OSS path passed entity IDs as top-level kwargs, which now raises `ValueError: Top-level entity parameters ... are not supported in search()`.

Changes:
- `before_run` now passes `filters=` for both client types (the OSS/Platform branch collapses to the shared 2.x convention).
- `app_id` is dropped from the **OSS** filters: OSS recognizes only `user_id`/`agent_id`/`run_id` as entities, so an `app_id` filter would be treated as a non-matching metadata filter and silently exclude all results. The Platform client keeps `app_id` (a recognized Platform entity).
- `add()` is unchanged — it still accepts top-level entity kwargs in 2.x.

No change to `Mem0ContextProvider`'s public API (constructor and hook signatures are identical). Not a breaking change for consumers.

Verified end to end against real `mem0ai==2.0.4` with an OSS `AsyncMemory`: the store (`after_run`) → search (`before_run`) round trip succeeds, where it raised `ValueError` before the fix.

### Contribution Checklist

- [ ] Is this a breaking change? No.

### DIFF
--- a/python/packages/mem0/agent_framework_mem0/_context_provider.py
+++ b/python/packages/mem0/agent_framework_mem0/_context_provider.py
@@ -60,7 +60,10 @@ class Mem0ContextProvider(ContextProvider):
             source_id: Unique identifier for this provider instance.
             mem0_client: A pre-created Mem0 MemoryClient or None to create a default client.
             api_key: The API key for authenticating with the Mem0 API.
-            application_id: The application ID for scoping memories.
+            application_id: The application ID for scoping memories. Platform-only:
+                the OSS ``AsyncMemory`` client does not recognize an application
+                scope (it scopes only by user_id/agent_id/run_id), so an OSS client
+                requires user_id or agent_id and application_id alone is rejected.
             agent_id: The agent ID for scoping memories.
             user_id: The user ID for scoping memories.
             context_prompt: The prompt to prepend to retrieved memories.
@@ -170,9 +173,14 @@ class Mem0ContextProvider(ContextProvider):
     # -- Internal methods ------------------------------------------------------
 
     def _validate_filters(self) -> None:
-        """Validates that at least one filter is provided."""
+        """Validates that at least one usable filter is provided for the configured client."""
         if not self.agent_id and not self.user_id and not self.application_id:
             raise ValueError("At least one of the filters: agent_id, user_id, or application_id is required.")
+        if isinstance(self.mem0_client, AsyncMemory) and not self.user_id and not self.agent_id:
+            raise ValueError(
+                "application_id is not supported by the OSS AsyncMemory client, which scopes "
+                "memories only by user_id/agent_id. Provide user_id or agent_id."
+            )
 
     def _build_filters(self) -> dict[str, Any]:
         """Build search filters from initialization parameters."""
--- a/python/packages/mem0/agent_framework_mem0/_context_provider.py
+++ b/python/packages/mem0/agent_framework_mem0/_context_provider.py
@@ -108,16 +108,17 @@ class Mem0ContextProvider(ContextProvider):
 
         filters = self._build_filters()
 
-        # AsyncMemory (OSS) expects user_id/agent_id/run_id as direct kwargs
-        # AsyncMemoryClient (Platform) expects them in a filters dict
-        search_kwargs: dict[str, Any] = {"query": input_text}
+        # mem0 >=2.0: both OSS (AsyncMemory) and Platform (AsyncMemoryClient) take
+        # entity IDs in a filters dict (top-level entity kwargs are rejected). OSS only
+        # recognizes user_id/agent_id/run_id as entities, so app_id (Platform-only) is
+        # dropped for the OSS client to avoid it being treated as a non-matching
+        # metadata filter that would silently exclude all results.
         if isinstance(self.mem0_client, AsyncMemory):
-            search_kwargs.update(filters)
-        else:
-            search_kwargs["filters"] = filters
+            filters = {key: value for key, value in filters.items() if key != "app_id"}
 
         search_response: _MemorySearchResponse_v1_1 | _MemorySearchResponse_v2 = await self.mem0_client.search(  # type: ignore[misc]
-            **search_kwargs,
+            query=input_text,
+            filters=filters,
         )
 
         if isinstance(search_response, list):
--- a/python/packages/mem0/tests/test_mem0_context_provider.py
+++ b/python/packages/mem0/tests/test_mem0_context_provider.py
@@ -211,20 +211,18 @@ class TestBeforeRun:
         assert "app_id" not in call_kwargs["filters"]
         assert "user_id" not in call_kwargs
 
-    async def test_oss_client_excludes_app_id(self, mock_oss_mem0_client: AsyncMock) -> None:
-        """OSS client with only application_id set: app_id must not leak into search filters."""
-        mock_oss_mem0_client.search.return_value = []
+    async def test_oss_client_rejects_application_id_only(self, mock_oss_mem0_client: AsyncMock) -> None:
+        """OSS client with only application_id set: before_run raises and search is never called."""
         provider = Mem0ContextProvider(source_id="mem0", mem0_client=mock_oss_mem0_client, application_id="app1")
         session = AgentSession(session_id="test-session")
         ctx = SessionContext(input_messages=[Message(role="user", contents=["Hello"])], session_id="s1")
 
-        await provider.before_run(
-            agent=None, session=session, context=ctx, state=session.state.setdefault(provider.source_id, {})
-        )  # type: ignore[arg-type]
+        with pytest.raises(ValueError, match="application_id is not supported"):
+            await provider.before_run(
+                agent=None, session=session, context=ctx, state=session.state.setdefault(provider.source_id, {})
+            )  # type: ignore[arg-type]
 
-        call_kwargs = mock_oss_mem0_client.search.call_args.kwargs
-        assert call_kwargs["filters"] == {}
-        assert "app_id" not in call_kwargs["filters"]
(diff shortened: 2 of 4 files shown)


python/packages/mem0/agent_framework_mem0/_context_provider.py
```diff
@@ -108,16 +108,17 @@ async def before_run(
...
         filters = self._build_filters()
 
-        # AsyncMemory (OSS) expects user_id/agent_id/run_id as direct kwargs
-        # AsyncMemoryClient (Platform) expects them in a filters dict
-        search_kwargs: dict[str, Any] = {"query": input_text}
+        # mem0 >=2.0: both OSS (AsyncMemory) and Platform (AsyncMemoryClient) take
+        # entity IDs in a filters dict (top-level entity kwargs are rejected). OSS only
+        # recognizes user_id/agent_id/run_id as entities, so app_id (Platform-only) is
+        # dropped for the OSS client to avoid it being treated as a non-matching
+        # metadata filter that would silently exclude all results.
```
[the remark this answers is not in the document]
youneshima: @microsoft-github-policy-service agree company="Mem0"

python/packages/mem0/agent_framework_mem0/_context_provider.py
```diff
@@ -108,16 +108,17 @@ async def before_run(
...
         if isinstance(self.mem0_client, AsyncMemory):
-            search_kwargs.update(filters)
-        else:
-            search_kwargs["filters"] = filters
+            filters = {key: value for key, value in filters.items() if key != "app_id"}
 
         search_response: _MemorySearchResponse_v1_1 | _MemorySearchResponse_v2 = await self.mem0_client.search(  # type: ignore[misc]
-            **search_kwargs,
+            query=input_text,
+            filters=filters,
```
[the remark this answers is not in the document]
youneshima: Verified in mem0 2.x: an entity-less filter set hard-raises before any data access (search at `memory/main.py:2609`, add at `memory/main.py:301`), so there is no unscoped or cross-tenant path on either read or write.
The valid point underneath (application_id is Platform-only for OSS) is now enforced in 251ee3c: `_validate_filters` rejects an OSS client created with only `application_id` and raises a clear ValueError at the boundary

python/packages/mem0/tests/test_mem0_context_provider.py
```diff
@@ -207,9 +207,24 @@ async def test_oss_client_all_scoping_params(self, mock_oss_mem0_client: AsyncMo
...
+        session = AgentSession(session_id="test-session")
+        ctx = SessionContext(input_messages=[Message(role="user", contents=["Hello"])], session_id="s1")
+
+        await provider.before_run(
+            agent=None, session=session, context=ctx, state=session.state.setdefault(provider.source_id, {})
+        )  # type: ignore[arg-type]
+
+        call_kwargs = mock_oss_mem0_client.search.call_args.kwargs
+        assert call_kwargs["filters"] == {}
+        assert "app_id" not in call_kwargs["filters"]
```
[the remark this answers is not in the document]
youneshima: Addressed in 251ee3c. The rewritten test (`test_oss_client_rejects_application_id_only`) asserts `before_run` raises ValueError and `search` is never called for the OSS + application_id-only case.

python/packages/mem0/agent_framework_mem0/_context_provider.py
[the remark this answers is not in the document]
youneshima: Done both in 251ee3c: `_validate_filters` raises a clear ValueError when an OSS client is created with only `application_id`, and the `application_id` arg docstring on `Mem0ContextProvider.__init__` now states it is Platform-only.

youneshima:
@microsoft-github-policy-service agree company="Mem0"

python/packages/mem0/agent_framework_mem0/_context_provider.py
whysosaket, line 78: We can throw in an error in case client is OSS client i.e AsyncMemory and application_id is passed as well.

Ideally should be done once we know what type of client is being initialized

whysosaket: or the surface can remain application_id but internally we can route it to run_id if this is passed for oss client

whysosaket, line 28: We can deprecate this as now search is always a list now.

python/packages/mem0/agent_framework_mem0/_context_provider.py
```diff
@@ -108,16 +111,17 @@ async def before_run(
...
+        # recognizes user_id/agent_id/run_id as entities, so app_id (Platform-only) is
+        # dropped for the OSS client to avoid it being treated as a non-matching
+        # metadata filter that would silently exclude all results.
         if isinstance(self.mem0_client, AsyncMemory):
-            search_kwargs.update(filters)
-        else:
-            search_kwargs["filters"] = filters
+            filters = {key: value for key, value in filters.items() if key != "app_id"}
 
         search_response: _MemorySearchResponse_v1_1 | _MemorySearchResponse_v2 = await self.mem0_client.search(  # type: ignore[misc]
```
whysosaket, line 122: As mentioned above _MemorySearchResponse_v1_1 can be deprecated

python/packages/mem0/agent_framework_mem0/_context_provider.py
```diff
@@ -169,9 +173,14 @@ def get_role_value(role: Any) -> str:
     # -- Internal methods ------------------------------------------------------
 
     def _validate_filters(self) -> None:
-        """Validates that at least one filter is provided."""
+        """Validates that at least one usable filter is provided for the configured client."""
         if not self.agent_id and not self.user_id and not self.application_id:
             raise ValueError("At least one of the filters: agent_id, user_id, or application_id is required.")
+        if isinstance(self.mem0_client, AsyncMemory) and not self.user_id and not self.agent_id:
```
whysosaket, line 179: I see the intent but messaging can be improved, right now even if users are sending application_id  (in OSS client) it would seem as if it matters but actually it doesn't

python/packages/mem0/agent_framework_mem0/_context_provider.py
whysosaket, line 170: If we're passing the application_id in metadata then criticism of it not being supported on OSS client becomes invalid.

Metadata are handled by both SDKs, and also it would break the filters, as direct filter={application_id: "app123"} cannot be used to filter metadata directly on Platform Client

whysosaket, line 192: Application ID as mentioned above when passed as metadata cannot directly be used in filter like this

whysosaket, line 190: agent_id and user_id are mutually exclusive and tend to break when used together when in same filter group

whysosaket: The Key we're using is application_id on R170 while saving, then while searching how can it be app_id???

whysosaket: This is not a major blocker but still an issue

eavanvalkenburg:
@youneshima can you catch up to main and fix the merge conflicts?

python/packages/mem0/agent_framework_mem0/_context_provider.py
```diff
@@ -108,16 +111,17 @@ async def before_run(
...
+        # mem0 >=2.0: both OSS (AsyncMemory) and Platform (AsyncMemoryClient) take
+        # entity IDs in a filters dict (top-level entity kwargs are rejected). OSS only
+        # recognizes user_id/agent_id/run_id as entities, so app_id (Platform-only) is
+        # dropped for the OSS client to avoid it being treated as a non-matching
+        # metadata filter that would silently exclude all results.
         if isinstance(self.mem0_client, AsyncMemory):
-            search_kwargs.update(filters)
-        else:
-            search_kwargs["filters"] = filters
+            filters = {key: value for key, value in filters.items() if key != "app_id"}
```
moonbox3, line 120: Could we update the corresponding platform add path to use this same 2.x `filters` shape before bumping the dependency? `before_run()` now searches with `filters={...}`, but `after_run()` still calls `AsyncMemoryClient.add()` with top-level `user_id`/`agent_id` and `metadata={"application_id": ...}`; in `mem0ai 2.0.4`, that becomes a `/v3/memories/add/` payload whose identity fields are not under `filters`. That can reject platform stores or write memories under a scope this new search path will not read back, and the existing store test still asserts the old payload shape.

eavanvalkenburg:
Superseded by #7004, which applies the mem0ai 2.x compatibility update on the current branch and keeps the Python checks passing.

python/packages/mem0/agent_framework_mem0/_context_provider.py
eavanvalkenburg: Addressed in replacement PR #7004, commit 5018e46a6: Platform AsyncMemoryClient.add now sends user_id, agent_id, and app_id under filters, while OSS AsyncMemory.add keeps its top-level user_id/agent_id signature.

PR #6137, opened 2026-05-28
Python: fix: avoid empty final response after function limits

### LINKED ISSUE #5769: Anthropic function limit fallback can return empty final response
## Summary

When the function invocation loop reaches `max_function_calls` or `max_iterations`, the core loop correctly stops additional tool calls and makes a final no-tool model call. With the Anthropic provider, that final fallback call can return `stop_reason=end_turn` with an empty `content` array, producing a blank final response.

## Why this matters

The safety limit itself works: tool calls are capped. The problem is the user-visible final turn. A production agent can hit a function-call/iteration limit and return no text instead of explaining that the limit was reached.

## Observed behavior

Using `agent-framework-anthropic` with Anthropic Opus 4.7:

- Tool function returns `call again` until called 3 times.
- Client config sets either:
  - `client.function_invocation_configuration["max_function_calls"] = 1`, or
  - `client.function_invocation_configuration["max_iterations"] = 1`
- SDK executes exactly one tool call, then performs the final fallback call with `tool_choice={"type":"none"}`.
- Raw Anthropic final response is:
  - `stop_reason="end_turn"`
  - `content=[]`

The resulting `Agent.run(...)` response has no text.

## Relevant SDK behavior

The core loop appears to intend a final no-tool answer after limit exhaustion:

- Non-streaming path sets `mutable_options["tool_choice"] = "none"` and calls the underlying chat client again after loop exhaustion.
- Streaming path does the same with a final inner stream.
- Core tests mock this final call as producing fallback text.

That assumption does not always hold for Anthropic when the last tool result says the model should continue.

## Minimal shape of the final Anthropic request

The final request sent by the SDK looked like this conceptually:

```json
{
  "tool_choice": {"type": "none"},
  "tools": [{"type": "custom", "name": "continue_until_three", "input_schema": {}}],
  "messages": [
    {"role": "user", "content": [{"type": "text", "text": "Start the loop."}]},
    {"role": "assistant", "content": [{"type": "tool_use", "id": "toolu_...", "name": "continue_until_three", "input": {"marker": "1"}}]},
    {"role": "user", "content": [{"type": "tool_result", "tool_use_id": "toolu_...", "content": [{"type": "text", "text": "call again"}], "is_error": false}]}
  ]
}
```

Anthropic returned `content=[]`.

Direct Anthropic comparison showed:

- Same messages + `tool_choice=none` -> empty content.
- Same messages with no tools -> empty content.
- Same messages plus explicit user instruction like `Tool call limit reached. Do not call more tools. Explain what happened in one sentence.` -> text response.

## Expected behavior

When the function invocation loop stops because a configured limit is reached, the SDK should not return an empty final assistant message. Possible fixes:

1. Append an explicit final instruction before the no-tool fallback call, e.g. explain that the tool/iteration limit was reached and produce a final answer without further tool use.
2. If the fallback call still returns empty content, synthesize a deterministic limit-reached assistant message.
3. Expose enough metadata to callers so they can reliably detect `limit_reached` and apply their own deterministic fallback.

## Environment

- Python SDK from `microsoft/agent-framework` local clone
- `agent-framework-anthropic` dependency pin locally loosened to allow `anthropic>=0.97.0,<1`
- `anthropic==0.101.0`
- Model: `claude-opus-4-7`
- Thinking: `{"type":"adaptive","display":"omitted"}`
- Output config: `{"effort":"low"}`

## Local proof

This was found while evaluating Microsoft Agent Framework adoption in a production Azure Functions assistant runtime. We are keeping a local guard to avoid returning blank user responses when a tool/iteration limit is reached.

he-yufeng:
﻿## Summary

Fixes #5769.

When the function invocation loop hits `max_iterations`, `max_function_calls`, or the error-stop fallback, the framework makes one last model call with tools disabled so the user gets a plain final answer. Some providers can return an empty assistant message for that call. In that case the SDK currently returns a blank final response even though the loop stopped because of a framework limit.

This adds a framework fallback only for those forced limit/stop paths. Normal empty model responses outside that fallback path are left alone.

## Verification

- `PYTHONPATH=python/packages/core python -m pytest python/packages/core/tests/core/test_function_invocation_logic.py -q -k "empty_final_response_uses_limit_fallback"`
- `PYTHONPATH=python/packages/core python -m pytest python/packages/core/tests/core/test_function_invocation_logic.py -q -k "max_iterations or max_function_calls"`
- `PYTHONPATH=python/packages/core python -m pytest python/packages/core/tests/core/test_function_invocation_logic.py -q`
- `python -m py_compile python/packages/core/agent_framework/_tools.py python/packages/core/tests/core/test_function_invocation_logic.py`
- `python -m ruff check python/packages/core/agent_framework/_tools.py python/packages/core/tests/core/test_function_invocation_logic.py`
- `git diff --check`

Note: pytest still emits the existing local Windows `.pytest_cache` permission warning in this checkout; tests pass.

### DIFF
--- a/python/packages/core/agent_framework/_tools.py
+++ b/python/packages/core/agent_framework/_tools.py
@@ -1856,6 +1856,33 @@ def _clear_internal_conversation_id(response: ChatResponse[Any]) -> ChatResponse
     return response
 
 
+_FUNCTION_INVOCATION_LIMIT_FALLBACK_TEXT = (
+    "Function invocation stopped because the configured limit was reached before the model returned a final answer."
+)
+
+
+def _response_has_final_assistant_content(response: ChatResponse[Any]) -> bool:
+    return bool(
+        response.messages
+        and response.messages[-1].role == "assistant"
+        and response.messages[-1].contents
+    )
+
+
+def _ensure_limit_fallback_response(response: ChatResponse[Any]) -> ChatResponse[Any]:
+    if _response_has_final_assistant_content(response):
+        return response
+
+    from ._types import Content, Message
+
+    fallback = Content.from_text(_FUNCTION_INVOCATION_LIMIT_FALLBACK_TEXT)
+    if response.messages and response.messages[-1].role == "assistant":
+        response.messages[-1].contents.append(fallback)
+    else:
+        response.messages.append(Message(role="assistant", contents=[fallback]))
+    return response
+
+
 def _extract_tools(
     options: dict[str, Any] | None,
 ) -> ToolTypes | Callable[..., Any] | Sequence[ToolTypes | Callable[..., Any]] | None:
@@ -2314,6 +2341,7 @@ class FunctionInvocationLayer(Generic[OptionsCoT]):
         from ._types import (
             ChatResponse,
             ChatResponseUpdate,
+            Content,
             ResponseStream,
             add_usage_details,
         )
@@ -2383,6 +2411,7 @@ class FunctionInvocationLayer(Generic[OptionsCoT]):
                 fcc_messages: list[Message] = []
                 response: ChatResponse[Any] | None = None
                 aggregated_usage: UsageDetails | None = None
+                limit_fallback_requested = False
 
                 loop_enabled = self.function_invocation_configuration.get("enabled", True)
                 max_iterations = self.function_invocation_configuration.get("max_iterations", DEFAULT_MAX_ITERATIONS)
@@ -2436,6 +2465,8 @@ class FunctionInvocationLayer(Generic[OptionsCoT]):
                         execute_function_calls=execute_function_calls,
                     )
                     if result.get("action") == "return":
+                        if limit_fallback_requested:
+                            response = _ensure_limit_fallback_response(response)
                         response.usage_details = aggregated_usage
                         return _clear_internal_conversation_id(response)
                     total_function_calls += result.get("function_call_count", 0)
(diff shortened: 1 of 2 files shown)


python/packages/core/agent_framework/_tools.py
```diff
@@ -1856,6 +1856,33 @@ def _clear_internal_conversation_id(response: ChatResponse[Any]) -> ChatResponse
     return response
 
 
+_FUNCTION_INVOCATION_LIMIT_FALLBACK_TEXT = (
+    "Function invocation stopped because the configured limit was reached before the model returned a final answer."
```
moonbox3, line 1860: Stepping back from line nits: is fabricating an assistant message the right approach here?

- We synthesize assistant-role content the model never produced. No precedent in core; the only framework strings in `_tools.py` are tool-role `function_result` payloads fed back to the model, never user-facing assistant text. And it's hardcoded English (no i18n) that lands in persisted history -> on replay it reads as if the model said it.
- .NET's `FunctionInvokingChatClient` (what we mirror) handles this same limit by stripping tools and returning the model's actual final response. No synthetic message, no fallback text, no signal. Pre-PR Python already does the same (strip tools -> final real call -> INFO log).
- The limit is a boolean set at ~6 sites across `_get_response` and `_stream`; one (approval-stop) is already missed -> the condition wants modeling once, not sprinkling. Plus we mutate `response.messages[-1].contents` in place, against the build-new convention for ChatResponse content.

Could we signal the limit instead of faking content? `finish_reason` is an open `NewType(str)` and we already carry control-flow via `additional_properties` (he internal conversation id) -> surfacing the limit there lets the caller decide on filler, keeps history honest, and matches .NET. If we do want user-facing text, that feels like an agent/caller-layer concern and should at least be marked framework-generated, not a hardcoded literal.

I understand based on the issue this is trying to solve that some clients will produce a final empty output. Perhaps the fix is to have it belong at that layer?

he-yufeng:
Closing this for now. The design concern is fair: synthesizing an assistant message that the model never produced is not a small behavioral fix, and I don't want to keep a questionable fallback PR in the queue. If the maintainers want this solved, I can reopen with a different approach after agreeing on the desired API/UX behavior first.

PR #6141, opened 2026-05-28
.NET: fix: avoid local history when service returns conversation id

### LINKED ISSUE #6120: .NET: [Bug]: InMemoryChatHistoryProvider used even when service stores history (OpenAI Responses)
### Description

I'm trying to use MAF against OpenAI Responses API, i.e. a service that stores history. I'm seeing InMemoryChatHistoryProvider storing messages in the session. 

My understanding was that in this case the in memory provider should "disengage" and the session should only store the conversation/response id. I must be doing something wrong, because this fundamental scenario appears to be broken.

### Code Sample

This test fails on the last assert (`hasInMemoryHistory` is `true`):

```csharp
using System.ClientModel.Primitives;
using Azure.Identity;
using Microsoft.Agents.AI;
using OpenAI;
using OpenAI.Responses;
using Xunit;

namespace MafBug.Tests;

#pragma warning disable OPENAI001

public sealed class InMemoryChatHistoryProviderBugTests
{
    const string Endpoint = "https://xxx.services.ai.azure.com/openai/v1";
    const string Model = "gpt-5-mini";

    [Fact]
    public async Task ResponsesApi_DoesNotPopulateInMemoryChatHistoryProvider()
    {
        var cancellationToken = TestContext.Current.CancellationToken;

        var openai = new OpenAIClient(
            new BearerTokenPolicy(new DefaultAzureCredential(), scope: "https://ai.azure.com/.default"),
            new OpenAIClientOptions { Endpoint = new Uri(Endpoint) });
        var responsesClient = openai.GetResponsesClient();

        var agent = responsesClient.AsAIAgent(
            model: Model,
            instructions: "You are a helpful assistant. Respond with very short answers.",
            name: "ResponsesAgent");

        var session = await agent.CreateSessionAsync(cancellationToken);

        var r1 = await agent.RunAsync("Hi", session, cancellationToken: cancellationToken);

        var typedSession = Assert.IsType<ChatClientAgentSession>(session);
        Assert.False(
            string.IsNullOrWhiteSpace(typedSession.ConversationId),
            "ChatClientAgentSession.ConversationId is empty.");

        var hasInMemoryHistory = session.TryGetInMemoryChatHistory(out var inMemoryMessages);
        var storedCount = inMemoryMessages?.Count ?? 0;

        Assert.False(
            hasInMemoryHistory,
            $"BUG: InMemoryChatHistoryProvider stored {storedCount} message(s) in " +
            $"the session even though the OpenAI Responses API is storing " +
            $"conversation '{typedSession.ConversationId}' server-side. ");
    }
}
```

```csproj
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>net10.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Azure.Identity" Version="1.21.0" />
    <PackageReference Include="Microsoft.Agents.AI" Version="1.7.0" />
    <PackageReference Include="Microsoft.Agents.AI.OpenAI" Version="1.7.0" />
    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="18.5.1" />
    <PackageReference Include="xunit.v3" Version="3.2.2" />
    <PackageReference Include="xunit.runner.visualstudio" Version="3.1.5" />
  </ItemGroup>

</Project>
```

```markdown

```

### Package Versions

1.7.0, also checked 1.6.2

### .NET Version

.net10

### Additional Context

_No response_

he-yufeng:
﻿## Summary
- fix #6120 by treating a service-returned ConversationId as active server-side history for the rest of the run
- skip the default InMemoryChatHistoryProvider notification after the service takes over history
- strengthen the existing regression test so the default provider stays empty when a ConversationId is returned

## To verify
- dotnet run --project dotnet/tests/Microsoft.Agents.AI.UnitTests/Microsoft.Agents.AI.UnitTests.csproj --framework net10.0 -- --filter-method "Microsoft.Agents.AI.UnitTests.ChatClientAgent_ChatHistoryManagementTests.RunAsync_DoesNotThrow_WhenNoChatHistoryProviderInOptionsAndConversationIdReturnedAsync"
- dotnet run --project dotnet/tests/Microsoft.Agents.AI.UnitTests/Microsoft.Agents.AI.UnitTests.csproj --framework net10.0 -- --filter-class "Microsoft.Agents.AI.UnitTests.ChatClientAgent_ChatHistoryManagementTests"
- dotnet build dotnet/src/Microsoft.Agents.AI/Microsoft.Agents.AI.csproj --no-restore
- dotnet build dotnet/tests/Microsoft.Agents.AI.UnitTests/Microsoft.Agents.AI.UnitTests.csproj --no-restore
- git diff --check

Notes: full solution `dotnet build dotnet/agent-framework-dotnet.slnx --no-restore` still fails locally on unrelated projects that have no restored assets and an existing MCP unit-test reference error.


he-yufeng:
Rebased this branch onto current `main` and pushed the updated head `b11fa102`.

I also fixed a rebase fallout in the new `SetResponseConversationId` helper: it now clones `ChatOptions` before adding a returned service conversation id, and it does not treat the local-history sentinel as a real service id. Without that, the current main branch's per-service-call persistence tests observed the sentinel through the same mutable `ChatOptions` reference.

Validation run locally from `dotnet/`:

- `dotnet build .\src\Microsoft.Agents.AI\Microsoft.Agents.AI.csproj -f net10.0 --tl:off` -> passed, 0 warnings/errors
- `dotnet run --project .\tests\Microsoft.Agents.AI.UnitTests\Microsoft.Agents.AI.UnitTests.csproj -f net10.0` -> passed, 1647 tests
- `git diff --check upstream/main..HEAD` -> passed

I used `dotnet run --project` for the xUnit v3/MTP test project because `dotnet test` is not usable with this checkout's current runner configuration on my Windows machine.

he-yufeng:
Rebased this branch onto current main and pushed the updated head $head.

I also addressed the open Copilot review threads:

- SetResponseConversationId now treats the service-returned conversation id as authoritative for follow-up calls by writing it into the cloned ChatOptions.
- The no-local-history assertion no longer depends on InMemoryChatHistoryProvider as a concrete type.
- Added a regression test covering a stale request ConversationId followed by a service-returned id inside a function-calling loop.

Validation run locally from dotnet/:

- dotnet build .\src\Microsoft.Agents.AI\Microsoft.Agents.AI.csproj -f net10.0 --tl:off -> passed, 0 warnings/errors
- dotnet run --project .\tests\Microsoft.Agents.AI.UnitTests\Microsoft.Agents.AI.UnitTests.csproj -f net10.0 -> passed, 1648 tests
- git diff --check upstream/main..HEAD -> passed

peibekwe:
Closing this one as https://github.com/microsoft/agent-framework/pull/6209 is a better solution to the issue.

Issue #3556, opened 2026-01-30
Python: Add middleware sample to show filtering messages between agents


moonbox3:
## Summary

Add middleware sample for message interception between agents.

## Files

| File | Line |
|------|------|
| sample: `group_chat_agent_manager.py` | before GroupChatBuilder |

## Context

```python
# TODO: think about using middleware to help filter messages between agents
workflow = (
    GroupChatBuilder()
    .with_agent_orchestrator(orchestrator_agent)
```

Use cases: content filtering, message transformation, access control between agents.

## Action Items

- [ ] Design middleware interface for message interception
- [ ] Document use cases and patterns

PR #6143, opened 2026-05-28
Python: Add middleware sample to show filtering messages between agents (#3556)

PratikWayase:
### Motivation and Context

This PR adds a sample demonstrating how middleware can be used to intercept and process messages between agents in a group chat workflow.

Currently, there is no built-in middleware hook in `GroupChatBuilder` for message interception. This sample provides a user-space pattern to achieve this behavior without modifying the core framework.

This addresses issue #3556 by illustrating how developers can implement:

* Content filtering
* Message transformation
* Access control between agents

---

### Description

This change updates the `group_chat_agent_manager.py` sample to include a middleware-style pattern for message interception.

Key aspects of the implementation:

* Introduces a simple `MessageInterceptor` callable interface:

  * Input: `Message`
  * Output: `Message` or `None` (to drop messages)

* Demonstrates multiple middleware use cases:

  * Content filtering (e.g., redact or drop sensitive messages)
  * Message transformation (e.g., tagging for audit/logging)
  * Access control (e.g., blocking specific agents)

* Implements a wrapper (`run_with_message_middleware`) that:

  * Intercepts streaming responses (`intermediate` events)
  * Processes final outputs (`output` events)
  * Applies middleware without requiring changes to the core API

* Documents patterns and trade-offs:

  * Middleware chaining
  * Message dropping vs redaction
  * Execution ordering considerations

---

### Contribution Checklist

* [x] The code builds clean without any errors or warnings
* [x] The PR follows the Contribution Guidelines
* [x] This is not a breaking change
* [x] All unit tests pass, and I have added new tests where possible

### DIFF
--- a/python/samples/03-workflows/orchestrations/group_chat_agent_manager.py
+++ b/python/samples/03-workflows/orchestrations/group_chat_agent_manager.py
@@ -2,11 +2,12 @@
 
 import asyncio
 import os
-from typing import cast
+from typing import AsyncIterator, Callable, Optional, cast
 
 from agent_framework import (
     Agent,
     AgentResponseUpdate,
+    Event,
     Message,
 )
 from agent_framework.foundry import FoundryChatClient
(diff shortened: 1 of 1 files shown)


python/samples/03-workflows/orchestrations/group_chat_agent_manager.py
moonbox3, line 1: I do not think this closes #3556 as written. The issue asks for message interception between agents, but this sample wraps `workflow.run()` and post-processes emitted events after the group chat internals have already consumed the participant response. That means a dropped/redacted message can still be appended to the orchestrator conversation and broadcast to other participants before the wrapper sees it.

There are also current API mismatches: `agent_framework.Event` is not exported, `Message` uses `contents=[...]` rather than `text=...`, `Message` has no `replace()` method, `WorkflowEvent` is not a dataclass so `dataclasses.replace(event, ...)` will not work, and streamed updates are emitted as `output` events rather than `intermediate` events.

I think the best way to make this sample match #3556 is to show a custom `BaseGroupChatOrchestrator` that applies the interceptor before participant messages are appended to group-chat state or broadcast to other participants.

An executor/participant wrapper can transform an `AgentExecutorResponse`, but that feels less aligned with the issue: the request is specifically about message interception between agents in group chat. The orchestrator is already the central routing point for participant responses, speaker selection, and broadcasts, so it is the clearest place to demonstrate content filtering, message transformation, and access control without implying that caller-facing `workflow.run()` events are the actual inter-agent message channel.

p.s. I'd also prefer a better name for the sample too, please. And it needs to be added to the workflows samples README. Thanks.

moonbox3 (changes requested):

eavanvalkenburg:
Considering the concerns raised by @moonbox3 and the lack of activity, I'm closing this one.

PR #6144, opened 2026-05-28
.NET: fix: accept stringified inline skill arguments

### LINKED ISSUE #6020: .NET: [Bug]: Arguments valueKind is String instead of Object when calling AgentInlineSkillScript via functionCall

he-yufeng:
﻿Fixes #6020.

## Summary
- accept inline skill arguments that arrive as a JSON string containing an object
- keep non-object arguments rejected, so arrays/plain strings still fail fast
- clone argument `JsonElement` values before invoking the delegate so parsed string-backed arguments survive beyond the temporary `JsonDocument`
- add a regression test for stringified object arguments

## To verify
- `dotnet run --project dotnet\tests\Microsoft.Agents.AI.UnitTests\Microsoft.Agents.AI.UnitTests.csproj --framework net10.0 -- --filter-class "Microsoft.Agents.AI.UnitTests.AgentSkills.AgentInlineSkillScriptTests"`
- `dotnet build dotnet\src\Microsoft.Agents.AI\Microsoft.Agents.AI.csproj --no-restore`
- `dotnet build dotnet\tests\Microsoft.Agents.AI.UnitTests\Microsoft.Agents.AI.UnitTests.csproj --no-restore`
- `git diff --check`

One initial test attempt ran concurrently with a build and hit a transient Windows/Defender lock on an `obj` DLL; the same test command passed when rerun serially.

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI/Skills/Programmatic/AgentInlineSkillScript.cs
+++ b/dotnet/src/Microsoft.Agents.AI/Skills/Programmatic/AgentInlineSkillScript.cs
@@ -85,25 +85,58 @@ internal sealed class AgentInlineSkillScript : AgentSkillScript
     /// </exception>
     private static AIFunctionArguments ConvertToFunctionArguments(JsonElement? arguments)
     {
-        if (arguments is null ||
-            arguments.Value.ValueKind == JsonValueKind.Null ||
-            arguments.Value.ValueKind == JsonValueKind.Undefined)
+        if (arguments is null)
         {
             return [];
         }
 
-        if (arguments.Value.ValueKind != JsonValueKind.Object)
+        var argumentElement = arguments.Value;
+        if (argumentElement.ValueKind == JsonValueKind.Null ||
+            argumentElement.ValueKind == JsonValueKind.Undefined)
+        {
+            return [];
+        }
+
+        using JsonDocument? parsedArguments = argumentElement.ValueKind == JsonValueKind.String
+            ? ParseStringArguments(argumentElement)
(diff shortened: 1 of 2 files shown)


he-yufeng:
Closing this duplicate version in favor of #6214, which is the current branch for the same #6020 fix and has the cleaner current validation coverage. Keeping one active PR for this topic.

PR #6150, opened 2026-05-28
Python: fix DevUI deployment Dockerfile auth args

fallintoplace:
## Summary
Fix the DevUI deployment Dockerfile generator to stop emitting the removed `--auth` CLI flag.

## Why
DevUI auth is enabled by default now, and non-loopback binds are configured with `DEVUI_AUTH_TOKEN` or `--auth-token`.
The generated Dockerfile was still launching `devui` with `--auth`, which is parsed as an abbreviated `--auth-token` and causes startup to fail because no token value is provided.

## What changed
- remove `--auth` from the generated DevUI Dockerfile command
- add a regression test covering the generated `CMD` line
- assert the Dockerfile does not inline `DEVUI_AUTH_TOKEN`

## Validation
- `uv run --group dev pytest packages/devui/tests/devui/test_deployment.py`
- `uv run --group dev ruff check packages/devui/agent_framework_devui/_deployment.py packages/devui/tests/devui/test_deployment.py`
- `uv run --group dev ruff format --check packages/devui/agent_framework_devui/_deployment.py packages/devui/tests/devui/test_deployment.py`


fallintoplace:

eavanvalkenburg (approved):

moonbox3 (approved):

PR #6156, opened 2026-05-28
.NET: Forward Magentic participant replies to manager

### LINKED ISSUE #6172: .NET: [Bug]: MagenticOrchestrator loses participant replies between rounds — manager loops to MaxRounds
### Description
In the .NET Magentic implementation, `MagenticOrchestrator.TakeTurnAsync`
discards the `messages` parameter on subsequent turns. The agent's reply
(e.g. HealthChecker → `[OK]`) is never appended to the manager's
`ChatHistory`, so each progress-ledger round sees only the manager's own
prior instructions, never the participants' answers. The manager
concludes "no team member has answered yet" every round and re-dispatches
the same speaker until `MaxRounds` is hit.

PR up at #6156 

### Minimal repro
A two-participant health-check workflow:

- `MagenticManager` (plain coordinator, no tools)
- `HealthChecker` — always replies `[OK]`
- `DatabaseChecker` — always replies `[OK]`
- Task: "Is the system healthy?"

**Python side**: completes in 3 rounds with the verdict "The system is healthy."
**.NET side**: never reaches `is_request_satisfied = true`; loops until `MaxRounds`.

### Expected
After a participant responds, its reply is part of the chat history
the manager sees on the next progress-ledger call, matching the Python
behaviour. The manager can then determine the task is satisfied and
synthesize a final answer.

### Actual
The manager's progress-ledger input on round N+1 contains only its own
instructions plus the original task ledger — no participant replies.
`is_request_satisfied` stays `false` indefinitely.

```markdown

```

### Code Sample

```markdown
AIAgent healthChecker = chatClient.AsAIAgent(
    name: "HealthChecker",
    description: "Checks the core application framework state.",
    instructions: """
        You only return [OK] and nothing else.
        If someone asks you what is the state or asks anything, you only return [OK].
        """);

AIAgent databaseChecker = chatClient.AsAIAgent(
    name: "DatabaseChecker",
    description: "Checks the database connection state.",
    instructions: """
        You only return [OK] and nothing else.
        If someone asks you what is the database state, you only return [OK].
        """);

AIAgent managerAgent = new ManagerSpyChatClient(chatClient).AsAIAgent(
    name: "MagenticManager",
    description: "Coordinates the health and database checkers to assess system health.",
    instructions: """
        Coordinate HealthChecker and DatabaseChecker to determine overall system health.
        Consult both participants. If either reports [NO], the system is unhealthy.
        Only if both report [OK] is the system healthy. Do not build or fix anything.
        Synthesize a final verdict and stop.
        """);

var workflow = new MagenticWorkflowBuilder(managerAgent)
    .AddParticipants([healthChecker, databaseChecker])
    .WithName("Health Check Magentic Workflow")
    .WithDescription("Coordinates two checkers to report system health.")
    .RequirePlanSignoff(false)
    .WithMaxRounds(6)
    .WithMaxStalls(1)
    .WithMaxResets(1)
    .Build();
```

```markdown

```

### Package Versions

Microsoft.Agents.AI: 1.7.0, Microsoft.Agents.AI.Workflows: 1.7.0

### .NET Version

.Net 10.0.300

### Additional Context

_No response_

ghominejad:
Fixes #6172 , #6173
### Motivation and Context

The .NET `MagenticOrchestrator` does not feed the participant agents'
responses back into the manager's chat history. The manager builds its
progress-ledger prompt from `MagenticTaskContext.ChatHistory`, so it
never sees any participant replies and keeps re-picking the same speaker
until `MaxRounds` terminates the run.

The Python port handles this in `_MagenticOrchestratorExecutor._handle_response`,
which extends `magentic_context.chat_history` with the participant's
messages before re-entering the inner loop. This PR restores parity.

with a two-participant workflow where both participants reply with a short status token, the Python sample terminates in 3 rounds with the expected verdict, while .NET loops until MaxRounds without ever incorporating either reply.

### Description

In `MagenticOrchestrator.TakeTurnAsync`, the `messages` parameter on
subsequent turns carries the participant agent's reply (forwarded by
`AIAgentHostExecutor` and accumulated by `ChatProtocolExecutor`). The
first-turn branch consumes `messages` when constructing the initial
`MagenticTaskContext`, but the subsequent-turn branch ignored it and went
straight to `RunCoordinationRoundAsync`, so the reply was dropped.

The fix appends the incoming `messages` to `_taskContext.ChatHistory`
before invoking the coordination round, mirroring the Python orchestrator's
`chat_history.extend(messages)` step. No public API change, no behavior
change on the first turn or on terminated contexts.

### Tests

- Added `RunCoordinationRound_Forwards_Participant_Reply_To_ManagerAsync` in
  `MagenticOrchestrationTests` — runs a 2-round scenario and asserts the
  worker's reply appears in the manager's round-2 progress-ledger input.
  Verified to fail without the fix.
- Added `RecordingReplayAgent` test helper (sibling of `TestReplayAgent`)
  that captures input messages per call, for any future test that needs to
  assert what the manager was actually handed.
- All `MagenticOrchestrationTests` pass

### Contribution Checklist


ghominejad:

lokitoth (approved):

peibekwe (approved):

PR #6185, opened 2026-05-30
.NET: [.NET] Fix Declarative Foreach dropping multi-field record values (fixes #6183)

### LINKED ISSUE #6183: .NET: [Bug]: Declarative Foreach collapses multi-field records to their first field in the loop value variable
### Description

In declarative workflows, when a `Foreach` action iterates over a table whose rows are **multi-field records**, the loop `value` variable receives only the **first field's value** of each record instead of the full record.

For example, iterating over `[{ name: "Alice", role: "Engineer" }, ...]` assigns `"Alice"` to `Local.currentItem` rather than the whole `{ name, role }` record, so `Local.currentItem.role` is unavailable inside the loop body.

This is distinct from #4195 (parsing produced *empty* records). Here the records are correctly populated, but `ForeachExecutor` deliberately reduces each one to its first property.

### 1. Workflow YAML

```yaml
kind: Workflow
name: repro-foreach-first-field
trigger:
  kind: OnConversationStart
  id: trigger
  actions:
    - kind: SetVariable
      id: set_people
      variable: Local.people
      value: |
        =Table(
          { name: "Alice", role: "Engineer" },
          { name: "Bob",   role: "Designer" },
          { name: "Carol", role: "PM" }
        )
    - kind: Foreach
      id: iterate_people
      items: =Local.people
      value: Local.currentItem
      index: Local.index
      actions:
        - kind: SendActivity
          id: show_item
          activity: "item={Local.currentItem} role={Local.currentItem.role}"
    - kind: EndWorkflow
      id: end
```

### 2. Expected behavior

`Local.currentItem` should be the full record for each iteration, so `show_item` prints e.g. `item={ name: "Alice", role: "Engineer" } role=Engineer`.

### 3. Actual behavior

`Local.currentItem` is just the first field's value (`"Alice"`), and `Local.currentItem.role` resolves to blank. The whole record is lost.

## Root Cause

[`ForeachExecutor.ExecuteAsync`](https://github.com/microsoft/agent-framework/blob/main/dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/ObjectModel/ForeachExecutor.cs#L47) collapses each row to its first property:

```csharp
// ForeachExecutor.cs, line 47
this._values = [.. tableValue.Values.Select(value => value.Properties.Values.First().ToFormula())];
```

`tableValue.Values` is a sequence of `RecordDataValue`. Calling `.Properties.Values.First()` keeps only the first field's value and discards the rest of the record.

The framework already has the correct conversion elsewhere — [`DataValueExtensions.ToFormula`](https://github.com/microsoft/agent-framework/blob/main/dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/Extensions/DataValueExtensions.cs#L53-L57) maps each row through `value.ToRecordValue()`, preserving all fields:

```csharp
TableDataValue tableValue =>
    FormulaValue.NewTable(
        tableValue.Values.FirstOrDefault()?.ParseRecordType() ?? RecordType.Empty(),
        tableValue.Values.Select(value => value.ToRecordValue())),
...
RecordDataValue recordValue => recordValue.ToRecordValue(),
```

`ForeachExecutor` simply does not use that path.

## Proposed Fix

In `ForeachExecutor.ExecuteAsync`, convert each row as a whole value so records keep every field:

```diff
- this._values = [.. tableValue.Values.Select(value => value.Properties.Values.First().ToFormula())];
+ this._values = [.. tableValue.Values.Select(value => value.ToFormula())];
```

Each `RecordDataValue` then routes through the `RecordDataValue recordValue => recordValue.ToRecordValue()` branch of `ToFormula`, producing a full `RecordValue`. This makes `Local.currentItem.<field>` work directly and removes the need for the common user-side workaround of declaring an `index` and using `Index(originalCollection, Local.index + 1).FieldName`.

## Tests Gap

The existing tests in [`ForeachExecutorTest.cs`](https://github.com/microsoft/agent-framework/blob/main/dotnet/tests/Microsoft.Agents.AI.Workflows.Declarative.UnitTests/ObjectModel/ForeachExecutorTest.cs) only build records with a **single** field (e.g. `value`/`item`), so `.First()` happens to return the correct value and the bug stays hidden. A test that iterates over multi-field records and asserts `value` is the full record would catch this regression.

## Related

- #4195 (closed) referenced the same line, but its fix was in `JsonDocumentExtensions.cs` (preventing *empty* records) and did not change `ForeachExecutor`. After that fix, records are populated correctly, yet `Foreach` still discards all but the first field — so this issue remains.

### Package Versions

Reproduced against `origin/main` (`ForeachExecutor.cs` line 47 unchanged as of filing; file last modified 2026-02-12, #3835).

### .NET Version

.NET 8.0

hanhan761:
## Summary

Fix \ForeachExecutor\ that collapsed multi-field records to only the first field's value in the loop value variable.

## Changes

- \ForeachExecutor.cs\ line 52: \alue.Properties.Values.First().ToFormula()\ → \alue.ToFormula()\

## Issue

Fixes #6183

## Verification

The existing tests in \ForeachExecutorTest.cs\ only build records with a single field, so they didn't catch this bug. The fix preserves the original behavior for single-field records while also correctly keeping all fields for multi-field records.

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/ObjectModel/ForeachExecutor.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/ObjectModel/ForeachExecutor.cs
@@ -49,7 +49,7 @@ internal sealed class ForeachExecutor : DeclarativeActionExecutor<Foreach>
         EvaluationResult<DataValue> expressionResult = this.Evaluator.GetValue(this.Model.Items);
         if (expressionResult.Value is TableDataValue tableValue)
         {
-            this._values = [.. tableValue.Values.Select(value => value.Properties.Values.First().ToFormula())];
+            this._values = [.. tableValue.Values.Select(value => value.ToFormula())];
         }
         else
         {


peibekwe:
https://github.com/microsoft/agent-framework/pull/6208

PR #6190, opened 2026-05-30
Python: persist pending request IDs through session serialization (fixes #5818)

### LINKED ISSUE #5818: Python: AgentSession Pending Request Resume Failure
I am using AgentSession to persist and restore workflow execution state in Python. The scenario that fails is Magentic orchestration with human plan review enabled.

When the workflow pauses for human plan review, the SSE stream emits a request_id. I persist the session state, including the pending request metadata, and then try to resume the same session by sending feedback for that same request_id. The session does not resume the pending request and instead errors.

Expected behavior :

If a session pauses on request_info and I restore that same AgentSession later, responding with the same request_id should resume execution successfully.

Question :

How should pending requests in `AgentSession` be handled after restoring a persisted session? Specifically, if human plan review in Magentic generates a `request_id` and that pending request is saved in session state, what is the correct way to resume it later using the same `request_id`?

hanhan761:
﻿## Summary

When a workflow pauses for human-in-the-loop input (e.g. Magentic plan review), the pending request IDs are kept in \WorkflowAgent._pending_requests\. If the session is serialized with \	o_dict()\ and later restored with \rom_dict()\ on a new agent instance, the pending requests are lost. The agent then starts a fresh run instead of resuming the paused workflow.

## Changes

Added two helper methods on \WorkflowAgent\:

- \_store_pending_requests_in_session()\ — saves pending request IDs to \session.state["_pending_requests"]\ after each run
- \_restore_pending_requests_from_session()\ — restores pending request IDs from \session.state\ before each run

Both \_run_impl\ and \_run_stream_impl\ call these methods, so the fix covers streaming and non-streaming execution.

Since the data is stored in \session.state\, the existing \	o_dict()\/\rom_dict()\ serialization handles it automatically — no changes needed to \AgentSession\.

## Related

Fixes #5818

### DIFF
--- a/python/packages/core/agent_framework/_workflows/_agent.py
+++ b/python/packages/core/agent_framework/_workflows/_agent.py
@@ -292,6 +292,7 @@ class WorkflowAgent(BaseAgent):
         # combine the messages
         session_messages: list[Message] = session_context.get_messages(include_input=True)
 
+        self._restore_pending_requests_from_session(provider_session)
         output_events: list[WorkflowEvent[Any]] = []
         async for event in self._run_core(
             session_messages,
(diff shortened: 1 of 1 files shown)


eavanvalkenburg:
Closing this, as there is still discussion on the approach we want to take.

PR #6202, opened 2026-05-30
Python: Fix auto function calling stripping explicit null arguments (fixes #5934)

### LINKED ISSUE #5934: Python: [Bug]: Auto function calling removes null arguments

hanhan761:
## Summary

Changes exclude_none=True to exclude_none=False in FunctionTool.invoke() and _auto_invoke_function() to preserve explicitly-provided 
ull values. When a model passes {"param": null} for a required-but-nullable parameter, exclude_none=True strips the key, causing the schema validation to report it as "missing required argument".

## Root Cause

model_dump(exclude_none=True) cannot distinguish between "field was never provided" (default None) and "field was explicitly set to None". This splits the model's intent when it provides 
ull for optional parameters.

## Changes

- python/packages/core/agent_framework/_tools.py: Changed exclude_none=True to exclude_none=False in 3 locations (lines 639, 648, 1495)
- python/packages/core/tests/core/test_tools.py: Added 	est_invoke_preserves_explicit_none_arguments test

## Issue

Fixes #5934

## Verification

- New test 	est_invoke_preserves_explicit_none_arguments passes
- All existing tests in test_tools.py continue to pass

### DIFF
--- a/python/packages/core/agent_framework/_tools.py
+++ b/python/packages/core/agent_framework/_tools.py
@@ -636,7 +636,7 @@ class FunctionTool(SerializationMixin):
                     parsed_arguments = dict(arguments)
                     if self.input_model is not None and not self._schema_supplied:
                         parsed_arguments = self.input_model.model_validate(parsed_arguments).model_dump(
-                            exclude_none=True
+                            exclude_none=False
                         )
                 elif isinstance(arguments, BaseModel):
                     if (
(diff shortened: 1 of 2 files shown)


moonbox3:
@hanhan761 you have multiple PRs open trying to fix the same issue. Why is that? #6199 along with this one.

moonbox3:
Please make sure you're not duplicating work. #5944 was the first PR introduced with the fix.

PR #6207, opened 2026-05-30
Python: re-role trailing assistant message to user for Anthropic compatibility (fixes #5008)

### LINKED ISSUE #5008: Python: [Bug]: Workflow fails with Anthropic models — assistant message prefill not supported
### Description

When two agents backed by Anthropic AnthropicClient are connected via WorkflowBuilder.add_edge(), the first agent's output (assistant-role messages) is passed directly as context to the second agent. Anthropic's API rejects this because it requires the conversation to end with a user message.

This means any multi-agent workflow using Anthropic models fails on the second agent invocation. The issue is in AgentExecutor with the default context_mode="full" — it forwards the full conversation history including assistant messages without re-roling them for providers that don't support assistant prefill.

Expected Behavior:

The workflow should handle message role conversion automatically when passing context between agents, regardless of the downstream provider. Either:

AgentExecutor should detect that the target agent's provider doesn't support assistant prefill and re-role assistant messages as user messages, or
The WorkflowBuilder should insert a role-conversion step when connecting agents backed by Anthropic.

### Code Sample

```markdown
#!/usr/bin/env python3
"""
Minimal repro: Agent Framework workflow fails with Anthropic models.

When two agents are connected via WorkflowBuilder, the first agent's output
(assistant-role messages) is passed directly to the second agent. Anthropic's
API rejects this because the conversation must end with a user message.

Error:
    anthropic.BadRequestError: 400 - This model does not support assistant
    message prefill. The conversation must end with a user message.

Expected: The workflow should automatically handle message role conversion
when the downstream agent uses Anthropic.
"""

from __future__ import annotations

import asyncio

from agent_framework import AgentResponseUpdate, WorkflowBuilder
from agent_framework_anthropic import AnthropicClient

load_dotenv()


async def main() -> None:
    client = AnthropicClient(model_id="claude-sonnet-4-6")

    agent_a = client.as_agent(
        name="AgentA",
        instructions="You are a helpful assistant. Respond briefly.",
    )

    agent_b = client.as_agent(
        name="AgentB",
        instructions="You are a helpful assistant. Summarize what you received.",
    )

    # AgentA's assistant output flows directly to AgentB.
    # Anthropic rejects this because the conversation ends with an assistant message.
    workflow = (
        WorkflowBuilder(start_executor=agent_a)
        .add_edge(agent_a, agent_b)
        .build()
    )

    async for event in workflow.run("Say hello", stream=True):
        if isinstance(event, AgentResponseUpdate) and event.text:
            print(event.text, end="", flush=True)

    print()


if __name__ == "__main__":
    asyncio.run(main())
```

```markdown
anthropic.BadRequestError: Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': 'This model does not support assistant message prefill. The conversation must end with a user message.'}
```

### Package Versions

agent-framework-rc6

### Python Version

3.12.1

### Additional Context

_No response_

hanhan761:
﻿## Summary

When two agents backed by Anthropic are connected via WorkflowBuilder, the first agent's output (assistant-role messages) is passed as context to the second agent. Anthropic's API rejects this with:

\\\
This model does not support assistant message prefill. The conversation must end with a user message.
\\\

## Changes

Modified \_prepare_messages_for_anthropic\ in the Anthropic chat client:

- After preparing all messages, check if the last message has role "assistant"
- If so, re-role it as "user" so the conversation ends with a valid user message
- Single-agent usage is unaffected (conversation already ends with user message in that flow)

## Related

Fixes #5008

### DIFF
--- a/python/packages/anthropic/agent_framework_anthropic/_chat_client.py
+++ b/python/packages/anthropic/agent_framework_anthropic/_chat_client.py
@@ -699,7 +699,7 @@ class RawAnthropicClient(
 
         Anthropic's API requires that the conversation ends with a user message.
         If the last message is from the assistant, a synthetic user turn is
-        appended to satisfy this constraint.
+        appended when it will not break Anthropic tool_use/tool_result pairing.
         """
         # first system message is passed as instructions
         if messages and isinstance(messages[0], Message) and messages[0].role == "system":
@@ -712,10 +712,18 @@ class RawAnthropicClient(
         # Anthropic requires the conversation to end with a user message.
         # Append a synthetic user turn so chained agent outputs work as
         # valid context for the next agent without rewriting the assistant message.
-        if result and result[-1].get("role") == "assistant":
+        if result and result[-1].get("role") == "assistant" and not self._message_has_tool_use(result[-1]):
             result.append({"role": "user", "content": "Continue"})
 
         return result
+
+    def _message_has_tool_use(self, message: dict[str, Any]) -> bool:
+        """Return whether an Anthropic message contains tool_use blocks."""
+        content = message.get("content")
+        return isinstance(content, list) and any(
+            isinstance(item, dict) and item.get("type") in {"tool_use", "mcp_tool_use", "server_tool_use"}
+            for item in content
+        )
 
     def _prepare_message_for_anthropic(self, message: Message) -> dict[str, Any]:
         """Prepare a Message for the Anthropic client.
--- a/python/packages/anthropic/agent_framework_anthropic/_chat_client.py
+++ b/python/packages/anthropic/agent_framework_anthropic/_chat_client.py
@@ -698,8 +698,8 @@ class RawAnthropicClient(
         as Anthropic expects system instructions as a separate parameter.
 
         Anthropic's API requires that the conversation ends with a user message.
-        If the last message is from the assistant, its role is changed to user
-        to satisfy this constraint.
+        If the last message is from the assistant, a synthetic user turn is
+        appended to satisfy this constraint.
         """
         # first system message is passed as instructions
         if messages and isinstance(messages[0], Message) and messages[0].role == "system":
--- a/python/packages/anthropic/agent_framework_anthropic/_chat_client.py
+++ b/python/packages/anthropic/agent_framework_anthropic/_chat_client.py
@@ -696,11 +696,26 @@ class RawAnthropicClient(
 
         This skips the first message if it is a system message,
         as Anthropic expects system instructions as a separate parameter.
+
+        Anthropic's API requires that the conversation ends with a user message.
+        If the last message is from the assistant, its role is changed to user
+        to satisfy this constraint.
         """
         # first system message is passed as instructions
         if messages and isinstance(messages[0], Message) and messages[0].role == "system":
-            return [self._prepare_message_for_anthropic(msg) for msg in messages[1:]]
-        return [self._prepare_message_for_anthropic(msg) for msg in messages]
+            msgs = list(messages[1:])
+        else:
+            msgs = list(messages)
+
+        result = [self._prepare_message_for_anthropic(msg) for msg in msgs]
+
+        # Anthropic requires the conversation to end with a user message.
+        # Re-role a trailing assistant message as user so chained agent
+        # outputs work as valid context for the next agent.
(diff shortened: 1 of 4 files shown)


eavanvalkenburg (approved):

eavanvalkenburg:
please address the comments @hanhan761

hanhan761:
All Copilot review comments have been addressed and resolved:

1. **exclude_none → exclude_unset=True, exclude_none=False** — applied to both `invoke()` and `_auto_invoke_function()` in `_tools.py`
2. **Test distinguishes omission from explicit None** — test now uses a non-None default and verifies different outputs
3. **Re-role → append synthetic user turn** — `_chat_client.py` now appends `{"role": "user", "content": "Continue"}` instead of rewriting the assistant message. Docstring updated accordingly.

@eavanvalkenburg This is ready for re-review.

python/packages/core/agent_framework/_tools.py
```diff
@@ -636,7 +636,7 @@ async def invoke(
                     parsed_arguments = dict(arguments)
                     if self.input_model is not None and not self._schema_supplied:
                         parsed_arguments = self.input_model.model_validate(parsed_arguments).model_dump(
-                            exclude_none=True
+                            exclude_unset=True, exclude_none=False
```
eavanvalkenburg: what does this have to do with Anthropic?

eavanvalkenburg (changes requested):
There are changes unrelated to the stated goal of this PR in here, touching the core package, please remove

moonbox3:
Closing. Re-open if/when necessary.

moonbox3:
Ah, I see the core changes were reverted. Apologies.

python/packages/anthropic/agent_framework_anthropic/_chat_client.py
```diff
@@ -696,11 +696,26 @@ def _prepare_messages_for_anthropic(self, messages: Sequence[Message]) -> list[d
...
+        else:
+            msgs = list(messages)
+
+        result = [self._prepare_message_for_anthropic(msg) for msg in msgs]
+
+        # Anthropic requires the conversation to end with a user message.
+        # Append a synthetic user turn so chained agent outputs work as
+        # valid context for the next agent without rewriting the assistant message.
+        if result and result[-1].get("role") == "assistant":
+            result.append({"role": "user", "content": "Continue"})
```
moonbox3, line 716: What happens when the trailing assistant message contains tool_use blocks, for example a thread persisted mid-loop or a pause_turn continuation? Anthropic requires the user message immediately after tool_use to carry matching tool_result blocks, so appending a plain text turn there turns a previously valid request into a guaranteed 400. If we keep this, it seems like we need to guard on the last message's content, not just its role.

eavanvalkenburg (approved):

python/packages/anthropic/agent_framework_anthropic/_chat_client.py
hanhan761: Addressed in commit 98cd205f1. The synthetic Continue turn is now only appended when the trailing assistant message does not contain tool_use-like blocks, so we don't break Anthropic's required tool_use/tool_result pairing. Added a regression test for the trailing tool_use case.

moonbox3:
@hanhan761 please fix the failing code quality check:

FAILED: pyright in packages/anthropic
/home/runner/work/agent-framework/agent-framework/python/packages/anthropic/agen
t_framework_anthropic/_chat_client.py
  /home/runner/work/agent-framework/agent-framework/python/packages/anthropic/ag
ent_framework_anthropic/_chat_client.py:724:40 - error: Type of "get" is 
partially unknown
    Type of "get" is "Overload[(key: Unknown, default: None = None, /) -> 
(Unknown | None), (key: Unknown, default: Unknown, /) -> Unknown, (key: Unknown,
default: _T@get, /) -> (Unknown | _T@get)]" (reportUnknownMemberType)
  /home/runner/work/agent-framework/agent-framework/python/packages/anthropic/ag
ent_framework_anthropic/_chat_client.py:725:17 - error: Type of "item" is 
unknown (reportUnknownVariableType)
2 errors, 0 warnings, 0 informations

moonbox3 (approved):

eavanvalkenburg (approved):

PR #6217, opened 2026-05-31
Python: fix: propagate created_at from response.completed into streaming update (#5347)

### LINKED ISSUE #5347: Python: [Bug]: OpenAI Responses streaming final response loses created_at, causing durabletask persistence warning
### Description

When using `OpenAIChatClient` from `agent_framework.openai` with `stream=True`, the final streamed response loses `created_at`.

More specifically:

- `await agent.run(..., stream=False)` returns an `AgentResponse` with `created_at` populated
- `await agent.run(..., stream=True)` followed by `await stream.get_final_response()` returns an `AgentResponse` with `created_at=None`

This becomes especially visible when using `agent-framework-durabletask`, because its durable entity path prefers `stream=True` first and then persists the final response. During persistence it logs:

```text
Invalid or missing created_at value in durable agent state; defaulting to current UTC time, None
```

The durable flow still works because it falls back to the current UTC time, but the final streamed response metadata is wrong and the warning is noisy.

## Minimal repro

```python
import asyncio
import os
from azure.identity.aio import DefaultAzureCredential, get_bearer_token_provider

async def main():
    async_credential = DefaultAzureCredential()
    token_provider = get_bearer_token_provider(
        async_credential,
        "https://cognitiveservices.azure.com/.default",
    )

    client = OpenAIChatClient(
        base_url=f"{os.environ['AZURE_OPENAI_ENDPOINT']}/openai/v1/",
        api_key=token_provider,
        model=os.environ["AZURE_OPENAI_CHAT_DEPLOYMENT"],
    )

    agent = Agent(
        name="Probe",
        instructions="Reply with ok only.",
        client=client,
    )

    non_stream = await agent.run("say ok")

    stream = await agent.run("say ok", stream=True)
    async for _ in stream:
        pass
    final = await stream.get_final_response()

    print("non_stream.created_at =", repr(non_stream.created_at))
    print("stream_final.created_at =", repr(final.created_at))

    await async_credential.close()

asyncio.run(main())
```

Observed output:

```text
non_stream.created_at = '2026-04-18T12:48:06.000000Z'
stream_final.created_at = None
```

## Additional validation

I also checked the raw final OpenAI streaming event. The final `response.completed` event still contains a populated timestamp on the embedded response object:

- `event.type == "response.completed"`
- `event.response.created_at` is populated

So the timestamp still exists in the raw streamed response, but it is not propagated into the final `ChatResponse` / `AgentResponse`.

## Impact

- noisy warning in durabletask samples/apps
- incorrect metadata on final streamed responses
- fallback timestamp masks the issue, so behavior still appears to work while metadata is wrong

### Code Sample

```markdown

```

```markdown

```

### Package Versions

agent-framework-core==1.0.0, agent-framework-openai==1.0.0, agent-framework-durabletask==1.0.0b260402

### Python Version

3.12

### Additional Context

_No response_

awesome-pro:
### Motivation and Context

Fixes #5347

`await agent.run(..., stream=True)` followed by `stream.get_final_response()` returned an `AgentResponse` with `created_at=None`, while the non-streaming path returned the correct timestamp. This caused the durabletask persistence warning:

```
Invalid or missing created_at value in durable agent state; defaulting to current UTC time, None
```

### Description

`_parse_chunk_from_openai` never populated `created_at` on the `ChatResponseUpdate` it returned. When the final `ChatResponse` is assembled from the list of streaming updates, `created_at` is only set when an update carries it — so it stayed `None`.

The `response.completed` event already has `event.response.created_at` (a Unix timestamp). This patch:
1. Adds a `created_at: str | None = None` local variable to `_parse_chunk_from_openai`
2. In the `response.completed` case, converts the timestamp to the same ISO-8601 format used by the non-streaming path
3. Passes it to the returned `ChatResponseUpdate`

### Contribution Checklist

### DIFF
--- a/python/packages/openai/agent_framework_openai/_chat_client.py
+++ b/python/packages/openai/agent_framework_openai/_chat_client.py
@@ -1975,6 +1975,7 @@ class RawOpenAIChatClient(  # type: ignore[misc]
         local_shell_tool_name = self._get_local_shell_tool_name(options.get("tools"))
         conversation_id: str | None = None
         response_id: str | None = None
+        created_at: str | None = None
         continuation_token: OpenAIContinuationToken | None = None
         model = self.model
         match event.type:
@@ -2156,6 +2157,10 @@ class RawOpenAIChatClient(  # type: ignore[misc]
                 response_id = event.response.id
                 conversation_id = self._get_conversation_id(event.response, options.get("store"))
                 model = event.response.model
+                if event.response.created_at is not None:
+                    created_at = datetime.fromtimestamp(event.response.created_at, tz=timezone.utc).strftime(
+                        "%Y-%m-%dT%H:%M:%S.%fZ"
+                    )
                 if event.response.usage:
                     usage = self._parse_usage_from_openai(event.response.usage)
                     if usage:
@@ -2521,6 +2526,7 @@ class RawOpenAIChatClient(  # type: ignore[misc]
             contents=contents,
             conversation_id=conversation_id,
             response_id=response_id,
+            created_at=created_at,
             role="assistant",
             model=model,
             continuation_token=continuation_token,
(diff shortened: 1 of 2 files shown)


awesome-pro:

PR #6222, opened 2026-05-31
.NET: Fix Magentic to share agent replies across team

### LINKED ISSUE #6223: .NET: [Bug]: Magentic participant message routing is the opposite of GroupChatHost and the Python reference
### What happened?

In a Magentic workflow, a participant that takes a later turn receives the
*instruction* the manager addressed to an earlier participant, but never
receives that earlier participant's *response*.

`MagenticOrchestrator` sends the per-round manager instruction untargeted, so
the orchestrator's fan-out edge delivers it to every participant rather than
only the selected speaker. Participant replies are never relayed to the other
participants - they reach only the manager's `ChatHistory`. So an idle agent
accumulates instructions meant for other agents, while never seeing what those agents actually
produced.

This is the opposite of the .NET `GroupChatHost` (same assembly), which tracks
the current speaker and broadcasts replies to all other participants, and the
opposite of the Python Magentic orchestrator, which targets the instruction at
the selected speaker and broadcasts responses to the others.

### What did you expect to happen?

Each participant should see the running conversation - prior participants'
*responses* - plus its own instruction, and should NOT see instructions
addressed to other participants. This matches `GroupChatHost` and the Python
reference.

**Before - sees the first agent's instruction (leaked via fan-out), not its reply:**

System: "..."
User: "Instruction meant for Agent1" <-- ⚠️ LEAKED: Untargeted fan-out
User: "Instruction meant for Agent2"

**After - sees the first agent's response, plus its own instruction:**

System: "..."
Assistant/Agent1: "Agent1's response" <-- BROADCASTED: Correct context
User: "Instruction meant for Agent2" <-- 🎯 TARGETED: Direct instruction

### Steps to reproduce the issue

1. Build a Magentic workflow with one manager and two participant agents
   (e.g. HealthChecker and DatabaseChecker), each instructed to reply with a
   short fixed marker.
2. Run it with a single task message and a `TurnToken`.
3. Have the manager select HealthChecker in round 1 and DatabaseChecker in
   round 2.
4. Inspect the message list DatabaseChecker receives on its turn (e.g. log the
   request reaching the second agent).

**Actual:** DatabaseChecker's input contains HealthChecker's *instruction* and
its own instruction, but not HealthChecker's response

### Code Sample

```markdown
AIAgent healthChecker = chatClient.AsAIAgent(
    name: "HealthChecker",
    description: "Checks the core application framework state.",
    instructions: """
        You only return [The core framework is healthy!] and nothing else.
        If someone asks you what is the state or asks anything, you only confirm it's healthy.
        """);

AIAgent databaseChecker = chatClient.AsAIAgent(
    name: "DatabaseChecker",
    description: "Checks the database connection state.",
    instructions: """
        You only return [The Database is healthy!] and nothing else.
        If someone asks you what is the database state, you only confirm it's healthy.
        """);

AIAgent managerAgent = chatClient.AsAIAgent(
    name: "MagenticManager",
    description: "Coordinates the health and database checkers to assess system health.",
    instructions: """
        Coordinate HealthChecker and DatabaseChecker to determine overall system health.
        Consult both participants. If either reports Unhealthy, the system is unhealthy.
        Only if both report [Healthy] is the system healthy. Do not build or fix anything.
        Synthesize a final verdict and stop.
        """);

var workflow = new MagenticWorkflowBuilder(managerAgent)
    .AddParticipants([healthChecker, databaseChecker])
    .WithName("Health Check Magentic Workflow")
    .WithDescription("Coordinates two checkers to report system health.")
    .RequirePlanSignoff(false)
    .WithMaxRounds(6)
    .WithMaxStalls(1)
    .WithMaxResets(1)
    .Build();
```

```markdown

```

### Package Versions

Microsoft.Agents.AI: 1.8.0, Microsoft.Agents.AI.Workflows:1.8.0

### .NET Version

.Net 10

### Additional Context

_No response_

PR open : #6222 (implements the fix)

### What The PR #6222 Fixes:
- [x] **Targeted Instructions:** Restricts the per-round manager instruction to the selected speaker only, preventing idle agents from accumulating unrelated instructions.
- [x] **Response Broadcasting:** Broadcasts each participant's response to all other participants so the running conversation context is maintained.
- [x] **Checkpoint Integrity:** Persists CurrentSpeakerExecutor across checkpoints so the exclusion
survives restore.

ghominejad:
Fixes #6223 

### Motivation and Context

`MagenticOrchestrator` sent the per-round instruction untargeted, so the fan-out edge
delivered it to every participant, and never relayed replies. A later speaker saw the
prior speaker's *instruction* but not its *response* — inverted from `GroupChatHost`
(same assembly) and the Python reference.

Net effect: an idle participant accumulated instructions addressed to *other*
agents but never saw other agents' *responses* - inverted from the intended
behavior, which breaks collaborative hand-offs and dilutes each agent's context.

The .NET `GroupChatHost` already implements the correct pattern (tracks the
current speaker, broadcasts replies to all others, persists that id across
checkpoints), and the Python orchestrator does the same. This PR aligns
Magentic with both.

### Description

- Target the per-round instruction/question at the selected speaker only, instead of an
  untargeted send.
- Broadcast each participant reply to the other participants as context
  (buffered, no `TurnToken`, so they do not act), excluding the responder via
  `_currentSpeakerExecutorId` - mirroring `GroupChatHost`.
- Persist `_currentSpeakerExecutorId` across checkpoints so the exclusion
  survives restore.
- Add a regression test (`RecordingEchoAgent`) asserting a later speaker
  receives the prior participant's *response* and not its *instruction*.

### Before / after:

**Before** - sees the first agent's instruction (leaked via fan-out), not its reply:

- System: "..."
- User: "Instruction meant for Agent1"  <-- ⚠️ LEAKED: Untargeted fan-out
- User: "Instruction meant for Agent2"

**After** - sees the first agent's response, plus its own instruction:

- System: "..."
- Assistant/Agent1: "Agent1's response"         <--  BROADCASTED: Correct context
- User: "Instruction meant for Agent2"    <-- 🎯 TARGETED: Direct instruction

### Contribution Checklist

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/Magentic/MagenticOrchestrator.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/Magentic/MagenticOrchestrator.cs
@@ -338,6 +338,7 @@ internal class MagenticOrchestrator(AIAgent managerAgent, List<AIAgent> team, Ta
     {
         bool wasStalled = taskContext.IsStalled;
         taskContext.Reset();
+        this._currentSpeakerExecutorId = null;
         await context.SendMessageAsync(new ResetChatSignal(), cancellationToken: cancellationToken).ConfigureAwait(false);
 
         await this.UpdatePlanAndDelegateAsync(taskContext, context, cancellationToken, replanAfterStall: wasStalled).ConfigureAwait(false);
@@ -348,9 +349,8 @@ internal class MagenticOrchestrator(AIAgent managerAgent, List<AIAgent> team, Ta
         List<ChatMessage> messages = [await this._manager.PrepareFinalAnswerAsync(taskContext, context, cancellationToken).ConfigureAwait(false)];
         await context.YieldOutputAsync(messages, cancellationToken).ConfigureAwait(false);
         taskContext.IsTerminated = true;
+        this._currentSpeakerExecutorId = null;
     }
-
-    private const string CurrentSpeakerStateKey = "CurrentSpeakerStateKey";
 
     protected internal override async ValueTask OnCheckpointingAsync(IWorkflowContext context, CancellationToken cancellationToken = default)
     {
@@ -361,7 +361,7 @@ internal class MagenticOrchestrator(AIAgent managerAgent, List<AIAgent> team, Ta
                                                               cancellationToken: cancellationToken)
                                        .AsTask();
 
-        Task currentSpeakerTask = context.QueueStateUpdateAsync(CurrentSpeakerStateKey,
+        Task currentSpeakerTask = context.QueueStateUpdateAsync(MagenticConstants.CurrentSpeakerStateKey,
                                                             this._currentSpeakerExecutorId,
                                                             cancellationToken: cancellationToken)
                                         .AsTask();
@@ -390,7 +390,7 @@ internal class MagenticOrchestrator(AIAgent managerAgent, List<AIAgent> team, Ta
 
         async Task LoadCurrentSpeakerAsync()
         {
-            this._currentSpeakerExecutorId = await context.ReadStateAsync<string?>(CurrentSpeakerStateKey, cancellationToken: cancellationToken)
+            this._currentSpeakerExecutorId = await context.ReadStateAsync<string?>(MagenticConstants.CurrentSpeakerStateKey, cancellationToken: cancellationToken)
                                                         .ConfigureAwait(false);
         }
     }
--- a/dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/Magentic/MagenticOrchestrator.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/Magentic/MagenticOrchestrator.cs
@@ -90,6 +90,7 @@ internal class MagenticOrchestrator(AIAgent managerAgent, List<AIAgent> team, Ta
 
     private MagenticTaskContext? _taskContext;
     private PortBinding? _planReviewPort;
+    private string? _currentSpeakerExecutorId;
 
     protected override ProtocolBuilder ConfigureProtocol(ProtocolBuilder protocolBuilder)
     {
@@ -196,13 +197,36 @@ internal class MagenticOrchestrator(AIAgent managerAgent, List<AIAgent> team, Ta
         else
         {
             // Subsequent turns: agent returned control, go directly to coordination (progress ledger only, no replan).
-            // Capture the participant's reply into the manager-visible chat history so the progress ledger can see it.
             if (messages is { Count: > 0 })
             {
+                // Capture the participant's reply into the manager-visible chat history so the progress ledger can see it.
                 this._taskContext.ChatHistory.AddRange(messages);
+
+                // Share the reply with the other participants except the replier
+                await this.BroadcastReplyToOtherParticipantsAsync(messages, context, cancellationToken).ConfigureAwait(false);
             }
             await this.RunCoordinationRoundAsync(this._taskContext, context, cancellationToken).ConfigureAwait(false);
         }
+    }
+
+    /// <summary>
+    /// Forwards a participant's reply to every other participant so they share the running conversation.
+    /// The messages are buffered (no <see cref="TurnToken"/> is sent) - they only become context for the participant's next turn.
+    /// </summary>
+    private ValueTask BroadcastReplyToOtherParticipantsAsync(
+        List<ChatMessage> messages, IWorkflowContext context, CancellationToken cancellationToken)
+    {
+        List<Task>? sendTasks = null;
+        foreach (AIAgent agent in team)
+        {
+            string executorId = AIAgentHostExecutor.IdFor(agent);
+            if (string.Equals(executorId, this._currentSpeakerExecutorId, StringComparison.Ordinal))
+            {
+                continue;
+            }
+            (sendTasks ??= []).Add(context.SendMessageAsync(messages, executorId, cancellationToken).AsTask());
+        }
+        return sendTasks is null ? default : new ValueTask(Task.WhenAll(sendTasks));
     }
 
     private ChatMessage? _fullTaskLedgerMessage;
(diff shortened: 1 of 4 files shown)


dotnet/tests/Microsoft.Agents.AI.Workflows.UnitTests/MagenticOrchestrationTests.cs
```diff
@@ -419,6 +419,84 @@ public async Task RunCoordinationRound_Forwards_Participant_Reply_To_ManagerAsyn
...
+
+        // Each participant takes exactly one turn.
+        healthChecker.RecordedInputs.Should().ContainSingle();
+        databaseChecker.RecordedInputs.Should().ContainSingle();
+
+        // The first speaker must NOT see the instruction targeted at the other speaker.
+        List<ChatMessage> healthInput = healthChecker.RecordedInputs[0];
+        healthInput.Should().Contain(m => m.Text.Contains(HealthInstruction), "the first speaker receives its own instruction");
+        healthInput.Should().NotContain(m => m.Text.Contains(DatabaseInstruction),
+            "an idle agent must not receive an instruction addressed to a different speaker");
```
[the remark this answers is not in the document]
ghominejad: fixed in `c8e3485`: Removed the trivially-true...

dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/Magentic/MagenticOrchestrator.cs
```diff
@@ -196,15 +197,38 @@ protected override async ValueTask TakeTurnAsync(List<ChatMessage> messages, IWo
...
+        List<ChatMessage> messages, IWorkflowContext context, CancellationToken cancellationToken)
+    {
+        List<Task>? sendTasks = null;
+        foreach (AIAgent agent in team)
+        {
+            string executorId = AIAgentHostExecutor.IdFor(agent);
+            if (string.Equals(executorId, this._currentSpeakerExecutorId, StringComparison.Ordinal))
+            {
+                continue;
+            }
```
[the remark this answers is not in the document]
ghominejad: Fixed in <c8e3485>: added an early-return guard when _currentSpeakerExecutorId is null/empty

dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/Magentic/MagenticOrchestrator.cs
```diff
@@ -316,6 +343,8 @@ private async ValueTask PrepareFinalAnswerAsync(MagenticTaskContext taskContext,
     }
 
     private const string CurrentTurnEmitUpdateEventsKey = nameof(CurrentTurnEmitUpdateEventsKey);
+    private const string CurrentSpeakerStateKey = nameof(_currentSpeakerExecutorId);
```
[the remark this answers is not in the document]
ghominejad: Fixed: the key changed from nameof(...) to "CurrentSpeakerStateKey"

dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/Magentic/MagenticOrchestrator.cs
```diff
@@ -196,15 +197,38 @@ protected override async ValueTask TakeTurnAsync(List<ChatMessage> messages, IWo
...
+    {
+        List<Task>? sendTasks = null;
+        foreach (AIAgent agent in team)
+        {
+            string executorId = AIAgentHostExecutor.IdFor(agent);
+            if (string.Equals(executorId, this._currentSpeakerExecutorId, StringComparison.Ordinal))
+            {
+                continue;
+            }
+            (sendTasks ??= []).Add(context.SendMessageAsync(messages, executorId, cancellationToken).AsTask());
```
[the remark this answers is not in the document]
ghominejad: Thanks — I looked into this, and the list payload is received correctly, so
I've left it as a single send.

`ChatProtocolExecutor` (the base class for `AIAgentHostExecutor`) registers a
handler for `IEnumerable<ChatMessage>`, not only for individual `ChatMessage`,
so a `List<ChatMessage>` routes to that handler.

dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/Magentic/MagenticOrchestrator.cs
```diff
@@ -315,7 +350,8 @@ private async ValueTask PrepareFinalAnswerAsync(MagenticTaskContext taskContext,
         taskContext.IsTerminated = true;
     }
 
-    private const string CurrentTurnEmitUpdateEventsKey = nameof(CurrentTurnEmitUpdateEventsKey);
+    private const string CurrentSpeakerStateKey = "CurrentSpeakerStateKey";
```
peibekwe: nit: For consistency, can you please use `nameof` for the key value? i.e. `nameof(_currentSpeakerExecutorId`

dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/Magentic/MagenticOrchestrator.cs
```diff
@@ -344,5 +387,11 @@ async Task LoadContextStateAsync()
                 this._taskContext = new MagenticTaskContext(state, team, limits, []);
             }
         }
+
+        async Task LoadCurrentSpeakerAsync()
+        {
+            this._currentSpeakerExecutorId = await context.ReadStateAsync<string?>(CurrentSpeakerStateKey, cancellationToken: cancellationToken)
```
peibekwe: `_currentSpeakerExecutorId` doesn't seem to be cleared on lifecycle transitions. e.g. `PrepareFinalAnswerAsync` and `ResetAndReplanAsync` don't clear it. A checkpoint snapshot taken in those transition windows might persist a stale id that survives restore. You can take a look at how `GroupChat` does this for `ResetAsync`.

dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/Magentic/MagenticOrchestrator.cs
ghominejad: Happy to drop the magic literal. One thought: Would nameof(CurrentSpeakerStateKey) work instead of the private field? It matches MagenticTaskContextKey in MagenticConstants. My only hesitation with nameof(_currentSpeakerExecutorId) is that the persisted key would then track a private field name. Fine to go with the field name if you'd prefer to keep it identical to GroupChatHost, though.

dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/Magentic/MagenticOrchestrator.cs
ghominejad: Done in 5ae36d8 — clears _currentSpeakerExecutorId in both ResetAndReplanAsync and PrepareFinalAnswerAsync

dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/Magentic/MagenticOrchestrator.cs
ghominejad: Went ahead and pushed nameof(CurrentSpeakerStateKey) in 5ae36d8. Easy to switch to nameof(_currentSpeakerExecutorId) for exact GroupChatHost parity if you'd prefer; just say the word.

dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/Magentic/MagenticOrchestrator.cs
peibekwe: Thank you.

dotnet/src/Microsoft.Agents.AI.Workflows/Specialized/Magentic/MagenticOrchestrator.cs
peibekwe: `nameof(CurrentSpeakerStateKey)` works. It still maintains consistency with other constants in the class.

peibekwe (approved):

westey-m (approved):

PR #6246, opened 2026-06-01
.NET - Fix missing id on function_call_output in Foundry Hosting

alliscode:
The Foundry storage layer was rejecting responses with "ID cannot be null or empty (Parameter 'id')" because function_call_output items emitted by OutputConverter had no id on the wire.

OutputItemFunctionToolCallOutput's public ctor only sets CallId and Output; Id is read-only and only the SDK's internal ctor populates it. OutputItemBuilder<T>.ApplyAutoStamps fills ResponseId and AgentReference but not Id, so the itemId passed to AddOutputItem<T>(itemId) was used only for event sequencing and the serialized item went out with id=null.

Switch to stream.OutputItemFunctionCallOutput(callId, output), the SDK convenience method that uses the internal ctor and stamps the id. Add a regression test asserting the added/done events carry a non-empty matching Id.

### Contribution Checklist

### DIFF
--- a/dotnet/tests/Microsoft.Agents.AI.Foundry.Hosting.UnitTests/OutputConverterTests.cs
+++ b/dotnet/tests/Microsoft.Agents.AI.Foundry.Hosting.UnitTests/OutputConverterTests.cs
@@ -704,6 +704,35 @@ public class OutputConverterTests
         Assert.Equal("[{\"id\":1}]", inner);
     }
 
+    // K-06e: Regression — the OutputItemFunctionToolCallOutput must have a populated Id
+    // and a matching wire id on the added/done events. The Foundry storage layer extracts
+    // a partition id from this field and throws "ID cannot be null or empty (Parameter 'id')"
+    // when it is missing.
+    [Fact]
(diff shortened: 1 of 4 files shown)


dotnet/tests/Microsoft.Agents.AI.Foundry.Hosting.UnitTests/OutputConverterTests.cs
```diff
@@ -704,6 +704,35 @@ public async Task ConvertUpdatesToEventsAsync_FunctionResultJsonElementArrayPayl
         Assert.Equal("[{\"id\":1}]", inner);
     }
 
+    // K-06e: Regression — the OutputItemFunctionToolCallOutput must have a populated Id
```
peibekwe, line 707: nit: I'm not sure if this K-06e is a code for something or shouldn't be there..

peibekwe (approved):

TaoChenOSU (approved):

peibekwe (approved):

Issue #6245, opened 2026-06-01
.NET: [Bug]: OutputConverter.FunctionResultContent creates OutputItemFunctionToolCallOutput with null id, causes HTTP 500 on persistence


enhendrickson:
## Summary

When a Foundry-hosted agent invokes a tool, the `OutputConverter.FunctionResultContent` case creates `OutputItemFunctionToolCallOutput` items with a null `Id` field. This causes the Responses API storage layer to reject the persistence request with HTTP 500, followed by a 409 on retry (confirming partial resource creation). While `Azure.AI.AgentServer.Responses v1.0.0-beta.5` gracefully catches this and emits a `response.failed` event instead of breaking the SSE stream, the root cause remains unfixed.

## Root Cause

In [OutputConverter.cs](https://github.com/microsoft/agent-framework/blob/main/dotnet/src/Microsoft.Agents.AI.Foundry.Hosting/OutputConverter.cs) (~line 450-460), the `FunctionResultContent` case:

```csharp
case FunctionResultContent functionResult:
{
    var outputText = EncodeFunctionResultAsJsonStringPayload(functionResult.Result);
    var itemId = GenerateItemId("fc");  // WRONG: "fc" is function_call prefix, not function_call_output
    var outputItem = new OutputItemFunctionToolCallOutput(
        functionResult.CallId,
        BinaryData.FromString(outputText));  // Id property stays null (get-only)
    var outputBuilder = stream.AddOutputItem<OutputItemFunctionToolCallOutput>(itemId);
    yield return outputBuilder.EmitAdded(outputItem);
    yield return outputBuilder.EmitDone(outputItem);
    break;
}
```

**Problem:** The 2-arg constructor `OutputItemFunctionToolCallOutput(callId, output)` doesn't set the `Id` property (it's get-only). When the item is serialized for persistence (using `ModelReaderWriterOptions.Json`), the `id` field is written as `null`, which violates the Responses API schema and causes HTTP 500.

**Why it happens:** The 2-arg public constructor is for deserialization. The internal 9-arg constructor (used by the SDK's convenience method) has the proper id setup:
```csharp
internal OutputItemFunctionToolCallOutput(
    string id,
    string callId,
    BinaryData output,
    ... other fields ...)
```

## Reproduction

1. Deploy a Foundry hosted agent with Foundry.Hosting 1.8.0-preview
2. Register any tool via `AddFoundryToolboxes()` or `AsOpenAIResponseTool()`
[...]

Issue #5941, opened 2026-05-18
Python: [Bug]: Multi-turn tool calls fail with both Responses API and Chat Completions API when used via `agent-framework-ag-ui` (Foundry project endpoint)


nomhiro:
## Summary

Using `agent-framework-ag-ui` from Python with a Microsoft Foundry project inference endpoint, the **first** turn of a tool-calling chat works correctly, but **any second or later turn that follows a tool call** fails with a 400 from the upstream OpenAI API.

The failure mode differs by client type but both surface a consistent root cause: when AG-UI replays conversation history on subsequent turns, the messages produced by the prior tool-calling turn are not paired correctly with their tool results before being sent back to the LLM.

This affects all four AG-UI features we tested (Agentic Chat, Human-in-the-Loop, Shared State, Agentic Generative UI) as soon as the agent has invoked any tool at least once.

### Code Sample

```markdown
## Reproduction

Minimal server (full repo: <will-be-linked>):


import os
from agent_framework import Agent, tool
from agent_framework.openai import OpenAIChatClient  # or OpenAIChatCompletionClient — both fail
from agent_framework_ag_ui import add_agent_framework_fastapi_endpoint
from fastapi import FastAPI
from fastapi.middleware.cors import CORSMiddleware
from dotenv import load_dotenv

load_dotenv()

@tool
def search_cars(car_class: str | None = None) -> list[dict]:
    return [{"car_id": "SUV-001", "class": "suv", "daily_jpy": 12800}]

client = OpenAIChatClient(  # also tried OpenAIChatCompletionClient — same outcome, different error
    model=os.environ["FOUNDRY_MODEL"],
    base_url=os.environ["FOUNDRY_BASE_URL"],  # ".../openai/v1"
    api_key=os.environ["FOUNDRY_API_KEY"],
)

agent = Agent(
    name="ShirokumaChat",
    instructions="You are a car rental booking assistant.",
    client=client,
    tools=[search_cars],
)

app = FastAPI()
app.add_middleware(CORSMiddleware, allow_origins=["*"], allow_methods=["*"], allow_headers=["*"])
add_agent_framework_fastapi_endpoint(app, agent, "/agentic_chat")


Steps:

1. Start the server on `localhost:8888`.
2. In AG-UI Dojo (running locally), open the Agentic Chat tab for `Microsoft Agent Framework (Python)`.
3. Send turn 1: "Show me your SUV cars" → tool is called, response streams correctly. ✅
[...]

Issue #5662, opened 2026-05-05
.NET: `externalLoop` + `ApprovalRequiredAIFunction` leaves dangling `function_call` in Azure Conversations API, breaks resume


alliscode:
## Summary

When a declarative workflow invokes an Azure agent with `externalLoop: true` and that agent has a tool wrapped in `ApprovalRequiredAIFunction`, the **first** turn correctly surfaces an `mcp_approval_request` to the client. On the **second** turn (after the client posts the `mcp_approval_response`), the request fails with:

```
HTTP 400: No tool output found for function call call_J9Y...
```

This blocks the canonical pattern of "agent with a tool that requires approval" inside a declarative workflow on Foundry hosting. The HITL bridge added in microsoft/agent-framework#5589 works correctly for the `InvokeFunctionTool + requireApproval` shape (workflow-level tool); only the agent-level / `externalLoop` shape is affected.

## Repro

Repo: `alliscode/foundry-samples-pr` (or any equivalent declarative-workflow hosted-agent setup using 1.4.x packages).

1. In `workflow.yaml`, define a single agent step using `InvokeAzureAgent` with `externalLoop: true` (mirroring `microsoft/agent-framework/dotnet/samples/03-workflows/Declarative/FunctionTools/FunctionTools.yaml`).
2. In `Program.cs`, register the target Azure agent with a tool wrapped in `ApprovalRequiredAIFunction`, e.g.:

    ```csharp
    var issueRefund = AIFunctionFactory.Create(IssueRefund);
    var approvalRequired = new ApprovalRequiredAIFunction(issueRefund);
    // register approvalRequired as a tool on the agent definition
    ```

3. Deploy to a Foundry hosted-agent project (`azd up` or equivalent).
4. Invoke the agent with a prompt that triggers the tool (e.g. "Issue a refund for order 123").

### Observed

- **Turn 1**: agent emits `function_call` for `IssueRefund`. `ApprovalRequiredAIFunction` halts; bridge surfaces `mcp_approval_request{name=IssueRefund, arguments=...}` to the client. ✅
- **Turn 2**: client POSTs `mcp_approval_response{approve=true}`. Server returns:

    ```
    HTTP 400 BadRequest
    No tool output found for function call call_J9Y...
    ```

### Expected

- Turn 2 resumes the agent loop with the approval result, the tool actually executes, and the assistant returns a final reply that incorporates the tool output.

## Root cause (working hypothesis)

[...]

Issue #5546, opened 2026-04-28
Python: [Bug]: Hosted toolbox MCP call_id from a previous turn is replayed as function_call_output without its function_call (400 from Responses API)


cristofima:
### What happened?

When a Foundry-hosted agent uses a toolbox MCP tool that completed successfully on a previous turn within the same `conversation_id`, the next turn fails with a 400 from the Azure OpenAI Responses API:

```
openai.BadRequestError: Error code: 400 - {
  'error': {
    'message': "No tool call found for function call output with call_id mcp_06b686e11f118cf40169f0e5badb3081979842929d5cf04920.",
    'type': 'invalid_request_error',
    'param': 'input',
    'code': None
  }
}
```

The error is raised inside `agent_framework_openai/_chat_client.py:555` when calling `client.responses.create(stream=True, **run_options)`. The `input` array sent to the Responses API contains a `function_call_output` whose `call_id` has no matching `function_call` in the same array, which violates the Responses API contract. The 400 happens before the model is invoked.

The same `conversation_id` is reused across turns, while `caresp_*` ids change. A retry of the failing prompt produces the exact same 400 with the same orphan `call_id`, which confirms the orphan is now persisted in conversation state and is replayed every time.

### What did you expect to happen?

A subsequent user message in the same conversation should be processed normally. Either:

- the framework should not replay a `function_call_output` from a prior turn without its matching `function_call` in the same `input` payload, or
- the hosted MCP tool calls should be persisted as a balanced pair (call + output) so the Responses API accepts them on replay.

### Steps to reproduce the issue

1. Build a Foundry hosted agent with `agent-framework-foundry-hosting==1.0.0a260428` and `agent-framework-core==1.2.1` / `agent-framework-foundry==1.2.1`.
2. Wire a Foundry Toolbox MCP connector backed by an OAuth-protected MCP server (in our case Logic Apps with Entra ID auth code + PKCE).
3. From the Foundry playground, send three prompts in the same conversation:
   - **Prompt 1**: a question that does not require the tool. Succeeds.
   - **Prompt 2**: a question that triggers the MCP tool. Returns `oauth_consent_request`. After the user authorizes and resends, the tool runs and returns the result. Succeeds.
   - **Prompt 3**: any follow-up question. Fails with the 400 above before the model streams anything.

### Code Sample

```markdown
Minimal shape of the agent wiring (full project is a Foundry hosted agent, but the failure is reproducible with this skeleton):


[...]

PR #6247, opened 2026-06-01
.NET: Fix: OutputConverter.FunctionResultContent creates OutputItemFunction…

enhendrickson:
## Summary

When a Foundry-hosted agent invokes a tool, the `OutputConverter.FunctionResultContent` case creates `OutputItemFunctionToolCallOutput` items with a null `Id` field. This causes the Responses API storage layer to reject the persistence request with HTTP 500, followed by a 409 on retry (confirming partial resource creation). While `Azure.AI.AgentServer.Responses v1.0.0-beta.5` gracefully catches this and emits a `response.failed` event instead of breaking the SSE stream, the root cause remains unfixed.

## Root Cause

In [OutputConverter.cs](https://github.com/microsoft/agent-framework/blob/main/dotnet/src/Microsoft.Agents.AI.Foundry.Hosting/OutputConverter.cs) (~line 450-460), the `FunctionResultContent` case:

```csharp
case FunctionResultContent functionResult:
{
    var outputText = EncodeFunctionResultAsJsonStringPayload(functionResult.Result);
    var itemId = GenerateItemId("fc");  // WRONG: "fc" is function_call prefix, not function_call_output
    var outputItem = new OutputItemFunctionToolCallOutput(
        functionResult.CallId,
        BinaryData.FromString(outputText));  // Id property stays null (get-only)
    var outputBuilder = stream.AddOutputItem<OutputItemFunctionToolCallOutput>(itemId);
    yield return outputBuilder.EmitAdded(outputItem);
    yield return outputBuilder.EmitDone(outputItem);
    break;
}
```

**Problem:** The 2-arg constructor `OutputItemFunctionToolCallOutput(callId, output)` doesn't set the `Id` property (it's get-only). When the item is serialized for persistence (using `ModelReaderWriterOptions.Json`), the `id` field is written as `null`, which violates the Responses API schema and causes HTTP 500.

**Why it happens:** The 2-arg public constructor is for deserialization. The internal 9-arg constructor (used by the SDK's convenience method) has the proper id setup:
```csharp
internal OutputItemFunctionToolCallOutput(
    string id,
    string callId,
    BinaryData output,
    ... other fields ...)
```

## Reproduction

1. Deploy a Foundry hosted agent with Foundry.Hosting 1.8.0-preview
2. Register any tool via `AddFoundryToolboxes()` or `AsOpenAIResponseTool()`
3. Invoke the agent with a prompt that triggers the tool
4. Observe in logs:
   - Initial POST `/storage/responses` returns HTTP 500
   - Azure.Core auto-retry with same request-id returns HTTP 409 (resource already exists)
   - SSE stream terminates (beta.4) or emits `response.failed` event (beta.5)

**Smoke test evidence** (from this session):
```
17:15:13.534927401Z  POST /storage/responses ... HTTP 500
17:15:13.535XXX...Z  Azure.Core retry with request-id d08ad4aa-17aa-42e5-9ae6-03b1ebb857c3
17:15:13.536XXX...Z  HTTP 409 Conflict (resource already exists)
17:15:13.537XXX...Z  Persistence failed before terminal event for response caresp_...
```

## Expected Behavior

Tool invocations should persist cleanly and the agent should return results to the client without storage errors.

## Current Workaround (beta.5)

Upgrade to `Azure.AI.AgentServer.Responses v1.0.0-beta.5`, which contains `ResponseOrchestrator.CreateStreamingAsync` exception handling that catches persistence failures and emits `response.failed` with code "storage_error". The SSE stream completes gracefully instead of terminating abruptly.

```xml
<PackageReference Include="Azure.AI.AgentServer.Responses" Version="1.0.0-beta.5" />
```

Client-side mitigation: listen for `response.failed` events with `code="storage_error"` and display a user-friendly message.

## Suggested Fix

Replace the manual 2-arg constructor usage with the SDK's convenience method, which handles id generation correctly:

```csharp
case FunctionResultContent functionResult:
{
    var outputText = EncodeFunctionResultAsJsonStringPayload(functionResult.Result);
    foreach (var evt in stream.OutputItemFunctionCallOutput(
        functionResult.CallId,
        BinaryData.FromString(outputText)))
    {
        yield return evt;
    }
    break;
}
```

The convenience method uses `IdGenerator.NewFunctionCallOutputItemId()` internally, which generates proper "fco_{partitionKey}{entropy}" formatted ids.

## Code Location

- **File:** `dotnet/src/Microsoft.Agents.AI.Foundry.Hosting/OutputConverter.cs`
- **Class:** `OutputConverter`
- **Method:** (various overloads of the iterator that yield from `FunctionResultContent` case)
- **Lines:** ~450-460 (exact line number may vary with version)

## Environment

- **Microsoft.Agents.AI.Foundry.Hosting:** 1.8.0-preview.260528.1
- **Azure.AI.AgentServer.Responses:** 1.0.0-beta.4 (buggy) → 1.0.0-beta.5 (graceful error)
- **Azure.AI.Projects:** 2.1.0-beta.1
- **Foundry Hosted Agent:** `/responses` protocol (streaming SSE)

## Related

- Similar pattern (orphan function_call items) fixed in #5662 via deferred emission
- MCP-specific function_call_output pairing fixed in #5546
- Multi-turn tool-call pairing issue #5941

I have submitted a ticket for this: #6245


moonbox3 (commented):
Is there a unit test we should be adding/updating related to this change?

enhendrickson:
@microsoft-github-policy-service agree company="Provisions Group"

enhendrickson:
> Is there a unit test we should be adding/updating related to this change?

good question, I just amended/pushed the one that was there @moonbox3

enhendrickson:
I also accepted someone else's changes that came in via MC - I'm guessing it was an agent coded solution to my bug, but either way, it was G2G... this PR is now ready to merge with both the fixed bug and updated test

alliscode:
@enhendrickson thanks for creating this. I believe the issue was already fixed by this PR: https://github.com/microsoft/agent-framework/pull/6246 and is included in the latest release. Please reopen this if you think it's not resolved.

PR #6257, opened 2026-06-02
Python: fix: handle string response in _get_metadata_from_chat_response (Closes #6234)

### LINKED ISSUE #6234: Python: [Bug]: OpenAIChatCompletionClient fails with 'str' object has no attribute 'system_fingerprint' when provider returns non-OpenAI conformant response
### What happened?
If the OpenAI Chat Completions provider returns a response body that is a **plain string** (rather than a ChatCompletion object, as expected by the framework client), the following stack trace will result:

```
Exception: ChatClientException: ("<class 'agent_framework_openai._chat_completion_client.OpenAIChatCompletionClient'> service failed to complete the prompt: 'str' object has no attribute 'system_fingerprint'", AttributeError("'str' object has no attribute 'system_fingerprint'"))
  ...
  File "agent_framework_openai/_chat_completion_client.py", line 687, in _parse_response_from_openai
    response_metadata = self._get_metadata_from_chat_response(response)
  File "agent_framework_openai/_chat_completion_client.py", line 791, in _get_metadata_from_chat_response
    "system_fingerprint": response.system_fingerprint,
AttributeError: 'str' object has no attribute 'system_fingerprint'
```

### What did you expect to happen?
A robust error explaining that the backend provider returned an invalid response, instead of an internal AttributeError. For OpenAI-incompatible provider responses, the client should surface an actionable error, not crash.

### Steps to reproduce the issue
1. Point agent-framework Python clients (OpenAIChatCompletionClient) to an OpenAI-compatible proxy/gateway or a backend that occasionally returns payloads that are not strictly compliant with OpenAI's response schema (such as returning plain text for errors, HTML error pages, or unwrapped response bodies).
2. Cause the provider to return a plain string or body that cannot be decoded into a ChatCompletion object.
3. Submit a prompt to the agent, or cause the agent to create a chat completion.
4. The client will raise AttributeError as described above.

### Remediation suggestion
Add a defensive type check or a `getattr` guard for `system_fingerprint` in `_get_metadata_from_chat_response`, and surface non-object provider responses as clear errors (with payload included for debugging).

This pattern was already used for other non-conformant response fields (see code handling streaming "delta": null, etc.).

---

**Note**: No existing issues matched this pattern in this repo as of June 2026. Let me know if more info is needed or for a PR.

### Code Sample

```markdown
Using the OpenAIChatCompletionClient against a non-OpenAI backend (proxy/gateway/custom deploy) that sometimes returns a plain string:


from agent_framework_openai import OpenAIChatCompletionClient
client = OpenAIChatCompletionClient(base_url="https://mynonopenai-endpoint/v1")
# Simulate a non-conformant backend response
response = await client.get_response([{"role": "user", "content": "hi"}])
# The request triggers the above error.
```

```markdown
Exception: ChatClientException: ("<class 'agent_framework_openai._chat_completion_client.OpenAIChatCompletionClient'> service failed to complete the prompt: 'str' object has no attribute 'system_fingerprint'", AttributeError("'str' object has no attribute 'system_fingerprint'"))
File "/home/site/wwwroot/.python_packages/lib/site-packages/agent_framework_openai/_chat_completion_client.py", line 554, in _get_response
return self._parse_response_from_openai( ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/home/site/wwwroot/.python_packages/lib/site-packages/agent_framework_openai/_chat_completion_client.py", line 687, in _parse_response_from_openai
response_metadata = self._get_metadata_from_chat_response(response)
File "/home/site/wwwroot/.python_packages/lib/site-packages/agent_framework_openai/_chat_completion_client.py", line 791, in _get_metadata_from_chat_response
"system_fingerprint": response.system_fingerprint,
AttributeError: 'str' object has no attribute 'system_fingerprint'
```

### Package Versions

agent-framework-openai: (latest main as of June 2026)
agent-framework-core: (latest main as of June 2026)

### Python Version

Python 3.12

### Additional Context

Can provide raw response if needed. Problem is easy to repro against OpenAI-compatible backends that serve non-spec responses (see discussion above).

kuishou68:
Closes #6234

## Problem

When a non-conformant OpenAI-compatible backend returns a plain string instead of a `ChatCompletion` object, `_parse_response_from_openai()` passes it directly to `_get_metadata_from_chat_response()`, which crashes with:

```
AttributeError: 'str' object has no attribute 'system_fingerprint'
```

## Fix

Two defensive changes:

1. **`_parse_response_from_openai()`**: Added an `isinstance(response, str)` check at the top of the method. If the response is a string, a descriptive `ChatClientException` is raised with the invalid payload included for debugging — following the same pattern used elsewhere in this file for error handling.

2. **`_get_metadata_from_chat_response()`**: Changed `response.system_fingerprint` to `getattr(response, "system_fingerprint", None)` as defense-in-depth, so even if an unexpected non-ChatCompletion object slips through, it won't crash with an `AttributeError`.

## Testing

- The change follows existing error-handling patterns already used in `_parse_response_from_openai` (see `BadRequestError` handling at lines 557-571).
- The `getattr` guard is consistent with how `logprobs` is already handled in `_get_metadata_from_chat_choice` (line 810).

Co-authored-by: lingxiu58 <86288566+lingxiu58@users.noreply.github.com>
Co-authored-by: pojian68 <232320289+pojian68@users.noreply.github.com>

### DIFF
--- a/python/packages/openai/agent_framework_openai/_chat_completion_client.py
+++ b/python/packages/openai/agent_framework_openai/_chat_completion_client.py
@@ -684,6 +684,13 @@ class RawOpenAIChatCompletionClient(  # type: ignore[misc]
 
     def _parse_response_from_openai(self, response: ChatCompletion, options: Mapping[str, Any]) -> ChatResponse:
         """Parse a response from OpenAI into a ChatResponse."""
+        if isinstance(response, str):
+            raise ChatClientException(
+                maybe_append_azure_endpoint_guidance(
+                    f"{type(self)} service returned an invalid response (str instead of ChatCompletion): {response!r}",
+                    azure_endpoint=self.azure_endpoint,
+                ),
+            )
         response_metadata = self._get_metadata_from_chat_response(response)
         messages: list[Message] = []
         finish_reason: FinishReason | None = None
@@ -788,7 +795,7 @@ class RawOpenAIChatCompletionClient(  # type: ignore[misc]
     def _get_metadata_from_chat_response(self, response: ChatCompletion) -> dict[str, Any]:
         """Get metadata from a chat response."""
         return {
-            "system_fingerprint": response.system_fingerprint,
+            "system_fingerprint": getattr(response, "system_fingerprint", None),
         }
 
     def _get_metadata_from_streaming_chat_response(self, response: ChatCompletionChunk) -> dict[str, Any]:


python/packages/openai/agent_framework_openai/_chat_completion_client.py
```diff
@@ -788,7 +795,7 @@ def _parse_text_from_openai(self, choice: Choice | ChunkChoice) -> Content | Non
     def _get_metadata_from_chat_response(self, response: ChatCompletion) -> dict[str, Any]:
         """Get metadata from a chat response."""
         return {
-            "system_fingerprint": response.system_fingerprint,
+            "system_fingerprint": getattr(response, "system_fingerprint", None),
```
moonbox3, line 798: Should we mirror this onto the streaming path too? The fix hardens non-streaming, but here `response.system_fingerprint` is still a direct deref, and `_parse_response_update_from_openai` has no str guard. A provider returning the same malformed payload while streaming hits the cryptic `AttributeError` again. The outer `except Exception` still wraps it into ChatClientException, so it's loud, just without the specific "str instead of ..." message this PR adds for non-streaming. Same contract on both paths?

python/packages/openai/agent_framework_openai/_chat_completion_client.py
```diff
@@ -684,6 +684,13 @@ def _prepare_options(self, messages: Sequence[Message], options: Mapping[str, An
 
     def _parse_response_from_openai(self, response: ChatCompletion, options: Mapping[str, Any]) -> ChatResponse:
         """Parse a response from OpenAI into a ChatResponse."""
+        if isinstance(response, str):
```
moonbox3, line 687: Are we covering the new branch? Couldn't find a test exercising the str -> ChatClientException path in `python/packages/openai/tests/openai/test_openai_chat_completion_client.py`. Existing direct-call tests make it nearly free: `client._parse_response_from_openai("some string", {})` inside `pytest.raises(ChatClientException)`, asserting substring `"str instead of ChatCompletion"` rather than the full message (Azure suffix is env-dependent). Pins the contract so a refactor can't quietly regress it back to the AttributeError.

eavanvalkenburg:
This is a duplicate of #6270 that solves for two issues, closing this one.

PR #6261, opened 2026-06-02
Python: Fix spurious Magentic custom manager warning

### LINKED ISSUE #4306: Python: MagenticBuilder logs spurious custom-manager warning with defaults
### Description
`MagenticBuilder` always emits a misleading warning when a custom manager is provided, even with no extra manager args.

In `_set_manager`, `_log_warning_if_constructor_args_provided` checks `max_stall_count`, which defaults to `3` (non-None), so the warning always fires:
> "Custom manager provided; all other manager arguments will be ignored."

### Code Sample
```python
from agent_framework_orchestrations import MagenticBuilder

# using any valid participants + custom manager
builder = MagenticBuilder(participants=[...], manager=my_manager)
# Actual: warning logged even when no constructor overrides were passed
# Expected: no warning unless override args were explicitly supplied
```

No exception. Incorrect warning behavior.

### Package Versions
agent-framework-orchestrations: source checkout at commit `02ba27493`

### Python Version
Python 3.13.5

### Additional Context
This can create false-positive warning noise in normal custom-manager usage.

moonbox3:
### Motivation and Context

Fixes #4306.

`MagenticBuilder` logged `Custom manager provided; all other manager arguments will be ignored.` whenever a custom `manager` or `manager_factory` was supplied, even when the caller did not pass any standard-manager options. This happened because `max_stall_count` defaulted to `3`, so the warning check could not distinguish an omitted value from an explicit override.

### Description

Updates `MagenticBuilder` to use the existing `_MISSING` sentinel for `max_stall_count` so omitted values remain distinguishable from explicitly supplied values. The builder resolves omitted `max_stall_count` to `3` only when constructing a `StandardMagenticManager`.

Adds regression coverage for custom manager and manager factory configurations that should not warn, preserves the warning when a standard-manager option is explicitly supplied with a custom manager, and verifies `manager_agent_factory` still receives the default `max_stall_count`.

### Contribution Checklist

### DIFF
--- a/python/packages/orchestrations/agent_framework_orchestrations/_magentic.py
+++ b/python/packages/orchestrations/agent_framework_orchestrations/_magentic.py
@@ -39,7 +39,7 @@ from ._base_group_chat_orchestrator import (
     ParticipantRegistry,
 )
 from ._participant_output_config import (
-    _MISSING,  # pyright: ignore[reportPrivateUsage]
+    UNSET,
     _coalesce_output_from,  # pyright: ignore[reportPrivateUsage]
     _coerce_intermediate_output_from,  # pyright: ignore[reportPrivateUsage]
     _ParticipantIntermediateOutputSelection,  # pyright: ignore[reportPrivateUsage]
@@ -1411,13 +1411,13 @@ class MagenticBuilder:
         task_ledger_plan_update_prompt: str | None = None,
         progress_ledger_prompt: str | None = None,
         final_answer_prompt: str | None = None,
-        max_stall_count: int | Sentinel = _MISSING,
+        max_stall_count: int | Sentinel = UNSET,
         max_reset_count: int | None = None,
         max_round_count: int | None = None,
         # Existing params
         enable_plan_review: bool = False,
         checkpoint_storage: CheckpointStorage | None = None,
-        output_from: Sequence[_ParticipantOutputSpecifier] | Literal["all"] | None = cast(Any, _MISSING),
+        output_from: Sequence[_ParticipantOutputSpecifier] | Literal["all"] | None = cast(Any, UNSET),
         intermediate_output_from: _ParticipantIntermediateOutputSelection = None,
     ) -> None:
         """Initialize the Magentic workflow builder.
@@ -1621,7 +1621,7 @@ class MagenticBuilder:
         progress_ledger_prompt: str | None = None,
         final_answer_prompt: str | None = None,
         # Limits
-        max_stall_count: int | Sentinel = _MISSING,
+        max_stall_count: int | Sentinel = UNSET,
         max_reset_count: int | None = None,
         max_round_count: int | None = None,
     ) -> None:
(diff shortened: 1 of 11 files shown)


python/packages/orchestrations/agent_framework_orchestrations/_magentic.py
```diff
@@ -1411,7 +1411,7 @@ def __init__(
         task_ledger_plan_update_prompt: str | None = None,
         progress_ledger_prompt: str | None = None,
         final_answer_prompt: str | None = None,
-        max_stall_count: int = 3,
+        max_stall_count: int = cast(Any, _MISSING),
```
eavanvalkenburg: [PEP661 is now final](https://peps.python.org/pep-0661/) so we should use that (from typing_extensions for now) to create the sentinel value, that is better then the `object()` approach used here.

eavanvalkenburg: also, personally I prefer a name like `UNSET`, instead of MISSING and since users will see this, as it is part of the init, we shouldn't have a underscore

moonbox3: Fixed in `32ac3da0`.

moonbox3: The agent couldn't figure out the update from MISSING -> UNSET. :)

python/packages/orchestrations/agent_framework_orchestrations/_magentic.py
```diff
@@ -1411,7 +1411,7 @@ def __init__(
         task_ledger_plan_update_prompt: str | None = None,
         progress_ledger_prompt: str | None = None,
         final_answer_prompt: str | None = None,
-        max_stall_count: int = 3,
+        max_stall_count: int | Sentinel = _MISSING,
```
moonbox3: Should be `UNSET` with no `_`

python/packages/orchestrations/agent_framework_orchestrations/_magentic.py
moonbox3: Fixed in `32ac3da0`.

python/packages/orchestrations/agent_framework_orchestrations/_magentic.py
moonbox3: Fixed

python/packages/orchestrations/agent_framework_orchestrations/_magentic.py
```diff
@@ -1411,13 +1411,13 @@ def __init__(
...
         final_answer_prompt: str | None = None,
-        max_stall_count: int = 3,
+        max_stall_count: int | Sentinel = UNSET,
         max_reset_count: int | None = None,
         max_round_count: int | None = None,
         # Existing params
         enable_plan_review: bool = False,
         checkpoint_storage: CheckpointStorage | None = None,
-        output_from: Sequence[_ParticipantOutputSpecifier] | Literal["all"] | None = cast(Any, _MISSING),
+        output_from: Sequence[_ParticipantOutputSpecifier] | Literal["all"] | None = cast(Any, UNSET),
```
eavanvalkenburg, line 1420: Do we need this cast?

eavanvalkenburg (approved):

giles17 (approved):

PR #6296, opened 2026-06-03
Python: [Generated by SRE Agent] Fix MCP allowed_tools empty list handling

chetantoshniwal:
## Summary

When `allowed_tools` is set to an empty list `[]`, the existing falsy check (`if not self.allowed_tools`) incorrectly treats it as unconfigured (same as `None`), causing **all** MCP tools to be exposed instead of none.

## Changes

- **`_mcp.py`**: Changed `if not self.allowed_tools:` to `if self.allowed_tools is None:` so that an empty list correctly results in no tools being available.
- **`test_mcp.py`**: Added test case for empty list `[]` to the existing parametrized `test_mcp_tool_allowed_tools` test.

## Behavior

| `allowed_tools` value | Before (bug) | After (fix) |
|------------------------|-------------|-------------|
| `None` | All tools exposed | All tools exposed (unchanged) |
| `[]` | All tools exposed ❌ | No tools exposed ✅ |
| `["tool_a"]` | Only tool_a | Only tool_a (unchanged) |


giles17 (approved):

python/packages/core/agent_framework/_mcp.py
```diff
@@ -544,7 +544,7 @@ def _prepare_message_for_mcp(
     @property
     def functions(self) -> list[FunctionTool]:
         """Get the list of functions that are allowed."""
-        if not self.allowed_tools:
+        if self.allowed_tools is None:
```
eavanvalkenburg, line 553: the docstring doesn't actually specify either way, what a empty list means, so let's clarify the docstring for the `allowed_tools` param to ensure this change is clear.

chetantoshniwal: Done in 20b50ed — clarified the `allowed_tools` docstring across all four classes (`MCPTool`, `MCPStdioTool`, `MCPStreamableHTTPTool`, `MCPWebsocketTool`) so `None` (expose all), `[]` (expose none), and a non-empty collection (expose listed) are each spelled out. Also reworded the existing line, which was a bit confusing on its own (read as "tools allowed to use this tool" rather than "tools this server exposes").

python/packages/core/agent_framework/_mcp.py
```diff
@@ -1592,7 +1596,11 @@ def __init__(
                 - A dict with keys `always_require_approval` or `never_require_approval`,
                   followed by a sequence of strings with the names of the relevant tools.
                 A tool should not be listed in both, if so, it will require approval.
-            allowed_tools: A list of tools that are allowed to use this tool.
+            allowed_tools: Optional allow-list of MCP tool names to expose as functions.
+                ``None`` (the default) exposes every tool advertised by the MCP server.
+                An empty collection (``[]``) exposes no tools, which is useful when the
```
eavanvalkenburg: reading this makes me think, that you should just set `load_tools` to False if you do not want to expose tools, let's add that to the docstring here. we can leave the behavior in as a guard or when people want to inspect the tools, but not actually use them.

chetantoshniwal: Good call — done in f83b940. Reframed the `allowed_tools` docstring across all four classes to recommend `load_tools=False` for the "no tools at all" case, and called out `[]` as a runtime guard / inspection-only path. Behavior is unchanged (`[]` still exposes nothing); the docstring now points readers at the cleaner idiom first.

eavanvalkenburg (approved):

Issue #6265, opened 2026-06-02
Python: Channels


eavanvalkenburg:
Tracking issue for the Python **hosting protocol-helper** work (formerly tracked as “channels”).

The accepted direction is helper-first: protocol packages convert protocol-native input/output into Agent Framework run values and protocol-native responses/events. App code owns the web framework routes, native SDK clients, authN/authZ, commands, delivery, deployment shape, and session/checkpoint storage policy.

The earlier host/channel registry model is no longer the v1 boundary. Issue titles that still say “channel” should be read as “protocol helper package + samples” until each slice is renamed or rewritten.

## Current Sprint 5 work

- [x] #6585 — Core state + Responses protocol helpers — PR #6891 🔍
- [x] #6588 — Telegram protocol helpers and samples — follow-up after #6891 🔍

## Merged / retained prior work

- [x] #6584 — ADRs and implementation spec — original PR #6578; helper-first ADR update in #6837 ✅
- [x] #6586 — Minimal Foundry Hosted Agent helper — PR #6593 ✅

## Remaining protocol-helper slices

- [x] #6587 — Invocations helpers and Foundry hosted agent sample
- [ ] #6589 — Activity Protocol helpers
- [ ] #6590 — Discord helpers
- [x] #6591 — A2A helpers and sample
- [x] #6592 — MCP helpers and sample

## Split rules

- Each protocol slice should provide only the Agent Framework-specific seam: helpers such as `<protocol>_to_run(...)`, `<protocol>_from_run(...)`, `<protocol>_from_streaming_run(...)`, `<protocol>_session_id(...)`, and command/media helpers when useful.
- App/web-framework code owns routes, auth, native SDK clients, webhook registration, background work, deployment shape, and response delivery.
- Samples ship with the protocol helper package they use, or with the last not-yet-landed helper package when a sample spans multiple protocols.
- New non-core package tests must use `tests/<unique_name>/` and must not include `tests/__init__.py`.
- New packages should follow the current typing setup: Pyright for source, no package-local source mypy task.
- Do not reintroduce the old host/channel registry, route aggregation, linking, multicast, cross-channel delivery, or durable delivery implementation into the v1 helper surface.
- If an older issue or PR references `feature/python-hosting` or the old host/channel API, reconcile it against the helper-first model before implementation.

## Follow-up / proposed work

Tracked by ADR-0028 rather than v1 implementation PRs:

- Identity linking and unlink flows
- Multicast and cross-channel delivery
[...]

PR #6305, opened 2026-06-03
Python: add agent-framework-hosting-mcp channel

eavanvalkenburg:
### Motivation and Context

Part of the hosting channels work tracked in #6265.

The host can already expose its target (an agent or workflow) over several channels (Activity protocol, Telegram, Discord, Teams, Responses). This adds a channel that exposes the same target as a **Model Context Protocol (MCP) tool**, so any MCP client can invoke the hosted agent/workflow as a single tool over Streamable HTTP.

### Description

Adds the `agent-framework-hosting-mcp` package providing `MCPChannel`:

- Builds an MCP `Server` + `StreamableHTTPSessionManager` and mounts it on the host's ASGI app (default path `/mcp`), with the session-manager lifecycle wired into host startup/shutdown.
- Exposes the host target as one MCP tool. Invocations route through the host pipeline (`ChannelContext.run` / `run_stream`) so sessions, cross-channel linking, and run/response hooks all apply.
- Maps the MCP request context to a `ChannelSession` isolation key and a `ChannelIdentity`; streaming output is forwarded as MCP progress notifications.

Includes unit tests (using an in-memory MCP client), a README, and workspace registration in the root `pyproject.toml`/`uv.lock`.

### Contribution Checklist

### DIFF
--- a/python/packages/hosting-mcp/agent_framework_hosting_mcp/_channel.py
+++ b/python/packages/hosting-mcp/agent_framework_hosting_mcp/_channel.py
@@ -18,11 +18,16 @@ typed inputs).
 
 from __future__ import annotations
 
-from collections.abc import Mapping
+import base64
+import json
+import re
+from collections.abc import Mapping, Sequence
 from contextlib import AbstractAsyncContextManager
-from typing import Any
+from dataclasses import asdict, is_dataclass
+from typing import Any, cast
 
 import mcp.types as types
+from agent_framework import Content, Message
 from agent_framework_hosting import (
     ChannelContext,
     ChannelContribution,
@@ -36,6 +41,7 @@ from agent_framework_hosting import (
 )
 from mcp.server.lowlevel import Server
 from mcp.server.streamable_http_manager import StreamableHTTPSessionManager
+from pydantic import AnyUrl
 from starlette.routing import Mount
 from starlette.types import Receive, Scope, Send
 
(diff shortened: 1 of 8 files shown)


python/packages/hosting-mcp/agent_framework_hosting_mcp/_channel.py
```diff
@@ -0,0 +1,282 @@
...
+    :meth:`agent_framework.Agent.as_mcp_server`).
+    """
+
+    name = "mcp"
+
+    def __init__(
+        self,
+        *,
+        path: str = "/mcp",
+        tool_name: str = _DEFAULT_TOOL_NAME,
```
eavanvalkenburg, line 189: the name should come from the agent name

python/packages/hosting-mcp/agent_framework_hosting_mcp/_channel.py
```diff
@@ -0,0 +1,282 @@
...
+        await self._server.request_context.session.send_progress_notification(
+            progress_token=progress_token,
+            progress=progress,
+            message=message,
+            related_request_id=request_id,
+        )
+
+    def _result_to_content(self, result: HostedRunResult[Any]) -> list[types.ContentBlock]:
+        """Convert a host result into MCP tool content (text only)."""
+        text = getattr(result.result, "text", None)
```
eavanvalkenburg: we have richer types then just text both in the agent run result and the output of a MCP tool, let's use them, that should also include structured outputs which MCP supports

python/packages/hosting-mcp/agent_framework_hosting_mcp/_channel.py
```diff
@@ -0,0 +1,282 @@
...
+        """Capture the host context and mount the Streamable-HTTP transport."""
+        self._ctx = context
+        self._server = self._build_server()
+        self._session_manager = StreamableHTTPSessionManager(
+            app=self._server,
+            json_response=self._json_response,
+            stateless=self._stateless,
+        )
+        return ChannelContribution(
+            routes=[Mount(self.path, app=self._handle_asgi)],
```
eavanvalkenburg: could we forgo this Mount setup altogether, by just manually exposing the `list_tool` and `call_tool` routes?

python/packages/hosting-mcp/tests/hosting_mcp/test_channel.py
eavanvalkenburg, line 1: we need integration tests, that spin up this host and connects to it with a MCP client

PR #6306, opened 2026-06-03
Python: add agent-framework-hosting-a2a channel

eavanvalkenburg:
### Motivation and Context

Part of the hosting channels work tracked in #6265.

The host can already expose its target (an agent or workflow) over several channels (Activity protocol, Telegram, Discord, Teams, Responses). This adds a channel that exposes the same target as a peer agent over the **Agent-to-Agent (A2A) protocol**, so A2A clients can discover and call the hosted agent/workflow.

### Description

Adds the `agent-framework-hosting-a2a` package providing `A2AChannel`:

- Serves an A2A agent card (default derived from the host target's name/description) plus JSON-RPC routes on the host's ASGI app.
- Handles requests with a host-routed `HostAgentExecutor` that drives the host pipeline (`ChannelContext.run` / `run_stream`) rather than wrapping the target directly, so sessions, cross-channel linking, and run/response hooks all apply.
- Maps the A2A context id to a `ChannelSession` isolation key and the caller to a `ChannelIdentity`. Non-streaming runs publish reply messages as task status updates; streaming runs publish incremental task artifacts.

Includes unit tests, a README, and workspace registration in the root `pyproject.toml`/`uv.lock`.

### Contribution Checklist

### DIFF
--- a/python/packages/hosting-a2a/agent_framework_hosting_a2a/_channel.py
+++ b/python/packages/hosting-a2a/agent_framework_hosting_a2a/_channel.py
@@ -1,0 +1,142 @@
+# Copyright (c) Microsoft. All rights reserved.
+
+"""A2A (Agent-to-Agent) channel for :mod:`agent_framework_hosting`.
+
+Exposes the hosted target as an A2A peer agent: it publishes an agent card and
+JSON-RPC routes, and drives every request through the host pipeline via
+:class:`HostAgentExecutor`.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+from typing import Any
+
+from a2a.server.request_handlers import DefaultRequestHandler
+from a2a.server.tasks import InMemoryTaskStore
+from a2a.types import AgentCapabilities, AgentCard, AgentInterface, AgentSkill
+from agent_framework_hosting import (
+    ChannelContext,
+    ChannelContribution,
+    ChannelResponseHook,
+    ChannelRunHook,
+)
+
+from ._executor import HostAgentExecutor
+
+try:
+    from a2a.server.routes import create_agent_card_routes, create_jsonrpc_routes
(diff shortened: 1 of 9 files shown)


python/packages/hosting-a2a/agent_framework_hosting_a2a/_channel.py
```diff
@@ -0,0 +1,142 @@
...
+    ChannelContext,
+    ChannelContribution,
+    ChannelResponseHook,
+    ChannelRunHook,
+)
+
+from ._executor import HostAgentExecutor
+
+try:
+    from a2a.server.routes import create_agent_card_routes, create_jsonrpc_routes
```
eavanvalkenburg: this dependency is part of the package right? Then we should also be able to import this without condition.

python/packages/hosting-a2a/agent_framework_hosting_a2a/_channel.py
```diff
@@ -0,0 +1,142 @@
...
+        name = self._agent_name or getattr(target, "name", None) or self.name
+        description = self._agent_description or getattr(target, "description", None) or f"{name} (A2A)"
+        return AgentCard(
+            name=name,
+            description=description,
+            version=self._agent_version,
+            default_input_modes=["text"],
+            default_output_modes=["text"],
+            capabilities=AgentCapabilities(streaming=self._streaming),
+            supported_interfaces=[AgentInterface(url=self._url, protocol_binding="JSONRPC")],
```
eavanvalkenburg: is there a way to enable other protocols?

python/packages/hosting-a2a/tests/hosting_a2a/test_channel.py
eavanvalkenburg, line 1: we also need a test that spins up the server and consumes it, preferable with the A2AAgent from the AF A2A package

PR #6310, opened 2026-06-03
Python: Fix per-service-call history persistence with server-storing clients

### LINKED ISSUE #5798: Python: OpenAI store=True can silently bypass external HistoryProvider persistence

eavanvalkenburg:
### Motivation and Context

When an `Agent` was configured with `require_per_service_call_history_persistence=True` together with a `HistoryProvider`, and the underlying chat client stored history server-side by default (e.g. `OpenAIChatClient`, where `STORES_BY_DEFAULT=True`), the external history provider was **silently never persisted**. The per-service-call middleware was skipped because the service was assumed to own history, and the once-per-run path also skipped the provider — so neither persisted.

Fixes #5798

### Description

**Unify persistence on the per-service-call middleware.** When `require_per_service_call_history_persistence=True` and a `HistoryProvider` exists, the `PerServiceCallHistoryPersistingMiddleware` is now always installed and owns persistence. `service_stores_history` only selects *how* the middleware behaves, never *whether* it persists:

- **Service does not store:** the middleware loads providers and drives the function loop with a local sentinel conversation id.
- **Service stores:** the middleware skips loading (the service owns history) and persists each service call while the real conversation id flows through. A warning is logged for any provider with `load_messages=True` whose load is bypassed.

The observable contract: with the flag on, persistence happens *per service call* — in a function-call → final-completion run, the function-call turn is persisted before the second call starts.

**Rationalize chat-options handling in `_prepare_run_context`:**

- `_merge_options` now skips `None` overrides and strips remaining `None` values in a single pass, so an unset `store` is never forwarded to the client and the service decides its own default (`STORES_BY_DEFAULT` is only an internal behavior hint).
- `store` and `conversation_id` are resolved once from a single combined view (`effective_options`) instead of probing both the agent-default and runtime dicts separately. The `InMemoryHistoryProvider` auto-injection and the per-service-call resolution now agree on `conversation_id` (an agent-level default is honored consistently).

Tests are added to `test_agents.py` as a scenario matrix (sync + streaming) that asserts the per-service-call persistence *timing* across storing/non-storing clients and `store` overrides.

### Contribution Checklist

### DIFF
--- a/python/packages/core/agent_framework/_agents.py
+++ b/python/packages/core/agent_framework/_agents.py
@@ -700,7 +700,9 @@ class RawAgent(BaseAgent, Generic[OptionsCoT]):  # type: ignore[misc]
                 drives the function loop with a local conversation; if it does, loading is skipped
                 (the service-managed conversation is the source of truth) and the middleware only
                 persists. A warning is logged for providers with ``load_messages=True`` when
-                loading is skipped because service-side storage is active.
+                loading is skipped because service-side storage is active. When no HistoryProvider
+                is present, this flag has no effect (no middleware is installed and nothing is
+                persisted).
             default_options: A TypedDict containing chat options. When using a typed agent like
                 ``Agent[OpenAIChatOptions]``, this enables IDE autocomplete for
                 provider-specific options including temperature, max_tokens, model,
--- a/python/packages/core/agent_framework/_agents.py
+++ b/python/packages/core/agent_framework/_agents.py
@@ -92,12 +92,16 @@ OptionsCoT = TypeVar(
 def _merge_options(base: dict[str, Any], override: dict[str, Any]) -> dict[str, Any]:
     """Merge two options dicts, with override values taking precedence.
 
+    ``None`` is treated as "unset": ``None`` overrides are skipped so they don't clobber a base
+    value, and the merged result is stripped of any remaining ``None`` values in a final pass so
+    unset options are never forwarded (e.g. an unset ``store`` is left for the service to default).
+
     Args:
         base: The base options dict.
         override: The override options dict (values take precedence).
 
     Returns:
-        A new merged options dict.
+        A new merged options dict containing no ``None`` values.
     """
     result = dict(base)
 
@@ -123,7 +127,7 @@ def _merge_options(base: dict[str, Any], override: dict[str, Any]) -> dict[str, 
             result["instructions"] = f"{result['instructions']}\n{value}"
         else:
             result[key] = value
-    return result
+    return {key: value for key, value in result.items() if value is not None}
 
 
 def _sanitize_agent_name(agent_name: str | None) -> str | None:
@@ -460,6 +464,9 @@ class BaseAgent(SerializationMixin):
         if provider_session is None and self.context_providers:
             provider_session = AgentSession()
 
+        # When per-service-call persistence is enabled, the per-service-call middleware owns
+        # HistoryProvider persistence (in both the local and service-managed cases), so skip
+        # them on the once-per-run path to avoid double persistence.
         per_service_call_history_required = self.require_per_service_call_history_persistence and any(
             isinstance(provider, HistoryProvider) for provider in self.context_providers
         )
@@ -686,11 +693,14 @@ class RawAgent(BaseAgent, Generic[OptionsCoT]):  # type: ignore[misc]
             description: A brief description of the agent's purpose.
             context_providers: Context providers to include during agent invocation.
             middleware: List of middleware to intercept agent and function invocations.
-            require_per_service_call_history_persistence: When True, history providers are invoked
-                around each model call instead of once per ``run()`` when the service
-                is not already storing history. If service-side storage is active for
-                the run, the agent skips local history providers and relies on the
-                service-managed conversation instead.
+            require_per_service_call_history_persistence: When True (and a HistoryProvider is
+                present), the provider always persists history via per-service-call middleware,
+                regardless of whether the client stores history server-side. If the client does
+                not store history, the middleware also loads providers around each model call and
+                drives the function loop with a local conversation; if it does, loading is skipped
+                (the service-managed conversation is the source of truth) and the middleware only
+                persists. A warning is logged for providers with ``load_messages=True`` when
+                loading is skipped because service-side storage is active.
             default_options: A TypedDict containing chat options. When using a typed agent like
                 ``Agent[OpenAIChatOptions]``, this enables IDE autocomplete for
                 provider-specific options including temperature, max_tokens, model,
@@ -791,22 +801,20 @@ class RawAgent(BaseAgent, Generic[OptionsCoT]):  # type: ignore[misc]
         self,
         *,
         session: AgentSession | None,
-        options: Mapping[str, Any] | None,
+        conversation_id: str | None,
         service_stores_history: bool,
     ) -> list[HistoryProvider]:
         history_providers = self._get_history_providers()
         if not self.require_per_service_call_history_persistence or not history_providers:
             return []
 
-        conversation_id = (
-            session.service_session_id
-            if session and session.service_session_id
-            else cast(str | None, (options or {}).get("conversation_id") or self.default_options.get("conversation_id"))
-        )
-        if service_stores_history:
-            return []
-
-        if conversation_id is not None:
+        # A live service-managed session id takes precedence over the resolved conversation id.
+        if session and session.service_session_id:
+            conversation_id = session.service_session_id
+        # Without service-side storage the middleware persists locally and drives the function
+        # loop with a local sentinel, which cannot be reconciled with an existing service-managed
+        # conversation. When the service stores history, an existing conversation id is expected.
+        if conversation_id is not None and not service_stores_history:
             raise AgentInvalidRequestException(
                 "require_per_service_call_history_persistence cannot be used "
                 "with an existing service-managed conversation."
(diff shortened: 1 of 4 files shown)


python/packages/core/agent_framework/_clients.py
```diff
@@ -597,10 +597,11 @@ def as_agent(
...
             middleware: List of middleware to intercept agent and function invocations.
-            require_per_service_call_history_persistence: Whether to require per-service-call
-                chat history persistence. When enabled, history providers are invoked around
-                each model call instead of once per ``run()`` when the service is not already
-                storing history.
+            require_per_service_call_history_persistence: When enabled (and a HistoryProvider is
+                present), the provider always persists history. If the client does not store
+                history server-side, history providers are invoked around each model call;
+                if it does, providers persist once per ``run()`` while loading is skipped and
+                the service-managed conversation is the source of truth.
```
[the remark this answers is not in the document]
eavanvalkenburg: Good catch — fixed in 1ee7950e6. The docstring now states that when the client stores history server-side the provider still persists after each model call; only provider *loading* is skipped (the service-managed conversation is the source of truth).

python/packages/core/agent_framework/_agents.py
```diff
@@ -686,11 +693,14 @@ def __init__(
...
-                the run, the agent skips local history providers and relies on the
-                service-managed conversation instead.
+            require_per_service_call_history_persistence: When True (and a HistoryProvider is
+                present), the provider always persists history via per-service-call middleware,
+                regardless of whether the client stores history server-side. If the client does
+                not store history, the middleware also loads providers around each model call and
+                drives the function loop with a local conversation; if it does, loading is skipped
+                (the service-managed conversation is the source of truth) and the middleware only
+                persists. A warning is logged for providers with ``load_messages=True`` when
+                loading is skipped because service-side storage is active.
```
TaoChenOSU: nit: this becomes sort of confusing. How about when this is true and there isn't a HistoryProvider?

python/packages/core/agent_framework/_agents.py
```diff
@@ -1188,10 +1212,30 @@ async def _prepare_run_context(
...
+        history_providers = self._get_history_providers()
+        if self.require_per_service_call_history_persistence and history_providers and service_stores_history:
+            for provider in history_providers:
+                if provider.load_messages:
+                    logger.warning(
+                        "HistoryProvider '%s' has load_messages=True but the chat client stores history "
+                        "server-side; skipping local history load and relying on the service-managed "
+                        "conversation. Set store=False to load from the provider, or load_messages=False "
+                        "to silence this warning.",
+                        provider.source_id,
```
TaoChenOSU, line 1238: Why not throw an exception at agent initialization?

python/packages/core/agent_framework/_sessions.py
```diff
@@ -652,7 +662,11 @@ async def _finalize_response(
         response: ChatResponse,
     ) -> ChatResponse:
         """Persist a model response and apply the local follow-up sentinel when needed."""
-        if response.conversation_id is not None and not is_local_history_conversation_id(response.conversation_id):
+        if (
```
moonbox3, line 668: What happens when `service_stores_history` is True but the client returns `conversation_id=None`? This guard is skipped, provider loading was already skipped (`_sessions.py` `_prepare_service_call_context`), and `session.service_session_id` only gets set when `conversation_id` is truthy (`_agents.py:1400`). So we persist this turn to the provider but the next run skips loading again with no service id to resume from, dropping cross-turn history silently with no error and no warning. Is a storing client guaranteed to always echo a `conversation_id`, or should we assert/warn in the storing branch when it comes back empty so the assumption can't fail quietly?

python/packages/core/agent_framework/_agents.py
```diff
@@ -791,22 +801,20 @@ def _resolve_per_service_call_history_providers(
...
-            return []
-
-        if conversation_id is not None:
+        # A live service-managed session id takes precedence over the resolved conversation id.
+        if session and session.service_session_id:
+            conversation_id = session.service_session_id
+        # Without service-side storage the middleware persists locally and drives the function
+        # loop with a local sentinel, which cannot be reconciled with an existing service-managed
+        # conversation. When the service stores history, an existing conversation id is expected.
+        if conversation_id is not None and not service_stores_history:
```
moonbox3, line 819: Are we covering the allow side of this boundary? The PR adds `and not service_stores_history`, so flag-on + storing client + an existing `conversation_id` now resumes instead of raising. Tests cover the raise side (`store=False` + `conversation_id`), but I couldn't find one for storing + `conversation_id` asserting it does NOT raise and that the id propagates to `service_session_id`. A regression dropping the `and not service_stores_history` qualifier would wrongly raise on the resume-a-stored-conversation path and nothing would catch it. Worth a test?

python/packages/core/agent_framework/_agents.py
```diff
@@ -1188,10 +1212,30 @@ async def _prepare_run_context(
...
+        # When require_per_service_call_history_persistence is set together with a
+        # HistoryProvider, the per-service-call middleware (installed below) always persists
+        # the provider. ``service_stores_history`` only selects how the middleware behaves:
+        # - service does not store: the middleware also loads providers and drives the function
+        #   loop with a local sentinel conversation id, or
+        # - service stores: the middleware skips loading (the service owns history) and simply
+        #   persists each service call while the real conversation id flows through.
+        # In the service-managed case loading is skipped, so warn for providers that expect to load.
+        history_providers = self._get_history_providers()
+        if self.require_per_service_call_history_persistence and history_providers and service_stores_history:
```
moonbox3, line 1230: This sits in `_prepare_run_context`, which runs once per `run()`, so a long-lived agent looping turns against a storing client with a `load_messages=True` provider re-logs the same WARNING every turn. Could that train users to tune it out? Wondering if we should gate it to once per session/agent (or first run only) so it stays a signal rather than per-turn noise.

python/packages/core/agent_framework/_agents.py
```diff
@@ -791,22 +801,20 @@ def _resolve_per_service_call_history_providers(
...
-            session.service_session_id
-            if session and session.service_session_id
-            else cast(str | None, (options or {}).get("conversation_id") or self.default_options.get("conversation_id"))
-        )
-        if service_stores_history:
-            return []
-
-        if conversation_id is not None:
+        # A live service-managed session id takes precedence over the resolved conversation id.
+        if session and session.service_session_id:
```
moonbox3, line 814: Are we covering a second run on the same session in storing mode? This precedence branch only fires once `service_session_id` is populated (run 2+), and every per-service-call test does a single run. A two-run test asserting persistence keeps happening, `service_session_id` stays stable, and loading stays skipped would lock down this path.

python/packages/core/agent_framework/_agents.py
eavanvalkenburg: Clarified in 8e09ec806: the docstring now states that when no HistoryProvider is present this flag has no effect (no middleware is installed and nothing is persisted).

python/packages/core/agent_framework/_agents.py
eavanvalkenburg: We intentionally keep this a runtime warning rather than raising at init: `store` is a runtime option (it can be overridden per `run()`), so whether the service stores history isn't known at construction time. A persist-only/store-only provider with a server-storing client is also a valid (if unusual) configuration we don't want to forbid — so a hard init exception would be both premature and potentially breaking. The warning surfaces the likely-misconfiguration without blocking it.

python/packages/core/agent_framework/_sessions.py
eavanvalkenburg: Addressed in 8e09ec806. PerServiceCallHistoryPersistingMiddleware._finalize_response now logs a WARNING every time the service is expected to store history but the response comes back with conversation_id=None — pointing out that cross-turn history may not resume and suggesting store=False to load/resume from the HistoryProvider instead. It's intentionally not deduped because this is an uncommon, easy-to-miss failure mode.

python/packages/core/agent_framework/_agents.py
eavanvalkenburg: Added test_psc_flag_on_storing_with_existing_conversation_id_does_not_raise in 8e09ec806: flag on + storing client + an existing conversation_id passed via options does not raise and the id propagates to session.service_session_id (mirrors the reject-side tests).

python/packages/core/agent_framework/_agents.py
eavanvalkenburg: We chose to keep this warning per-run (not deduped): settings like store can change between runs, so a later run may legitimately confirm the situation is resolved, and an unusual-but-allowed setup warrants a standing per-run signal rather than a one-time message that scrolls away.

python/packages/core/agent_framework/_agents.py
eavanvalkenburg: Added test_psc_flag_on_storing_two_runs_same_session in 8e09ec806: runs twice in storing mode on the same session, asserting persistence happens each run, service_session_id stays stable, and provider loading stays skipped (get_calls == 0).

TaoChenOSU (approved):

moonbox3 (approved):

PR #6311, opened 2026-06-03
.NET: Bug fixes for AGUI hosting and workflows

westey-m:
### Description

Bug fixes for AGUI hosting and workflows

### Contribution Checklist

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/ObjectModel/InvokeMcpToolExecutor.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/ObjectModel/InvokeMcpToolExecutor.cs
@@ -28,6 +28,12 @@ internal sealed class InvokeMcpToolExecutor(
     DeclarativeActionExecutor<InvokeMcpTool>(model, state)
 {
     /// <summary>
+    /// Snapshot of evaluated parameters at approval-request time.
+    /// Used to prevent TOCTOU attacks where state mutates during the approval window.
+    /// </summary>
+    private ApprovalSnapshot? _approvalSnapshot;
+
+    /// <summary>
     /// Step identifiers for the MCP tool invocation workflow.
     /// </summary>
     public static class Steps
@@ -75,6 +81,10 @@ internal sealed class InvokeMcpToolExecutor(
 
         if (requireApproval)
         {
+            // Snapshot the evaluated parameters to prevent TOCTOU attacks.
+            // If state mutates during the approval window, the approved values are used on resume.
+            this._approvalSnapshot = new ApprovalSnapshot(serverUrl, serverLabel, toolName, arguments, connectionName);
+
             // Create tool call content for approval request.
             // Transport headers (e.g. Authorization) are intentionally excluded from the
             // approval event: they must not cross into the externally-surfaced approval request.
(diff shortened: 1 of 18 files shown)


dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/ObjectModel/InvokeMcpToolExecutor.cs
```diff
@@ -27,6 +27,12 @@ internal sealed class InvokeMcpToolExecutor(
     WorkflowFormulaState state) :
     DeclarativeActionExecutor<InvokeMcpTool>(model, state)
 {
+    /// <summary>
+    /// Snapshot of evaluated parameters at approval-request time.
+    /// Used to prevent TOCTOU attacks where state mutates during the approval window.
+    /// </summary>
+    private ApprovalSnapshot? _approvalSnapshot;
+
```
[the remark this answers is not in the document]
westey-m: Addressed. The snapshot is now persisted to workflow state using the single-key container pattern (matching `AIAgentHostState`, `RoundRobinGroupChatManagerState`, etc.):

- `OnCheckpointingAsync` persists `_approvalSnapshot` to state via `QueueStateUpdateAsync`
- `OnCheckpointRestoredAsync` restores it via `ReadStateAsync<ApprovalSnapshot>`
- `CompleteAsync` clears the state key after successful execution
- Changed `ApprovalSnapshot` from `private readonly struct` → `internal sealed record` for clean serialization
- State is executor-scoped by the engine, addressing the concurrent runs concern

Also fixed the net472 build error (`ValueTask.CompletedTask` → `default(ValueTask)`) and added a checkpoint/restore test that verifies the snapshot survives the cycle.

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/ObjectModel/InvokeMcpToolExecutor.cs
```diff
@@ -137,13 +149,14 @@ public async ValueTask CaptureResponseAsync(
...
         }
 
-        // Approved - now invoke the tool
-        string serverUrl = this.GetServerUrl();
-        string? serverLabel = this.GetServerLabel();
-        string toolName = this.GetToolName();
-        Dictionary<string, object?>? arguments = this.GetArguments();
+        // Approved - use the snapshot from approval-request time to prevent TOCTOU attacks.
+        // Headers are re-evaluated (they may contain auth secrets that should not be persisted).
+        string serverUrl = this._approvalSnapshot?.ServerUrl ?? this.GetServerUrl();
```
peibekwe, line 154: `ApprovalSnapshot` isn't checkpoint-durable, and this line would probably silently fallback and re-open TOCTOU on resume. I'd recommend using `DurableProperty` to initialize `ApprovalSnapshot` i.e. `DurableProperty<ApprovalSnapshot> ...` (That will handle saving and restoring from context during checkpointing). We have an example of this in `QuestionExecutor`, so you can take a look at that. You can also add a checkpoint test to cover this hole.

When you switch to DurableProperty, then after successful invocation or failure, we will need to clear the values in the DurableProperty, to ensure that if the same executor is reused in something like `GoToAction`, the stale values won't exist.

peibekwe: nit: Also, with `DurableProperty`, we can throw if the snapshot doesn't exist instead of falling back to re-evaluating.

peibekwe: I just refreshed and see that you've addressed this using the `OnCheckpointRestoredAsync`. Ignore this :)

peibekwe (approved):

semenshi (approved):

dotnet/src/Microsoft.Agents.AI.Workflows.Declarative/ObjectModel/InvokeMcpToolExecutor.cs
westey-m: Thanks, I'll get it merged.

javiercn (approved):
AG-UI changes look good.

PR #6319, opened 2026-06-04
Python: MCP long-running task support in Python

### LINKED ISSUE #6258: Python: Port MCP LRO support to Python
Port the MCP LRO support that's implemented in C# to Python.

peibekwe:
### Motivation and Context

Ports the MCP long-running task (SEP-2663) feature from .NET to Python.  When an MCP server advertises a tool with `execution.taskSupport == "required"`, `MCPTool` now transparently drives the `tools/call` - `tasks/get` (polled) - `tasks/result` lifecycle, so the agent's function-calling loop sees a normal synchronous tool result.
 
This is purely additive, those tools could not complete via plain `tools/call` before, so the new path fixes a previously broken case without changing any existing behavior. Matches The work done in #5994 

Fixes #6258 

### Contribution Checklist


python/packages/core/agent_framework/_mcp.py
```diff
@@ -1363,35 +1496,280 @@ async def call_tool(self, tool_name: str, **kwargs: Any) -> str | list[Content]:
...
+        return None, legacy
+
+    async def _poll_task_until_terminal(self, task_id: str) -> types.GetTaskResult:
+        """Poll ``tasks/get`` until the task reaches a terminal status."""
+        from anyio import ClosedResourceError
+        from mcp import types
+        from mcp.shared.exceptions import McpError
+
+        while True:
+            for attempt in range(2):
```
moonbox3: What happens if the server never leaves `working`? `while True` has no overall deadline or attempt cap; only a terminal status exits, and poll delay clamps to <= 5s. `default_ttl` is sent as server-side retention (1593-1596), not a client wait bound, so a wedged or non-cooperative server hangs the tool call (and the whole agent turn) forever with no error and nothing above debug. Should we add a client deadline (like an `MCPTaskOptions` max-wait) that raises `ToolExecutionException` when exceeded?

python/packages/core/agent_framework/_mcp.py
```diff
@@ -1363,35 +1496,280 @@ async def call_tool(self, tool_name: str, **kwargs: Any) -> str | list[Content]:
...
+                try:
+                    request = types.ClientRequest(
+                        types.GetTaskRequest(params=types.GetTaskRequestParams(taskId=task_id))
+                    )
+                    # Use lenient Result then coerce: GetTaskResult.ttl is required by
+                    # schema but servers may legitimately omit it.
+                    lenient = await self.session.send_request(request, types.Result)  # type: ignore[union-attr]
+                    snapshot = self._coerce_get_task_result(lenient, task_id)
+                    break
+                except (ClosedResourceError, McpError) as ex:
```
moonbox3: Should a single slow `tasks/get` kill the whole task? With `request_timeout` set it becomes the session `read_timeout_seconds`, so one slow poll raises `McpError(REQUEST_TIMEOUT)`. `_is_connection_lost` only matches "session terminated", so that timeout (and any non-connection McpError) falls through here and aborts the entire long-running task, defeating the point of LRO. And this raises `ToolExecutionException`, not `CancelledError`, so the cancel arm at 1573-1576 never fires and the remote task is left running orphaned. Could a transient poll error keep polling instead, and could this path also fire best-effort `tasks/cancel`?

python/packages/core/agent_framework/_mcp.py
```diff
@@ -1363,35 +1496,280 @@ async def call_tool(self, tool_name: str, **kwargs: Any) -> str | list[Content]:
...
+        raw.pop("_meta", None)
+
+        task_field = raw.get("task")
+        if isinstance(task_field, dict):
+            task_id_val = cast(dict[str, Any], task_field).get("taskId")
+            if isinstance(task_id_val, str):
+                return task_id_val, None
+
+        try:
+            legacy = types.CallToolResult.model_validate(raw)
```
moonbox3, line 1700: Does this break the no-double-start invariant AGENTS.md describes? Phase 1 deliberately won't retry `tools/call` on connection loss because the server may have already started the op. Here the server accepted the call and returned a success-shaped (just unparseable) payload, yet we re-issue a plain `tools/call`, which would execute a side-effecting tool twice. Should an unparseable-but-successful response raise instead of silently re-calling?

python/packages/core/agent_framework/_mcp.py
moonbox3: Why hardcode `range(2)`? It encodes try-once-then-reconnect-once, with `attempt == 0` doubling as the may-reconnect sentinel and a `for...else` give-up at 1685-1686 that is unreachable (every path breaks/continues/raises, hence the `# pragma: no cover`). The same block is duplicated verbatim in `_fetch_task_result` (1758-1786). Could a shared `_send_with_one_reconnect(...)` helper replace the magic bound, the dead else, and the duplication? Bumping `range(2)` higher would also break the give-up message path, since it keys off the non-zero attempt.

python/packages/core/agent_framework/_mcp.py
```diff
@@ -149,6 +152,40 @@ def _url_origin(url: Any) -> tuple[str, str, int | None]:
...
+# Internal polling bounds for MCP long-running tasks. Not user-tunable today;
+# promote to MCPTaskOptions if a concrete need arises.
+_MCP_TASK_MIN_POLL_INTERVAL = timedelta(milliseconds=500)
+_MCP_TASK_MAX_POLL_INTERVAL = timedelta(seconds=5)
+_MCP_TASK_CANCEL_TIMEOUT = timedelta(seconds=5)
+_MCP_TASK_TERMINAL_STATUSES: frozenset[str] = frozenset({"completed", "failed", "cancelled", "input_required"})
+
+
+@experimental(feature_id=ExperimentalFeature.MCP_LONG_RUNNING_TASKS)
+@dataclass
```
moonbox3: Should `MCPTaskOptions` be frozen? `__post_init__` validates `default_ttl >= 0` only at construction, so `opts.default_ttl = timedelta(seconds=-5)` afterward silently bypasses it and sends a negative ttl. A single instance shared across tools (stdio/HTTP/ws all take `task_options`) can also be mutated mid-flight, since `default_ttl` is read at create time and `cancel_remote_task_on_local_cancellation` at cancel time. The setter already swaps the whole object, so freezing loses nothing.

python/packages/core/agent_framework/_mcp.py
```diff
@@ -1400,16 +1778,59 @@ async def call_tool(self, tool_name: str, **kwargs: Any) -> str | list[Content]:
...
+
+    async def _try_cancel_task(self, task_id: str) -> None:
+        """Send ``tasks/cancel`` swallowing every failure; bounded by ``_MCP_TASK_CANCEL_TIMEOUT``."""
+        from mcp import types
+
+        async def _send() -> None:
+            request = types.ClientRequest(types.CancelTaskRequest(params=types.CancelTaskRequestParams(taskId=task_id)))
+            try:
+                await self.session.send_request(request, types.CancelTaskResult)  # type: ignore[union-attr]
+            except Exception:
```
moonbox3: Should a failed cancel be this quiet? Inner `except Exception` logs at `debug`, then the outer `contextlib.suppress` swallows everything again, including the 5s `wait_for` timeout. If the cancel never lands (server down, timed out) the remote task keeps running and nothing above `debug` says cleanup failed, so "why is this task still running after I cancelled" yields nothing. Could we log the failure/timeout at `warning`, and drop the redundant outer `suppress(Exception)` the inner handler already covers?

python/packages/core/agent_framework/_mcp.py
```diff
@@ -1363,35 +1496,280 @@ async def call_tool(self, tool_name: str, **kwargs: Any) -> str | list[Content]:
...
+                )
+                break
+            except (ClosedResourceError, McpError) as ex:
+                if not self._is_connection_lost(ex):
+                    error_message = ex.error.message if isinstance(ex, McpError) else str(ex)
+                    raise ToolExecutionException(error_message, inner_exception=ex) from ex
                 if attempt == 0:
-                    # First attempt failed, try reconnecting.
-                    logger.info("MCP connection closed or terminated unexpectedly. Reconnecting...")
+                    logger.info("MCP connection lost during tasks/result; reconnecting (task_id=%s).", task_id)
```
moonbox3: Are we covering this reconnect path? `_fetch_task_result`'s reconnect-once-then-give-up is structurally identical to the poll loop's (which has reconnect + second-disconnect tests), but I couldn't find a test that drops the connection during `tasks/result`. A regression here (wrong attempt index, missing `continue`, swallowed error) would ship silently. Worth a test that disconnects on the first `tasks/result` and asserts one reconnect + retry, plus the second-failure "task state unknown" path?

python/packages/core/agent_framework/_mcp.py
```diff
@@ -149,6 +152,40 @@ def _url_origin(url: Any) -> tuple[str, str, int | None]:
...
+@dataclass
+class MCPTaskOptions:
+    """Options controlling how MCPTool drives the MCP long-running task lifecycle.
+
+    When an MCP server advertises a tool with ``execution.taskSupport == "required"``,
+    the framework transparently drives the SEP-2663 ``tools/call`` → ``tasks/get``
+    (polled) → ``tasks/result`` lifecycle so the agent sees a normal tool result.
+
+    Attributes:
+        default_ttl: Optional default time-to-live forwarded to the server as
```
TaoChenOSU: Does time to live here mean the server auto cancels the task?

python/packages/core/agent_framework/_mcp.py
```diff
@@ -149,6 +152,40 @@ def _url_origin(url: Any) -> tuple[str, str, int | None]:
...
+        cancel_remote_task_on_local_cancellation: If True (default), a local
+            cancellation of the awaiting coroutine triggers a best-effort
+            ``tasks/cancel`` on the server before re-raising ``CancelledError``.
+    """
+
+    default_ttl: timedelta | None = None
+    cancel_remote_task_on_local_cancellation: bool = True
+
+    def __post_init__(self) -> None:
+        if self.default_ttl is not None and self.default_ttl.total_seconds() < 0:
```
TaoChenOSU: Can it be 0? It felt like 0 wouldn't make sense.

python/packages/core/agent_framework/_mcp.py
```diff
@@ -1363,35 +1496,280 @@ async def call_tool(self, tool_name: str, **kwargs: Any) -> str | list[Content]:
...
+                parsed = parser(fallback_result)
+                text = (
+                    "\n".join(c.text for c in parsed if c.type == "text" and c.text)
+                    if isinstance(parsed, list)
+                    else str(parsed)
+                )
+                raise ToolExecutionException(text or str(parsed))
+            return parser(fallback_result)
+
+        assert task_id is not None  # noqa: S101  # nosec B101 - protected by the branch above
```
TaoChenOSU: nit: prefer checking it explicitly and raise a clear exception if it's not set.

python/packages/core/agent_framework/_mcp.py
peibekwe: Added `MCPTaskOptions.max_task_wait` that bounds the entire post-create lifecycle and fires a best-effort tasks/cancel on expiry.

python/packages/core/agent_framework/_mcp.py
peibekwe: Good catch! Transient `REQUEST_TIMEOUT` on tasks/get now keeps polling (bounded by `max_task_wait`), and every abandonment path fires tasks/cancel via a unified outer handler, so the remote task is never silently orphaned.

python/packages/core/agent_framework/_mcp.py
peibekwe: Good catch! Removed the unsafe fallback; an unparseable success now raises `ToolExecutionException` instead of re-issuing `tools/call`.

python/packages/core/agent_framework/_mcp.py
peibekwe: Made `MCPTaskOptions` a frozen dataclass and documented immutability in the docstring and `AGENTS.md`.

python/packages/core/agent_framework/_mcp.py
peibekwe: Rewrote `_try_cancel_task` to log at warning and dropped the redundant outer contextlib.suppress.

python/packages/core/agent_framework/_mcp.py
peibekwe: Added `test_fetch_task_result_reconnects_during_fetch` and `test_fetch_task_result_second_disconnect_raises_task_state_unknown_and_cancels` to cover the reconnect + second-failure paths.

python/packages/core/agent_framework/_mcp.py
peibekwe: Extracted a shared helper used by both and removed the duplicated block.

python/packages/core/agent_framework/_mcp.py
peibekwe: Clarified the docstring: default_ttl is the server's task-record retention period after terminal status (so the client can still read tasks/get/tasks/result); it doesn't cancel a running task.

python/packages/core/agent_framework/_mcp.py
peibekwe: Good point! Tightened validation to require default_ttl > 0 (a zero TTL would expire the record before any client could read it). Updated the test accordingly.

python/packages/core/agent_framework/_mcp.py
peibekwe: Replaced the assert.

TaoChenOSU (approved):

moonbox3 (approved):

PR #6342, opened 2026-06-04
Python: Improve AG-UI tool call argument protocol conformance

White-Mouse:
### Motivation & Context

AG-UI `TOOL_CALL_ARGS` events carry a tool-call identifier. The converter previously attached every argument fragment to the currently active call without checking that identifier. Out-of-order or malformed events could therefore associate arguments with a different tool call.

- **What are the major changes?** Normalize the supported tool-call ID field names, ignore argument fragments whose ID does not match the active call, and clear call state only for the matching `TOOL_CALL_END`.
- **What is the impact of these changes?** Valid event streams keep their existing behavior. Mismatched fragments are ignored instead of being rebound to another call.
- **What do you want reviewers to focus on?** The matching rules when an event omits its ID, which preserve compatibility with streams that do not repeat the ID on every event.

Validation completed from `python/`:

- `uv run pytest -q packages/ag-ui/tests/ag_ui/test_event_converters.py packages/ag-ui/tests/ag_ui/test_ag_ui_client.py` (45 passed)
- `uv run ruff check ...`
- `uv run ruff format --check ...`
- `uv run poe pyright -P ag-ui`
- `uv run poe test-typing -P ag-ui`

### Related Issue

Not linked yet.

### Contribution Checklist

### DIFF
--- a/python/packages/ag-ui/agent_framework_ag_ui/_event_converters.py
+++ b/python/packages/ag-ui/agent_framework_ag_ui/_event_converters.py
@@ -31,6 +31,16 @@ class AGUIEventConverter:
         self.thread_id: str | None = None
         self.run_id: str | None = None
 
+    @staticmethod
+    def _get_tool_call_id(event: dict[str, Any]) -> str | None:
+        """Return the tool call ID from either AG-UI field spelling."""
+        tool_call_id = event.get("toolCallId")
+        if tool_call_id is None:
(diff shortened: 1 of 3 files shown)


moonbox3:
Please link the PR to an open issue

moonbox3 (approved):

eavanvalkenburg (approved):

PR #6383, opened 2026-06-08
Python: parse structured response value from final message

### LINKED ISSUE #6366: Python: [Bug]: AgentResponse.value throws pydantic ValidationError parse errors in multi-message scenarios when using with_request_info()
### Description

`_types.py`
Classes: `ChatResponse()` and `AgentResponse()`
Methods: `text()` and `value()`

When utilizing `response_format` (such as a Pydantic BaseModel), the framework relies on lazy evaluation to parse the output when the `.value` property is accessed. However, `.value` feeds `self.text` directly to the JSON/Pydantic parser, and `self.text` blindly concatenates the text of all messages in the response history (`return "".join(msg.text for msg in self.messages)`)

In multi-turn agent loops (e.g., involving intermediate tool calls or reasoning traces), this concatenates conversational text with the final JSON payload, creating an invalid string like:

`{ "skill_name": "building-permit-compliance"}{"compliant": True}`

The `with_request_info` Nuance: This architectural flaw is often masked during standard automated runs because the orchestrator simply passes the raw `.text` to the next agent without ever touching `.value`. However, if a developer uses `.with_request_info()` (e.g., for Human-in-the-Loop workflows), the framework is forced to pause, serialize the event state, and access `.value`. This triggers the lazy evaluation, feeds the concatenated invalid string into the parser, and immediately crashes the workflow with a `ValidationError` or `JSONDecodeError`

Providing `response_format`, `skills_provider` (or any tool), and `with_request_info` will cause a PydanticValidation because the multi-turn conversation is concatenated before the call to `model_validate_json` in the `_parse_structured_response_value` function

### Code Sample

```python
def create_compliance_agent():
    chat_client = OpenAIChatCompletionClient(
        model="not_used",  # required but not used since we set the model in the agent options
        base_url=os.getenv("HF_API_BASE_URL"),
        api_key=os.getenv("HF_API_KEY"),
    )

    skills_provider = SkillsProvider.from_paths(
        skill_paths=pathlib.Path(__file__).parent / "skills"
    )

    return chat_client.as_agent(
        instructions=(
            "You are a building permit compliance agent that determines whether a building permit application is compliant based on the extracted data from the application and the relevant building codes."
            "You will be given the extracted data from a building permit application and you should determine whether the application is compliant with the information in the building permit compliance skill."
            "You MUST use the load_skill tool to read the 'building-permit-compliance' skill before making any decisions."
        ),
        name="building_permit_compliance_agent",
        tools=[],
        context_providers=[skills_provider],
        default_options={
            "model": MODEL,
            "response_format": ComplianceResult,
        },
    )

def create_workflow():
    data_agent = create_data_agent()
    compliance_agent = create_compliance_agent()
    return (
        SequentialBuilder(
            participants=[read_pdf, data_agent,
                          compliance_agent],
            chain_only_agent_responses=True,
        )
        .with_request_info(agents=[compliance_agent])
        .build()
    )
```

The input received by the pydantic model (notice how the call to the skills provider is sent too because of how `self.text` is accessed:

```markdown
{
  "skill_name": "building-permit-compliance"
}
{
  "compliant": false,
  "reasons": [
    "Contractor \"Pacific Coast Contractors LLC\" is not on the list of approved contractors.",
    "The application pertains to a commercial renovation rather than a solar panel installation, which falls outside the scope of the solar panel compliance criteria."
  ]
}



pydantic_core._pydantic_core.ValidationError: 1 validation error for ComplianceResult
  Invalid JSON: trailing characters at line 5 column 1 [type=json_invalid, input_value='{\n  "skill_name": "buil...ance criteria."\n  ]\n}', input_type=str]
    For further information visit https://errors.pydantic.dev/2.13/v/json_invalid
```

### Package Versions

agent-framework-core: 1.8.0

### Python Version

Python 3.13.5

### Additional Context

There are ways to manually work around this; however, it would be good to know if the behaviour is expected and by design, or if it is unexpected and will be fixed.

liuzemei:
### Motivation and Context

Fixes #6366.

When an agent response contains intermediate messages and a final structured JSON payload, `AgentResponse.value` currently parses `self.text`, which concatenates all message text. That can join intermediate tool/context JSON with the final response JSON and cause `ValidationError` / `JSONDecodeError` from trailing characters.

### Description

- Parse structured `ChatResponse.value` and `AgentResponse.value` from the last non-empty message text instead of the public concatenated `.text` value.
- Keep `.text` behavior unchanged for callers that want the full concatenated response text.
- Add regression coverage for Pydantic response formats and JSON schema mapping response formats.

### Testing

```bash
uv run --group dev pytest packages/core/tests/core/test_types.py -q -k 'final_message or empty_text or invalid_schema or chat_response_with_format or mapping_response_format'\nuv run pytest packages/core/tests/core/test_types.py -q\nuv run pytest packages/core/tests/core/test_middleware.py -q\nuv run pytest packages/core/tests/workflow/test_workflow.py -q -k 'AgentResponse or response or output'\nuv run ruff format --check packages/core/agent_framework/_types.py packages/core/tests/core/test_types.py\nuv run ruff check packages/core/agent_framework/_types.py packages/core/tests/core/test_types.py\ngit diff --check\n```\n\n### Contribution Checklist\n\n- [x] The code builds clean without any errors or warnings\n- [x] The PR follows the Contribution Guidelines\n- [x] All unit tests pass, and I have added new tests where possible\n- [ ] **Is this a breaking change?** If yes, add "[BREAKING]" prefix to the title of the PR.

### DIFF
--- a/python/packages/core/tests/core/test_types.py
+++ b/python/packages/core/tests/core/test_types.py
@@ -869,6 +869,79 @@ def test_chat_response_with_mapping_response_format() -> None:
     assert response.value["response"] == "Hello"
 
 
+def test_chat_response_value_parses_final_message_with_response_format() -> None:
+    """ChatResponse.value should ignore intermediate messages when parsing structured output."""
+    response = ChatResponse(
+        messages=[
+            Message(role="assistant", contents=['{"skill_name": "building-permit-compliance"}']),
+            Message(role="assistant", contents=['{"response": "Hello"}']),
+        ],
+        response_format=OutputModel,
+    )
+
+    assert response.text == '{"skill_name": "building-permit-compliance"}\n{"response": "Hello"}'
+    assert response.value is not None
+    assert response.value.response == "Hello"
+
+
+def test_agent_response_value_parses_final_message_with_response_format() -> None:
+    """AgentResponse.value should ignore intermediate messages when parsing structured output."""
+    response = AgentResponse(
+        messages=[
+            Message(role="assistant", contents=['{"skill_name": "building-permit-compliance"}']),
+            Message(role="assistant", contents=['{"response": "Hello"}']),
+        ],
+        response_format=OutputModel,
+    )
+
+    assert response.text == '{"skill_name": "building-permit-compliance"}{"response": "Hello"}'
+    assert response.value is not None
+    assert response.value.response == "Hello"
+
+
+def test_agent_response_mapping_value_parses_final_message() -> None:
+    """AgentResponse.value should parse the final message for JSON schema mappings."""
+    response = AgentResponse(
+        messages=[
+            Message(role="assistant", contents=['{"skill_name": "building-permit-compliance"}']),
+            Message(role="assistant", contents=['{"response": "Hello"}']),
+        ],
+        response_format={"type": "object", "properties": {"response": {"type": "string"}}},
+    )
+
+    assert response.value is not None
+    assert isinstance(response.value, dict)
+    assert response.value["response"] == "Hello"
+
+
+def test_chat_response_value_ignores_trailing_non_assistant_message() -> None:
+    """ChatResponse.value should parse the final assistant message when later tool output exists."""
+    response = ChatResponse(
+        messages=[
+            Message(role="assistant", contents=['{"response": "Hello"}']),
+            Message(role="tool", contents=["tool output is not structured JSON"]),
+        ],
+        response_format=OutputModel,
+    )
+
+    assert response.value is not None
+    assert response.value.response == "Hello"
+
+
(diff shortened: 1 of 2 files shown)


liuzemei:

python/packages/core/tests/core/test_types.py
eavanvalkenburg, line 1: we also need some tests for how this behaves with streaming, so constructing a number of streaming messages, then combining those and then calling the parser

eavanvalkenburg:
and some tests are failing, please have a look at those as well @liuzemei

moonbox3 (approved):

eavanvalkenburg (approved):

PR #6400, opened 2026-06-08
Python: fix: execute non-approval tools in mixed approval batches

### LINKED ISSUE #6385: Python: [Bug]: Mixed Tool Batch Applies Approval Wrapper To All Tool Calls
## Summary
When the model emits multiple tool calls in one assistant turn, and only one of those tools has approval_mode set to always_require, the framework wraps all function calls in function_approval_request.

In AG-UI this results in confirm_changes being emitted for every tool call in the batch, including tools that should execute without approval.

## Impact
- Non-sensitive tools are incorrectly blocked behind approval dialogs.
- UI shows duplicate or extra approval prompts.
- User flow becomes confusing and slower.

## Reproduction
1. Register two tools:
   - add_comment with approval_mode="always_require"
   - search_work_items with default approval_mode (never_require)
2. Send a prompt that causes the model to emit both function calls in the same assistant message.
3. Observe AG-UI events and stored messages.

## Expected
Only add_comment should be wrapped as function_approval_request and mapped to confirm_changes.
search_work_items should execute directly.

## Actual
Both add_comment and search_work_items are wrapped as function_approval_request, producing two confirm_changes approvals.

## Root Cause
In _try_execute_function_calls, approval_needed is computed as a single boolean for the whole batch. If any function call targets a tool in approval_tools, the code returns function_approval_request for all function_call items.

Current behavior location:
- agent_framework/_tools.py
- Function: _try_execute_function_calls
- Branch: if approval_needed then return Content.from_function_approval_request for every function_call in function_calls

## Proposed Fix
Handle function calls per item instead of all-or-nothing by batch:

1. Split function_calls into groups:
   - approval_calls: function_call where tool is in approval_tools
   - executable_calls: function_call where tool is known and does not require approval
   - declaration_only_calls: function_call where tool is declaration_only or additional_tool
2. Return approval request content only for approval_calls.
3. Execute executable_calls normally.
4. Preserve existing declaration_only behavior.
5. Keep terminate_on_unknown_calls behavior unchanged.

## Notes
A temporary mitigation is to disable multiple tool calls at model level:
- default_options={"allow_multiple_tool_calls": False}

This reduces incidence but does not fix the underlying batching logic.

## Applied workaround:
I disallow multiple tool calls for now.
``` default_options={"allow_multiple_tool_calls": False} ```

Disclaimer: To my understanding non-critical tool calls should not be wrapped. But I may not be in the complete picture of the planned/intended approval concept of the library.

### Code Sample

```markdown

```

```markdown
What gets emitted in MessagesSnapshot:

`
messages:<...>,
"tool_calls": [        {
            "id": "f991b6dc-fa33-4065-9e3c-f7373cc023b7",
            "name": null,
            "role": "assistant",
            "content": null,
                {
                    "id": "call_ZPd6rILF7c5QoA9yqJ3jb9cj",
                    "type": "function",
                    "function": {
                        "name": "add_comment",
                        "arguments": "{\"id\": 69239, \"comment\": \"test\"}"
                    },
                    "encrypted_value": null
                },
                {
                    "id": "call_AgE9A3WI1iu3FrIb7D3ZDYpR",
                    "type": "function",
                    "function": {
                        "name": "search_work_items",
                        "arguments": "{\"work_item_type\": \"Task\", \"assigned_to\": \"@me\", \"top\": 200}"
                    },
                    "encrypted_value": null
                },
                {
                    "id": "1c54f251-4dcf-4d51-b61e-d774fd084842",
                    "type": "function",
                    "function": {
                        "name": "confirm_changes",
                        "arguments": "{\"function_name\": \"add_comment\", \"function_call_id\": \"call_ZPd6rILF7c5QoA9yqJ3jb9cj\", \"function_arguments\": {\"id\": 69239, \"comment\": \"test\"}, \"steps\": [{\"description\": \"Execute add_comment\", \"status\": \"enabled\"}]}"
                    },
                    "encrypted_value": null
                },
                {
                    "id": "2cc96972-0c2b-469f-aa3d-df24efc3336b",
                    "type": "function",
                    "function": {
                        "name": "confirm_changes",
                        "arguments": "{\"function_name\": \"search_work_items\", \"function_call_id\": \"call_AgE9A3WI1iu3FrIb7D3ZDYpR\", \"function_arguments\": {\"work_item_type\": \"Task\", \"assigned_to\": \"@me\", \"top\": 200}, \"steps\": [{\"description\": \"Execute search_work_items\", \"status\": \"enabled\"}]}"
                    },
                    "encrypted_value": null
                }
            ],
            "encrypted_value": null
        }
}
```

### Package Versions

agent-framework-ag-ui

### Python Version

_No response_

### Additional Context

_No response_

jstar0:
## Summary

Fixes #6385.

When an assistant response contains a mixed batch of tool calls, the auto-invocation path currently uses one batch-level approval flag. If any tool in the batch requires approval, every function call is converted to a `function_approval_request`, including tools configured with `approval_mode="never_require"`. That blocks non-sensitive tools behind the approval flow and prevents their normal results from being returned.

## Changes

This change makes approval handling per function call:

- Calls to tools with `approval_mode="always_require"` are returned as approval requests.
- Calls that do not require approval continue through the normal execution path in the same batch.
- Mixed results preserve batch order, so executed results and approval requests line up with the original function calls.
- Executed tool results in a mixed approval batch are emitted as `tool` messages, while approval requests remain attached to the assistant message.
- Updated the `tool()` documentation note to match the per-call behavior.

## Verification

```bash
uv run pytest packages/core/tests/core/test_function_invocation_logic.py::test_mixed_batch_only_approval_tool_gets_wrapped
uv run pytest packages/core/tests/core/test_function_invocation_logic.py
uv run pytest packages/core/tests/workflow/test_agent_executor_tool_calls.py
uv run pytest packages/core/tests/core/test_tools.py packages/core/tests/core/test_middleware.py packages/core/tests/core/test_middleware_context_result.py packages/core/tests/core/test_middleware_with_agent.py packages/core/tests/core/test_middleware_with_chat.py
uv run ruff check packages/core/agent_framework/_tools.py packages/core/tests/core/test_function_invocation_logic.py
uv run pyright packages/core/agent_framework/_tools.py packages/core/tests/core/test_function_invocation_logic.py
```

## Contribution Checklist

- [x] Relevant unit tests pass, and I have added new tests where possible

### DIFF
--- a/python/packages/core/agent_framework/_tools.py
+++ b/python/packages/core/agent_framework/_tools.py
@@ -1238,8 +1238,8 @@ def tool(
     Note:
         When approval_mode is set to "always_require", the function will not be executed
         until explicit approval is given, this only applies to the auto-invocation flow.
-        It is also important to note that if the model returns multiple function calls, some that require approval
-        and others that do not, it will ask approval for all of them.
+        If the model returns multiple function calls, only the calls for tools that require approval
+        will ask for approval.
 
     Example:
 
@@ -1675,8 +1675,8 @@ async def _try_execute_function_calls(
     Returns:
         A tuple of:
         - A list of Content containing the results of each function call,
-          or the approval requests if any function requires approval,
-          or the original function calls if any are declaration only.
+          approval requests for function calls that require approval,
+          or original function calls for declaration-only requests.
         - Always False; termination via middleware is no longer supported.
     """
     from ._types import Content
@@ -1685,20 +1685,18 @@ async def _try_execute_function_calls(
     # The live tools list (when tools is the run-local list) is exposed on the
     # FunctionInvocationContext so tools can add/remove tools during the run.
     live_tools: list[ToolTypes] | None = cast("list[ToolTypes]", tools) if isinstance(tools, list) else None
-    approval_tools = [tool_name for tool_name, tool in tool_map.items() if tool.approval_mode == "always_require"]
+    approval_tools = {tool_name for tool_name, tool in tool_map.items() if tool.approval_mode == "always_require"}
     logger.debug(
         "_try_execute_function_calls: tool_map keys=%s, approval_tools=%s",
         list(tool_map.keys()),
-        approval_tools,
+        list(approval_tools),
     )
     declaration_only = [tool_name for tool_name, tool in tool_map.items() if tool.declaration_only]
     configured_additional_tools = config.get("additional_tools") or []
     additional_tool_names = [tool.name for tool in configured_additional_tools]
-    # check if any are calling functions that need approval
-    # if so, we return approval request for all
-    approval_needed = False
-    declaration_only_flag = False
-    for fcc in function_calls:
+    deferred_results_by_index: dict[int, Content] = {}
+    function_calls_to_execute: list[tuple[int, Content]] = []
+    for idx, fcc in enumerate(function_calls):
         fcc_name = getattr(fcc, "name", None)
         logger.debug(
             "Checking function call: type=%s, name=%s, in approval_tools=%s",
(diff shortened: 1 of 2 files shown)


jstar0:

python/packages/core/agent_framework/_tools.py
```diff
@@ -2108,12 +2107,17 @@ def _handle_function_call_results(
...
-        new_items = [fccr for fccr in function_call_results if fccr.type != "function_call"]
+        new_items = [
+            fccr for fccr in function_call_results if fccr.type not in {"function_call", "function_result"}
+        ]
         if new_items:
             if response.messages and response.messages[0].role == "assistant":
                 response.messages[0].contents.extend(new_items)
             else:
                 response.messages.append(Message(role="assistant", contents=new_items))
+        function_result_items = [fccr for fccr in function_call_results if fccr.type == "function_result"]
```
moonbox3, line 2118: Could we apply this same split to the streamed update path? This keeps non-streaming mixed approval responses role-correct, but `_process_function_requests` still returns the unsplit `function_call_results`, and the streaming loop emits them all with `update_role == "assistant"`. Then `stream.get_final_response()` contains a `function_result` on an assistant message, so persisting it for approval continuation can send provider-invalid history. Should the streaming path yield the tool results as a separate `role="tool"` update before the approval request?

eavanvalkenburg:
We have a different design for this in mind and will get that going this week, so closing this PR, thanks for the look @jstar0

PR #3908, opened 2026-02-13
Python: Replace wildcard imports with explicit imports

eavanvalkenburg:
Fixes #3605

## Summary
Replaced all wildcard imports (`from ... import *`) with explicit symbol imports across the Python codebase.

## Changes
- **Core package** (`agent_framework/__init__.py`): Replaced 10 wildcard imports with explicit imports for ~100 symbols
- **OpenAI subpackage** (`agent_framework/openai/__init__.py`): Replaced 6 wildcard imports with explicit imports
- **Lab namespace packages**: Replaced 3 wildcard imports with explicit imports and added `__all__` declarations for re-exports
- **CODING_STANDARD.md**: Added explicit guidance prohibiting wildcard imports with rationale

## Impact
- **API**: No breaking changes - all previously exported symbols remain available
- **Clarity**: Makes public API explicit and improves code navigation
- **Performance**: Eliminates unnecessary symbol resolution during imports
- **Tooling**: Better IDE autocomplete, go-to-definition, and type checking

## Testing
- Ran core test suite: 812 passed, 18 skipped
- Verified ruff checks pass on all modified files
- Confirmed wildcard import count: 19 → 0

## Wildcard Count
- **Before**: 19 wildcard imports
- **After**: 0 wildcard imports

PR #6407, opened 2026-06-08
[BREAKING] Python: Add `reset` to workflow

### LINKED ISSUE #4280: [Feature]: Allow checkpoint creation at boundary SuperStepStartedEvent
### Description

I'm building a complex workflow framework that supports the concept of "backtracking" within a depth-first-search scenario. I'm trying to set up a system where, within a workflow, each agent/executor can provide multiple output options to the next executor. Consider the following basic workflow:

start --> executorA --> executorB --> finalOutput

Suppose executorA determines there are 2 possible outputs, O1 and O2, it could give to executorB. In my framework, during the first pass, executorA will generate O1 and give that to executorB. Now, suppose executorB immediately crashes when reading O1. This is the perfect opportunity to have the workflow automatically backtrack to executorA, where he will quickly decide to return O2 instead. Then executorB can try running O2 and hopefully generate the desired output.

My issue with the way checkpoints are saved is that they are tied to the SuperStepCompleted event. In my example, the only natural places agent-framework supports checkpointing is between executorA and executorB, and after executorB runs. This means that executorA can never be "backtracked" to, and his other branch path O2 can never be revisited.

There are several ways I can write code to make my framework support the functionality I want, but it feels very inelegant. Essentially, I would like an additional entry point where agent-framework can invoke my executorA's implementation of `OnCheckpointRestoredAsync`.

### Code Sample

```markdown

```

### Language/SDK

Both

TaoChenOSU:
### Motivation and Context

When a workflow is hosted as an agent in Foundry Hosted Agent, a new message comes in without a conversation ID or a previous response ID should trigger a new workflow run, without any residual states in the workflow from previous interactions.

This is currently not possible unless a new workflow instance is built from the builder, which is impossible in hosted agent since the workflow instance is only created once when a session starts. This PR addresses the limitation by introducing 
`reset` on the workflow class, which allows clients to reset a workflow to its original state. A workflow now persists an initial checkpoint created automatically prior to the first run and will use that checkpoint for reset.

Closes #4280, #4588

### Description

1. Refactor _workflow.py, _runner.py and _runner_context.py such that a workflow contains only the minimal amount of state of a run. Responsibilities are also made to be more clearly distributed across workflow, runner and runner_context.
2. Add `reset` to workflow.
3. Use `reset` in `_responses.py` restore the workflow to its initial state when a request doesn't have a previous response id nor conversation id.
4. Add tests.

### Contribution Checklist

This is considered a breaking change because the runner module is removed from the top-level export. This is a low risk breaking change as we don't expect too many customers are using the runner directly. The runner was designed to be internal so this encourages the correct usage.


python/packages/core/agent_framework/_workflows/_workflow.py
```diff
@@ -760,9 +757,6 @@ def run(
                 client_kwargs=client_kwargs,
             ),
             finalizer=functools.partial(self._finalize_events, include_status_events=include_status_events),
```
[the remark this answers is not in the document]
TaoChenOSU: Removed reserve() on runner.

python/packages/core/agent_framework/_workflows/_workflow.py
```diff
@@ -478,6 +474,50 @@ def get_executors_list(self) -> list[Executor]:
...
+              Use this method only when necessary, for example to capture the initial state of the workflow prior to the
+              first run.
+            - Creating a checkpoint manually will alter the checkpoint lineage. The new checkpoint will become the
+              parent of the next checkpoint created automatically (if checkpointing is enabled by providing a storage).
+        """
+        if self._is_run_active():
+            raise WorkflowException(
+                "Cannot create checkpoint while a workflow run is active. "
+                "Checkpointing is only allowed between runs when the workflow is idle."
+            )
```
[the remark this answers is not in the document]
TaoChenOSU: A workflow may have in flight message.

python/packages/core/agent_framework/__init__.py
```diff
@@ -481,7 +482,6 @@
     "RoleLiteral",
     "RubricScore",
     "RunContext",
-    "Runner",
```
moonbox3, line 501: Should we treat dropping `Runner` from the top-level exports as a breaking change? It has been in `__all__` since #3908 and shipped in the released 1.x line, so `from agent_framework import Runner` now raises `ImportError` for anyone who imported it, yet the PR is not marked `[BREAKING]`. Could we either keep a lazy `__getattr__` re-export with a `DeprecationWarning`? We shouldn't be breaking from one version to the next without some heads up to developers.

python/packages/core/agent_framework/_workflows/_workflow.py
```diff
@@ -789,51 +838,57 @@ async def _run_core(
...
+                yield event
+        finally:
+            # Clear the active-run weakref so a subsequent ``run()`` is allowed.
+            # ``run()`` set this synchronously after constructing the ResponseStream;
+            # we clear it here once the run has finished (success, error, early
+            # close, or partial iteration). This is in-band, so by the time the
+            # caller's stream is later garbage collected, ``_active_run`` is already
+            # ``None`` (or has been replaced by a newer run's weakref) - no GC-time
+            # finalizer is needed.
+            self._active_run = None
```
moonbox3: Could a late finalizer here reset a newer run's guard? `_run_core`'s `finally` clears `self._active_run` unconditionally. If a caller abandons a partially consumed stream and then starts another run, the abandoned async generator is closed by asyncio's async-gen finalizer hook at GC time, which runs this `finally` and sets `_active_run = None` while the new run is still live. 

I confirmed the ordering locally: the dropped generator's `finally` fires on a later loop tick and overwrites the value the new run just set. That silently reopens the concurrency guard, so a subsequent `run()` (or a `create_checkpoint()`, which also gates only on `_is_run_active()`) could proceed against an already-active run and mutate shared runner/context/state concurrently. This is exactly the "replaced by a newer run's weakref" case the inline comment assumes is safe. 

Should we compare-and-clear instead, only resetting when `self._active_run()` is still this run's own stream, so a stale finalizer can't clobber a newer run?

python/packages/core/agent_framework/__init__.py
TaoChenOSU: Added the breaking prefix in the title and a short text in the description.

python/packages/core/agent_framework/_workflows/_workflow.py
TaoChenOSU: How can the new run be started if the workflow still holds the reference to the old run?

moonbox3: The workflow only holds a weakref to the ResponseStream, not a strong reference. So if the caller partially consumes the stream and then drops their reference, `_active_run()` can become None, which lets the next `run()` pass `_is_run_active()` and install a weakref to the new stream.
 
The stale part is the old async generator’s cleanup: its finally can still run later via async-gen finalization. When it does, it unconditionally executes `self._active_run = None`, which can clear the weakref that now points at the newer run. So the issue is not “how can B start while A is strongly held?” A is not strongly held by the workflow. The issue is: A's delayed cleanup can mutate the workflow after B has already become the active run.
 
That’s why compare-and-clear is safer: only clear `_active_run` if it still points to this run's own stream.

eavanvalkenburg: yeah, looking at this, indeed a weakref here seems to defeat the purpose of holding a reference at all, and I'm also a bit confused as to the goals here, either we want to support concurrent executions or not, let's first clarify that (inside and outside of the context of FHA) and then figure out the best way to do that.

TaoChenOSU: The purpose is that we want to avoid concurrent execution but still want to allow customers to run a workflow (same instance multiple times).

Previously, we used a `_is_running` flag to indicate if a workflow is running. We had this flag in both the workflow and the runner, which is very confusing and error prone. This is part of the refactoring to make the workflow module and the runner module more focus on their own unique responsibilities:
1. Workflow: manages the topology and running state
2. Runner: manages the run state and the pregel loop
3. Context: manages the runtime, including message passing, events, and checkpointing.

python/packages/core/agent_framework/__init__.py
moonbox3: The point I am getting at: we can't break from one version to the next. The proper way to do this is add a deprecation warning for this import and then in a later iteration, we remove it.

python/packages/core/agent_framework/_workflows/_workflow.py
```diff
@@ -478,6 +474,50 @@ def get_executors_list(self) -> list[Executor]:
         """Get the list of executors in the workflow."""
         return list(self.executors.values())
 
+    async def create_checkpoint(self, checkpoint_storage: CheckpointStorage | None) -> CheckpointID:
```
eavanvalkenburg: Should we make this a public method?  Workflow.create_checkpoint() is added as a general API but the only clear use is Foundry hosting’s “restore to initial state” trick. That is not a core workflow requirement. It is a host adapter requirement. Can we instead keep this inside foundry_hosting, or introduce a hosting-specific workflow factory/reset strategy?

python/packages/foundry_hosting/agent_framework_foundry_hosting/_responses.py
```diff
@@ -590,6 +595,14 @@ async def _handle_inner_workflow(
             # any future async resources owned by the workflow are entered here.
             await self._ensure_agent_ready()
 
+            # Create a checkpoint to store the initial state of the workflow, if it doesn't already exist.
+            # This allows us to restore to a clean slate when no conversation_id or previous_response_id
+            # is supplied in a request.
+            if self._initial_checkpoint_id is None:
+                self._initial_checkpoint_id = await self._agent.workflow.create_checkpoint(
```
eavanvalkenburg: so we are doing double work here, we are creating a inital checkpoint, and then later on, seeing if we can restore a different one? That is a lot of machinery to emulate “construct a clean workflow.” The simpler version is: for no context id, use a fresh workflow instance. can we not solve that in the hosted adapter, not in core?

python/packages/core/agent_framework/__init__.py
```diff
@@ -489,7 +490,6 @@
...
```
eavanvalkenburg, line 501: we should not remove this here, it is a public API and we cannot just remove it without warning, at the very least we need to start by marking it deprecated here.

python/packages/foundry_hosting/agent_framework_foundry_hosting/_responses.py
eavanvalkenburg: _initial_checkpoint_id is checked and assigned without a lock. Two first requests can both see None, both call create_checkpoint(), and both mutate the same workflow runner/context. Then both proceed to restore/run on the same workflow instance. So this design depends on shared mutable workflow state and then pretends it is request-isolated, that does not solve the actual issue.

eavanvalkenburg (changes requested):
So left some comments, but after looking more closely it doesn't actually solve the issue with FHA, and adds breaking changes where none are needed.
 
Might I suggest a simpler version?

Keep normal workflows normal: one workflow instance has state; independent runs should use independent workflow instances. For Foundry hosting, add a hosted-only mechanism:

   - accept a workflow/agent factory for stateless requests, or
   - keep a private hosted reset implementation inside foundry_hosting, or
   - require explicit checkpoint-enabled hosted workflows and document that hosting owns persistence.

Do not add a public checkpoint API and a breaking top-level import change to make one host’s lifecycle model fit everyone else.

python/packages/core/agent_framework/_workflows/_workflow.py
TaoChenOSU: Good call!

Replaced this method with reset, which is more targeted.

We have considered a hosting-specific workflow factory, but that didn't land because we cannot rely on just the workflow factory, all executors/agents must also be created by factories too, which adds a lot more complexity on the customer side.

python/packages/foundry_hosting/agent_framework_foundry_hosting/_responses.py
TaoChenOSU: Removed. `reset` now handles everything.

TaoChenOSU: Removed

python/packages/core/agent_framework/__init__.py
moonbox3: This was my point in https://github.com/microsoft/agent-framework/pull/6407#discussion_r3417037962

python/packages/core/agent_framework/_workflows/_runner.py
```diff
@@ -63,99 +63,123 @@ def __init__(
...
+
+        while self._iteration < self._max_iterations:
+            logger.info(f"Starting superstep {self._iteration + 1}")
+            yield WorkflowEvent.superstep_started(iteration=self._iteration + 1)
+
+            # Run iteration concurrently with live event streaming: we poll
+            # for new events while the iteration coroutine progresses.
+            iteration_task = asyncio.create_task(self._run_iteration())
+            try:
+                while not iteration_task.done():
```
moonbox3, line 149: I've asked a similar question on a previous iteration / related change before: could abandoning a streaming run leave `_run_iteration()` running after the workflow has been marked idle? If the async generator is closed after yielding an event, Python injects `GeneratorExit`, not `asyncio.CancelledError`, so the cancellation block below does not run and the child task can keep mutating the shared runner/context. With the weakref run guard, dropping the `ResponseStream` can then permit `reset()` or another `run()` while that orphan task is still delivering messages, so should this cleanup cancel `iteration_task` in a `finally` whenever it is not done?

Ashutosh0x:
## Review: Workflow Reset

This is a critical feature for Foundry Hosted Agent scenarios. When a new message arrives without `conversation_id` or `previous_response_id`, the workflow must reset cleanly — great catch.

### Questions / Edge Cases to Consider

1. **Partial reset**: What happens if `reset()` is called mid-execution while a step is actively running? Is there a guard or does it raise?
2. **Checkpoint interaction**: If a workflow has checkpoint storage configured, does `reset()` also clear the persisted checkpoint, or only the in-memory state? If only in-memory, a subsequent `resume_from_checkpoint()` could restore stale state.
3. **Event ordering**: Does the host need to emit a `workflow_reset` event (or similar) to downstream listeners before re-running?
4. **Nested workflows**: If Workflow A orchestrates Workflow B, does calling `reset()` on A cascade to B?

### Suggested Test Cases

- `test_reset_clears_all_step_states`
- `test_reset_during_active_execution_raises`
- `test_reset_then_run_produces_clean_output`
- `test_reset_does_not_affect_persisted_checkpoint` (or does — depends on design intent)

Happy to contribute edge case tests if you'd like!

PR #6414, opened 2026-06-09
Python: Add tool approval middleware

### LINKED ISSUE #6385: Python: [Bug]: Mixed Tool Batch Applies Approval Wrapper To All Tool Calls

eavanvalkenburg:
### Motivation and Context

Python currently surfaces approval prompts for every tool call in a mixed tool-call batch when any one tool requires approval. This can block safe read-only tools behind unnecessary approval prompts and makes host approval UX noisy.

Fixes #6385

### Description

- Adds a session-backed mixed-batch bypass in the Python function-invocation loop so known non-approval-required tool calls are hidden, stored, and reinjected as approved responses when the visible approval flow resumes.
- Adds experimental `ToolApprovalMiddleware` with standing approval rules, tool+arguments rules, heuristic `auto_approval_rules`, queued approval prompts, and streaming/non-streaming support.
- Adds public helper APIs for "always approve this tool" and "always approve this exact tool+arguments" responses.
- Adds focused core tests and a regular-client sample showing host/user policy decisions for approval, rejection, standing approvals, and heuristic auto-approval.

### Contribution Checklist

### DIFF
--- a/python/packages/core/agent_framework/_harness/_tool_approval.py
+++ b/python/packages/core/agent_framework/_harness/_tool_approval.py
@@ -6,7 +6,7 @@ import copy
 import inspect
 import json
 from asyncio import sleep
-from collections.abc import AsyncIterable, Awaitable, Callable, Mapping, MutableMapping, Sequence
+from collections.abc import AsyncIterable, Awaitable, Callable, Iterable, Mapping, MutableMapping, Sequence
 from typing import Any, Literal, cast
 
 from .._feature_stage import ExperimentalFeature, experimental
@@ -65,7 +65,8 @@ def _content_from_state(value: Any) -> Content:
 def _contents_from_state(values: Any) -> list[Content]:
     if not isinstance(values, list):
         return []
-    return [_content_from_state(value) for value in cast(list[Any], values)]
+    state_items = list(cast(Iterable[Any], values))
+    return [_content_from_state(value) for value in state_items]
 
 
 def _content_to_state(content: Content) -> dict[str, Any]:
@@ -296,8 +297,10 @@ def _get_always_approve_scope(response: Content) -> ToolApprovalScope | None:
         return None
     metadata_mapping = cast(Mapping[str, Any], metadata)
     scope = metadata_mapping.get(ALWAYS_APPROVE_SCOPE_PROPERTY)
-    if scope in (ALWAYS_APPROVE_TOOL, ALWAYS_APPROVE_TOOL_WITH_ARGUMENTS):
-        return scope
+    if scope == ALWAYS_APPROVE_TOOL:
+        return ALWAYS_APPROVE_TOOL
+    if scope == ALWAYS_APPROVE_TOOL_WITH_ARGUMENTS:
+        return ALWAYS_APPROVE_TOOL_WITH_ARGUMENTS
     return None
 
 
@@ -418,6 +421,12 @@ class ToolApprovalMiddleware(AgentMiddleware):
                         content for content in update.contents if content.type != "function_approval_request"
                     ]
                     if remaining_contents:
+                        raw_finish_reason = update.finish_reason
+                        finish_reason: FinishReasonLiteral | FinishReason | None
+                        if isinstance(raw_finish_reason, str):
+                            finish_reason = FinishReason(raw_finish_reason)
+                        else:
+                            finish_reason = cast(FinishReasonLiteral | FinishReason | None, raw_finish_reason)
                         yield AgentResponseUpdate(
                             contents=remaining_contents,
                             role=update.role,
@@ -426,7 +435,7 @@ class ToolApprovalMiddleware(AgentMiddleware):
                             response_id=update.response_id,
                             message_id=update.message_id,
                             created_at=update.created_at,
-                            finish_reason=cast(FinishReasonLiteral | FinishReason | None, update.finish_reason),
+                            finish_reason=finish_reason,
                             continuation_token=update.continuation_token,
                             additional_properties=update.additional_properties,
                             raw_representation=update.raw_representation,
--- a/python/packages/core/agent_framework/_harness/_tool_approval.py
+++ b/python/packages/core/agent_framework/_harness/_tool_approval.py
@@ -1,0 +1,583 @@
+# Copyright (c) Microsoft. All rights reserved.
+
+from __future__ import annotations
+
+import copy
+import inspect
+import json
+from asyncio import sleep
+from collections.abc import AsyncIterable, Awaitable, Callable, Mapping, MutableMapping, Sequence
+from typing import Any, Literal, cast
+
+from .._feature_stage import ExperimentalFeature, experimental
+from .._middleware import AgentContext, AgentMiddleware
+from .._serialization import SerializationMixin
+from .._sessions import AgentSession
+from .._tools import _TOOL_APPROVAL_STATE_KEY  # pyright: ignore[reportPrivateUsage]
+from .._types import (
+    AgentResponse,
+    AgentResponseUpdate,
+    Content,
+    FinishReason,
+    FinishReasonLiteral,
+    Message,
+    ResponseStream,
+)
+
+DEFAULT_TOOL_APPROVAL_SOURCE_ID = _TOOL_APPROVAL_STATE_KEY
+ALWAYS_APPROVE_PROPERTY = "tool_approval"
+ALWAYS_APPROVE_SCOPE_PROPERTY = "always_approve"
+ALWAYS_APPROVE_TOOL: Literal["tool"] = "tool"
+ALWAYS_APPROVE_TOOL_WITH_ARGUMENTS: Literal["tool_with_arguments"] = "tool_with_arguments"
+
+_RULES_KEY = "rules"
+_QUEUED_APPROVAL_REQUESTS_KEY = "queued_approval_requests"
+_COLLECTED_APPROVAL_RESPONSES_KEY = "collected_approval_responses"
+
+ToolApprovalScope = Literal["tool", "tool_with_arguments"]
+ToolApprovalRuleCallback = Callable[[Content], bool | Awaitable[bool]]
+
+
+def _parse_function_arguments(function_call: Content) -> dict[str, Any]:
+    arguments = function_call.parse_arguments()
+    return dict(arguments or {})
+
+
+def _serialize_argument_value(value: Any) -> str:
+    return json.dumps(value, sort_keys=True, separators=(",", ":"), default=str)
+
+
+def _serialize_arguments(function_call: Content) -> dict[str, str] | None:
+    arguments = _parse_function_arguments(function_call)
+    if not arguments:
+        return None
+    return {key: _serialize_argument_value(value) for key, value in arguments.items()}
+
+
+def _content_from_state(value: Any) -> Content:
+    if isinstance(value, Content):
+        return value
+    if isinstance(value, Mapping):
+        return Content.from_dict(cast(Mapping[str, Any], value))
+    raise TypeError(f"Expected Content or mapping state item, got {type(value).__name__}.")
+
+
+def _contents_from_state(values: Any) -> list[Content]:
+    if not isinstance(values, list):
+        return []
+    return [_content_from_state(value) for value in cast(list[Any], values)]
+
+
+def _content_to_state(content: Content) -> dict[str, Any]:
+    return content.to_dict()
+
+
+@experimental(feature_id=ExperimentalFeature.HARNESS)
+class ToolApprovalRule(SerializationMixin):
+    """A standing rule for approving future matching tool calls."""
+
+    tool_name: str
+    arguments: dict[str, str] | None
+
+    def __init__(self, tool_name: str, arguments: Mapping[str, str] | None = None) -> None:
+        """Initialize a tool approval rule.
+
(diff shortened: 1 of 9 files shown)


python/packages/core/agent_framework/_tools.py
```diff
@@ -1912,6 +1934,69 @@ def _is_hosted_tool_approval(content: Any) -> bool:
...
+    """Return the shared tool-approval state bag for the invocation session."""
+    if invocation_session is None:
+        return None
+    raw_state = invocation_session.state.get(_TOOL_APPROVAL_STATE_KEY)
+    if isinstance(raw_state, dict):
+        return cast(dict[str, Any], raw_state)
+    if raw_state is not None:
+        raise TypeError(
+            f"Session state for {_TOOL_APPROVAL_STATE_KEY!r} must be a dict, got {type(raw_state).__name__}."
+        )
```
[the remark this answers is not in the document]
eavanvalkenburg: Fixed in bc0cd9d03. `_get_tool_approval_state()` now accepts a restored `ToolApprovalState`, serializes it back into the shared session-state mapping, and continues storing/popping hidden mixed-batch approval requests without raising. I also added `test_mixed_batch_accepts_restored_tool_approval_state` to cover this path.

python/packages/core/agent_framework/_tools.py
```diff
@@ -2157,8 +2242,14 @@ async def _process_function_requests(
...
     max_errors: int,
     execute_function_calls: Callable[..., Awaitable[tuple[list[Content], bool, bool]]],
+    invocation_session: AgentSession | None = None,
 ) -> FunctionRequestResult:
+    from ._types import Message
+
     if prepped_messages is not None:
+        pending_auto_approval_responses = _pop_auto_approvable_approval_responses(invocation_session)
+        if pending_auto_approval_responses:
+            prepped_messages.append(Message(role="user", contents=pending_auto_approval_responses))
```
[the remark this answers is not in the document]
eavanvalkenburg: Fixed in 390340eb9. Hidden auto-approvable requests are now popped/replayed only when the current prepared messages include an explicit `function_approval_response`, so unrelated later turns leave the hidden requests queued. I also added `test_hidden_mixed_batch_requests_do_not_replay_on_unrelated_turn` to cover the stale replay case.

python/packages/core/agent_framework/_tools.py
```diff
@@ -1719,15 +1721,35 @@ async def _try_execute_function_calls(
...
+                or tool is None
+                or tool_name in declaration_only
+                or tool_name in additional_tool_names
+            ):
+                visible_requests.append(approval_request)
+                continue
+            if invocation_session is None:
+                visible_requests.append(approval_request)
+                continue
+            auto_approvable_requests.append(approval_request)
```
TaoChenOSU: Questions: 
1. what qualifies as an auto approvable request? 
2. Does auto approval mean the next time the tool is invoked, it no longer needs approval?

python/packages/core/agent_framework/_harness/_tool_approval.py
```diff
@@ -0,0 +1,592 @@
...
+def _parse_function_arguments(function_call: Content) -> dict[str, Any]:
+    arguments = function_call.parse_arguments()
+    return dict(arguments or {})
+
+
+def _serialize_argument_value(value: Any) -> str:
+    return json.dumps(value, sort_keys=True, separators=(",", ":"), default=str)
+
+
+def _serialize_arguments(function_call: Content) -> dict[str, str] | None:
```
moonbox3: Should this preserve empty arguments as `{}` instead of `None`? `create_always_approve_tool_with_arguments_response()` promises an exact tool+arguments rule, but approving a `{}` call stores `arguments=None`, and `_matches_rule()` treats that as a tool-wide wildcard. That means one no-argument approval can silently auto-approve later calls to the same tool with arbitrary arguments.

python/packages/core/agent_framework/_tools.py
```diff
@@ -2157,8 +2250,21 @@ async def _process_function_requests(
...
+    from ._types import Message
+
     if prepped_messages is not None:
+        has_explicit_approval_response = any(
+            content.type == "function_approval_response"
+            for message in prepped_messages
+            if isinstance(message, Message)
+            for content in message.contents
+        )
+        if has_explicit_approval_response:
```
moonbox3: What happens when a user abandons one mixed approval flow, starts another, and approves the newer request? The hidden requests are appended to one session-wide list and `_pop_auto_approvable_approval_responses()` drains all of them on any explicit approval response, so stale hidden tool calls from the abandoned flow can execute in the later unrelated flow. Could the hidden requests be keyed to the visible approval request or batch so only the matching hidden calls replay?

python/packages/core/agent_framework/_harness/_tool_approval.py
```diff
@@ -0,0 +1,592 @@
...
+
+
+def _matches_rule(request: Content, rules: Sequence[ToolApprovalRule]) -> bool:
+    function_call = _function_call_from_request(request)
+    if function_call is None:
+        return False
+    for rule in rules:
+        if rule.tool_name != function_call.name:
+            continue
+        if rule.arguments is None:
```
moonbox3, line 319: From a security perspective, should standing approvals include the hosted-tool trust boundary? The core function-invocation path treats `server_label` on hosted approval requests as the boundary that must pass through untouched, but this matcher only checks tool name and arguments. A standing approval for `delete_file` on one MCP server could auto-approve an approval request with the same tool name from another server, unless the rule also matches `server_label` or hosted approvals are excluded from this middleware path.

python/packages/core/agent_framework/_harness/_tool_approval.py
```diff
@@ -0,0 +1,592 @@
...
+        await self._drain_auto_approvable_queue(state)
+        if next_queued := self._pop_next_queued_request(state):
+            _save_state(context.session, state, source_id=self.source_id)
+            context.result = self._response_for_queued_request(next_queued, stream=context.stream)
+            return
+        if context.stream:
+            context.result = self._process_stream(context, call_next, state)
+            return
+
+        while True:
```
moonbox3, line 388: Could this preserve the configured function-invocation budgets across auto-approved turns? Each all-auto-approved pass re-enters `call_next()` from this unbounded middleware loop, while the inner function-invocation layer reinitializes `total_function_calls` and its `max_iterations` loop for each call. A model that repeatedly requests an auto-approved side-effecting tool can keep getting fresh budgets instead of stopping at `max_function_calls` or `max_iterations`; the streaming loop has the same shape.

python/packages/core/agent_framework/_tools.py
eavanvalkenburg: Clarified in 5d0e44b05 by renaming the mixed-batch hidden state to “already approved” rather than “auto approvable”. In the core loop this means a known `never_require` tool call from the same mixed batch is already approved for that batch only; it is hidden and replayed when the matching visible approval response resumes. It does not create a future standing approval. Future standing approvals are still only created by the explicit ToolApprovalMiddleware helpers/rules.

python/packages/core/agent_framework/_harness/_tool_approval.py
eavanvalkenburg: Fixed in 5d0e44b05. `ToolApprovalRule.arguments is None` is now documented and reserved for tool-wide rules, while argument-scoped no-argument approvals store `{}` and only match future no-argument calls. Added `test_tool_approval_middleware_empty_arguments_rule_is_not_tool_wide` to verify a `{}` approval does not auto-approve a later call with arguments.

python/packages/core/agent_framework/_tools.py
eavanvalkenburg: Fixed in 5d0e44b05. Hidden already-approved requests are now stored in groups keyed to the visible approval request ids from their batch, and replay pops only the group matching the current explicit approval response ids. Added `test_hidden_mixed_batch_requests_replay_only_for_matching_visible_approval` for the abandoned-batch/newer-batch case.

python/packages/core/agent_framework/_harness/_tool_approval.py
eavanvalkenburg: Fixed in 5d0e44b05. Standing approval rules now include the hosted-tool `server_label` boundary when present, and matching requires the same server label. This keeps hosted approvals working while preventing same-named tools on different hosted servers from sharing approvals. Added `test_tool_approval_middleware_standing_rules_include_hosted_server_boundary`.

python/packages/core/agent_framework/_harness/_tool_approval.py
eavanvalkenburg: Fixed in 5d0e44b05. ToolApprovalMiddleware now passes a shared private function-invocation budget state through repeated inner calls, and the function-invocation layer checks shared `total_function_calls` before making the next model call. Added `test_tool_approval_middleware_auto_approved_loops_share_function_call_budget` to verify repeated auto-approved re-entry respects `max_function_calls`.

moonbox3 (approved):

westey-m (approved):

PR #6418, opened 2026-06-09
Python: host MAF workflows on a standalone Durable Task worker

### LINKED ISSUE #4877: Python: [Feature]: Durable Task Scheduler Support in Agent Framework Workflows
### Description

We have been working with the sample:

https://github.com/Azure-Samples/durable-task-extension-for-agent-framework/tree/main/samples/python/azure-container-apps/agentic-travel-planner

Using this setup, we validated the following use cases:

✅ HITL (Human-in-the-Loop) across replicas
One replica can pause for approval, and another replica can resume and complete the workflow.
✅ Distributed orchestration using Durable Task Scheduler
Demonstrates reliable execution across multiple orchestration instances.

**Gap Identified**

While Durable Task Scheduler works well for orchestration scenarios, it is not natively integrated with Microsoft Agent Framework workflows.

Is Durable Task support planned for Agent Framework workflows?

If already supported, could you please share the relevant GitHub repo or documentation URL?

### Code Sample

```markdown

```

### Language/SDK

Python

ahmedmuhsin:
# Summary

Enables a Microsoft Agent Framework `Workflow` to run as a **durable orchestration on a standalone Durable Task worker** (Azure Container Apps, AKS, a VM, or locally), not just inside Azure Functions. The workflow execution engine is extracted into `agent-framework-durabletask` behind a host-agnostic context protocol, and the Azure Functions host delegates to that shared engine instead of maintaining its own copy. The standalone host ships with a complete client surface: workflow lifecycle, human-in-the-loop, async start, and typed event streaming.

# Motivation

Previously, MAF workflow execution lived only in the Azure Functions host (`agent-framework-azurefunctions`). Running the same workflow on other compute required reimplementing the orchestrator. Both hosts ultimately drive a Durable Task backend, so the execution logic belongs in the shared `durabletask` package, with each host providing a thin adapter.

# What's included

- **Host-agnostic workflow engine (`durabletask`)**: a `WorkflowOrchestrationContext` protocol + `DurableTaskWorkflowContext` adapter, the superstep orchestrator, serialization helpers, a capturing runner context, and the shared non-agent activity body. Adds `DurableAIAgentWorker.configure_workflow` (auto-registers agent executors as entities, non-agent executors as activities, and the orchestrator), `plan_workflow_registration` (the shared "what to register" decision), and a persistent-event-loop async bridge (`run_agent_coroutine`) that fixes a cross-loop hang. The engine modules live in a private `_workflows/` subpackage.
- **A complete standalone client (`DurableWorkflowClient`)**: workflow lifecycle and HITL, plus an async surface (`run_workflow` for start-with-optional-wait, `stream_workflow` for a typed `WorkflowEvent` stream) so external callers can drive and observe a workflow without Azure Functions.
- **Azure Functions delegates to the shared engine**: `AgentFunctionApp` reuses the shared orchestrator, activity body, and registration planner instead of its own copies (a large net reduction); runs agent entity coroutines on the shared loop.
- **Core lazy re-exports**: `WORKFLOW_ORCHESTRATOR_NAME` and `DurableWorkflowClient` are exposed from the `agent_framework.azure` namespace.
- **Executor identity + typed input**: `AgentExecutor(agent, id=...)` registers the durable entity under the executor id the orchestrator dispatches to (so a custom id works); the client's initial JSON payload is reconstructed to the start executor's declared input type, mirroring in-process delivery.
- **Samples + tests**: standalone samples under `samples/04-hosting/durabletask` (`08_workflow`, `09_workflow_hitl`, and `10_workflow_streaming`) plus integration tests for the standalone host.

# Notable design points

- **Client interface segregation:** workflow lifecycle + HITL live on a dedicated `DurableWorkflowClient`; `DurableAIAgentClient` stays agent-only. The worker stays combined (`DurableAIAgentWorker`) because registration aggregates over one underlying worker, so "hosts aggregate, callers specialize."
- **Async client and event streaming:** the standalone client mirrors the async surface that in-process core workflows and the .NET durable host already expose. `run_workflow` starts a workflow and optionally awaits its output; `stream_workflow` yields typed `WorkflowEvent` objects as the run progresses (the same event type core streams via `Workflow.run(..., stream=True)`). Streaming is brokerless: the orchestrator publishes accumulated events to the orchestration custom status after each superstep (live execution only, never on replay), and the client consumes them by polling. Non-agent executors run inside an activity that captures their real events with data payloads; agent executors surface coarse invoked/completed lifecycle events. Granularity is per executor and per yielded output, not token-level; token-level streaming would require an external broker and is left as a follow-up. Because the underlying Durable Task client is synchronous, the async methods dispatch blocking calls to a worker thread so they never block the event loop.
- **Cross-loop hang fix:** shared chat clients/credentials bind their asyncio primitives to the event loop of first use; running all agent coroutines on one persistent loop prevents a silent permanent hang when entities land on different worker threads.
- **HITL is host-agnostic:** the shared orchestrator pauses via `ctx.request_info` and waits for an external event named by the request id; `DurableWorkflowClient.get_pending_hitl_requests` / `send_hitl_response` encapsulate that contract. The orchestration waits indefinitely for the response, matching MAF core's `request_info` and the .NET durable host (no built-in timeout); a configurable timeout can be added later once core defines the contract.
- **Synchronous edge conditions:** the durable orchestrator evaluates edge conditions synchronously, matching the .NET durable host. Async edge conditions are not currently supported on the durable host, so an edge with an async condition is not traversed. Broader async-condition support (evaluating routing decisions as cached activities) is planned as a follow-up.

# Testing

- **Unit:** 422 passed (durabletask + azurefunctions), including workflow event serialization round-trips and the streaming/async client methods.
- **Integration:** Azure Functions and standalone durabletask integration tests pass, including all workflow and HITL paths.
- **Standalone samples (live):** `08_workflow` (legitimate + spam paths) and `09_workflow_hitl` (approve + reject) verified end-to-end against the DTS emulator.

# Notes

- `test_08_dt_workflow` uses a unique per-run task hub, so it needs the DTS emulator started with `DTS_USE_DYNAMIC_TASK_HUBS=true` (as the `azure-functions-integration-setup` action does). It is not a regression and runs in CI via the integration test job.

# Related Issue

Fixes #4877


cgillum:
> The DTS emulator only provisions the default task hub, so test_08_dt_workflow (which uses unique per-run task hubs) requires a real Durable Task Scheduler -- a pre-existing emulator limitation, not a regression.

@ahmedmuhsin the DTS emulator can be set up to support multiple task hub beyond just `default`. You can see how we set it up in `.github\actions\azure-functions-integration-setup\action.yml`:

```bash
docker run -d --name dts-emulator -p 8080:8080 -p 8082:8082 -e DTS_USE_DYNAMIC_TASK_HUBS=true mcr.microsoft.com/dts/dts-emulator:latest
```

The `DTS_USE_DYNAMIC_TASK_HUBS=true` is what allows using task hub names other than `default`.

ahmedmuhsin:
Thanks @cgillum, you're right. The tests use a unique per-run task hub, so it was the emulator running without `DTS_USE_DYNAMIC_TASK_HUBS=true`, not a hard limitation. I've fixed the note in the description.

python/packages/durabletask/agent_framework_durabletask/_workflows/client.py
```diff
@@ -0,0 +1,202 @@
...
+    def __init__(self, client: TaskHubGrpcClient):
+        """Initialize the workflow client wrapper.
+
+        Args:
+            client: The durabletask client instance to wrap.
+        """
+        self._client = client
+        logger.debug("[DurableWorkflowClient] Initialized with client type: %s", type(client).__name__)
+
+    def start_workflow(self, input: Any = None, *, instance_id: str | None = None) -> str:
```
kshyju, line 67: The .NET implementation has an async method to start a workflow (with an option to make it wait for sync wait if needed). Should the python also offer a similar async method?

python/packages/durabletask/agent_framework_durabletask/_workflows/client.py
```diff
@@ -0,0 +1,202 @@
...
+        Returns:
+            The runtime status name (e.g. ``"RUNNING"``, ``"COMPLETED"``), or
+            ``None`` if no state is available for the instance.
+        """
+        state = self._client.get_orchestration_state(instance_id)
+        if state is None:
+            return None
+        return state.runtime_status.name
+
+    def get_pending_hitl_requests(self, instance_id: str) -> list[dict[str, Any]]:
```
kshyju, line 243: The .NET implementation has a stream method (async). Should we have a similar one in python as well? Does the python core (in-proc workflow) has that?

python/packages/durabletask/agent_framework_durabletask/_workflows/client.py
ahmedmuhsin: Good call, added `run_workflow(input, *, wait=True, instance_id=None, timeout_seconds=300)`. By default it starts the workflow and awaits its deserialized output; pass `wait=False` to get the instance id back immediately and drive it yourself (via `stream_workflow` or the HITL methods). This mirrors the shape of the .NET `RunAsync`. The underlying durabletask client is synchronous, so the blocking start/await calls run through `asyncio.to_thread` to keep the event loop free.

python/packages/durabletask/agent_framework_durabletask/_workflows/client.py
ahmedmuhsin: Yes on both. Core in-process workflows stream via `Workflow.run(message, stream=True)`, which yields `WorkflowEvent` objects. I added `stream_workflow(instance_id)` on the durable client to match that surface: an async iterator that yields the same typed `WorkflowEvent` objects, with each event data payload already reconstructed, ending when the workflow reaches a terminal state.

The mechanism differs from in-proc because of the durable boundary. It is brokerless and polls the orchestration custom status, into which the orchestrator publishes accumulated events after each superstep (live execution only, never on replay). Non-agent executors run inside an activity that captures their real events with data; agent executors currently surface coarse invoked/completed lifecycle events. Granularity is per executor and per yielded output, not token-level. Token-level streaming through the durable boundary would need an external broker, which I left as a follow-up.

There is an end-to-end example in `samples/04-hosting/durabletask/10_workflow_streaming`.

kshyju:
My agent review flagged we are missing "sub workflow" support in this PR. @ahmedmuhsin

python/packages/durabletask/agent_framework_durabletask/_workflows/orchestrator.py
```diff
@@ -0,0 +1,911 @@
...
+    if isinstance(start_executor, AgentExecutor):
+        if isinstance(raw_value, str):
+            return raw_value
+        if isinstance(raw_value, (dict, list)):
+            return json.dumps(raw_value)
+        return str(raw_value)
+
+    input_type = _select_primary_input_type(start_executor)
+    if input_type is None:
+        return raw_value
```
kshyju, line 528: My agent review flagged this line saying we should consider sanitizing this value, Can you double check?

kshyju (approved):

python/packages/durabletask/agent_framework_durabletask/_workflows/orchestrator.py
```diff
@@ -0,0 +1,911 @@
...
+            continue
+
+        aggregated: list[Any] = []
+        aggregated_sources: list[str] = []
+        for src in group.source_executor_ids:
+            for msg, msg_source in pending_sources[src]:
+                aggregated.append(msg)
+                aggregated_sources.append(msg_source)
+
+        target_id = group.target_executor_ids[0]
```
gavin-aguiar, line 397: We should check if target_executor_ids have any ids

python/packages/durabletask/agent_framework_durabletask/_workflows/orchestrator.py
```diff
@@ -0,0 +1,911 @@
...
+        logger.debug("FanIn group %s ready, delivering %d messages to %s", group.id, len(aggregated), target_id)
+
+        if target_id not in next_pending_messages:
+            next_pending_messages[target_id] = []
+
+        first_source = aggregated_sources[0] if aggregated_sources else "__fan_in__"
```
gavin-aguiar, line 403: Same with aggregated_sources. check if the list contains anything

python/packages/durabletask/agent_framework_durabletask/_async_bridge.py
```diff
@@ -0,0 +1,89 @@
...
+    live loop across all invocations, preventing cross-loop hangs.
+
+    Args:
+        coro: The coroutine to execute.
+
+    Returns:
+        The coroutine's result.
+    """
+    loop = _ensure_loop()
+    future = asyncio.run_coroutine_threadsafe(coro, loop)
```
gavin-aguiar, line 88: Should we add a timeout for this or else this will block the code from proceeding. Does durabletask enforce a timeout at the activity level?

python/packages/durabletask/agent_framework_durabletask/_workflows/client.py
```diff
@@ -0,0 +1,310 @@
...
+            runtime_status = state.runtime_status.name if state is not None else None
+            if runtime_status in terminal_statuses:
+                return
+
+            if deadline is not None and time.monotonic() >= deadline:
+                raise TimeoutError(f"Workflow '{instance_id}' did not complete within {timeout_seconds}s")
+
+            await asyncio.sleep(poll_interval_seconds)
+
+    def get_pending_hitl_requests(self, instance_id: str) -> list[dict[str, Any]]:
```
gavin-aguiar, line 243: get and send hitl request should be async right since  run_workflow / stream_workflow  are already async.

python/packages/durabletask/agent_framework_durabletask/_workflows/orchestrator.py
ahmedmuhsin: A FanInEdgeGroup always has exactly one target by construction, so target_executor_ids is never empty here and the [0] is safe. If it somehow were empty I'd rather this raise than silently skip a delivery the readiness check just approved, so I'll leave it as is.

python/packages/durabletask/agent_framework_durabletask/_workflows/orchestrator.py
ahmedmuhsin: Same thing here. The readiness check right above requires every source to have pending messages, and a group needs at least two sources to even construct, so aggregated always has content by this point. aggregated_sources[0] is also already guarded with the __fan_in__ fallback.

python/packages/durabletask/agent_framework_durabletask/_async_bridge.py
ahmedmuhsin: durabletask doesn't cancel a running activity, it lets it finish on the worker, so a timeout here would just abandon the result instead of stopping the work. Agent calls can also run for minutes, so a fixed cap would break legit runs. That kind of timeout is better handled at the orchestration layer with a timer, so I'd leave this one off.

python/packages/durabletask/agent_framework_durabletask/_workflows/client.py
ahmedmuhsin: The underlying durabletask client is fully sync, so async here would just mean wrapping in to_thread. These are single shot calls in the same family as start_workflow and get_runtime_status, which are also sync. run_workflow and stream_workflow are async because they block or poll over time. Making only the HITL methods async would be less consistent, not more. Happy to convert the whole primitive set to async as a follow-up if we want a fully async surface.

python/packages/durabletask/agent_framework_durabletask/_workflows/orchestrator.py
ahmedmuhsin: Sub-workflow support is a new feature rather than something this refactor touches. I'd like to keep this PR scoped to standalone hosting and track sub-workflows as a follow-up. I can open an issue for it.

gavin-aguiar (approved):

PR #6451, opened 2026-06-10
Python: Integrate shell tool into harness agent

### LINKED ISSUE #6447: Python: Add Shell feature to harness agent

westey-m:
### Motivation and Context

The .NET `HarnessAgent` already supports an optional shell capability, but the Python
`create_harness_agent` did not. This change brings the Python harness to parity by adding
the shell tool as an opt-in feature, enabled when the caller supplies a shell executor.

### Description

Adds the shell tool as an optional capability to `create_harness_agent`:

- New `SupportsShellTool` `@runtime_checkable` protocol in `_clients.py` (mirrors
  `SupportsWebSearchTool`), exported from the package.
- Two new `create_harness_agent` parameters: `shell_executor` and
  `shell_environment_provider_options`. When `shell_executor` is provided, the harness wires
  the shell tool via `client.get_shell_tool(func=executor.as_function())` and adds a
  `ShellEnvironmentProvider`.
- If the chat client does not support shell tools, a warning is logged and **both** the shell
  tool and the environment provider are skipped (the provider is not useful without an
  execution path).
- `ShellEnvironmentProvider` is imported lazily at runtime because the shell types live in the
  separate `agent-framework-tools` package (which depends on core), so the new params are typed
  loosely with the expected real types documented in the docstring.
- Note: `shell_executor` must be a shell tool exposing `as_function()` (e.g. `LocalShellTool`/
  `DockerShellTool`), since Python's `ShellExecutor` protocol lacks `as_function()` (unlike .NET).

Includes new unit tests (tool + provider wired, custom options forwarded, warn-and-skip when
unsupported, nothing added by default) and a harness README update.

Closes #6447

### Contribution Checklist

### DIFF
--- a/python/packages/core/agent_framework/_harness/_agent.py
+++ b/python/packages/core/agent_framework/_harness/_agent.py
@@ -15,7 +15,7 @@ from collections.abc import Callable, Sequence
 from typing import TYPE_CHECKING, Any
 
 from .._agents import Agent, SupportsAgentRun
-from .._clients import SupportsWebSearchTool
+from .._clients import SupportsShellTool, SupportsWebSearchTool
 from .._compaction import CompactionProvider, ContextWindowCompactionStrategy, ToolResultCompactionStrategy
 from .._feature_stage import ExperimentalFeature, experimental
 from .._sessions import ContextProvider, HistoryProvider, InMemoryHistoryProvider
@@ -106,6 +106,7 @@ def _assemble_context_providers(
     skills_paths: Sequence[str] | None,
     background_agents: Sequence[SupportsAgentRun] | None,
     background_agents_instructions: str | None,
+    shell_context_provider: ContextProvider | None,
     extra_context_providers: Sequence[ContextProvider] | None,
 ) -> list[ContextProvider]:
     """Assemble the ordered list of context providers."""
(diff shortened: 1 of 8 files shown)


eavanvalkenburg (approved):

python/packages/core/agent_framework/_harness/_agent.py
```diff
@@ -137,13 +138,48 @@ def _assemble_context_providers(
...
+        )
+        return None, None
+
+    # Imported lazily: the shell types live in the separate agent-framework-tools package,
+    # which depends on core, so core cannot import them at module load time.
+    from agent_framework_tools.shell import ShellEnvironmentProvider
+
+    shell_tool = client.get_shell_tool(func=shell_executor.as_function())
+    shell_provider = ShellEnvironmentProvider(shell_executor, shell_environment_provider_options)
+    return shell_tool, shell_provider
```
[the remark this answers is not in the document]
westey-m: Good catch — addressed in 851b32fe7. `_assemble_shell` now validates that `shell_executor` exposes a callable `as_function()` upfront (before the client capability check) and raises a clear `TypeError` naming the offending type and the expected contract, instead of a cryptic `AttributeError`. Added unit tests covering both the supporting and non-supporting client paths.

python/packages/core/agent_framework/_harness/_agent.py
```diff
@@ -174,6 +220,8 @@ def create_harness_agent(
     skills_paths: Sequence[str] | None = None,
     background_agents: Sequence[SupportsAgentRun] | None = None,
     background_agents_instructions: str | None = None,
+    shell_executor: Any = None,
+    shell_environment_provider_options: Any = None,
```
TaoChenOSU: nit: should this be typed `ShellEnvironmentProviderOptions | None`?

python/packages/core/agent_framework/_harness/_agent.py
```diff
@@ -174,6 +220,8 @@ def create_harness_agent(
     skills_paths: Sequence[str] | None = None,
     background_agents: Sequence[SupportsAgentRun] | None = None,
     background_agents_instructions: str | None = None,
+    shell_executor: Any = None,
```
TaoChenOSU: nit: same here

python/packages/core/agent_framework/_harness/_agent.py
westey-m: @TaoChenOSU, my understand is that these cannot be typed here, because the shell package depends on the core package rather than the other way around.  So there is some ad hoc type checking done above in _assemble_shell.  Let me know if this is not correct.

TaoChenOSU: I think you're concern about circular dependency, you can do the following:
```
if TYPE_CHECKING:
    from agent_framework import ShellEnvironmentProviderOptions

def __init__(self, ..., shell_environment_provider_options: "ShellEnvironmentProviderOptions | None" = None, ...):
```

westey-m: You're right, @TaoChenOSU — thanks for the nudge. Done in b12823b50: both `shell_executor` and `shell_environment_provider_options` are now typed (`ShellExecutor | None` / `ShellEnvironmentProviderOptions | None`) via a `TYPE_CHECKING` import. Since that block never executes at runtime there's no circular dependency, and the lazy runtime import of `ShellEnvironmentProvider` inside `_assemble_shell` is retained. One detail: the types live in `agent_framework_tools.shell` (not `agent_framework` — they aren't re-exported from core). Verified core's pyright resolves them cleanly. Note `ShellExecutor` is a protocol without `as_function()`, so I invoke the validated `getattr` result rather than `shell_executor.as_function()`.

python/packages/core/agent_framework/_harness/_agent.py
westey-m: Typed in b12823b50 as well — see the reply on the `shell_environment_provider_options` thread for details.

TaoChenOSU (approved):

PR #6463, opened 2026-06-10
.NET: Fix CopySessionConfig() and CopyResumeSessionConfig() to preserve SessionConfig.Streaming value

### LINKED ISSUE #4732: .NET: [Bug]: CopySessionConfig() hardcodes Streaming = true, ignoring SessionConfig.Streaming value
### Description

`GitHubCopilotAgent.CopySessionConfig()` and `CopyResumeSessionConfig()` in `Microsoft.Agents.AI.GitHub.Copilot` always hardcode `Streaming = true` on the copied config, ignoring any value the caller sets on `source.Streaming`. There are scenarios where we want to disable streaming
 
This means there is **no way to disable streaming** when using `AsAIAgent()` with the GitHub Copilot SDK, even though `SessionConfig.Streaming` exists as a settable property.
 
## Steps to Reproduce
 
```csharp
var sessionConfig = new SessionConfig
{
    Streaming = false, // caller explicitly disables streaming
    SystemMessage = "You are a helpful assistant.",
    OnPermissionRequest = PermissionHandler.ApproveAll,
};
 
var agent = copilotClient.AsAIAgent(sessionConfig);
// Agent always uses streaming regardless — CopySessionConfig ignores source.Streaming
```
### Root Cause
```csharp
internal static SessionConfig CopySessionConfig(SessionConfig source)
{
    return new SessionConfig
    {
        Model = source.Model,
        ReasoningEffort = source.ReasoningEffort,
        // ... all other properties copied from source ...
        InfiniteSessions = source.InfiniteSessions,
        Streaming = true  // ← hardcoded, should be: source.Streaming
    };
}
```

### Package Versions

Microsoft.Agents.AI.GitHub.Copilot (verified in 1.0.0-preview.260311.1, likely since 1.0.0-preview.260225.1)

### .NET Version

_No response_

### Additional Context

_No response_

giles17:
### Motivation and Context

`CopySessionConfig()` and `CopyResumeSessionConfig()` in `GitHubCopilotAgent` hardcoded `Streaming = true`, making it impossible to disable streaming when using `AsAIAgent()` with the GitHub Copilot SDK even though `SessionConfig.Streaming` is a settable property.

Fixes #4732

### Description

The root cause was that both copy methods unconditionally set `Streaming = true` on the output config instead of propagating the caller's value. The fix changes both methods to use `source.Streaming ?? true`, preserving an explicitly set `false` value while defaulting to `true` when unset (maintaining backward compatibility). Unit tests are added to verify that `Streaming = false` is preserved and that a null value defaults to `true` for both `CopySessionConfig` and `CopyResumeSessionConfig`.

### Contribution Checklist

---

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.GitHub.Copilot/GitHubCopilotAgent.cs
+++ b/dotnet/src/Microsoft.Agents.AI.GitHub.Copilot/GitHubCopilotAgent.cs
@@ -145,11 +145,12 @@ public sealed class GitHubCopilotAgent : AIAgent, IAsyncDisposable
         // Ensure the client is started
         await this.EnsureClientStartedAsync(cancellationToken).ConfigureAwait(false);
 
-        // Create or resume a session with streaming enabled
+        // Create or resume a session with streaming enabled by default
         SessionConfig sessionConfig = this._sessionConfig != null
             ? CopySessionConfig(this._sessionConfig)
             : new SessionConfig { Streaming = true };
 
+        bool isStreaming = sessionConfig.Streaming ?? true;
         CopilotSession copilotSession;
         if (typedSession.SessionId is not null)
         {
@@ -178,7 +179,7 @@ public sealed class GitHubCopilotAgent : AIAgent, IAsyncDisposable
                         break;
 
                     case AssistantMessageEvent assistantMessage:
-                        channel.Writer.TryWrite(this.ConvertToAgentResponseUpdate(assistantMessage));
+                        channel.Writer.TryWrite(this.ConvertToAgentResponseUpdate(assistantMessage, isStreaming));
                         break;
 
                     case AssistantUsageEvent usageEvent:
@@ -326,12 +327,18 @@ public sealed class GitHubCopilotAgent : AIAgent, IAsyncDisposable
         };
     }
 
-    internal AgentResponseUpdate ConvertToAgentResponseUpdate(AssistantMessageEvent assistantMessage)
+    /// <summary>
+    /// Converts an <see cref="AssistantMessageEvent"/> to an <see cref="AgentResponseUpdate"/>.
+    /// When streaming is enabled, text was already delivered via delta events, so only raw metadata is emitted.
+    /// When streaming is disabled, the full message text is emitted as <see cref="TextContent"/>.
+    /// </summary>
+    internal AgentResponseUpdate ConvertToAgentResponseUpdate(AssistantMessageEvent assistantMessage, bool isStreaming)
     {
-        AIContent content = new()
-        {
-            RawRepresentation = assistantMessage
-        };
+        // When streaming, text was already delivered via AssistantMessageDeltaEvent.
+        // When not streaming, this is the only opportunity to emit the response text.
+        AIContent content = isStreaming
+            ? new AIContent { RawRepresentation = assistantMessage }
+            : new TextContent(assistantMessage.Data?.Content ?? string.Empty) { RawRepresentation = assistantMessage };
 
         return new AgentResponseUpdate(ChatRole.Assistant, [content])
         {
--- a/dotnet/src/Microsoft.Agents.AI.GitHub.Copilot/GitHubCopilotAgent.cs
+++ b/dotnet/src/Microsoft.Agents.AI.GitHub.Copilot/GitHubCopilotAgent.cs
@@ -271,19 +271,20 @@ public sealed class GitHubCopilotAgent : AIAgent, IAsyncDisposable
     }
 
     /// <summary>
-    /// Copies all supported properties from a source <see cref="SessionConfig"/> into a new instance
-    /// with <see cref="SessionConfigBase.Streaming"/> set to <c>true</c>.
+    /// Copies all supported properties from a source <see cref="SessionConfig"/> into a new instance,
+    /// preserving <see cref="SessionConfigBase.Streaming"/> from the source (defaulting to <c>true</c> if unset).
     /// </summary>
     internal static SessionConfig CopySessionConfig(SessionConfig source)
     {
         SessionConfig copy = source.Clone();
-        copy.Streaming = true;
+        copy.Streaming = source.Streaming ?? true;
         return copy;
     }
 
     /// <summary>
     /// Copies all supported properties from a source <see cref="SessionConfig"/> into a new
-    /// <see cref="ResumeSessionConfig"/> with <see cref="SessionConfigBase.Streaming"/> set to <c>true</c>.
+    /// <see cref="ResumeSessionConfig"/>, preserving <see cref="SessionConfigBase.Streaming"/>
+    /// from the source (defaulting to <c>true</c> if unset).
     /// </summary>
     internal static ResumeSessionConfig CopyResumeSessionConfig(SessionConfig? source)
     {
@@ -306,7 +307,7 @@ public sealed class GitHubCopilotAgent : AIAgent, IAsyncDisposable
             SkillDirectories = source?.SkillDirectories,
             DisabledSkills = source?.DisabledSkills,
             InfiniteSessions = source?.InfiniteSessions,
-            Streaming = true
+            Streaming = source?.Streaming ?? true
         };
     }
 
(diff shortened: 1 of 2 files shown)


dotnet/src/Microsoft.Agents.AI.GitHub.Copilot/GitHubCopilotAgent.cs
```diff
@@ -271,19 +271,20 @@ private ResumeSessionConfig CreateResumeConfig()
...
-    /// Copies all supported properties from a source <see cref="SessionConfig"/> into a new instance
-    /// with <see cref="SessionConfigBase.Streaming"/> set to <c>true</c>.
+    /// Copies all supported properties from a source <see cref="SessionConfig"/> into a new instance,
+    /// preserving <see cref="SessionConfigBase.Streaming"/> from the source (defaulting to <c>true</c> if unset).
     /// </summary>
     internal static SessionConfig CopySessionConfig(SessionConfig source)
     {
         SessionConfig copy = source.Clone();
-        copy.Streaming = true;
+        copy.Streaming = source.Streaming ?? true;
```
giles17, line 281: This change exposes a code path the agent does not handle correctly yet. After `Streaming=false` is preserved here (same issue for line 310), `ConvertToAgentResponseUpdate(AssistantMessageEvent)` still returns only raw `AIContent` at lines 329-342. The nearby test (`GitHubCopilotAgentTests.cs:291-309`) says that is correct only because the message content was "already delivered via delta events." In a non-streaming session there may be no delta events, so the final assistant text is silently dropped. The PR needs an end-to-end non-streaming response path, not just config-copy changes.

giles17 (commented):
## Automated Code Review

**Reviewers:** 5 | **Confidence:** 95%

### ✓ Correctness
> The PR correctly fixes the bug where CopySessionConfig and CopyResumeSessionConfig were hardcoding Streaming=true, ignoring the caller's explicit value. The fix uses null-coalescing (source.Streaming ?? true / source?.Streaming ?? true) which preserves an explicit false, preserves an explicit true, and defaults to true when unset. The Clone() call in CopySessionConfig already copies all properties, but the subsequent assignment correctly handles the null→true default. Tests cover all meaningful cases. No issues found.

### ✓ Security Reliability
> This PR correctly fixes the streaming configuration preservation bug by replacing hardcoded `Streaming = true` with null-coalescing logic (`source.Streaming ?? true` and `source?.Streaming ?? true`). The change properly preserves an explicitly set `false` while defaulting to `true` when unset. No security or reliability concerns: no injection risks, no resource leaks, null handling is correct (CopyResumeSessionConfig uses `source?.Streaming` for its nullable parameter), and backward compatibility is maintained for calers that don't explicitly set Streaming.

### ✓ Test Coverage
> The test coverage for this change is thorough. The PR adds four new tests covering both `CopySessionConfig` and `CopyResumeSessionConfig` for the two key scenarios: `Streaming = false` (the bug fix) and `Streaming = null` (backward-compatible default to `true`). The existing tests (`CopySessionConfig_CopiesAllProperties`, `CopyResumeSessionConfig_CopiesAllProperties`, `CopyResumeSessionConfig_WithNullSource_ReturnsDefaults`) continue to validate backward compatibility. The explicit `Streaming = true` case is not separately tested, but it's logically equivalent to the null-defaults-to-true path from an output perspective, making it a non-gap. Assertions are meaningful and specific (Assert.True/Assert.False on the boolean property).

### ✓ Failure Modes
> The PR correctly fixes the streaming value preservation bug. Both CopySessionConfig (using Clone + null-coalescing override) and CopyResumeSessionConfig (using manual construction with null-conditional + null-coalescing) properly preserve explicit Streaming=false while defaulting to true when unset. No silent failure modes, lost errors, or operational issues found.

### ✗ Design Approach
> I found one blocking design gap. This PR now preserves `Streaming=false`, but the agent’s response-conversion path still assumes assistant text was already delivered via delta events. In the non-streaming case enabled by this change, that assumption is false, so final assistant text can be silently dropped instead of surfaced in the response.

### Flagged Issues
- [ ] Preserving `SessionConfig.Streaming = false` is not end-to-end supported: `GitHubCopilotAgent.cs:280` and `:310` now allow non-streaming sessions, but `GitHubCopilotAgent.cs:329-342` converts `AssistantMessageEvent` into raw `AIContent` only, and `GitHubCopilotAgentTests.cs:291-309` explicitly documents that this is safe only because the text was 'already delivered via delta events'. With streaming disabled, that premise no longer holds, so the assistant's final text can be lost.

---

dotnet/tests/Microsoft.Agents.AI.GitHub.Copilot.UnitTests/GitHubCopilotAgentTests.cs
```diff
@@ -235,11 +302,62 @@ public void ConvertToAgentResponseUpdate_AssistantMessageEvent_DoesNotEmitTextCo
...
+        const string TestId = "agent-id";
+        var agent = new GitHubCopilotAgent(copilotClient, ownsClient: false, id: TestId, tools: null);
+
+        // Act
+        AgentResponseUpdate result = agent.ConvertToAgentResponseUpdate(assistantMessage, isStreaming: false);
+
+        // Assert - should emit empty TextContent rather than throwing.
+        Assert.Empty(result.Text);
+        Assert.Contains(result.Contents, c => c is TextContent);
+    }
```
giles17, line 362: The production code has `assistantMessage.Data?.Content ?? string.Empty`, handling `Data` being null. Consider adding a test where `Data` is null (not just where `Content` is empty) to cover that null-propagation path and document the expected behavior.

giles17 (commented):
## Automated Code Review

**Reviewers:** 5 | **Confidence:** 91%

### ✓ Correctness
> The PR correctly adds conditional text content emission based on streaming mode. When streaming, text was already delivered via delta events so only raw metadata is emitted (avoiding duplicates). When not streaming, the full message text is properly emitted as TextContent. The isStreaming variable is correctly derived from the resolved sessionConfig.Streaming value, captured safely by the event subscription closure, and the null-conditional handling of assistantMessage.Data?.Content is correct. No correctness issues found.

### ✓ Security Reliability
> The PR cleanly adds non-streaming support to ConvertToAgentResponseUpdate by threading an isStreaming boolean derived from the session config. Null handling is defensive (Data?.Content ?? string.Empty), the captured variable in the closure is a value type so there's no race condition, and tests cover both streaming and non-streaming paths including the empty-content edge case. No security or reliability issues found.

### ✓ Test Coverage
> The PR adds good test coverage for the new `isStreaming` parameter on `ConvertToAgentResponseUpdate`. The streaming path, non-streaming path, and empty-content edge case are all tested with meaningful assertions (content equality, type checks, reference identity). One minor gap: the production code handles `assistantMessage.Data?.Content ?? string.Empty` (null-coalescing on `Data`), but no test exercises the `Data = null` scenario in non-streaming mode. This is a low-severity gap since the null-coalescing ensures it won't crash, but a test would document the expected behavior.

### ✓ Failure Modes
> The PR correctly handles the streaming/non-streaming distinction for AssistantMessageEvent conversion. The `isStreaming` flag is derived from the same session config sent to the server, null handling is proper throughout (`Data?.Content ?? string.Empty`), and the code paths for both modes are well-tested. No concrete failure modes or silent error paths were identified.

### ✓ Design Approach
> I did not find a design-approach issue in this diff. The change is narrowly scoped to the demonstrated bug: `AgentResponseUpdate.Text` is defined as the concatenation of `TextContent` items in `Contents` (`dotnet/src/Microsoft.Agents.AI.Abstractions/AgentResponseUpdate.cs:91-97`), so emitting `TextContent` only for non-streaming assistant messages is consistent with the framework contract and with the new tests. I also did not find repo evidence that a better pattern should be used here instead of keying off the session's streaming configuration.

---

dotnet/src/Microsoft.Agents.AI.GitHub.Copilot/GitHubCopilotAgent.cs
giles17: Fixed in `79b662af`.

dotnet/tests/Microsoft.Agents.AI.GitHub.Copilot.UnitTests/GitHubCopilotAgentTests.cs
giles17: Fixed in `79b662af`.

giles17 (commented):
## Automated Code Review

**Reviewers:** 5 | **Confidence:** 94% | **Result:** All clear

Reviewed: Correctness, Security Reliability, Test Coverage, Failure Modes, Design Approach

---

peibekwe (approved):

westey-m (approved):

PR #6468, opened 2026-06-10
Python: Bug fix for declarative workflows

peibekwe:
### Description

Bug fix for declarative workflows.

### Contribution Checklist

### DIFF
--- a/python/packages/declarative/agent_framework_declarative/_workflows/_declarative_base.py
+++ b/python/packages/declarative/agent_framework_declarative/_workflows/_declarative_base.py
@@ -269,6 +269,9 @@ class DeclarativeWorkflowState:
     - Conversation: Conversation history
     """
 
+    # Sentinel marking "no prior value" for temporary-key bookkeeping.
+    _MISSING: Any = object()
+
     def __init__(self, state: State, env_config: DeclarativeEnvConfig | None = None):
         """Initialize with a State instance.
 
@@ -492,6 +495,15 @@ class DeclarativeWorkflowState:
         else:
             raise ValueError(f"Cannot append to non-list at path '{path}'")
 
+    def _clear_local_path(self, name: str) -> None:
+        """Remove ``name`` from the ``Local`` namespace, if present."""
+        state_data = self.get_state_data()
+        local = cast(dict[str, Any], state_data.get("Local"))
+        if local is None or name not in local:
+            return
+        local.pop(name, None)
+        self.set_state_data(state_data)
+
     def eval(self, expression: str) -> Any:
         """Evaluate a PowerFx expression with the current state.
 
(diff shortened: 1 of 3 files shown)


python/packages/declarative/tests/test_declarative_state_path_safety.py
```diff
@@ -0,0 +1,251 @@
...
+# pyright: reportPrivateUsage=false, reportUnknownVariableType=false
+# pyright: reportGeneralTypeIssues=false
+
+"""Path-segment validation tests for DeclarativeWorkflowState.
+
+Path segments handed to ``get``/``set``/``append`` and ``{Variable.Path}``
+placeholders in ``interpolate_string`` must be valid declarative identifiers
+(``[A-Za-z][A-Za-z0-9_]*``). These tests pin that contract and the related
+behavior across legitimate and invalid inputs.
+"""
```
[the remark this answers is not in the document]
peibekwe: Addressed.

python/packages/declarative/agent_framework_declarative/_workflows/_declarative_base.py
```diff
@@ -692,7 +720,7 @@ def _preprocess_custom_functions(self, formula: str) -> str:
                 if isinstance(replacement, str):
                     if len(replacement) > MAX_INLINE_LENGTH:
                         # Store long strings in a temp variable to avoid PowerFx expression limit
-                        temp_var_name = f"_TempMessageText{temp_var_counter}"
+                        temp_var_name = f"TempMessageText{temp_var_counter}"
```
moonbox3: Could this preserve or isolate any existing `Local.TempMessageText0` value? Since the counter restarts for each preprocessing call and the scratch variable now uses a normal public `Local` name, evaluating a long `MessageText(...)` expression can permanently overwrite workflow state that the user already stored at `Local.TempMessageText0`. Could we generate a collision-free scratch path or snapshot and restore/delete the temporary value after PowerFx evaluation?

python/packages/declarative/agent_framework_declarative/_workflows/_declarative_base.py
```diff
@@ -866,8 +897,8 @@ def replace_var(match: re.Match[str]) -> str:
             value = self.get(var_path)
             return str(value) if value is not None else ""
 
-        # Match {Variable.Path} patterns
-        pattern = r"\{([A-Za-z][A-Za-z0-9_.]*)\}"
+        # Match {Variable.Path} patterns where each segment is a declarative identifier.
+        pattern = r"\{([A-Za-z][A-Za-z0-9_]*(?:\.[A-Za-z][A-Za-z0-9_]*)*)\}"
```
moonbox3: Should interpolation keep matching non-empty dotted state paths and let `get()` decide whether each segment is safe? `get()` still supports arbitrary dict-key segments, but this matcher now leaves placeholders such as `{Local.bag._id}`, `{Local.bag.1}`, or conversation-id keyed paths literal even when they resolve safely through dict lookup. That looks like a backwards-compatible surface break for SendActivity templates that previously resolved those state values. Are we okay with that?

eavanvalkenburg (approved):

python/packages/declarative/agent_framework_declarative/_workflows/_declarative_base.py
peibekwe: Good point! Restored the underscore-prefixed `_TempMessageText` name and made `eval` snapshot the prior value before each write, restoring it in finally; so user state at that path is preserved and the temp key is removed even if PowerFx raises.

python/packages/declarative/agent_framework_declarative/_workflows/_declarative_base.py
peibekwe: Tightened the matcher so root segment must be an identifier, follow-on segments accept any non-empty dict-key (so `{Local.bag._id}`, numeric keys, and UUID-style conversation-id paths now resolve), while non-state literals like `{foo-bar}` stay literal.

moonbox3 (approved):

PR #6487, opened 2026-06-11
.NET: Scope argument-based standing approvals correctly in ToolApprovalAgent

### LINKED ISSUE #6486: Python: .NET: Harden argument scoping for standing approval rules in ToolApprovalAgent
### Summary
Review and harden how `ToolApprovalAgent` records and matches **standing approval rules** in `Microsoft.Agents.AI` (`dotnet/src/Microsoft.Agents.AI/Harness/ToolApproval/`). The goal is to ensure an argument-scoped standing approval stays scoped exactly as the user intended for all argument shapes, and that the .NET behavior matches the existing Python implementation in `agent_framework/_harness/_tool_approval.py`.

### Scope
- `ToolApprovalAgent.cs`, `ToolApprovalRule.cs`, `ToolApprovalRequestContentExtensions.cs`
- Align the argument-normalization/matching contract with the Python harness, whose `ToolApprovalRule` semantics already document the intended behavior.

### Acceptance criteria
- Argument-scoped standing approvals match only the intended calls across all argument shapes.
- Behavior is consistent between the .NET and Python tool-approval harnesses.
- Regression tests cover the standing-rule argument-matching paths.

rogerbarreto:
Fixes #6486

Scopes argument-based standing approvals correctly in `ToolApprovalAgent`. An argument-scoped standing approval (the "always approve with exact arguments" path) now records an empty argument set instead of `null` when the approved call has no arguments, so it matches only future no-argument calls. `null` stays reserved exclusively for tool-level approvals, keeping the two scopes distinct. Aligns .NET behavior with the existing Python harness.

### Changes
- `ToolApprovalAgent.SerializeArguments` returns a non-null (empty) dictionary for empty/null args.
- Clarified XML docs on `ToolApprovalRule.Arguments` semantics.
- Regression tests: no-argument standing-approval flow, `MatchesRule` argument scoping, empty-args rule serialization.

### Validation
- Build: 0 warnings / 0 errors.
- Tests: 1652 (net10.0) + 1542 (net472) pass.
- `dotnet format` clean on changed projects.


westey-m (approved):

SergeyMenshykh (approved):

peibekwe (approved):

rpelevin:
This is a good boundary to lock down in the approval harness.

The important distinction is that "no arguments" and "all arguments" are different approval scopes. Treating an exact-arguments approval for an empty call as the same shape as a tool-level approval can silently widen authority on the next run.

I would keep three cases separate in the rule model:

- tool-level rule: the operator approved the tool regardless of arguments;
- exact empty-arguments rule: the operator approved only the no-argument call shape;
- exact argument rule: the operator approved the specific normalized argument set.

That gives the matcher a clean invariant:

a standing approval can only match a future call whose tool identity and argument scope are the same as the approved call.

The regression surface I would preserve across both runtimes is:

1. empty-arguments approval does not match a later argument-bearing call;
2. argument-bearing approval does not match missing, extra, reordered, or differently serialized arguments;
3. tool-level approval is the only state allowed to ignore arguments;
4. persisted approval rules round-trip without converting empty args to tool-level scope;
5. Python and .NET produce the same match result for the same normalized call envelope.

That keeps "always approve with exact arguments" from becoming a broader standing capability than the user granted.

PR #6500, opened 2026-06-12
docs: ADR-0033 feature-usage bitmask in the User-Agent

eavanvalkenburg:
### Motivation & Context

Agent Framework currently identifies SDK traffic but does not show which
framework capabilities are observed together. This docs-only design
defines a transparent, opt-out feature-usage mask for approved first-party
request paths whose User-Agent telemetry can be ingested.

V1 starts at 128 bits so normal package and capability growth does not
immediately force a registry v2.

- **What are the major changes?** Adds ADR-0033, SPEC-004, and the per-language
  registry. The design uses a 128-bit process-global monotonic mask encoded as
  `(feat=vN.<hex>)`, with independent Python/.NET numbering and a dedicated
  mask-only opt-out. Python's existing whole-User-Agent opt-out also suppresses
  its mask. Bits are sticky process-lifetime Booleans: repeated requests carrying
  a bit are not additional feature uses or request-level attribution.
  Emission requires both an approved client/pipeline and an approved actual HTTPS
  origin on every request/redirect hop; custom origins are default-deny. OpenAI
  hooks preserve SDK transport defaults through `DefaultAsyncHttpxClient`. Usage
  is marked at first meaningful activation rather than DI construction. Optional
  packages own private `FeatureIndex` declarations, with a repository test
  enforcing registry parity, full coverage, valid range, and no overlapping
  indexes. The candidate registry contains 63 Python and 52 .NET indexes,
  including separate declarative-agent and declarative-workflow signals,
  Foundry evals, and stable skill categories.
- **What is the impact of these changes?** Documentation only; runtime behavior
  is unchanged. The proposal establishes the future wire format, privacy and
  redirect boundaries, activation rules, package release independence,
  v1-to-v2 migration, and implementation plan. A full v1 payload remains bounded
  to 32 hex characters (43 added User-Agent bytes including the wrapper).
- **What do you want reviewers to focus on?** The destination classifier,
  activation semantics, package-local index ownership and uniqueness checks,
  declarative agent/workflow split, 128-bit boundary, privacy implications of a
  process-global sticky mask, its deliberately non-countable interpretation, and
  the .NET two-lane snapshot semantics.

### Related Issue

Part of #7240. No other open PR addresses this issue.

### Contribution Checklist

### DIFF
--- a/docs/decisions/0027-feature-usage-bitmask-user-agent.md
+++ b/docs/decisions/0027-feature-usage-bitmask-user-agent.md
@@ -1,0 +1,281 @@
+---
+status: proposed
+contact: eavanvalkenburg
+date: 2026-06-12
+deciders: eavanvalkenburg
+consulted:
+informed:
+---
+
+# Feature-usage bitmask in the User-Agent
+
+## Context and Problem Statement
+
+We can see which Agent Framework packages are installed and that *some* framework
+call happened (via the existing `agent-framework-python/{version}` User-Agent),
+but we have no usage-based signal about **which features are actually exercised**
+at runtime, nor which are used *together* (e.g. workflows + MCP + Foundry). How
+can we collect a lightweight, privacy-respecting signal of feature usage for the
+traffic we can actually read, without standing up new event pipelines?
+
+The detailed mechanism is in [SPEC-002](../specs/002-feature-usage-telemetry.md);
+the per-language bit lists are in
+[`docs/feature-usage-bit-registry.json`](../feature-usage-bit-registry.json).
+
+## Decision Drivers
+
+- **Transparency** — openly documented, human-decodable, user-controllable. No
+  hidden or obfuscated telemetry.
+- **First-party scope / no third-party leakage** — emit only to Azure/Foundry
+  endpoints (the telemetry we can ingest); never leak a feature fingerprint into
+  third-party logs we cannot read.
+- **Live signal** — reflect features exercised *so far*, re-evaluated per request,
+  not frozen at client construction.
+- **Low cost / few moving parts** — reuse telemetry already in the request path;
+  near-zero runtime overhead; as little machinery as the job needs.
+- **Privacy** — encode only coarse boolean feature usage; no identifiers,
+  arguments, prompts, or payloads.
+
+## Considered Options
+
+The options below are grouped by the decisions that matter: the **transport**,
+the **granularity**, and the **registry sharing model**.
+
+### Transport
+
+#### A. User-Agent token, first-party only, per request (chosen)
+
+Stamp a `(feat=...)` comment onto the UA, but only on Azure/Foundry clients, and
+re-evaluate it per request.
+
+- Good, reuses telemetry already sent to the one backend we can read.
+- Good, per-request stamping reflects the live mask (not frozen at construction).
+- Good, first-party scoping means no fingerprint leaks to third-party providers.
+- Good, maps onto .NET's existing per-request UA pipeline policies unchanged.
+- Bad, no signal for traffic that never hits a first-party endpoint (accepted —
+  we couldn't read it anyway).
+
+#### B. User-Agent token on all clients
+
+- Good, simplest to wire (one static header).
+- Bad, sends a deployment fingerprint to OpenAI/Anthropic/AWS/Google logs we
+  cannot read — privacy leak for zero benefit.
+- Bad, baked into static `default_headers`, so it freezes at client construction
+  and reports a near-empty mask.
+
+#### C. OpenTelemetry span/resource attribute
+
+- Good, precise per-call usage; no UA change.
+- Bad (**privacy — the main reason to hold it**), a span attribute broadcasts the
+  feature-combination fingerprint into the user's **general** telemetry pipeline,
+  which is typically exported to third-party APM vendors (Datadog, Honeycomb, …).
+  That re-introduces exactly the fingerprint leakage the first-party-only UA
+  scoping (A) was chosen to avoid — just into a different set of third parties.
+- Bad (secondary), also a cardinality footgun (a growing, combinatorial value
+  must never become a metric dimension).
+- Neutral, for the team's own goal it reaches us only if the user exports to
+  Azure Monitor and we query it.
+- **Deferred, not rejected.** The version prefix lets us add it later **if** the
+  privacy review blesses a broadly-emitted mask (or a scoped/redacted variant)
+  and a concrete query needs the per-call precision.
+
+#### D. Bespoke usage events
+
+- Good, richest detail and flexibility.
+- Bad, new data flow and cost; larger privacy surface; heavy to build and review;
+  overkill for a coarse "which features" signal.
+
+#### E. Install/import-time signal only (status quo-ish)
+
+- Good, zero new runtime work.
+- Bad, measures installation, not usage; cannot capture feature combinations —
+  does not solve the problem.
+
+### Granularity
+
+#### F. Per package, with core broken out per feature/provider (chosen)
+
+- Good, ~50 bits (Python) / ~40 (.NET) fit a **64-bit** mask, which keeps .NET's
+  accumulator lock-free (`Interlocked.Or`) and the registry hand-maintainable.
+- Good, matches the actual questions ("which orchestration / which built-in
+  provider / which package?") — each orchestration pattern and each built-in
+  context/history provider gets its own bit, since they serve different purposes.
+- Neutral, cannot distinguish sub-features *within* a provider package (e.g.
+  openai chat vs embeddings) until a bit is promoted.
+
+#### G. Per construct (one bit per instantiable type)
+
+- Good, finest detail.
+- Bad, ~96 bits forces a 128-bit mask, which forfeits .NET's lock-free
+  `Interlocked.Or` (needs a lock / `UInt128`).
+- Bad, ~96 call sites across two SDKs; the sheer count pushes toward code
+  generation and extra tests — machinery to manage machinery.
+- Bad, precision nobody's decision actually needs.
+
+### Registry sharing model
+
+#### H. Per-language bit lists (chosen)
+
+Each SDK owns an independent list; the decoder picks the list using the language
+already present in the UA product token.
+
+- Good, **no cross-language coordination**: each SDK numbers and evolves its
+  features independently; adding a Python feature never touches .NET numbering.
+- Good, no null placeholders for one-SDK features, no "same bit, same meaning"
+  rule, no SDK-aware decode caveats.
+- Good, decoding is trivial: language (from UA) + version -> list -> AND.
+- Neutral, two small lists to maintain instead of one (but they were going to
+  diverge anyway — the packages differ).
+
+#### I. Single shared cross-language registry
+
+- Good, one list, one number space.
+- Bad, forces synchronized numbering and null placeholders for features that
+  exist in only one SDK, plus SDK-aware decode rules.
+- Bad, the synchronization is pure accidental complexity — **the language is
+  already in the User-Agent**, so sharing the number space buys nothing.
+
+### Registry maintenance
+
+#### J. Hand-written enum + parity test (chosen)
+
+- Good, ~40 members that change a few times a year; a 10-line test (enum vs JSON
+  list) is enough.
+- Good, no build step, no generator to own.
+
+#### K. Code-generate the enums from the registry
+
+- Bad, a generator + drift test + schema test to maintain a short list of
+  integer constants; justified only by the per-construct bit count we rejected.
+
+### Representation (how the mask is rendered as text)
+
+All examples below encode the same mask — bits 0, 2, 16, 22, 27 set
+(agent + workflow + sequential-orchestration + foundry.chat_client + openai, in
+the Python v1 list) = decimal `138477573`.
+
+#### L. Decimal — `feat=v1.138477573`
+
+- Good, human-familiar; trivial to parse.
+- Neutral, no visual alignment to bit/nibble boundaries; slightly longer than hex
+  for large masks. No advantage over hex.
+
+#### M. Hex (chosen) — `feat=v1.8410005`
+
+- Good, compact (≤16 chars for a 64-bit mask).
+- Good, decodes with one stdlib call in every language (`int(x, 16)` /
+  `Convert.ToUInt64(x, 16)`); nibble boundaries are eyeball-able.
+- Good, lowercase, no `0x` prefix, no leading zeros — unambiguous and stable.
+
+#### N. Binary / bit-list — `feat=v1.1000010000010000000000000101` or `feat=v1.0,2,16,22,27`
+
+- Good, most directly human-readable ("which bits").
+- Bad, longest form in the UA; the bit-list needs delimiter handling and grows
+  with the number of set bits.
+
+#### O. Alphabet / base-N (e.g. Crockford base32 `feat=v1.442005`, base62 `feat=v1.9n2lf`)
+
+- Good, shortest representation.
+- Bad, needs a custom alphabet + decode table on both ends; base62 is
+  case-sensitive (fragile through case-normalizing intermediaries); not
+  eyeball-able. Premature optimization for a value that is already ≤16 chars in
+  hex.
+
+## Decision Outcome
+
+Chosen: **a per-request, first-party-only User-Agent `(feat=...)` token (A),
+with per-package granularity (F), per-language bit lists (H), hand-written enums
+kept honest by a parity test (J), rendered as lowercase hex (M).**
+
+This is the smallest design that answers the question. A 64-bit mask accumulates
+from universal `mark_feature_used()` calls; the token is stamped per request only
+on Azure/Foundry clients (live, no third-party leak); each SDK owns an
+independent bit list selected by the language already in the UA; the mask is
+rendered as hex (`feat=v1.8410005`). OTel (C) is deferred — mainly because a
+broadly-emitted span attribute would leak the fingerprint into the user's general
+telemetry, against the first-party-only stance — but left open behind the version
+prefix. Per-construct granularity (G), a shared registry (I), codegen (K), and the
+decimal/binary/base-N representations (L, N, O) are rejected as complexity or
+length the problem does not require.
+
+### Consequences
+
+- Good, adds usage signal at near-zero cost, no new data flow, few moving parts.
+- Good, transparent (public registry, human-decodable token) and disabled by the
+  existing User-Agent opt-out.
+- Good, first-party-only + per-request emission gives a live mask and no
+  third-party fingerprint leak.
+- Good, 64-bit keeps .NET lock-free; per-language lists remove all cross-language
+  sync; hand-written enums avoid a codegen toolchain.
+- Neutral, the token's reach equals first-party traffic; broader per-call signal
+  (OTel) can be added later if needed.
+- Bad, each feature must add a `mark_feature_used()` call, and first-party clients
+  need a per-request hook (small, mirrors existing patterns).
+
+## Registry versioning and migration (v1 → v2)
+
+The token carries a **per-language** version (`feat=v1.<hex>`); a version bump is
+independent for Python and .NET.
+
+- **Additive growth stays on v1 — no bump.** Allocating a new feature to a
+  reserved/unused bit is backward-compatible: an older decoder simply sees an
+  unknown high bit and ignores it. Normal package growth never needs a new
+  version.
+- **A bump (v2) is required only for breaking changes:** renumbering or
+  re-partitioning existing bits, changing the *meaning* of an already-assigned
+  bit, or widening beyond 64-bit. Within a version a bit is **never** reused or
+  reassigned — that invariant is what lets old decoders stay correct.
+- **Mixed-version coexistence is the norm.** A fleet runs many SDK releases at
+  once, so `v1` and `v2` tokens appear simultaneously for a long time (old SDKs
+  keep emitting `v1`). The decoder keeps **every** published `(language,
+  version)` table and selects by the token's version; the `v1` table is retained
+  indefinitely for historical decode.
+- **Unknown version → do not guess.** A decoder without the `vN` table must
+  record "unknown registry version" rather than decode against an older table —
+  bit meanings may differ across versions, so mis-attribution is worse than
+  no data.
+- **Producing v2:** publish the v2 table alongside v1 in the registry doc, bump
+  that SDK's `FeatureBit` enum + version constant; the SDK emits `v2` from the
+  release it ships in. Prefer staying on v1 (additive) and reserving a clean v2
+  for an eventual deliberate re-partition.
+
+## Limitations
+
+| Limitation | Caused by (choice) | Why we accepted it |
+| --- | --- | --- |
+| **No signal for self-hosted or third-party-only traffic.** If a process never calls Azure/Foundry, we see nothing. | First-party-only emission (A) | We can't read third-party logs anyway, and must not leak a fingerprint into them. Reach traded for privacy. |
+| **No OTel / per-call signal in v1.** | OTel deferred (C) — primarily on **privacy** grounds | A broadly-emitted span attribute would push the fingerprint into the user's general telemetry / third-party APM vendors, undoing the first-party-only scoping. Left open to add later if there is a compelling reason to add. |
+| **Mask reflects "usage so far," not the whole session.** Early requests carry fewer bits than later ones. | Process-global accumulator + per-request stamping | Honest and still useful; the team aggregates across requests. The per-request design is what makes it *grow* rather than freeze. |
+| **No per-agent / per-call attribution.** The mask is one process-wide value — "this process used X", not "this agent/call used X". | Single global accumulator (simplicity) | Per-call attribution is what the deferred OTel span path would add; not needed for portfolio-level questions. |
+| **Coarse granularity.** Can't distinguish sub-features (e.g. openai chat vs embeddings, which shell tool). | Per-package granularity (F) + 64-bit (keeps .NET lock-free) | Matches the actual questions; finer bits can be promoted later behind the version prefix. |
+| **Fingerprinting risk is reduced, not eliminated.** A feature-combination mask is still a deployment signature, and it transits intermediaries (proxies/CDNs) even when first-party-scoped. | Emitting any feature-combination value | Scope + opt-out + coarse granularity mitigate it; residual risk is the subject of the privacy review below. |
+
+## Open Questions (for decider discussion)
+
+These are unresolved and should be decided before/at approval:
+
+1. **Privacy / telemetry-acceptance review (blocking).** Is a coarse,
+   first-party-only, opt-out-able feature-combination mask acceptable telemetry?
+   Even scoped, it transits intermediaries and is a deployment fingerprint. This
+   is a **release precondition**. Possible outcomes that would change the design:
+   require a dedicated opt-out flag (Q2), coarser granularity, hashing, or
+   explicit opt-in.
+2. **Dedicated opt-out flag?** v1 reuses `AGENT_FRAMEWORK_USER_AGENT_DISABLED`
+   (mask dies with the whole UA). Do we add a mask-only flag now (keep base UA,
+   drop the fingerprint), or wait until asked / until the privacy review requires
+   it?
+3. **When (if ever) to add the OTel path?** Held back mainly for **privacy**: a
+   span attribute broadcasts the fingerprint into the user's general telemetry
+   and onward to third-party APM vendors, contradicting the first-party-only
+   stance. It also carries a metric-cardinality hazard. Would the privacy review
+   allow a broadly-emitted mask, a scoped/redacted variant, or none? Decide if/when
+   to revisit.
+
+## More Information
+
+- Mechanism & API: [SPEC-002](../specs/002-feature-usage-telemetry.md)
+- Per-language bit lists: [`docs/feature-usage-bit-registry.json`](../feature-usage-bit-registry.json)
+- Encoding / opt-out / governance prose: [feature-usage-bit-registry.md](../specs/feature-usage-bit-registry.md)
+- Existing accumulator pattern: `python/packages/core/agent_framework/_telemetry.py`
+- .NET emission policies: `dotnet/src/Microsoft.Agents.AI.Foundry/AgentFrameworkUserAgentPolicy.cs`,
+  `dotnet/src/Microsoft.Agents.AI.Foundry.Hosting/HostedAgentUserAgentPolicy.cs`
--- a/docs/decisions/0027-feature-usage-bitmask-user-agent.md
+++ b/docs/decisions/0027-feature-usage-bitmask-user-agent.md
@@ -216,6 +216,47 @@ a shared registry (I), codegen (K), and the decimal/binary/base-N representation
 - Bad, each feature must add a `mark_feature_used()` call, and first-party clients
   need a per-request hook (small, mirrors existing patterns).
 
+## Prior art
+
+SDK telemetry-in-the-User-Agent is well-established; this design is closest to
+AWS's, and conventional in the rest. Summary of what comparable SDKs do:
+
+| SDK | What's in the UA / headers | Usage-based? | Opt-out | Closest to ours? |
+| --- | --- | --- | --- | --- |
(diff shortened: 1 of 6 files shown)


docs/decisions/0027-feature-usage-bitmask-user-agent.md
```diff
@@ -0,0 +1,395 @@
...
+- Good, fits our **mixed feature lifecycle**: many features are *not* bound to an
+  outbound request — an `Agent`, a workflow/orchestration, or a context/history
+  provider is constructed once and lives for the session/process. A process-wide
+  mask is the only scope that can represent them at all.
+- Good, trivial and cheap: one OR under a lock (Python) / `Interlocked.Or`
+  (.NET); no per-request state plumbing.
+- Neutral, coarser than per-call — early requests carry fewer bits than later
+  ones, and the token says "this process used X", not "this call used X".
+
+#### Q. Per-request set, reset between calls (botocore's model — rejected)
```
TaoChenOSU: nit: why do these two start with P and Q?

docs/decisions/0027-feature-usage-bitmask-user-agent.md
```diff
@@ -0,0 +1,395 @@
...
+
+#### F. Per package, with core broken out per feature/provider (chosen)
+
+- Good, ~50 bits (Python) / ~40 (.NET) fit a **64-bit** mask, which keeps .NET's
+  accumulator lock-free (`Interlocked.Or`) and the registry hand-maintainable.
+- Good, matches the actual questions ("which orchestration / which built-in
+  provider / which package?") — each orchestration pattern and each built-in
+  context/history provider gets its own bit, since they serve different purposes.
+- Neutral, cannot distinguish sub-features *within* a provider package (e.g.
+  openai chat vs embeddings) until a bit is promoted.
```
TaoChenOSU: The endpoints they hit are different, no?

docs/decisions/0033-feature-usage-bitmask-user-agent.md
```diff
@@ -0,0 +1,395 @@
...
+#### K. Code-generate the enums from the registry
+
+- Bad, a generator + drift test + schema test to maintain a short list of
+  integer constants; justified only by the per-construct bit count we rejected.
+
+### Representation (how the mask is rendered as text)
+
+All examples below encode the same mask — bits 0, 2, 16, 22, 27 set
+(agent + workflow + sequential-orchestration + foundry.chat_client + openai, in
+the Python v1 list) = decimal `138477573`.
```
TaoChenOSU: Have we gone over how the registry is versioned?

docs/decisions/0027-feature-usage-bitmask-user-agent.md
```diff
@@ -0,0 +1,395 @@
...
+   Even scoped, it transits intermediaries and is a deployment fingerprint. This
+   is a **release precondition**. Possible outcomes that would further change the
+   design: coarser granularity, hashing, or explicit opt-in (a dedicated mask-only
+   opt-out flag is already included — see below).
+2. **When (if ever) to add the OTel path?** Held back mainly for **privacy**: a
+   span attribute broadcasts the fingerprint into the user's general telemetry
+   and onward to third-party APM vendors, contradicting the first-party-only
+   stance. It also carries a metric-cardinality hazard. Would the privacy review
+   allow a broadly-emitted mask, a scoped/redacted variant, or none? Decide if/when
+   to revisit.
```
TaoChenOSU: OTel requires much more set up on the user side. I am not sure how we can make this option work, given that the data may not even be available to us even if the customer is using Foundry.

docs/decisions/0027-feature-usage-bitmask-user-agent.md
eavanvalkenburg: yeah, so the question for us is, if we need that level of detail or do we care about openai vs gemini usage, rather then openai completions vs openai responses, this is one of the things we need to figure out when implementing.

docs/decisions/0033-feature-usage-bitmask-user-agent.md
eavanvalkenburg: there is a section on that topic, the basics are to only up the version when we need to remove keys, adding can be done inplace (within the limits of the number of bits)

docs/decisions/0027-feature-usage-bitmask-user-agent.md
eavanvalkenburg: agreed, I also doubt if this can ever work, left this in for completeness.

docs/specs/feature-usage-bit-registry.md
```diff
@@ -0,0 +1,233 @@
...
+  would broadcast the fingerprint into the user's general telemetry / third-party
+  APM vendors). Left open behind the version prefix; see
+  [ADR-0033](../decisions/0033-feature-usage-bitmask-user-agent.md#considered-options).
+
+## Bit table — Python (`agent-framework-python`, version 1)
+
+Layout: core feature + provider bits 0–15 (contiguous, with room to grow),
+orchestration patterns 16–21, provider/integration packages from 22.
+
+| Bit | Id | Feature | Marked at (representative) |
```
chetantoshniwal: Lets add Skills (might even be good to track different skill types) and foundry evals.

docs/specs/feature-usage-bit-registry.md
```diff
@@ -0,0 +1,233 @@
...
+
+  ```text
+  agent-framework-python/1.2.3 (feat=v1.23)
+  ```
+
+- **Decoding:** read the **language** from the product token, pick that table;
+  read `vN`, pick that version; `AND` the hex mask against each bit. Unknown bits
+  (newer SDK than the decoder's copy) are ignored.
+
+## Emission scope (where the mask is sent)
```
chetantoshniwal, line 91: Let's add a tenet to help and guide on what should be added and what should not be for further additions.

docs/specs/feature-usage-bit-registry.md
eavanvalkenburg: Addressed in e6027b1da. Foundry evals remain explicit bits in both registries (`foundry.evals`: Python 52, .NET 51). Skills now retain the provider bit and add stable user-facing categories: Python file/in-memory/MCP sources, and .NET file/in-memory/inline/class/MCP skills. Implementation decorators such as caching/filtering/deduplication stay under the parent capability.

docs/specs/feature-usage-bit-registry.md
eavanvalkenburg: Addressed in e6027b1da. The registry now has an **Allocation tenet**: a bit must represent a stable framework-owned capability, answer a concrete product/support question, have an actual-use mark point, and pass privacy review. It explicitly excludes install/import state, aliases, wrappers, internal helpers, decorators, customer/runtime values, and speculative distinctions without a named query/decision owner.

docs/specs/feature-usage-bit-registry.md
eavanvalkenburg: Final rebased commit: 2ce288423 (same changes described above).

docs/specs/feature-usage-bit-registry.md
eavanvalkenburg: Final rebased commit: 2ce288423 (same changes described above).

docs/specs/004-feature-usage-telemetry.md
```diff
@@ -0,0 +1,438 @@
...
+
+Because the existing base-UA values are static, eligible first-party clients
+install a **per-request hook** that calls `apply_feature_token()` on each
+outgoing request:
+
+- **OpenAI-SDK clients created by Agent Framework**: construct the underlying
+  client with
+  `http_client=httpx.AsyncClient(event_hooks={"request": [_stamp_feat_hook]})`,
+  where the hook mutates the fully merged `request.headers["User-Agent"]`.
+  Gate on the existing `use_azure` result in
```
moonbox3: Could gating on `use_azure` leak this process-wide fingerprint to caller-selected endpoints? That result becomes true for an explicit Azure credential or endpoint, while the same settings path accepts an arbitrary `base_url` / `AZURE_OPENAI_BASE_URL`, so a framework-created client routed through a customer gateway or non-Microsoft OpenAI-compatible endpoint would still receive `(feat=...)`. 

Could emission require a separate telemetry-approved destination classification and suppress the token for custom origins and cross-origin redirects?

docs/specs/004-feature-usage-telemetry.md
```diff
@@ -0,0 +1,438 @@
...
+    token = get_feature_token()  # None when disabled or mask == 0
+    base = _strip_feature_comment(user_agent)
+    return f"{base} (feat={token})" if token else base
+```
+
+Because the existing base-UA values are static, eligible first-party clients
+install a **per-request hook** that calls `apply_feature_token()` on each
+outgoing request:
+
+- **OpenAI-SDK clients created by Agent Framework**: construct the underlying
```
moonbox3, line 270: Should we preserve the OpenAI SDK's transport defaults when installing this hook? Passing a plain `httpx.AsyncClient` changes `follow_redirects` from `true` to `false` and replaces the SDK's connection-limit and timeout defaults, so enabling telemetry could alter request behavior and throughput. Could this use `DefaultAsyncHttpxClient(event_hooks=...)` or otherwise preserve the equivalent defaults?

docs/specs/004-feature-usage-telemetry.md
```diff
@@ -0,0 +1,438 @@
...
+
+
+def get_feature_token() -> str | None:
+    """Return ``v<version>.<hex_mask>`` for the accumulated mask, or None."""
+    if not _feature_mask_enabled() or _feature_mask == 0:
+        return None
+    return f"v{REGISTRY_VERSION}.{_feature_mask:x}"
+```
+
+- **Per package/feature, usage-based:** `mark_feature_used()` is called the first
```
moonbox3: Should operational clients and providers mark usage on their first public operation rather than construction? DI containers commonly instantiate services that are never called, and constructor marking would permanently add those unused features to the process-global mask even though the stated signal is "actually exercised." Could construction remain the mark point only where construction itself exercises the capability?

docs/specs/004-feature-usage-telemetry.md
```diff
@@ -0,0 +1,438 @@
...
+- **Token is safe by construction.** The emitted value is `v{int}.{hex}` —
+  characters limited to `[0-9a-fv.]` — so no header-injection sanitization is
+  required. A 128-bit mask is at most 32 hex characters (contrast botocore,
+  which must sanitize and cap arbitrary component strings).
+- **Private API.** `mark_feature_used`, `get_feature_token`, `apply_feature_token`
+  and the mask itself are internal helpers; only the emitted token and the
+  per-language registry tables are the stable, decodable contract.
+- **No import cycles:** the call lives in each package's own module, so `core`
+  never imports optional packages. Each package references its bit via the shared
+  `FeatureBit` IntEnum in `agent_framework._telemetry`.
```
moonbox3: Could we call this `FeatureIndex` or `FeatureBitIndex`? The enum values are positions 0..127, not bit values, so FeatureBit invites callers to OR the ordinal directly instead of shifting it. It looks like that ambiguity has already produced an incorrect `Interlocked.Or(..., (long)bit)` implementation in an earlier revision of this doc.

docs/specs/004-feature-usage-telemetry.md
```diff
@@ -0,0 +1,438 @@
...
+
+### Bit constants
+
+`core` defines a hand-written `FeatureBit` IntEnum in `_telemetry.py` alongside
```
moonbox3: Have we considered the release coupling created by keeping every optional-package bit in core? Adding a telemetry bit for any independently versioned provider would require a core release and then a provider dependency-floor update before that provider can ship it. Could package-local index constants, validated against the published registry, preserve the dependency direction without requiring coordinated releases?

docs/specs/004-feature-usage-telemetry.md
eavanvalkenburg: Addressed in 37b7732e3. Emission now requires both an approved framework pipeline and the actual normalized HTTPS origin to match that pipeline’s reviewed first-party allowlist. Credentials/`use_azure` alone do not approve a destination; custom gateways are default-deny, and the hook re-evaluates every redirect hop and removes `(feat=...)` before an unapproved/cross-origin request.

docs/specs/004-feature-usage-telemetry.md
eavanvalkenburg: Addressed in 37b7732e3. The OpenAI design now requires `DefaultAsyncHttpxClient(event_hooks=...)`, not a plain `httpx.AsyncClient`, so the SDK’s redirect, timeout, connection-limit, and pool defaults remain intact. Caller-owned transports are not replaced.

docs/specs/004-feature-usage-telemetry.md
eavanvalkenburg: Addressed in 37b7732e3. Marking now uses first meaningful activation. Operational clients, tools, providers, and hosts mark on their first real operation/participation; constructor marking is allowed only when construction itself activates or registers the capability. DI instantiation alone is explicitly not usage.

docs/specs/004-feature-usage-telemetry.md
eavanvalkenburg: Addressed in 37b7732e3. `FeatureBit` was renamed to `FeatureIndex` throughout. The declarations contain positions `0..127`, are not flags, and `mark_feature_used(index)` performs the shift.

docs/specs/004-feature-usage-telemetry.md
eavanvalkenburg: Addressed in 37b7732e3. Optional packages now own private package-local `FeatureIndex` declarations; core owns only the accumulator API and core indexes. A repository validation test is required to assert exact registry parity, complete non-reserved coverage, `0..127` range, and no duplicate/overlapping indexes across packages for each language/version.

docs/specs/feature-usage-bit-registry.md
```diff
@@ -0,0 +1,264 @@
...
+| 66 | `azure_cosmos` | Azure Cosmos history / checkpoint store | `agent_framework_azure_cosmos` |
+| 67 | `azure_contentunderstanding` | Azure Content Understanding context provider | `agent_framework_azure_contentunderstanding.ContentUnderstandingContextProvider` |
+| 68 | `redis` | Redis context / history provider | `agent_framework_redis` |
+| 69 | `mem0` | Mem0 memory provider | `agent_framework_mem0.Mem0ContextProvider` |
+| 70 | `purview` | Purview client | `agent_framework_purview.PurviewClient` |
+| 71 | `a2a` | A2A agent / executor | `agent_framework_a2a.A2AAgent` / `A2AExecutor` |
+| 72 | `ag_ui` | AG-UI chat client / agent | `agent_framework_ag_ui` |
+| 73 | `chatkit` | ChatKit integration | `agent_framework_chatkit` |
+| 74 | `devui` | DevUI served | `agent_framework_devui.serve` |
+| 75 | `declarative` | Declarative agent / workflow | `agent_framework_declarative` |
```
peibekwe: As discussed, we should separate out declarative agents and declarative workflows. The usage and investment needs for both are different.

docs/specs/feature-usage-bit-registry.md
```diff
@@ -0,0 +1,264 @@
...
+| 56 | `copilotstudio` | Copilot Studio agent | `Microsoft.Agents.AI.CopilotStudio.CopilotStudioAgent` |
+| 57 | `github_copilot` | GitHub Copilot agent | `Microsoft.Agents.AI.GitHub.Copilot.GitHubCopilotAgent` |
+| 58 | `azure_cosmos` | Cosmos history / checkpoint store | `Microsoft.Agents.AI.CosmosChatHistoryProvider` |
+| 59 | `valkey` | Valkey chat-history provider | `Microsoft.Agents.AI.Valkey.ValkeyChatHistoryProvider` |
+| 60 | `mem0` | Mem0 memory provider | `Microsoft.Agents.AI.Mem0.Mem0Provider` |
+| 61 | `purview` | Purview integration | `Microsoft.Agents.AI.Purview` |
+| 62 | `a2a` | A2A agent | `Microsoft.Agents.AI.A2A.A2AAgent` |
+| 63 | `hosting.ag_ui` | AG-UI hosting endpoint | `Microsoft.Agents.AI.Hosting.AGUI.AspNetCore.AGUIEndpointRouteBuilderExtensions.MapAGUIServer` |
+| 64 | `devui` | DevUI served | `Microsoft.Agents.AI.DevUI` |
+| 65 | `declarative` | Declarative agent factory | `Microsoft.Agents.AI.ChatClientPromptAgentFactory` |
```
peibekwe: The same point for dotnet as well. We need one for declarative agents and another for declarative workflow.

TaoChenOSU (approved):

docs/specs/feature-usage-bit-registry.md
eavanvalkenburg: Addressed in a6e574f87. Python now allocates adjacent indexes `75` (`declarative.agent`) and `76` (`declarative.workflow`). Because v1 has not shipped, every later candidate row was shifted by one rather than consuming a distant reserved slot.

docs/specs/feature-usage-bit-registry.md
eavanvalkenburg: Addressed in a6e574f87. .NET now allocates adjacent indexes `65` (`declarative.agent`) and `66` (`declarative.workflow`), with later unreleased v1 rows shifted by one.

docs/decisions/0033-feature-usage-bitmask-user-agent.md
```diff
@@ -0,0 +1,589 @@
...
+- Good, compact (≤32 chars for a 128-bit mask).
+- Good, decodes with one stdlib call in every language (`int(x, 16)` /
+  two 64-bit lane parses in .NET); nibble boundaries are eyeball-able.
+- Good, lowercase, no `0x` prefix, no leading zeros — unambiguous and stable.
+
+#### N. Binary / bit-list — `feat=v1.100000001000000000000000100000000000000000000000000000101` or `feat=v1.0,2,32,48,56`
+
+- Good, most directly human-readable ("which bits").
+- Bad, longest form in the UA; the bit-list needs delimiter handling and grows
+  with the number of set bits.
```
rogerbarreto: Bit list should be separated from the binary option.

I `feat=v1.0,1,2,3,4,5` is clearly eyeballing transparent while, this option should also come with strong recommendations on how many bits should be present in a single call at the same time. Do we have an AVG estimate, are you thinking in over using the bits?

What is extra the processing cost of adding the bits detection for building up the UA?

Would be nice also to provide what is a char count in a 128 bits scenario for every option so we have a table of how many bytes are expected to be added per request

Are we thinking in providing up to 128 bits of information in our calls?

docs/decisions/0033-feature-usage-bitmask-user-agent.md
```diff
@@ -0,0 +1,589 @@
...
+
+- Good, human-familiar; trivial to parse.
+- Neutral, no visual alignment to bit/nibble boundaries; slightly longer than hex
+  for large masks. No advantage over hex.
+
+#### M. Hex (chosen) — `feat=v1.101000100000005`
+
+- Good, compact (≤32 chars for a 128-bit mask).
+- Good, decodes with one stdlib call in every language (`int(x, 16)` /
+  two 64-bit lane parses in .NET); nibble boundaries are eyeball-able.
```
rogerbarreto: Suggest removing any jargons from the ADR. "nibble boundaries are eyeball-able" ???

Also, suggest another option where the "nibble" boundaries are separated by a "." separator char, for ease of looking, possibly something that could be compatible with a `int(x, 16)` call with the benefit of extra visibility.

```suggestion
  two 64-bit lane parses in .NET); and because each character always stands for the same group of four bits, you can read off which features are set just by looking.
```

docs/decisions/0033-feature-usage-bitmask-user-agent.md
```diff
@@ -0,0 +1,589 @@
...
+`mark_feature_used()` calls (so it spans build/start/participation activations
+that aren't bound to any service request — the per-request set model (S2) can't);
+the token is **stamped per request** only when both the client/pipeline and the
+actual HTTPS origin are approved, so custom origins and cross-origin redirects
+cannot inherit the fingerprint; each
+SDK owns an independent bit list selected by the language already in the UA; the
+mask is rendered as hex (`feat=v1.101000100000005`). **Two opt-out env vars are
+provided:** a dedicated `AGENT_FRAMEWORK_FEATURE_MASK_DISABLED` that drops only
+the mask while keeping the base SDK identity/version User-Agent, and the existing
+`AGENT_FRAMEWORK_USER_AGENT_DISABLED` that drops the whole contribution. OTel (C)
```
rogerbarreto: I'm not sure we do have proposed or existing anything that drops the MAF - UA stamping, not sure if this should be added as part of this ADR, suggest limiting the env var toggling for the Feature Bit Stamping, in the scope of this PR.

```suggestion
the mask while keeping the base SDK identity/version User-Agent
```

docs/decisions/0033-feature-usage-bitmask-user-agent.md
```diff
@@ -0,0 +1,589 @@
...
+normal growth; it does not waive the registry's
+[allocation tenet](../specs/feature-usage-bit-registry.md#allocation-tenet).
+
+### Consequences
+
+- Good, adds usage signal at near-zero cost, no new data flow, few moving parts.
+- Good, transparent (public registry, human-decodable token) and disabled by
+  **two** opt-out env vars: a dedicated `AGENT_FRAMEWORK_FEATURE_MASK_DISABLED`
+  (mask only) and `AGENT_FRAMEWORK_USER_AGENT_DISABLED` (whole UA; existing in
+  Python and added to .NET with this work).
```
rogerbarreto: See above.

```suggestion
- Good, transparent (public registry, human-decodable token) and disabled by
  **one** opt-out env vars: a dedicated `AGENT_FRAMEWORK_FEATURE_MASK_DISABLED`
  (mask only).
```

docs/specs/004-feature-usage-telemetry.md
```diff
@@ -0,0 +1,466 @@
...
+    with _feature_mask_lock:
+        _feature_mask |= 1 << index
+
+
+def get_feature_token() -> str | None:
+    """Return ``v<version>.<hex_mask>`` for the accumulated mask, or None."""
+    if not _feature_mask_enabled() or _feature_mask == 0:
+        return None
+    return f"v{REGISTRY_VERSION}.{_feature_mask:x}"
+```
```
rogerbarreto, line 122: Are feature bit's a global setting? Isn't it per request and by usage isolation?

Not sure this will represent the actual journey of the feature usage. 

i.e:
```
Agent 2 will send UA bits only with what was used
Time1: Agent1 -> Function Call  (bit = Agent + FC + Chat)
Time2: Agent2 -> Chat Normal    (bit = `Agent + Chat)

---
Agent 2 usage will send UA with FC while it didn't use FC.

Time1: Agent1 -> Function Call (bit = Agent + FC + Chat)
Time2: Agent2 -> Chat (bit = Agent + FC + Chat).
```

Would be nice in the ADR to show what is a real use case and what to expect from the bits , for clarity.

docs/specs/004-feature-usage-telemetry.md
```diff
@@ -0,0 +1,466 @@
...
+regardless of provider. Only **emission** is scoped. A user who never calls a
+first-party endpoint emits no token; this is the honest, intended behaviour (no
+third-party leakage, no signal we couldn't read anyway).
+
+The existing base User-Agent behavior (`agent-framework-python/{version}` plus
+any dynamically detected hosting prefix) is unchanged; packages continue using
+their current `default_headers`, `user_agent`, suffix, or policy mechanisms.
+`get_user_agent()` stays base-only (no `feat=`). The `feat=` token is
+**separate**, added **only** by eligible Azure/Foundry clients, and
+**re-evaluated on each request** so it reflects the mask accumulated so far. A
```
rogerbarreto, line 233: Correct me if I'm wrong

Not sure about the mask accumulation being a global setting would be meaningful for the tracking. 

If a feature was used only once for the lifecycle of the application promoting it in everysubsequent call may give us the wrong indication that the feature is actually being used in the given request.

Accumulating strategy also penalized the request and bits, as they will grow the longer the application is running and cannot be trusted by its sums or counts, we only can know that the feature was used (not how often, or for the given request).

docs/specs/feature-usage-bit-registry.md
```diff
@@ -0,0 +1,266 @@
...
+| 53 | `foundry.toolbox` | Foundry Toolbox MCP tool | `agent_framework_foundry_hosting.FoundryToolbox` |
+| 54 | `foundry_local` | Foundry Local client | `agent_framework_foundry_local.FoundryLocalClient` |
+| 55 | `foundry_hosting` | Foundry hosting layer | `agent_framework_foundry_hosting.ResponsesHostServer` / `InvocationsHostServer` |
+| 56 | `openai` | OpenAI clients | `agent_framework_openai` |
+| 57 | `anthropic` | Anthropic clients | `agent_framework_anthropic` |
+| 58 | `bedrock` | AWS Bedrock clients | `agent_framework_bedrock` |
+| 59 | `gemini` | Gemini chat client | `agent_framework_gemini` |
+| 60 | `mistral` | Mistral embedding client | `agent_framework_mistral` |
+| 61 | `ollama` | Ollama clients | `agent_framework_ollama` |
+| 62 | `claude` | Claude Agent SDK agent | `agent_framework_claude` |
```
rogerbarreto, line 159: Suggest moving all third party indexes to the end of the table, possibly a reserved 3P slot 80-128, something that fits, or a dedicated bit just for those, 3P bit table.

docs/specs/feature-usage-bit-registry.md
```diff
@@ -0,0 +1,266 @@
...
+| 74 | `devui` | DevUI served | `agent_framework_devui.serve` |
+| 75 | `declarative.agent` | Declarative agent definitions | `agent_framework_declarative.AgentFactory` |
+| 76 | `declarative.workflow` | Declarative workflow definitions | `agent_framework_declarative.WorkflowFactory` |
+| 77 | `durabletask` | Durable task runtime | `agent_framework_durabletask` |
+| 78 | `azurefunctions` | Azure Functions agent host | `agent_framework_azurefunctions` |
+| 79 | `tools` | Shell tools | `agent_framework_tools.shell.LocalShellTool` / `DockerShellTool` |
+| 80 | `monty` | Monty CodeAct provider | `agent_framework_monty.MontyCodeActProvider` |
+| 81 | `hyperlight` | Hyperlight CodeAct provider | `agent_framework_hyperlight.HyperlightCodeActProvider` |
+| 82 | `azure_cosmos_memory` | Azure Cosmos DB semantic-memory provider | `agent_framework_azure_cosmos_memory.CosmosMemoryContextProvider` |
+| 83 | `hosting` | App-owned agent/workflow hosting state | `agent_framework_hosting.AgentState` / `WorkflowState` |
```
rogerbarreto, line 180: Suggest using more specific names, as `hosting` + `a2a` could mean completely different from `hosting.a2a`. 

The provided Id's should be specific which low-to-no margin for misrepresentation. Suggest looking for other cases where this could happen

```suggestion
| 83 | `hosting.core` | App-owned agent/workflow hosting state | `agent_framework_hosting.AgentState` / `WorkflowState` |
```

docs/specs/feature-usage-bit-registry.md
```diff
@@ -0,0 +1,266 @@
...
+| 60 | `mem0` | Mem0 memory provider | `Microsoft.Agents.AI.Mem0.Mem0Provider` |
+| 61 | `purview` | Purview integration | `Microsoft.Agents.AI.Purview` |
+| 62 | `a2a` | A2A agent | `Microsoft.Agents.AI.A2A.A2AAgent` |
+| 63 | `hosting.ag_ui` | AG-UI hosting endpoint | `Microsoft.Agents.AI.Hosting.AGUI.AspNetCore.AGUIEndpointRouteBuilderExtensions.MapAGUIServer` |
+| 64 | `devui` | DevUI served | `Microsoft.Agents.AI.DevUI` |
+| 65 | `declarative.agent` | Declarative agent definitions | `Microsoft.Agents.AI.PromptAgentFactory.CreateAsync` |
+| 66 | `declarative.workflow` | Declarative workflow definitions | `Microsoft.Agents.AI.Workflows.Declarative.DeclarativeWorkflowBuilder.Build` |
+| 67 | `durabletask` | Durable task runtime | `Microsoft.Agents.AI.DurableTask` |
+| 68 | `azurefunctions` | Azure Functions agent host | `Microsoft.Agents.AI.Hosting.AzureFunctions` |
+| 69 | `tools` | Shell tools | `Microsoft.Agents.AI.Tools.Shell.ShellExecutor` |
```
rogerbarreto: Tools is a very broad identifier, suggest getting more specific naming for the enumerators.
```suggestion
| 69 | `tools.shell` | Shell tools | `Microsoft.Agents.AI.Tools.Shell.ShellExecutor` |
```

docs/specs/feature-usage-bit-registry.md
eavanvalkenburg: Addressed in 982a92cd7 with an explicit allocation rule rather than reindexing. The provider/integration range is not partitioned by vendor ownership because several packages span first- and third-party services, ownership can change, and protocols/storage do not fit a stable taxonomy. Indexes are allocation space; the destination allowlist independently controls first-party-only emission.

docs/decisions/0033-feature-usage-bitmask-user-agent.md
eavanvalkenburg: Addressed in 982a92cd7. This ADR now introduces only the dedicated mask-only flag across both SDKs. The existing Python whole-User-Agent flag still suppresses the Python mask, while adding an equivalent .NET whole-UA switch is explicitly outside this decision.

docs/specs/feature-usage-bit-registry.md
eavanvalkenburg: Addressed in 982a92cd7. The registry now defines its naming rule. Python keeps `hosting` because it is the actual base `agent-framework-hosting` package, distinct from the separate hosting integration packages. The narrower .NET `AIHostAgent` row is renamed to `hosting.agent`.

docs/decisions/0033-feature-usage-bitmask-user-agent.md
eavanvalkenburg: Addressed in 982a92cd7. The jargon was removed and the text now says directly that each hex character maps to four consecutive bit positions. A dotted grouped-hex form is documented as considered but rejected because separators add bytes and must be stripped before standard hex parsing.

docs/specs/004-feature-usage-telemetry.md
eavanvalkenburg: Addressed in 982a92cd7. The process-global accumulator remains intentional. The ADR/spec now explain that per-operation scoping would add hot-path enable/reset/context propagation and expose a more privacy-sensitive behavioral trace. Bits are defined as sticky binary observations and raw request rows must never be interpreted as invocation frequency.

docs/specs/004-feature-usage-telemetry.md
eavanvalkenburg: Addressed in 982a92cd7. The spec now includes an Agent A/Agent B timeline showing intentional carry-over: a later request may contain an MCP bit set by earlier work in the same process. It explicitly says this is process-level observed-once state, not evidence that the later agent or request used MCP and not a usage count.

docs/decisions/0033-feature-usage-bitmask-user-agent.md
eavanvalkenburg: Addressed in 982a92cd7. The outcome and consequences now describe one new `AGENT_FRAMEWORK_FEATURE_MASK_DISABLED` opt-out. The pre-existing Python whole-UA behavior is noted separately rather than standardized as new cross-language behavior.

docs/decisions/0033-feature-usage-bitmask-user-agent.md
eavanvalkenburg: Addressed in 982a92cd7. Binary and bit-list are now separate alternatives. The ADR adds exact added-User-Agent byte counts for the example, all currently assigned Python/.NET rows, and the full 128-bit bound; it also states that there is no defensible average or smaller per-request bit budget and describes the bounded marking/stamping work.

docs/specs/feature-usage-bit-registry.md
eavanvalkenburg: Addressed in 982a92cd7. The shell-tool identifier is now `tools.shell` in both language tables, and the personal KQL decoder snapshot was updated to match.

giles17 (approved):

PR #6515, opened 2026-06-15
Python: Capture context provider instructions in agent telemetry

### LINKED ISSUE #5291: Python: AgentTelemetryLayer records gen_ai.system_instructions before ContextProvider.before_run() extends them
## Description

`AgentTelemetryLayer._trace_agent_invocation()` captures `gen_ai.system_instructions` from `merged_options` **before** calling `execute()`, which is where `AgentMiddlewareLayer` runs `ContextProvider.before_run()` and processes `context.extend_instructions()` calls.

This means the telemetry span only records the **base** agent instructions, not the dynamically extended instructions from context providers.

## Reproduction

```python
class UserMemoryProvider(ContextProvider):
    async def before_run(self, *, agent, session, context, state, **kwargs):
        context.extend_instructions(self.source_id, "The user's name is Alice.")

agent = Agent(
    client=client,
    name="MemoryAgent",
    instructions="You are a friendly assistant.",
    context_providers=[UserMemoryProvider()],
)
session = agent.create_session()
result = await agent.run("Hello", session=session)
```

## Expected

`gen_ai.system_instructions` in the `invoke_agent` span contains:
```json
[{"type": "text", "content": "You are a friendly assistant."}, {"type": "text", "content": "The user's name is Alice."}]
```

## Actual

`gen_ai.system_instructions` only contains the base instructions:
```json
[{"type": "text", "content": "You are a friendly assistant."}]
```

## Root Cause

In `observability.py`, `AgentTelemetryLayer._trace_agent_invocation()` — both the streaming and non-streaming paths capture `system_instructions` before `execute()` is called:

```python
# Non-streaming path (~line 1636):
with _get_span(...) as span:
    _capture_messages(
        span=span,
        messages=messages,
        system_instructions=_get_instructions_from_options(dict(merged_options)),  # base only
    )
    response = await execute()  # context providers run HERE, too late
```

The call order is:
```
AgentTelemetryLayer:
  1. Create span
  2. Record gen_ai.system_instructions  <-- only base instructions
  3. Call execute() -->
     AgentMiddlewareLayer:
       4. context_provider.before_run()  <-- extends instructions HERE
       5. Call LLM
       6. context_provider.after_run()
  7. End span
```

## Suggested Fix

Move the `_capture_messages()` call to **after** `execute()` returns (or at minimum, re-capture `system_instructions` post-execute). Alternatively, the middleware layer could update the span attribute after providers run.

## Current Workaround

Add a `ContextProvider` as the **last** provider that patches the span manually:

```python
from opentelemetry import trace

class InstructionsTelemetryFixProvider(ContextProvider):
    def __init__(self):
        super().__init__("telemetry_fix")

    async def before_run(self, *, agent, session, context, state, **kwargs):
        span = trace.get_current_span()
        if span.is_recording():
            all_instructions = [agent.instructions] if isinstance(agent.instructions, str) else list(agent.instructions)
            if context.instructions:
                all_instructions.extend(context.instructions)
            otel = [{"type": "text", "content": i} for i in all_instructions]
            span.set_attribute("gen_ai.system_instructions", json.dumps(otel))

agent = Agent(
    ...,
    context_providers=[UserMemoryProvider(), InstructionsTelemetryFixProvider()],  # must be LAST
)
```

## Environment

- `agent-framework`: 1.0.1
- `opentelemetry-sdk`: 1.40
- Python: 3.10
- OS: Windows

eavanvalkenburg:
### Motivation & Context

`AgentTelemetryLayer` captured `gen_ai.system_instructions` before context providers ran, so the `invoke_agent` span only reported base instructions and missed instructions added through `ContextProvider.before_run()`.

Fixes #5291

  - Track the active `invoke_agent` span while agent telemetry is running.
  - Let agent-owned chat telemetry update that span with the final system instructions once context providers have contributed instructions.
  - Add regression coverage for non-streaming, streaming, and unrelated nested chat calls.
  - Agent telemetry now reports the same final system instructions sent to the chat client, including context-provider additions.
  - Unrelated nested chat telemetry is guarded from overwriting the agent span instructions.
  - The active agent span guard in `observability.py`, especially the parent/session checks that limit updates to agent-owned chat spans.

### Related Issue

Fixes #5291

### Contribution Checklist

### DIFF
--- a/python/packages/core/agent_framework/observability.py
+++ b/python/packages/core/agent_framework/observability.py
@@ -2344,7 +2344,7 @@ def _instructions_preserve_existing_agent_instructions(
 
     if not isinstance(existing_items_obj, list):
         return False
-    existing_items = cast(list[Any], existing_items_obj)
+    existing_items = cast(list[object], existing_items_obj)
 
     existing_contents: list[str] = []
     for item in existing_items:
--- a/python/packages/core/agent_framework/observability.py
+++ b/python/packages/core/agent_framework/observability.py
@@ -112,10 +112,6 @@ INNER_ACCUMULATED_USAGE: Final[contextvars.ContextVar[UsageDetails | None]] = co
     "inner_accumulated_usage", default=None
 )
 
-ACTIVE_AGENT_SPAN: Final[contextvars.ContextVar[trace.Span | None]] = contextvars.ContextVar(
-    "active_agent_span", default=None
-)
-
 OTEL_METRICS: Final[str] = "__otel_metrics__"
 TOKEN_USAGE_BUCKET_BOUNDARIES: Final[tuple[float, ...]] = (
     1,
@@ -1492,14 +1488,15 @@ class ChatTelemetryLayer(Generic[OptionsCoT]):
         )
 
         if stream:
+            agent_span = trace.get_current_span()
             span = _start_streaming_span(attributes, OtelAttr.REQUEST_MODEL)
 
             if OBSERVABILITY_SETTINGS.SENSITIVE_DATA_ENABLED and messages and span.is_recording():
                 system_instructions = _get_instructions_from_options(opts)
-                _capture_active_agent_system_instructions(
+                _capture_current_agent_system_instructions(
+                    agent_span,
                     span,
                     system_instructions,
-                    is_agent_chat_call=merged_client_kwargs.get("session") is not None,
                 )
                 _capture_messages(
                     span=span,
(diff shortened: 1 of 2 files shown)


python/packages/core/agent_framework/observability.py
```diff
@@ -112,6 +112,10 @@
     "inner_accumulated_usage", default=None
 )
 
+ACTIVE_AGENT_SPAN: Final[contextvars.ContextVar[trace.Span | None]] = contextvars.ContextVar(
```
TaoChenOSU: We don't need to track this ourselves. We can use `trace.get_current_span`

eavanvalkenburg:
Updated based on the review feedback. The session-based ownership heuristic was too broad, so I removed both the explicit active-span tracking and the attempted provenance marker. The updated implementation now uses `trace.get_current_span()` and only updates the current parent `invoke_agent` span when the chat instructions preserve/extend the instructions already captured on that agent span. I also strengthened the nested-chat regression to pass the live session from `after_run`, covering the overwrite case called out by the automated review.

python/packages/core/agent_framework/observability.py
eavanvalkenburg: Addressed in the latest commits. I removed the self-tracked active span/provenance machinery and now use `trace.get_current_span()` from chat telemetry. The remaining guard only updates the parent `invoke_agent` span when the chat instructions preserve/extend the instructions already captured on that agent span, so later nested chat calls cannot clobber it.

TaoChenOSU (approved):

moonbox3 (approved):

PR #6523, opened 2026-06-15
.NET: Rebuild Hyperlight sandbox after tool registry updates

eavanvalkenburg:
### Motivation & Context

Hyperlight CodeAct providers should keep the reusable sandbox aligned with the provider-owned tool registry. When a tool registration is updated, subsequent run snapshots need to ensure sandbox callbacks are refreshed instead of relying only on stable tool names in the reuse key.

This change makes tool registry updates participate in sandbox reuse decisions while preserving the existing per-run snapshot behavior.

  - Add a tool registry version to Hyperlight run snapshots and include it in the sandbox configuration fingerprint.
  - Increment the provider registry version when `AddTools(...)` accepts tool registrations, including same-name replacements.
  - Keep the standalone `HyperlightExecuteCodeFunction` on a fixed version because it captures immutable options at construction time.
  - Add unit coverage proving same-name tool snapshots with different registry versions produce different fingerprints.
  - Subsequent provider snapshots after `AddTools(...)` force a lazy sandbox rebuild before execution, so registered callbacks match the current provider registry.
  - Existing snapshots remain isolated from later provider mutations.
  - No public API or breaking behavior change.
  - Whether the registry version is captured at the right boundary and whether the lazy rebuild behavior preserves the intended snapshot contract.

### Related Issue

No public issue. No matching open PR found.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.Hyperlight/Internal/SandboxExecutor.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Hyperlight/Internal/SandboxExecutor.cs
@@ -46,7 +46,7 @@ internal sealed class SandboxExecutor : IDisposable
             IReadOnlyList<FileMount> fileMounts,
             IReadOnlyList<AllowedDomain> allowedDomains,
             string? hostInputDirectory,
-            long toolRegistryVersion = 0)
+            Guid toolRegistryVersion = default)
         {
             this.Tools = tools;
             this.FileMounts = fileMounts;
@@ -69,7 +69,7 @@ internal sealed class SandboxExecutor : IDisposable
 
         public string? HostInputDirectory { get; }
 
-        public long ToolRegistryVersion { get; }
+        public Guid ToolRegistryVersion { get; }
 
         /// <summary>
         /// Stable fingerprint of the configuration that materially affects how
@@ -84,7 +84,7 @@ internal sealed class SandboxExecutor : IDisposable
             IReadOnlyList<FileMount> fileMounts,
             IReadOnlyList<AllowedDomain> allowedDomains,
             string? hostInputDirectory,
-            long toolRegistryVersion = 0)
+            Guid toolRegistryVersion = default)
         {
             var sb = new StringBuilder();
             sb.Append("tools=");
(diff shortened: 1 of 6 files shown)


SergeyMenshykh (commented):
## Review

Correct, surgical fix for a real sandbox-reuse bug: `ComputeFingerprint` keys tools by **name only**, so replacing a tool with a same-name/different-callback registration kept the stale warm sandbox. Folding in the thread-safe `_toolRegistryVersion` correctly forces the rebuild. `toolRegistryVersion: 0` for the immutable standalone function is right.

**One ask — strengthen the test.** The new test only proves `ComputeFingerprint` differs when handed `1` vs `2`, which is nearly tautological. It never verifies the actual motivation: that **`AddTools(...)` with a same-name replacement** bumps the version and changes the snapshot fingerprint. Please add a test that calls `AddTools` twice (same name, different callback) and asserts the two resulting snapshots/fingerprints differ — closing the loop `AddTools -> version bump -> fingerprint change -> rebuild`.

Otherwise LGTM. Worth confirming CI is green (build/test boxes unchecked).

SergeyMenshykh (approved):

eavanvalkenburg:
@SergeyMenshykh addressed in 07167dec4: added provider-level coverage that exercises `AddTools(...)` with a same-name replacement and asserts the captured `execute_code` snapshot fingerprint changes. I used a small internal test-visible fingerprint property to avoid reflection.

dotnet/src/Microsoft.Agents.AI.Hyperlight/Internal/SandboxExecutor.cs
```diff
@@ -61,6 +69,8 @@ public RunSnapshot(
 
         public string? HostInputDirectory { get; }
 
+        public long ToolRegistryVersion { get; }
```
westey-m: There is a small risk that if this was running over a very long period of time, ToolRegisteryVersion might overflow.  Should we consider Guid instead?

westey-m (approved):

dotnet/src/Microsoft.Agents.AI.Hyperlight/Internal/SandboxExecutor.cs
eavanvalkenburg: Addressed in ada57bc64: switched the tool registry generation from `long` to a `Guid`, generate a fresh `Guid.NewGuid()` for accepted `AddTools(...)` mutations, and use `Guid.Empty` for the immutable standalone execute-code snapshot.

westey-m (approved):

westey-m (approved):

Issue #5366, opened 2026-04-20
Add deterministic validation at execution boundary for agent actions


devincapriola:
The framework enables agents to orchestrate workflows and trigger real-world actions (tool calls, API requests, etc).

From the documentation and examples, it appears that developers are expected to implement their own safeguards and validation logic within the application layer.

This creates a potential gap at the moment of execution:

- actions may be constructed across multiple steps or agents  
- parameters can evolve during orchestration  
- execution may occur without a final, deterministic validation of the exact payload being committed  

While upstream validation, policies, and guardrails are important, they do not guarantee that the final executed action is explicitly verified at the execution boundary.

### Suggestion

Introduce a system-level validation step before execution that verifies:

- agent identity / context  
- action being performed  
- parameters (schema + expected values)  
- destination or external system  
- timestamp / validity window  
- optional nonce or replay protection  

If validation fails → execution does not occur.

This could be implemented as an execution wrapper or middleware layer, ensuring validation is consistently enforced across all agents and workflows rather than relying on individual implementations.

Happy to share a reference implementation pattern if useful.

chetantoshniwal:
Can you clarify with your concrete scenario on what exactly you are proposing and how it will be generic.

devincapriola:
### Example scenario

An agent is orchestrating a workflow to trigger a payment API:

1. Agent A determines an action → “send payment”
2. Parameters are built across steps (amount, destination, metadata)
3. Another agent or step modifies or enriches those parameters
4. Final API call is executed

At each step, validation may occur, but the final payload being executed is still assembled at runtime.

### The gap

Even if:
- schema validation passes  
- policies are applied upstream  
- guardrails are in place  

There isn’t necessarily a single deterministic check right before execution that verifies:

- the exact action  
- final parameters  
- destination  
- execution context  

### What I’m proposing

A generic validation layer at the execution boundary, regardless of the specific tool or API.

Conceptually:

Where `validate(...)` can include:
- schema + expected value checks  
- permission/policy checks  
- environment or tenant constraints  
- replay protection or validity window  

### Why this is generic

This pattern applies across:
- API calls (payments, emails, data writes)
- tool execution
- code execution
- multi-agent workflows

The key idea is having a consistent enforcement point right before execution, rather than relying on validation being distributed across earlier steps or individual components.

moonbox3:
[...]

PR #6528, opened 2026-06-15
Python: Samples: deterministic action-boundary validation middleware (#5366)

eeee2345:
This adds a function-middleware sample that answers #5366: a single deterministic enforcement point that validates a tool call right before it executes.

ATRValidationMiddleware subclasses FunctionMiddleware, inspects the validated arguments in FunctionInvocationContext.arguments, and raises MiddlewareTermination before calling call_next() when the arguments match a known attack pattern, so the tool never runs. The decision is deterministic (no model call, no network), which is what #5366 asks for at the execution boundary.

The check is a small, self-contained deny-list illustrating the pattern. To enforce a full, maintained ruleset instead of the illustrative subset, install the open-source engine (pip install pyatr) and swap the matcher; the sample documents this inline.

Files
- python/samples/02-agents/middleware/atr_validation_middleware.py
- a row in the middleware samples README

Local checks
- ruff clean (line-length 120; long patterns split across adjacent string literals)
- runs standalone: prints one allowed call and one blocked call

Disclosure: the deny-list patterns mirror the open-source Agent Threat Rules (ATR) project, which I maintain. The sample has no ATR dependency; the reference is for users who want the full ruleset.


python/samples/02-agents/middleware/atr_validation_middleware.py
```diff
@@ -0,0 +1,155 @@
...
+instead of this illustrative subset, install the engine (``pip install pyatr``) and replace
+``_matches_attack_pattern`` with a call into it; see
+https://github.com/Agent-Threat-Rule/agent-threat-rules.
+
+Because the validation is deterministic and happens at the execution boundary, the decision is
+reproducible and auditable -- no model is in the enforcement path.
+"""
+
+
+# A small, illustrative subset that mirrors ATR rule intent (prompt injection, exfiltration,
```
eavanvalkenburg: would it be possible to pull in the rules laid out in the ATR repo? that would make this even more powerfull! Or would it make sense to wait for this to be completed: https://github.com/Agent-Threat-Rule/agent-threat-rules/tree/main/engines/python ?

eeee2345:
@eavanvalkenburg yes — pulling in the real ATR ruleset is the right move, and it makes the sample much stronger than the illustrative deny-list it has now. The rules ship in the `agent-threat-rules` package (and `pyatr` on the Python side), load locally, and run deterministically with no model call, so the middleware can evaluate tool args against the actual published ruleset instead of a hardcoded subset. I don't think it needs to wait on anything — the package is published and stable, so the sample can load the rules today.

I'll update the sample to load ATR directly, and while I'm in there I'll fix the points from the review: the `.` not matching newlines is a real gap for multiline injection payloads (I'll add the DOTALL flag), move `load_dotenv()` into `main()`, switch `print` to a module logger, and include the matched rule id in the block message for auditability. Thanks for the careful look.

eeee2345:
@eavanvalkenburg done — pushed the fixes. The sample now loads the published ATR ruleset and runs the real pyatr engine over the tool args (with a deny-list fallback only when pyatr isn't installed), plus the review points: DOTALL on the fallback patterns, load_dotenv() moved into main(), a module logger instead of print(), and the matched ATR rule id surfaced in both the log line and the MiddlewareTermination message. The in-repo engines/python path is still interface-only, so it uses the published pyatr package for now.

python/samples/02-agents/middleware/atr_validation_middleware.py
```diff
@@ -0,0 +1,205 @@
...
+def _detect_with_atr(text: str) -> str | None:
+    """Run the real ATR engine over *text*; return the matched rule id, or None.
+
+    Evaluates the text as a ``tool_call`` event so it is checked against the rules' ``tool_args``
+    conditions. Returns the highest-severity rule id when one or more rules fire (``evaluate``
+    sorts matches critical-first), otherwise None. Returns None (so the caller falls back) when
+    pyatr is not installed.
+    """
+    try:
+        from pyatr import AgentEvent, ATREngine
```
eavanvalkenburg: we should have a header in this script with the dependencies, including pyatr, then we can just do a normal import and get rid of the fallback, or only keep some of the fallback rules as a comment to allow people to understand the types of rules.

eavanvalkenburg (commented):
with the addition of using pyatr we should simplify the sample!

eeee2345:
@eavanvalkenburg agreed — the fallback makes more sense to drop now that the sample runs the real engine. Pushed: a dependency header declaring `pyatr` (with the install line), a plain top-level import, and the try/except fallback path removed so detection is a single straight-line call. Left three of the old deny-list pattern shapes inline as a short comment so a reader still sees what the rules encode, without the sample carrying a second code path.

Typing stays green with pyatr absent (the env CI uses): ty / pyrefly / pyright-samples / ruff all clean. Runtime checks out too — benign args pass through, the injection arg trips a rule and the middleware blocks the tool call.

Disclosure: I maintain ATR/pyatr.

python/samples/02-agents/middleware/atr_validation_middleware.py
```diff
@@ -0,0 +1,177 @@
+# Copyright (c) Microsoft. All rights reserved.
+
+# Dependencies (beyond agent-framework + azure-identity):
```
eavanvalkenburg: this is not the syntax for a PEP 723 compatible definition of script dependencies, there are a bunch of those in the repo already, just search for `PEP 723`

eavanvalkenburg (approved):
Looks good, just fix the script dependency syntax, I'll get someone on the team to also have a look

eavanvalkenburg:
@eeee2345 please make sure all comments that are addressed are set to resolved

moonbox3 (approved):

eeee2345:
@eavanvalkenburg done — resolved the addressed threads. Quick recap of how each was handled:

- Dependencies/import: the PEP 723 `# /// script` header now declares `agent-framework` + `pyatr`, and the module does a plain top-level import.
- `load_dotenv()`: moved inside `main()` so importing the module has no side effects.
- Inline regex patterns: removed — detection now delegates to the real `pyatr` engine (`load_default_rules()` + `evaluate()`), so the newline/`DOTALL` concern no longer applies.
- `print()` in middleware: the `process()` middleware only uses `logger`; the remaining `print()`s are demo-only inside `main()`.
- Termination message: now includes the matched rule id — `ATR validation blocked tool '<name>' (rule: <id>)`.

Ready for review whenever you have a chance — thanks!

PR #6536, opened 2026-06-16
Python: Improve hosted session scoping

moonbox3:
### Motivation & Context

This change tightens a few Python hosting behaviors around scoped, multi-turn state so that hosted and UI-backed flows behave more predictably in multi-user and local scenarios. It also aligns the Python adapters more closely with the intended ownership boundaries for per-thread and per-response state.

  - Refines AG-UI workflow scoping so resolver-provided scope is applied consistently to live workflow instances.
  - Adds lightweight hosted-session context handling for Foundry Hosting workflow checkpoint paths.
  - Updates package documentation and regression coverage for the adjusted scoping behavior.
  - Python hosted workflows and AG-UI workflow-factory usage get more consistent state separation and clearer configuration behavior.
  - Existing default AG-UI behavior remains unchanged unless a scope resolver is provided.
  - Foundry Hosting gains stricter default handling for workflow checkpoint state, with an explicit non-strict path for local-only scenarios.
  - Whether the new scoping hooks fit the current Python hosting APIs cleanly.
  - Whether the strict and non-strict checkpoint behavior is understandable and covered by the tests.

### Related Issue

#6558 tracking updating docs around this change.

### Contribution Checklist


python/packages/foundry_hosting/agent_framework_foundry_hosting/_responses.py
```diff
@@ -118,6 +118,63 @@
 logger = logging.getLogger(__name__)
 
 _AZURE_RESPONSES_MESSAGE_ROLE_TYPE = f"{MessageRole.__module__}:{MessageRole.__qualname__}"
+_HOSTED_SESSION_CONTEXT_FILE_NAME = "_hosted_session_context.json"
+
```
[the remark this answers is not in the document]
moonbox3: Addressed by moving the hosted-session marker to a non-checkpoint filename so checkpoint scans do not attempt to decode it.

python/packages/foundry_hosting/agent_framework_foundry_hosting/_responses.py
```diff
@@ -118,6 +118,63 @@
...
+        chat_id: Stable chat or conversation isolation key for the current caller.
+    """
+
+    user_id: str
+    chat_id: str
+
+    def __post_init__(self) -> None:
+        """Validate that hosted session identity values are usable as ownership keys."""
+        if not self.user_id or not self.chat_id:
+            raise ValueError("Hosted session context requires non-empty user_id and chat_id values.")
```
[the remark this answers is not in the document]
moonbox3: Addressed by rejecting whitespace-only hosted session keys before stamping.

moonbox3:
Addressed the flagged issue by rejecting whitespace-only hosted session keys before stamping.

python/packages/foundry_hosting/agent_framework_foundry_hosting/_responses.py
```diff
@@ -118,6 +118,63 @@
...
+        chat_id: Stable chat or conversation isolation key for the current caller.
+    """
+
+    user_id: str
+    chat_id: str
+
+    def __post_init__(self) -> None:
+        """Validate that hosted session identity values are usable as ownership keys."""
+        if not self.user_id.strip() or not self.chat_id.strip():
+            raise ValueError("Hosted session context requires non-empty user_id and chat_id values.")
```
eavanvalkenburg, line 140: this is inconsistent with the check, the check is `or` while this states `and`. And I would prefer a regular class for this

python/packages/foundry_hosting/agent_framework_foundry_hosting/_responses.py
```diff
@@ -118,6 +118,63 @@
...
+
+
+class HostedSessionIsolationKeyProvider(Protocol):
```
eavanvalkenburg: do we really need a protocol for this? seems to me this is really specific to Foundry Hosting and if people really need to to override the behavior they can just subclass `PlatformHostedSessionIsolationKeyProvider`

python/packages/foundry_hosting/agent_framework_foundry_hosting/_responses.py
```diff
@@ -118,6 +118,63 @@
...
+            context: The Responses API execution context.
+            request: The current create-response request.
+
+        Returns:
+            The hosted session ownership context, or None when unavailable.
+        """
+        ...
+
+
+class PlatformHostedSessionIsolationKeyProvider:
```
eavanvalkenburg: this could be a callable as well, since it's just one method, much simpler and easier to override

python/packages/foundry_hosting/agent_framework_foundry_hosting/_responses.py
eavanvalkenburg: scratch this, see my other comment, a callable is enough here.

eavanvalkenburg (commented):
left some comments and I'm also a bit confused why this touches both agui and foundry hosting...

python/packages/foundry_hosting/agent_framework_foundry_hosting/_responses.py
moonbox3: Addressed the wording mismatch by documenting that both values must be non-whitespace. Deferred switching this value object to a regular class; the frozen dataclass keeps the two-field ownership context immutable and value-comparable without extra boilerplate.

python/packages/foundry_hosting/agent_framework_foundry_hosting/_responses.py
moonbox3: Addressed by replacing the provider/protocol shape with a sync-or-async `hosted_session_context_resolver` callable.

python/packages/foundry_hosting/agent_framework_foundry_hosting/_responses.py
moonbox3: Addressed by removing the public platform provider class and using a private default resolver callable for the platform isolation values.

moonbox3:
@eavanvalkenburg,

> left some comments and I'm also a bit confused why this touches both agui and foundry hosting...

The two changes cover the same underlying scoping gap, but in two different hosting surfaces. AG-UI's `workflow_factory` path keeps live workflow instances in memory by thread id. We already had Snapshot Scope for persisted snapshots, but the resolver was only applied when snapshot storage was configured. This update makes the same scope apply to the live workflow cache too, so `threadId` alone is not the effective cache boundary.

Foundry Hosting has a separate persistence path: workflow checkpoints are restored by `previous_response_id` / `conversation_id`. That path needed an ownership context persisted beside the checkpoint, so resume checks the current request context before loading checkpoint state.

Of course AG-UI and Foundry Hosting are not coupled in code, but they both had session/state lookup paths where the caller-supplied id needed to be paired with an application/platform scope. That’s why the PR touches both.

eavanvalkenburg (approved):

TaoChenOSU:
We are chatting with Ravi to better understand the user isolation story in FHA. My understand is that the sandbox already provides user-isolation, so we probably don't need to implement it in our layer.

PR #6547, opened 2026-06-16
Python: [BREAKING] Port FileMemoryProvider and integrate FileMemoryProvider & FileAccess into the harness agent

### LINKED ISSUE #6415: .NET: Consolidate Memory implementations between Python and C#
The python memory provider supports fine topic based memories with automatic extraction and injection.
The dotnet memory provider supports course grained memory files without extraction or injection.

These serve different purposes, and we may want to consider supporting both in both languages.

DECISIONS:
Let's replicate FileBasedMemory to python and make that the default.
We can port the fine grained topic to .net later.
Let's also add a sample for showing the HarnessAgent with Foundry Memory (preferably also including other foundry features, e.g. toolbox)

- [x] Port FileMemoryProvider to python (In review)
- [ ] Port FileMemoryProvider to python (done)
- [ ] Add sample showing FileMemoryProvider and Foundry Memory together in .net (In review)
- [ ] Add sample showing FileMemoryProvider and Foundry Memory together in .net (done)
- [ ] Add sample showing FileMemoryProvider and Foundry Memory together in python (In review)
- [ ] Add sample showing  FileMemoryProvider and Foundry Memory together in python (done)

### LINKED ISSUE #6546: Python: Integrate FileAccess into HarnessAgent

westey-m:
### Motivation & Context

The .NET harness already ships a file-backed memory provider and integrates file access into the `HarnessAgent`, but the Python harness lagged behind. This change brings the Python `create_harness_agent` to parity by:

- Porting the .NET `FileMemoryProvider` to Python so harness agents can persist session-scoped memories as files that survive compaction.
- Integrating both `FileMemoryProvider` and `FileAccessProvider` into `create_harness_agent` as on-by-default, disableable, store-overridable features — modeled on the .NET `HarnessAgent`/`HarnessAgentOptions` behavior.

This contributes to the broader effort to consolidate the memory implementations between Python and C# and to make the Python harness feature-complete for file-centric agent scenarios (data processing, long-running research with durable notes).

  - New `agent_framework/_harness/_file_memory.py` implementing `FileMemoryProvider`, a session-scoped, file-backed memory provider built on the existing `AgentFileStore` abstraction (Pythonic port of the .NET `FileMemoryProvider`). Exported from the package public API.
  - `create_harness_agent` now wires up `FileMemoryProvider` and `FileAccessProvider` on by default, with new kwargs to disable each (`disable_file_memory`, `disable_file_access`) or override their stores (`file_memory_store`, `file_access_store`). Defaults use `FileSystemAgentFileStore` rooted at `{cwd}/agent-file-memory` and `{cwd}/working`. The previous ad-hoc memory wiring (`disable_memory`/`memory_store`) was removed in favor of this provider.
  - Tests: added `test_harness_file_memory.py` and updated `test_harness_agent.py` for the new wiring.
  - Harness console sample formatters fixed to match the real Python snake_case tool/arg names (previously used .NET-style PascalCase names that never matched).
  - New sample `samples/02-agents/harness/harness_data_processing.py` (+ bundled `working/sales.csv`) replicating the .NET `Harness_Step03_DataProcessing` sample: a data-analyst harness agent over a folder of CSV data, run in the interactive harness console.
  - Docs: updated `python/packages/core/AGENTS.md`.

  - Python harness agents gain durable file memory and integrated file access out of the box, matching .NET. The previous `disable_memory`/`memory_store` parameters are removed (replaced by the new provider-based kwargs).

  - The `FileMemoryProvider` port — concurrency handling (per-loop/per-folder write locks) and parity with the .NET semantics, and the default-on wiring/override behavior in `create_harness_agent`.

### Related Issue

Fixes #6415
Fixes #6546

### Contribution Checklist


python/packages/core/agent_framework/_harness/_file_memory.py
```diff
@@ -0,0 +1,343 @@
...
+            """Save a memory file with the given name and content. Overwrites the file if it already exists. Include a description for large files to provide a summary that helps with future discovery."""  # noqa: E501
+            try:
+                normalized = _normalize_relative_path(file_name)
+            except ValueError as exc:
+                return f"Could not save file '{file_name}': {exc}"
+            if _is_internal_file(normalized):
+                return (
+                    f"Could not save file '{file_name}': the file name is reserved for internal use. "
+                    "Please choose a different file name."
+                )
```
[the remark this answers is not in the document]
westey-m: Fixed in 292c0a1e8. File memory is intentionally a flat, session-scoped space, so nested names are now rejected up front: `file_memory_save_file` returns a friendly "must not be saved into a subdirectory" message, and `file_memory_read_file`/`file_memory_delete_file` treat a nested name as not found. A new `_is_nested_path` helper checks for `/` after normalization (which also covers `\`). Added `test_tools_reject_nested_paths`.

python/packages/core/agent_framework/_harness/_file_memory.py
```diff
@@ -0,0 +1,343 @@
...
+            async with self._write_lock(working_folder):
+                try:
+                    await self.store.write_file(path, content)
+                    if description and description.strip():
+                        await self.store.write_file(desc_path, description)
+                    else:
+                        await self.store.delete_file(desc_path)
+                    await self._rebuild_index(working_folder)
+                except OSError as exc:
+                    return f"Could not save file '{file_name}': {exc.strerror or exc}"
```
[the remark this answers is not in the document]
westey-m: Fixed in 292c0a1e8. `file_memory_save_file` now also catches `ValueError` (e.g. symlink/reparse-point detection from the store) and returns it as a tool-friendly message instead of letting it escape. Covered by `test_tools_surface_store_value_errors`.

python/packages/core/agent_framework/_harness/_file_memory.py
```diff
@@ -0,0 +1,343 @@
...
+        async def file_memory_read_file(file_name: str) -> str:
+            """Read the content of a memory file by name. Returns the file content or a message indicating the file was not found."""  # noqa: E501
+            try:
+                normalized = _normalize_relative_path(file_name)
+            except ValueError as exc:
+                return f"Could not read file '{file_name}': {exc}"
+            try:
+                content = await self.store.read_file(_combine_paths(working_folder, normalized))
+            except OSError as exc:
+                return f"Could not read file '{file_name}': {exc.strerror or exc}"
```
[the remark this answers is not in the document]
westey-m: Fixed in 292c0a1e8. `file_memory_read_file` now catches `ValueError` (non-UTF8 content, symlink/reparse detection) and returns a readable message. Covered by `test_tools_surface_store_value_errors`.

python/packages/core/agent_framework/_harness/_file_memory.py
```diff
@@ -0,0 +1,343 @@
...
+
+            path = _combine_paths(working_folder, normalized)
+            desc_path = _combine_paths(working_folder, _description_file_name(normalized))
+            async with self._write_lock(working_folder):
+                try:
+                    deleted = await self.store.delete_file(path)
+                    await self.store.delete_file(desc_path)
+                    await self._rebuild_index(working_folder)
+                except OSError as exc:
+                    return f"Could not delete file '{file_name}': {exc.strerror or exc}"
```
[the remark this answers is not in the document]
westey-m: Fixed in 292c0a1e8. `file_memory_delete_file` now catches `ValueError` raised during delete/rebuild and returns a user-visible message. Covered by `test_tools_surface_store_value_errors`.

python/packages/core/agent_framework/_harness/_file_memory.py
```diff
@@ -0,0 +1,343 @@
...
+            [
+                file_memory_save_file,
+                file_memory_read_file,
+                file_memory_delete_file,
+                file_memory_list_files,
+                file_memory_search_files,
+            ],
+        )
+
+        index_content = await self.store.read_file(_combine_paths(working_folder, _MEMORY_INDEX_FILE_NAME))
```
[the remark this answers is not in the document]
westey-m: Fixed in 292c0a1e8. The `memories.md` read in `before_run` is now wrapped in `try/except (OSError, ValueError)`; on failure we log a warning and skip index injection for that run (it self-heals on the next successful save/delete that rebuilds the index). Covered by `test_before_run_skips_injection_when_index_unreadable`.

python/packages/core/agent_framework/_harness/_file_memory.py
```diff
@@ -0,0 +1,343 @@
...
+        working_folder = self._resolve_working_folder(context)
+
+        if working_folder:
+            await self.store.create_directory(working_folder)
+
+        @tool(name="file_memory_save_file", approval_mode="never_require")
+        async def file_memory_save_file(file_name: str, content: str, description: str | None = None) -> str:
+            """Save a memory file with the given name and content. Overwrites the file if it already exists. Include a description for large files to provide a summary that helps with future discovery."""  # noqa: E501
+            try:
+                normalized = _normalize_relative_path(file_name)
```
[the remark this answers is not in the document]
westey-m: Fixed in 292c0a1e8. Rather than make every discovery surface recursive, we keep file memory flat and reject paths containing a separator up front in `file_memory_save_file` (with read/delete treating nested names as not found). A new `_is_nested_path` helper checks for `/` after normalization (`\` is normalized to `/`). Covered by `test_tools_reject_nested_paths`.

python/packages/core/agent_framework/_harness/_agent.py
```diff
@@ -241,8 +254,10 @@ def create_harness_agent(
     todo_provider: TodoProvider | None = None,
     disable_mode: bool = False,
     mode_provider: AgentModeProvider | None = None,
-    disable_memory: bool = False,
-    memory_store: MemoryStore | None = None,
+    disable_file_memory: bool = False,
+    file_memory_store: AgentFileStore | None = None,
+    disable_file_access: bool = False,
+    file_access_store: AgentFileStore | None = None,
```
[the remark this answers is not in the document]
westey-m: Good catch. This is a breaking change (the old `disable_memory`/`memory_store` kwargs are removed and file memory + file access are now on by default), which is why the title carries the `[BREAKING]` prefix. We will reconcile the PR metadata accordingly — adding the `breaking change` label and updating the contribution checklist so it no longer states "This is not a breaking change."

python/packages/core/agent_framework/_harness/_file_access.py
```diff
@@ -82,15 +82,18 @@ def _compile_search_regex(pattern: str) -> re.Pattern[str]:
...
     """
     if len(pattern) > _MAX_SEARCH_PATTERN_LENGTH:
         raise ValueError(
             f"Regex pattern is too long ({len(pattern)} characters). "
             f"Maximum supported length is {_MAX_SEARCH_PATTERN_LENGTH} characters."
         )
-    return re.compile(pattern, flags=re.IGNORECASE)
+    try:
+        return re.compile(pattern, flags=re.IGNORECASE)
+    except re.error as exc:
```
eavanvalkenburg: not sure how much value this adds, if the regex already gives a error message, then why would we catch that, it only makes it so that you have to read the message to understand the error, where before it was either a ValueError or something else

python/packages/core/agent_framework/_harness/_file_memory.py
```diff
@@ -0,0 +1,382 @@
...
+    async def before_run(
+        self,
+        *,
+        agent: Any,
+        session: AgentSession,
+        context: SessionContext,
+        state: dict[str, Any],
+    ) -> None:
+        """Inject file-memory tools, instructions, and the memory index."""
+        del agent, session, state
```
eavanvalkenburg: this is not needed

python/packages/core/agent_framework/_harness/_file_memory.py
```diff
@@ -0,0 +1,382 @@
...
+
+        working_folder = self._resolve_working_folder(context)
+
+        if working_folder:
+            await self.store.create_directory(working_folder)
+
+        @tool(name="file_memory_save_file", approval_mode="never_require")
```
eavanvalkenburg: we should add a fixed description schema to this, this will not have param level descriptions and it is faster to have a fixed schema when possible

python/packages/core/agent_framework/_harness/_file_memory.py
```diff
@@ -0,0 +1,382 @@
...
+                )
+            if _is_internal_file(normalized):
+                return (
+                    f"Could not save file '{file_name}': the file name is reserved for internal use. "
+                    "Please choose a different file name."
+                )
+
+            path = _combine_paths(working_folder, normalized)
+            desc_path = _combine_paths(working_folder, _description_file_name(normalized))
+            async with self._write_lock(working_folder):
```
eavanvalkenburg: I'm wondering if we are overcomplicating this, we are guarding against the same write from multiple agents against the same file at the same time, with a lot of complex machinery and weakrefs etc. I think for a v1 just a simple lock when writing should be enough, as I don't expect many people to set this up to be shared file store between agents and even then the odds of them writing at exactly the same time are slim

python/packages/core/agent_framework/_harness/_file_access.py
westey-m: Reverted in 111f8653e: `_compile_search_regex` no longer wraps `re.error` as a `ValueError`. Since `FunctionTool.invoke` re-raises, an invalid pattern now propagates from the search tools to the model as a tool error so it can correct the regex and retry, keeping the original error message intact. The length-cap check still raises `ValueError`. Tests updated to expect the raised `re.error`.

python/packages/core/agent_framework/_harness/_file_memory.py
westey-m: Removed in 111f8653e: dropped the `del agent, session, state` line from `FileMemoryProvider.before_run` (and the matching one in `FileAccessProvider.before_run` for consistency).

python/packages/core/agent_framework/_harness/_file_memory.py
westey-m: Simplified in 111f8653e: replaced the per-event-loop, per-folder `_WRITE_LOCKS_BY_LOOP` weakref machinery with a single `self._write_lock = asyncio.Lock()` created in `__init__`, and removed the now-unused `weakref`/`ClassVar` imports. This serializes writes per provider instance, which is sufficient for v1.

python/packages/core/agent_framework/_harness/_file_memory.py
westey-m: Done in 111f8653e: added module-level Pydantic input models (with `Field(description=...)` per parameter) and wired them in via `schema=` on the `@tool` decorators. Applied to all five FileMemory tools and all six FileAccess tools for consistency. The fixed schema is built once at import (no per-`before_run` reflection) and now carries param-level descriptions; the tool-level description still comes from each function's docstring.

python/packages/core/agent_framework/_harness/_agent.py
```diff
@@ -151,8 +155,17 @@ def _assemble_context_providers(
     if not disable_mode:
         providers.append(mode_provider or AgentModeProvider())
 
-    if not disable_memory and memory_store is not None:
-        providers.append(MemoryContextProvider(store=memory_store))
+    # File-based session memory (on by default). Default store is rooted at
+    # ``{cwd}/agent-file-memory``; the provider isolates memories per session
+    # via its default ``scope=session_id``.
+    if not disable_file_memory:
+        memory_store = file_memory_store or FileSystemAgentFileStore(Path.cwd() / "agent-file-memory")
```
moonbox3, line 162: Could this defer creating the default filesystem stores until the providers actually run or until the first file operation? `FileSystemAgentFileStore.__init__` immediately calls `mkdir`, so a default `create_harness_agent(...)` now writes under the process CWD during construction (`agent-file-memory` here, and `working` below) and raises in read-only CWDs before any agent run starts. 

Since both providers are on by default, wouldn't this break existing harness construction in serverless/container/test environments that do not allow writing to the working directory?

eavanvalkenburg (approved):

python/packages/core/agent_framework/_harness/_agent.py
westey-m: Good catch — fixed in f6e436528. Rather than deferring store instantiation in `create_harness_agent`, I fixed the root cause in `FileSystemAgentFileStore`: its `__init__` no longer calls `mkdir`, so construction performs no filesystem writes and is safe in read-only CWDs. The root directory is now created lazily on the first `write_file` / `create_directory`; every read/list/search/exists/delete operation already tolerated a missing root (returning None/empty), and `write_file` self-creates the directory tree via `parent.mkdir(...)`. This benefits all callers of the store, not just the harness defaults. Updated the class/`__init__` docstrings and added `test_filesystem_store_does_not_create_root_until_write`.

TaoChenOSU (approved):

TaoChenOSU (approved):

PR #6548, opened 2026-06-16
Python: fix: concurrent_agents sample incorrectly treats output as list[Message]

benke520:
## Problem

The `concurrent_agents.py` sample treats `get_outputs()` results as `list[Message]` directly:

```python
for output in outputs:
    messages: list[Message] | Any = output
    for i, msg in enumerate(messages, start=1):
```

However, at runtime the default `ConcurrentBuilder` aggregator yields `AgentResponse` objects, not `list[Message]`. This causes incorrect iteration behavior.

## Verification

```python
outputs = events.get_outputs()
for output in outputs:
    print(type(output).__name__)  # prints "AgentResponse", not "list"
```

## Fix

Check `isinstance(output, AgentResponse)` and iterate `output.messages` instead of treating the output as a raw list.

## Testing

Ran the sample locally with `agent-framework` v1.8.0 and confirmed the output is correctly printed with the fix.


eavanvalkenburg:
@benke520 there are merge conflicts and open comments, please have a look!

python/samples/03-workflows/orchestrations/concurrent_agents.py
```diff
@@ -77,8 +76,9 @@ async def main() -> None:
     if outputs:
         print("===== Final Aggregated Conversation (messages) =====")
         for output in outputs:
-            messages: list[Message] | Any = output
-            for i, msg in enumerate(messages, start=1):
+            if not isinstance(output, AgentResponse):
+                continue
+            for i, msg in enumerate(output.messages, start=1):
```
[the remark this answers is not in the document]
benke520: Thanks for the feedback. The default ConcurrentBuilder aggregator (_AggregateAgentConversations) always yields AgentResponse — see its ctx.yield_output(AgentResponse(...)) call. Since this sample uses the default aggregator without a custom callback, AgentResponse is the only possible output type here. That said, I've added a warning log for unexpected types so it's not entirely silent if someone adapts the sample with a custom aggregator.

benke520:
> @benke520 there are merge conflicts and open comments, please have a look!

Thanks @eavanvalkenburg — both addressed! Merge conflicts resolved (rebased onto latest main) and replied to the Copilot review comment. Also added a warning print for unexpected output types to avoid silent skipping.

benke520:
> @benke520 there are merge conflicts and open comments, please have a look!

Thanks @eavanvalkenburg — both addressed! Merge conflicts resolved (rebased onto latest main) and replied to the Copilot review comment. Also added a warning print for unexpected output types to avoid silent skipping.

python/samples/03-workflows/orchestrations/concurrent_agents.py
```diff
@@ -80,8 +79,10 @@ async def main() -> None:
     if outputs:
         print("===== Final Aggregated Conversation (messages) =====")
         for output in outputs:
-            messages: list[Message] | Any = output
-            for i, msg in enumerate(messages, start=1):
+            if not isinstance(output, AgentResponse):
+                print(f"[WARNING] Unexpected output type: {type(output).__name__}, skipping.")
```
moonbox3: I'd prefer to not include this warning in the sample.

python/samples/03-workflows/orchestrations/concurrent_agents.py
```diff
@@ -80,8 +79,9 @@ async def main() -> None:
     if outputs:
         print("===== Final Aggregated Conversation (messages) =====")
         for output in outputs:
-            messages: list[Message] | Any = output
-            for i, msg in enumerate(messages, start=1):
+            if not isinstance(output, AgentResponse):
```
moonbox3, line 82: The `AgentResponse` handling looks correct. Could you also update the sample documentation and expected output to match the actual contract? The description still says the default aggregator returns `list[Message]`, and the example output includes the user prompt even though the aggregator returns only the participants’ assistant messages. Also, the `Message` import appears to be unused after this change.

benke520: Good catch - addressed in b861dff: (1) Removed unused Message import, (2) Updated docstring to say the default aggregator yields AgentResponse objects containing only participants' assistant messages instead of list[Message], (3) Removed the 01 [user] entry from sample output since the aggregator intentionally excludes the user prompt, (4) Renumbered remaining entries (researcher=01, marketer=02, legal=03)

moonbox3 (approved):

eavanvalkenburg (approved):

PR #6556, opened 2026-06-16
Python: Align serialized tool format to OTel GenAI tool def format

### LINKED ISSUE #6300: Python: gen_ai.tool.definitions emitted as single JSON blob, breaks Aspire Tools view
We use agent-framework-core in a Python FastAPI chat backend that streams tool-calling responses through an Azure OpenAI client, with traces shipped to a local Aspire dashboard for debugging.

The Python SDK serialises `gen_ai.tool.definitions` as one JSON string covering the whole tools array instead of an array of per-tool JSON strings. The Aspire dashboard's GenAI panel's Tools tab then renders every function as a bare `function` row with no name, description, or parameters. The data is on the span, it just is not parseable in the structured view.

Image

### Where

`python/packages/core/agent_framework/observability.py`, `OTEL_ATTR_MAP["tools"]`:

```python
"tools": (
    OtelAttr.TOOL_DEFINITIONS,
    lambda tools: (
        json.dumps(tools_dict, ensure_ascii=False)
        if (tools_dict := __import__("agent_framework._tools", fromlist=["_tools_to_dict"])._tools_to_dict(tools))
        else None
    ),
    True,
    None,
),
```

### Repro

agent-framework-core 1.2.2 (also identical on `main`), Aspire 13.2 image. Configure an agent with two or more function tools, run a turn, open the `chat <model>` span in Aspire and switch to the Tools tab.

### Suggested fix

Use the correct JSON schema OTel requires:
https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/registry/attributes/gen-ai.md#gen-ai-tool-definitions

TaoChenOSU:
### Motivation & Context

Closes #6300 

Our JSON-serialized tool doesn't align with what OTel GenAI Semantic Conventions define: https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-tool-definitions.json

Tool serialization logic has been centralized to `observability.py`.

Bug fix

Please review on if the serialization logic covers all tool shapes.

### Related Issue

Fixes #6300 

Aspire Dashboard
Before:
image

After:
image

### Contribution Checklist


python/packages/core/agent_framework/observability.py
```diff
@@ -2209,11 +2353,7 @@ def _get_instructions_from_options(options: Any) -> str | list[str] | None:
     # Tools with validation - returns None if no valid tools
     "tools": (
         OtelAttr.TOOL_DEFINITIONS,
-        lambda tools: (
-            json.dumps(tools_dict, ensure_ascii=False)
-            if (tools_dict := __import__("agent_framework._tools", fromlist=["_tools_to_dict"])._tools_to_dict(tools))
-            else None
-        ),
+        lambda tools: json.dumps(tools_dict, ensure_ascii=False) if (tools_dict := _tools_to_dict(tools)) else None,
         True,
```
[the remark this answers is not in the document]
TaoChenOSU: This was not the case:
image

moonbox3 (approved):

python/packages/core/agent_framework/observability.py
```diff
@@ -2209,11 +2353,7 @@ def _get_instructions_from_options(options: Any) -> str | list[str] | None:
     # Tools with validation - returns None if no valid tools
     "tools": (
         OtelAttr.TOOL_DEFINITIONS,
-        lambda tools: (
-            json.dumps(tools_dict, ensure_ascii=False)
-            if (tools_dict := __import__("agent_framework._tools", fromlist=["_tools_to_dict"])._tools_to_dict(tools))
-            else None
-        ),
+        lambda tools: json.dumps(tools_dict, ensure_ascii=False) if (tools_dict := _tools_to_dict(tools)) else None,
```
eavanvalkenburg, line 2424: can we look for a way to cache this, I'm concerned that recalculating all of these on every run is going to have latency impact, which we would like to avoid...

TaoChenOSU: I don't think we can cache this because tools can be dynamically provided per run. We should prioritize accuracy over speed for observability scenarios. Also, these are most often very small objects, which aren't likely the bottleneck. Let's avoid optimize prematurely. If we see performance issue, we will then optimize.

eavanvalkenburg: this is not coming from optimizing prematurely but from experience where bottlenecks come from, so I think we should be able to cache tools that were already parsed into this, so not a while list, but the items in the list, can be just in-memory cache that only works within the run, but even that saves time, there are a lot of checks happening in this kind of stuff and each one has some latency, so it does add up

Necmttn:
Golden fixtures should split tool definition serialization from tool call serialization.

A tool definition has name, description, input schema, and maybe output schema. A tool call instance has id, selected name, arguments, and result/error later. If tests cover both shapes side by side, future changes are less likely to leak runtime arguments into `gen_ai.tool.definitions` or lose schema fields when a tool has nested objects/arrays. Include a no-description tool and a nested-schema tool.

---

_Generated with [ax](https://github.com/Necmttn/ax)._

python/packages/core/agent_framework/observability.py
TaoChenOSU: Added

giles17 (approved):

PR #6570, opened 2026-06-17
Python: consolidate dependency maintenance workflow

### LINKED ISSUE #6569: Python: consolidate dependency maintenance automation
Python dependency updates are currently handled by Dependabot plus separate manual workflows for dev dependency upgrades and dependency range validation.

That creates competing PR streams for the Python uv workspace, where dependencies can be declared in multiple `pyproject.toml` files but are resolved into one shared `python/uv.lock`. Separate update PRs can conflict or force unnecessary rebasing.

Create one Python dependency-maintenance workflow that runs weekly and can be triggered manually. It should:

- update Python dev dependency pins
- refresh the shared `python/uv.lock`
- run dependency bounds test scenarios
- run repo-wide upper-bound validation
- create issues for failed upper-bound candidates
- open or update a single Python dependency-maintenance PR

Keep Python Dependabot as a temporary fallback on a later weekday until the workflow has proven reliable, then remove it.

eavanvalkenburg:
### Motivation & Context

Python dependency updates need to be coordinated across the uv workspace because dependency declarations can appear in multiple `pyproject.toml` files while the workspace has one shared `python/uv.lock`. Separate automation PRs for dev dependency updates and dependency range updates can conflict with each other and require unnecessary rebasing.

This change consolidates Python dependency maintenance into one scheduled/manual workflow that updates and validates the generated dependency state together. Python Dependabot stays as a temporary Thursday fallback until the new workflow has proven reliable.

- **What are the major changes?** Add a single `Python - Dependency Maintenance` workflow that runs Mondays at 04:00 UTC, folds in dev dependency repinning and dependency upper-bound validation, applies a 7-day dependency release cutoff, and opens/updates one automation PR.
- **What is the impact of these changes?** The weekly Python automation produces at most one dependency PR and avoids newly published packages by using `UV_EXCLUDE_NEWER` plus matching PyPI upload-time filtering. If upper-bound validation fails, the workflow keeps only dev dependency updates that still pass final checks and creates issues for failed range candidates.
- **What do you want reviewers to focus on?** Review the fallback behavior for failed range validation, the schedule/Dependabot overlap, and whether the generated PR body follows the repository template closely enough for automation.

### Related Issue

Fixes #6569

No other open PR was found for this issue.

### Contribution Checklist


.github/workflows/python-dependency-maintenance.yml
```diff
@@ -60,7 +91,7 @@ jobs:
             const reportPath = "python/scripts/dependencies/dependency-range-results.json"
```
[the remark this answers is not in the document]
eavanvalkenburg: Addressed in 82bf32317: added a separate issue-creation step for dependency bounds test failures. When validate-dependency-bounds-test fails before upper-bound validation runs, the workflow now reads dependency-bounds-test-results.json, creates/dedupes a tracking issue, and still falls back to dev-only updates if those pass final validation.

.github/workflows/python-dependency-maintenance.yml
```diff
@@ -0,0 +1,362 @@
...
+
+jobs:
+  dependency-maintenance:
+    name: Dependency Maintenance
+    runs-on: ubuntu-latest
+    env:
+      # Match the existing Python dependency maintenance workflows. Reevaluate if package
+      # installability starts differing across supported Python versions.
+      UV_PYTHON: "3.13"
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
moonbox3: From a security perspective, should we avoid exposing the write-capable `GITHUB_TOKEN` to the dependency resolution and validation  steps? This job installs and executes newly selected PyPI versions, and those subprocesses inherit `GH_TOKEN` while it has  contents/issues/PR write permissions. Could we scope the token only to the issue, push, and PR publishing steps instead of at this current level?

TaoChenOSU (approved):

.github/workflows/python-dependency-maintenance.yml
eavanvalkenburg: Addressed in 992657bf9: removed `GH_TOKEN` from the job-level environment so dependency resolution, installs, validation, and final checks no longer inherit the write-capable token. The token is now provided only to the GitHub API/CLI publishing steps that need it (`github-token` for github-script issue creation and step-scoped `GH_TOKEN` for `gh pr` create/update).

moonbox3 (approved):

PR #6579, opened 2026-06-18
Python: adjust checkpoint encoding handling

moonbox3:
### Motivation & Context

Checkpoint persistence should preserve ordinary workflow state values while keeping internal encoding metadata clearly separated from  user data. This change refines checkpoint encoding behavior for dictionaries that contain reserved metadata-like keys and tightens  reconstruction behavior without changing the public checkpoint storage API.

  - Encode dictionaries containing checkpoint-reserved keys through the existing pickle envelope so they round-trip without introducing  a new JSON wrapper shape.
  - Limit automatic framework/OpenAI reconstruction to concrete classes while preserving explicit `allowed_checkpoint_types` behavior.
  - Add regression coverage for reserved-key dictionaries, `FileCheckpointStorage` round-trips, and restricted decode behavior.
  - Existing checkpoints continue to restore through the current decode path.
  - Newly saved checkpoint state with reserved-key dictionaries round-trips as user data.
  - Applications can still opt in to additional checkpoint types with `allowed_checkpoint_types`.
  - Compatibility of reserved-key dictionary handling and the restricted decode behavior for framework/OpenAI classes.

### Related Issue

N/A.

### Contribution Checklist

- [x] **This is not a breaking change.** If it _is_ a breaking change, add the `breaking change` label (or add "[BREAKING]" to the title  prefix, before or after any language prefix) — a workflow keeps the label and title prefix in sync automatically.


python/packages/core/agent_framework/_workflows/_checkpoint_encoding.py
```diff
@@ -217,17 +233,18 @@ def _encode(value: Any) -> Any:
 
     # Recursively encode dict values (keys become strings)
     if isinstance(value, dict):
-        return {str(k): _encode(v) for k, v in value.items()}  # type: ignore
+        typed_dict = cast(dict[Any, Any], value)  # type: ignore[redundant-cast]
+        encoded_dict: dict[str, Any] = {str(k): _encode(v) for k, v in typed_dict.items()}
+        if _RESERVED_DICT_KEYS.intersection(encoded_dict):
+            return _encode_pickle(value)
+        return encoded_dict
```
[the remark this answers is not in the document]
moonbox3: Addressed in 4ac3f74ad by checking for reserved keys before recursively encoding dict values, while preserving the existing key-stringification behavior for collision detection.

eavanvalkenburg (approved):

moonbox3:
@copilot fix the merge conflicts

moonbox3:
@copilot resolve the merge conflicts in this pull request

python/packages/core/agent_framework/_workflows/_checkpoint_encoding.py
```diff
@@ -69,6 +73,13 @@
 # Module prefix for OpenAI SDK types that are always allowed
 _OPENAI_MODULE_PREFIX = "openai.types."
 
+_DENIED_ALLOWED_TYPE_KEYS: frozenset[str] = frozenset({
+    "agent_framework._workflows._checkpoint_encoding:_RestrictedUnpickler",
+    "agent_framework._workflows._checkpoint_encoding:_base64_to_unpickle",
+    "agent_framework._workflows._checkpoint_encoding:decode_checkpoint_value",
+    "agent_framework._workflows._checkpoint_encoding:encode_checkpoint_value",
```
TaoChenOSU, line 81: Question:
1. why are methods included here? 
2. What does "denied_allowed" mean?

python/packages/core/agent_framework/_workflows/_checkpoint_encoding.py
```diff
@@ -128,13 +139,18 @@ def __init__(self, data: bytes, allowed_types: frozenset[str]) -> None:
...
-            return super().find_class(module, name)
+        if type_key in _DENIED_ALLOWED_TYPE_KEYS:
+            raise pickle.UnpicklingError(f"Checkpoint deserialization blocked for type '{type_key}'.")
+
+        if type_key in _BUILTIN_ALLOWED_TYPE_KEYS or type_key in self._allowed_types:
+            return super().find_class(module, name)  # type: ignore[no-any-return]  # nosec
+
+        if module.startswith(_FRAMEWORK_MODULE_PREFIX) or module.startswith(_OPENAI_MODULE_PREFIX):
+            if "." in name:
+                raise pickle.UnpicklingError(f"Checkpoint deserialization blocked for type '{type_key}'.")
```
TaoChenOSU, line 153: nit: some comment here would be nice for why the type is blocked when the name contains a period.

python/packages/core/agent_framework/_workflows/_checkpoint_encoding.py
```diff
@@ -217,17 +233,18 @@ def _encode(value: Any) -> Any:
 
     # Recursively encode dict values (keys become strings)
     if isinstance(value, dict):
-        return {str(k): _encode(v) for k, v in value.items()}  # type: ignore
+        typed_dict = cast(dict[Any, Any], value)  # type: ignore[redundant-cast]
+        if any(str(k) in _RESERVED_DICT_KEYS for k in typed_dict):
```
TaoChenOSU, line 240: Question: When will the value to be encoded contain reserved keys?

Should we raise instead if the value provided contains reserved keys?

python/packages/core/agent_framework/_workflows/_checkpoint_encoding.py
moonbox3: Addressed in the follow-up change. These entries are module-level globals that pickle can resolve during unpickling, not just classes. Since `agent_framework.*` is otherwise auto-allowed, these checkpoint helper callables need an explicit block so a crafted payload cannot call them during reconstruction. I also renamed `_DENIED_ALLOWED_TYPE_KEYS` to `_BLOCKED_FRAMEWORK_GLOBAL_KEYS` and added a comment to make the intent clearer.

python/packages/core/agent_framework/_workflows/_checkpoint_encoding.py
moonbox3: Addressed by adding a comment. Dotted pickle global names can traverse attributes on an otherwise allowed module, so the framework/OpenAI prefix allowlist stays limited to concrete top-level classes rather than helper callables reachable through module attributes. Explicitly listed `allowed_types` still win before this check.

python/packages/core/agent_framework/_workflows/_checkpoint_encoding.py
moonbox3: I kept the current behavior here. Checkpoint state can contain arbitrary user/provider dictionaries, including metadata-shaped keys like `__type__` or `__pickled__`. Those names are reserved for the checkpoint encoding envelope, but they should not be invalid as user state. Pickling the whole marker-shaped dict preserves the original data and keeps decode unambiguous, including nested dictionaries and `FileCheckpointStorage` round-trips.

TaoChenOSU (approved):

giles17 (approved):

PR #6544, opened 2026-06-16
.NET: Integrate LoopAgent into HarnessAgent with TodoCompletionLoopEvaluator

westey-m:
### Motivation & Context

The core `LoopAgent` (added in #6384) makes it possible to re-invoke an agent until configurable criteria are met. This PR builds on that to make looping a first-class, opt-in capability of `HarnessAgent`, delivering the "loop until completion" enforcement scenario for harnesses.

It also adds a `TodoCompletionLoopEvaluator` so a harness can keep working autonomously until every todo item is complete — optionally scoped to specific agent modes — which is exactly the kind of completion-enforcement behavior the harness experience needs.

### Description & Review Guide

- **What are the major changes?**
  - `HarnessAgentOptions` gains `LoopEvaluators` (`IEnumerable<LoopEvaluator>?`) and `LoopAgentOptions`. When at least one evaluator is supplied, `HarnessAgent.BuildAgent` wraps the harness in a `LoopAgent` as the outermost decorator; when none are supplied, behavior is unchanged.
  - New `TodoCompletionLoopEvaluator` (+ `TodoCompletionLoopEvaluatorOptions`): resolves `TodoProvider` (and `AgentModeProvider` when `Modes` is set) from `context.Agent.GetService<T>()`, and re-invokes with a customizable feedback message while incomplete todos remain. `Modes == null` applies in all modes; a non-null empty set is an error; mode matching is ordinal.
  - The `Harness_Step01_Research` sample now demonstrates the evaluator scoped to `"execute"` mode with a `MaxIterations` safety cap.
- **What is the impact of these changes?**
  - Purely additive and opt-in. Existing harnesses with no `LoopEvaluators` behave exactly as before.
- **What do you want reviewers to focus on?**
  - Provider resolution via the `GetService` decorator chain (no provider constructor params), and the decorator ordering that makes `LoopAgent` outermost.
  - The null-vs-empty `Modes` semantics.

### Related Issue

Fixes #6478

### Contribution Checklist

- [x] The code builds clean without any errors or warnings
- [x] All unit tests pass, and I have added new tests where possible
- [x] The PR follows the [Contribution Guidelines](https://github.com/microsoft/agent-framework/blob/main/CONTRIBUTING.md)
- [x] This PR is linked to an issue and there is no other open PR for this issue (see Related Issue above).
[...]

PR #6607, opened 2026-06-18
Python: [BREAKING] Integrate looping into HarnessAgent

### LINKED ISSUE #6478: .NET: Integrate Looping into HarnessAgent
- [x] .net (in review)
- [x] .net (done)
- [x] python (in review)
- [x] python (done)

westey-m:
### Motivation & Context

The core `AgentLoopMiddleware` makes it possible to re-invoke an agent until configurable criteria are met. This PR builds on that to make looping a first-class, opt-in capability of the harness, delivering the "loop until completion" enforcement scenario for harnesses.

It also adds shared `todos_remaining` / `todos_remaining_message` helpers so a harness can keep working autonomously until every todo item is complete — optionally scoped to specific agent modes — which is exactly the kind of completion-enforcement behaviour the harness experience needs. This is the Python counterpart of the .NET work in #6544.

  - `create_harness_agent` gains `loop_should_continue`, `loop_next_message`, and `loop_max_iterations` keyword arguments. When `loop_should_continue` is supplied, the harness is wrapped in an `AgentLoopMiddleware` wired as the **outermost** middleware (each iteration is a full agent run, including tool approval); when it is `None`, behaviour is unchanged.
  - **Approval escape hatch**: if a loop iteration returns a pending tool-approval request, the loop stops and returns it to the caller *before* evaluating `should_continue` or injecting `next_message`, so looping is HITL-safe even when wrapped around `ToolApprovalMiddleware`.
  - New shared helpers in `_loop.py`, exported from the package: `todos_remaining(*, modes=None)` resolves the `TodoProvider` (and `AgentModeProvider` when `modes` is set) from `agent.context_providers` and loops while incomplete todos remain (`modes=None` applies in all modes; a non-empty sequence gates by mode, case-insensitively; an empty sequence raises `ValueError`). `todos_remaining_message` is a `next_message` callable that lists the still-open todos. These merge and replace the former provider-argument helper.
  - The `harness_research.py` sample demonstrates the loop scoped to `"execute"` mode with a `loop_max_iterations` safety cap; the two `agent_loop_middleware_*` samples switch to the no-argument `todos_remaining()`.
  - Tests and `AGENTS.md` updated; conforms to the new split type-checker setup (pyright on source; pyright/mypy/pyrefly/ty/zuban on tests/samples).

  - Looping is opt-in and additive for `create_harness_agent` (no loop unless `loop_should_continue` is set).
  - **Breaking**: the public `todos_remaining(provider)` signature is removed and replaced by `todos_remaining(*, modes=None)`, which resolves the provider from the running agent's `context_providers` instead of taking it as an argument. Callers passing a provider must update to the new form.

  - The middleware ordering (loop outermost vs. tool approval) and the approval escape-hatch semantics.
  - The merged `todos_remaining` API shape and its mode-gating behaviour.

### Related Issue

Fixes #6478

### Contribution Checklist


python/packages/core/agent_framework/_harness/_loop.py
```diff
@@ -794,3 +804,107 @@ def _should_continue(*, session: Any = None, **kwargs: Any) -> bool:
...
+    """``next_message`` callable that reminds the agent which todos are still open.
+
+    Designed to pair with :func:`todos_remaining` as a loop's ``next_message`` (e.g.
+    ``create_harness_agent``'s ``loop_next_message``): between iterations it resolves the harness
+    :class:`~agent_framework.TodoProvider` from the agent, lists the still-open todo items, and
+    instructs the agent to complete them all before finishing.
+
+    Returns ``None`` when the session/agent/provider is unavailable or no todos are open, so the loop
+    falls back to its default "continue" nudge.
+    """
```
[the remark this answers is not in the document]
westey-m: Good catch — fixed in eb9de4e88. The `todos_remaining_message` docstring now states that returning `None` makes the loop reuse the previous iteration's messages verbatim under the default `fresh_context=False` (skipping progress injection), and only falls back to `DEFAULT_NEXT_MESSAGE` when `fresh_context=True`.

python/packages/core/AGENTS.md
```diff
@@ -120,7 +120,10 @@ agent_framework/
...
+- **`todos_remaining(*, modes=None)`** / **`todos_remaining_message`** - Helper factories for todo-driven loops (the Python counterpart of .NET's `TodoCompletionLoopEvaluator`), designed for `create_harness_agent` but usable with any agent that registers a `TodoProvider` via `context_providers`. They resolve the `TodoProvider`/`AgentModeProvider` from the *running agent* (`agent.context_providers`, via `_resolve_context_provider`) rather than taking the provider as an argument, so they can be wired directly into `loop_should_continue`/`loop_next_message`. `todos_remaining` returns a `should_continue` predicate that loops while any todo is open; pass `modes=[...]` to gate looping to specific operating modes (case-insensitive; honors the `AgentModeProvider`'s `source_id`/`available_modes`), `modes=None` (default) applies in every mode, and an empty sequence raises `ValueError`. `todos_remaining_message` is a `next_message` callable that lists the still-open todo titles and tells the agent to finish them, returning `None` (→ default nudge) when the session/agent/provider is unavailable or nothing is open.
```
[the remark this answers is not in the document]
westey-m: Fixed in eb9de4e88. The AGENTS.md bullet no longer says "→ default nudge"; it now describes the real behaviour: returning `None` reuses the previous iteration's messages verbatim under the default `fresh_context=False`, and uses `DEFAULT_NEXT_MESSAGE` only when `fresh_context=True`.

moonbox3 (approved):

Ashutosh0x:
## Review: HarnessAgent Looping Integration

Interesting @westey-m — integrating looping directly into HarnessAgent simplifies the pattern significantly. Instead of external retry/loop logic, the agent can self-iterate.

### Questions

1. **Loop termination**: What controls the exit condition? Is it a max-iterations guard, an explicit stop signal from the agent, or both? Both would be safest — agents that loop indefinitely are a production risk.

2. **Checkpoint interaction**: Does each loop iteration create a new checkpoint? If so, can the agent resume from a specific iteration on crash recovery?

3. **Observability**: Are loop iterations surfaced as individual spans in the OTel trace? This would make debugging loop behavior much easier (I see @TaoChenOSU is actively building OTel support in the Python side).

### Context

I contributed an ADR for dynamic agent resolution (PR #6643) and bug fixes in the Python hosting stack (PR #6642). The looping pattern would interact with the hosting channels — specifically, how does a looping agent's output get projected to a single Responses API response? Eager to understand the design here.

Solid work!

python/packages/core/agent_framework/_harness/_loop.py
```diff
@@ -794,3 +804,111 @@ def _should_continue(*, session: Any = None, **kwargs: Any) -> bool:
...
+    :func:`~agent_framework.create_harness_agent` wired up instead of constructing their own.
+    """
+    return next(
+        (provider for provider in getattr(agent, "context_providers", []) if isinstance(provider, provider_type)),
+        None,
+    )
+
+
+def todos_remaining(*, modes: Sequence[str] | None = None) -> ShouldContinueCallable:
+    """Build a ``should_continue`` predicate that loops while the harness ``TodoProvider`` has open items.
```
eavanvalkenburg:
```suggestion
    """Build a ``should_continue`` predicate that loops while the Agent's ``TodoProvider`` has open items.
```

python/packages/core/agent_framework/_harness/_loop.py
```diff
@@ -794,3 +804,111 @@ def _should_continue(*, session: Any = None, **kwargs: Any) -> bool:
...
+
+    This resolves the :class:`~agent_framework.TodoProvider` from the running agent
+    (``agent.context_providers``) rather than taking it as an argument, so it can be used directly
+    with :func:`~agent_framework.create_harness_agent` (whose providers are built internally) as well
+    as with any agent that registers a ``TodoProvider`` via ``context_providers``. It is the Python
+    counterpart of the .NET ``TodoCompletionLoopEvaluator``.
+
+    Args:
+        modes: When provided, the loop only continues while the agent's current operating mode (read
```
eavanvalkenburg: and i have a bit of trouble wrapping my head around what this means, i think we need to make the param name more self-explanatory, something like `looping_modes` maybe?

python/packages/core/agent_framework/_harness/_loop.py
westey-m: Thanks — applied your suggestion in 72266c7ae. The first docstring line now reads "...loops while the Agent's ``TodoProvider`` has open items." (it works with any agent that registers a `TodoProvider` via `context_providers`, not just the harness).

python/packages/core/agent_framework/_harness/_loop.py
westey-m: Good call — renamed `modes` to `looping_modes` in 72266c7ae. Updated the signature, docstring `Args`, and the `ValueError` message in `_loop.py`, plus all references in `packages/core/AGENTS.md`, the `harness_research.py` sample, and `test_harness_loop.py`.

eavanvalkenburg (approved):

PR #6610, opened 2026-06-19
Python: [wip]: Prototype foundry hosting workflow adapter

moonbox3:
An example of using a foundry hosted adapter for workflows


moonbox3:
@TaoChenOSU please use this as the basis of forming the proper abstractions around foundry-hosting and not low-level core workflow apis.

PR #5100, opened 2026-04-05
Python: Add include_reference_source_data to agentic search requests

mvanhorn:
## Summary

`_agentic_search()` in `AzureAISearchContextProvider` constructs `KnowledgeBaseRetrievalRequest` without passing `knowledge_source_params`, so `include_reference_source_data` is never set. This causes `ref.source_data` to always be `None` on returned references, even when the knowledge source has `source_data_fields` configured.

## Changes

- Pass `knowledge_source_params=[KnowledgeSourceParams(..., include_reference_source_data=True)]` to both `KnowledgeBaseRetrievalRequest` constructions in `_agentic_search()` (the minimal reasoning path at line 828 and the non-minimal path at line 836)
- Add `KnowledgeSourceParams` to both `TYPE_CHECKING` and runtime import blocks
- Add two unit tests verifying the parameter is passed for both reasoning effort paths

## Testing

- Existing 119 tests pass
- New tests verify `knowledge_source_params` is present with correct `knowledge_source_name` and `include_reference_source_data=True` for both the minimal and non-minimal reasoning paths

Fixes #5095

This contribution was developed with AI assistance (Claude Code).

Issue #5095, opened 2026-04-04
Python: AzureAISearchContextProvider agentic mode: `source_data` is always `None` on references — missing `include_reference_source_data` parameter


dantelmomsft:
### Description

When using `AzureAISearchContextProvider` in `agentic` mode, the `source_data` field on `KnowledgeBaseReference` objects in the retrieval response is always `None`, even when the knowledge source has `source_data_fields` correctly configured (e.g. `id`, `content`, `source_file_name`, `offset`, `page_number`).

This prevents consumers from accessing structured citation metadata (document names, page numbers, content excerpts) from agentic retrieval references.

### Root Cause

The Azure AI Search agentic retrieval API requires callers to **opt-in** to receiving source data on references at request time by passing `include_reference_source_data=True` via `knowledge_source_params` on the `KnowledgeBaseRetrievalRequest`.

The `source_data_fields` on the knowledge source definition only controls **which** fields to return — the caller must also set this runtime flag per request.

`_agentic_search()` (line ~808 in `_context_provider.py`) builds the `KnowledgeBaseRetrievalRequest` without `knowledge_source_params`, so the flag is never set:

```python
# Current code — no knowledge_source_params
retrieval_request = KnowledgeBaseRetrievalRequest(
    messages=kb_messages,
    retrieval_reasoning_effort=reasoning_effort,
    output_mode=output_mode,
    include_activity=True,
    # knowledge_source_params is missing
)
```

### Expected Behavior

When a knowledge source has `source_data_fields` configured, `ref.source_data` on each `KnowledgeBaseSearchIndexReference` should contain a dict with the configured fields, e.g.:

```python
{
    'id': 'aHR0cHM6Ly8...',
    'content': 'Overview\nIntroducing PerksPlus...',
    'source_file_name': 'integration-test/PerksPlus.pdf',
    'offset': 0,
    'page_number': 3,
}
```

### Actual Behavior

`ref.source_data` is always `None` for every reference returned by agentic retrieval.

### Steps to Reproduce

1. Create a search index with a knowledge source that has `source_data_fields` configured
2. Create a knowledge base referencing that knowledge source
3. Use `AzureAISearchContextProvider` in `agentic` mode with `knowledge_base_name` pointing to the existing KB
4. Run a query — inspect `ref.source_data` on the returned references → always `None`

### Verified Fix

Passing `knowledge_source_params` with `include_reference_source_data=True` on the retrieval request resolves the issue:

```python
[...]

PR #6612, opened 2026-06-19
Python: support knowledge_source_params in AzureAISearchContextProvider

### LINKED ISSUE #5560: [Feature]: AzureAISearchContextProvider lacks a way to set KnowledgeSourceParams
### Description

KnowledgeSourceParams are crucial to configure AI Search with respect to the retrieval of extra fields (IncludeReferenceSourceData field) and many other advanced behaviours.

### Code Sample

```markdown
Add knowledgeSourceParams in the constructor of AzureAISearchContextProvider:


scc = AzureAISearchContextProvider(..., knowledgeSourceParams=[ ... ])


var scc = new AzureAISearchContextProvider(..., knowledgeSourceParams: [ ... ]);
```

### Language/SDK

Both

Athosone:
## Summary

Adds an optional keyword-only `knowledge_source_params: list[KnowledgeSourceParams] | None = None`
to `AzureAISearchContextProvider`, forwarded verbatim into both
`KnowledgeBaseRetrievalRequest` constructions in `_agentic_search`.

This lets callers set per-source agentic retrieval options — most importantly
`filter_add_on` (server-side OData filtering), and also
`include_reference_source_data` and the other `KnowledgeSourceParams` fields —
without subclassing the provider or reaching into private methods.

## Behavior

- **Back-compatible:** default `None` means retrieval requests are built exactly as before.
- **Agentic-only:** the parameter is omitted from the semantic overload, so passing it
  in semantic mode is a type error; a runtime `ValueError` also guards untyped callers.

## Relationship to existing work

Superset of #5095 / #5100 (which hardcodes `include_reference_source_data`): this exposes
the full `KnowledgeSourceParams` surface generically.

## Tests

- `knowledge_source_params` reaches the request in both the minimal (intents) and
  non-minimal (messages) agentic branches.
- Default path sends `None` (back-compat).
- Passing the parameter in semantic mode raises.

`uv run poe test -P azure-ai-search` (121 passed, 97% coverage), `uv run poe syntax -P azure-ai-search`,
and `uv run poe pyright -P azure-ai-search` all pass.

Closes #5560

### DIFF
--- a/python/packages/azure-ai-search/agent_framework_azure_ai_search/_context_provider.py
+++ b/python/packages/azure-ai-search/agent_framework_azure_ai_search/_context_provider.py
@@ -64,6 +64,7 @@ if TYPE_CHECKING:
         KnowledgeBaseRetrievalResponse,
         KnowledgeRetrievalIntent,
         KnowledgeRetrievalSemanticIntent,
+        KnowledgeSourceParams,
     )
     from azure.search.documents.knowledgebases.models import (
         KnowledgeRetrievalLowReasoningEffort as KBRetrievalLowReasoningEffort,
@@ -239,6 +240,7 @@ class AzureAISearchContextProvider(ContextProvider):
         azure_openai_api_key: str | None = None,
         knowledge_base_output_mode: KnowledgeBaseOutputModeLiteral = "extractive_data",
         retrieval_reasoning_effort: RetrievalReasoningEffortLiteral = "minimal",
+        knowledge_source_params: list[KnowledgeSourceParams] | None = None,
         agentic_message_history_count: int = _DEFAULT_AGENTIC_MESSAGE_HISTORY_COUNT,
         env_file_path: str | None = None,
         env_file_encoding: str | None = None,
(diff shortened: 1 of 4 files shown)


eavanvalkenburg:
there was a recent update to the this class, can you catch up with main and see if this is still relevant @Athosone

python/packages/azure-ai-search/agent_framework_azure_ai_search/_context_provider.py
```diff
@@ -830,6 +847,7 @@ async def _agentic_search(self, messages: list[Message]) -> list[Message]:
                 retrieval_reasoning_effort=reasoning_effort,
                 output_mode=output_mode,
                 include_activity=True,
+                knowledge_source_params=self._knowledge_source_params,
```
moonbox3, line 850: Should this preserve the source-data params that are now built on main before forwarding caller params? Current main resolves `_knowledge_source_names` and sends `include_reference_source_data=True`; this branch sends `None` on the default path, so existing agentic retrieval loses `ref.source_data` again after the conflict is resolved this way. Could we merge caller-supplied fields with the resolved-source defaults instead of replacing them?

moonbox3:
Please re-open when ready to address the conflicts, and if the fix is needed as it may not be any longer.

Issue #4893, opened 2026-03-25
[Feature]: Align concepts between AG-UI and A2A protocols


lixiaoqiang:
### Description

Currently, session_id corresponds to thread_id in AG-UI and context_id in A2A, and message_id corresponds to message_id / artifact_id in both AG-UI and A2A.
However, we are missing the corresponding concept mapping between task_id in A2A and run_id in AG-UI.

|AG-UI|A2A|agentframework|
|-|-|-|
|thread\_id|context\_id|session\_id|
|run\_id|task\_id|??|
|message\_id|message\_id/artifact\_id|message\_id|

### Code Sample

```markdown

```

### Language/SDK

Both

moonbox3:
Thanks for filing this, @lixiaoqiang. It's a fair observation about the missing cell. We'll discuss as a team how we want to handle it in MAF (including how it should interact with workflow runs, which raises a related question). Tagging @eavanvalkenburg so he's aware.

eavanvalkenburg:
After looking at this more closely, I don't think `run_id` in AG-UI and `task_id` in A2A should be mapped to the same Agent Framework concept.

They have different lifetimes and different semantics:

- AG-UI `run_id` is per-request / per-run event correlation. It lets the AG-UI frontend and backend correlate `RUN_STARTED`, streamed events, and `RUN_FINISHED` for one invocation. If the caller does not provide it, the AG-UI wrapper can generate one. The wrapped agent generally does not need to know it.
- A2A `task_id` is a protocol task identifier. It can affect future A2A calls: for example, it may be used to continue an `INPUT_REQUIRED` task, to subscribe/poll an in-progress task, or to relate later protocol behavior to a task. That makes it durable protocol state in a way AG-UI `run_id` is not.

So the table should not have a shared `run_id`/`task_id` row in Agent Framework. The closer mapping is:

| AG-UI | A2A | Agent Framework |
|---|---|---|
| `thread_id` | `context_id` | session / conversation identity |
| `run_id` | — | AG-UI event correlation owned by the AG-UI wrapper |
| — | `task_id` | A2A task/protocol state, and sometimes continuation-token payload |
| `message_id` | `message_id` / `artifact_id` | message / response identity |

That means we do not need to align A2A `task_id` with AG-UI `run_id`. If we later need a generic telemetry correlation concept, that should be handled as run/telemetry context, not by treating these protocol fields as equivalent session state.

eavanvalkenburg:
[...]

Issue #4673, opened 2026-03-13
Python: [Bug]: No Native Conversation ID Propagation


sphenry:
### Description

Framework has no concept of a client-scoped conversation_id that flows through A2A handoffs, tools, and telemetry.

Cannot correlate end-to-end conversation traces across orchestrator and sub-agents without custom plumbing.

A solution is Manual contextVar propagation across all server handlers + filtering framework-injected conversation_id kwarg from MCP tool calls

### Code Sample

```markdown

```

### Error Messages / Stack Traces

```markdown

```

### Package Versions

python-1.0.0rc4

eavanvalkenburg:
@sphenry could you provide a bit more context here?

PR #6630, opened 2026-06-19
docs: Add Python session identity ADR

eavanvalkenburg:
### Motivation & Context

Python protocol and service integrations currently use session, task, response, and continuation identifiers with overlapping terminology. This ADR documents the lifetime boundaries for those identities so follow-up implementation work for native conversation propagation and A2A task/session behavior has a clearer design basis.

The ADR keeps AG-UI out of the session identity decision, preserves `service_session_id` as an opaque service continuation handle, and focuses the remaining open design question on how Python should store additional durable continuation state across calls.

  - Adds ADR 0029 for Python identity lifetimes across sessions, tasks, service continuation, response IDs, and continuation tokens.
  - Records the chosen direction to split identity storage by lifecycle.
  - Lists the remaining durable-state options: protocol-specific `AgentSession` subclasses, namespaced `AgentSession.state`, or richer service-owned `service_session_id` values.
  - Adds an A2A appendix for `task_id` vs `reference_task_ids` behavior to validate in implementation.
  - No runtime behavior changes.
  - Provides design guidance for future Python work on #4673 and #4893.
  - Whether the ADR correctly separates service continuation, durable protocol state, response identity, and continuation tokens.
  - Whether the remaining durable-state options are framed clearly enough for a follow-up decision.

### Related Issue

Related to #4673 and #4893. Does not close them.

### Contribution Checklist

### DIFF
--- a/docs/decisions/0029-python-agent-session-identity.md
+++ b/docs/decisions/0029-python-agent-session-identity.md
@@ -1,0 +1,237 @@
+---
+status: proposed
+contact: eavanvalkenburg
+date: 2026-06-19
+deciders: eavanvalkenburg, moonbox3, TaoChenOSU, chetantoshnival
+consulted: westey-m
+informed:
+---
+
+# Python identity lifetimes for sessions, tasks, and continuation
+
+## Context and Problem Statement
+
+Python `AgentSession` currently carries a local `session_id`, an optional opaque service continuation
+`service_session_id`, and provider state. `service_session_id` is any service-owned value that lets that service continue
+a conversation, session, or thread; chat clients happen to map it through the abstract `conversation_id` option, but
+other agent types can use it differently. It is not a generic correlation field, and generic correlation should not
+require parsing or understanding that opaque service-owned value.
+
+The related issues mix values with different lifetimes:
+
+- **Session / conversation identity**: values that group a multi-turn interaction. Examples: A2A `context_id`, OpenAI
+  Responses `conversation` (`conv_*`) or response-chain continuation (`previous_response_id`).
+- **Task identity**: values that identify a protocol task and may affect future protocol calls. Example: A2A `task_id`.
+- **Message / response identity**: values that identify an output message or response. Examples: A2A `message_id` /
+  `artifact_id`, OpenAI Responses response id (`resp_*`).
+- **Continuation token**: a framework resume payload for in-progress work. It may contain the same underlying value as a
+  protocol id, such as A2A `task_id`, but it only exists when there is an unfinished operation to resume.
+
+These values should not automatically live in the same object just because they all help "continue" something. A value
+belongs in `AgentSession` only when it is needed to continue future calls across turns. A value that identifies one
+result belongs on the response or message. A value that resumes in-progress work belongs in a `ContinuationToken`.
+
+An `AgentSession` created for one agent is not expected to be guaranteed to work against another agent. When a session is
+used with an incompatible agent, protocol, or service, the framework should still help users understand what is wrong as
+early as possible, preferably before calling out to the remote service.
+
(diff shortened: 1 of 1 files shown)


giles17 (approved):

docs/decisions/0029-python-agent-session-identity.md
```diff
@@ -0,0 +1,321 @@
...
+contact: eavanvalkenburg
+date: 2026-06-19
+deciders: eavanvalkenburg, moonbox3, TaoChenOSU, chetantoshnival
+consulted: westey-m
+informed:
+---
+
+# Python identity lifetimes for sessions, tasks, and continuation
+
+## Context and Problem Statement
```
TaoChenOSU, line 12: For someone who is not so familiar with this, it's a bit difficult to visualize the problem. Is it possible to include some code that can show the gap we currently have?

docs/decisions/0029-python-agent-session-identity.md
```diff
@@ -0,0 +1,321 @@
...
+
+Keep the common `service_session_id` case as a plain string. When an agent/service needs more than one service-owned
+continuation value, allow `service_session_id` to be a typed structured value, such as a `TypedDict`. The main session ID
+used for `gen_ai.conversation.id` should still be extracted by the owning agent, not inferred by generic telemetry code.
+
+Examples:
+
+```python
+simple_session = AgentSession(
+    service_session_id="resp_123",
```
TaoChenOSU, line 170: Why is the session id a response ID?

docs/decisions/0029-python-agent-session-identity.md
```diff
@@ -0,0 +1,321 @@
...
+- Good, because it avoids new public fields and avoids a subclass requirement.
+- Good, because `AgentSession.state` already exists for provider/session state.
+- Neutral, because helper APIs can hide the raw dictionary access.
+- Bad, because stringly typed state is easier to corrupt and harder to validate.
+- Bad, because generic consumers need helper APIs anyway; directly reading nested dictionaries is not a good abstraction.
+- Bad, because users may accidentally overwrite or persist invalid protocol state.
+
+## Decision
+
+Chosen direction: **split identity by lifecycle**.
```
TaoChenOSU: Which option is this?

TaoChenOSU (commented):
This sounds like the gap that exists today is primarily in A2A. Is it necessary to introduce a new abstraction for me? My understanding of the problem is very limited. Please correct me if I am wrong.

eavanvalkenburg:
@TaoChenOSU Thanks — your review question on whether this needs a new abstraction is now reflected directly in the Decision section. The direction is to avoid a new top-level identity abstraction and keep continuation identity under `service_session_id` (Option B), while keeping run correlation in existing run/telemetry context. The immediate implementation gap is mostly A2A mapping clarity, but we keep the lifecycle split consistent across providers.

docs/decisions/0029-python-agent-session-identity.md
eavanvalkenburg: For OpenAI you can either continue a conversation using `resp_...` (passed to previous_response_id) or `conv_...` (passed to conversation) depending on how you start it, see https://developers.openai.com/api/docs/guides/conversation-state

docs/decisions/0029-python-agent-session-identity.md
eavanvalkenburg: During the investigation for this, I found more and more that this was often not done correctly, so wanted to clarify that this is the main decision driver for how to deal with ids/tokens/etc that are emitted by different protocols at different times. Added some more detail on this.

TaoChenOSU (approved):

PR #6631, opened 2026-06-19
.NET: [BREAKING] Support archive-type skills in AgentMcpSkillsSource

### LINKED ISSUE #6077: .NET: Support MCP skills of archive type
Support MCP skills of the `archive` type.

SergeyMenshykh:
### Motivation & Context

The MCP skills source previously discovered only `skill-md` index entries, where a skill's `SKILL.md` and sibling resources are fetched on demand from the server. The Agent Skills discovery spec also defines `archive` entries, where a skill is distributed as a single packaged archive (ZIP / TAR / gzip-compressed TAR) that unpacks into the skill's namespace. Without archive support, skills published in that format are silently unusable.

This change adds `archive`-type skill support so an MCP server can advertise packaged skills, which are downloaded, safely unpacked to a local directory, and served like file-based skills - while keeping the strict guarantee that MCP-delivered scripts are never executed.

  - `AgentMcpSkillsSource` now dispatches index entries to per-type loaders via a new `IMcpSkillEntryLoader` strategy: `SkillMdEntryLoader` (existing) and `ArchiveEntryLoader` (new).
  - `ArchiveEntryLoader` downloads, extracts, and serves each archive skill through an internal `AgentFileSkillsSource`, pruning directories the server no longer advertises.
  - `AgentMcpSkillArchiveExtractor` does the hardened unpacking: path-traversal guards, link skipping, and size/count limits.
  - New `AgentMcpSkillsSourceOptions` for the extraction directory, resource extensions/depth, and limits.

  - Additive only. `skill-md` discovery is unchanged; entry types with no registered loader are skipped as before. Archive-bundled scripts are surfaced as readable resources only - the inner file source is created with no allowed script extensions and no script runner, so they are never discovered as runnable scripts.

### Contribution Checklist

- [ ] **This is not a breaking change.** If it _is_ a breaking change, add the `breaking change` label (or add "[BREAKING]" to the title prefix, before or after any language prefix) - a workflow keeps the label and title prefix in sync automatically.

Closes: https://github.com/microsoft/agent-framework/issues/6077

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.Mcp/Skills/AgentMcpSkillsSourceOptions.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Mcp/Skills/AgentMcpSkillsSourceOptions.cs
@@ -1,0 +1,97 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using Microsoft.Shared.DiagnosticIds;
+
+namespace Microsoft.Agents.AI;
+
+/// <summary>
+/// Configuration options for <see cref="AgentMcpSkillsSource"/>.
+/// </summary>
+[Experimental(DiagnosticIds.Experiments.AgentsAIExperiments)]
+public sealed class AgentMcpSkillsSourceOptions
+{
+    /// <summary>
+    /// Gets or sets the base directory that archive-type skills are extracted to and served from.
+    /// </summary>
+    /// <remarks>
+    /// Archives are extracted beneath this directory as <c>{ArchiveSkillsDirectory}/{skill-name}/</c>.
+    /// When <see langword="null"/>, the source extracts to a per-instance unique location of
+    /// <c>{currentDirectory}/{guid}/{skill-name}/</c>, where the GUID is generated once per
+    /// <see cref="AgentMcpSkillsSource"/> instance so that multiple sources never overwrite one
+    /// another. Set this to a fixed value to get a predictable, reusable extraction location.
+    /// When set, each source must use its own unique directory: the source treats the directory as
+    /// exclusively its own and, on every discovery, prunes any sub-directory that the MCP server no
+    /// longer advertises. Pointing two sources at the same directory would therefore cause them to
+    /// delete each other's extracted skills.
+    /// </remarks>
+    public string? ArchiveSkillsDirectory { get; set; }
+
+    /// <summary>
+    /// Gets or sets the allowed file extensions for resources discovered in extracted archive-type skills.
+    /// </summary>
+    /// <remarks>
+    /// When <see langword="null"/>, defaults to <c>.md</c>, <c>.json</c>, <c>.yaml</c>, <c>.yml</c>,
+    /// <c>.csv</c>, <c>.xml</c>, and <c>.txt</c>.
+    /// </remarks>
+    public IEnumerable<string>? ArchiveResourceExtensions { get; set; }
+
+    /// <summary>
+    /// Gets or sets the maximum depth to search for resource files within each extracted archive-type
+    /// skill directory. A value of <c>1</c> searches only the skill root directory. A value of <c>2</c>
+    /// searches the root and one level of subdirectories.
+    /// </summary>
+    /// <remarks>
+    /// When <see langword="null"/>, the source uses the default depth of <c>2</c>.
+    /// </remarks>
+    public int? ArchiveResourceSearchDepth { get; set; }
+
+    /// <summary>
+    /// Gets or sets the maximum number of files that may be extracted from a single archive-type skill.
+    /// </summary>
+    /// <remarks>
+    /// Guards against excessive-file-count denial-of-service archives. When <see langword="null"/>, the
+    /// source uses a default of <c>20</c>, sized for a typical well-formed skill (a handful of files).
+    /// Raise this for archive-type skills that legitimately bundle many files. An archive that exceeds
+    /// the limit is skipped.
+    /// </remarks>
+    public int? ArchiveMaxFileCount { get; set; }
+
+    /// <summary>
+    /// Gets or sets the maximum size, in bytes, of a downloaded archive-type skill resource.
+    /// </summary>
+    /// <remarks>
+    /// Guards against archive resources that are too large to materialize safely. When
+    /// <see langword="null"/>, the source uses a default of <c>1 MB</c>, sized for a typical
+    /// well-formed skill archive. Raise this for archive-type skills that legitimately require
+    /// larger archive payloads. An archive that exceeds the limit is skipped.
+    /// </remarks>
+    public long? ArchiveMaxSizeBytes { get; set; }
+
+    /// <summary>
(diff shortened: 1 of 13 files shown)


dotnet/src/Microsoft.Agents.AI.Mcp/Skills/AgentSkillsProviderBuilderMcpExtensions.cs
```diff
@@ -15,12 +15,13 @@ public static class AgentSkillsProviderBuilderMcpExtensions
     /// </summary>
     /// <param name="builder">The builder to extend.</param>
     /// <param name="client">An MCP client connected to a server exposing Agent Skills resources.</param>
+    /// <param name="options">Optional options that control archive-distributed skill handling.</param>
     /// <returns>The builder instance for chaining.</returns>
-    public static AgentSkillsProviderBuilder UseMcpSkills(this AgentSkillsProviderBuilder builder, McpClient client)
+    public static AgentSkillsProviderBuilder UseMcpSkills(this AgentSkillsProviderBuilder builder, McpClient client, AgentMcpSkillsSourceOptions? options = null)
     {
         _ = Throw.IfNull(builder);
         _ = Throw.IfNull(client);
```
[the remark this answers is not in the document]
SergeyMenshykh: This is the change for experimental API which allows breaking changes.

dotnet/tests/Microsoft.Agents.AI.Mcp.UnitTests/Skills/AgentMcpSkillsSourceArchiveTests.cs
```diff
@@ -0,0 +1,693 @@
...
+        Assert.False(File.Exists(Path.Combine(this._extractionRoot, "skill", "escaped.txt")));
+        Assert.True(File.Exists(Path.Combine(target, "SKILL.md")));
+    }
+
+    [Fact]
+    public void Extract_ArchiveExceedsDefaultFileCount_Throws()
+    {
+        // Arrange - more files than the default cap (50).
+        var entries = Enumerable.Range(0, AgentMcpSkillArchiveExtractor.DefaultMaxFileCount + 1)
+            .Select(i => ($"file{i}.txt", "x"))
```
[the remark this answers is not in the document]
SergeyMenshykh: Fixed in 60d33d14c: removed the hardcoded '50' from the comment. It now reads 'more files than the default cap' without citing a number that can drift from the constant.

dotnet/tests/Microsoft.Agents.AI.Mcp.UnitTests/Skills/AgentMcpSkillsSourceArchiveTests.cs
```diff
@@ -0,0 +1,693 @@
...
+        {
+            if (Directory.Exists(this._extractionRoot))
+            {
+                Directory.Delete(this._extractionRoot, recursive: true);
+            }
+        }
+        catch (IOException)
+        {
+            // Best-effort cleanup.
+        }
```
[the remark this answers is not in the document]
SergeyMenshykh: Fixed in 60d33d14c: the catch now handles both IOException and UnauthorizedAccessException.

SergeyMenshykh:
> ArchiveEntryLoader.cs:81-84 delegates each extracted archive root to AgentFileSkillsSource, which recursively treats nested SKILL.md files as separate skills (AgentFileSkillsSource.cs:160-176). This conflicts with the archive loader's single-namespace contract (lines 17-21) and can cause one archive entry to surface multiple unadvertised skills.

The fix for this issue is outside the scope of this PR and will be addressed separately.

SergeyMenshykh:
> ArchiveEntryLoader.LoadAsync reconciles/prunes on-disk directories from the filtered archiveEntries set, so an archive skill that is still present in skill://index.json but temporarily fails validation (e.g., missing url) is treated as unadvertised and its extracted directory is deleted. That conflicts with the documented ArchiveSkillsDirectory contract in AgentMcpSkillsSourceOptions.cs:25-28, which says pruning applies only to subdirectories the MCP server no longer advertises.

An entry with a missing url or invalid name can't be materialized - keeping its stale directory around would mean serving potentially outdated content for a skill the server can no longer provide correctly. Pruning it and re-downloading when the entry is fixed again is the safer default. I'll update the doc comment to clarify that pruning applies to entries that aren't actionable, not just unadvertised ones.

westey-m (approved):

dotnet/src/Microsoft.Agents.AI.Mcp/Skills/Loaders/ArchiveEntryLoader.cs
```diff
@@ -0,0 +1,403 @@
...
+
+    /// <summary>
+    /// Reconciles the on-disk archive skill directories with the current set of advertised entries.
+    /// When no entries are advertised, all existing directories are removed. Otherwise, only stale
+    /// directories (not in the current advertised set) are pruned.
+    /// </summary>
+    private void ReconcileArchiveSkillDirectories(string? archiveSkillsDirectory, List<McpSkillIndexEntry> archiveEntries)
+    {
+        // Nothing to reconcile if the directory hasn't been established yet.
+        if (string.IsNullOrEmpty(archiveSkillsDirectory) || !Directory.Exists(archiveSkillsDirectory))
```
rogerbarreto, line 95: So many small conditions, ideally we should add UT coverage to all those specific pathways ensuring they are reached via the public API counterpart.

dotnet/src/Microsoft.Agents.AI.Mcp/Skills/AgentMcpSkillsSourceOptions.cs
```diff
@@ -0,0 +1,98 @@
...
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using Microsoft.Shared.DiagnosticIds;
+
+namespace Microsoft.Agents.AI;
+
+/// <summary>
+/// Configuration options for <see cref="AgentMcpSkillsSource"/>.
+/// </summary>
+[Experimental(DiagnosticIds.Experiments.AgentsAIExperiments)]
```
rogerbarreto: The other components introduced are not experimental, but this is, I see the package is `alpha` do we need to have experimental flag?

```suggestion
```

rogerbarreto (approved):

dotnet/src/Microsoft.Agents.AI.Mcp/Skills/Loaders/ArchiveEntryLoader.cs
SergeyMenshykh: - Added 5 tests covering the untested condition branches (missing name, invalid name chars, missing url, unsupported format, text-only blob) in bf1605c65.

dotnet/src/Microsoft.Agents.AI.Mcp/Skills/AgentMcpSkillsSourceOptions.cs
SergeyMenshykh: - Removed the [Experimental] attribute — alpha package version already signals experimental status.

rogerbarreto (approved):

PR #6642, opened 2026-06-20
Python: Fix ResponsesChannel session continuity and harden _result_to_text for workflows

Ashutosh0x:
### Motivation and Context

Fixes three issues flagged by automated DevFlow review on #6580:

1. **Session isolation key continuity** - In local (non-Foundry) hosting, the first turn's checkpoint/history data is stored under the initial `response_id`, but subsequent turns mint a *new* `response_id` as their isolation key. Checkpoint storage, `FileHistoryProvider`, and other session-scoped stores can't find data from the previous turn, silently dropping conversation state across turns.

2. **Non-streaming workflow result rendering** - `ResponsesChannel._handle` calls `_result_to_text(result.result)` on line 225. When the target is a workflow, `result.result` is a `WorkflowRunResult` whose `get_outputs()` may return an empty list. The current `_result_to_text` joins the empty list into `'''`, yielding a blank Responses envelope. Workflows that carry their final state on `get_final_state()` instead of through output executors are unrenderable.

3. **Local sample session bootstrap** - The `local_responses/app.py` and `local_responses_workflow/app.py` samples use `FileHistoryProvider` / checkpoint storage respectively, but the first turn's session is bootstrapped with a per-request `response_id` as isolation key.

### Description

**`_channel.py` - Session continuity fix:** When no externally-provided session exists, the fallback now uses `previous_response_id or response_id` as the isolation key.

**`_channel.py` - `_result_to_text` hardening:** When `get_outputs()` returns an empty list, falls back to `get_final_state()` before the generic `str()` cast.

**`test_channel.py` - New tests:** Added 3 tests for empty-output workflow results, final-state fallback, and plain-string rendering.

### Contribution Checklist


eavanvalkenburg:
The underlying code has not been merged, so this PR does not make sense yet, closing.

Issue #2988, opened 2025-12-20
.NET: Support dynamic agent resolution in AG-UI endpoints (MapAGUI with factory delegate)


mattbrailsford:
## Summary

The current `MapAGUI` extension method only accepts a pre-created `AIAgent` instance, which doesn't support scenarios where the agent needs to be resolved dynamically at request time (e.g., multi-tenant applications, agent selection by ID/alias from a database).

## Current API

```csharp
public static IEndpointConventionBuilder MapAGUI(
    this IEndpointRouteBuilder endpoints,
    string pattern,
    AIAgent aiAgent)  // Pre-created instance
```

## Problem

For dynamic agent resolution scenarios like:
- `POST /agents/{agentId}/stream` - resolve agent by ID from database
- Multi-tenant applications where agent configuration varies per tenant
- Agent marketplaces where users select from multiple agents

We cannot use `MapAGUI` because the agent must be resolved at request time, not at startup.

Additionally, the types needed to implement this ourselves are `internal`:
- `RunAgentInput` - cannot use for model binding
- `AGUIServerSentEventsResult` - cannot return from custom endpoint

This forces consumers to reimplement AG-UI request/response handling.

### Add factory delegate overload

```csharp
public static IEndpointConventionBuilder MapAGUI(
    this IEndpointRouteBuilder endpoints,
    string pattern,
    Func<HttpContext, CancellationToken, Task<AIAgent?>> agentFactory)
```

Usage:
```csharp
endpoints.MapAGUI("/agents/{agentId}/stream", async (httpContext, ct) =>
{
    var agentId = httpContext.GetRouteValue("agentId")?.ToString();
    return await agentResolver.ResolveAsync(agentId, ct);
});
```

## Environment

- Package: `Microsoft.Agents.AI.Hosting.AGUI.AspNetCore` v1.0.0-preview.251219.1
- Framework: .NET 10

TheEagleByte:
I've put together a PR for this which should address everything! Hopefully we can get it merged soon.

avogiatzis:
@TheEagleByte appreciate the work you've done on the PR, where there any developments on the architectural discussions? We're also looking to adopt AGUI but the static agent is an issue as we're looking to build agents dynamically, fetching configurations from the Database etc.

TheEagleByte:
[...]

PR #6643, opened 2026-06-20
.NET: ADR for unified dynamic agent resolution across AG-UI, Responses, and A2A

Ashutosh0x:
### Motivation and Context

This ADR addresses the cross-channel consistency requirement raised by @javiercn on #3162:

> *'Before we move forward with this change, we'd rather have a broader discussion to establish the pattern here, as this needs to be applied not only to AG-UI but to Open AI responses and A2A and we'd rather do everything simultaneously.'*

All three hosting channel endpoint builders (MapAGUI, MapOpenAIResponses, MapA2AHttpJson) currently resolve the AIAgent instance **once at endpoint registration time**. This singleton-capture prevents dynamic per-request agent resolution needed for multi-tenant platforms.

@halllo further identified that AgentSessionStore is also singleton-captured, breaking scoped/transient DI registrations.

### Description

Proposes an **IAgentResolver** interface in the hosting core plus factory-delegate overloads for all three channel types:

1. **IAgentResolver interface** - shared DI-registered resolver for route-based/claims-based agent selection
2. **Factory delegate overloads** - inline Func<HttpContext, CancellationToken, ValueTask<AIAgent?>> for quick prototyping
3. **Per-request session store resolution** - fixes the singleton-capture bug by resolving from HttpContext.RequestServices

The ADR analyzes 3 options (HttpContextRoutingAgent workaround, factory delegates only, IAgentResolver + factory dual API) and recommends the dual API approach.

### References
- #3162 - .NET: Support dynamic agent resolution in AG-UI endpoints
- #2988 - Original issue
- #2343 - Earlier PR by @halllo

### Contribution Checklist
- [ ] All unit tests pass (no code changes, ADR only)
- [ ] **Is this a breaking change?** No - design proposal only

### DIFF
--- a/docs/decisions/0029-unified-dynamic-agent-resolution.md
+++ b/docs/decisions/0029-unified-dynamic-agent-resolution.md
@@ -129,7 +129,7 @@ public static IServiceCollection AddAgentResolver<TResolver>(this IServiceCollec
     where TResolver : class, IAgentResolver
 {
     services.AddHttpContextAccessor();
-    services.AddSingleton<IAgentResolver, TResolver>();
+    services.AddScoped<IAgentResolver, TResolver>();
     return services;
 }
 ```
--- a/docs/decisions/0029-unified-dynamic-agent-resolution.md
+++ b/docs/decisions/0029-unified-dynamic-agent-resolution.md
@@ -1,0 +1,223 @@
+---
+status: proposed
+contact: Ashutosh0x
+date: 2026-06-20
+consulted: "@javiercn, @TheEagleByte, @halllo, @Davidlmkh"
+---
+
+# Unified Dynamic Agent Resolution Across AG-UI, OpenAI Responses, and A2A Endpoints
+
+## Context and Problem Statement
+
+All three hosting channel endpoint builders (`MapAGUI`, `MapOpenAIResponses`, `MapA2AHttpJson`) currently resolve the `AIAgent` instance and its `AgentSessionStore` **once at endpoint registration time** (app startup). This singleton-capture pattern prevents dynamic per-request agent resolution, which is required for multi-tenant agent platforms where route parameters (e.g., `/agents/{agentId}`) determine which agent handles each request.
+
+Community contributors have proposed factory-delegate overloads for AG-UI (#3162, #2343), but the maintainers (@javiercn) correctly noted that this pattern must be applied consistently across all three channel types simultaneously — not just AG-UI.
+
+### Current singleton-capture pattern (all three channels)
+
+```csharp
+// AG-UI — AGUIEndpointRouteBuilderExtensions.cs:105
+var agentSessionStore = endpoints.ServiceProvider.GetKeyedService<AgentSessionStore>(aiAgent.Name);
+var hostAgent = new AIHostAgent(aiAgent, agentSessionStore);
+
+// Responses — EndpointRouteBuilderExtensions.Responses.cs:71
+var executor = new AIAgentResponseExecutor(agent);
+
+// A2A — A2AEndpointRouteBuilderExtensions.cs:69
+var a2aServer = endpoints.ServiceProvider.GetKeyedService<A2AServer>(agentName);
+```
+
+### Pain points identified by the community
+
+1. **No dynamic routing** — Cannot serve different agents based on route parameters (#2988, #3162)
+2. **AgentSessionStore scoping broken** — Even scoped/transient session stores are captured as singletons at startup (@halllo, #3162 comment)
+3. **HttpContextRoutingAgent workaround is brittle** — Requires overriding 5+ methods, couples agent logic to HTTP transport, `SerializeSession()` was not async (@halllo, @TheEagleByte)
+4. **Forces per-agent infrastructure** — Without dynamic resolution, multi-agent hosts need separate endpoints per agent or fork the repo (@Davidlmkh)
+
+## Decision Drivers
+
+- **Consistency**: The same pattern must work across AG-UI, OpenAI Responses, and A2A
+- **ASP.NET Core idiom**: Follow established patterns (`AddDbContext<T>()`, `AddAuthentication<THandler>()`, minimal API delegates)
+- **Separation of concerns**: Agent resolution is a routing/infrastructure concern, not an agent behavior concern
+- **Backward compatibility**: Existing single-agent `MapXxx(pattern, agent)` overloads must continue to work
+- **Session store scoping**: Must support per-request session store resolution, not singleton capture
+
+## Considered Options
+
+### Option 1: HttpContextRoutingAgent (current workaround)
+
+A delegating `AIAgent` subclass that uses `IHttpContextAccessor` to resolve the real agent at runtime.
+
+**Pros**: Works today without framework changes
+**Cons**: Couples agents to HTTP, requires overriding all virtual methods, session store still singleton-captured, reduces cohesion
+
+### Option 2: Factory delegate overloads (PR #3162 approach)
+
+Add `MapXxx` overloads accepting `Func<HttpContext, CancellationToken, ValueTask<AIAgent?>>`:
+
+```csharp
+app.MapAGUI("/agents/{agentId}", async (context, ct) =>
+{
+    var agentId = context.GetRouteValue("agentId")?.ToString();
+    return await agentRepository.GetAgentByIdAsync(agentId, ct);
+});
+```
+
+**Pros**: Familiar minimal API pattern, maximum flexibility
+**Cons**: Channel-specific implementation needed for each `MapXxx`
+
+### Option 3: `IAgentResolver` interface with DI registration (recommended)
+
+Define a shared `IAgentResolver` interface in the hosting core:
+
+```csharp
+public interface IAgentResolver
+{
+    ValueTask<AIAgent?> ResolveAgentAsync(HttpContext context, CancellationToken cancellationToken = default);
+}
+```
+
+Registration:
+```csharp
+builder.Services.AddAgentResolver<RouteBasedAgentResolver>();
+```
+
+Endpoint mapping (no agent parameter):
+```csharp
+app.MapAGUI("/agents/{agentId}");
+app.MapOpenAIResponses("/agents/{agentId}/v1/responses");
+app.MapA2AHttpJson("/agents/{agentId}/a2a");
+```
+
+Each `MapXxx` overload resolves the agent per-request via `IAgentResolver` from DI.
+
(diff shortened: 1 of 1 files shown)


Ashutosh0x:
cc @javiercn — This ADR directly addresses the cross-channel consistency requirement you raised on #3162:

> *'Before we move forward with this change, we'd rather have a broader discussion to establish the pattern here, as this needs to be applied not only to AG-UI but to Open AI responses and A2A and we'd rather do everything simultaneously.'*

I analyzed all three endpoint builders (`MapAGUI`, `MapOpenAIResponses`, `MapA2AHttpJson`) and found the same singleton-capture pattern in each. The ADR proposes a dual API approach — `IAgentResolver` interface for DI-registered resolvers + factory delegate overloads for inline resolution — that works identically across all three channel types.

Would love your architectural feedback, particularly on:
1. Whether the `IAgentResolver` abstraction captures the right granularity
2. Whether per-request `AgentSessionStore` resolution from `HttpContext.RequestServices` is the right fix for the scoping bug @halllo identified
3. Whether this should live in `Microsoft.Agents.AI.Hosting` or a more specific package

Happy to iterate based on your guidance!

javiercn:
I don't feel strongly about what we do in this regard, provided we are consistent and that other folks like @DeagleGross and @ReubenBond @rogerbarreto or @westey-m are happy with the approach.

They might have different opinions on alternative ways in which this can be done/achieved.

That said, the IHttpContextAccessor pattern is not the end of the world, especially if the framework provided it. You can just wrap the agent, or the frameworks could simply put the HttpContext instance on the additional properties of the options, and you get to run your resolution logic within a delegating handler which you can already create with Use(...)

Ashutosh0x:
Thanks @javiercn — really appreciate the feedback and the direction!

You're right that the `IHttpContextAccessor` pattern is pragmatic and already familiar to ASP.NET Core developers. Let me think through both approaches:

### Option A: `IHttpContextAccessor` + `Use(...)` (your suggestion)

`csharp
app.MapAGUI(/agents/{agentId}, agent)
    .Use(async (context, next) =>
    {
        var httpContext = context.HttpContext;
        var agentId = httpContext.GetRouteValue(agentId);
        // resolve agent from DI or factory
        var agent = resolver.Resolve(agentId);
        // set on additional properties
        httpContext.Items[ResolvedAgent] = agent;
        await next(context);
    });
`

**Pros**: No new abstractions needed, uses existing middleware pipeline, developers already know this pattern.

**Cons**: The initial `agent` parameter is still required at registration time (even if ignored at runtime), and the resolution logic is duplicated across each endpoint unless wrapped in a shared middleware.

### Option B: `IAgentResolver` (ADR proposal)

**Pros**: Single abstraction, enforced consistency, testable via DI.

**Cons**: New interface to learn, more opinionated.

### My take

Given your preference for consistency without being overly prescriptive, I think the **best path forward** is:

1. **Add `HttpContext` to the options/additional properties** (as you suggested) — this is a minimal, non-breaking change that unblocks dynamic resolution for all three channels
2. **Document the `Use(...)` pattern** as the recommended approach for per-request agent selection
3. **Defer `IAgentResolver`** as a potential future convenience layer if the community gravitates toward it

I'll update the ADR to incorporate this direction. Happy to defer to @DeagleGross, @ReubenBond, @rogerbarreto, and @westey-m on which approach feels most natural for the .NET hosting surface.

Also addressing Copilot's review — will fix the `AddSingleton` → `AddScoped` registration for resolvers that depend on scoped services (DbContext, etc.).

Ashutosh0x:
Addressed Copilot's review — pushed `ee49e2b5`: changed `AddSingleton<IAgentResolver>` → `AddScoped<IAgentResolver>`.

This ensures resolvers that depend on scoped services (`DbContext`, per-request repositories) work correctly. Per-request resolution is the core intent of this ADR, so scoped lifetime is the natural fit.

Thanks for the catch @copilot!

Ashutosh0x:
## ADR Revised — Incorporating @javiercn's Feedback

Updated the ADR (commit `a04a40fe`) based on Javier's architectural guidance:

### What Changed

1. **Added Option 4** — `Use(...)` middleware + HttpContext on additional properties (Javier's suggestion). Full code examples showing both inline middleware and framework-level `AgentRequestOptions` approaches.

2. **Restructured Decision Outcome** as a phased approach:
   - **Phase 1** (minimal, non-breaking): Expose `HttpContext` on options, document `Use(...)` pattern, fix per-request `AgentSessionStore` resolution
   - **Phase 2** (if community demand): Factory delegate overloads + `IAgentResolver` convenience layer

3. **Added Maintainer Feedback section** with @javiercn's direct quote and key takeaways

4. **Expanded consulted list** per Javier's request: cc @DeagleGross @ReubenBond @rogerbarreto @westey-m

The revision prioritizes **minimal framework changes using established ASP.NET Core patterns** (middleware pipeline, `IHttpContextAccessor`) rather than introducing new abstractions.

Would love input from the team on whether Phase 1 captures the right scope. Happy to iterate further!

docs/decisions/0029-unified-dynamic-agent-resolution.md
```diff
@@ -0,0 +1,290 @@
...
+
+**Pros**: No new abstractions needed, uses existing ASP.NET Core middleware pipeline, developers already know this pattern, `IHttpContextAccessor` is well-established
+**Cons**: Initial `agent` parameter still required at registration (even if overridden at runtime), resolution logic duplicated unless wrapped in shared middleware
+
+### Option 3: `IAgentResolver` interface with DI registration (recommended)
+
+Define a shared `IAgentResolver` interface in the hosting core:
+
+```csharp
+public interface IAgentResolver
```
DeagleGross, line 116: I dont see this being much different from resolving `AIAgent` from the registered instances in the DI container. Default implementation of this will do basically same `GetRequiredService(agentname)` or similar. Is there any specific use-case where this is helping to solve some problem?

docs/decisions/0029-unified-dynamic-agent-resolution.md
```diff
@@ -0,0 +1,290 @@
...
+### Option 1: HttpContextRoutingAgent (current workaround)
+
+A delegating `AIAgent` subclass that uses `IHttpContextAccessor` to resolve the real agent at runtime.
+
+**Pros**: Works today without framework changes
+**Cons**: Couples agents to HTTP, requires overriding all virtual methods, session store still singleton-captured, reduces cohesion
+
+### Option 2: Factory delegate overloads (PR #3162 approach)
+
+Add `MapXxx` overloads accepting `Func<HttpContext, CancellationToken, ValueTask<AIAgent?>>`:
```
DeagleGross, line 57: This makes sense, but ideally for demo-purposes can you please share the sample where overloads today dont allow you to do X, and with the new overload you can do it? That will be a "proof" of why those overloads are important to exist

DeagleGross (commented):
Hey, thanks for filing and working on this! Recently not being an active contributor to agent-framework, but placed a couple of comments below. I am sure @westey-m can provide more insights here.

docs/decisions/0029-unified-dynamic-agent-resolution.md
Ashutosh0x: Great question @DeagleGross! You are right that a default IAgentResolver would basically do GetRequiredService. The key difference is **when and how** the agent name is determined:

**Today (DI only):** The agent is resolved **once at startup** in MapAGUI(pattern, agent). You can register multiple named agents, but there is no hook to select between them per-request based on route data.

**Where dynamic resolution adds value:**
1. **Route-based** - /agents/{agentId} where agentId comes from the URL at request time
2. **Claims-based** - Different tenants get different agents based on JWT claims
3. **Database-driven** - Agent definitions stored in a DB, not statically in DI

That said, per @javiercn feedback, I have revised the ADR to deprioritize IAgentResolver in favor of **Phase 1**: expose HttpContext on options and use Use(...) middleware. This achieves the same result without a new abstraction. IAgentResolver is now Phase 2 only if community demand warrants it. Does that align with your thinking?

docs/decisions/0029-unified-dynamic-agent-resolution.md
Ashutosh0x: Absolutely! Here is a concrete before/after:

**Today — you CANNOT do this:**
`csharp
// Goal: serve different agents based on URL
// e.g. GET /agents/weather/ag-ui  vs  GET /agents/search/ag-ui

// This does NOT work — agent is captured at startup:
app.MapAGUI(/agents/{agentId}, weatherAgent);
// ^ weatherAgent is always used, {agentId} route param is ignored
`

**Workaround today — brittle HttpContextRoutingAgent:**
`csharp
// You have to create a wrapper agent that overrides every method:
public class RoutingAgent : AIAgent
{
    private readonly IHttpContextAccessor _accessor;
    public override async Task<AgentResponse> RunAsync(...)
    {
        var ctx = _accessor.HttpContext;
        var id = ctx.GetRouteValue(agentId);
        var real = await _repo.GetAgent(id);
        return await real.RunAsync(...); // delegate everything
    }
    // Must also override: RunStreamingAsync, CreateSessionAsync,
    // GetSessionAsync, SerializeSession (not async!), etc.
}
`

**With factory delegate overload — clean and simple:**
`csharp
app.MapAGUI(/agents/{agentId}, async (context, ct) =>
{
    var agentId = context.GetRouteValue(agentId)?.ToString();
    return await agentRepo.GetAgentAsync(agentId, ct);
});
// Session store also resolved per-request automatically
`

The core problem: today the agent is a **startup-time constant**, but multi-tenant platforms need it to be a **request-time variable**. The workaround requires duplicating the entire AIAgent surface area.

Ashutosh0x:
## Phase 1 Implementation PR Created

I've opened **#6659** with the actual implementation of the factory delegate overload for MapAGUI, following the approach discussed here.

**ADR (this PR)** → **Implementation (#6659)**

The implementation resolves agents and session stores per-request from `HttpContext.RequestServices` instead of capturing at startup, exactly as proposed in the revised Phase 1 approach.

javiercn:
This ADR is aligned with the cross-host direction we want. We have opened #7733 as the canonical issue for the implementation requirement.

The investigation confirmed that AG-UI, OpenAI Chat Completions, OpenAI Responses, and A2A all currently capture hosted agents or protocol services from the root provider. `AddAIAgent` already supports scoped/transient lifetimes, but the protocol integrations do not honor them.

#7733 also expands the requirement to include a consistent request-time hosted-agent selector across all hosting extensions. The selector should choose a stable hosted registration/identity before the agent, session store, and protocol services are resolved from `HttpContext.RequestServices`.

Please use #7733 for the canonical implementation discussion. A separate ADR can be authored from that work if the final solution requires one.

PR #6360, opened 2026-06-05
Python: Prefer explicit AG-UI resume payloads

gezw:
## Summary

- make explicit AG-UI resume payloads take precedence over message-derived responses
- preserve message-derived responses when no explicit resume payload is present
- add workflow-run regression coverage for stale message approvals

## Details

`run_workflow_stream()` can receive resume responses from both the explicit `resume` payload and from prior AG-UI messages. When both sources contain an entry for the same pending request id, the explicit resume payload should be treated as the caller's current decision and should not be overwritten by stale message history.

This change centralizes the merge behavior so message-derived responses are used as a fallback and explicit resume values win on conflicts.

## Validation

From `python/`:

- `uv run --group dev pytest -q --disable-warnings packages/ag-ui/tests/ag_ui/test_workflow_run.py`
- `uv run --group dev ruff check packages/ag-ui/agent_framework_ag_ui/_workflow_run.py packages/ag-ui/tests/ag_ui/test_workflow_run.py`
- `uv run --group dev ruff format --check packages/ag-ui/agent_framework_ag_ui/_workflow_run.py packages/ag-ui/tests/ag_ui/test_workflow_run.py`
- `uv run --group dev python -m py_compile packages/ag-ui/agent_framework_ag_ui/_workflow_run.py packages/ag-ui/tests/ag_ui/test_workflow_run.py`
- `git diff --check`

PR #6646, opened 2026-06-20
Python: add checkpointing support to AgentFrameworkWorkflow.run() in agent-framework-ag-ui

### LINKED ISSUE #6632: Python: [Feature]: Add Checkpointing Support to AgentFrameworkWorkflow.run() in agent-framework-ag-ui
## Summary

The `AgentFrameworkWorkflow` implementation provided by `agent-framework-ag-ui` currently exposes a `run()` method that only accepts a `RunAgentInput`. This differs from the core `agent-framework` workflow implementations, which support checkpointing and workflow state persistence.

This limitation prevents AG-UI-based workflows from taking advantage of one of the framework's key capabilities: resumable execution through checkpointing.

## Problem

Today, the AG-UI adapter provides its own `AgentFrameworkWorkflow` abstraction, but its `run()` API does not expose any mechanism for:

* Creating checkpoints during execution
* Restoring workflow state from a checkpoint
* Resuming interrupted executions
* Integrating with existing checkpoint storage providers

As a result, applications built on `agent-framework-ag-ui` lose functionality that is otherwise available when using standard workflows in `agent-framework`.

This creates an inconsistency in the developer experience and makes AG-UI workflows less suitable for long-running or fault-tolerant agent scenarios.

## Proposed Solution

Extend the `AgentFrameworkWorkflow.run()` API in `agent-framework-ag-ui` to support checkpointing in a manner consistent with the core `agent-framework` workflow APIs.

Potential approaches could include:

### Option 1: Support Workflow Context

Allow `run()` to accept the same workflow context or execution options used by the core framework, enabling checkpoint providers to be configured and utilized during execution.

### Option 2: Expose Checkpoint Parameters

Add optional checkpoint-related arguments to `run()`, such as:

```python
await workflow.run(
    input=run_agent_input,
    checkpoint_storage=checkpoint_storage,
    checkpoint_id=checkpoint_id,
)
```

### Option 3: Align with Core Workflow API

Refactor the AG-UI workflow implementation to use the same execution model as the standard framework workflows so that checkpointing works without introducing AG-UI-specific APIs.

## Benefits

* Feature parity between `agent-framework` and `agent-framework-ag-ui`
* Support for long-running agent executions
* Improved fault tolerance and recovery
* Easier integration with persistent storage backends (PostgreSQL, Redis, etc.)
* Consistent developer experience across workflow implementations

## Use Case

Many production deployments require durable execution. For example, an AG-UI workflow may orchestrate multiple agents, tool calls, and user interactions over an extended period of time.

If the process crashes or is redeployed, the workflow currently must be restarted from the beginning. With checkpointing support, execution could resume from the most recent checkpoint, matching the capabilities available in the core framework.

### Code Sample

```markdown

```

### Language/SDK

Python

vaibhav-patel:
## Summary

Fixes #6632. Adds workflow checkpointing to the ag-ui adapter's
`AgentFrameworkWorkflow.run()`, bringing it to parity with the core
`agent_framework.Workflow.run()` checkpointing API.

`run()` now accepts two optional keyword arguments:
- `checkpoint_storage: CheckpointStorage | None` — enables checkpoint creation at each
  superstep (forwarded to the core workflow).
- `checkpoint_id: str | None` — resumes a run from a persisted checkpoint.

This delegates to the existing core capability rather than reinventing it, so the public
surface mirrors `Workflow.run(checkpoint_storage=..., checkpoint_id=...)`.

## Details

- Resume honors the core API contract: `message` is never passed alongside
  `checkpoint_id`; incoming messages are forwarded only as request-info `responses`
  (so `responses` + `checkpoint_id` is a restore-then-send).
- Params may also be supplied via the `input_data` keys `__ag_ui_checkpoint_storage` /
  `__ag_ui_checkpoint_id`, letting the FastAPI endpoint (which calls `run(input_data)`
  positionally) opt in without a signature change; explicit keyword args take precedence.
- Workflow checkpointing is independent from AG-UI thread-snapshot persistence; resume
  bypasses snapshot-hydration early-returns to reach the core restore path.

## Backward compatibility

The existing `run(RunAgentInput)` / `run(input_data)` signature is preserved — the new
params are optional and default to `None`, and the non-checkpoint path still calls
`run_workflow_stream(input_data, workflow)` with its original two-argument convention.

## Tests

New focused tests cover checkpoint creation, resume-from-checkpoint, input-data-keyed
params, and the unchanged default path. Full ag-ui suite green (824 passed); ruff
lint/format clean; no new pyright errors.


python/packages/ag-ui/tests/ag_ui/test_workflow_agent.py
```diff
@@ -7,13 +7,25 @@
...
-from agent_framework import Workflow, WorkflowBuilder, WorkflowContext, executor
+from agent_framework import (
+    InMemoryCheckpointStorage,
+    Workflow,
+    WorkflowBuilder,
+    WorkflowContext,
+    executor,
+    handler,
+)
+from agent_framework._workflows._executor import Executor
```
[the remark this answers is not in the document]
vaibhav-patel: Thanks — switched to the public `from agent_framework import Executor` in b7f3384.

python/packages/ag-ui/agent_framework_ag_ui/_workflow.py
```diff
@@ -288,7 +323,23 @@ async def run(self, input_data: dict[str, Any]) -> AsyncGenerator[BaseEvent]:
...
+        # FastAPI endpoint (which calls ``run(input_data)`` positionally) can still opt
+        # into checkpointing without changing its call site.
+        if checkpoint_id is None:
+            checkpoint_id = cast(str | None, input_data.get(_CHECKPOINT_ID_INPUT_KEY))
+        if checkpoint_storage is None:
+            checkpoint_storage = cast(CheckpointStorage | None, input_data.get(_CHECKPOINT_STORAGE_INPUT_KEY))
+
+        # A checkpoint resume legitimately carries no new messages; it must reach the
+        # core workflow's restore path rather than replaying a stored thread snapshot.
+        if (
```
moonbox3: Could checkpoint resumes with snapshot persistence also preserve the stored AG-UI thread history? This new `checkpoint_id is None` gate correctly avoids hydration, but a checkpoint-only resume with no new messages then continues into the snapshot-save path; `_reconstruct_messages_from_thread_snapshot()` leaves `raw_messages` empty, and the builder is only seeded with stored history for AG-UI `resume_payload`, so `snapshot_store.save(... snapshot_builder.build())` can replace the existing thread snapshot with only the resumed output. I reproduced this with a stored two-message snapshot: after a successful checkpoint resume, the saved snapshot contained only the new assistant output, silently dropping the prior replayable transcript.

python/packages/ag-ui/agent_framework_ag_ui/_workflow_run.py
```diff
@@ -640,11 +668,27 @@ def _drain_open_message() -> list[TextMessageEndEvent]:
...
     try:
-        if responses:
-            event_stream = workflow.run(responses=responses, stream=True, **fwd_kwargs)
+        if checkpoint_id is not None:
+            # Resume from a checkpoint. ``message`` is mutually exclusive with
+            # ``checkpoint_id`` in the core API, so incoming messages are only
+            # forwarded as request-info responses (``responses``), never as a new
+            # start-executor message. ``responses`` + ``checkpoint_id`` performs a
+            # restore-then-send in a single call.
+            run_kwargs: dict[str, Any] = {"checkpoint_id": checkpoint_id, **checkpoint_kwargs}
```
moonbox3: Could checkpoint resumes coerce AG-UI resume responses after the checkpoint has restored its pending requests? `responses` are coerced before restore using `pending_before_run`, which is empty for a cold checkpoint restore, so this new path forwards raw JSON resume payloads into core. A JSON `function_approval_response` that succeeds through the existing non-checkpoint AG-UI path now fails after restore with `Response type mismatch ... expected Content, got dict`, which means checkpointed approval workflows cannot be resumed through AG-UI.

moonbox3:
@vaibhav-patel please also look at the failing code quality CI/CD checks thanks

python/packages/ag-ui/agent_framework_ag_ui/_workflow.py
vaibhav-patel: Good catch — fixed. On a checkpoint-only resume `raw_messages` is empty, so `_reconstruct_messages_from_thread_snapshot()` returns the empty incoming list and the snapshot builder was seeded with nothing, letting `snapshot_store.save()` overwrite the stored transcript with only the new output. I broadened the builder-seeding condition to also seed from the stored history when `checkpoint_id is not None and not raw_messages` (it previously only handled the `resume_payload` case), so the saved snapshot after a checkpoint resume now contains the prior replayable messages plus the newly produced output. Added `test_workflow_checkpoint_only_resume_preserves_thread_snapshot`, which fails on the old code and passes now.

python/packages/ag-ui/agent_framework_ag_ui/_workflow_run.py
vaibhav-patel: Fixed. On a cold checkpoint restore the live workflow instance has no pending requests yet, so `pending_before_run` is empty and the resume responses were forwarded to core uncoerced — hence `Response type mismatch … expected Content, got dict`. I now restore the checkpoint up front (best-effort, idempotent — `workflow.run(checkpoint_id=…)` restores again) so `pending_before_run` reflects the post-restore pending set and the resume contract + coercion run against it, exactly like the non-checkpoint path. I gated this on `resume_payload is not None` so a pure checkpoint restore still surfaces its pending interrupts rather than tripping the `WORKFLOW_RESUME_REQUIRED` contract from #6360 (which I rebased onto). Added `test_workflow_run_resume_content_response_after_checkpoint_restore` — a fresh workflow instance resuming a JSON `function_approval_response` from a checkpoint — which fails on the old code and passes now.

One note: the pre-restore reaches into `workflow._runner.restore_from_checkpoint(...)` (private, guarded/best-effort, consistent with the existing `_pending_request_events` helper) — happy to switch to a public "restore to inspect pending" entry point if you'd prefer one.

vaibhav-patel:
@moonbox3 CI is green now — fixed the `I001` import-sort in `test_workflow_agent.py`, and the "Invalid override of method `run`" failures (mypy/pyrefly/ty/zuban) by updating the three test-double `run()` overrides in `test_endpoint.py` to accept `**kwargs` so they're LSP-compatible with the new keyword-only parent signature. `ruff check`, `ruff format --check`, all five typing checkers, and the full ag-ui suite (892 passed) pass locally.

Thanks for the thorough review — the two resume bugs you reproduced are fixed too (details on the inline threads).

python/packages/ag-ui/agent_framework_ag_ui/_workflow_run.py
```diff
@@ -146,6 +154,32 @@ async def _pending_request_events(workflow: Workflow) -> dict[str, Any]:
...
+    so this best-effort restore hydrates them first. It is idempotent: the subsequent
+    ``workflow.run(checkpoint_id=...)`` restores again from the same checkpoint. Any
+    failure is swallowed here and surfaced by that core run instead.
+    """
+    runner = getattr(workflow, "_runner", None)
+    restore_from_checkpoint = getattr(runner, "restore_from_checkpoint", None)
+    if restore_from_checkpoint is None:
+        return
+    try:
+        await restore_from_checkpoint(checkpoint_id, checkpoint_storage)
```
moonbox3: Could we avoid restoring the checkpoint twice to inspect its pending requests? This pre-restore invokes every executor's `on_checkpoint_restore()` hook, then `workflow.run(checkpoint_id=..., responses=...)` restores again and invokes the hooks a second time. Custom restore hooks are not required to be idempotent, so this can duplicate restoration work or fail workflows that correctly expect one restore per resume. Could the response coercion be handled through a core/public restore-and-send path without executing user restore hooks twice?

vaibhav-patel: Good catch — you're right that the pre-restore made `on_checkpoint_restore()` fire twice, and custom hooks aren't guaranteed idempotent.

I've reworked it so the resume no longer restores twice. Instead of restoring the checkpoint into the live workflow to surface its pending requests, the ag-ui layer now loads the persisted `WorkflowCheckpoint` (via the runtime `CheckpointStorage`, falling back to the workflow's build-time context storage through `RunnerContext.load_checkpoint`, which is explicitly documented as not mutating context state) and reads `pending_request_info_events` directly. That gives the resume-contract check and response coercion the same post-restore pending set without touching workflow state or running any executor hook. `workflow.run(checkpoint_id=..., responses=...)` then performs the one real restore, so the restore — and every custom `on_checkpoint_restore` — runs exactly once per resume.

Added a regression test (`test_workflow_run_resume_restores_checkpoint_exactly_once`) with an executor whose `on_checkpoint_restore` increments a counter; it asserts the count is 1 after a checkpointed resume (fails at 2 on the previous double-restore code). The existing `test_workflow_run_resume_content_response_after_checkpoint_restore` and the rest of the ag-ui suite stay green, along with ruff and all five type checkers.

python/packages/ag-ui/agent_framework_ag_ui/_workflow_run.py
```diff
@@ -146,6 +155,59 @@ async def _pending_request_events(workflow: Workflow) -> dict[str, Any]:
...
+    runtime override) takes precedence; otherwise the workflow's build-time checkpoint
+    storage, exposed through its runner context, is used. Loading only reads the persisted
+    ``WorkflowCheckpoint`` -- it never mutates workflow state and never invokes executor
+    ``on_checkpoint_restore`` hooks. Any failure is swallowed here and surfaced by the
+    core ``workflow.run(checkpoint_id=...)`` call instead.
+    """
+    try:
+        if checkpoint_storage is not None:
+            return await checkpoint_storage.load(checkpoint_id)
+        runner_context = getattr(workflow, "_runner_context", None)
```
TaoChenOSU: This is a very hack approach and may break anytime.

python/packages/ag-ui/agent_framework_ag_ui/_workflow_run.py
```diff
@@ -856,11 +954,27 @@ def _drain_open_message() -> list[TextMessageEndEvent]:
...
+            # restore-then-send in a single call.
+            run_kwargs: dict[str, Any] = {"checkpoint_id": checkpoint_id, **checkpoint_kwargs}
+            if responses:
+                run_kwargs["responses"] = responses
+            event_stream = workflow.run(stream=True, **run_kwargs, **fwd_kwargs)
+        elif responses:
+            event_stream = workflow.run(responses=responses, stream=True, **checkpoint_kwargs, **fwd_kwargs)
         else:
-            event_stream = workflow.run(message=messages, stream=True, **fwd_kwargs)
+            event_stream = workflow.run(message=messages, stream=True, **checkpoint_kwargs, **fwd_kwargs)
```
TaoChenOSU: You can simply forward everything to the run method, including None. The method will raise if conflicting arguments are provided.

moonbox3:
Any follow up here, @vaibhav-patel? Thank you.

moonbox3:
Helping get this across the finish line.

moonbox3 (approved):

python/packages/ag-ui/agent_framework_ag_ui/_endpoint.py
```diff
@@ -136,6 +141,16 @@ def add_agent_framework_fastapi_endpoint(
     else:
         raise TypeError("agent must be SupportsAgentRun, Workflow, AgentFrameworkAgent, or AgentFrameworkWorkflow.")
 
+    if checkpoint_storage is not None:
+        if not isinstance(protocol_runner, AgentFrameworkWorkflow):
+            raise ValueError("checkpoint_storage is only supported when the endpoint exposes a workflow.")
+        if (
+            protocol_runner.checkpoint_storage is not None
+            and protocol_runner.checkpoint_storage is not checkpoint_storage
```
TaoChenOSU, line 151: nit: can this be simplified to `protocol_runner.checkpoint_storage is not checkpoint_storage`? And why do we need to assign `checkpoint_storage` again here when it was already assigned during creation?

moonbox3: good catch - updated

TaoChenOSU (approved):

Issue #4869, opened 2026-03-23
.NET: [Bug]: AGUIChatClient sets ConversationId in response despite being stateless


ArturDorochowicz:
### Description

I have an AGUIChatClient working against Agent Framework's MapAGUI endpoint. 

I see that the response from `GetResponseAsync` has `ConversationId` set. Reading https://learn.microsoft.com/en-us/dotnet/ai/ichatclient#stateless-vs-stateful-clients tells me that it signifies a stateful service, but if I pass it in ChatOptions to a subsequent call with just the new input, the model's response indicates it has no previous context.

This (?) also seems to confuse AIAgent. If I do AsAIAgent on AGUIChatClient, then CreateSessionAsync and pass the session to multiple RunAsync calls with just the new input, the model's response indicates it has no previous context.

     

### Code Sample

```markdown
using var client = new AGUIChatClient(httpClient, "endpoint");
var agent = client.AsAIAgent();

var session = await agent.CreateSessionAsync();
var r1 = await agent.RunAsync("What is the area of France?", session);

var confused = await agent.RunAsync("What is its capital?", session);
// response indicates that model has no previous context
```

### Error Messages / Stack Traces

```markdown

```

### Package Versions

Microsoft.Agents.AI.AGUI 1.0.0-preview.260311.1

### .NET Version

.net 10

danroth27:
@chetantoshniwal  It looks like this is no longer an issue with AGUI.Client, so I think this can be closed.

javiercn:
This is resolved by the current `AGUI.Client` implementation.

The former in-tree client surfaced the AG-UI `threadId` as MEAI `ConversationId`. In MEAI, a non-null `ConversationId` indicates that the remote service owns conversation history. `ChatClientAgent` therefore stopped managing and sending local history, even though the AG-UI client was stateless.

The current `AGUI.Client` intentionally leaves `ConversationId` null. When wrapped with `AsAIAgent`, the preserved `AgentSession` continues managing local history:

```csharp
using var client = new AGUIChatClient(new(httpClient, "/ag-ui"));
AIAgent agent = client.AsAIAgent(name: "agui-client");

AgentSession session = await agent.CreateSessionAsync();

await agent.RunAsync("What is the area of France?", session);
await agent.RunAsync("What is its capital?", session);
```

We verified that the second turn receives the accumulated history while `ConversationId` remains null.

[...]

PR #6653, opened 2026-06-21
.NET: Replace MAF AG-UI abstractions with the AG-UI C# SDK abstractions

javiercn:
> [!NOTE]
> This PR consumes the external AG-UI .NET SDK packages (`AGUI.*`) at `0.0.1`, now published on [NuGet.org](https://www.nuget.org/packages/AGUI.Abstractions) (they ship from the AG-UI repo — see [ag-ui-protocol/ag-ui#1963](https://github.com/ag-ui-protocol/ag-ui/pull/1963)).

## Summary

Replaces Microsoft Agent Framework's in-tree AG-UI abstractions (`Microsoft.Agents.AI.AGUI`) with the external **AG-UI C# SDK** packages, so MAF consumes the protocol implementation maintained by the AG-UI team instead of carrying its own parallel copy.

The framework keeps its own thin ASP.NET hosting glue; only the *protocol abstractions* move out.

**Net change:** 112 files, **+523 / −10,352** — almost entirely deletions of the vendored protocol code and its tests.

## What changes

**Removed** (the vendored AG-UI implementation):
- `src/Microsoft.Agents.AI.AGUI/` — the internal `AGUIChatClient`, event/message types, JSON converters, `AGUIJsonSerializerContext`, role/tool/event constants, and the `ChatResponseUpdate ↔ AG-UI` extensions.
- `tests/Microsoft.Agents.AI.AGUI.UnitTests/` — the unit tests for that vendored copy (now covered upstream in the AG-UI SDK's own test suites).

**Added** package references (`Directory.Packages.props`, restored from NuGet):
- `AGUI.Abstractions`, `AGUI.Formatting`, `AGUI.Protobuf`, `AGUI.Client`, `AGUI.Server` — all `0.0.1`.

**Kept and re-pointed** at the external primitives:
- `Microsoft.Agents.AI.Hosting.AGUI.AspNetCore` retains its own ASP.NET surface but now layers it over the framework-agnostic `AGUI.Server` primitives — `RunAgentInput.ToChatRequestContext(...)` and `ChatResponseUpdateAGUIExtensions.AsAGUIEventStreamAsync(...)`.
- **Renamed the hosting entry points** `AddAGUI` → `AddAGUIServer` and `MapAGUI` → `MapAGUIServer` (matching the sibling `AddA2AServer` convention), so the bare `AddAGUI`/`MapAGUI` names stay free for the AG-UI SDK to use itself in the future. The package's public surface is now exactly these two methods.
- **Simplified the JSON wiring:** deleted the `AGUIJsonSerializerOptions` static holder; the type-info resolvers are now registered inside `AddAGUIServer` via an internal `IConfigureOptions<JsonOptions>` (`ConfigureAGUIJsonOptions`, wired with `TryAddEnumerable`), with resolver order matching the A2A hosting package (agent-abstractions first, then the AG-UI wire context).
- The samples under `samples/02-agents/AGUI/*` and `samples/05-end-to-end/AGUI*` and the hosting integration/unit tests are migrated to the external API.

## Why

- **Single source of truth.** The AG-UI protocol (events, messages, tools, interrupts, state, multimodal, SSE + protobuf transports) is maintained once, in the AG-UI C# SDK, and stays wire-compatible with the TypeScript/Python SDKs. MAF no longer has to track protocol changes in a second implementation.
- **Less surface to own.** ~10k lines of protocol code and tests leave the framework; MAF keeps only the ASP.NET hosting integration that is genuinely framework-specific.
- **Same model.** The SDK is built on `Microsoft.Extensions.AI` — `IChatClient` remains the only integration point on both the client and server sides, so the programming model is unchanged.

## Migration guide (for existing consumers of `Microsoft.Agents.AI.AGUI`)

**Hosting rename.** If you host an agent over AG-UI, the two hosting entry points are renamed
(the old names are gone):

```csharp
// Before
builder.Services.AddAGUI();
app.MapAGUI("/", agent);

// After
builder.Services.AddAGUIServer();
app.MapAGUIServer("/", agent);
```

If you used the **abstractions** directly, three more things change:

**1. Namespaces** — the single `Microsoft.Agents.AI.AGUI` namespace splits along the SDK's package boundaries:

| Was | Now |
|---|---|
| `Microsoft.Agents.AI.AGUI` (client) | `AGUI.Client` |
| `Microsoft.Agents.AI.AGUI` (server adapters) | `AGUI.Server` |
| `Microsoft.Agents.AI.AGUI` (events/messages/tools) | `AGUI.Abstractions` |

**2. `AGUIChatClient` construction** — the positional constructor becomes options-based:

```csharp
// Before
using Microsoft.Agents.AI.AGUI;
var chatClient = new AGUIChatClient(
    httpClient,
    serverUrl,
    jsonSerializerOptions: AGUIClientSerializerContext.Default.Options);

// After
using AGUI.Client;
var chatClient = new AGUIChatClient(new(httpClient, serverUrl)
{
    JsonSerializerOptions = AGUIClientSerializerContext.Default.Options,
});
```

**3. Recovering the originating AG-UI input on the server** — read it back from `ChatOptions` via the SDK extension:

```csharp
using AGUI.Abstractions;
using AGUI.Server;

if (!chatOptions.TryGetRunAgentInput(out RunAgentInput? agentInput))
{
    // not an AG-UI-originated request
}
```

**4. Package references** — drop the project/package reference to `Microsoft.Agents.AI.AGUI` and add the `AGUI.*` packages you use (`AGUI.Client` for clients, `AGUI.Server` + `AGUI.Abstractions` for server/hosting, plus `AGUI.Protobuf` if you opt into the protobuf transport).

## Notes for reviewers

- The endpoint shape is unchanged: `MapPost` → `input.ToChatRequestContext(...)` → `chatClient.GetStreamingResponseAsync(...)` → `.AsAGUIEventStreamAsync(ctx, ct)` → negotiated SSE/protobuf result.
- The public `Microsoft.Agents.AI.Hosting.AGUI.AspNetCore` surface is now just `AddAGUIServer` + `MapAGUIServer` (renamed from `AddAGUI`/`MapAGUI`; the SSE result and JSON configurator are internal).
- The `AGUI.* 0.0.1` packages are published on NuGet.org, so the build restores cleanly; hosting unit (9/9) and integration (32/32) tests pass against them.


javiercn:
## Review comment dispositions

Thanks for the thorough automated review. Triaged below, with fixes pushed in the latest revision.

### `AGUIServerSentEventsResult.cs` — flush `Utf8JsonWriter` after serialization
**Dismissed (no change).** `JsonSerializer.Serialize(Utf8JsonWriter, value, JsonTypeInfo)` flushes the writer at the end of the call, so the bytes are committed to the `IBufferWriter` before `SseFormatter` frames them — no manual `Flush()` is needed and small payloads are not truncated. This path is also unchanged by the migration and is covered by `AGUIServerSentEventsResultTests`.

### `AGUIServerSentEventsResult.cs` — `ex.Message` forwarded to the client
**Fixed.** The error event now sends a generic client-facing message (`"An error occurred while streaming the agent response."`); the full exception is still recorded server-side via `LogStreamingError`. (Polyfill-only path, `#if !NET10_0_OR_GREATER`.)

### `AGUIEndpointRouteBuilderExtensions.cs` — invalid/empty body 400 path untested
**Dismissed (no change).** The `[FromBody] RunAgentInput?` binding + `null → 400` behavior is ASP.NET Core minimal-API model binding, not logic owned by this glue, so a test here would be asserting framework behavior. The endpoint's own logic (context conversion, session resolution, streaming) is covered by the integration suite.

### Design Approach — `ConversationId` absence and `RawRepresentationFactory` continuation

These reflect an **intentional** AG-UI SDK design, not a regression (microsoft/agent-framework#4869):

- **`AGUIChatClient` is stateless and deliberately never surfaces a `ConversationId`.** In M.E.AI semantics a returned `ConversationId` means "the service is storing this conversation; send only deltas next turn" — but the AG-UI server is stateless, so advertising one would truncate history. This mirrors the reference TypeScript client, which owns a stable client-side thread id, always sends full history, and never adopts a thread id from the response. So `BasicStreamingTests` asserting `ConversationId == null` is codifying the correct contract.
- **Thread/run identity is carried on the wire events' raw representation** (`RunStartedEvent.ThreadId` / `RunFinishedEvent.ThreadId`, run id via `ResponseId`), and **continuation/branching sets `RunAgentInput.ThreadId`/`ParentRunId` through `ChatOptions.RawRepresentationFactory`** — the AG-UI-native way to set wire-level fields. That is the public, intended pattern (not a bypass), so `SessionPersistenceTests` is exercising the real continuation path callers use.

**Docs/samples updated to match** (the contract mismatch the review correctly flagged): the surviving AG-UI samples and READMEs no longer present `ConversationId` as thread identity — they read the thread id from the `RUN_STARTED` raw representation and document the stateless model:
- `samples/02-agents/AGUI/Step01/Step02/Step03/Step05/Client/Program.cs` and `samples/02-agents/AGUI/README.md`
- `samples/05-end-to-end/AGUIClientServer/AGUIClient/Program.cs` and its `README.md`
- `AGUIWebChat`'s Blazor client is left as-is: it already uses the idiomatic M.E.AI stateless/stateful pattern (`statefulMessageCount = ConversationId is not null ? … : 0`), which naturally sends full history against the always-null AG-UI client.

The non-blocking note about the serializer-options test is acknowledged; it can be tightened in a follow-up since it doesn't affect the migration's correctness.

javiercn:
## Follow-up: renamed the public hosting API + slimmed the JSON wiring

To avoid squatting on `AddAGUI` / `MapAGUI` — names the AG-UI .NET SDK should be free to use itself in the future — and to match the in-repo A2A convention (`AddA2AServer`, `MapA2A*`), the two public methods are renamed:

| Before | After |
|---|---|
| `services.AddAGUI()` | `services.AddAGUIServer()` |
| `endpoints.MapAGUI(...)` | `endpoints.MapAGUIServer(...)` |

The package's entire public surface is now exactly these two methods. Supporting changes:

- Deleted the `AGUIJsonSerializerOptions` static holder. The JSON type-info resolvers are now registered inside `AddAGUIServer` via an internal `IConfigureOptions<JsonOptions>` (`ConfigureAGUIJsonOptions`), wired with `TryAddEnumerable` (the idiomatic, idempotent options pattern). Resolver order is preserved (AG-UI wire context first, then Agent Framework abstractions).
- Renamed the extension class to `AGUIServerServiceCollectionExtensions` (and its file) to match the A2A naming.
- Updated all call sites: the AG-UI sample servers, the `02-agents/AGUI` + `AGUIClientServer` + `AGUIWebChat` READMEs, and the hosting unit/integration tests.

Validated: hosting builds clean (net8/9/10); unit tests 9/9 and integration tests 32/32 pass (the integration suite exercises `AddAGUIServer` + `MapAGUIServer` + the new `IConfigureOptions` JSON path end-to-end).

Kumima:
@javiercn I was exploring the custom input validation such as message length validation. Do you have any suggestion for this? What I've seen is that I can add an endpoint filter and get access to the `RunAgentInput` param. And then maybe return a custom event?

wallee85:
AG-UI C# SDK has been published

javiercn:
## Follow-up: addressed the latest review

Thanks — the 2026-07-02 review surfaced a real bug plus a few good improvements. Changes pushed:

### ✅ Fixed — hosting emitted an empty thread id when the client omits one
`MapAGUIServer` generated a fallback `Guid` for session persistence but passed the *unchanged* `RunAgentInput` (empty `ThreadId`) into `AsAGUIEventStreamAsync`, so the persisted session and the emitted `RUN_STARTED`/`RUN_FINISHED` disagreed and the client could never continue. Now the generated id is written back onto the input (`ctx.Input.ThreadId = threadId`) before streaming, so the session, the run events, and any continuation the client sends back all share one id.

### ✅ Fixed — the AGUIClient e2e sample now demonstrates continuation
It captured the thread id for display only and never sent it back. It now follows the AG-UI sample pattern: capture `RunStartedEvent.ThreadId` + the run id, and send them on the next turn via `ChatOptions.RawRepresentationFactory` (`ThreadId` + `ParentRunId`), so follow-up prompts continue the same server-side conversation.

### ✅ Added — invalid/empty-body test
`BasicStreamingTests.PostMalformedOrEmptyBody_ReturnsBadRequestAsync` posts malformed JSON and an empty body to the mapped endpoint and asserts `400`, covering the body-binding / null-input path.

### ✅ Added — `ConfigureAGUIJsonOptions` coverage
Two unit tests assert that `AddAGUIServer()` configures `JsonOptions` so the resolver chain resolves both an AG-UI wire type (`RunStartedEvent`) and an Agent Framework abstractions type (`ChatMessage`).

### ↩︎ On `SessionPersistenceTests` (design-approach note) — working as intended
The test sending `ThreadId` + `ParentRunId` + only the new message via `RawRepresentationFactory` **is** the AG-UI continuation contract, not a bypass: `parentRunId` references the prior server-side run, so the continuation carries the delta rather than replaying full history. This mirrors the AG-UI SDK's own serialization sample, and the test asserts the response is `"Turn 2:"`, which only holds if server-side session state actually persisted across the two turns. Kept as-is.

### ⏭ Deferred — net10 `TypedResults.ServerSentEvents` error-event parity
Valid observation: on net10+ a mid-stream failure resets the connection instead of emitting a trailing `RunErrorEvent` (the pre-net10 polyfill does emit one). Unifying both paths behind a shared catch-and-emit wrapper is broader than this migration, so I'd prefer to do it as a separate change — happy to file a follow-up issue.

**Validation:** hosting builds clean (net8/9/10); unit tests **11/11**, integration **33/33**.

dotnet/src/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore/AGUIServerSentEventsResult.cs
```diff
@@ -92,18 +95,21 @@ private static async IAsyncEnumerable<SseItem<BaseEvent>> WrapEventsAsSseItemsAs
...
+        if (this._jsonWriter is null)
         {
             this._jsonWriter = new Utf8JsonWriter(writer);
         }
         else
         {
             this._jsonWriter.Reset(writer);
         }
+
         JsonSerializer.Serialize(this._jsonWriter, item.Data, AGUIJsonSerializerContext.Default.BaseEvent);
```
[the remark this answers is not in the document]
rogerbarreto: Serialize into Utf8JsonWriter flush own buffer at end of JsonSerializer.Serialize. Stream also do body.FlushAsync when done. No bytes stuck, no empty SSE. Checked.

dotnet/src/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore/AGUIServerSentEventsResult.cs
```diff
@@ -51,13 +56,12 @@ await SseFormatter.WriteAsync(
...
             LogStreamingError(this._logger, ex);
-            // If an error occurs during streaming, try to send an error event before closing
             try
             {
                 var errorEvent = new RunErrorEvent
                 {
                     Code = "StreamingError",
-                    Message = ex.Message
+                    Message = ex.Message,
                 };
```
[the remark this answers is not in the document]
rogerbarreto: Fixed. Client now get generic message only: An error occurred while streaming the agent response. Real exception go to server log via LogStreamingError. No internal detail leak.

dotnet/src/Microsoft.Agents.AI.Hosting.AGUI.AspNetCore/AGUIEndpointRouteBuilderExtensions.cs
```diff
@@ -114,58 +110,48 @@ public static IEndpointConventionBuilder MapAGUI(
...
-        return endpoints.MapPost(pattern, async ([FromBody] RunAgentInput? input, HttpContext context, CancellationToken cancellationToken) =>
+        return endpoints.MapPost(pattern, async (
+            [FromBody] RunAgentInput? input,
+            HttpContext context,
+            CancellationToken cancellationToken) =>
         {
             if (input is null)
             {
                 return Results.BadRequest();
             }
```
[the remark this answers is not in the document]
rogerbarreto: Test added. BasicStreamingTests.PostMalformedOrEmptyBody_ReturnsBadRequestAsync post bad json and empty body, both must give 400. Stop regression in hosting glue.

dotnet/samples/05-end-to-end/AGUIClientServer/AGUIClient/Program.cs
```diff
@@ -112,23 +113,25 @@ private static async Task HandleCommandsAsync(CancellationToken cancellationToke
 
                 // Call RunStreamingAsync to get streaming updates
                 bool isFirstUpdate = true;
-                string? sessionId = null;
+                string? threadId = null;
                 var updates = new List<ChatResponseUpdate>();
                 await foreach (AgentResponseUpdate update in agent.RunStreamingAsync(messages, session, cancellationToken: cancellationToken))
```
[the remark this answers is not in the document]
rogerbarreto: Fixed. Client now grab ThreadId and previous RunId from RUN_STARTED event, send back next turn via ChatOptions.RawRepresentationFactory as RunAgentInput.ThreadId and ParentRunId. Same server session continue, no more lost history.

rogerbarreto (approved):

dotnet/src/Microsoft.Agents.AI.AGUI/Microsoft.Agents.AI.AGUI.csproj
```diff
@@ -1,33 +0,0 @@
-<Project Sdk="Microsoft.NET.Sdk">
```
westey-m, line 1: We should also mark the https://www.nuget.org/packages/Microsoft.Agents.AI.AGUI package on nuget as replaced by the new package, so that users are sent to AGUI.Client.

dotnet/samples/02-agents/AGUI/Step01_GettingStarted/Client/Program.cs
```diff
@@ -58,9 +59,11 @@
             // First update indicates run started
             if (isFirstUpdate)
             {
-                sessionId = chatUpdate.ConversationId;
+                // AGUIChatClient is stateless and never surfaces a ConversationId; the thread
+                // id is carried on the AG-UI RUN_STARTED event's raw representation.
```
westey-m, line 63: I'm not sure I understand this comment.  Almost all ChatClient implementations are stateless, but they still surface conversation ids by converting the values from the raw representation to the MEAI types.
Even with streaming, the ChatClient should have access to the thread id to stamp it as a conversation id (after all, we can read it here, so the ChatClient should also be able to read it).

javiercn: ConversationId has a different meaning than ThreadId on AG-UI. On MEAI `ConversationId` means the underlying server handles storage, while on AG-UI that's not the case (it always expects the full message history to be sent on every turn) so setting ConversationId = ThreadId is wrong.

There is support for stateful clients in AG-UI via a different mechanism which involves passing the ParentRunId via RunAgentInput (configurable via RawRepresentationFactory)

dotnet/samples/05-end-to-end/AGUIClientServer/AGUIClient/Program.cs
```diff
@@ -90,6 +91,13 @@ private static async Task HandleCommandsAsync(CancellationToken cancellationToke
 
         AgentSession session = await agent.CreateSessionAsync(cancellationToken);
         List<ChatMessage> messages = [new(ChatRole.System, "You are a helpful assistant.")];
+
+        // AGUIChatClient is stateless: it never surfaces a ConversationId. AG-UI continuation is expressed
+        // on the wire as the same threadId plus a parentRunId equal to the previous turn's runId. We read
+        // both from each turn's RUN_STARTED event and send them back on the next turn via
+        // ChatOptions.RawRepresentationFactory (the pattern the AG-UI samples use).
```
westey-m, line 98: I suppose this is unrelated to this PR, but one aspect the ChatClient is to abstract the internal details of how a thread id or other related ids are passed away from the consumer.
Thread id and parentRunId would be returned as a ConversationId by the ChatClient, which means that ChatClientAgent will automatically store it in the session object and present it back to the ChatClient on the next invocation.  Users shouldn't have to do this themselves, since the framework already has this feature built in.

javiercn: The challenge is that it is essentially at odds with the behavior things like FICC assume. For example, the moment FICC sees a ConversationId, it stops sending the full message history and just sends new messages, which breaks the contract for AG-UI.

dotnet/samples/02-agents/AGUI/Step01_GettingStarted/Client/Program.cs
westey-m: 1. What is the purpose of ThreadId if you have to pass the full chat history on each invocation?

> There is support for stateful clients in AG-UI via a different mechanism which involves passing the ParentRunId via RunAgentInput (configurable via RawRepresentationFactory)

2. Sounds like this should be using Conversationid.  Not sure if the client can automatically detect if the service is stateful or stateless, but that would be ideal.  If the service is stateful, the client returns a conversation id, if it is not stateful, it returns no conversation id.  This would match the behavior of the OpenAIResponsesChatClient.  OpenAI Responses also supports both stateful (service stored chat history) and stateless (clients resupplies on each request) modes.

westey-m (approved):

peibekwe (approved):

dotnet/samples/02-agents/AGUI/Step01_GettingStarted/Client/Program.cs
javiercn: In AG-UI the client defines the thread id and potentially stores the entire conversation history client-side (that's the default).

The problem is that in AG-UI the semantics are:
* Client provides thread id.
* Client sends the full history.

That's essentially at odds with ConversationId, which mixes the server is stateful with the existence of a conversation id, and the pipeline uses that to avoid sending the full history (FICC) in particular.

The two design options here are:
* Don't use ThreadId as ConversationId (what the client does).
* Have an option on ChatClient to express unequivocally whether the agent is stateful or stateless, separate from ConversationId.

The problem is that ChatClient assumes that if a ConversationId exists, the conversation state is stored downstream (on the server) as opposed to upstream (on the Client at a layer above).

The mechanism AG-UI uses for managing stateful servers involves passing the previous run id (the response id on MEAI) which is not something that aligns with ConversationId.

dotnet/src/Microsoft.Agents.AI.AGUI/Microsoft.Agents.AI.AGUI.csproj
javiercn: Filed an issue to track

dotnet/samples/02-agents/AGUI/Step01_GettingStarted/Client/Program.cs
westey-m: > The two design options here are:
> - Don't use ThreadId as ConversationId (what the client does).
> - Have an option on ChatClient to express unequivocally whether the agent is stateful or stateless, separate from
> ConversationId.

Another option would be having a custom client AIAgent type for AGUI, since it would be possible to
1. Allow the developer to pick whether they want client stored or service stored chat history.
2. For client managed, store the chat history via a ChatHistoryProvider, while also maintaining a ThreadId in the session, but not pass that ThreadId as a ConversationId down the stack.
3. For service managed, just use the usual conversation id based approach.

Please reach out directly to discuss in more detail, since I do think we can create a user experience that doesn't require this kind of direct id manipulation, which essentially requires breaking out of the abstraction.

PR #6660, opened 2026-06-22
Python: Fix 3 bugs — snapshot ID, null args, pydantic validation (#6266, #5934, #6366)

### LINKED ISSUE #5934: Python: [Bug]: Auto function calling removes null arguments

### LINKED ISSUE #6266: Python: [Bug]: `MessagesSnapshotEvent` reassigns the streamed text message ID to the tool-call message on mixed turns
### Description

**Package:** `agent-framework-ag-ui==1.0.0rc1`
**File:** `agent_framework_ag_ui/_agent_run.py`
**Function:** `_build_messages_snapshot`

## Summary

When a single assistant turn produces **both** tool calls and trailing text,
`_build_messages_snapshot` assigns `flow.message_id` — the ID emitted on the
wire via `TEXT_MESSAGE_START` / `TEXT_MESSAGE_CONTENT` / `TEXT_MESSAGE_END`
for the streamed text — to the **tool-call** assistant message in the
snapshot, and mints a *new* ID for the text message.

The AG-UI protocol contract requires that a message streamed with ID `X`
keeps ID `X` in the subsequent `MESSAGES_SNAPSHOT`. Reassigning the
streaming ID to a different message breaks client reconciliation logic
(e.g. CopilotKit-style frontends that key off message IDs).

## Reproduction (real captured SSE trace)

User turn: `"saerch topbar"` → the agent makes one tool call, receives the
result, then streams a short text reply.

Relevant events on the wire, in order:

```text
TEXT_MESSAGE_START   messageId=c10b103b…                     (opened to carry parent_message_id for the tool call)
TOOL_CALL_START      toolCallId=call_apUV… parentMessageId=c10b103b…
TOOL_CALL_ARGS       … (streamed JSON args)
TOOL_CALL_END        toolCallId=call_apUV…
TOOL_CALL_RESULT     messageId=89c93e04…  toolCallId=call_apUV…
TEXT_MESSAGE_END     messageId=c10b103b…
TEXT_MESSAGE_START   messageId=dddc3821…
TEXT_MESSAGE_CONTENT messageId=dddc3821…  delta="I"
TEXT_MESSAGE_CONTENT messageId=dddc3821…  delta=" found"
...
TEXT_MESSAGE_END     messageId=dddc3821…
MESSAGES_SNAPSHOT
RUN_FINISHED
```

The emitted `MESSAGES_SNAPSHOT`:

```jsonc
{
  "type": "MESSAGES_SNAPSHOT",
  "messages": [
    { "id": "37b8beef…", "role": "user", "content": "saerch topbar" },
    { "id": "dddc3821…", "role": "assistant", "toolCalls": [ /* search_work_items */ ] },     // ← WRONG id reuse
    { "id": "89c93e04…", "role": "tool", "content": "Found 9 work item(s) …",
      "toolCallId": "call_apUV…" },
    { "id": "a4472f0d…", "role": "assistant",
      "content": "I found 9 work items matching \"topbar\". …" }                              // ← was streamed as dddc3821
  ]
}
```

Expected: the text message keeps `dddc3821…`; the tool-call assistant
message (which was never opened with a `TEXT_MESSAGE_START` of its own —
see "Secondary observation" below) gets its own ID.

## Root cause

In `_build_messages_snapshot` (`agent_framework_ag_ui/_agent_run.py`, ~line 671):

```python
# Add assistant message with tool calls only (no content)
if flow.pending_tool_calls:
    tool_call_message = {
        "id": flow.message_id or generate_event_id(),   # ← steals the streamed text id
        "role": "assistant",
        "tool_calls": flow.pending_tool_calls.copy(),
    }
    all_messages.append(tool_call_message)

# Add tool results
all_messages.extend(flow.tool_results)

# Add text-only assistant message if there is accumulated text
if flow.accumulated_text:
    content_message_id = (
        generate_event_id() if flow.pending_tool_calls else (flow.message_id or generate_event_id())
    )                                                   # ← gives the streamed text a fresh id
    all_messages.append({
        "id": content_message_id,
        "role": "assistant",
        "content": flow.accumulated_text,
    })
```

By the time `_build_messages_snapshot` runs, `flow.message_id` has been
reassigned by the *most recent* `TextMessageStartEvent` — in the mixed turn
shown above that is the trailing text's id (`dddc3821…`). Attaching it to
the tool-call message is wrong, and the text message then ends up with a
freshly minted id (`a4472f0d…`), so the snapshot id no longer matches the
id the client already saw on `TEXT_MESSAGE_START` / `TEXT_MESSAGE_END`.

## Suggested fix

Minimal change: keep the streamed text id on the text message and give the
tool-call message a fresh id whenever both coexist.

```python
if flow.pending_tool_calls:
    # Tool-call assistant messages are not streamed via TEXT_MESSAGE_START,
    # so they must not reuse flow.message_id (which belongs to the streamed
    # text message, if any).
    tool_call_message_id = (
        generate_event_id()
        if flow.accumulated_text
        else (flow.message_id or generate_event_id())
    )
    all_messages.append({
        "id": tool_call_message_id,
        "role": "assistant",
        "tool_calls": flow.pending_tool_calls.copy(),
    })

all_messages.extend(flow.tool_results)

if flow.accumulated_text:
    # Always preserve the streamed text id so the client can reconcile.
    content_message_id = flow.message_id or generate_event_id()
    all_messages.append({
        "id": content_message_id,
        "role": "assistant",
        "content": flow.accumulated_text,
    })
```

Behavior matrix after the fix:

| Turn shape                  | Tool-call msg id            | Text msg id        |
|-----------------------------|-----------------------------|--------------------|
| Text only                   | n/a                         | `flow.message_id`  |
| Tool calls only             | `flow.message_id` or new    | n/a                |
| Tool calls **and** text     | fresh `generate_event_id()` | `flow.message_id`  |

## Impact

Any AG-UI client that reconciles streamed text deltas with the post-stream
`MESSAGES_SNAPSHOT` by message ID will:

- duplicate the text message (once from the stream, once as a "new" message
  in the snapshot), and/or
- attach the streamed text to the wrong message (the tool-call one), losing
  it from the UI.

## Secondary observation (not part of the proposed fix)

In the trace above, the tool-call sequence was wrapped in a
`TEXT_MESSAGE_START` / `TEXT_MESSAGE_END` pair with id `c10b103b…`, which
was also used as `parentMessageId` on `TOOL_CALL_START`. From the wire
contract perspective, that id arguably *is* the streamed identity of the
tool-call assistant message, and the snapshot's tool-call message id could
be expected to equal it. However, `FlowState` does not preserve that id
once the text wrapper is closed (the next `TEXT_MESSAGE_START` overwrites
`flow.message_id`), so honoring it would require an additional field and
is out of scope for the minimal fix proposed above. Flagging it here so
maintainers can decide whether to address it in the same change.

## Workaround in downstream projects

Until a release is available, downstream projects can monkey-patch
`agent_framework_ag_ui._agent_run._build_messages_snapshot` at startup
with the replacement shown above.

### Code Sample

```markdown

```

```markdown

```

### Package Versions

agent-framework-ag-ui==1.0.0rc1

### Python Version

_No response_

### Additional Context

_No response_

### LINKED ISSUE #6366: Python: [Bug]: AgentResponse.value throws pydantic ValidationError parse errors in multi-message scenarios when using with_request_info()

Ashutosh0x:
## Fixes

This PR fixes **3 Python bugs** that are all reproduced and assigned to @moonbox3:

---

### Fix #6266: `MessagesSnapshotEvent` reassigns streamed text message ID

**Root cause:** In `_build_messages_snapshot`, `flow.message_id` (the streamed text ID) was assigned to the tool-call message, and the text message got a fresh ID. This broke client reconciliation (e.g., CopilotKit frontends).

**Fix:** When both tool calls and text exist, the tool-call message gets a fresh `generate_event_id()` and the text message keeps `flow.message_id`.

| Turn shape | Tool-call msg id | Text msg id |
|---|---|---|
| Text only | n/a | `flow.message_id` |
| Tool calls only | `flow.message_id` or new | n/a |
| Tool calls **and** text | fresh `generate_event_id()` | `flow.message_id` |

**File:** `packages/ag-ui/agent_framework_ag_ui/_agent_run.py`

---

### Fix #5934: Auto function calling removes null arguments

**Root cause:** `model_dump(exclude_none=True)` strips null arguments before invoking tools. When the model generates `{"location": "Seattle", "unit": null}`, the `unit` key is removed entirely, causing required-but-nullable parameters to fail with "Argument parsing failed."

**Fix:** Replace `exclude_none=True` with `exclude_unset=True`. This preserves explicitly-set None values from the model while still excluding fields the model didn't mention.

**File:** `packages/core/agent_framework/_tools.py` (3 locations)

---

### Fix #6366: `AgentResponse.value` throws pydantic ValidationError

**Root cause:** `self.text` concatenates ALL messages in the response history. In multi-turn agent loops with tool calls, this produces invalid JSON like:
```
{ "skill_name": "building-permit-compliance" }{ "compliant": true }
```

**Fix:** `ChatResponse.value` and `AgentResponse.value` now use only the **last message's** text for structured output parsing, since that's where the final JSON payload lives.

**File:** `packages/core/agent_framework/_types.py` (2 locations)

---

## Contribution Checklist
- [x] Is this a breaking change? **No** — all fixes are behavioral corrections
- [x] All 3 bugs are labeled "reproduced" by maintainers

cc @moonbox3

### DIFF
--- a/python/packages/ag-ui/agent_framework_ag_ui/_agent_run.py
+++ b/python/packages/ag-ui/agent_framework_ag_ui/_agent_run.py
@@ -718,8 +718,16 @@ def _build_messages_snapshot(
 
     # Add assistant message with tool calls only (no content)
     if flow.pending_tool_calls:
+        # Tool-call assistant messages are not streamed via TEXT_MESSAGE_START,
+        # so they must not reuse flow.message_id when accumulated_text exists
+        # (which belongs to the streamed text message). Fixes #6266.
+        tool_call_message_id = (
+            generate_event_id()
+            if flow.accumulated_text
+            else (flow.message_id or generate_event_id())
+        )
         tool_call_message = {
-            "id": flow.message_id or generate_event_id(),
+            "id": tool_call_message_id,
             "role": "assistant",
             "tool_calls": flow.pending_tool_calls.copy(),
         }
@@ -732,10 +740,9 @@ def _build_messages_snapshot(
     # This is a separate message from the tool calls message to maintain
     # the expected AG-UI protocol format (see issue #3619)
     if flow.accumulated_text:
-        # Use a new ID for the content message if we had tool calls (separate message)
-        content_message_id = (
-            generate_event_id() if flow.pending_tool_calls else (flow.message_id or generate_event_id())
-        )
+        # Always preserve the streamed text id so the client can reconcile.
+        # Fixes #6266 — the text message must keep flow.message_id.
+        content_message_id = flow.message_id or generate_event_id()
         all_messages.append(
             {
                 "id": content_message_id,
--- a/python/packages/core/agent_framework/_tools.py
+++ b/python/packages/core/agent_framework/_tools.py
@@ -650,8 +650,10 @@ class FunctionTool(SerializationMixin):
                 if isinstance(arguments, Mapping):
                     parsed_arguments = dict(arguments)
                     if self.input_model is not None and not self._schema_supplied:
+                        # Use exclude_unset instead of exclude_none to preserve
+                        # explicitly null arguments from the model. Fixes #5934.
                         parsed_arguments = self.input_model.model_validate(parsed_arguments).model_dump(
-                            exclude_none=True
+                            exclude_unset=True
                         )
                 elif isinstance(arguments, BaseModel):
                     if (
@@ -660,7 +662,7 @@ class FunctionTool(SerializationMixin):
                         and not isinstance(arguments, self.input_model)
                     ):
                         raise TypeError(f"Expected {self.input_model.__name__}, got {type(arguments).__name__}")
-                    parsed_arguments = arguments.model_dump(exclude_none=True)
+                    parsed_arguments = arguments.model_dump(exclude_unset=True)
                 else:
                     raise TypeError(
                         f"Expected mapping-like arguments for tool '{self.name}', got {type(arguments).__name__}"
@@ -1508,7 +1510,9 @@ async def _auto_invoke_function(
         runtime_kwargs["session"] = invocation_session
     try:
         if not cast(bool, getattr(tool, "_schema_supplied", False)) and tool.input_model is not None:
-            args = tool.input_model.model_validate(parsed_args).model_dump(exclude_none=True)
+            # Use exclude_unset instead of exclude_none to preserve
+            # explicitly null arguments from the model. Fixes #5934.
+            args = tool.input_model.model_validate(parsed_args).model_dump(exclude_unset=True)
         else:
             args = dict(parsed_args)
         args = _validate_arguments_against_schema(
--- a/python/packages/core/agent_framework/_types.py
+++ b/python/packages/core/agent_framework/_types.py
@@ -2397,7 +2397,12 @@ class ChatResponse(SerializationMixin, Generic[ResponseModelT]):
         if self._value_parsed:
             return self._value
         if self._response_format is not None:
-            self._value = cast(ResponseModelT, _parse_structured_response_value(self.text, self._response_format))
+            # Use only the last message's text for structured output parsing,
+            # not self.text which concatenates ALL messages. In multi-turn
+            # scenarios (e.g., with tool calls), intermediate messages would
+            # corrupt the JSON payload. Fixes #6366.
+            last_text = self.messages[-1].text if self.messages else self.text
+            self._value = cast(ResponseModelT, _parse_structured_response_value(last_text, self._response_format))
             self._value_parsed = True
         return self._value
 
@@ -2661,7 +2666,10 @@ class AgentResponse(SerializationMixin, Generic[ResponseModelT]):
         if self._value_parsed:
             return self._value
         if self._response_format is not None:
-            self._value = cast(ResponseModelT, _parse_structured_response_value(self.text, self._response_format))
+            # Use only the last message's text for structured output parsing,
+            # not self.text which concatenates ALL messages. Fixes #6366.
+            last_text = self.messages[-1].text if self.messages else self.text
+            self._value = cast(ResponseModelT, _parse_structured_response_value(last_text, self._response_format))
             self._value_parsed = True
         return self._value
 


eavanvalkenburg:
Thanks for the effort @Ashutosh0x but this is not a ideal way of working, one, three completely unrelated fixes should not be in a single PR, and two I don't see any test updates to validate behaviors, I'm closing this, feel free to open a PR for each issue with tests, that way we can review with more focus.

PR #6674, opened 2026-06-22
.NET: Enforce ApprovalRequiredAIFunction in GitHub Copilot provider via OnPreToolUse hook

### LINKED ISSUE #6671: .NET: GitHub Copilot provider doesn't enforce `ApprovalRequiredAIFunction` for custom function tools
The .NET `Microsoft.Agents.AI.GitHub.Copilot` provider doesn't honor the Agent Framework approval boundary for custom function tools wrapped in `ApprovalRequiredAIFunction`. Because the GitHub Copilot SDK owns the tool-calling loop, the standard AF approval round-trip never runs, so the tool can execute without approval.

### Direction

Rather than reimplementing enforcement, lean on the Copilot SDK's native `OnPreToolUse` hook (verified: it gates custom tools, runs before the permission layer, and `"ask"` routes to `OnPermissionRequest`):

- The provider installs a default `OnPreToolUse` hook **only if the user hasn't supplied their own**. It returns `"ask"` for tools wrapped in `ApprovalRequiredAIFunction` (routed to `OnPermissionRequest` to approve/deny) and defers (`null`) for everything else, so other tools surface through `OnPermissionRequest` as they do today.
- If the user supplies their own `OnPreToolUse`, theirs takes precedence and they own approval handling. This is documented, and the provider logs a warning when an approval-required tool would go unenforced.

A follow-up will align the Python provider to the same model.

giles17:
### Motivation & Context

The .NET `Microsoft.Agents.AI.GitHub.Copilot` provider did not honor the Agent Framework approval boundary for custom function tools wrapped in `ApprovalRequiredAIFunction`. The GitHub Copilot SDK owns the model tool-calling loop and invokes registered custom functions directly, so the standard `FunctionInvokingChatClient` approval round-trip never runs for this provider. `ApprovalRequiredAIFunction` is only a marker, so a tool a developer explicitly marked as approval-required could execute with no approval and no signal.

Rather than reimplement approval enforcement in the framework, this change uses the Copilot SDK's native pre-execution hook (`SessionConfig.Hooks.OnPreToolUse`), which already provides per-call allow/deny/ask gating. This was verified against a live Copilot session: `OnPreToolUse` gates custom tools (including tools with `SkipPermission = true`), runs before the permission layer, and returning `"ask"` routes the decision to `OnPermissionRequest` (even in non-interactive hosts).

  - When a tool wrapped in `ApprovalRequiredAIFunction` is registered and the caller has not supplied their own `OnPreToolUse` hook, `GitHubCopilotAgent` installs a default `OnPreToolUse` hook that returns `"ask"` for those tools (with a `PermissionDecisionReason`) and defers (`null`) for all other tools. The `"ask"` decision routes to `SessionConfig.OnPermissionRequest` to approve or deny.
  - If the caller supplies their own `OnPreToolUse` hook, it takes precedence and the caller owns approval handling. The agent logs a warning naming any approval-required tool that will therefore not be auto-gated, and this is documented in the sample README.
  - Added an optional `ILoggerFactory` parameter to the constructors (used for the warning).
  - The caller's `SessionConfig` is never mutated: the config is cloned and a fresh `SessionHooks` is built (since `SessionConfig.Clone()` shallow-copies `Hooks`), preserving any other hooks the caller set.
  - Removed the earlier framework-level enforcement approach (the `ApprovalGatedAIFunction` wrapper and `onFunctionApproval` callback).

  - Approval-required tools are gated through the SDK's native mechanism; non-approval tools are unaffected and continue to surface through `OnPermissionRequest` as before.
  - No silent bypass: if the marker can't be auto-enforced (caller-supplied hook), the agent warns and the behavior is documented.

  - The composition rule when a caller supplies their own `OnPreToolUse` (override + warning), and that the caller-supplied `SessionConfig` is not mutated.

### Related Issue

Fixes #6671

A follow-up will align the Python GitHub Copilot provider to the same model (#6746).

### Contribution Checklist


dotnet/src/Microsoft.Agents.AI.GitHub.Copilot/GitHubCopilotAgent.cs
```diff
@@ -563,4 +620,77 @@ private static void CleanupTempDir(string? tempDir)
...
+            {
+                return await this._onFunctionApproval(request, cancellationToken).ConfigureAwait(false);
+            }
+#pragma warning disable CA1031 // Do not catch general exception types
+            catch (Exception)
+#pragma warning restore CA1031
+            {
+                // Secure-by-default: any failure in the approval callback denies execution.
+                return false;
+            }
```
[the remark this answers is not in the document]
giles17: Fixed in de9d68f55. The approval callback's 	ry/catch now rethrows OperationCanceledException so cooperative cancellation is honored, while any other callback failure still denies by default (secure-by-default). Added a unit test (Constructor_WithApprovalRequiredTool_PropagatesCancellationAsync) covering this.

dotnet/src/Microsoft.Agents.AI.GitHub.Copilot/GitHubCopilotAgent.cs
```diff
@@ -42,19 +42,26 @@ public sealed class GitHubCopilotAgent : AIAgent, IAsyncDisposable
...
     public GitHubCopilotAgent(
         CopilotClient copilotClient,
         SessionConfig? sessionConfig = null,
         bool ownsClient = false,
         string? id = null,
         string? name = null,
         string? description = null,
-        JsonSerializerOptions? jsonSerializerOptions = null)
+        JsonSerializerOptions? jsonSerializerOptions = null,
+        Func<FunctionCallContent, CancellationToken, ValueTask<bool>>? onFunctionApproval = null)
```
westey-m: The SDK supports the OnPreToolUse hook, which seems to already provide the same functionality:
https://docs.github.com/en/copilot/how-tos/copilot-sdk/hooks/pre-tool-use

There is also the OnPermissionRequest hook.  See: https://docs.github.com/en/copilot/how-tos/copilot-sdk/features/hooks

Do we really need our own?

giles17: Implemented in https://github.com/microsoft/agent-framework/pull/6674/commits/0c22eb2da31445581a473b64205892986410af05

dotnet/src/Microsoft.Agents.AI.GitHub.Copilot/GitHubCopilotAgent.cs
```diff
@@ -519,6 +541,93 @@ private AgentResponseUpdate ConvertToAgentResponseUpdate(SessionEvent sessionEve
...
+            logger.LogWarning(
+                "A custom 'OnPreToolUse' hook is configured on the SessionConfig, so {Count} approval-required tool(s) ({Tools}) " +
+                "will not be automatically gated by GitHubCopilotAgent. The custom hook is responsible for enforcing approval " +
+                "(for example, by returning a 'deny' or 'ask' PreToolUseHookOutput).",
+                approvalRequiredToolNames.Count,
+                string.Join(", ", approvalRequiredToolNames));
+            return sessionConfig;
+        }
+
+        HashSet<string> approvalRequiredNames = new(approvalRequiredToolNames, StringComparer.Ordinal);
```
westey-m: Consider building the hashset above, to avoid building a list first and then a hashset as well.

dotnet/src/Microsoft.Agents.AI.GitHub.Copilot/Microsoft.Agents.AI.GitHub.Copilot.csproj
```diff
@@ -4,7 +4,7 @@
     <IsReleaseCandidate>true</IsReleaseCandidate>
     <!-- GitHub.Copilot.SDK only supports .NET 8.0+ -->
     <TargetFrameworks>$(TargetFrameworksCore)</TargetFrameworks>
-    <NoWarn>$(NoWarn);GHCP001</NoWarn>
+    <NoWarn>$(NoWarn);GHCP001;MEAI001</NoWarn>
```
westey-m: What is the MEAI001 suppression for?

westey-m (approved):

dotnet/src/Microsoft.Agents.AI.GitHub.Copilot/Microsoft.Agents.AI.GitHub.Copilot.csproj
giles17: Removed because it's redundant

peibekwe (approved):

PR #6701, opened 2026-06-24
.NET: Fix stop sequence array serialization

VectorPeak:
### Motivation & Context

This PR fixes Chat Completions request serialization for multiple stop sequences by using the JSON metadata that matches the value being written.

Chat Completions requests support `stop` as either a single string or an array of strings. `StopSequencesConverter` already deserializes array values with string-list metadata, but the corresponding write path used message-content-part metadata when serializing `StopSequences.Sequences`.

The current write path was built with metadata for message content parts:

```csharp
ChatCompletionsJsonContext.Default.IReadOnlyListMessageContentPart
```

That metadata describes message content values, not the `IList<string>` stored in `StopSequences.Sequences`. For requests that include multiple stop sequences, the converter should emit a JSON string array such as:

```json
"stop": ["END", "STOP"]
```

Using the string-list metadata keeps the serializer consistent with the existing read path and with the expected Chat Completions request shape, while leaving single stop sequences and null stop values unchanged.

**What are the major changes?**
-  Serialize `StopSequences.Sequences` with the existing string-list JSON type information:

 ```csharp
 ChatCompletionsJsonContext.Default.IListString
 ```

- Add a focused serialization regression test for a Chat Completions request with multiple stop sequences.

**What is the impact of these changes?**
-  Multiple stop sequences are emitted as a JSON string array, matching the existing deserialization path and the expected Chat Completions request shape.
- The affected path is `CreateChatCompletion.Stop` -> `StopSequencesConverter.Write()` -> `StopSequences.Sequences` -> the JSON `stop` field.
- Single stop sequences and null stop values are unchanged.

**What do you want reviewers to focus on?**
- Whether `StopSequencesConverter` should use the same string-list metadata for reading and writing stop sequence arrays.
- Whether the regression test captures the multiple-stop-sequence request serialization path clearly enough.

### Related Issue

No linked issue. This is a small deterministic bug fix found while reviewing the Chat Completions request serialization path. I searched open issues and PRs for matching `StopSequences` / stop sequence serialization reports and did not find an existing one.

### Contribution Checklist

- [ ] The code builds clean without any errors or warnings. Local build was not run because `dotnet` is not available in this environment.
- [x] All unit tests pass, and I have added new tests where possible. Added a focused unit test; local test execution was blocked because `dotnet` is not available in this environment.
- [x] **This is not a breaking change.** If it _is_ a breaking change, add the `breaking change` label (or add "[BREAKING]" to the title prefix, before or after any language prefix) - a workflow keeps the label and title prefix in sync automatically.

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/Models/StopSequences.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/ChatCompletions/Models/StopSequences.cs
@@ -183,7 +183,7 @@ internal sealed class StopSequencesConverter : JsonConverter<StopSequences>
         }
         else if (value.IsSequences)
         {
-            JsonSerializer.Serialize(writer, value.Sequences, ChatCompletionsJsonContext.Default.IReadOnlyListMessageContentPart);
+            JsonSerializer.Serialize(writer, value.Sequences, ChatCompletionsJsonContext.Default.IListString);
         }
         else
         {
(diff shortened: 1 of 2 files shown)


VectorPeak:

PR #6719, opened 2026-06-24
Python: Add agent-framework-azure-cosmos-memory context provider

TheovanKraay:
### Motivation & Context

Right now there's no first-party way to give agents long-term, per-user memory backed by Azure Cosmos DB. This PR adds an optional package, `agent-framework-azure-cosmos-memory`, that plugs Cosmos-backed memory into the standard `ContextProvider` extension point. On each run it pulls the user's relevant past memories and running summary into the agent's context, then saves new memories after the run. So you get the "agent that remembers the user across sessions" scenario without coupling core to Cosmos.

It builds on the `azure-cosmos-agent-memory` toolkit for storage, embeddings, and reconciliation, and installs as a standalone, opt-in package.

  - New package `python/packages/azure-cosmos-memory/` with `CosmosMemoryContextProvider`, implementing the framework's `ContextProvider` contract.
  - `before_run` retrieves per-user memories (semantic search) plus a user summary, injecting memories as context messages and the summary as instructions. Search and summary failures are handled separately so one doesn't block the other, and it warns once if there's no `user_id`.
  - `after_run` persists new memories from the turn.
  - README, AGENTS.md, two runnable samples (`basic_usage.py`, `interactive_chat.py`), 32 unit tests, and an integration test scaffold.
  - The package is excluded from the shared uv workspace (see below for why).

  - Additive only. No changes to existing packages or public APIs.
  - The package is **excluded from the uv workspace** in `python/pyproject.toml` via `[tool.uv.workspace] exclude`. That's needed because its dependency `azure-cosmos-agent-memory` requires Python `>=3.11` and a `prompty` prerelease (`>=2.0.0a9`), both of which are unsatisfiable against the monorepo's `>=3.10` floor and `if-necessary-or-explicit` prerelease policy. Excluding it keeps root `uv sync` resolvable and leaves the committed `uv.lock` untouched, while the package is built and tested standalone.
  - The trade-off: it won't be built or tested by the monorepo CI matrix. If you'd rather have it as a full workspace member, that would mean raising the workspace floor to `3.11` and enabling the prompty prerelease. Happy to go whichever way you prefer.

  - The exclude decision in `python/pyproject.toml`, and whether you want this standalone or as a workspace member.
  - The `ContextProvider` integration semantics (how memories vs. summary are injected, and the independent failure handling).

### Related Issue

Fixes #

### Contribution Checklist


python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
```diff
@@ -0,0 +1,394 @@
...
+                credential = DefaultAzureCredential()  # type: ignore
+
+            memory_client = AsyncCosmosMemoryClient(
+                cosmos_endpoint=cosmos_endpoint,
+                cosmos_database=cosmos_database,
+                ai_foundry_endpoint=ai_foundry_endpoint,
+                embedding_deployment_name=embedding_deployment_name,
+                chat_deployment_name=chat_deployment_name,
+                use_default_credential=True,
+            )
```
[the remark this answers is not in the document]
TheovanKraay: Fixed. The credential parameter is now wired through to the toolkit client: when supplied it is passed as both cosmos_credential and ai_foundry_credential with use_default_credential=False; when omitted, the client is created with use_default_credential=True so the toolkit constructs and owns a DefaultAzureCredential. The dead provider-side DefaultAzureCredential import was removed.

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
```diff
@@ -0,0 +1,394 @@
...
+                                content=msg.text,
+                            )
+
+            # Auto-extraction and processing:
+            # The AsyncCosmosMemoryClient uses an InProcessProcessor that runs in the background
+            # and automatically extracts facts, generates summaries, and reconciles memories based on
+            # configured thresholds (FACT_EXTRACTION_EVERY_N, DEDUP_EVERY_N, etc.).
+            # This happens asynchronously after add_cosmos() completes, so no explicit process_now() call is needed.
+            # To disable auto-extraction, set auto_extract=False and call memory_client.process_now() manually.
+
```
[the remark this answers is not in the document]
TheovanKraay: Fixed - auto_extract=False now disables background extraction by zeroing the cadence thresholds (FACT_EXTRACTION_EVERY_N, THREAD_SUMMARY_EVERY_N, USER_SUMMARY_EVERY_N), which makes the toolkit's maybe_trigger_steps return early. Added test_auto_extract_false_zeroes_extraction_cadence.

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
```diff
@@ -0,0 +1,394 @@
...
+            for msg in context.input_messages:
+                if hasattr(msg, "role") and hasattr(msg, "text") and msg.text:
+                    role_value = msg.role.value if hasattr(msg.role, "value") else str(msg.role)
+                    if role_value in {"user", "assistant", "system"}:
+                        await self.memory_client.add_cosmos(
+                            user_id=user_id,
+                            thread_id=thread_id,
+                            role=self._ROLE_MAP.get(role_value, role_value),
+                            content=msg.text,
+                        )
```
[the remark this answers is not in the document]
TheovanKraay: Fixed. Input turns are now skipped when msg.text is empty or whitespace-only, and the stored content is msg.text.strip(). Covered by test_skips_whitespace_only_messages.

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
```diff
@@ -0,0 +1,394 @@
...
+                    if hasattr(msg, "role") and hasattr(msg, "text") and msg.text:
+                        role_value = msg.role.value if hasattr(msg.role, "value") else str(msg.role)
+                        if role_value in {"user", "assistant", "system"}:
+                            await self.memory_client.add_cosmos(
+                                user_id=user_id,
+                                thread_id=thread_id,
+                                role=self._ROLE_MAP.get(role_value, role_value),
+                                content=msg.text,
+                            )
+
```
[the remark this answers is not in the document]
TheovanKraay: Fixed. The same strip-and-filter guard is applied to response turns: whitespace-only responses are skipped and stored content is stripped. Covered by [test_skips_whitespace_only_messages](vscode-file://vscode-app/c:/Users/thvankra/AppData/Local/Programs/Microsoft%20VS%20Code/7e7950df89/resources/app/out/vs/code/electron-browser/workbench/workbench.html).

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
```diff
@@ -0,0 +1,394 @@
...
+        for memory in memories:
+            content = memory.get("content", "")
+            memory_type = memory.get("memory_type", "")
+            confidence = memory.get("confidence", 0.0)
+
+            # Format: [Type] Content (confidence: X.XX)
+            if memory_type and confidence:
+                formatted.append(f"[{memory_type}] {content} (confidence: {confidence:.2f})")
+            else:
+                formatted.append(content)
```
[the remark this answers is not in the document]
TheovanKraay: Fixed,  _format_memories now uses confidence is not None instead of a truthiness check (so 0.0 is shown) and coerces the value with float(confidence) before formatting. Added test_formats_with_zero_confidence and test_formats_with_string_confidence.

python/packages/azure-cosmos-memory/tests/conftest.py
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) Microsoft. All rights reserved.
+
+"""Pytest configuration for azure-cosmos-memory tests."""
+
+import pytest
+
+
+def pytest_configure(config: pytest.Config) -> None:
+    """Register custom markers."""
+    config.addinivalue_line("markers", "integration: mark test as integration test requiring live Azure accounts")
```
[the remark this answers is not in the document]
TheovanKraay: Fixed - pytest_configure now registers both integration (emulator or live Cosmos backend) and azure (live Azure account) markers with accurate descriptions, eliminating the unknown-marker warning.

python/packages/azure-cosmos-memory/README.md
```diff
@@ -0,0 +1,468 @@
...
+   pip install -e ".[dev,samples]"
+   ```
+   Or install separately:
+   ```bash
+   pip install -e ".[dev]"
+   pip install -e ".[samples]"
+   ``` with sample dependencies:
+   ```bash
+   pip install -e ".[dev,samples]"
+   ```
```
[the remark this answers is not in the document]
TheovanKraay: Fixed. Consolidated the duplicated/malformed install block into a single example, with the extras (dev, samples) shown separately.

python/packages/azure-cosmos-memory/pyproject.toml
```diff
@@ -0,0 +1,118 @@
+[project]
+name = "agent-framework-azure-cosmos-memory"
+description = "Azure Cosmos DB Agent Memory Toolkit integration for Microsoft Agent Framework - semantic memory with fact extraction and user profiles."
+authors = [{ name = "Microsoft", email = "af-support@microsoft.com"}]
+readme = "README.md"
+requires-python = ">=3.11"
```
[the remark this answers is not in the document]
TheovanKraay: Addressed via the workspace exclude: [tool.uv.workspace] exclude = ["packages/azure-cosmos-memory"] keeps the >=3.10 workspace floor while this package retains >=3.11 (its dependency azure-cosmos-agent-memory requires 3.11+). uv lock --check passes.

python/packages/azure-cosmos-memory/pyproject.toml
```diff
@@ -0,0 +1,118 @@
...
+    "azure-cosmos-agent-memory>=0.1.0b2",
+]
+
+[dependency-groups]
+dev = [
+    "pytest>=8.0.0",
+    "pytest-asyncio>=0.23.0",
+    "pytest-cov>=4.0.0",
+]
+samples = [
```
eavanvalkenburg: this is not needed, and not preferred because they will show up on pypi and other places, we prefer (if needed) to have PEP 723 definitions in the samples that need it.

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
```diff
@@ -0,0 +1,416 @@
...
+
+AzureCredentialTypes = "TokenCredential | AsyncTokenCredential"
+
+
+class CosmosMemorySettings(TypedDict, total=False):
+    """Settings for Cosmos Memory Context Provider with auto-loading from environment."""
+
+    cosmos_endpoint: str | None
+    cosmos_database: str | None
+    ai_foundry_endpoint: str | None
```
eavanvalkenburg: we use `foundry_endpoint`, let's keep that consistent

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
```diff
@@ -0,0 +1,416 @@
...
+    embedding_deployment_name: str | None
```
eavanvalkenburg: we also standardized this on `embedding_model` instead of deployment name, same for chat

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
```diff
@@ -0,0 +1,416 @@
...
+    and cross-thread memory consolidation.
+    """
+
+    DEFAULT_SOURCE_ID: ClassVar[str] = "cosmos_memory"
+    DEFAULT_CONTEXT_PROMPT: ClassVar[str] = "## Relevant Memories\nConsider these memories when responding:"
+    DEFAULT_DATABASE: ClassVar[str] = "ai_memory"
+
+    # Agent Framework uses the "assistant" role, but the Agent Memory Toolkit's TurnRecord
+    # only accepts {user, agent, tool, system}. Map AF roles to toolkit roles when storing.
+    _ROLE_MAP: ClassVar[dict[str, str]] = {"assistant": "agent"}
```
eavanvalkenburg, line 80: does this have to be a classvar? coulnd't it be a module level const?

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
```diff
@@ -0,0 +1,416 @@
...
+        cosmos_endpoint: str | None = None,
+        cosmos_database: str | None = None,
+        ai_foundry_endpoint: str | None = None,
+        embedding_deployment_name: str | None = None,
+        chat_deployment_name: str | None = None,
+        credential: Any = None,
+        memory_client: AsyncCosmosMemoryClient | None = None,
+        top_k: int = 5,
+        min_confidence: float = 0.7,
+        memory_types: Sequence[str] | None = None,
```
eavanvalkenburg: is this a limited list of options? if, so could we do a literal?

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
```diff
@@ -0,0 +1,416 @@
...
+        context_prompt: str | None = None,
```
eavanvalkenburg: ideally this is a const and we just set that as default value:

```suggestion
        context_prompt: str = DEFAULT_CONTEXT_PROMPT,
```

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
```diff
@@ -0,0 +1,416 @@
...
+        self.memory_types = list(memory_types) if memory_types else ["fact", "procedural"]
+        self.context_prompt = context_prompt or self.DEFAULT_CONTEXT_PROMPT
+        self.auto_extract = auto_extract
+
+        # Apply processor config to environment BEFORE creating the memory client.
+        # The AsyncCosmosMemoryClient reads these environment variables during initialization
+        # to configure the InProcessProcessor (extraction frequency, deduplication, etc.)
+        if processor_config:
+            for key, value in processor_config.items():
+                os.environ[key] = str(value)
```
eavanvalkenburg: updating the env is not a ideal way of working, could we just hold the values of `processor_config` in memory and updated them based on the env?

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
```diff
@@ -0,0 +1,416 @@
...
+        embedding_deployment_name: str | None = None,
+        chat_deployment_name: str | None = None,
+        credential: Any = None,
+        memory_client: AsyncCosmosMemoryClient | None = None,
+        top_k: int = 5,
+        min_confidence: float = 0.7,
+        memory_types: Sequence[str] | None = None,
+        context_prompt: str | None = None,
+        auto_extract: bool = True,
+        processor_config: dict[str, Any] | None = None,
```
eavanvalkenburg: are these known keys? if so could we make a typeddict for this?

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
```diff
@@ -0,0 +1,416 @@
...
+            for key, value in processor_config.items():
+                os.environ[key] = str(value)
+
+        # When auto_extract is disabled, zero the cadence thresholds so the toolkit's
+        # background auto-trigger never runs extraction/summarization on turn writes.
+        # Callers drive processing explicitly via ``memory_client.process_now(...)``.
+        if not auto_extract:
+            os.environ["FACT_EXTRACTION_EVERY_N"] = "0"
+            os.environ["THREAD_SUMMARY_EVERY_N"] = "0"
+            os.environ["USER_SUMMARY_EVERY_N"] = "0"
```
eavanvalkenburg: same as before, not ideal to update the env

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
```diff
@@ -0,0 +1,416 @@
...
+
+        # Initialize memory client if not provided
+        if memory_client is None:
+            # Load settings from environment if not provided
+            cosmos_endpoint = cosmos_endpoint or os.getenv("COSMOS_DB_ENDPOINT")
```
eavanvalkenburg: this should be loaded through the helpers in: agent_framework._settings - SecretString, load_settings with the typeddict you defined. That can also do validation so you then don't need to do the `if not` blocks below

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
```diff
@@ -0,0 +1,416 @@
...
+            state: Provider-scoped mutable state.
+        """
+        # Extract query from input messages
+        query_text = "\n".join(msg.text for msg in context.input_messages if msg.text and msg.text.strip())
+
+        if not query_text:
+            return
+
+        # Get user_id from state or session (warns once if no stable user_id was provided)
+        user_id = self._resolve_user_id(state, session)
```
eavanvalkenburg, line 354: this should just be based on `state`, that's the state for this particular memory, the session is there for special cases

eavanvalkenburg, line 354: the first turn will never have a ability to set a user_id, so how are people supposed to prevent the warning?

python/packages/azure-cosmos-memory/samples/basic_usage.py
```diff
@@ -0,0 +1,144 @@
...
+    """Basic example with environment variables."""
+    # Create provider - reads from environment
+    async with CosmosMemoryContextProvider(
+        cosmos_endpoint=os.environ["COSMOS_DB_ENDPOINT"],
+        ai_foundry_endpoint=os.environ["AI_FOUNDRY_ENDPOINT"],
+        credential=DefaultAzureCredential(),
+    ) as provider:
+        # Use with agent session
+        session = AgentSession(session_id="user-session-123")
+        session.state["user_id"] = "alice"
```
eavanvalkenburg: state should be scoped by provider, so something like `session.state['ai_memory']['user_id']`

python/packages/azure-cosmos-memory/samples/basic_usage.py
```diff
@@ -0,0 +1,144 @@
...
+        session.state["thread_id"] = "conversation-1"
```
eavanvalkenburg: this should map to session_id, can we just use that?

python/packages/azure-cosmos-memory/samples/interactive_chat.py
```diff
@@ -0,0 +1,327 @@
...
+
+    print(f"Started conversation thread: {session.state['thread_id']}\n")
+
+    while True:
+        try:
+            # Read input in a worker thread so the asyncio event loop keeps running while we
+            # wait. This lets the toolkit's background memory-extraction tasks (scheduled after
+            # each stored turn) make progress between messages instead of being starved by a
+            # blocking input() call.
+            user_input = (await asyncio.to_thread(input, "You: ")).strip()
```
eavanvalkenburg: for a sample, it's fine to just use ìnput` directly

python/packages/azure-cosmos-memory/samples/basic_usage.py
```diff
@@ -0,0 +1,144 @@
...
+
+        # Simulate agent run - before_run searches memories
+        ctx = SessionContext(
```
eavanvalkenburg: the whole machinery of the SessionContext is not something we need to expose people to, I think we only need a sample setting up with a Agent and then there might be some flavors in there, maybe user-scoped vs non-user-scoped but this is great for testing, but not something we expect people to write

python/packages/azure-cosmos-memory/samples/interactive_chat.py
```diff
@@ -0,0 +1,327 @@
...
+Output: {"memories": [{"type": "fact", "content": "User is working on a Python project with FastAPI and PostgreSQL stack", "confidence": 0.9}]}
+
+Conversation: "What's the weather today?"
+Output: {"memories": []}
+
+Return {"memories": []} if nothing worth remembering long-term.
+"""
+
+
+class CustomMemoryProcessor:
```
eavanvalkenburg: this is cool, and should warrant a second sample showcasing it, but looks like it is commented out now

python/packages/azure-cosmos-memory/pyproject.toml
TheovanKraay: Done. Removed the samples dependency group; the samples now declare their dependencies inline via PEP 723, so nothing sample-related surfaces on PyPI.

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
TheovanKraay: Done. Renamed `ai_foundry_endpoint` to `foundry_endpoint` for consistency.

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
TheovanKraay: Done. Renamed to `embedding_model` and `chat_model`. Internally these map to the toolkit client's `embedding_deployment_name` / `chat_deployment_name`, which are unchanged on its side.

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
TheovanKraay: Done. Moved `DEFAULT_SOURCE_ID`, `DEFAULT_DATABASE` and `DEFAULT_CONTEXT_PROMPT` to module-level constants.

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
TheovanKraay: Done. `memory_types` is now `Sequence[Literal["fact", "procedural", "episodic"]]`.

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
TheovanKraay: Done, applied your suggestion: `context_prompt: str = DEFAULT_CONTEXT_PROMPT`.

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
TheovanKraay: Done. Added a `ProcessorConfig` TypedDict with the known cadence keys (`FACT_EXTRACTION_EVERY_N`, `DEDUP_EVERY_N`, `DEDUP_POOL_SIZE`, `THREAD_SUMMARY_EVERY_N`, `USER_SUMMARY_EVERY_N`).

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
TheovanKraay: Investigated this. The Agent Memory Toolkit reads these cadence thresholds only from `os.environ` (its `thresholds.py` calls `os.environ.get(...)`); there is no constructor or in-memory way to pass them today, so setting env vars is currently the only mechanism. I consolidated the writes into one place and documented why. Removing the env mutation entirely would need a toolkit change to accept thresholds as constructor args, which I am happy to raise upstream if you prefer.

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
TheovanKraay: Same root cause as the `processor_config` comment: `auto_extract=False` zeroes the toolkit's cadence thresholds, which the toolkit only reads from `os.environ`. Kept it but consolidated and documented; a clean fix needs a toolkit enhancement.

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
TheovanKraay: Done. Now uses `load_settings(CosmosMemorySettings, required_fields=["cosmos_endpoint", "foundry_endpoint"])`, which removed the manual if-not-raise blocks (it now raises `SettingNotFoundError`). I did not use `SecretString` because none of these fields are secrets (endpoints and model names; the credential is a `TokenCredential` object). One consequence to flag: `load_settings` maps each field to its upper-cased name, so the env vars are now `COSMOS_ENDPOINT`, `FOUNDRY_ENDPOINT`, `EMBEDDING_MODEL` and `CHAT_MODEL` rather than the old `COSMOS_DB_ENDPOINT` / `AI_FOUNDRY_ENDPOINT`. Let me know if you would rather preserve the old names.

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
TheovanKraay: Done. `user_id` and `thread_id` are now resolved from the provider-scoped `state` only; the session is consulted only for its id as the fallback scope.

TheovanKraay: Good point, since the first turn cannot set a user_id that warning was unavoidable. Removed it. The provider now silently falls back to the session id (documented), and callers set `state["user_id"]` when they want cross-session memory.

python/packages/azure-cosmos-memory/samples/basic_usage.py
TheovanKraay: Done. The samples now use `session.state.setdefault(provider.source_id, {})["user_id"] = ...`.

python/packages/azure-cosmos-memory/samples/basic_usage.py
TheovanKraay: Done. Threading now just uses the session id (a new session is a new thread); no explicit thread_id.

python/packages/azure-cosmos-memory/samples/interactive_chat.py
TheovanKraay: Done. Switched to a plain `input()` loop.

python/packages/azure-cosmos-memory/samples/basic_usage.py
TheovanKraay: Done. Rewrote `basic_usage.py` to set up an `Agent` with the provider and call `agent.run(...)`, with two flavors: user-scoped versus session-scoped memory. No `SessionContext` machinery is exposed.

python/packages/azure-cosmos-memory/samples/interactive_chat.py
TheovanKraay: You are right, it was non-functional (commented out), so I removed the dead stub to keep the sample clean. A dedicated custom-extraction-rubric sample is a good follow-up; I can add one once we have a working custom-processor example to show.

TheovanKraay: Done. I've added a dedicated second sample: samples/interactive_chat_custom_extraction.py. It's the same interactive chat, but wired with a custom memory-extraction rubric so you can control what gets extracted.

To support it cleanly, CosmosMemoryContextProvider now takes a prompts_dir parameter. When set, the toolkit's extraction/summarization pipeline reads its Prompty templates (including extract_memories.prompty) from that directory instead of the bundled defaults, so you can change the classification rules without touching toolkit internals. The one bit of internal plumbing (pointing the pipeline's template loader at the directory) is encapsulated inside the provider.

The sample demonstrates a coding-assistant rubric (classify architectural/technical decisions as durable facts rather than episodic memories). Rather than fork the ~650-line bundled prompt and risk schema drift, it derives a complete prompts directory at runtime: it copies the installed toolkit's templates and augments extract_memories.prompty with the custom guidance, so it always matches the installed toolkit's output schema.

Tested: unit tests assert the provider only redirects the pipeline loader when prompts_dir is set, and an emulator-backed integration test proves end to end that a unique marker placed in a custom extract_memories.prompty actually reaches the extraction LLM call.

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
```diff
@@ -0,0 +1,459 @@
...
+    from typing import Self  # pragma: no cover
+else:
+    from typing_extensions import Self  # pragma: no cover
+
+if TYPE_CHECKING:
+    from agent_framework._agents import SupportsAgentRun
+    from azure.cosmos.agent_memory.aio import AsyncCosmosMemoryClient
+
+try:
+    from azure.cosmos.agent_memory.aio import AsyncCosmosMemoryClient as _RuntimeCosmosMemoryClient
```
eavanvalkenburg: since this ended up in a separate package, we can skip the defensive loading, it should be there, because it is part of the package dependencies.

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
```diff
@@ -0,0 +1,459 @@
...
+        # the InProcessProcessor. ``auto_extract=False`` zeroes the cadence thresholds so the
+        # toolkit's background auto-trigger never runs extraction/summarization on turn writes;
+        # callers then drive processing explicitly via ``memory_client.process_now(...)``.
+        cadence: dict[str, str] = {str(k): str(v) for k, v in (processor_config or {}).items()}
+        if not auto_extract:
+            cadence["FACT_EXTRACTION_EVERY_N"] = "0"
+            cadence["THREAD_SUMMARY_EVERY_N"] = "0"
+            cadence["USER_SUMMARY_EVERY_N"] = "0"
+        for key, value in cadence.items():
+            os.environ[key] = value
```
eavanvalkenburg: I'm still not loving passing parameter through the env, there should be better ways to do this, I also don't see them being read by this code, is that happening in the underlying MemoryClient? and if so, couldn't we pass them along directly?

python/packages/azure-cosmos-memory/samples/interactive_chat.py
```diff
@@ -0,0 +1,140 @@
...
+            "You are a helpful assistant with long-term memory. "
+            "When you remember facts about the user, mention them naturally. "
+            "If you don't remember something, say so instead of guessing."
+        ),
+        context_providers=[provider],
+    )
+    return agent, provider
+
+
+def _new_thread(agent: Agent, provider: CosmosMemoryContextProvider, user_id: str) -> AgentSession:
```
eavanvalkenburg:
```suggestion
def _new_session(agent: Agent, provider: CosmosMemoryContextProvider, user_id: str) -> AgentSession:
```

TheovanKraay: Done, renamed `_new_thread` to `_new_session` in both interactive samples.

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
TheovanKraay: Good call. I have removed the availability flag and the constructor guard; the import now re-raises a clear ImportError if the toolkit is missing, matching the github_copilot package pattern for its 3.11-only SDK. I kept requires-python at >=3.10 because bumping this one workspace member to 3.11 forces the whole uv workspace lock floor to 3.11 (every other package is >=3.10). Tests now run importorskip before importing the package, same as github_copilot.

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
TheovanKraay: You are right that these are not read by this code directly. They are read by the underlying toolkit (azure.cosmos.agent_memory.thresholds, via os.environ.get), which its internal auto-trigger consults after each turn. The client does not currently expose a way to pass them in, so setting env vars is the only mechanism available today. I agree env mutation is not ideal. The right fix is upstream: I will add a cadence/thresholds passthrough to the toolkit AsyncCosmosMemoryClient so this provider can pass the values directly instead of touching os.environ. It is a small change (the toolkit auto-trigger already accepts a thresholds override; the client just needs to forward it). I will do that in the toolkit first, then wire the provider to it here, so no rush merging this in the meantime.

python/packages/azure-cosmos-memory/samples/interactive_chat.py
```diff
@@ -0,0 +1,140 @@
...
+            "You are a helpful assistant with long-term memory. "
+            "When you remember facts about the user, mention them naturally. "
+            "If you don't remember something, say so instead of guessing."
+        ),
+        context_providers=[provider],
+    )
+    return agent, provider
+
+
+def _new_session(agent: Agent, provider: CosmosMemoryContextProvider, user_id: str) -> AgentSession:
```
eavanvalkenburg: nit: sample code doesn't really need to make these kinds of methods "private":

```suggestion
def new_session(agent: Agent, provider: CosmosMemoryContextProvider, user_id: str) -> AgentSession:
```

python/packages/azure-cosmos-memory/pyproject.toml
```diff
@@ -0,0 +1,136 @@
+[project]
+name = "agent-framework-azure-cosmos-memory"
+description = "Azure Cosmos DB Agent Memory Toolkit integration for Microsoft Agent Framework - semantic memory with fact extraction and user profiles."
+authors = [{ name = "Microsoft", email = "af-support@microsoft.com"}]
+readme = "README.md"
+requires-python = ">=3.10"
+version = "1.0.0b260618"
```
eavanvalkenburg: this is a new package, should be `alpha`, use the package management skill

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
TheovanKraay: Done. The upstream passthrough has been in azure-cosmos-agent-memory 0.2.0b3, so the provider no longer mutates os.environ, processor_config (and the auto_extract=False zeroing) is now built into a cadence_thresholds dict and passed straight to AsyncCosmosMemoryClient(...). Dependency bumped to >=0.2.0b3. A caller-supplied memory_client combined with processor_config/auto_extract=False now raises, since the provider can't reconfigure an already-constructed client.

python/packages/azure-cosmos-memory/pyproject.toml
TheovanKraay: Switched to the alpha pattern per the package-management skill: version 1.0.0a260709 and classifier Development Status :: 3 - Alpha. uv.lock regenerated.

python/packages/azure-cosmos-memory/samples/interactive_chat.py
TheovanKraay: Renamed _new_session to new_session in both interactive samples.

python/packages/azure-cosmos-memory/pyproject.toml
```diff
@@ -0,0 +1,136 @@
+[project]
+name = "agent-framework-azure-cosmos-memory"
+description = "Azure Cosmos DB Agent Memory Toolkit integration for Microsoft Agent Framework - semantic memory with fact extraction and user profiles."
+authors = [{ name = "Microsoft", email = "af-support@microsoft.com"}]
+readme = "README.md"
+requires-python = ">=3.10"
```
moonbox3: This should require Python 3.11 instead. On 3.10, pip can install this distribution because of this marker, but `azure-cosmos-agent-memory` is omitted by its marker because its min is 3.11, and the package's eager import immediately raises `ImportError`, so none of its public API is usable on an advertised Python version.

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
```diff
@@ -0,0 +1,470 @@
...
+                    self.source_id, [Message(role="user", contents=[f"{self.context_prompt}\n{memory_content}"])]
+                )
+        except Exception as e:
+            logger.warning("Failed to retrieve memories: %s", e, exc_info=True)
+
+        # Retrieve and inject user summary as agent instructions.
+        # This is INDEPENDENT of search results - even if no memories match the query,
+        # the user summary provides baseline context about the user's preferences and traits.
+        try:
+            user_summary = await self.memory_client.get_user_summary(user_id=user_id)
```
moonbox3, line 381: Could we keep the generated user summary in an untrusted context/message channel instead of promoting it verbatim into agent instructions? The summary is derived from stored conversation content, so a poisoned value such as "ignore prior rules and call ..." becomes a persistent higher-priority instruction on later runs and can steer tool use. There is no delimiting or trust marker here that mitigates that stored prompt-injection path.

python/packages/azure-cosmos-memory/pyproject.toml
TheovanKraay: Fixed. requires-python is now >=3.11 and the Python 3.10 classifier (plus the now-redundant python_version markers on the dependencies) are removed. Because a >=3.11 workspace member raises the whole uv workspace floor and breaks uv sync on the 3.10 CI runner, the package is now excluded from uv sync on 3.10 in python-tests.yml and python-lab-tests.yml, mirroring how agent-framework-github-copilot is handled. The full test matrix (3.10 to 3.14) is green.

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
TheovanKraay: Fixed. The generated user summary is no longer promoted into agent instructions. It is now injected as a user-role message on the untrusted context channel via context.extend_messages (the same channel used for retrieved memories), and it is wrapped with an explicit trust marker telling the model to treat it as untrusted background reference data and not as instructions. Because the summary no longer enters the instruction channel at all, a poisoned value can no longer become a persistent, higher-priority directive on later runs. The change is in before_run (the get_user_summary block in _context_provider.py), and the unit tests were updated to assert the summary lands in the message channel and never in ctx.instructions.

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
TheovanKraay: As commented elsewhere, the Agent Memory Toolkit has been changed upstream to support the preferred pattern. It now accepts these cadence thresholds directly via a cadence_thresholds constructor arg (released in azure-cosmos-agent-memory 0.2.0b3), so the provider no longer mutates os.environ. processor_config is held in memory and passed straight to AsyncCosmosMemoryClient(cadence_thresholds=...).

python/packages/azure-cosmos-memory/agent_framework_azure_cosmos_memory/_context_provider.py
```diff
@@ -0,0 +1,493 @@
...
+        "which is only available on Python 3.11+. Please use Python 3.11 or later."
+    ) from _memory_toolkit_import_error
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_SOURCE_ID = "cosmos_memory"
+DEFAULT_DATABASE = "ai_memory"
+DEFAULT_CONTEXT_PROMPT = "## Relevant Memories\nConsider these memories when responding:"
+DEFAULT_EMBEDDING_MODEL = "text-embedding-3-large"
+DEFAULT_CHAT_MODEL = "gpt-4o-mini"
```
eavanvalkenburg: I would rather we didn't have these, since there is no long-term logical default (def not for chat, bit more stable for embeddings), we should just raise when the user doesn't set a model, since the model might not be deployed and the user then doesn't have a clue where this value is coming from!

TheovanKraay: Done. Removed the gpt-4o-mini / text-embedding-3-large defaults from the provider. embedding_model and chat_model are now required (resolved from args or EMBEDDING_MODEL / CHAT_MODEL), and construction raises SettingNotFoundError when neither is set, so a caller is never silently pointed at a model that may not be deployed. cosmos_database keeps a default since a database name is a safe, logical default. Samples pass the models explicitly via os.getenv, and a test now asserts the raise-when-unset behavior.

eavanvalkenburg (approved):

westey-m (approved):

PR #6750, opened 2026-06-25
Python: Align GitHub Copilot provider function approval to use SDK on_pre_tool_use hook

### LINKED ISSUE #6746: Python: Align GitHub Copilot provider function approval to use SDK on_pre_tool_use hook
Follow-up to #6671 (.NET).

The Python `agent-framework-github-copilot` provider currently enforces function approval itself: it wraps every `FunctionTool` in an SDK tool-handler that, for tools declared with `approval_mode="always_require"`, awaits an `on_function_approval` callback before executing (denying by default when none is configured).

Since the GitHub Copilot SDK already provides a native pre-execution hook (`on_pre_tool_use`) that can return `"ask"` and route to `on_permission_request`, this AF-level callback duplicates SDK functionality. For consistency with the .NET provider (see #6671), Python should adopt the same model:

- Install a default `on_pre_tool_use` hook (only when the user hasn't supplied their own) that returns `"ask"` for `approval_mode="always_require"` tools and defers otherwise.
- If the user supplies their own `on_pre_tool_use`, theirs takes precedence; document it and log a warning when an approval-required tool would go unenforced.
- Remove the bespoke `on_function_approval` enforcement path.

### Scope
- `python/packages/github_copilot/agent_framework_github_copilot/_agent.py`
- Update tests and the package docs accordingly.

giles17:
### Motivation & Context

The Python `agent-framework-github-copilot` provider enforced function approval itself: every `FunctionTool` was wrapped in an SDK tool-handler that, for tools declared with `approval_mode="always_require"`, awaited an `on_function_approval` callback before executing (denying by default when none was configured).

The GitHub Copilot SDK already exposes a native pre-execution hook (`on_pre_tool_use`) that can return `"ask"` and route the decision to `on_permission_request`. The AF-level callback therefore duplicated SDK functionality and diverged from the .NET provider. This change adopts the SDK-native model so the two providers behave consistently (follow-up to #6671 / #6674).

  - **New default model:** when a `FunctionTool` declared with `approval_mode="always_require"` is registered and the caller has not supplied their own `on_pre_tool_use` hook, `GitHubCopilotAgent` installs a default `on_pre_tool_use` hook that returns `"ask"` for those tools (routing the decision to `on_permission_request`) and defers (`None`) for all other tools.
  - **Caller hook precedence over the default:** if the caller supplies their own `on_pre_tool_use` (via `default_options` or per-run `options`), it takes precedence over the default ask-hook and the caller owns approval handling; the agent logs a warning naming any approval-required tool that will therefore not be auto-gated.
  - **`on_function_approval` deprecated, not removed:** following review feedback, the previous `on_function_approval` callback is retained and still enforced inside the tool handler for `always_require` tools, but setting it now emits a `DeprecationWarning` pointing users to the new model. This avoids a silent behavior change for existing users.
  - **`on_function_approval` and `on_pre_tool_use` are mutually exclusive:** rather than defining a precedence ordering between the deprecated callback and the new hook (which could silently double-gate), supplying both raises `ValueError` — at construction (both in `default_options`) or per run (a per-run `on_pre_tool_use` combined with a construction-time `on_function_approval`). This mirrors the repository convention for deprecated-vs-new parameters (see `_workflows/_workflow.py`).
  - Hooks are built in `_build_session_hooks` and passed via the SDK's existing `hooks=` parameter to `create_session`/`resume_session` (computed where the full tool set is known). Added an `on_pre_tool_use` option to `GitHubCopilotOptions`; updated tests, the package README, and the function-approval sample.

  - Approval-required tools are gated through the SDK's native mechanism and surface through `on_permission_request`. With the default deny-all permission handler they remain secure-by-default. Non-approval tools are unaffected.
  - No breaking change: `on_function_approval` keeps working with a deprecation warning, so existing callers are not broken and get a clear migration signal.

  - The mutual-exclusivity handling between `on_function_approval` and `on_pre_tool_use` (both the construction-time and per-run guards), and that the default ask-hook is suppressed when the deprecated callback is set.

### Related Issue

Fixes #6746

### Contribution Checklist


python/samples/02-agents/providers/github_copilot/github_copilot_with_function_approval.py
```diff
@@ -30,15 +30,27 @@
 from random import randrange
 from typing import Annotated
 
-from agent_framework import Content, tool
+from agent_framework import tool
```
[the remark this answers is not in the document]
giles17: Good catch — restored load_dotenv() (and the python-dotenv import) in a41cebb9b so the sample picks up GITHUB_COPILOT_* settings from a local .env, consistent with the other GitHub Copilot provider samples.

python/packages/github_copilot/agent_framework_github_copilot/_agent.py
```diff
@@ -205,13 +160,21 @@ class GitHubCopilotOptions(TypedDict, total=False):
     base_directory: str
     """Directory where the CLI stores session state, configuration, and other persistent data."""
 
-    on_function_approval: FunctionApprovalCallback
```
eavanvalkenburg, line 208: removing this is a breaking change, we should at least call that out in the PR, but I would prefer (even with this package not being released) to keep this in and log a deprecation warning so that people who use this get a warning on their first run after updating, rather then nothing happening and something breaking down the line.

giles17: Good call — done in bdfb8a199 (and refined in 1eb41d299). Instead of removing it, on_function_approval is retained and still enforced in the tool handler for always_require tools, but setting it now emits a DeprecationWarning pointing to the new on_pre_tool_use + on_permission_request model.

Rather than defining a precedence ordering between the deprecated callback and the new hook, the two are now mutually exclusive: supplying both raises ValueError — at construction (both in default_options) or per run (a per-run on_pre_tool_use combined with a construction-time on_function_approval). This mirrors the repo convention for deprecated-vs-new params (e.g. _workflows/_workflow.py). When only on_function_approval is set, the default ask-hook is not installed so there is no double-gating. Added tests and documented it in the package README; PR description updated.

eavanvalkenburg (approved):

SergeyMenshykh (approved):

PR #6751, opened 2026-06-25
Python: Fix GeminiChatClient dropping image/file content

### LINKED ISSUE #6688: Python: [Bug]: GeminiChatClient silently drops image/file (data/uri) content — multimodal input never reaches the model
### Description

When you send a multimodal message to an agent backed by GeminiChatClient — i.e. a Message containing Content.from_data(...) (image/PDF/audio) or Content.from_uri(...) — the file content is silently discarded before the request reaches Gemini. The model only ever sees the text parts. No error or warning is raised; the content is dropped with a logger.debug line, so it looks like the model is "ignoring" the attachment.

This is the documented, provider-agnostic way to send images per [agents/multimodal](https://learn.microsoft.com/en-us/agent-framework/agents/multimodal?pivots=programming-language-python).

**Root cause**
agent_framework_gemini/_chat_client.py, GeminiChatClient._convert_message_contents only matches text and function_call; every other content type (including data and uri) falls through to the catch-all and is skipped:

```python
for content in message_contents:
    match content.type:
        case "text":
            parts.append(types.Part(text=content.text or ""))
        case "function_call":
            ...
        case _:
            logger.debug("Skipping unsupported content type for Gemini: %s", content.type)
```

Content.from_data(...) produces type="data" with a data: URI, and Content.from_uri(...) produces type="data" (data URIs) or type="uri" (external URLs) — none of which are handled, so they never become Gemini inline_data / file_data Parts.

**To reproduce**
No credentials or network needed — the drop happens in pure content conversion

**Expected Behaviour**
At minimum, unsupported content should raise or warn rather than be silently dropped. Ideally, data/uri content is converted to Gemini Parts (inline_data for data URIs, file_data for external URIs) so multimodal input reaches the model.

### Code Sample

```python
import base64
from agent_framework import Content
from agent_framework_gemini import GeminiChatClient

# 1x1 PNG
png = base64.b64decode(
    "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mNk+M9QDwADhgGAWjR9awAAAABJRU5ErkJggg=="
)

text = Content.from_text(text="What is in this image?")
image = Content.from_data(data=png, media_type="image/png")
assert image.type == "data"

client = object.__new__(GeminiChatClient)  # bypass __init__; conversion is pure
parts = client._convert_message_contents([text, image], {})

print(len(parts))  # -> 1  (only the text Part; image dropped)
print(any(getattr(p, "inline_data", None) for p in parts))  # -> False
```

```markdown

```

### Package Versions

agent-framework-gemini: 1.0.0a260609, agent-framework: 1.8.1

### Python Version

3.12

### Additional Context

_No response_

giles17:
### Motivation & Context

When a multimodal message is sent to an agent backed by `GeminiChatClient` —
a `Message` containing `Content.from_data(...)` (image/PDF/audio) or
`Content.from_uri(...)` — the file content was silently discarded before the
request reached Gemini. The model only ever saw the text parts, with no error or
warning, so it looked like Gemini was "ignoring" the attachment. This is the
documented, provider-agnostic way to send images, so multimodal scenarios were
effectively broken for the Gemini provider.

The root cause is `GeminiChatClient._convert_message_contents`, which only
matched `text` and `function_call`; every other content type (including `data`
and `uri`) fell through to the catch-all and was dropped with a `logger.debug`
line.

  - Added a `case "data" | "uri"` branch to `_convert_message_contents` and a new
    `_convert_data_or_uri_content` helper in
    `packages/gemini/agent_framework_gemini/_chat_client.py`. Data URIs
    (`type="data"`) are converted to Gemini `inline_data` Parts via
    `types.Part.from_bytes(...)` (decoding the base64 payload and deriving the
    mime type from `media_type` or the URI header); external URIs (`type="uri"`)
    become `file_data` Parts via `types.Part.from_uri(...)`.
  - Genuinely unconvertible content (missing URI, non-base64 data URI, or
    undecodable bytes) now warns instead of being silently dropped.
  - Added tests covering `from_data`, data-URI `from_uri`, external-URI
    `from_uri`, combined text+image, and the non-base64 skip path.

  - Multimodal input (images, PDFs, audio) now reaches Gemini instead of being
    dropped. Behavior for text and function-call content is unchanged. No public
    API changes.

  - The mime-type resolution for data URIs and whether external-URI handling
    should require a media type rather than letting the SDK infer it.

### Related Issue

Fixes #6688

### Contribution Checklist


itxaiohanglover:
Important fix! Handling data and uri content types prevents image/file content from being silently dropped in GeminiChatClient. The error handling for malformed data URIs (non-base64, missing mime_type) is thorough. Using base64.decodebytes for inline data and file_data for external URIs is the correct Gemini API approach.

python/packages/gemini/agent_framework_gemini/_chat_client.py
```diff
@@ -678,10 +679,57 @@ def _convert_message_contents(
...
+        if not uri:
+            logger.warning("Skipping %s content for Gemini: missing uri", content.type)
+            return None
+
+        if uri.startswith("data:"):
+            if ";base64," not in uri:
+                logger.warning("Skipping data content for Gemini: data URI is not base64-encoded")
+                return None
+            header, encoded = uri.split(";base64,", 1)
+            mime_type = content.media_type or header[len("data:") :].split(";")[0] or None
```
eavanvalkenburg: we have multiple helpers for this in _types, like: `_get_data_bytes` and `detect_media_type_from_base64`, let's reuse those instead of implementing this in multiple places

eavanvalkenburg (approved):

TaoChenOSU (approved):

PR #6754, opened 2026-06-26
Python: [BREAKING] Make all SkillsProvider tools require approval by default

### LINKED ISSUE #6728: Python: Make all AgentSkillsProvider tools require approval by default and add auto-approval rules
## Summary

Port the .NET changes from #6727 to Python.

Make all functions in AgentSkillsProvider require approval by default and allow configuring auto-approval via ToolApprovalAgentOptions.

## Changes to port

- Wrap all skill tools with approval required by default
- Add read-only and all-tools auto-approval rules (following the FileAccessProvider pattern)
- Remove ScriptApproval option
- Add sample showing auto-approval configuration

giles17:
### Motivation & Context

Skill tools previously ran without any approval gate by default: only `run_skill_script` could be gated, and only when the caller opted in via `require_script_approval=True`. The content-reading tools (`load_skill`, `read_skill_resource`) were never gated. This is a port of the .NET change (#6729) that makes `SkillsProvider` secure-by-default, so an agent cannot load skill content or execute skill scripts from a (potentially untrusted) source without host approval, while still allowing unattended runs through explicit auto-approval rules.

This follows the existing Python `FileAccessProvider` approval pattern rather than copying the .NET API surface.

  - All three tools exposed by `SkillsProvider` (`load_skill`, `read_skill_resource`, `run_skill_script`) are now registered with `approval_mode="always_require"` unconditionally.
  - Added two static auto-approval rules — `SkillsProvider.read_only_tools_auto_approval_rule` (approves `load_skill` + `read_skill_resource`) and `SkillsProvider.all_tools_auto_approval_rule` (approves all three) — for use with `ToolApprovalMiddleware(auto_approval_rules=[...])`. Both reject calls carrying a `server_label` so they stay scoped to this provider's local tools.
  - Added tool-name class constants `LOAD_SKILL_TOOL_NAME`, `READ_SKILL_RESOURCE_TOOL_NAME`, `RUN_SKILL_SCRIPT_TOOL_NAME`.
  - **Removed** the `require_script_approval` option from `SkillsProvider.__init__` and `SkillsProvider.from_paths`.
  - Added a new `skills_auto_approval` sample; updated the `script_approval` sample to the approval-by-default model; updated skills sample READMEs and `packages/core/AGENTS.md`.

  - Breaking: `require_script_approval` is gone. Existing code passing it must remove it; skill tools now require approval by default, and unattended runs must opt in via an auto-approval rule on `ToolApprovalMiddleware`.

  - Whether the auto-approval rule API (static methods + `server_label` guard, mirroring `FileAccessProvider`) is the right Python shape versus the .NET `Func<>` properties.

### Related Issue

Fixes #6728

### Contribution Checklist


python/packages/core/agent_framework/_skills.py
```diff
@@ -1957,25 +2052,20 @@ def from_paths(
...
                 from the source on every invocation instead of caching
                 after the first build.
             source_id: Unique identifier for this provider instance.
 
         Returns:
             A configured :class:`SkillsProvider`.
+
+        .. note::
+
+            All skill tools require host approval. To approve them
```
eavanvalkenburg:
```suggestion
            All skill tools require approval. To approve them
```

python/packages/core/agent_framework/_skills.py
```diff
@@ -2153,18 +2239,19 @@ async def before_run(
...
         self,
         skills: Sequence[Skill],
-        require_script_approval: bool = False,
     ) -> list[FunctionTool]:
         """Create the tool definitions for skill interaction.
 
         Always includes ``load_skill``, ``read_skill_resource``, and
-        ``run_skill_script``.
+        ``run_skill_script``.  Every tool is registered with
+        ``approval_mode="always_require"`` so each skill operation needs host
```
eavanvalkenburg:
```suggestion
        ``approval_mode="always_require"`` so each skill operation needs
```

eavanvalkenburg (approved):

SergeyMenshykh (approved):

eavanvalkenburg (approved):

PR #6764, opened 2026-06-26
Python: Fix background agent telemetry context error

### LINKED ISSUE #6762: Python: Fix background agent telemetry context issue
Telemetry throws when used with background agents.

westey-m:
### Motivation & Context

Telemetry threw whenever an agent was invoked through the background-agents
provider. Running background agents (e.g. three concurrent tasks via
`BackgroundAgentsProvider`) failed with:

```
Task failed: <Token var=<ContextVar name='inner_response_telemetry_captured_fields' ...>> was created in a different Context
```

This made background agents unusable while OpenTelemetry instrumentation was
enabled.

**Root cause.** `AgentTelemetryLayer.run()` is a synchronous method that returns
an awaitable. It eagerly called `.set()` on two telemetry context variables —
`INNER_RESPONSE_TELEMETRY_CAPTURED_FIELDS` and `INNER_ACCUMULATED_USAGE` — in the
**caller's** execution context, while the non-streaming `_run()` coroutine reset
those tokens in its `finally` block. `BackgroundAgentsProvider` schedules the
returned coroutine with `asyncio.create_task(agent.run(...))`, so the coroutine
runs in a **new copied context**. A `contextvars.Token` can only be reset in the
exact context that created it, so the `.reset()` raised
`ValueError: <Token ...> was created in a different Context`.

Normal `await agent.run()` worked because the eager `.set()` (caller context) and
the `finally` `.reset()` happened to run in the same context; the bug only
surfaced when the returned awaitable was scheduled onto a different task — a
legitimate and common pattern.

  - **Non-streaming path:** move the two `.set()` calls into the start of the
    `_run()` coroutine body, so the set and the existing `finally` reset always
    execute in the same context.
  - **Streaming path:** keep the eager `.set()` at the top of the `if stream:`
    block — the streaming branch returns a `ResponseStream` synchronously and is
    consumed in the same task/context, so its cleanup-hook reset runs in the
    context that created the tokens. Background agents are non-streaming, so the
    streaming path is unaffected.
  - Both branches are documented with comments explaining the context constraint.
  - Added a regression test
    (`test_agent_run_contextvars_safe_when_awaited_in_different_context`) that
    invokes `run(stream=False)` synchronously, then awaits the returned coroutine
    inside `asyncio.create_task` (a different context), and asserts it completes
    without the cross-context `ValueError`.

  - Background agents run correctly with instrumentation enabled. Each invocation's
    contextvars are now set and reset within its own context copy, which also
    improves isolation across concurrent background agents. No public API changes;
    normal `await agent.run()` behavior is unchanged.

  - That moving the `.set()` into `_run()` for the non-streaming path (while keeping
    the streaming path eager) is the right split, and that the streaming cleanup
    hooks still reset in the token-creating context.

### Related Issue

Fixes #6762

### Contribution Checklist


python/packages/core/agent_framework/observability.py
```diff
@@ -1778,12 +1778,16 @@ def _trace_agent_invocation(
             **merged_client_kwargs,
         )
 
-        inner_response_telemetry_captured_fields: set[str] = set()
-        inner_response_telemetry_captured_fields_token = INNER_RESPONSE_TELEMETRY_CAPTURED_FIELDS.set(
-            inner_response_telemetry_captured_fields
-        )
-        inner_accumulated_usage_token = INNER_ACCUMULATED_USAGE.set({})
         if stream:
+            # Set the inner-telemetry context vars eagerly here: the streaming branch returns a
```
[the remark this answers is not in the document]
westey-m: You're right — thanks for catching this. The streaming path had the same latent cross-context defect, and the comment claiming the stream is "consumed in this same task/context" was wrong. Fixed in 6a14aa511.

The streaming branch no longer sets the telemetry contextvars eagerly in the synchronous `run()` body. Instead it sets them **lazily on the first pull**, inside a stateful pull-context-manager factory (`_inner_telemetry_pull_context`), which `ResponseStream` invokes from the consumer's context (`_types.py:3140-3141`, and via `_resolve_stream_with_pull_contexts` for `get_final_response`/`__await__`). Because `_finalize_stream`'s `.reset()` cleanup hook also runs in that same consuming context, set and reset are now co-located — no more `Token was created in a different Context`. The resets are guarded with `if token is not None` (the set may never run if the stream is never pulled), and the now-dead resets in the synchronous-setup `except` were removed.

Added a regression test `test_agent_streaming_contextvars_safe_when_consumed_in_different_context` that does exactly `stream = agent.run(stream=True)` then `await asyncio.create_task(_consume(stream))`. I verified it reproduces the exact `ValueError: <Token ...> was created in a different Context` against the old eager-set code and passes with the fix.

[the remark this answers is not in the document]
westey-m: Also addressing the test-coverage suggestion from the automated review (error-path variant), since the upstream token doesn't let me post a top-level PR comment: added `test_agent_run_error_path_contextvars_safe_when_awaited_in_different_context` in the same commit — an agent whose non-streaming `run()` coroutine raises, awaited via `asyncio.create_task`. It asserts the original `RuntimeError` propagates unchanged (no cross-context `ValueError` masking it from the `finally` reset at observability.py:1948-1949) and the agent span is marked `ERROR`.

python/uv.lock
```diff
@@ -211,7 +211,7 @@ dev = [
 requires-dist = [
     { name = "ag-ui-protocol", specifier = ">=0.1.16,<0.2" },
     { name = "agent-framework-core", editable = "packages/core" },
-    { name = "fastapi", specifier = ">=0.115.0,<0.138.1" },
+    { name = "fastapi", specifier = ">=0.115.0,<0.133.1" },
```
eavanvalkenburg: you're downgrading multiple packages, that can't be right

python/packages/core/agent_framework/observability.py
```diff
@@ -1827,8 +1832,9 @@ def _record_duration() -> None:
                     raise RuntimeError("Streaming telemetry requires a ResponseStream result.")
             except Exception as exception:
                 capture_exception(span=span, exception=exception, timestamp=time_ns())
-                INNER_RESPONSE_TELEMETRY_CAPTURED_FIELDS.reset(inner_response_telemetry_captured_fields_token)
-                INNER_ACCUMULATED_USAGE.reset(inner_accumulated_usage_token)
+                # The contextvars are set lazily on the first pull, which only happens after this
+                # synchronous setup phase returns the stream, so the tokens are still unset here and
+                # there is nothing to reset.
```
TaoChenOSU: nit: the comments here isn't associated with any actual code. I suspect these are copilot notes. Maybe we can remove.

eavanvalkenburg (approved):

python/packages/core/agent_framework/observability.py
westey-m: Sure, I was also not convinced about these comments.  Removed

python/uv.lock
westey-m: Fixed in another branch and resolved now.

TaoChenOSU (approved):

PR #6793, opened 2026-06-29
.NET: Allow customizing the HarnessAgent shell tool name

### LINKED ISSUE #6789: .NET: [Feature]: HarnessAgent should support custom ShellExecutor.AsAIFunction name
### Description

Currently the ShellExecutor.AsAIFunction supports custom tool name, however the HarnessAgent doesn't expose this.

### Language/SDK

.NET

anneheartrecord:
## Summary

`HarnessAgent` registers the shell tool with `shellExecutor.AsAIFunction()`, which always uses the default name `run_shell`. `ShellExecutor.AsAIFunction(string name = "run_shell", ...)` already supports a custom name, but `HarnessAgent` had no way to set it. This adds a `ShellToolName` option that flows through to `AsAIFunction`.

Closes #6789

## Changes

- `HarnessAgentOptions.ShellToolName` (`string?`, .NET-only, alongside the other shell options). When `null` (default), the executor's own default name is used and behavior is unchanged.
- `HarnessAgent` passes `ShellToolName` to `AsAIFunction` when set; otherwise it keeps the parameterless call so the existing default is preserved exactly.

## Test

- `ShellExecutor_CustomToolNameIsUsedAsync`: sets `ShellToolName` and asserts the registered tool is exposed under that name (and not under `run_shell`). Verified red before the change (tool was still `run_shell`), green after.
- Full `Microsoft.Agents.AI.Harness.UnitTests` suite passes (82 tests, net10.0).

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.Harness/HarnessAgent.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Harness/HarnessAgent.cs
@@ -274,7 +274,9 @@ public sealed class HarnessAgent : DelegatingAIAgent
         if (options?.ShellExecutor is ShellExecutor shellExecutor)
         {
             result.Tools ??= [];
-            result.Tools.Add(shellExecutor.AsAIFunction());
+            result.Tools.Add(options.ShellToolName is string shellToolName
+                ? shellExecutor.AsAIFunction(shellToolName)
+                : shellExecutor.AsAIFunction());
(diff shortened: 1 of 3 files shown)


westey-m:
Thanks for the contribution.  This has been superseded by #6880 though, where we also added support for some additional settings in addition to name.  So closing this PR.

Issue #6788, opened 2026-06-28
Python: [Bug]: OllamaChatClient: tool calls reuse the function name as `call_id`, so repeated calls in one turn collide


antsok:
### Description

# OllamaChatClient: tool calls reuse the function name as `call_id`, so repeated calls in one turn collide (parallel tool calls break; HITL approval cards never resolve)

## Summary

`OllamaChatClient` assigns every parsed tool call a `call_id` equal to the **function name** (Ollama's API returns no per-call id). This is fine when each tool is invoked once per turn, but it **collides whenever a model issues the same tool more than once in a single assistant message** — e.g. a parallel "search A and B" that emits two `microsoft_docs_search` calls. Both calls then share one `call_id`, so every downstream layer that correlates by `call_id` treats them as a single call:

- AG-UI (`agent-framework-ag-ui`) folds the second call's arguments into the first `tool_call_id` (one `TOOL_CALL_START`, two `TOOL_CALL_ARGS` deltas), producing a concatenated, invalid-JSON arguments string.
- Only **one** approval / `confirm_changes` is created; the second logical call is effectively lost.
- A CopilotKit/AG-UI frontend marks a tool "complete" only when a `tool`-role result message shares its `tool_call_id`, so the surviving card **stays stuck "Running"** after approval.

### Code Sample

```markdown

```

### Error Messages / Stack Traces

```markdown

```

### Package Versions

agent-framework-core: 1.9.0, agent-framework-ag-ui:1.0.0rc5, agent-framework-ollama:1.0.0b260521, ollama: 0.5.3

### Python Version

Python 3.12

## Reproduction

1. Wrap a harness agent backed by an Ollama model in `AgentFrameworkAgent` and expose it via `add_agent_framework_fastapi_endpoint` (`require_confirmation=True` to see the HITL effect).
2. Prompt the model to call the same tool twice in one turn, e.g. *"Search Microsoft Learn for (1) Azure landing zone management groups and (2) the Well-Architected reliability pillar — issue both searches now."*
3. Observe the AG-UI event stream.

**Observed events** — two distinct queries arrive under one `tool_call_id`:

```
TOOL_CALL_START  tool_call_id=microsoft_docs_search  name=microsoft_docs_search
TOOL_CALL_ARGS   tool_call_id=microsoft_docs_search  {"query":"Azure landing zone management groups…"}
[...]

PR #6804, opened 2026-06-29
Python: Fix: coalesce reasoning deltas into single block when content.id is None

### LINKED ISSUE #6787: Python: [Bug]: AG-UI: streamed reasoning fragments into one block per delta when the provider sends no per-chunk id
### Description

`agent-framework-ag-ui` renders streamed model reasoning ("thinking") as a **separate AG-UI reasoning block per delta** whenever the chat client emits reasoning content without a stable `id`. The reasoning emitter keys each block on `content.id or generate_event_id()`, so `id=None` deltas each get a fresh id, closing and reopening the block on every chunk.

A CopilotKit (or any AG-UI) frontend then shows **one "Thought" accordion per token batch** instead of a single expandable reasoning block for the turn.

This is an **internal inconsistency in ag-ui**: its *text* emitter already tolerates `id=None` (it keys on a turn-level `flow.message_id` and ignores `content.id`), so streamed answer text correctly coalesces into one message — but the *reasoning* emitter does not, even though it already tracks an equivalent turn-level `flow.reasoning_message_id`.

Reproduced with `agent-framework-ollama` (whose streamed `text_reasoning` chunks carry `id=None`), but it affects any provider that streams reasoning without per-chunk ids.

### Code Sample

```markdown
1. Wrap a harness agent backed by a streaming "thinking" Ollama model in `AgentFrameworkAgent` and expose it via `add_agent_framework_fastapi_endpoint`.
2. Stream a run that makes the model think for more than one chunk (any non-trivial prompt).
3. Observe the AG-UI event stream (or a CopilotKit client).
```

```markdown

**Observed events** — the `Start → Content → End` triplet repeats once per reasoning delta, each with a different `message_id`:


REASONING_START            message_id=r1
REASONING_MESSAGE_START     message_id=r1 role=reasoning
REASONING_MESSAGE_CONTENT   message_id=r1 delta="The user wants"
REASONING_MESSAGE_END       message_id=r1
REASONING_END              message_id=r1
REASONING_START            message_id=r2          # next delta -> brand-new block
REASONING_MESSAGE_CONTENT   message_id=r2 delta=" to deploy an AI landing zone…"
REASONING_MESSAGE_END       message_id=r2
…


The matching answer **text** in the same turn streams under a single `message_id` — the asymmetry that pinpoints the bug.
```

### Package Versions

agent-framework-core:1.9.0, agent-framework-ag-ui:1.0.0rc5, agent-framework-ollama: 1.0.0b260521, ollama: 0.5.3

### Python Version

Python 3.12

## Expected

All reasoning deltas of a single assistant turn share one `message_id`, producing **one** `REASONING_START → (many) REASONING_MESSAGE_CONTENT → REASONING_END` block — mirroring how `_emit_text` coalesces streamed text. A subsequent turn (e.g. after a tool call) opens its own block.

## Actual

One block per delta → the frontend renders a separate "Thought for a few seconds" accordion per chunk, with the reasoning text chopped mid-sentence across them.

## Root cause

In `python/packages/ag_ui/agent_framework_ag_ui/_run_common.py`, the two emitters treat a missing `content.id` differently.

**Text (tolerant — keys on turn-level id, ignores `content.id`):**

```python
def _emit_text(content, flow, skip_text=False):
    ...
    if not flow.message_id:
        flow.message_id = generate_event_id()      # one id per turn
        ...
    events.append(TextMessageContentEvent(message_id=flow.message_id, delta=content.text))
```

**Reasoning (fragments — keys on per-content id):**

```python
def _emit_text_reasoning(content, flow=None):
    ...
    message_id = content.id or generate_event_id()   # <-- new id every delta when content.id is None
    if flow is not None:
        if flow.reasoning_message_id != message_id:   # always true -> close + reopen each delta
            events.extend(_close_reasoning_block(flow))
            events.append(ReasoningStartEvent(message_id=message_id))
            events.append(ReasoningMessageStartEvent(message_id=message_id, role="reasoning"))
            flow.reasoning_message_id = message_id
        ...
```

When `content.id is None`, `generate_event_id()` returns a fresh value per call, so `flow.reasoning_message_id != message_id` is true on every delta and the block is closed/reopened each time. The streaming machinery (`flow.reasoning_message_id`, `_close_reasoning_block`) is already there to keep one block per turn; the emitter just discards it by re-keying on `content.id`.

## Impact

- Any AG-UI consumer (CopilotKit, etc.) shows fragmented per-chunk reasoning cards instead of one expandable block — a visible UX regression for "thinking" models.
- Triggers for **any** provider whose streamed reasoning content lacks per-chunk ids;

PratikWayase:
### Motivation & Context

1. **Why is this change required?** When a provider (like Ollama) streams model reasoning ("thinking"), it often does not provide a stable `content.id` per chunk, resulting in `content.id=None` for every delta.
2. **What problem does it solve?** The AG-UI reasoning emitter (`_emit_text_reasoning`) was calling `generate_event_id()` on every single delta when `content.id` was `None`. This caused the stream to emit a `REASONING_START -> CONTENT -> REASONING_END` lifecycle per token, instead of one lifecycle per turn.
3. **What scenario does it contribute to?** Any AG-UI frontend (e.g., CopilotKit) connected to a "thinking" model (like DeepSeek-R1 via Ollama) was rendering dozens of fragmented "Thought for a few seconds" accordions, with sentences chopped mid-word across them. 

  Changed the `message_id` assignment logic in `_emit_text_reasoning()` (`_run_common.py`). It now reuses the existing `flow.reasoning_message_id` when `content.id` is `None`, falling back to `generate_event_id()` only on the very first delta of a turn. This brings it into parity with how `_emit_text()` already handles text deltas via `flow.message_id`. Added 3 regression tests.
  Frontends will now correctly receive a single `REASONING_START -> (many) REASONING_MESSAGE_CONTENT -> REASONING_END` block per turn. This is a non-breaking change; providers that *do* supply a `content.id` continue to use it exactly as before.
  Please verify the 5-line logic swap in `_emit_text_reasoning` and the 3 new tests in `TestReasoningCoalescing` to ensure the coalescing behavior correctly mirrors `_emit_text`.

####  Remaining Work (Out of Scope for this PR)
The original issue report contained a **second, distinct bug** regarding parallel tool calls. When Ollama issues the same tool twice in one turn, `agent-framework-ollama`'s `_parse_tool_calls_from_ollama` reuses the function name as the `call_id`, causing a collision. **That bug is located in `agent-framework-ollama` and will be addressed in a separate PR.**

### Related Issue

Partially addresses #6788 

### Contribution Checklist

### DIFF
--- a/python/packages/ag-ui/agent_framework_ag_ui/_run_common.py
+++ b/python/packages/ag-ui/agent_framework_ag_ui/_run_common.py
@@ -944,7 +944,12 @@ def _emit_text_reasoning(content: Content, flow: FlowState | None = None) -> lis
     if not text and content.protected_data is None:
         return []
 
-    message_id = content.id or generate_event_id()
+    if content.id:
+        message_id = content.id
+    elif flow is not None and flow.reasoning_message_id:
+        message_id = flow.reasoning_message_id
+    else:
+        message_id = generate_event_id()
 
     events: list[BaseEvent] = []
 
--- a/python/packages/ag-ui/tests/ag_ui/test_run_common.py
+++ b/python/packages/ag-ui/tests/ag_ui/test_run_common.py
@@ -6,6 +6,11 @@ import logging
 
 import pytest
 from ag_ui.core import EventType
+from ag_ui.core.events import (
+    ReasoningMessageContentEvent,
+    ReasoningMessageStartEvent,
+    ReasoningStartEvent,
+)
 from agent_framework import Content
 
 from agent_framework_ag_ui import state_update
@@ -13,7 +18,9 @@ from agent_framework_ag_ui._orchestration._predictive_state import PredictiveSta
 from agent_framework_ag_ui._run_common import (
     FlowState,
     _build_run_finished_event,
+    _close_reasoning_block,
     _emit_mcp_tool_result,
+    _emit_text_reasoning,
     _emit_tool_result,
     _extract_resume_payload,
     _extract_tool_result_state,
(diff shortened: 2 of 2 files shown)


antsok:
Fixes #6787

moonbox3 (approved):

moonbox3:
There's one file conflict to fix, please.

TaoChenOSU (approved):

PR #6809, opened 2026-06-29
Python: fix: preserve function-call name when merging streaming deltas

Osamaali313:
### Problem

`Content._add_function_call_content` (the `+` operator for `function_call` content, used by `ChatResponse.from_updates` / `AgentResponse.from_updates` to merge streaming deltas) builds the merged function name like this:

```python
name=getattr(self, "name", getattr(other, "name", None)),
```

`Content.__init__` always assigns `self.name` (defaulting to `None`), so the attribute is **never missing**. `getattr(self, "name", <fallback>)` therefore always returns `self.name` — even when it is `None` — and the `<fallback>` (`other.name`) is never consulted.

As a result, when two `function_call` contents are merged and only the **second** carries the name, the name is silently dropped. This happens for any streaming path where the function name arrives in a later delta than the first chunk:

```python
a = Content.from_function_call(call_id="1", name=None, arguments='{"a":')
b = Content.from_function_call(call_id="1", name="get_weather", arguments="1}")

(a + b).name   # -> None          ❌  (name dropped)
(b + a).name   # -> "get_weather" ✅  (works only when self has the name)
```

The asymmetry confirms it's an oversight: the sibling fields in the same `return` use the correct "either side" pattern — `exception=getattr(self, "exception", None) or getattr(other, "exception", None)` on the very next line, and `_add_text_reasoning_content` uses `self.id or other.id`.

### Fix

Mirror the adjacent `exception` field:

```python
name=getattr(self, "name", None) or getattr(other, "name", None),
```

After the fix, the name is preserved regardless of which side carries it, both-`None` stays `None`, and the normal both-set case is unchanged (self wins).

### Tests

Extended `test_function_call_content_add_merging_and_errors` to cover the late-name (`a + b` and `b + a`) and both-`None` cases. Verified RED→GREEN: the new assertions fail on `main` and pass with the fix; the full `tests/core/test_types.py` suite stays green (267 passed).


python/packages/core/tests/core/test_types.py
```diff
@@ -536,6 +536,18 @@ def test_function_call_content_add_merging_and_errors():
     with raises(ContentError):
         _ = a + b
 
+    # name merging: when the first chunk has no name (e.g. a streaming delta where
+    # the function name arrives later), the merged content must keep the name from
+    # whichever side provides it, regardless of order.
+    a = Content.from_function_call(call_id="1", name=None, arguments='{"a":')
```
moonbox3: Could this keep the test typing gate red? `Content.from_function_call` still annotates `name` as `str`, and `uv run poe test-typing -P core --checker pyright` reports this new `name=None` call, plus the two below it, as `reportArgumentType`. If nameless streaming chunks are valid here, should the API signature be widened to `str | None`; otherwise these tests need to construct the nameless `Content` without violating the public factory type?

Osamaali313:
Good catch — fixed in d82c14f. I went with the second option you offered (construct the nameless delta via `Content("function_call", ...)` rather than the factory), so the public `from_function_call(name: str)` signature stays strict and the pyright `reportArgumentType` gate clears. The factory just wraps that same constructor, so the merged objects and assertions are unchanged (test still passes).

I chose this over widening `from_function_call` to `name: str | None` because a nameless call is really a *streaming-delta intermediate* rather than something a caller should mint through the public factory — and production already treats `name` as optionally-absent on that path (e.g. the `content.name` guards in `_compaction.py` / `_evaluation.py`). But if you'd prefer to make nameless deltas first-class and widen the factory signature instead, I'm happy to switch — just say the word.

moonbox3 (approved):

TaoChenOSU (approved):

PR #6811, opened 2026-06-29
[BREAKING] Python: Foundry Hosted Agent V2 protocol upgrade

TaoChenOSU:
### Motivation & Context

Upgrade the `agent-framework-foundry-hosting` package to the latest from `azure-ai-agentserver-xxx`s, which contain a breaking change that updates the hosting protocol to 2.0.0 from 1.0.0, which is a breaking change.

Related materials:
- https://github.com/Azure/azure-sdk-for-python/pull/47584

1. Upgrade the `azure-ai-agentserver-xxx` dependencies
2. Scope checkpoints and approval store by user id
3. Add a new Foundry toolbox integration that relays the call id to the toolbox service as it's now required
4. Add a new Foundry toolbox skills integration
5. Add tests
6. Update all samples to 2.0.0

  - Dependency updates
  - Scope file based storage by user id
  - New Foundry toolbox integration
  - New Foundry toolbox skills integration
  - Breaking as the 2.0.0 is not backward compatible and the latest `agent-framework-foundry-hosting` that supports 1.0.0 will be `1.0.0a260625`

### Contribution Checklist


python/packages/foundry_hosting/agent_framework_foundry_hosting/_toolbox.py
```diff
@@ -0,0 +1,263 @@
...
+        if "http_client" in kwargs:
+            raise TypeError("FoundryToolbox manages its own http_client; pass 'credential' instead.")
+
+        endpoint = url or _resolve_toolbox_endpoint()
+        tool_name = name or os.environ.get("TOOLBOX_NAME") or _toolbox_name_from_endpoint(endpoint)
+
+        http_client = httpx.AsyncClient(
+            auth=_ToolboxAuth(credential, token_scope),
+            timeout=timeout,
+        )
```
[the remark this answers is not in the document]
TaoChenOSU: The header is no longer needed. Samples will be updated in a following commit.

python/samples/04-hosting/foundry-hosted-agents/responses/01_basic/agent.yaml
```diff
@@ -3,10 +3,10 @@ kind: hosted
 name: agent-framework-agent-basic-responses
 protocols:
   - protocol: responses
-    version: 1.0.0
+    version: 2.0.0
```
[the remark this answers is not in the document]
TaoChenOSU: No issue was created to track this as this is an urgent requirement

python/packages/foundry_hosting/agent_framework_foundry_hosting/_toolbox.py
```diff
@@ -0,0 +1,263 @@
...
+                :class:`~agent_framework.MCPStreamableHTTPTool`. Do not pass
+                ``http_client``; this class manages its own.
+        """
+        if "http_client" in kwargs:
+            raise TypeError("FoundryToolbox manages its own http_client; pass 'credential' instead.")
+
+        endpoint = url or _resolve_toolbox_endpoint()
+        tool_name = name or os.environ.get("TOOLBOX_NAME") or _toolbox_name_from_endpoint(endpoint)
+
+        http_client = httpx.AsyncClient(
```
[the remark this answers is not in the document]
TaoChenOSU: This is no longer needed. Will remove in other places too

python/packages/foundry_hosting/agent_framework_foundry_hosting/_toolbox.py
```diff
@@ -0,0 +1,263 @@
...
+                ``TOOLBOX_ENDPOINT`` or from ``FOUNDRY_PROJECT_ENDPOINT`` plus
+                ``TOOLBOX_NAME``.
+            name: The local tool name. When ``None``, it is taken from ``TOOLBOX_NAME``
+                or derived from the endpoint path.
+            token_scope: The token scope to request. Defaults to the Foundry data-plane
+                scope.
+            load_prompts: Whether to load prompts from the toolbox. Defaults to ``False``
+                because toolboxes expose tools.
+            timeout: Request timeout in seconds for the underlying HTTP client.
+            kwargs: Additional keyword arguments forwarded to
```
eavanvalkenburg: if we don't want to expose http_client through kwargs, then we shouldn't expose kwargs, instead expose the things we want to expose. Because things like `allowed_tools` might also make sense here? (kwargs on MCPStreamableHTTPTool are also not passed through)

python/packages/foundry_hosting/agent_framework_foundry_hosting/_toolbox.py
```diff
@@ -0,0 +1,263 @@
...
+        """Close the MCP session and the toolbox-owned HTTP client."""
+        try:
+            await super().close()
+        finally:
+            client = self._httpx_client
+            if self._owns_http_client and client is not None:
+                self._owns_http_client = False
+                await client.aclose()
+
+    def as_skills_provider(
```
eavanvalkenburg, line 186: this makes me think we should have this on the base MCPTool @semenshi what do you think?

python/packages/foundry_hosting/agent_framework_foundry_hosting/_toolbox.py
eavanvalkenburg, line 1: shouldn't this be in the foundry package, instead of this foundry_hosting? or does this only work on FHA?

python/samples/04-hosting/foundry-hosted-agents/responses/01_basic/agent.yaml
rogerbarreto, line 6: This 2.0.0 change should be reflected in every single response sample.

python/packages/foundry_hosting/agent_framework_foundry_hosting/_toolbox.py
TaoChenOSU: This is still debatable. We currently don't know how toolbox will behave outside of the context of hosted agent due to auth. Auth is done via per request id to the foundry hosted agent service.

python/samples/04-hosting/foundry-hosted-agents/responses/01_basic/agent.yaml
TaoChenOSU: Yes, I will have a follow up PR to update all the samples.

python/packages/foundry_hosting/agent_framework_foundry_hosting/_toolbox.py
TaoChenOSU: Good call, but we don't know at the moment how toolbox works outside of FHA. I am still waiting for more clarity.

eavanvalkenburg: yeah, this is more generic, because regular MCPTool will work and they can expose Skills as well.

python/packages/foundry_hosting/agent_framework_foundry_hosting/_toolbox.py
eavanvalkenburg: So, I am worried about breaking changes if we do need to move it, at least, people would need to install a different package if they want to use it, which might cause issues...

python/packages/foundry_hosting/agent_framework_foundry_hosting/_toolbox.py
```diff
@@ -0,0 +1,244 @@
...
+
+    def __init__(
+        self,
+        credential: TokenCredential,
+        *,
+        url: str | None = None,
+        name: str | None = None,
+        token_scope: str = DEFAULT_TOOLBOX_SCOPE,
+        load_prompts: bool = False,
+        timeout: float = _DEFAULT_TIMEOUT,
```
eavanvalkenburg, line 138: now we don't have ways to control broader behavior, such as not loading tools when you only want to use skills, I think we need to map a couple of more of the underlying params here, or is the strategy, to add if we get feedback?

python/packages/foundry_hosting/agent_framework_foundry_hosting/_toolbox.py
TaoChenOSU: This module already depends on the agentserver package. It's also tricky if we place it in the foundry package.

python/packages/foundry_hosting/agent_framework_foundry_hosting/_toolbox.py
TaoChenOSU: For `allowed_tools` and other arguments, we can add later. AFAIK, toolbox has other features such as tool filtering that happens on the server side.

python/packages/foundry_hosting/agent_framework_foundry_hosting/_toolbox.py
TaoChenOSU: Good catch, added back `load_tools`. For other params, we can always add later if toolbox supports them.

eavanvalkenburg (approved):

python/packages/foundry_hosting/agent_framework_foundry_hosting/_toolbox.py
eavanvalkenburg: @copilot can you create a issue to track this?

alliscode (approved):

PR #6817, opened 2026-06-30
Python: Allow custom argument parsing for skill scripts

### LINKED ISSUE #6543: Python: Allow custom argument marshaling for skill scripts
See the .NET PR that adds custom argument marshaling for skill scripts: https://github.com/microsoft/agent-framework/pull/6498. Do the same for Python, if applicable.

giles17:
### Motivation & Context

vLLM and some other OpenAI-compatible backends send tool-call arguments in a non-conforming shape (for example, as a JSON string `"{\"query\":\"hello\"}"` instead of a JSON object). Inline skill scripts bind arguments by keyword name and expect a JSON object (a `dict`), so arguments arriving in another shape cannot be handled and the call fails.

This mirrors the scenario fixed on the .NET side, letting callers plug in their own argument conversion logic so skill scripts work against these backends.

  - Adds a public `SkillScriptArgumentParser` type alias: an optional callable `(raw args: dict | list[str] | str | None) -> dict | None` that converts the raw `args` value into named keyword arguments before an inline script runs.
  - The parser can be supplied at three levels, mirroring the .NET PR:
    - `InlineSkillScript` constructor — per script.
    - `InlineSkill` constructor — default for scripts added via `@skill.script`.
    - `ClassSkill` constructor — default for scripts discovered via `@ClassSkill.script`.
  - In `InlineSkillScript.run()`, the parser is applied first, before the existing inline list-args guard. A raw `str` reaching an inline script with no parser configured raises a clear `TypeError`.
  - Exports the alias from `agent_framework` and documents it in the core package `AGENTS.md`.

  - Opt-in and fully backward compatible: every new parameter defaults to `None`, leaving existing behavior unchanged (arguments are used as-is, expecting a `dict` or `None`).
  - The parser output type is constrained to `dict | None`, so the type system enforces the inline-script contract (inline scripts bind by keyword name).
  - File-based scripts are unaffected — their `SkillScriptRunner` already owns argument handling, matching the .NET scoping to inline scripts.

  - Whether the three-level placement of the hook matches the .NET design, and the chosen parser input/output typing.

### Related Issue

Closes #6543

### Contribution Checklist


python/packages/core/agent_framework/_skills.py
```diff
@@ -74,6 +74,24 @@
 # region Models
 
 
+SkillScriptArgumentMarshaler: TypeAlias = Callable[[dict[str, Any] | list[str] | str | None], dict[str, Any] | None]
```
eavanvalkenburg: I'm not a fan of the term marshaler, in python that has a very specific meaning that doesn't fit here: https://en.wikipedia.org/wiki/Marshalling_(computer_science)#Marshalling_in_different_programming_languages so something like `SkillScriptArgumentParser` or `SkillScriptArgumentSerializer` makes more sense for Python

SergeyMenshykh (approved):

python/packages/core/agent_framework/_skills.py
giles17: Noted. Thanks for the call out! I've addressed it in the latest commit going with `SkillScriptArgumentParser`

python/packages/core/agent_framework/_skills.py
```diff
@@ -336,17 +354,28 @@ def __init__(
...
+        argument_parser: SkillScriptArgumentParser | None = None,
     ) -> None:
         """Initialize an InlineSkillScript.
 
         Args:
             name: Identifier for this script (e.g. ``"analyze"``).
             description: Optional human-readable summary.
             function: Callable (sync or async) that implements the script.
+
+        Keyword Args:
```
TaoChenOSU: nit: Why is `argument_parser` in its own `Keyword Args` section?

TaoChenOSU (approved):

TaoChenOSU (approved):

PR #6822, opened 2026-06-30
Python: Fix: Ollama parallel tool calls collide on same call_id

### LINKED ISSUE #6788: Python: [Bug]: OllamaChatClient: tool calls reuse the function name as `call_id`, so repeated calls in one turn collide

PratikWayase:
### Motivation & Context

1. **Why is this change required?** Ollama's API does not return a unique per-call ID for tool invocations. 
2. **What problem does it solve?** `_parse_tool_calls_from_ollama` was using the function name as the `call_id`. When a model issued the same tool twice in one turn (e.g., parallel searches), both calls shared one `call_id`. This caused AG-UI to merge their arguments into invalid JSON, silently drop the second call, and leave HITL approval cards permanently stuck on "Running".
3. **What scenario does it contribute to?** Any prompt that triggers parallel/repeated tool calls via Ollama (e.g., "Search for A and B at the same time") now correctly executes both tools and renders two independently-resolving approval cards in CopilotKit/AG-UI.
4. **If it fixes an open issue, please link to the issue below.** Fixes #6788

  Changed `_parse_tool_calls_from_ollama` to generate a unique, deterministic `call_id` using `f"{name}:{index}:{blake2s(args_json).hexdigest()[:8]}"`. Changed `_format_tool_message` to strip this suffix (`split(":", 1)[0]`) so Ollama's API still receives the bare function name required to correlate the tool result in its chat history. Added 2 regression tests.

  AG-UI frontends now correctly receive `N` independent tool call lifecycles for `N` parallel calls, preventing argument corruption and fixing stuck HITL approval states. This is fully backward compatible—single tool calls simply get a `:0:hash` suffix that is transparently stripped on the return trip.

  Please verify the `blake2s` hash generation ensures uniqueness across identical function names, and confirm that `split(":", 1)[0]` safely restores the bare tool name even if a tool's original name contains a colon.

### Related Issue

Fixes #6788

### Related PRs

The original issue #6788 tracked two distinct bugs across two separate packages. To keep changes atomic and easy to review, they were split into two PRs:
- **Part 1 (Reasoning fragmentation in `ag-ui`):** #6804 - *Python: Fix: coalesce reasoning deltas into single block when content.id is None*
- **Part 2 (Tool call collision in `ollama`):** *This PR*

### Contribution Checklist


python/packages/ollama/agent_framework_ollama/_chat_client.py
```diff
@@ -524,7 +525,14 @@ def _format_tool_message(self, message: Message) -> list[OllamaMessage]:
                     tool_text = "\n".join(text_parts) if text_parts else ""
                 else:
                     tool_text = str(item.result) if item.result is not None else ""
-                messages.append(OllamaMessage(role="tool", content=tool_text, tool_name=item.call_id))
+
+                # CHANGED: Strip the unique suffix (index:hash) to restore the bare
```
eavanvalkenburg: we don't need to know what changed, only comments that explain the code make sense

python/packages/ollama/agent_framework_ollama/_chat_client.py
```diff
@@ -524,7 +525,14 @@ def _format_tool_message(self, message: Message) -> list[OllamaMessage]:
...
+                # function name that Ollama's API expects to correlate the tool result.
+                tool_name = item.call_id.split(":", 1)[0] if item.call_id else ""
+
+                messages.append(
+                    OllamaMessage(role="tool", content=tool_text, tool_name=tool_name)
+                )  # <-- CHANGED: Was item.call_id
```
eavanvalkenburg: same here

python/packages/ollama/agent_framework_ollama/_chat_client.py
```diff
@@ -568,11 +576,19 @@ def _parse_response_from_ollama(
 
     def _parse_tool_calls_from_ollama(self, tool_calls: Sequence[OllamaMessage.ToolCall]) -> list[Content]:
         resp: list[Content] = []
-        for tool in tool_calls:
+        for index, tool in enumerate(tool_calls):
+            name = tool.function.name
+            args = tool.function.arguments if isinstance(tool.function.arguments, dict) else {}
+
+            args_json = json.dumps(args, sort_keys=True)
+            args_hash = hashlib.blake2s(args_json.encode()).hexdigest()[:8]
```
eavanvalkenburg: do we need all this? could we just do a uuid4? because this will also provide the same id if the args are the same, but that should not necessarily be the same call?

eavanvalkenburg:
@PratikWayase  please address and resolve the comments and get the checks to pass!

moonbox3:
@PratikWayase please also resolve all comments if/when they're addressed. It helps us better review the PR. Thanks.

moonbox3 (approved):

TaoChenOSU (approved):

Issue #6803, opened 2026-06-29
Python: Realtime Agent Abstractions


eavanvalkenburg:
### Description

Realtime abstractions for python

### Code Sample

```markdown

```

### Language/SDK

Python

PR #6831, opened 2026-06-30
Python: Add realtime agent ADR

eavanvalkenburg:
### Motivation & Context

Python realtime agent work needs an agreed architecture before implementation starts. This adds a proposed ADR for the Python realtime agent design, covering direct realtime model sessions, STT → agent → TTS pipeline composition, agent compatibility, event/data models, and the provider/session protocol boundary.

  - Adds `docs/decisions/0031-python-realtime-agent.md` as a proposed ADR.
  - Chooses a dedicated `RealtimeAgent` built on `BaseAgent` + agent middleware/telemetry layers rather than extending `Agent` or inheriting chat-specific `RawAgent`.
  - Defines the proposed roles of `AgentConversation`, `RealtimeEvent`, `RealtimePipeline`, STT/TTS abstractions, and the direct realtime client/session protocols.
  - Includes appendices comparing Semantic Kernel Python realtime and .NET `IRealtimeClient`, plus canonical code shapes.
  - Documentation-only architecture proposal. No runtime behavior changes.
  - Establishes the intended shape for future Python realtime implementation PRs.
  - Whether the `RealtimeAgent` / `AgentConversation` split is the right Python surface.
  - Whether the `RealtimeEvent` and provider protocol boundaries are narrow enough.
  - Whether `RealtimePipeline(STT, agent, TTS)` is staged appropriately.

### Related Issue

Related to #6803. This ADR does not close the implementation issue.

### Contribution Checklist


docs/decisions/0035-python-realtime-agent.md
```diff
@@ -0,0 +1,1141 @@
...
+
+- **Option 1: Add `RealtimeAgent` as the realtime-specific agent type**
+- **Option 2: Extend `Agent` to also be the realtime agent type**
+
+## Pros and Cons of the Options
+
+### Option 1: Add `RealtimeAgent` as the realtime-specific agent type
+
+Add `RawRealtimeAgent` / `RealtimeAgent` as first-class agent types for realtime. They should satisfy
+`SupportsAgentRun` for half-duplex compatibility and reuse the same base/middleware/telemetry stack where practical, but
```
TaoChenOSU, line 101: Could we add some description on the input and output shape of the clients? If the client accepts and outputs audio without texts, how will our current middleware/telemetry stack support audio?

docs/decisions/0035-python-realtime-agent.md
```diff
@@ -0,0 +1,1141 @@
...
+    await conversation.send("Hi there, I'm based in Amsterdam.")
+    async for update in conversation:
+        ...
+```
+
+In phase 2, sandwich setups should use a concrete `RealtimePipeline` class that composes speech-to-text, an inner
+`SupportsAgentRun`, and text-to-speech. The constructor order should follow the audio-to-audio flow:
+
+```python
+pipeline = RealtimePipeline(
```
TaoChenOSU, line 160: I wouldn't consider this as realtime. Based on the definition from OpenAI: https://developers.openai.com/api/docs/guides/realtime, what we have here is more like a request based audio API. 

Also, my guess is that this will not support streaming. Will the interaction still look like the live API:
```
async with agent.start_conversation() as conversation:
    await conversation.send("Hi there, I'm based in Amsterdam.")
    async for update in conversation:
```?

File: docs/decisions/0035-python-realtime-agent.md [hunk 3]
```diff
@@ -0,0 +1,1141 @@
...
+- `RealtimeAgent.run(..., stream=True)` owns the bounded response/`ResponseStream` projection for half-duplex scenarios.
+
+Entering and exiting the async context should be deterministic:
+
+- `__aenter__` creates or enters the underlying `RealtimeClientSessionProtocol`, applies initial session options, replays
+  any initial messages/history that belong to the conversation, installs tools/session settings, and starts the single
+  receive loop that maps `RealtimeEvent` values to `AgentResponseUpdate` values.
+- `__aenter__` should not implicitly request a model response unless the caller passed an explicit option such as
+  `create_response=True`.
+- `__aexit__` stops accepting new sends, requests provider/session close, cancels the receive loop if it is still
```
moonbox3, line 345: Should close finalize only the authored user, assistant, and tool turns through the normal `SessionContext` / context-provider pipeline? Today `HistoryProvider` stores outputs only from `context.response.messages` during `after_run`, but this full-duplex path defines neither a final response/provider callback nor a realtime-specific filter and writes only continuation state.

docs/decisions/0035-python-realtime-agent.md
```diff
@@ -0,0 +1,1141 @@
...
+The framework should define small provider-neutral audio bridge shapes, for example an audio frame type and source/sink
+protocols, so callers can connect microphone/speaker helpers or hosted media streams to `RealtimeConversation`. Concrete
+device drivers and ACS-specific web apps should remain samples or optional integrations, but users should not have to
+drop below the agent/conversation API to build the common loops:
+
+- local microphone -> `conversation.send(Content.from_data(...))`;
+- conversation audio content -> local speaker;
+- ACS websocket audio data -> `conversation.send(Content.from_data(...))`;
+- conversation audio content -> ACS websocket audio data;
+- VAD/barge-in events -> local or hosted "stop current playback" signal.
```
moonbox3, line 366: Should the audio sink report playback position back to the conversation? Stopping local playback or cancelling the response does not remove already-generated but unheard audio and transcript from OpenAI conversation context; `conversation.item.truncate` needs the `item_id`, `content_index`, and `audio_end_ms`. Without that reverse path, subsequent turns can reason over words the user never heard.

docs/decisions/0035-python-realtime-agent.md
```diff
@@ -0,0 +1,1141 @@
...
+Add provider-neutral direct realtime client/session protocols in core. The shape should be small:
+
+- a realtime client creates a live realtime session;
+- a realtime session sends `RealtimeEvent` commands;
+- a realtime session yields provider events mapped to `AgentResponseUpdate`;
+- raw provider events remain reachable through `raw_representation`.
+
+Add `RawRealtimeAgent` / `RealtimeAgent` over this client/session shape:
+
+- `run(..., stream=False)` opens an ephemeral realtime session, sends one turn, drains until the terminal response event,
```
moonbox3, line 400: Should bounded `run()` have an explicit terminal-error contract in addition to a terminal response event? OpenAI emits most request errors as recoverable `error` events and keeps the session open, so rejected input or `response.create` can produce no `response.done`. Mapping that event to `Content.from_error` leaves both variants waiting forever unless the triggering event is correlated and the run fails or finalizes, with a timeout fallback.

docs/decisions/0035-python-realtime-agent.md
```diff
@@ -0,0 +1,1141 @@
...
+  and returns `AgentResponse`;
+- `run(..., stream=True)` does the same but yields `AgentResponseUpdate` values through `ResponseStream`;
```
moonbox3, line 402: Could streaming define deterministic cleanup when iteration stops early? The existing `ResponseStream` runs cleanup only at exhaustion or error and exposes no `aclose()` or async-context-manager contract, so breaking after one update leaves the ephemeral socket and receive task live. Should the realtime stream expose an explicit close or cancel path?

docs/decisions/0031-python-realtime-agent.md
```diff
@@ -0,0 +1,1042 @@
...
+- Python uses structural `Protocol` types instead of C# interfaces.
+- Python uses async context managers instead of `IAsyncDisposable`.
+- Python includes `Message`/`Content` send conveniences to reduce boilerplate in samples and hosting bridges.
+- Python includes `RealtimePipeline` as a first-class sandwich composition target, while .NET's ADR focuses on direct
+  realtime clients.
+
+## More Information
+
+- .NET realtime ADR draft: `https://github.com/rogerbarreto/agent-framework-public/blob/features/realtime-agent/docs/decisions/0030-realtime-ai-agent.md`
+- Semantic Kernel realtime client ADR: `https://github.com/microsoft/semantic-kernel/blob/main/docs/decisions/0065-realtime-api-clients.md`
```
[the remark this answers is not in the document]
eavanvalkenburg: Addressed in 21077967d: changed the external references in More Information from inline-code URLs to normal Markdown links so they render as clickable links.

docs/decisions/0035-python-realtime-agent.md
eavanvalkenburg: Addressed in 21077967d: expanded the ADR to describe provider-level input/output shape (`send(RealtimeEvent)`, `receive() -> AsyncIterable[RealtimeEvent]`), `RealtimeConversation.send(...)` normalization using `AgentRunInputs` / `normalize_messages(...)`, audio-only content as `Content.from_data(...)`, and middleware/telemetry expectations for audio-only updates.

docs/decisions/0035-python-realtime-agent.md
eavanvalkenburg: Addressed in 21077967d: clarified that `start_conversation()` is the full-duplex live API with `async for update in conversation`, while `run(...)` is only the half-duplex projection over an ephemeral realtime session. Also made direct realtime phase 1 and sandwich pipelines phase 2.

docs/decisions/0035-python-realtime-agent.md
eavanvalkenburg: Addressed in 21077967d: clarified close/finalization behavior. `RealtimeConversation.close()` writes durable continuation state and should run normal context-provider/history finalization only for authored user/assistant/tool messages, not protocol events, VAD/rate-limit/session events, or raw audio-buffer events.

docs/decisions/0035-python-realtime-agent.md
eavanvalkenburg: Addressed in 21077967d: added the playback-progress/truncation concern to the audio bridge section. The ADR now calls out that providers such as OpenAI need item id, content index, and `audio_end_ms` so unheard generated audio/transcript can be truncated after barge-in.

docs/decisions/0035-python-realtime-agent.md
eavanvalkenburg: Addressed in 21077967d: added a terminal-error contract for bounded `run(...)`. Correlated provider error events should fail or finalize the bounded run rather than waiting forever for `response.done`, with timeout/cancellation fallback for missing terminal events.

docs/decisions/0035-python-realtime-agent.md
eavanvalkenburg: Addressed in 21077967d: added deterministic cleanup requirements for `run(..., stream=True)` when callers stop iteration early, including an explicit close/cancel path or stream wrapper that closes the ephemeral realtime session and receive task.

moonbox3:
@eavanvalkenburg please re-open when you're ready to move forward with this.

docs/decisions/0035-python-realtime-agent.md
```diff
@@ -0,0 +1,1184 @@
...
+`RealtimeConversation` has these responsibilities:
+
+- **Own the single receive loop**. The provider session is single-reader. `RealtimeConversation` is the only component that
+  calls `RealtimeClientSessionProtocol.receive(...)`.
+- **Project events to updates**. It maps each provider `RealtimeEvent` to one `AgentResponseUpdate` and exposes only
+  agent-level updates to callers.
+- **Serialize outbound writes**. It should prevent overlapping sends from corrupting provider state. A simple async lock
+  around `send(...)` and `interrupt(...)` is enough unless
+  a provider proves otherwise.
+- **Own lifecycle and cleanup**. It is an async context manager. Exiting it closes the provider session, cancels the
```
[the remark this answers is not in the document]
eavanvalkenburg: Addressed in 591b7ea7e. The ADR now requires a finite configurable update queue, fail-fast overflow with an observable backpressure error and provider cleanup, no silent event loss, and terminal state outside the bounded queue so close/error paths cannot deadlock when it is full. It also clarifies that normal open-ended iteration ends only on explicit close/context exit, with provider/receive/backpressure failures as exceptional termination.

docs/decisions/0035-python-realtime-agent.md
```diff
@@ -0,0 +1,1184 @@
...
+- Bad, because sandwich pipelines require more cancellation and backpressure plumbing than direct realtime model sessions.
+
+## Validation
+
+Validation should happen in stages:
+
+1. ADR review validates the architecture before implementation.
+2. Unit tests with fake realtime sessions validate direct realtime model behavior, half-duplex finalization, streaming
+   finalization, event preservation, cleanup, receive-loop errors, and single-reader behavior.
+3. Unit tests validate `Content.from_realtime_event(...)` serialization and `AgentResponse.from_updates(...)`
```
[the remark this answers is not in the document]
eavanvalkenburg: Addressed in 591b7ea7e. The ADR no longer proposes generic `AgentResponse.from_updates(...)` aggregation for realtime. It now requires event classification plus realtime-specific bounded-response and conversation-history finalizers that reconstruct only authored messages; control/VAD/lifecycle/rate-limit/raw-buffer events remain stream-visible and available through raw representations but cannot enter response messages or durable history.

docs/decisions/0035-python-realtime-agent.md
```diff
@@ -0,0 +1,1184 @@
...
+  `RealtimeEvent.content` at the provider/session boundary.
+
+Add `RawRealtimeAgent` / `RealtimeAgent` over this client/session shape:
+
+- `run(..., stream=False)` opens an ephemeral realtime session, sends one turn, drains until the terminal response event,
+  and returns `AgentResponse`;
+- `run(..., stream=True)` does the same but yields `AgentResponseUpdate` values through `ResponseStream`;
+- `start_conversation(...)` returns `RealtimeConversation` for full-duplex live use.
+
+`RealtimeAgent` must satisfy `SupportsAgentRun` so it can participate in half-duplex agent scenarios, including
```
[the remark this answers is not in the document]
eavanvalkenburg: Addressed in 591b7ea7e. I narrowed the compatibility claim to half-duplex workflows, hosting surfaces, and utilities whose contracts accept `SupportsAgentRun`. The ADR now explicitly excludes the current `HandoffBuilder`, documents why its concrete `Agent` requirement is necessary today, and defers realtime handoffs until that orchestration is independently generalized.

docs/decisions/0035-python-realtime-agent.md
```diff
@@ -0,0 +1,1184 @@
...
+
+```python
+from agent_framework import Content
+from agent_framework.realtime import RealtimeConversation
+
+
+async def from_realtime_to_acs(websocket: WebSocket, content: Content) -> None:
+    await websocket.send(json.dumps({
+        "kind": "AudioData",
+        "audioData": {"data": content.data},
```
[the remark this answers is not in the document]
eavanvalkenburg: Addressed in 591b7ea7e. The ACS example now reads the public `Content.uri` data-URI representation, validates that it is base64 audio data (including decoding validation), and forwards only the encoded payload expected by ACS. It no longer references the nonexistent `Content.data` attribute or private helpers.

docs/decisions/0035-python-realtime-agent.md
```diff
@@ -0,0 +1,1223 @@
...
+
+Add `RawRealtimeAgent` / `RealtimeAgent` over this client/session shape:
+
+- `run(..., stream=False)` opens an ephemeral realtime session, sends one turn, drains until the terminal response event,
+  and returns `AgentResponse`;
+- `run(..., stream=True)` does the same but yields `AgentResponseUpdate` values through `ResponseStream`;
+- `start_conversation(...)` returns `RealtimeConversation` for full-duplex live use.
+
+`RealtimeAgent` must satisfy `SupportsAgentRun` so it can participate in half-duplex agent scenarios whose contracts accept
+the structural protocol, including protocol-based workflows, hosting surfaces, and agent-like utilities. This does not make
```
moonbox3, line 406: Following up on the earlier `SupportsAgentRun` compatibility discussion: should we either require `RealtimeAgent.run()` to accept the hosting argument superset or narrow "hosting surfaces" to the ones verified against the protocol? `AgentMCPTool` is typed against `SupportsAgentRun` but always calls `run(..., options=...)` at `_agent_tool.py:161`, and AG-UI can pass both `tools=` and `options=` at `_agent_run.py:2506` and `_agent_run.py:2521`; neither keyword is in the structural protocol. An implementation that satisfies the stated protocol can therefore pass validation and fail with `TypeError` as soon as one of those hosts invokes it.

docs/decisions/0035-python-realtime-agent.md
```diff
@@ -0,0 +1,1223 @@
...
+
+Comparison with Semantic Kernel Python realtime and canonical code shapes are included in the appendices so the main
+decision stays readable while still making the proposed API concrete.
+
+The selected architecture has these layers.
+
+### Agent-level live conversation
+
+Introduce an experimental `RealtimeConversation` as the Agent Framework live full-duplex surface. It is an async context
+manager / async disposable object, not an `AgentSession`.
```
moonbox3, line 181: Should the experimental boundary cover the whole realtime API family rather than only `RealtimeConversation`? This decision also introduces `RealtimeAgent`, `RawRealtimeAgent`, `RealtimeEvent`, the client/session protocols, and `Content.from_realtime_event()` into released packages, where an undecorated public API is treated as released. Defining one experimental feature ID for the family here would keep those coupled contracts free to evolve together instead of accidentally stabilizing everything except the conversation type.

docs/decisions/0035-python-realtime-agent.md
```diff
@@ -0,0 +1,1223 @@
...
+
+The existing `AgentMiddlewareLayer` and `AgentTelemetryLayer` should be reused only where their current `run(...)`
+contract fits realtime. Current `AgentMiddlewareLayer` is shaped around the chat `Agent.run(...)` signature, including
+chat options, tool forwarding, compaction, tokenizer, and chat/function middleware forwarding. If `RawRealtimeAgent.run`
+does not intentionally support that same compatible shape, the realtime implementation should factor out the reusable
+middleware/telemetry pieces or add small realtime-specific adapter layers rather than inheriting chat assumptions.
+
+Middleware and telemetry support for audio-only interactions should be explicit. Middleware that only inspects text must
+not assume every update has text. Telemetry should record provider/model/session metadata and timing, but raw audio
+payloads should not be captured by default. Token/usage metrics are provider-dependent for realtime audio and should be
```
moonbox3, line 440: How should full-duplex `start_conversation()` traffic enter the middleware and telemetry stack? The existing layers only wrap `run()`, while the ADR makes `send()` and the receive iterator the primary audio path and defers dedicated realtime middleware. Naming the interception points for inbound sends, outbound updates, and conversation close would prevent hosts from having one policy and observability path for bounded calls and a separate uninstrumented path for live sessions.

eavanvalkenburg:
Closing until we pick it up

PR #6837, opened 2026-06-30
Python: Refocus hosting channels ADR on protocol helpers

eavanvalkenburg:
### Motivation & Context

The Python hosting channels ADR was accepted before the API was released. Further design exploration showed that the most useful framework-owned surface is narrower than a channel route-contribution framework: protocol-specific conversion helpers plus optional shared execution state.

This update reduces the proposed framework responsibility so web frameworks and native protocol SDKs continue to own routing, auth, middleware, background work, command handling, and client calls.

  - Rewrites ADR-0027 around protocol-specific helpers such as `responses_to_af(...)`, `response_from_af(...)`, `telegram_to_af(...)`, and `telegram_from_af(...)`.
  - Keeps `AgentFrameworkHost` only as an optional shared execution-state holder for app, target, session store, and possible workflow checkpointing.
  - Moves channel contribution, host-owned hooks, command projection, and protocol dispatch out of the proposed v1 public surface.
  - Narrows the accepted-but-unreleased Python hosting design before it becomes public API.
  - Makes the proposed protocol helper surface usable from FastAPI, Starlette, Azure Functions, Django, tests, or native SDK handlers.
  - Whether the ADR draws the right boundary between Agent Framework-owned protocol conversion/state and app/web-framework-owned routing/client code.
  - Whether keeping a small optional `AgentFrameworkHost` for stateful execution and future workflow checkpointing is justified.

### Related Issue

N/A — design ADR update for accepted-but-unreleased API. Checked for similar open PRs; #6151 is a .NET implementation PR and does not cover this Python ADR refocus.

### Contribution Checklist

### DIFF
--- a/docs/decisions/0027-hosting-channels.md
+++ b/docs/decisions/0027-hosting-channels.md
@@ -1,144 +1,349 @@
 ---
 status: accepted
 contact: eavanvalkenburg
-date: 2026-06-11
+date: 2026-06-30
 deciders: eavanvalkenburg
 ---
 
-# Python minimal hosting core and pluggable channels
+# Python protocol helpers and optional execution state
 
 ## Context and Problem Statement
 
-Agent Framework has several protocol-specific hosting surfaces. App authors who want one agent or workflow on multiple protocols must compose servers, routes, middleware, session handling, and lifecycle code by hand.
+Agent Framework needs to help applications expose agents and workflows over external protocols such as OpenAI
+Responses, Telegram, Activity Protocol, and future transports.
 
-We will introduce a small Python hosting core that owns the common server shape and leaves protocol details inside channel packages. The first public contract must be intentionally narrow so Python can ship a base contract before adding identity linking, proactive delivery, or multicast behavior. Other language implementations may reuse the same conceptual boundary, but this ADR records the Python decision.
+The first version of this ADR chose a host/channel model: channel packages contributed routes, middleware, commands,
+lifecycle callbacks, hooks, and protocol dispatch to a common host object. That design was accepted before it was
+released. Implementation experiments showed that the most valuable part is narrower: translating protocol-native
+payloads into Agent Framework inputs and translating Agent Framework results back to protocol-native payloads.
+
+FastAPI, Starlette, Azure Functions, Django, Telegram SDKs, Bot Framework SDKs, and other app frameworks already own
+route registration, dependency injection, middleware, authentication, background tasks, lifecycle, and native client
+calls. Agent Framework should not duplicate those surfaces unless a specific hosting environment requires it.
 
 ## Decision Drivers
 
-- Keep the first host easy to explain: one app, one hostable target, one or more channels.
-- Reuse Agent Framework's existing agent, workflow, session, history, and checkpoint primitives.
-- Let channel packages own protocol parsing, protocol responses, authentication details, and native command surfaces.
-- Make session continuity explicit through a channel-supplied `ChannelSession(isolation_key=...)`.
+- Keep the released surface small enough to explain without first teaching a channel framework.
+- Provide reusable Agent Framework run translation that works with FastAPI and other web frameworks.
+- Let app/framework code own route declaration, auth, middleware, native SDK clients, command handling, and background
+  work.
+- Keep stateful execution support explicit: session lookup, session reset, and workflow checkpointing may still need a
+  small AF-owned home.
 - Avoid approving cross-channel identity and delivery semantics before their safety model is reviewed.
 
 ## Considered Options
 
-1. Keep only protocol-specific hosts.
-2. Ship a large hosting core with identity linking, authorization, background delivery, active-channel routing, and multicast in v1.
-3. Ship a minimal host/channel core now and track linking/multicast as follow-up work.
+1. Create protocol-specific hosts.
+2. Ship a full host/channel framework with route contribution and channel hooks.
+3. Ship protocol conversion helpers plus optional execution state.
 
-### Keep only protocol-specific hosts
+### Create protocol-specific hosts
 
-- Good: no new abstraction or package surface.
-- Neutral: each protocol can continue evolving independently.
-- Bad: every multi-channel app still has to compose servers, lifecycle, and session handling by hand.
+- Good: no new shared abstraction.
+- Neutral: each protocol host can evolve independently.
+- Bad: every package reinvents AF input/result mapping, session-key conventions, and stateful execution helpers.
 
-### Ship the large cross-channel host in v1
+### Ship a full host/channel framework
 
-- Good: the richest cross-channel scenarios are available immediately.
-- Neutral: the host becomes the natural place to demonstrate identity and delivery policy.
-- Bad: v1 becomes a security-sensitive identity and delivery system before the safety model is reviewed.
+- Good: one object can assemble routes, channels, session handling, hooks, and lifecycle callbacks.
+- Good: app code using the supported host shape can be short.
+- Bad: the framework owns concerns already handled by web frameworks, protocol SDKs and/or other services.
+- Bad: users must understand `Channel`, contribution, hook, and host-dispatch concepts before they can see how a request
+  becomes `agent.run(...)`.
+- Bad: the abstraction is hard to reuse outside the chosen ASGI shape.
 
-### Ship the minimal core now
+### Ship protocol helpers plus optional execution state
 
-- Good: the host/channel boundary can be implemented, tested, and explained without solving linking and durable delivery at the same time.
-- Neutral: apps that need richer behavior must build it locally or wait for ADR-0028 follow-up work.
-- Bad: proactive delivery and multicast scenarios are deliberately absent from v1.
+- Good: protocol packages provide the Agent Framework run value directly: `<protocol>_to_run(...)` and
+  `<protocol>_from_run(...)` style helpers.
+- Good: apps keep native FastAPI, Starlette, Azure Functions, Django, Bot Framework, or Telegram SDK code.
+- Good: helper functions can be tested without an ASGI app or host pipeline.
+- Good: a small state object can still own target-coupled state such as an agent/workflow target, a `SessionStore`, and
+  future workflow checkpoint coordination.
+- Good: provides maximum configurability in handling input and outputs (outside of the conversions)
+- Bad: building a first iteration of a new Host is more verbose.
+- Bad: samples show more explicit route/client code than a fully assembled channel host.
 
 ## Decision Outcome
 
-Chosen option: **minimal host/channel core now, follow-up enhancements later**.
+Chosen option: **protocol helpers plus optional execution state**.
 
-`AgentFrameworkHost` owns:
+Protocol packages own:
 
-- one application object,
-- one hostable target (`SupportsAgentRun` agent-compatible object or a `Workflow`), and
-- one or more channels.
+- parsing protocol-native input into Agent Framework run input and options;
+- rendering `AgentResponse`, `AgentResponseUpdate`, workflow results, or workflow updates back into protocol-native
+  response/event payloads;
+- protocol-specific isolation/session id helper functions when useful, such as `telegram_chat_session_id(update)`;
+- protocol-specific typing/update event helpers where the protocol has a native concept.
 
-Channels own:
+Application or web-framework code owns:
 
-- contributed routes, middleware, commands, and lifecycle callbacks,
-- protocol-native request parsing into `ChannelRequest`,
-- protocol-native rendering of the originating response, and
-- any channel-specific authentication or signature validation.
+- HTTP route declaration and route grouping;
+- dependency injection;
+- authentication and authorization;
+- middleware;
+- background tasks and webhook acknowledgement policy;
+- native protocol SDK clients and outbound calls;
+- command registration and command dispatch;
+- request/response status codes and framework-specific error handling;
+- choosing the isolation/session id source for the current deployment and route.
 
-The host owns:
+`AgentFrameworkState`, if provided, is limited to shared execution state:
 
-- route/lifecycle aggregation,
-- invocation of the target,
-- `ChannelSession(isolation_key=...)` to `AgentSession` resolution and caching,
-- `reset_session(isolation_key=...)`,
-- host-level middleware, including Foundry isolation middleware only when the Foundry hosting environment flag is present,
-- invocation of per-channel hooks (`ChannelRunHook`, `ChannelResponseHook`, `ChannelStreamUpdateHook`), and
-- workflow checkpoint wiring through an explicit `checkpoint_location`.
+- one first-class hostable target: either a `SupportsAgentRun` agent-compatible object or a `Workflow`;
+- a `SessionStore` instance or factory;
+- optional workflow checkpoint execution state.
 
-`ChannelIdentity`, when present, is request metadata only. In v1 it is not a linking, authorization, or delivery key.
+It is **not** an app object, channel registry, or route owner. It does not own FastAPI/Starlette setup, route
+contribution, protocol dispatch, command projection, or native SDK calls.
 
-### Trust boundary for `isolation_key`
+### Helper naming
 
-The host treats `ChannelSession.isolation_key` as a session partition key, not as proof of identity. Channels or host middleware must authenticate and authorize any externally supplied value before passing it to the host. For example, a Responses caller must not be allowed to choose an arbitrary `previous_response_id` or header-derived key unless the platform or middleware has already established that the caller owns that conversation. The host deliberately does not infer that trust from the string itself.
+Helpers should be protocol-specific, not generic. Prefer:
 
-### Hook ownership
+- `responses_to_run(...)`
+- `responses_from_run(...)`
+- `responses_stream_event_from_run(...)` if streaming needs a separate event helper
+- `telegram_to_run(...)`
+- `telegram_from_run(...)`
 
-Channels provide hook configuration and protocol-native context. The host invokes those hooks as part of the common invocation pipeline:
+Avoid a generic `protocol_to_run(...)` name in public samples because it hides the protocol-specific contract behind a
+second abstraction.
 
-- `ChannelRunHook` runs after channel parsing and before target invocation.
-- `ChannelResponseHook` runs after target invocation and before the originating channel serializes its response.
-- `ChannelStreamUpdateHook` is applied by the host while the channel consumes streamed updates because streaming serialization is protocol-specific.
+### Session continuity
 
-`ChannelStreamUpdateHook` is an update hook, not a final-response sanitizer. Channels that use it for redaction or filtering must also apply equivalent policy to any final response they render. Channels choose whether the response is streaming before run hooks execute.
+Session continuity remains explicit. Run parsing and isolation/session id selection are separate operations because
+isolation can come from more than one source:
 
-This keeps hook call conventions centralized while leaving protocol payload parsing and response formatting in channel packages.
+- protocol input, such as OpenAI Responses `previous_response_id`, a Telegram chat id, or an Activity conversation id;
+- running environment, such as Foundry Hosted Agents user/chat isolation context;
+- app-specific trusted middleware or route state.
 
-### State owned by v1
+The app chooses which helper to call for that route and deployment. For example:
 
-`state_dir` is limited to host-owned local files for reset-session aliases and workflow checkpoint path derivation. It does not store linked identities, active-channel state, response-routing state, continuation records, durable runner queues, or delivery attempts. Those storage concerns belong to ADR-0028.
+- `responses_session_id(body)` from `agent-framework-hosting-responses`, which can return either a `resp_*` previous
+  response id or a `conv_*` conversation id when present;
+- `telegram_chat_session_id(update)` from `agent-framework-hosting-telegram`;
+- `foundry_user_isolation_key()` or `foundry_chat_isolation_key()` from `agent-framework-foundry-hosting`.
+
+Keep these helpers outside `responses_to_run(...)`, `telegram_to_run(...)`, and other run-input parsers. That makes the
+trust boundary visible: using a request-derived key is a different decision than using a platform-provided isolation key.
+
+A `SessionStore` resolves that key into an `AgentSession`:
+
+For agent targets:
+
+```python
+session = await state.session_store.get(session_id)
+result = await state.target.run(messages, session=session, options=options)
+```
+
+For workflow targets, app code adapts the protocol helper output into the workflow's expected input and invokes the
+workflow through the state object's target:
+
+```python
+result = await state.target.run(message=workflow_input)
+```
+
+`SessionStore.reset(session_id)` rotates or clears the current session for non-persisted servers. Persisted stores can
+implement the same async interface later.
+
+The session id is a partition key, not proof of identity. App or platform code must authenticate and authorize any
+externally supplied key before using it.
+
+### Workflow checkpoints
+
+Workflow checkpointing is execution state, not protocol state. A small state object may help coordinate checkpoint
+storage for workflow targets, but protocol helper packages should not own checkpoint layout, route lifecycle, or durable
+execution.
 
 ## Non-goals for v1
 
-The following are deliberately **not** part of the v1 contract:
+The following remain outside the v1 contract:
 
-- cross-channel identity linking (`IdentityLinker`, `local_identity_link`, or `agent-framework-hosting-entra`),
-- identity allowlists or authorization policy (`IdentityAllowlist`, `AuthPolicy`),
-- response routing beyond the originating channel (`ResponseTarget`, active channel, specific linked channel, `all_linked`),
-- push or payload codecs (`ChannelPush`, `ChannelPushCodec`),
-- background/continuation delivery,
-- durable task runners (`DurableTaskRunner`, `InProcessTaskRunner`),
-- retry/replay policy (`RetryPolicy`),
-- fan-out, multicast, or all-linked delivery,
-- confidentiality tiers and `LinkPolicy`, and
+- cross-channel identity linking (`IdentityLinker`, `local_identity_link`, or `agent-framework-hosting-entra`);
+- identity allowlists or authorization policy (`IdentityAllowlist`, `AuthPolicy`);
+- response routing beyond the originating protocol (`ResponseTarget`, active channel, specific linked channel,
+  `all_linked`);
+- push or payload codecs (`ChannelPush`, `ChannelPushCodec`);
+- background/continuation delivery;
+- durable task runners (`DurableTaskRunner`, `InProcessTaskRunner`);
+- retry/replay policy (`RetryPolicy`);
+- fan-out, multicast, or all-linked delivery;
+- confidentiality tiers and `LinkPolicy`;
 - a host-level multi-agent router.
 
-These areas are follow-up enhancements covered by [ADR-0028](0028-hosting-linking-multicast-enhancements.md). They are not prerequisites for shipping or using the v1 host.
+These areas are follow-up enhancements covered by [ADR-0028](0028-hosting-linking-multicast-enhancements.md). They are
+not prerequisites for shipping or using the v1 protocol-helper surface. ADR-0028 was written against the earlier
+host/channel framing and must be revised to align with this protocol-helper and execution-state boundary before those
+enhancements are implemented.
 
 ## Consequences
 
 Positive:
 
-- The host/channel model can be implemented and tested without designing a security-sensitive identity graph.
-- Existing and new channel packages can share one Starlette app, middleware stack, lifecycle, and target invocation path.
-- Session continuity is explicit and debuggable: two channels share history only when they produce the same `isolation_key`.
-- Hook invocation is centralized in the host, so channels do not each invent the call convention.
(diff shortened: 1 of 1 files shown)


docs/decisions/0027-hosting-channels.md
```diff
@@ -1,143 +1,213 @@
...
+- background/continuation delivery;
+- durable task runners (`DurableTaskRunner`, `InProcessTaskRunner`);
+- retry/replay policy (`RetryPolicy`);
+- fan-out, multicast, or all-linked delivery;
+- confidentiality tiers and `LinkPolicy`;
 - a host-level multi-agent router.
 
-These areas are follow-up enhancements covered by [ADR-0028](0028-hosting-linking-multicast-enhancements.md). They are not prerequisites for shipping or using the v1 host.
+These areas are follow-up enhancements covered by [ADR-0028](0028-hosting-linking-multicast-enhancements.md). They are
+not prerequisites for shipping or using the v1 protocol-helper surface.
```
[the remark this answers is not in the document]
eavanvalkenburg: Addressed in 05da96ea2: ADR-0027 now explicitly notes that ADR-0028 was written against the earlier host/channel framing and must be revised to align with the protocol-helper/execution-state boundary before implementation.

docs/decisions/0027-hosting-channels.md
```diff
@@ -1,143 +1,213 @@
...
 
-- route/lifecycle aggregation,
-- invocation of the target,
-- `ChannelSession(isolation_key=...)` to `AgentSession` resolution and caching,
-- `reset_session(isolation_key=...)`,
-- host-level middleware, including Foundry isolation middleware only when the Foundry hosting environment flag is present,
-- invocation of per-channel hooks (`ChannelRunHook`, `ChannelResponseHook`, `ChannelStreamUpdateHook`), and
-- workflow checkpoint wiring through an explicit `checkpoint_location`.
+- `responses_to_af(...)`
+- `response_from_af(...)`
```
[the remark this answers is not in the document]
eavanvalkenburg: Addressed in 05da96ea2: the helper names now use protocol-specific run terminology (`responses_to_run(...)` / `responses_from_run(...)`) instead of the older `*_af` naming, so the Responses helper is no longer generic or inconsistent.

docs/decisions/0027-hosting-channels.md
```diff
@@ -1,143 +1,213 @@
...
-- Built-in channel tests prove that routes, commands, startup, and shutdown callbacks are contributed by channels and aggregated by the host.
-- Session tests prove that identical `ChannelSession.isolation_key` values resolve to the same cached `AgentSession`, and `reset_session` rotates that mapping.
-- Channel tests prove that each channel renders only its own originating response; there is no host-level push, multicast, or active-channel delivery path.
-- Workflow tests or samples use an explicit `checkpoint_location`.
-- Foundry isolation middleware is documented and covered by integration or contract tests, including the non-Foundry case where raw isolation headers are ignored.
-- The v1 API and packages do not expose the removed symbols or packages listed in [Non-goals for v1](#non-goals-for-v1).
-- The Python spec is updated to match this simplified contract and uses "public", "stable", or "released" terminology for Agent Framework APIs.
+Before this ADR is considered implemented:
+
+- A Responses sample uses normal FastAPI route code plus `responses_to_af(...)`, `response_from_af(...)`, and
```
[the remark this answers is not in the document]
eavanvalkenburg: Addressed in 05da96ea2: the validation gate now refers to `responses_to_run(...)` / `responses_from_run(...)`, matching the helper naming section.

docs/decisions/0027-hosting-channels.md
```diff
@@ -1,144 +1,422 @@
...
+
+The app chooses which helper to call for that route and deployment. For example:
+
+- `responses_session_id(body)` from `agent-framework-hosting-responses`, which can return either a `resp_*` previous
+  response id or a `conv_*` conversation id when present;
+- `telegram_session_id(update)` from `agent-framework-hosting-telegram`, which can choose the chat, user, thread, or
+  other Telegram-native partitioning logic for that helper;
+- `activity_session_id(activity)`, `discord_session_id(interaction_or_message)`, or
+  `a2a_session_id(request_context)` from their respective protocol packages;
+- `foundry_user_isolation_key()` or `foundry_chat_isolation_key()` from `agent-framework-foundry-hosting`.
```
moonbox3, line 188: With the Validation Gates section removed, is anything still requiring these helpers to fail closed outside Foundry? The previous revision explicitly gated on tests proving raw isolation headers are ignored in the non-Foundry case. This revision keeps the helpers as the trusted platform-provided key source but drops that requirement, so an implementation that reads isolation headers verbatim would let a caller spoof another user's isolation key when the app runs outside Foundry.

docs/decisions/0027-hosting-channels.md
```diff
@@ -1,144 +1,422 @@
...
+```
+
+If a route wants to resume from a prior checkpoint, it explicitly chooses the checkpoint and passes it to
+`workflow.run(...)`:
+
+```python
+storage = await state.get_or_create_checkpoint_storage(session_id)
+target = await state.get_target()
+latest = await storage.get_latest(workflow_name=target.name)
+result = await target.run(
```
moonbox3: Should this resume snippet drop `message` when `latest` exists? `Workflow.run` validates that `message` and `checkpoint_id` are mutually exclusive (`_workflows/_workflow.py` raises "Cannot provide both 'message' and 'checkpoint_id'"), so as written this raises `ValueError` exactly when a checkpoint is found. Since the appendix imports the real `Workflow` and `CheckpointStorage`, this reads as a contract against the existing API rather than a future one.

docs/decisions/0027-hosting-channels.md
```diff
@@ -1,144 +1,422 @@
...
+    def build(self) -> Workflow: ...
+
+
+class SessionStore:
+    async def get(self, session_id: str) -> AgentSession | None: ...
+    async def set(self, session_id: str, session: AgentSession) -> None: ...
+    async def delete(self, session_id: str) -> None: ...
+
+
+class CheckpointStore:
```
moonbox3: Can a non-in-memory `CheckpointStore` actually implement `set`? `CheckpointStorage` is a protocol whose implementations hold live clients (the Cosmos-backed one holds a Cosmos client), so a Redis or Cosmos backed store cannot persist a live storage instance by value, yet the async get/set/delete surface and the "plain async storage" wording at line 96 signal out-of-process backends are the intent. Would mapping `session_id` to storage config/ids, or using one partitioned `CheckpointStorage`, keep multi-replica session continuity implementable?

docs/decisions/0027-hosting-channels.md
```diff
@@ -1,144 +1,422 @@
...
+    session_id = responses_session_id(body)
+    # create a new response_id for this run
+    response_id = create_response_id()
+
+    # in this space, the developer can make any adjustments to the request, i.e.:
+    run["options"]["store"] = False
+    run["options"].pop("model", None)
+
+    # load the session (or create a new one) - this is optional
+    session = await state.get_or_create_session(session_id or response_id)
```
moonbox3: Should this sketch show, or at least note, an ownership check on the caller-supplied id? After the shared api key check, any caller can send another conversation's `resp_*` or `conv_*` id and inherit that `AgentSession` with its full history, which is exactly the pattern line 187 says apps must authorize first. The Django sketch repeats the same shape, and these examples are what samples and implementations will copy. Even a `# verify this caller owns session_id before loading` comment would keep the sketches aligned with the trust note.

docs/decisions/0027-hosting-channels.md
```diff
@@ -1,144 +1,422 @@
...
+        try:
+            body = json.loads(request.body)
+        except json.JSONDecodeError:
+            return HttpResponseBadRequest("invalid json")
+
+        run = responses_to_run(body)
+        session_id = responses_session_id(body)
+        response_id = create_response_id()
+        options = run["options"]
+        session = async_to_sync(state.get_or_create_session)(session_id or response_id)
```
moonbox3: Has this Django shape been tried against a second request? Each `async_to_sync` call in a sync view runs on a fresh event loop, but the module-level `OpenAIChatClient` caches an `AsyncOpenAI` client whose pooled httpx connections bind to the loop of the first `target.run`, so later requests typically fail with "Event loop is closed". Could the sketch use an async view (`async def post`) so the whole request runs on one live loop?

eavanvalkenburg: Addressed in ca599835f: the Django sketch is now an async view (`async def post`) and uses await for state.get_or_create_session, state.get_target, target.run, and state.set_session. The example no longer uses async_to_sync around a module-level async client.

docs/decisions/0027-hosting-channels.md
eavanvalkenburg: Addressed in ca599835f: the ADR no longer frames durable CheckpointStore as storing live CheckpointStorage clients by value. It now distinguishes local/in-memory session_id -> CheckpointStorage state from durable/multi-replica session_id -> checkpoint_id (or equivalent cursor/config) state, and notes that WorkflowRunResult/events do not emit checkpoint ids by default; app/state code that owns storage can observe the latest id via storage.get_latest(workflow_name=target.name) after a run.

docs/decisions/0027-hosting-channels.md
eavanvalkenburg: Addressed in ca599835f: both the FastAPI and Django sketches now include an explicit comment before loading the session that the app must verify this caller owns the supplied session id; API-key auth alone does not prove ownership of a caller-supplied resp_* or conv_* id.

docs/decisions/0027-hosting-channels.md
eavanvalkenburg: Addressed in ca599835f: ADR-0027 now explicitly says platform-provided isolation helpers must fail closed outside their trusted hosting environment. It also states Foundry-specific helpers may read values established by Foundry hosting middleware, but must not treat raw request headers as trusted Foundry isolation outside Foundry, and implementations must test that non-Foundry requests do not accept spoofable isolation headers.

docs/decisions/0027-hosting-channels.md
eavanvalkenburg: Addressed in ca599835f: the workflow resume snippet now branches. If no checkpoint id is available it calls run(message=..., checkpoint_storage=...). If a checkpoint id exists it calls run(checkpoint_id=..., checkpoint_storage=...) and does not pass message, matching the current Workflow.run validation.

docs/decisions/0027-hosting-channels.md
```diff
@@ -1,144 +1,468 @@
...
+# Python protocol helpers and optional execution state
 
 ## Context and Problem Statement
 
-Agent Framework has several protocol-specific hosting surfaces. App authors who want one agent or workflow on multiple protocols must compose servers, routes, middleware, session handling, and lifecycle code by hand.
+Agent Framework needs to help applications expose agents and workflows over external protocols such as OpenAI
+Responses, Telegram, Activity Protocol, and future transports.
 
-We will introduce a small Python hosting core that owns the common server shape and leaves protocol details inside channel packages. The first public contract must be intentionally narrow so Python can ship a base contract before adding identity linking, proactive delivery, or multicast behavior. Other language implementations may reuse the same conceptual boundary, but this ADR records the Python decision.
+The first version of this ADR chose a host/channel model: channel packages contributed routes, middleware, commands,
```
TaoChenOSU: nit: perhaps do not mention the first version at all if it won't exist in the code base.

docs/decisions/0027-hosting-channels.md
```diff
@@ -1,144 +1,468 @@
...
+hides the protocol-specific contract behind a second abstraction.
+
+Protocol packages should consider these helper families. This table is a set of examples, not a required protocol or
+checklist. Not every protocol needs every helper, but when a protocol has the concept the naming should stay consistent:
+
+| Helper family | Shape | Purpose |
+| --- | --- | --- |
+| Run conversion | `<protocol>_to_run(...)` | Convert one protocol-native call/update/request into `Agent.run` or `Workflow.run` values. |
+| Final rendering | `<protocol>_from_run(...)` | Convert a final `AgentResponse` / workflow result into protocol-native response payloads or operations. |
+| Stream rendering | `<protocol>_stream_from_run(...)` | Convert `ResponseStream` / workflow updates into protocol-native events or operations. |
```
TaoChenOSU: nit: `<protocol>_from_streaming_run(...)`

docs/decisions/0027-hosting-channels.md
```diff
@@ -1,144 +1,468 @@
...
+
+`agent.run(...)` may update the session object (for example, with service continuation state), so the explicit store call
+belongs after the run, not before it.
+
+The session id is a partition key, not proof of identity. App or platform code must authenticate and authorize any
+externally supplied key before using it.
+
+### Workflow checkpoints
+
+Workflow checkpointing is execution state, not protocol state. `WorkflowState` pairs a workflow target with checkpoint
```
TaoChenOSU, line 235: Is a "workflow target" a workflow instance? How does this target survive across sessions?

docs/decisions/0027-hosting-channels.md
eavanvalkenburg: good point

docs/decisions/0027-hosting-channels.md
eavanvalkenburg: See line 89, it refers to: `- ``WorkflowState``: one ``Workflow``, ``WorkflowBuilder``-shaped builder, orchestration builder, or workflow factory;`

In other words, depending on the workflow, hosting api, use case and other things, you can choose to build a workflow anew for each run, or reuse a existing one, the WorkflowState has a target attribute that can be set (WorkflowState itself is optional), if needed augmented with a small lookup between i.e. response_id and checkpoint_id to continue when needed (where response_id could also be chat_id for telegram, or whatever identifier there is for a particular channel)

eavanvalkenburg: And also note (and I added this as well) that the developer chooses where they intend to run, and whether that is a persistent server, like a Container App, or a transient thing like AzFunctions, because that obviously has major implications for how to design storage and which is why I have set this up with either a concrete instance of a agent or a callable, so that both options are present.

docs/decisions/0027-hosting-channels.md
```diff
@@ -1,144 +1,476 @@
...
+- background tasks and webhook acknowledgement policy;
+- native protocol SDK clients and outbound calls;
+- command registration and command dispatch;
+- request/response status codes and framework-specific error handling;
+- choosing the isolation/session id source for the current deployment and route.
+
+The optional execution-state helpers, if provided, are limited to shared execution state:
+
+- `AgentState`: one `SupportsAgentRun`-compatible target plus a `SessionStore`;
+- `WorkflowState`: one `Workflow`, `WorkflowBuilder`-shaped builder, orchestration builder, or workflow factory;
```
moonbox3, line 94: Could a concrete `Workflow` target share runner state across callers in a persistent host? `Workflow.run` preserves state on the instance between calls, so a singleton `WorkflowState(Workflow)` can carry one authorized session's executor/shared state into another unless every run restores an isolated checkpoint or builds a fresh workflow. Should this constrain concrete `Workflow` targets to single-session use and recommend a builder/factory for multi-session hosting?

docs/decisions/0027-hosting-channels.md
```diff
@@ -1,144 +1,476 @@
...
+
+For workflow targets, app code adapts the protocol helper output into the workflow's expected input and invokes the
+workflow through the state object's target:
+
+```python
+# session_id must already be authenticated and authorized for this caller
+target = await state.get_target()
+result = await target.run(message=workflow_input, checkpoint_storage=checkpoint_storage)
+latest = await checkpoint_storage.get_latest(workflow_name=target.name)
+if latest is not None:
```
moonbox3, line 253: Could this race when `checkpoint_storage` is shared across sessions? `get_latest(workflow_name=target.name)` is global to the workflow name in the current `CheckpointStorage` protocol, so session B can save a later checkpoint between session A's run and this lookup, causing A's cursor to resume B's state. 

The ownership checks above do not mitigate a checkpoint id selected after authorization; should the sketch require per-session scoped storage here or capture the checkpoint id from this run's `save(...)` path instead of querying global latest?

docs/decisions/0027-hosting-channels.md
eavanvalkenburg: so, later on I mention that this is up to the developer and depends on many things, persistent servers require a different setup, likely with a workflow factory/builder rather then a single concrete workflow, but if you are running a transient server (like a azfunc) that distinction is somewhat less relevant, but I wanted to give the developer this choice, rather then making a choice and then having to build in parameters to control this behavior.

docs/decisions/0027-hosting-channels.md
eavanvalkenburg: yes, see my note on the other comment, and it might be useful to see if we need to add something to checkpoint storage to account for this

docs/decisions/0027-hosting-channels.md
eavanvalkenburg: and of course our samples need to be very explicit about this.

docs/decisions/0027-hosting-channels.md
```diff
@@ -1,144 +1,476 @@
...
+calls. Agent Framework should not duplicate those surfaces unless a specific hosting environment requires it.
 
 ## Decision Drivers
 
-- Keep the first host easy to explain: one app, one hostable target, one or more channels.
-- Reuse Agent Framework's existing agent, workflow, session, history, and checkpoint primitives.
-- Let channel packages own protocol parsing, protocol responses, authentication details, and native command surfaces.
-- Make session continuity explicit through a channel-supplied `ChannelSession(isolation_key=...)`.
+- Keep the released surface small enough to explain without first teaching a channel framework.
+- Provide reusable Agent Framework run translation that works with FastAPI and other web frameworks.
```
rogerbarreto: Not sure why FastAPI is getting named specifically, if we mention names ideally for an abstraction I would expect 2+ named options to start with.

```suggestion
- Provide reusable Agent Framework run translation that works with all web frameworks mentioned, starting with FastAPI and Responses implementations.
```

eavanvalkenburg: mostly because it is currently one of the most popular ones for Python, I use FastAPI and Django below in the sketch of the implementation, adding Django here as an example as well.

docs/decisions/0027-hosting-channels.md
```diff
@@ -1,144 +1,476 @@
...
+- Let app/framework code own route declaration, auth, middleware, native SDK clients, command handling, and background
+  work.
+- Keep stateful execution support explicit: session lookup/storage and workflow checkpoint lookup/storage may still need
+  a small AF-owned home.
 - Avoid approving cross-channel identity and delivery semantics before their safety model is reviewed.
```
rogerbarreto: Not sure this is ADR related, sounds like instructions for the Agent.

```suggestion
```

eavanvalkenburg: it related to the scope of this ADR, but we can remove.

docs/decisions/0027-hosting-channels.md
```diff
@@ -1,144 +1,476 @@
...
 
-- Good: the richest cross-channel scenarios are available immediately.
-- Neutral: the host becomes the natural place to demonstrate identity and delivery policy.
-- Bad: v1 becomes a security-sensitive identity and delivery system before the safety model is reviewed.
+- Good: one object can assemble routes, channels, session handling, hooks, and lifecycle callbacks.
+- Good: app code using the supported host shape can be short.
+- Bad: the framework owns concerns already handled by web frameworks, protocol SDKs and/or other services.
+- Bad: users must understand `Channel`, contribution, hook, and host-dispatch concepts before they can see how a request
+  becomes `agent.run(...)`.
+- Bad: the abstraction is hard to reuse outside the chosen ASGI shape.
```
rogerbarreto: What is ASGI?

```suggestion
- Bad: the abstraction is hard to reuse outside the chosen <ASGI> shape.
```

eavanvalkenburg: Asynchronous Server Gateway Interface, a standard way of defining webservers for python (as opposed to WSGI which is non-async, given the nature of agents, async is def preferred here)

docs/decisions/0027-hosting-channels.md
```diff
@@ -1,144 +1,476 @@
...
 
-1. Keep only protocol-specific hosts.
-2. Ship a large hosting core with identity linking, authorization, background delivery, active-channel routing, and multicast in v1.
-3. Ship a minimal host/channel core now and track linking/multicast as follow-up work.
+1. Create protocol-specific hosts.
+2. Ship a full host/channel framework with route contribution and channel hooks.
+3. Ship protocol conversion helpers plus optional execution state.
 
-### Keep only protocol-specific hosts
+### Create protocol-specific hosts
```
rogerbarreto: I would put the numbers in the options so I don't have to come back to know which is which as well as a short description what is the actual option approach, possibly with a mermaid graph or something visual.

```suggestion
### 1. Create protocol-specific hosts

Small description of the proposal
```

docs/decisions/0027-hosting-channels.md
eavanvalkenburg: made the language clearer now

docs/decisions/0027-hosting-channels.md
```diff
@@ -1,144 +1,476 @@
...
+- Good: small state objects can still own target-coupled state: `AgentState` pairs an agent target with a `SessionStore`,
+  and `WorkflowState` resolves a workflow target while reusing the existing `CheckpointStorage` abstraction.
+- Good: provides maximum configurability in handling input and outputs (outside of the conversions)
+- Bad: building a first iteration of a new Host is more verbose.
+- Bad: samples show more explicit route/client code than a fully assembled channel host.
 
 ## Decision Outcome
 
-Chosen option: **minimal host/channel core now, follow-up enhancements later**.
+Chosen option: **protocol helpers plus optional execution state**.
```
rogerbarreto: nit.

```suggestion
Chosen option: **3. Ship protocol helpers plus optional execution state**.
```

docs/decisions/0027-hosting-channels.md
eavanvalkenburg: good point

docs/decisions/0027-hosting-channels.md
```diff
@@ -1,144 +1,476 @@
 ---
 status: accepted
 contact: eavanvalkenburg
-date: 2026-06-11
+date: 2026-06-30
 deciders: eavanvalkenburg
```
rogerbarreto, line 5: nit

```suggestion
date: 2026-07-07
deciders: eavanvalkenburg
consulted: rogerbarreto
```

docs/decisions/0027-hosting-channels.md
```diff
@@ -1,144 +1,476 @@
...
+
+Examples:
+
+- `responses_to_run(...)`, `responses_from_run(...)`, `responses_from_streaming_run(...)`,
+  `responses_session_id(...)`;
+- `telegram_to_run(...)`, `telegram_from_run(...)`, `telegram_from_streaming_run(...)`,
+  `telegram_session_id(...)`, `telegram_command(...)`;
+- `activity_to_run(...)`, `activity_from_run(...)`, `activity_session_id(...)`, `activity_command(...)`;
+- `discord_to_run(...)`, `discord_from_run(...)`, `discord_session_id(...)`, `discord_command(...)`.
+
```
rogerbarreto, line 129: Can we have a more complete example exercising one provider like "Telegram" how would that be used, like a small more contextualized code snippet.

```suggestion
- `responses_to_run(...)`, `responses_from_run(...)`, `responses_from_streaming_run(...)`,
  `responses_session_id(...)`;
- `telegram_to_run(...)`, `telegram_from_run(...)`, `telegram_from_streaming_run(...)`,
  `telegram_session_id(...)`, `telegram_command(...)`;
- `activity_to_run(...)`, `activity_from_run(...)`, `activity_session_id(...)`, `activity_command(...)`;
- `discord_to_run(...)`, `discord_from_run(...)`, `discord_session_id(...)`, `discord_command(...)`.

Real use-case examples including more pience of code...
```

docs/decisions/0027-hosting-channels.md
```diff
@@ -1,144 +1,476 @@
...
+performs acknowledgements, sends/edits messages, resolves protected file URLs, applies rate limits, and registers
+handlers.
+
+### Security responsibilities for application builders
+
+The application builder owns the trust boundary. Protocol helper packages can parse native payloads and expose candidate
+ids or operations, but they do not authenticate callers, authorize access to state, or decide which side effects are
+allowed.
+
+Application code that uses these helpers must:
```
rogerbarreto: We might also capitalize the individual items.

```suggestion
Application code that uses these helpers are responsible for:
```

docs/decisions/0027-hosting-channels.md
eavanvalkenburg: that is in the appendix, wanted to keep a more complete sample together that shows how they are all used, without introducing it here.

docs/decisions/0027-hosting-channels.md
```diff
@@ -1,144 +1,476 @@
...
+- authorize command/action effects such as reset, cancel, approve, submit, or tool invocation after parsing them;
+- opt in explicitly before resolving protected media/resource/file URLs and passing them to a remote model provider;
+- persist post-run session or checkpoint state only after `agent.run(...)`, `workflow.run(...)`, or stream finalization has
+  updated that state.
+
+For Foundry specifically, helpers may read values established by Foundry hosting middleware, but must not treat raw
+request headers as trusted Foundry isolation when the app is running outside Foundry. Implementations must test that
+non-Foundry requests do not accept spoofable isolation headers as platform-provided keys.
+
+For durable workflow checkpointing, the checkpoint boundary must be at least as specific as the authorized session/tenant
```
rogerbarreto: Durable could have many meanings here, is maybe Long running could better describe it.

```suggestion
For long-running workflow checkpointing, the checkpoint boundary must be at least as specific as the authorized session/tenant
```

docs/decisions/0027-hosting-channels.md
```diff
@@ -1,144 +1,476 @@
...
+```
+
+If the protocol mints a new continuation id as part of the response being created (for example, OpenAI Responses
+`resp_*` ids), store the **post-run** session explicitly under that new id:
+
+```python
+session = await state.get_or_create_session(previous_response_id)
+target = await state.get_target()
+result = await target.run(messages, session=session, options=options)
+await state.set_session(response_id, session)
```
rogerbarreto, line 224: Not sure why is the caller responsibility to update the session and keep track of the new post-run id, AFAIK the session instance object already carries this responsibility.

docs/decisions/0027-hosting-channels.md
```diff
@@ -1,144 +1,476 @@
...
+- identity allowlists or authorization policy (`IdentityAllowlist`, `AuthPolicy`);
+- response routing beyond the originating protocol (`ResponseTarget`, active channel, specific linked channel,
+  `all_linked`);
+- push or payload codecs (`ChannelPush`, `ChannelPushCodec`);
+- background/continuation delivery;
+- durable task runners (`DurableTaskRunner`, `InProcessTaskRunner`);
+- retry/replay policy (`RetryPolicy`);
+- fan-out, multicast, or all-linked delivery;
+- confidentiality tiers and `LinkPolicy`;
+- a host-level multi-agent router.
```
rogerbarreto: Will the protocol helper option be able to cover the proposed items in the list in the future?

docs/decisions/0027-hosting-channels.md
```diff
@@ -1,144 +1,476 @@
...
-- Built-in channel tests prove that routes, commands, startup, and shutdown callbacks are contributed by channels and aggregated by the host.
-- Session tests prove that identical `ChannelSession.isolation_key` values resolve to the same cached `AgentSession`, and `reset_session` rotates that mapping.
-- Channel tests prove that each channel renders only its own originating response; there is no host-level push, multicast, or active-channel delivery path.
-- Workflow tests or samples use an explicit `checkpoint_location`.
-- Foundry isolation middleware is documented and covered by integration or contract tests, including the non-Foundry case where raw isolation headers are ignored.
-- The v1 API and packages do not expose the removed symbols or packages listed in [Non-goals for v1](#non-goals-for-v1).
-- The Python spec is updated to match this simplified contract and uses "public", "stable", or "released" terminology for Agent Framework APIs.
 
-## More Information
+class WorkflowState:
```
rogerbarreto, line 374: Suggest also adding the changes to the PR, so any extra findings including use-cases, code paths, limitations, are reflected into the ADR observation.

docs/decisions/0027-hosting-channels.md
eavanvalkenburg: because this is specific to the way you want to run your server, if you do not want persistence, then you wouldn't call this method.

docs/decisions/0027-hosting-channels.md
eavanvalkenburg: did an update to rewrite these and noted how they are impacted by the new design

moonbox3 (approved):

rogerbarreto (approved):

Issue #6711, opened 2026-06-24
Python: Add AgentSkillsSourceContext to AgentSkillsSource.get_skills


SergeyMenshykh:
**Problem**: `AgentSkillsSource.get_skills` has no access to invocation context, preventing skill sources from making context-aware decisions (e.g., filtering skills per agent).

**Proposal**: Add an `AgentSkillsSourceContext` parameter to `get_skills` that exposes the invoking agent. The context is created internally by the skills provider and passed through the source pipeline.

This mirrors the .NET change in #6710.

PR #6797, opened 2026-06-29
.NET: Add AgentSkillsSourceContext to AgentSkillsSource.GetSkillsAsync

SergeyMenshykh:
## Summary

Adds an `AgentSkillsSourceContext` (carrying `Agent` and optional `Session`) to `AgentSkillsSource.GetSkillsAsync`, so skill sources, filters, and caching can make context-aware decisions. The context is built by `AgentSkillsProvider` from the `InvokingContext` and flows through all sources and decorators.

## Changes

- `AgentSkillsSourceContext` exposes `Agent` and `Session`, supplied from the invoking context.
- `FilteringAgentSkillsSource` predicate now receives an `AgentSkillFilterContext` bundling the skill and the source context.
- `CachingAgentSkillsSource` supports per-key cache isolation via `CachingAgentSkillsSourceOptions.CacheIsolationKeySelector`; a null selector keeps the existing shared-cache behavior.

Closes: https://github.com/microsoft/agent-framework/issues/6710

PR #6847, opened 2026-06-30
Python: [BREAKING] Extract caching from SkillsProvider into a CachingSkillsSource decorator

### LINKED ISSUE #6767: Python: Extract caching from SkillsProvider into a CachingSkillsSource decorator
Extract caching out of `SkillsProvider` into a dedicated `CachingSkillsSource` decorator, so caching becomes a composable layer in the skills-source pipeline instead of logic baked into the provider (alongside `Aggregating`/`Filtering`/`DeduplicatingSkillsSource`).

Today `SkillsProvider` owns caching internally (the `_cached_context` field, `_get_or_create_context`, and the `disable_caching` flag). This mixes context construction with caching and prevents controlling where caching sits relative to the other source decorators.

This mirrors the .NET work in #6765 to keep the two implementations aligned.

giles17:
### Motivation & Context

`SkillsProvider` currently bakes caching into itself: the `_cached_context`
field, the `_get_or_create_context()` method, and a `disable_caching` branch in
`before_run`. This mixes context construction with caching and prevents caching
from being a composable layer in the skills-source pipeline (alongside
`Aggregating`/`Filtering`/`Deduplicating`/`DelegatingSkillsSource`).

This change extracts caching into a dedicated `CachingSkillsSource` decorator so
caching becomes a swappable pipeline layer rather than logic welded into the
provider. It mirrors the .NET work in #6768 to keep the two implementations
aligned, and is a prerequisite for #6711 (adding context to skill-source
retrieval).

  - New `CachingSkillsSource(DelegatingSkillsSource)` decorator (experimental,
    `SKILLS` feature) that caches the inner source's skills list. It shares a
    single in-flight fetch across concurrent callers (via an `asyncio.Lock` with
    a double-checked cache) and only stores the result on success, so a failed
    fetch is not cached and the next call retries.
  - `SkillsProvider` now wraps its resolved source in a `CachingSkillsSource` by
    default and rebuilds instructions/tools from the (source-cached) skills each
    run. The provider's internal cache (`_cached_context`,
    `_get_or_create_context`) is removed; `before_run` is now branch-free.
  - `disable_caching=True` (on `SkillsProvider(...)` and `from_paths(...)`) now
    skips wrapping the source instead of toggling an internal flag — preserving
    the existing public kwarg and its meaning. It is the Python equivalent of
    .NET's builder `DisableCaching()`.
  - `CachingSkillsSource` is exported from `agent_framework`.

  - Caching is now composable: callers can place `CachingSkillsSource` explicitly
    in a pipeline, and `disable_caching=True` opts out of the default wrap.
  - Behavioral nuance: only the skills list is cached now (not the built
    instructions/tools, which are cheap, deterministic functions of the skills
    and are rebuilt each run). The expensive discovery in `get_skills()` is still
    cached. This is a `[BREAKING — experimental]` change to skills internals.

  - The concurrency/retry semantics of `CachingSkillsSource.get_skills` and that
    the default-wrap vs `disable_caching` behavior faithfully mirrors .NET #6768.

### Related Issue

Fixes #6767

### Contribution Checklist


eavanvalkenburg (approved):

SergeyMenshykh:
LGTM. This PR also add cache isolation key selector https://github.com/microsoft/agent-framework/pull/6797/changes#diff-a80018eeb1587fa83185680bb40b0a428a32fb71d302eed0224f7ade8d3ebfeb for .NET cache source. Same needed to be done on python as well. Not necessary in this PR.

SergeyMenshykh (approved):

PR #6850, opened 2026-07-01
Python: Add progressive MCP disclosure

### LINKED ISSUE #5821: Python: add progressive MCP discovery/dispatch mode for large MCP servers
### Feature request

The Python SDK has good MCP primitives today: `MCPStreamableHTTPTool` connects to MCP servers, loads remote tools, filters with `allowed_tools`, exposes remote tools as `FunctionTool`s, parses MCP results, raises on `isError`, and injects OTel metadata.

For large production MCP servers, it would be useful to also support a progressive model-facing MCP mode:

- expose a small stable discovery tool, e.g. `list_mcp_tools(server?: str)`
- expose a small stable dispatch tool, e.g. `call_mcp(server: str, tool: str, arguments: dict)`
- back both with SDK MCP transports/sessions/result parsing/OTel
- resolve `call_mcp` against the SDK-loaded filtered function list before execution
- keep `allowed_tools` / approval / remote-name metadata semantics intact
- avoid exposing `MCPStreamableHTTPTool.call_tool()` raw to the model

### Why

Direct first-class MCP exposure is excellent when the remote surface is small. With larger MCP servers or multiple servers, exposing every remote schema up front can add significant prompt/tool-definition overhead and can force skills/prompts to be coupled to every remote tool shape.

A progressive mode lets the model discover exact remote tools and schemas only when needed, while preserving SDK-owned MCP transport, telemetry, result parsing, and error behavior.

### Current workaround

Applications can implement this as a thin compatibility adapter by creating SDK `MCPStreamableHTTPTool` instances internally, reading their filtered `.functions`, and exposing local `FunctionTool`s for `list_mcp_tools` / `call_mcp`. That works, but it duplicates policy-sensitive behavior that would be better owned by the SDK.

### Suggested API shape

One possible shape:

```python
mcp = MCPStreamableHTTPTool(..., allowed_tools=[...])
progressive_tools = mcp.as_progressive_tools(
    list_tool_name="list_mcp_tools",
    call_tool_name="call_mcp",
)
agent = Agent(..., tools=[*progressive_tools])
```

Or an agent/tool option such as `exposure_mode="progressive"`.

The important behavior is that the model sees the small progressive surface, while the SDK still owns connection lifecycle, allowed tool filtering, approval handling, MCP result parsing, exceptions, and OTel propagation.

eavanvalkenburg:
### Motivation & Context

Large MCP servers can expose many tools, which frontloads a lot of schema into the model context even when only a small subset is needed. This adds progressive disclosure support for MCP tools so agents can discover, load, and unload MCP tool schemas on demand while keeping the existing `allowed_tools` boundary intact.

  - Added `use_progressive_disclosure` and `always_load` constructor options to all Python MCP tool classes immediately after `allowed_tools` on the transport constructors, while preserving positional compatibility for the base `MCPTool` constructor.
  - Progressive mode exposes prefix-aware `list_mcp_tools`, `load_tool`, and `unload_tool` loader tools, plus any allowed tools selected by `always_load`; other allowed tools are added through `FunctionInvocationContext.add_tools(...)` when loaded and can be removed through `FunctionInvocationContext.remove_tools(...)` when unloaded.
  - `load_tool` and `unload_tool` now accept either a single tool name or a list of tool names, applying successful changes in batches and returning per-tool model-visible results for mixed outcomes.
  - Persisted dynamically loaded MCP tool names on the `MCPTool` instance so loaded tools remain visible across later runs, including approval continuations, until explicitly unloaded.
  - Kept `allowed_tools` as the outer policy boundary, reserved loader tool names to avoid collisions, surfaced recoverable loader failures as model-visible string results, and emitted the progressive-tools experimental warning deterministically at construction.
  - Added a runnable MCP sample demonstrating progressive disclosure with a self-spawned stdio MCP server, including load and unload behavior.
  - Added tests for transport constructors, filtering, prefixed loader names, idempotent loading, batch load/unload, mixed batch results, unloading protection for `always_load`, persisted loaded state, approval preservation, runtime kwargs/header-provider behavior, loader error messages, warning behavior, and `load_tools=False` validation.
  - Existing MCP tool behavior remains unchanged unless progressive disclosure is explicitly enabled.
  - Agents can keep initial MCP tool context smaller while still allowing the model to discover, load, and later unload one or more permitted tools during the run.
  - The progressive disclosure semantics around `allowed_tools`, `always_load`, prefixed loader names, persisted loaded state, batch load/unload behavior, and the use of the existing progressive tool exposure mechanism.

### Related Issue

Fixes #5821

### Contribution Checklist


python/packages/core/agent_framework/_mcp.py
```diff
@@ -807,25 +822,132 @@ def _prepare_message_for_mcp(
...
+        )
+        return any(name in names for name in candidate_names)
+
+    def _progressive_functions(self) -> list[FunctionTool]:
+        """Return the initial model-facing function list for progressive disclosure."""
+        initial_functions = list(self._progressive_loader_tools())
+        initial_functions.extend(
+            func for func in self._filtered_functions() if self._function_matches_names(func, self._always_load_names)
+        )
+        return initial_functions
```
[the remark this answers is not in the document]
eavanvalkenburg: Addressed in d039c1b3e: progressive initial functions now skip any always-loaded MCP tool whose local name collides with the reserved loader names, with test coverage for the collision case.

python/packages/core/agent_framework/_mcp.py
```diff
@@ -807,25 +822,132 @@ def _prepare_message_for_mcp(
...
+            tools.append({
+                "name": func.name,
+                "remote_name": remote_name if isinstance(remote_name, str) else func.name,
+                "description": func.description,
+                "parameters": func.parameters(),
+                "approval_mode": func.approval_mode,
+                "loaded": self._is_progressive_function_loaded(func, ctx),
+                "always_loaded": self._function_matches_names(func, self._always_load_names),
+            })
+        return tools
```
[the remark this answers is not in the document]
eavanvalkenburg: Addressed in d039c1b3e: docs_list/list_mcp discovery now omits allowed MCP tools whose local name collides with the loader names, and the not-available message also only advertises loadable tools. Added coverage for the discovery filter.

python/packages/core/agent_framework/_mcp.py
```diff
@@ -807,25 +822,132 @@ def _prepare_message_for_mcp(
...
+    def _resolve_progressive_function(self, tool_name: str) -> FunctionTool:
+        """Resolve a progressive disclosure tool name against allowed MCP functions."""
+        matches = [func for func in self._filtered_functions() if self._function_matches_names(func, {tool_name})]
+        matches.extend(func for func in self._filtered_functions() if func.name == tool_name and func not in matches)
+        if not matches:
+            available = ", ".join(func.name for func in self._filtered_functions()) or "none"
+            raise ToolExecutionException(f"MCP tool '{tool_name}' is not available. Available tools: {available}.")
+        if len(matches) > 1:
+            raise ToolExecutionException(f"MCP tool name '{tool_name}' is ambiguous.")
+        return matches[0]
```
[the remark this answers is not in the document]
eavanvalkenburg: Addressed in d039c1b3e: load_tool now detects loader-name collisions before ctx.add_tools(...) and raises a clear ToolExecutionException telling callers to set tool_name_prefix or exclude the colliding MCP tool. Added focused coverage.

python/samples/02-agents/mcp/mcp_progressive_disclosure.py
```diff
@@ -0,0 +1,151 @@
...
+
+import asyncio
+import sys
+from typing import Any
+
+from agent_framework import Agent, MCPStdioTool
+from agent_framework.openai import OpenAIChatClient
+from dotenv import load_dotenv
+
+"""
```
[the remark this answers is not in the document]
eavanvalkenburg: Addressed in d039c1b3e: the sample documentation string is now assigned to __doc__ so it is intentional top-level documentation under the sample file-order style.

eavanvalkenburg:
Addressed the DevFlow flagged issue in d039c1b3e: progressive disclosure now treats the loader names as reserved in the progressive surface, filters always_load/discovery collisions, and returns a clear load_tool error that points users to tool_name_prefix or excluding the colliding MCP tool. Added regression tests for these cases.

python/packages/core/agent_framework/_mcp.py
```diff
@@ -807,24 +822,160 @@ def _prepare_message_for_mcp(
...
-            if any(name in allowed_names for name in candidate_names):
-                filtered_functions.append(func)
-        return filtered_functions
+        if any(tool_item is func for tool_item in ctx.tools):
+            return f"MCP tool '{func.name}' is already available."
+        try:
+            ctx.add_tools(func)
+        except ValueError as ex:
+            raise ToolExecutionException(str(ex), inner_exception=ex) from ex
+        return f"Loaded MCP tool '{func.name}'. It is available on the next model iteration."
```
moonbox3: Have we considered what happens when a progressively loaded tool requires approval and the approval comes back in a new run? `functions` rebuilds the surface as loaders plus `always_load` only, so the next `run()`'s tool map no longer contains the loaded tool, and the approved call hits the `tool is None` hosted-tool pass-through in `_tools.py` (around line 1490) and never executes. Since `always_require` approval always spans runs, that seems to make progressive disclosure incompatible with approval round-trips. Should loaded-tool state persist across runs, or should that limitation be called out?

python/packages/core/agent_framework/_mcp.py
```diff
@@ -807,24 +822,160 @@ def _prepare_message_for_mcp(
...
+                f"MCP tool '{func.name}' conflicts with progressive disclosure loader tool name(s): "
+                f"{loader_names}. Set tool_name_prefix or exclude the colliding MCP tool."
             )
-            if any(name in allowed_names for name in candidate_names):
-                filtered_functions.append(func)
-        return filtered_functions
+        if any(tool_item is func for tool_item in ctx.tools):
+            return f"MCP tool '{func.name}' is already available."
+        try:
+            ctx.add_tools(func)
```
moonbox3: Is there a reason a stable-looking constructor flag routes through the experimental `add_tools` (`@experimental(ExperimentalFeature.PROGRESSIVE_TOOLS)`)? The first model-triggered `load_tool` call emits an `ExperimentalWarning` mid-run, and under warnings-as-errors (a common pytest/CI setup) that raises inside the tool call and gets masked to `Error: Function failed.`, so loading silently never works. Should `use_progressive_disclosure` warn as experimental at construction so the signal shows up deterministically, or should `add_tools` be promoted to stable first?

python/packages/core/agent_framework/_mcp.py
```diff
@@ -395,6 +397,8 @@ def __init__(
         description: str | None = None,
         approval_mode: (Literal["always_require", "never_require"] | MCPSpecificApproval | None) = None,
         allowed_tools: Collection[str] | None = None,
+        use_progressive_disclosure: bool = False,
```
moonbox3: Could we make these two keyword-only or append them at the end of the signature? The base `MCPTool.__init__` has no `*`, so a subclass calling `super().__init__(name, desc, approval, allowed, "github")` that previously bound `tool_name_prefix` now silently binds a truthy `use_progressive_disclosure` and drops the prefix, with no TypeError. The shipped transports are keyword-only so they are unaffected, but `MCPTool` itself is documented for subclassing.

python/packages/core/agent_framework/_mcp.py
```diff
@@ -807,24 +822,160 @@ def _prepare_message_for_mcp(
...
+        if not matches:
+            available = (
+                ", ".join(
+                    func.name
+                    for func in self._filtered_functions()
+                    if not self._function_collides_with_progressive_loader(func)
+                )
+                or "none"
+            )
+            raise ToolExecutionException(f"MCP tool '{tool_name}' is not available. Available tools: {available}.")
```
moonbox3, line 967: What does the model actually see when it typos a tool name here in the default config? The invocation loop converts `ToolExecutionException` to a bare `Error: Function failed.` unless `include_detailed_errors=True` (default False), so this `Available tools: ...` guidance, and the ambiguity and loader-collision hints below, never reach the model. Since these loader errors are model-facing recovery guidance rather than internal details, could the loaders return the message as a string result instead of raising?

python/packages/core/agent_framework/_mcp.py
eavanvalkenburg: Fixed in b26a0b2fd by persisting progressively loaded MCP tool names on the MCPTool instance. The `functions` surface now includes previously loaded tools on later runs, so approval continuations can resolve the approved tool from the tool map instead of falling through as an unknown hosted tool. Added focused coverage for the persisted loaded surface.

python/packages/core/agent_framework/_mcp.py
eavanvalkenburg: Fixed in b26a0b2fd. Progressive MCP now emits an ExperimentalWarning deterministically at construction when `use_progressive_disclosure=True`, and the internal `add_tools` / `remove_tools` calls are wrapped to suppress the progressive-tools warning during model-triggered load/unload. Added a warnings-as-errors test around load and unload.

python/packages/core/agent_framework/_mcp.py
eavanvalkenburg: Fixed in b26a0b2fd by moving `use_progressive_disclosure` and `always_load` to the end of the base `MCPTool.__init__` signature, preserving positional `tool_name_prefix` compatibility for subclasses. The transport constructors remain keyword-only and keep the requested ordering immediately after `allowed_tools`. Added a positional compatibility test.

python/packages/core/agent_framework/_mcp.py
eavanvalkenburg: Fixed in b26a0b2fd. Recoverable loader failures now return model-visible string results instead of raising `ToolExecutionException`, so missing/filtered tools, ambiguous names, and loader-name collisions surface actionable guidance to the model under the default error settings. Framework misuse, such as invoking the loader without a live tools list, still raises. Added tests for the model-facing messages.

python/packages/core/agent_framework/_mcp.py
```diff
@@ -807,24 +837,221 @@ def _prepare_message_for_mcp(
...
+                "remote_name": remote_name if isinstance(remote_name, str) else func.name,
+                "description": func.description,
+                "parameters": func.parameters(),
+                "approval_mode": func.approval_mode,
+                "loaded": self._is_progressive_function_loaded(func, ctx),
+                "always_loaded": self._function_matches_names(func, self._always_load_names),
+            })
+        return tools
+
+    async def _load_progressive_mcp_tool(self, ctx: FunctionInvocationContext, tool: str) -> str:
```
TaoChenOSU: Should this tool take a list of strings for the tool names? Currently for every tool that the model wants to load, it has to issue parallel tools call (if the service even supports it) or multiple calls in sequence just to load the tools, essentially doubling the tool call counts and iteration number, which adds significant latency (mainly model roundtrips). This will also consume requests if customers are constrained by request quotas.

TaoChenOSU: Same for unloading

moonbox3 (approved):

python/packages/core/agent_framework/_mcp.py
eavanvalkenburg: good one, adding now.

eavanvalkenburg: same

TaoChenOSU (approved):

PR #6860, opened 2026-07-01
Python: Sample for using WorkIQ MCP server using a gateway for labelling and IFC policy evaluation  

shrutitople:
### Motivation & Context

FIDES enforces information-flow-control policy locally today. This change lets a FIDES-secured agent delegate policy decisions to an external FIDES Gateway fronting its MCP servers, while still tracking labels and surfacing approvals locally. It also adds a runnable WorkIQ email/teams sample as a reference.

SecureMCPToolProxy(gateway_policy=True) delegates per-call policy checks to the gateway's eval_policy (allow/deny/ask).
ConfidentialityLabel gains a readers lattice [PRIVATE] with a readers frozenset, intersected on combine via ConfidentialityLabel.combine, and parses gateway _meta IFC labels including the top-level "$" scope.
New [workiq-email-example.py] sample ([--cli] using ToolApprovalMiddleware, plus README and tests.


python/packages/core/agent_framework/security.py
```diff
@@ -106,22 +106,156 @@ def __str__(self) -> str:
...
+        PRIVATE: Singleton for private content with no specific readers.
     """
 
-    PUBLIC = "public"
-    PRIVATE = "private"
-    USER_IDENTITY = "user_identity"
+    PUBLIC: ConfidentialityLabel  # set after class body
+    PRIVATE: ConfidentialityLabel  # set after class body
+
+    def __init__(self, value: str | list[str] = "public", *, readers: frozenset[str] | None = None) -> None:
```
eavanvalkenburg: The literal will tell users that they made a mistake (misspelling or a non-list name) and the Sequence is to allow covariant typing.

```suggestion
    def __init__(self, value: Literal['public' | 'private'] | Sequence[str] = "public", *, readers: frozenset[str] | None = None) -> None:
```

python/packages/core/agent_framework/security.py
```diff
@@ -106,22 +106,156 @@ def __str__(self) -> str:
...
+            value: Either ``"public"``, ``"private"``, or a list of reader
+                identities. A list implies ``"private"`` restricted to those readers.
+            readers: Optional frozenset of reader identities. Only applied when
+                ``value`` is ``"private"``; ignored for ``"public"``.
+
+        Raises:
+            ValueError: If ``value`` is not ``"public"``, ``"private"``, or a list.
+        """
+        if isinstance(value, list):
+            self._level = "private"
```
eavanvalkenburg:
```suggestion
            self._level: Literal['public' | 'private']  = "private"
```

python/samples/02-agents/security/workiq-email-example.py
```diff
@@ -0,0 +1,262 @@
...
+            approval_on_violation=True,
+            quarantine_chat_client=None,
+        )
+
+        agent = await stack.enter_async_context(
+            Agent(
+                client=main_client,
+                name="WorkIQSecureAgent",
+                instructions=AGENT_INSTRUCTIONS,
+                tools=all_tools,
```
eavanvalkenburg: it is not necessary to unpack the tools list for both, you can just pass the two MCP servers in:

```suggestion
                tools=[secure_mcp_1, secure_mcp_2],
```

python/samples/02-agents/security/workiq-email-example.py
```diff
@@ -0,0 +1,262 @@
...
+        level=logging.WARNING,
+        format="%(asctime)s [%(levelname)s] %(name)s: %(message)s",
+        force=True,
+    )
+    logging.getLogger("agent_framework.security").setLevel(logging.DEBUG if debug else logging.WARNING)
+    try:
+        if mode == "cli":
+            asyncio.run(run_cli_async(debug=debug, gateway_port=gateway_port, auto_approve=auto_approve))
+        else:
+            asyncio.run(run_devui_async(debug=debug, gateway_port=gateway_port))
```
eavanvalkenburg: again, this complicates showing the exact setup for this feature, and adds a lot of overhead on the because of the two interfaces, remember that samples are not for dev purposes, it is to show how something works, less is more!

python/samples/02-agents/security/workiq-email-example.py
```diff
@@ -0,0 +1,262 @@
...
+            approvals = []
+            for req in result.user_input_requests:
+                function_call = req.function_call
+                tool_name = function_call.name if function_call is not None else "<unknown>"
+                if auto_approve:
+                    approved = True
+                    print(f"[auto-approve] gateway policy flagged '{tool_name}' -> approving")
+                else:
+                    prompt = f"Approve tool '{tool_name}' flagged by gateway policy? [y/N] "
+                    answer = (await asyncio.to_thread(input, prompt)).strip().lower()
```
eavanvalkenburg: this is also not needed for a sample, just use `answer = input(prompt)`

python/samples/02-agents/security/workiq-email-example.py
```diff
@@ -0,0 +1,262 @@
...
+FOUNDRY_MODEL = os.getenv("FOUNDRY_MODEL", "gpt-5")
+
+AGENT_INSTRUCTIONS = (
+    "Act as my high-autonomy productivity agent: read and respond to my emails/teams messages, "
+    "infer intent, and complete tasks using available WorkIQ MCP server tools without asking "
+    "unnecessary questions. Reply if you have to, do not ask me for confirmation."
+)
+
+
+async def run_cli_async(*, debug: bool = False, gateway_port: int = 9090, auto_approve: bool = False) -> None:
```
eavanvalkenburg: wiring the auto-approval should a rule in the ToolApprovalMiddleware, that makes the other code much simpler as well

python/packages/core/agent_framework/security.py
```diff
@@ -106,22 +106,156 @@ def __str__(self) -> str:
 
 
 @experimental(feature_id=ExperimentalFeature.FIDES)
-class ConfidentialityLabel(str, Enum):
+class ConfidentialityLabel:
```
eavanvalkenburg, line 109: I'm never a big fan of this kind of enum-looking but not quite it features... and I'm thinking if we maybe could do `Literal['public' | 'private'] | Sequence[str]` as the value for a confidentiality param, with helpers functions for combining and comparison, that might simplify things a bit and it saves a additonial class concept that users then do not need to know

python/packages/core/agent_framework/security.py
```diff
@@ -106,22 +106,158 @@ def __str__(self) -> str:
...
+
+    def is_at_most(self, other: ConfidentialityLabel) -> bool:
+        """Return ``True`` if this label is no more restrictive than *other*.
+
+        Uses the confidentiality hierarchy (``PUBLIC`` < ``PRIVATE``): returns
+        ``True`` when this label's level is less than or equal to *other*'s
+        level. Useful for checking whether content may flow to a destination
+        that allows up to a given confidentiality level.
+        """
+        return self._priority <= other._priority
```
moonbox3, line 222: Should reader-set containment be part of this comparison? The PR adds reader-restricted private labels and combines them by intersection, but this lets `ConfidentialityLabel(["Alice"])` flow to `ConfidentialityLabel(["Bob"])` because both have priority `1`. Since `max_allowed_confidentiality` uses this path, reader-restricted data can be sent to the wrong private audience.

python/packages/core/agent_framework/security.py
```diff
@@ -3167,6 +3529,104 @@ async def _wrapped(*args: Any, **kwargs: Any) -> Any:
...
+
+    def _conf_wire(conf: ConfidentialityLabel) -> list[str]:
+        """Convert to gateway wire format: always a list of readers.
+
+        The gateway's ``IFCLabels`` model expects ``confidentiality`` as a
+        ``list[str]``.  An empty list means public (no reader restriction).
+        """
+        if conf.readers:
+            return sorted(conf.readers)
+        if conf.is_private:
```
moonbox3, line 3559: Should the gateway payload preserve `PRIVATE` separately from `PUBLIC` here? `_conf_wire()` documents `[]` as public/no reader restriction, but it also serializes `ConfidentialityLabel.PRIVATE` with no readers as `[]`. With `gateway_policy=True`, an `allow` returns before built-in confidentiality checks run, so legacy private context can be evaluated remotely as public. Could we encode the private level explicitly or fall back to local enforcement when the label cannot be represented faithfully on the gateway wire?

python/packages/core/agent_framework/security.py
```diff
@@ -3167,6 +3529,104 @@ async def _wrapped(*args: Any, **kwargs: Any) -> Any:
...
+        try:
+            result = await mcp_tool.call_tool("eval_policy", call=call_payload)
+            # MCP tool results are list[Content]; parse the structured content
+            if isinstance(result, list) and result:
+                first = cast("Any", result[0])
+                text: str = cast("str", first.text) if hasattr(first, "text") else str(first)
+                try:
+                    return cast(dict[str, Any], json.loads(text))
+                except (json.JSONDecodeError, TypeError):
+                    return {"decision": "allow", "message": str(text)}
```
moonbox3, line 3615: Should parse failures from `eval_policy` become `error` instead of `allow`? In gateway mode an allow skips `_fallback_to_builtin_policy`, so non-JSON text, a schema mismatch, or an empty/unstructured result lets the protected tool call execute even though the policy decision was not actually evaluated. Could these parse/no-response branches return `decision: "error"` or request approval so the fallback path handles them?

moonbox3:
Please re-open when ready to fix the conflict and address the remaining feedback. Thank you.

PR #6843, opened 2026-06-30
.NET: Add skill approval options

SergeyMenshykh:
### Motivation & Context

Agent skills currently require approval for every provider tool. Some hosts need to selectively disable approval for trusted skill operations.

### Description & Review Guide

- **What are the major changes?** Added per-tool options to disable approval for loading skills, reading resources, and running scripts.
- **What is the impact of these changes?** Existing behavior is unchanged by default; callers can opt out per tool.
- **What do you want reviewers to focus on?** Option naming and default behavior.

### Related Issue

Fixes #6844, https://github.com/microsoft/agent-framework/issues/6825

Related Python issue: #6845

### Contribution Checklist

- [x] The code builds clean without any errors or warnings
- [ ] All unit tests pass, and I have added new tests where possible
- [x] The PR follows the [Contribution Guidelines](https://github.com/microsoft/agent-framework/blob/main/CONTRIBUTING.md)
- [x] This PR is linked to an issue and there is no other open PR for this issue (see Related Issue above).
- [x] **This is not a breaking change.** If it _is_ a breaking change, add the `breaking change` label (or add "[BREAKING]" to the title prefix, before or after any language prefix) — a workflow keeps the label and title prefix in sync automatically.

PR #6867, opened 2026-07-01
Python: Allow disabling approval for SkillsProvider tools

### LINKED ISSUE #6845: Python: Allow disabling approval for AgentSkillsProvider tools
AgentSkillsProvider should allow disabling approval per skill tool for trusted skill scenarios while keeping approval required by default.

Related .NET issue: #6844

giles17:
### Motivation & Context

`SkillsProvider` currently registers all three of its tools (`load_skill`, `read_skill_resource`, `run_skill_script`) with `approval_mode="always_require"`, so every skill operation prompts for host approval. Some hosts run skills from a trusted source and need to selectively disable approval for specific skill operations without losing the safe-by-default behavior.

This is the Python counterpart of the .NET change in #6843 (which added per-tool approval options to `AgentSkillsProvider`).

  - Added three keyword-only arguments (all default `False`) to `SkillsProvider.__init__` and `SkillsProvider.from_paths`: `disable_load_skill_approval`, `disable_read_skill_resource_approval`, and `disable_run_skill_script_approval`.
  - When a flag is set, the corresponding tool is registered with `approval_mode="never_require"` (via a small `_approval_mode()` helper) instead of `"always_require"`, so it runs without approval.
  - Updated the class, constructor, `from_paths`, `_create_tools`, and auto-approval-rule docstrings, plus `packages/core/AGENTS.md`.
  - Added unit tests covering each flag individually, all flags together, and `from_paths` forwarding.

  Existing behavior is unchanged by default — every skill tool still requires approval. Callers can now opt individual tools out of approval for trusted skills. The `.NET`-style naming (`Disable*Approval`) is preserved via Python's `disable_*` convention.

  Argument naming and the default (approval still required) behavior.

### Related Issue

Fixes #6845

Related .NET PR: #6843

### Contribution Checklist


TaoChenOSU (approved):

python/packages/core/agent_framework/_skills.py
```diff
@@ -2145,6 +2198,9 @@ def from_paths(
             source,
             instruction_template=instruction_template,
             disable_caching=disable_caching,
+            disable_load_skill_approval=disable_load_skill_approval,
```
moonbox3: Could this break existing `SkillsProvider` subclasses that inherit `from_paths()`? Because this classmethod returns `cls`, it now passes the three new kwargs even when the caller leaves them at their defaults, so a subclass with the previous compatible `__init__(..., instruction_template, disable_caching, source_id)` shape will start raising `TypeError` on `MyProvider.from_paths(...)` without opting into this feature. Could we preserve the default path for subclasses, or otherwise avoid forwarding the new kwargs unless they are explicitly enabled?

SergeyMenshykh (approved):

python/packages/core/agent_framework/_skills.py
giles17: Thanks that's a good catch. I've gone ahead to forward them only if enabled in the latest commit

moonbox3 (approved):

PR #6892, opened 2026-07-02
.NET: Return 404 for a response created against a nonexistent conversation

rogerbarreto:
### Motivation & Context

A create-response request that references a nonexistent conversation id previously failed
mid-execution and returned HTTP 500. This returns a clean 404 instead, matching the Conversations
API.

- Validate conversation existence up front in `ValidateRequestAsync`; a missing conversation returns
  `conversation_not_found`.
- Add `ResponseErrorCodes` to map validation codes to an HTTP status (404 vs 400) in one place.
- Add unit and HTTP integration tests.

### Related Issue

N/A.

### Contribution Checklist

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/InMemoryResponsesService.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/InMemoryResponsesService.cs
@@ -156,9 +156,26 @@ internal sealed class InMemoryResponsesService : IResponsesService, IDisposable
         {
             return new ResponseError
             {
-                Code = "invalid_request",
+                Code = ResponseErrorCodes.InvalidRequest,
                 Message = "Mutually exclusive parameters: 'conversation' and 'previous_response_id'. Ensure you are only providing one of: 'previous_response_id' or 'conversation'."
             };
+        }
+
+        // When a conversation store is configured and the request references a conversation,
+        // verify it exists up front. This surfaces a missing conversation as a clean not-found
+        // error (mapped to HTTP 404, consistent with the Conversations API) instead of failing
+        // mid-execution and reporting a generic server error.
+        if (this._conversationStorage is not null && request.Conversation?.Id is { Length: > 0 } conversationId)
+        {
+            var conversation = await this._conversationStorage.GetConversationAsync(conversationId, cancellationToken).ConfigureAwait(false);
+            if (conversation is null)
+            {
+                return new ResponseError
+                {
+                    Code = ResponseErrorCodes.ConversationNotFound,
+                    Message = $"Conversation with id '{conversationId}' not found."
+                };
+            }
         }
 
         return await this._executor.ValidateRequestAsync(request, cancellationToken).ConfigureAwait(false);
--- a/dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/ResponseErrorCodes.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/ResponseErrorCodes.cs
@@ -1,0 +1,38 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.Agents.AI.Hosting.OpenAI.Responses;
+
+/// <summary>
+/// Well-known error codes returned by response request validation, together with their mapping to
+/// the HTTP status code a handler should return for each. Centralizing the codes avoids scattering
+/// string literals and keeps the status mapping in a single place.
+/// </summary>
+internal static class ResponseErrorCodes
+{
+    /// <summary>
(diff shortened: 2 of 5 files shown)


dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/InMemoryResponsesService.cs
```diff
@@ -156,11 +156,28 @@ public InMemoryResponsesService(IResponseExecutor executor, InMemoryStorageOptio
...
             };
         }
 
+        // When a conversation store is configured and the request references a conversation,
+        // verify it exists up front. This surfaces a missing conversation as a clean not-found
+        // error (mapped to HTTP 404, consistent with the Conversations API) instead of failing
+        // mid-execution and reporting a generic server error.
+        if (this._conversationStorage is not null && request.Conversation?.Id is { Length: > 0 } conversationId)
+        {
+            var conversation = await this._conversationStorage.GetConversationAsync(conversationId, cancellationToken).ConfigureAwait(false);
```
westey-m, line 170: An optimization to consider but perhaps for a separate pr:
This call is an additional call that we need to wait for to complete before continuing on to processing the request, which adds latency.  It would be nice if e.g. ListItemsAsync just threw if the conversation didn't exist and that could be turned into a ResponseError right in the main processing method.
So instead of two round trips to the store (DB?), it's one round trip and less latency.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/ResponseErrorCodes.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// <summary>
+    /// Maps a validation error code to the HTTP status code a handler should return. Not-found codes
+    /// map to <see cref="StatusCodes.Status404NotFound"/>; every other validation failure maps to
+    /// <see cref="StatusCodes.Status400BadRequest"/>.
+    /// </summary>
+    /// <param name="code">The <see cref="Models.ResponseError.Code"/> to map.</param>
+    /// <returns>The HTTP status code to return for the given error code.</returns>
+    public static int ToHttpStatusCode(string? code) => code switch
+    {
+        ConversationNotFound => StatusCodes.Status404NotFound,
```
westey-m: I'm not sure this maps exactly to the typical semantics of a rest 404.  Typically a 404 would be if the resource doesn't exist, and in this case the call is to create a new resource in the form of a response, i.e. `POST /v1/responses`, so I would expect it to 404 only if there is no `responses` resource collection.

IMO BadRequest probably matches better.

westey-m: Actually, I tried this on OpenAI.  Looks like they also return 404, so even though I think it's weird, it's probably best to match the behavior.  Funny enough the error type is invalid_request_error which semantically means bad request, but the status code is not found. I got this response:

```
FAILED with HTTP status 404.
Error body:
{
  "error": {
    "message": "Conversation with id 'conv_b038a3ed9e0649789334272d74a61ce7' not found.",
    "type": "invalid_request_error",
    "param": null,
    "code": null
  }
}
```

westey-m (approved):

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/InMemoryResponsesService.cs
rogerbarreto: Yeah, my initial assumption was that if you are using InMemoryResponses service, naturally you would have InMemoryStorage, I think worth adding the improvement in this PR. Thanks!

SergeyMenshykh (approved):

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/InMemoryResponsesService.cs
rogerbarreto: Good idea, but tested live OpenAI: stream=true with bad conversation still give HTTP 404 + JSON body BEFORE stream open, not a stream event. So OpenAI check conversation up front. Move check into ListItemsAsync path would make streaming return 200 then failed event, break that parity. So keep the pre-flight check. Single-lookup optimization better as separate PR like you say.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/Responses/ResponseErrorCodes.cs
rogerbarreto: Yes, matched it now. Body match OpenAI exact: message "Conversation with id '<id>' not found.", type invalid_request_error, param null, code null. Status stay 404 like you confirm. Test assert this for both stream and non-stream.

westey-m (approved):

PR #6895, opened 2026-07-02
Python: Add SkillsSourceContext to SkillsSource.get_skills

### LINKED ISSUE #6711: Python: Add AgentSkillsSourceContext to AgentSkillsSource.get_skills

giles17:
### Motivation & Context

`SkillsSource.get_skills` took no arguments, so skill sources and decorators had no access to the invocation context. This prevented context-aware behavior such as filtering skills per agent or isolating the skills cache per agent/tenant.

This adds a `SkillsSourceContext` (the invoking agent plus optional session) that the skills provider builds and threads through the entire source pipeline, so sources, filters, and caching can make context-aware decisions.

Python port of the .NET change in #6797.

  - New frozen, experimental `SkillsSourceContext(agent, session)`.
  - `SkillsSource.get_skills` and all sources/decorators (`File`, `InMemory`, `Delegating`, `Deduplicating`, `Filtering`, `Caching`, `Aggregating`, `MCP`) now accept and forward a `SkillsSourceContext`.
  - `FilteringSkillsSource` predicate is now context-aware: `Callable[[Skill, SkillsSourceContext], bool]`.
  - `CachingSkillsSource` gains an optional `cache_isolation_key_selector` for per-key cache isolation; a `None` selector (the default) keeps the existing shared-cache behavior.
  - `SkillsProvider` builds the context from `before_run`'s `agent`/`session` and passes it into the pipeline.
  - `SkillsSourceContext` is exported from the public API; `foundry_hosting`'s toolbox skill source, tests, and docs are updated accordingly.

  - Enables per-agent/session skill selection and per-key cache isolation.
  - Signature changes touch the `@experimental` skills API only. Default behavior is unchanged: context-ignoring leaf sources and the default (unkeyed) cache behave exactly as before.

  - The `CachingSkillsSource` rewrite from a single shared slot to per-key buckets with per-key locks (double-checked locking, retry-on-failure preserved), and that the default path stays equivalent to the previous single-bucket behavior.

### Related Issue

Fixes #6711

### Contribution Checklist


python/packages/core/agent_framework/_skills.py
```diff
@@ -2225,10 +2228,14 @@ async def _create_context(self) -> tuple[Sequence[Skill], str | None, list[Funct
...
 
+        Args:
+            source_context: Contextual information about the agent and session
+                requesting skills, forwarded to the source pipeline.
+
         Returns:
             A tuple of ``(skills, instructions, tools)``.
         """
-        skills = await self._source.get_skills()
+        skills = await self._source.get_skills(source_context)
```
moonbox3, line 2323: Could this leak skills across agents or tenants when a context-aware source is used with the provider’s default caching? `SkillsProvider` wraps the whole source in a shared `CachingSkillsSource`, so the first invocation’s filtered result is reused for later `SkillsSourceContext`s before predicates or key selectors can run again. In a multi-tenant host, one tenant or agent can prime private/admin skills into the advertised tools and another tenant then receives those skill names and handlers. Could the provider either expose a cache isolation selector or avoid the unkeyed outer cache when the source pipeline is context-aware?

python/packages/core/agent_framework/_skills.py
```diff
@@ -2675,13 +2711,17 @@ def __init__(
...
+    async def get_skills(self, context: SkillsSourceContext) -> list[Skill]:
         """Discover and return all file-based skills from configured paths.
 
         Scans directories for ``SKILL.md`` files, parses their frontmatter,
         discovers resource and script files, and returns populated
         :class:`Skill` instances.
 
+        Args:
+            context: Contextual information about the agent and session
+                requesting skills. Unused by this source.
```
eavanvalkenburg: what does `Unused by this source` mean here? seems like the param is used at least

giles17: Good catch — reworded in d5df2d5cb. The param is part of the  get_skills  contract; this source just doesn't use it to change which skills are returned. Updated the docstring to say that instead of the misleading "Unused by this source."

python/packages/core/agent_framework/_skills.py
giles17: Good catch — fixed in d5df2d5cb. The provider no longer auto-wraps a caller-supplied  SkillsSource  in the default cache. Default caching now applies only to the built-in, context-independent file/in-memory leaf sources ( Deduplicating(Caching(leaf)) ), matching the .NET provider. A custom pipeline is used as-is, so a context-aware source is re-evaluated per  SkillsSourceContext  — no cross-agent/tenant replay. Callers who want caching compose  CachingSkillsSource  (optionally with  cache_isolation_key_selector ) themselves. Added a leak-prevention test.

moonbox3 (approved):

eavanvalkenburg (approved):

PR #6906, opened 2026-07-03
.NET: [BREAKING] Align ShellPolicy allow/deny semantics with Python

### LINKED ISSUE #6899: ShellTool: Add deny all by default option
The shell tool allows configuring what commands should be allowed and which blocked.
We should have the ability to block everything except a selected set of allowed commands.

westey-m:
### Motivation & Context

The `.NET` `ShellPolicy` (`Microsoft.Agents.AI.Tools.Shell`) and the Python
`ShellPolicy` (`agent_framework_tools.shell`) share the same intent — a layered
allow/deny UX pre-filter for shell commands — but had **incompatible evaluation
semantics**. In particular, the .NET version had **no way to express "deny
everything except an explicitly allowed set"**, which is exactly what
issue #6899 asks for. Its allow list behaved as *exceptions* that
short-circuited the deny list rather than as an exclusive allow list.

This change aligns the .NET behavior with the Python implementation so operators
get a consistent policy model across languages, including the requested
deny-all-except-allowed mode.

  - **Deny-first evaluation order** (matching Python): `empty → deny patterns →
    allow list → custom callback → default allow`. A deny match now wins over an
    allow match (previously allow short-circuited deny).
  - **Allow list is now exclusive**: when supplied, any command matching *none*
    of the allow patterns is denied — this provides the "deny everything except
    the allowed set" behavior requested in #6899.
  - **`null` vs empty allow list distinction**: a `null` allow list disables the
    allow rule entirely; a supplied-but-empty allow list denies every command
    (matching Python).
  - **New `custom` callback**: an optional
    `Func<ShellRequest, ShellPolicyOutcome?>` that gets the final say and may
    override the default allow.
  - Updated XML docs and added/rewrote unit tests covering deny precedence,
    exclusive allow list (match/non-match), empty/null allow list, and the
    custom callback.

  - **Breaking** for existing preview users of `ShellPolicy`: the allow-list
    meaning and evaluation precedence change. The feature is experimental, but
    flagged breaking so preview users are aware. Released features are unchanged.
  - .NET-idiomatic naming/shape is retained (no renames of `ShellPolicyOutcome`,
    `Allowed`, `WorkingDirectory`, etc.).

  - The evaluation order and the `null` vs empty allow-list semantics in
    `ShellPolicy.Evaluate`.

### Related Issue

Fixes #6899

### Contribution Checklist

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellPolicy.cs
+++ b/dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellPolicy.cs
@@ -118,24 +118,29 @@ public readonly struct ShellPolicyOutcome : IEquatable<ShellPolicyOutcome>
 /// </para>
 /// <para>
 /// <b>No default patterns.</b> A <see cref="ShellPolicy"/> constructed
-/// with no arguments has an empty deny list and an empty allow list —
-/// it will allow any non-empty command. Operators who want pre-execution
-/// rejection of specific shapes must supply their own
-/// <paramref>denyList</paramref>.
+/// with no arguments has an empty deny list and no allow list (allow list
+/// disabled) — it will allow any non-empty command. Operators who want
+/// pre-execution rejection of specific shapes must supply their own
+/// <paramref>denyList</paramref>, or an <paramref>allowList</paramref> to
+/// deny everything except the explicitly allowed commands.
 /// </para>
 /// <para>
-/// <b>Evaluation order — allow short-circuits deny.</b> Allow patterns are
-/// checked first; a match returns immediately without consulting the deny
-/// list. Use allow patterns sparingly (and prefer narrowly anchored regexes
-/// like <c>^git\s+status$</c> rather than substring matches), because an
-/// over-broad allow pattern can re-enable a command that the deny list was
-/// supposed to block.
+/// <b>Evaluation order — deny-first, the allow list is exclusive.</b> Deny
+/// patterns are checked first and a match wins immediately. If an allow list
+/// is supplied, it is treated as exclusive: any command that matches
+/// <em>none</em> of the allow patterns is denied. Supplying an empty allow
+/// list therefore denies every command; leaving the allow list
+/// <see langword="null"/> disables the allow list entirely. An optional
+/// <c>custom</c> callback gets the final say and may override the outcome.
+/// Prefer narrowly anchored regexes (like <c>^git\s+status$</c>) over
+/// substring matches when building an allow list.
 /// </para>
 /// </remarks>
 public sealed class ShellPolicy
 {
     private readonly IReadOnlyList<Regex> _denies;
-    private readonly IReadOnlyList<Regex> _allows;
+    private readonly IReadOnlyList<Regex>? _allows;
+    private readonly Func<ShellRequest, ShellPolicyOutcome?>? _custom;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="ShellPolicy"/> class.
@@ -145,10 +150,21 @@ public sealed class ShellPolicy
     /// empty collection disables the deny list entirely.
     /// </param>
     /// <param name="allowList">
-    /// Optional explicit-allow patterns. A match here short-circuits the
-    /// deny list and is useful when the caller knows the command is safe.
+    /// Optional allow-list patterns. When <see langword="null"/> the allow
+    /// list is disabled. When supplied (including as an empty collection) any
+    /// command matching none of the patterns is denied — an empty collection
+    /// therefore denies every command.
     /// </param>
-    public ShellPolicy(IEnumerable<string>? denyList = null, IEnumerable<string>? allowList = null)
+    /// <param name="custom">
+    /// Optional callback that gets the final say. It runs after the deny and
+    /// allow lists have passed; returning a non-<see langword="null"/> outcome
+    /// overrides the default allow, while <see langword="null"/> leaves the
+    /// default in place.
+    /// </param>
+    public ShellPolicy(
+        IEnumerable<string>? denyList = null,
+        IEnumerable<string>? allowList = null,
+        Func<ShellRequest, ShellPolicyOutcome?>? custom = null)
     {
         var deny = new List<Regex>();
         if (denyList is not null)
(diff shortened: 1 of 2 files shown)


dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellPolicy.cs
```diff
@@ -118,24 +118,29 @@ public bool Equals(ShellPolicyOutcome other) =>
...
-/// supposed to block.
+/// <b>Evaluation order — deny-first, the allow list is exclusive.</b> Deny
+/// patterns are checked first and a match wins immediately. If an allow list
+/// is supplied, it is treated as exclusive: any command that matches
+/// <em>none</em> of the allow patterns is denied. Supplying an empty allow
+/// list therefore denies every command; leaving the allow list
+/// <see langword="null"/> disables the allow list entirely. An optional
+/// <c>custom</c> callback gets the final say and may override the outcome.
+/// Prefer narrowly anchored regexes (like <c>^git\s+status$</c>) over
+/// substring matches when building an allow list.
```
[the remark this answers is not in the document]
westey-m: Good catch — reworded the remarks so the `custom` callback is described as running last (after the deny and allow lists pass) and overriding only the default allow, explicitly noting it cannot re-enable a command already rejected by the deny or allow list.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellPolicy.cs
```diff
@@ -189,19 +207,38 @@ public ShellPolicyOutcome Evaluate(ShellRequest request)
...
+            {
+                return ShellPolicyOutcome.Deny("command does not match allow list");
+            }
+        }
+
+        if (this._custom is not null)
+        {
+            var overrideOutcome = this._custom(request);
+            if (overrideOutcome is { } outcome)
+            {
```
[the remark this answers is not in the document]
westey-m: Intentionally leaving this as a bare allow. This PR aligns .NET with the Python implementation, and Python returns a reasonless `allow` decision on every allow path (including allow-list matches). Adding a .NET-only reason here would re-introduce a cross-language divergence, so declining to keep parity.

dotnet/tests/Microsoft.Agents.AI.Tools.Shell.UnitTests/LocalShellExecutorTests.cs
```diff
@@ -37,13 +37,78 @@ public void Policy_DenyList_BlocksDestructiveRm()
...
+        var policy = new ShellPolicy(custom: _ => null);
+        Assert.True(policy.Evaluate(new ShellRequest("echo hello")).Allowed);
+    }
+
+    [Fact]
+    public void Policy_Custom_RunsAfterDenyAndAllowLists()
+    {
+        // Deny-list match short-circuits before the custom callback runs, so a
+        // permissive custom callback cannot re-enable a denied command.
+        var policy = new ShellPolicy(
```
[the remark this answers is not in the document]
westey-m: Fixed. Renamed to `Policy_Custom_DoesNotRunWhenDenyListMatches` (which is what it actually asserts) and added `Policy_Custom_DoesNotOverrideAllowListDenial` to cover allow-list ordering: allow list set, non-matching command, permissive custom callback → still denied.

SergeyMenshykh (approved):

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellPolicy.cs
```diff
@@ -118,24 +118,31 @@ public bool Equals(ShellPolicyOutcome other) =>
...
+/// a deny); it cannot re-enable a command already rejected by the deny list
+/// or the allow list. Prefer narrowly anchored regexes (like
+/// <c>^git\s+status$</c>) over substring matches when building an allow list.
 /// </para>
 /// </remarks>
 public sealed class ShellPolicy
 {
     private readonly IReadOnlyList<Regex> _denies;
-    private readonly IReadOnlyList<Regex> _allows;
+    private readonly IReadOnlyList<Regex>? _allows;
```
rogerbarreto: nit. I would keep the consistencty with the parameters provided in the policy.

```suggestion
    private readonly IReadOnlyList<Regex>? _allowList;
```

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellPolicy.cs
```diff
@@ -160,24 +178,26 @@ public ShellPolicy(IEnumerable<string>? denyList = null, IEnumerable<string>? al
...
 
-        var allow = new List<Regex>();
         if (allowList is not null)
         {
+            var allow = new List<Regex>();
             foreach (var pattern in allowList)
             {
                 allow.Add(new Regex(pattern, RegexOptions.Compiled | RegexOptions.IgnoreCase));
             }
+            this._allows = allow;
```
rogerbarreto: nit
```suggestion
            this._allows = allowList.Select((p) => new Regex(...));
```

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellPolicy.cs
```diff
@@ -189,19 +209,38 @@ public ShellPolicyOutcome Evaluate(ShellRequest request)
...
+            var matched = false;
+            foreach (var allow in this._allows)
             {
-                return ShellPolicyOutcome.Deny($"matched deny pattern: {deny}");
+                if (allow.IsMatch(command))
+                {
+                    matched = true;
+                    break;
+                }
+            }
```
rogerbarreto: nit

```suggestion
            var matched = this._allows.Any(IsMatch(command));
```

dotnet/tests/Microsoft.Agents.AI.Tools.Shell.UnitTests/LocalShellExecutorTests.cs
```diff
@@ -37,13 +37,90 @@ public void Policy_DenyList_BlocksDestructiveRm()
     }
 
     [Fact]
-    public void Policy_AllowList_OverridesDeny()
+    public void Policy_DenyList_TakesPrecedenceOverAllowList()
```
rogerbarreto, line 40: I would also add a precedence for a greater pattern for denial.

So a more broad deny pattern also override a specific allow pattern.

rogerbarreto (approved):

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellPolicy.cs
westey-m: Done — renamed the fields to match the constructor parameters: `_denies`→`_denyList` and `_allows`→`_allowList` (renamed both so the pair stays consistent).

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellPolicy.cs
westey-m: Done — both lists are now compiled via LINQ: `allowList?.Select(pattern => new Regex(pattern, RegexOptions.Compiled | RegexOptions.IgnoreCase)).ToArray()` (and the same for the deny list), preserving the compiled + case-insensitive options.

dotnet/src/Microsoft.Agents.AI.Tools.Shell/ShellPolicy.cs
westey-m: Done — replaced the foreach/flag loop with `var matched = this._allowList.Any(allow => allow.IsMatch(command));`.

dotnet/tests/Microsoft.Agents.AI.Tools.Shell.UnitTests/LocalShellExecutorTests.cs
westey-m: Added `Policy_BroadDenyList_OverridesSpecificAllowList`: a broad deny pattern (`git push`) overrides a more specific allow pattern (`^git push origin main$`) for the command `git push origin main` → denied.

rogerbarreto (approved):

PR #6907, opened 2026-07-03
Python: Fix Foundry reasoning MCP compaction

### LINKED ISSUE #6883: Python: [Bug]: Orphaned mcp_call items cause HTTP 400 in multi-turn reasoning model flows
### Description

Problem highlighted in below issue has not been resolved for customer. 
https://github.com/microsoft/agent-framework/issues/6074

### Code Sample

```markdown
Attached the code sample used by client. 
[run_maf_workflow.txt](https://github.com/user-attachments/files/29597439/run_maf_workflow.txt)
[run_workflow.txt](https://github.com/user-attachments/files/29597438/run_workflow.txt)
```

```markdown
HTTP 400 — Item 'mcp_<id>' of type 'mcp_call' was provided without its
required 'reasoning' item: 'rs_<id>'.
```

### Package Versions

agent-framework==1.8.0

### Python Version

3.13

### Additional Context

Issue is INTERMITTENT — does not fire on every run. We need to try 5-6 times to reproduce.

godququ5-code:
## Summary
- Preserve Foundry reasoning output as `text_reasoning` with the provider reasoning id when reconstructing response history.
- Treat `mcp_server_tool_call` as a tool-call assistant item during compaction so reasoning + MCP call pairs stay atomic.
- Add focused regressions for both Foundry reasoning reconstruction paths and sliding-window MCP compaction.

Fixes #6883

## Root cause
Foundry response reconstruction converted reasoning summaries into ordinary text content, dropping the reasoning id. Core compaction also only recognized `function_call` assistant messages as tool-call items, so an MCP call following a reasoning item could be grouped separately and survive after its reasoning prefix was excluded.

## Test plan
- `uv run pytest packages/core/tests/core/test_compaction.py packages/foundry_hosting/tests/test_responses.py -q`
- `uv run prek run --files packages/core/agent_framework/_compaction.py packages/core/tests/core/test_compaction.py packages/foundry_hosting/agent_framework_foundry_hosting/_responses.py packages/foundry_hosting/tests/test_responses.py`

Note: the pytest command exits 0, but the local environment emits unrelated OpenTelemetry exporter shutdown noise against `127.0.0.1:7897` after the tests finish.


godququ5-code:

python/packages/core/agent_framework/_compaction.py
```diff
@@ -74,16 +74,16 @@ def _has_content_type(message: Message, content_type: str) -> bool:
     return any(content.type == content_type for content in message.contents)
 
 
-def _has_function_call(message: Message) -> bool:
-    return _has_content_type(message, "function_call")
+def _has_tool_call(message: Message) -> bool:
```
moonbox3, line 86: Now that MCP-call messages classify as tool_call groups, ToolResultCompactionStrategy will start collapsing older MCP groups too, but its summary builder only reads function_call and function_result contents (around lines 874-888 in this file), so an MCP group collapses to the literal `[Tool results: no results]` while the real output gets excluded. Should the summary builder also handle mcp_server_tool_call / mcp_server_tool_result so we don't lose the tool output that strategy exists to preserve?

python/packages/core/agent_framework/_compaction.py
```diff
@@ -74,16 +74,16 @@ def _has_content_type(message: Message, content_type: str) -> bool:
...
+    return any(content.type in {"function_call", "mcp_server_tool_call"} for content in message.contents)
 
 
 def _has_reasoning(message: Message) -> bool:
     return _has_content_type(message, "text_reasoning")
 
 
 def _is_tool_call_assistant(message: Message) -> bool:
-    return message.role == "assistant" and _has_function_call(message)
+    return message.role == "assistant" and _has_tool_call(message)
```
moonbox3, line 95: Have we thought about the other hosted tool-call content types here? Assistant messages carrying code_interpreter_tool_call, shell_tool_call, or image_generation_tool_call still classify as plain assistant text, so a reasoning item followed by one of those calls splits into separate groups and can reproduce the same "provided without its required 'reasoning' item" 400 this PR fixes for MCP.

python/packages/foundry_hosting/agent_framework_foundry_hosting/_responses.py
```diff
@@ -1110,7 +1110,7 @@ async def _item_to_message(item: Item, *, approval_storage: ApprovalStorage | No
         reason_contents: list[Content] = []
         if reasoning.summary:
             for summary in reasoning.summary:
-                reason_contents.append(Content.from_text(summary.text))
+                reason_contents.append(Content.from_text_reasoning(id=reasoning.id, text=summary.text))
```
moonbox3, line 1119: What happens when the reasoning item has no summary, which is the default unless summaries are explicitly requested? reason_contents stays empty, the reasoning id is dropped entirely, and the following mcp_call is still reconstructed, so replay can still produce the orphaned mcp_call 400 from #6883. The outbound path has the same gap: _content_to_events only emits a reasoning item when content.text is not None (around line 1757), so an id-only text_reasoning marker never round-trips at all.

moonbox3: Also, could this break hosted agents backed by non-Responses clients? The reconstructed text_reasoning carries no protected_data, so the Anthropic client turns it into a thinking block without the required signature (packages/anthropic _chat_client.py around line 919), and the Chat Completions client falls past its `text_reasoning if protected_data is not None` case into the default content path, which serializes `content.to_dict()` as an unknown text_reasoning content part (_chat_completion_client.py around lines 909 and 1022). Both are request 400s where the previous plain text content was valid everywhere.

python/packages/foundry_hosting/agent_framework_foundry_hosting/_responses.py
```diff
@@ -1383,7 +1383,7 @@ async def _output_item_to_message(item: OutputItem, *, approval_storage: Approva
         contents: list[Content] = []
         if reasoning.summary:
             for summary in reasoning.summary:
-                contents.append(Content.from_text(summary.text))
+                contents.append(Content.from_text_reasoning(id=reasoning.id, text=summary.text))
```
moonbox3, line 1410: With reconstructed history now containing text_reasoning, any replay without service-side storage trips drops_reasoning_without_storage in the OpenAI Responses client (packages/openai _chat_client.py around line 1480), which collects every mcp_server_tool_call call_id across the entire history and strips all MCP calls and results from the request, not just the pair adjacent to the dropped reasoning. Is silently losing all prior MCP context the intended trade-off for this flow, or should that drop be scoped to calls actually paired with a dropped reasoning item?

moonbox3 (approved):

eavanvalkenburg (approved):

PR #6925, opened 2026-07-06
Python: [BREAKING]: Canonicalize AG-UI interrupt and resume handling

moonbox3:
### Motivation & Context

AG-UI agent and workflow interrupts need a single protocol-level pause/resume contract that matches the canonical AG-UI `Interrupt` and `ResumeEntry` models. The current integration exposes an undesired interrupt shape, and several resume/cancel paths can leave approval, workflow, or snapshot state stale.

  - Emits interrupted runs as `RUN_FINISHED.outcome.type == "interrupt"` with canonical `outcome.interrupts`.
  - Handles tool approvals, confirmation prompts, and workflow `request_info` pauses through canonical AG-UI interrupt/resume data.
  - Serializes typed and compatible legacy client resume inputs into canonical `ResumeEntry` arrays.
  - Validates missing, unknown, invalid, resolved, and cancelled resume entries through streamed `RUN_ERROR` events.
  - Keeps snapshot hydration aligned with canonical interrupt outcomes and clears only the completed/cancelled interrupt state.
  - Updates AG-UI docs, examples, and tests, and raises the `ag-ui-protocol` dependency to the version that contains the canonical models.

  - This is a release-candidate breaking wire-contract change for interrupted runs: clients should read `RUN_FINISHED.outcome.interrupts` instead of the legacy top-level interrupt field.
  - Resume requests should use canonical `resume` entries with `interruptId`, `status`, and optional `payload`.
  - Cancelled approvals and workflow input requests no longer leave stale pending state that wedges later requests.

  - Whether the canonical interrupt/resume wire shape is the right public AG-UI contract.
  - Approval and workflow resume validation, especially cancellation and multi-interrupt cases.
  - Snapshot-backed hydrate/resume behavior for interrupted threads.

### Contribution Checklist


python/packages/ag-ui/README.md
```diff
@@ -132,6 +132,87 @@ async def get_weather(city: str) -> Content:
...
+        "reason": "tool_call",
+        "message": "Approve tool call get_weather?",
+        "toolCallId": "tool_call_1",
+        "responseSchema": {
+          "type": "object",
+          "properties": {
+            "approved": { "type": "boolean" },
+            "arguments": { "type": "object" }
+          },
+          "required": ["approved"]
```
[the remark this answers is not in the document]
moonbox3: Addressed

python/packages/ag-ui/README.md
```diff
@@ -132,6 +132,87 @@ async def get_weather(city: str) -> Content:
...
+        },
+        "metadata": {
+          "agentFramework": {
+            "kind": "function_approval"
+          }
+        }
```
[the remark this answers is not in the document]
moonbox3: Addressed

python/packages/ag-ui/agent_framework_ag_ui/_workflow_run.py
```diff
@@ -412,6 +537,48 @@ def _coerce_responses_for_pending_requests(
...
+def _coerce_responses_for_pending_requests_strict(
+    responses: dict[str, Any],
+    pending_events: dict[str, Any],
+) -> tuple[dict[str, Any], RunErrorEvent | None]:
+    """Coerce resume responses or return RUN_ERROR for invalid pending response payloads."""
+    if not responses or not pending_events:
+        return responses, None
+
+    normalized: dict[str, Any] = {}
+    pending_by_id = {str(request_id): event for request_id, event in pending_events.items()}
```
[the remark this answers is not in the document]
moonbox3: Addressed

python/packages/ag-ui/agent_framework_ag_ui/_run_common.py
```diff
@@ -108,12 +128,313 @@ def _extract_resume_payload(input_data: dict[str, Any]) -> Any:
...
+    entries: list[dict[str, Any]] = []
+    for item in candidates:
+        model_dump = getattr(item, "model_dump", None)
+        if callable(model_dump):
+            item = model_dump(by_alias=True, exclude_none=True)
+        if not isinstance(item, dict):
+            return [], "Each resume entry must be an object."
+
+        item_dict = cast(dict[str, Any], item)
+        interrupt_id = item_dict.get("interruptId") or item_dict.get("interrupt_id") or item_dict.get("id")
```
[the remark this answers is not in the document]
moonbox3: Addressed

python/packages/ag-ui/agent_framework_ag_ui/_agent_run.py
```diff
@@ -437,7 +468,339 @@ def _pending_approval_arguments(entry: PendingApprovalEntry) -> str | None:
     return entry["arguments"]
 
 
-def _evict_oldest_approvals(registry: dict[str, PendingApprovalEntry], max_size: int = 10_000) -> None:
+def _pending_approval_request_id(entry: PendingApprovalEntry) -> str | None:
```
eavanvalkenburg: I'm never a big fan of these kinds of helper functions that are only used once...

python/packages/ag-ui/agent_framework_ag_ui/_agent_run.py
```diff
@@ -437,7 +468,339 @@ def _pending_approval_arguments(entry: PendingApprovalEntry) -> str | None:
...
+    except json.JSONDecodeError:
+        return None
+    return cast(dict[str, Any], parsed) if isinstance(parsed, dict) else None
+
+
+def _thread_has_pending_approvals(
+    pending_approvals: dict[PendingApprovalKey, PendingApprovalEntry] | None,
+    thread_id: str,
+) -> bool:
+    if pending_approvals is None:
```
eavanvalkenburg: might as well cover the empty dict case early:
```suggestion
    if not pending_approvals:
```

python/packages/ag-ui/agent_framework_ag_ui/_agent_run.py
```diff
@@ -437,7 +468,339 @@ def _pending_approval_arguments(entry: PendingApprovalEntry) -> str | None:
...
+        return False
+    return any(key[0] == thread_id for key in pending_approvals)
+
+
+def _pending_approval_interrupt_ids(
+    pending_approvals: dict[PendingApprovalKey, PendingApprovalEntry] | None,
+    thread_id: str,
+) -> set[str]:
+    if pending_approvals is None:
```
eavanvalkenburg: same here

python/packages/ag-ui/agent_framework_ag_ui/_client.py
```diff
@@ -430,17 +430,18 @@ async def _streaming_impl(
 
         converter = AGUIEventConverter()
 
+        available_interrupts = options.get("available_interrupts")
```
eavanvalkenburg:
```suggestion
        available_interrupts = options.get("available_interrupts", options.get("availableInterrupts"))
```

python/packages/ag-ui/agent_framework_ag_ui/_event_converters.py
```diff
@@ -185,6 +185,13 @@ def _handle_run_finished(self, event: dict[str, Any]) -> ChatResponseUpdate:
         }
         if "interrupt" in event:
             additional_properties["interrupt"] = event.get("interrupt")
+        if "outcome" in event:
+            outcome = event.get("outcome")
+            additional_properties["outcome"] = outcome
+            if isinstance(outcome, dict) and outcome.get("type") == "interrupt":
```
eavanvalkenburg: is there ever a case where we would accept outcome not to be a dict, and what should happen? wonderign if this should just fail loudly instead of silently ignoring

TaoChenOSU (approved):

python/packages/ag-ui/agent_framework_ag_ui/_run_common.py
```diff
@@ -108,12 +128,318 @@ def _extract_resume_payload(input_data: dict[str, Any]) -> Any:
...
-        return RunFinishedEvent(run_id=run_id, thread_id=thread_id, interrupt=interrupts)  # type: ignore[call-arg]
+        canonical_interrupts = [
+            canonical for interrupt in interrupts if (canonical := _canonical_interrupt(interrupt)) is not None
+        ]
+        if canonical_interrupts:
+            return RunFinishedEvent(
+                run_id=run_id,
+                thread_id=thread_id,
+                outcome=RunFinishedInterruptOutcome(interrupts=canonical_interrupts),
+            )
```
contextablemark, line 442:
```suggestion
            )
        logger.warning(
            "run %s: %d interrupt(s) present but none carried an id/interruptId; "
            "emitting RUN_FINISHED with no interrupt outcome",
            run_id, len(interrupts),
        )
```

Minor / defensive: if `interrupts` is non-empty but every entry filters out (i.e., `_canonical_interrupt `returns `None` for all of them, which happens only when none carry an `id` / `interruptId` (line 289–291) ) control falls through to the bare `return` and emits a RUN_FINISHED with no outcome, byte-identical to a normal completion and with no log. A genuinely-paused run would look complete to the client.

In practice this is unreachable from framework-produced interrupts (tool-approval and workflow request-info always attach an id, and a real schema violation makes `Interrupt.model_validate` raise rather than return `None`), so it's only reachable via externally-injected / hand-seeded snapshot data. So this is purely defensive and not worth throwing an error over, but probably worth a log since this is going to turn into a head-scratcher otherwise if it ever does happen.

python/packages/ag-ui/agent_framework_ag_ui/_types.py
```diff
@@ -83,14 +86,14 @@ class AGUIRequest(BaseModel):
...
         validation_alias=AliasChoices("availableInterrupts", "available_interrupts"),
-        description="List of interrupts that can be resumed by the server",
+        description="Canonical AG-UI interrupts that can be resumed by the server",
     )
-    resume: dict[str, Any] | None = Field(
+    resume: Any | None = Field(
         None,
-        description="Resume payload containing interrupt responses",
+        description="Resume payload for continuing interrupted runs",
     )
```
contextablemark: `resume: Any` is the only untyped field on an otherwise fully-canonical request surface : `available_interrupts` right above it is `list[Interrupt]`, but `resume` advertises no shape in the generated schema, so a client generating from the OpenAPI/JSON schema is told "send anything" even though the canonical contract is `list[ResumeEntry]`. Runtime is fine (the normalizers accept both canonical and legacy); this is purely the advertised contract, and it's out of step with the sibling field so seems off.

I'd recommend typing it and folding the legacy shapes in via a before-validator (this suggestion covers the annotation; it also needs `BeforeValidator` added to the `pydantic` import and the small helper below, but `ResumeEntry` is already imported):

```suggestion
    resume: Annotated[list[ResumeEntry], BeforeValidator(_coerce_legacy_resume)] | None = Field(
        None,
        description="Resume payload for continuing interrupted runs",
    )
```

```python
def _coerce_legacy_resume(v):
    if v is None:
        return v
    if isinstance(v, dict) and "interrupts" in v:  # unwrap legacy container
        v = v["interrupts"]
    if not isinstance(v, list):
        return v  # let pydantic reject non-lists
    out = []
    for e in v:
        if not isinstance(e, dict):
            out.append(e)
            continue
        iid = e.get("interruptId") or e.get("interrupt_id") or e.get("id") or e.get("toolCallId")
        payload = e.get("payload", e.get("value", e.get("response")))
        entry = {"interruptId": iid, "status": e.get("status", "resolved")}
        if payload is not None:
            entry["payload"] = payload
        out.append(entry)
    return out
```

I checked this against `ag_ui.core` 0.1.19 + pydantic 2.13.4: it keeps `test_agui_request_resume_accepts_legacy_object_shapes` green (the container coerces cleanly), advertises the canonical shape in the schema, and 422s only genuinely-malformed input (id-less / bad status) that `_strict_resume_entries` already rejects downstream. One design note: it shifts that rejection from a streamed `RUN_ERROR` to a 422 on ingress which is fine if that error channel is acceptable, but I wanted to flag the surface change.

If you'd rather keep strictly zero behavior change, a schema-only override works — but `WithJsonSchema` must be given a **dereferenced** schema: passing `TypeAdapter(list[ResumeEntry]).json_schema()` directly raises `KeyError: '#/$defs/ResumeEntry'` at schema-gen because the `$ref` isn't hoisted into the model's root `$defs`. Inline the `ResumeEntry` definition into `items` so it's self-contained, then it generates fine (though the advertised schema will then be stricter than the `Any` runtime actually enforces - maybe a bit more breakage than you'd want).

contextablemark (commented):
See attached comments.

python/packages/ag-ui/agent_framework_ag_ui/_agent_run.py
moonbox3: Addressed in 2569b276c. I inlined the single-use pending approval request/interrupt id helper lookups into `_pending_approval_alias_keys` and removed the helpers.

python/packages/ag-ui/agent_framework_ag_ui/_agent_run.py
moonbox3: Addressed in 2569b276c. `_thread_has_pending_approvals` now uses `if not pending_approvals:` so `None` and `{}` both return early.

python/packages/ag-ui/agent_framework_ag_ui/_agent_run.py
moonbox3: Addressed in 2569b276c. `_pending_approval_interrupt_ids` now also uses `if not pending_approvals:` to cover the empty dict case early.

python/packages/ag-ui/agent_framework_ag_ui/_client.py
moonbox3: Addressed in 2569b276c. The fallback lookup is now `options.get("available_interrupts", options.get("availableInterrupts"))`.

python/packages/ag-ui/agent_framework_ag_ui/_event_converters.py
moonbox3: Addressed in 2569b276c. I kept the converter tolerant: it preserves the raw non-dict `outcome` for remote protocol compatibility, but now logs a warning instead of silently ignoring the malformed interrupt shape.

python/packages/ag-ui/agent_framework_ag_ui/_run_common.py
moonbox3: Addressed in 2569b276c. `_build_run_finished_event` now logs a warning when non-empty interrupt input all canonicalizes away, and `test_build_run_finished_event_logs_when_interrupts_all_drop` covers it.

python/packages/ag-ui/agent_framework_ag_ui/_types.py
moonbox3: Addressed in 2569b276c. `AGUIRequest.resume` is now `Annotated[list[ResumeEntry], BeforeValidator(_coerce_legacy_resume)] | None`, with legacy container and single-entry mapping coercion preserved. Malformed resume shapes now fail request validation as discussed, and tests cover schema, legacy coercion, and rejection.

contextablemark (approved):
Looks good

giles17 (approved):

PR #6951, opened 2026-07-07
.NET: Add support for file content messages in AG-UI 

### LINKED ISSUE #6931: .NET: [Feature]: Send file attachments and text together in one AG-UI message
### Description

According to this thread: https://github.com/ag-ui-protocol/ag-ui/issues/280
The AG-UI protocol has been updated to include file and text uploads. I am building a chat assistant using .NET AG-UI as a server with a React frontend (Assistant-UI). I was building a functionality where users can include attachments in their message. I have deployed a file-capable model but the only restriction is the .NET AG-UI package that does not allow sending files, whilst the main protocol does support it.

### Language/SDK

Both

ElderJames:
### Motivation & Context

This commit is required to correctly support AG-UI message payloads that contain mixed structured content, not just plain-string content.

It solves the limitation where AG-UI messages were treated as single text payloads, which prevented reliable round-trip handling for multi-part content such as text + media/data blocks.

The change enables richer message flows and serialization compatibility for scenarios involving document/image/audio/video blocks, URI/data payloads, hosted file references, and mixed content in both directions (AGUIMessage ↔ ChatMessage).

This change introduces first-class AG-UI content-block modeling and dual-direction serialization support for mixed message contents.

  - Added content-block data model and JSON converter:
    - New `AGUIMessageContent`, `AGUIMessageContentBlock`, and `AGUIMessageContentSource` in `AGUIMessageContent.cs`.
    - New `AGUIMessageContentJsonConverter` to serialize/deserialize `content` as either:
      - single string (legacy text path), or
      - array of content blocks (new structured path).
  - Updated `AGUIMessage.Content` type from `string` to `AGUIMessageContent` (`AGUIMessage.cs`).
  - Extended `AGUIJsonSerializerContext` registrations to include new content-block types for source-generation serializers.
  - Enhanced `AGUIChatMessageExtensions` with:
    - conversion from `AGUIMessage` to `ChatMessage` for block-based content,
    - conversion from `ChatMessage` to `AGUIMessage` producing content blocks,
    - explicit handling for `TextContent`, `UriContent`, `DataContent`, `HostedFileContent`, tool results/calls, and assistant message edge cases,
    - support for mapping media types and source modes (`url`, `data`, hosted file id).
  - Added/updated tests:
    - `AGUIChatMessageExtensionsTests`
    - `AGUIJsonSerializerContextTests`
    - coverage for mixed/blocked content conversion and JSON round-trip validation.

  - Enables structured AG-UI payloads for mixed media/text content across request/response boundaries.
  - Improves interoperability and correctness of AG-UI serialization/deserialization in .NET.
  - Keeps behavior compatible when a message is pure text by preserving text-only optimization paths.
  - Expands test coverage for serialization and conversion paths, reducing regressions around content handling.

  - `AGUIMessage.Content` type migration impact: verify all call paths in AGUI conversion flow expect `AGUIMessageContent` and not raw strings.
  - Converter behavior in `AGUIMessageContentJsonConverter` (`string` vs block array parsing) for backward compatibility.
  - Mapping rules in `AGUIChatMessageExtensions` for:
    - assistant messages with tool calls + additional text content,
    - mixed content (`TextContent` + `UriContent`/`DataContent`),
    - media block type resolution (`image/audio/video/document/file/binary`).
  - Test expectations in both conversion and serializer-context tests to ensure both directions remain stable.

### Related Issue

Fixes #6931

### Contribution Checklist


ElderJames:

Kumima:
AG-UI hosting structure is under big change, see:https://github.com/microsoft/agent-framework/pull/6653

rogerbarreto (commented):
@ElderJames Thanks for the contribution! The direction is solid: AG-UI multimodal content maps cleanly onto the MEAI content types, and the bidirectional approach in `AGUIChatMessageExtensions` is the right place for it. A few things to align with the AG-UI spec and the existing Python adapter before this is ready.

### Spec reference
Using the canonical `ag-ui-protocol` schema (`sdks/python/ag_ui/core/types.py`):
- Only `UserMessage.content` is `Union[str, List[InputContent]]`. `developer`, `system`, `assistant`, `tool`, `reasoning` content is a plain string.
- Canonical media parts (`image`/`audio`/`video`/`document`) carry only `type`, `source`, and optional `metadata`. The flat `url`/`data`/`id`/`filename`/`mimeType` fields exist only on the deprecated `binary` part.
- `source` is `{type:"data", value:<base64>, mimeType}` (mimeType required) or `{type:"url", value, mimeType?}`.

### Blocking

1. **`DataContent` emits base64 into the block `url` field.** `BuildMediaContentBlock` always sets `Url = value`, and for the `DataContent` branch `value` is the base64 string. The result is a block with `"url": "<base64...>"`, which is malformed. The correct payload is already in `source.value`, so `url` should not be set for data sources.

2. **`HostedFileContent` emits `type:"document"` with `source: null`.** Per spec `DocumentInputContent.source` is required, so this is non-conformant. An id-only reference should use the deprecated `binary` shape (which allows `id`) or be resolved to a real source. The Python adapter maps a binary id to `ag-ui://binary/{id}`; matching that keeps the two languages consistent.

### Please consider

3. **Content widened to all roles.** `AGUIMessage.Content` becomes `AGUIMessageContent` for every role, but only user messages may carry block arrays. Text-only messages collapse back to a string so the common path is fine, but the outbound path can still emit arrays for non-user roles. Suggest constraining block-array emission to the user role.

4. **Canonical vs deprecated shape mixing.** Emitting `type:"image"`/`"document"` with extra top-level `url`/`mimeType`/`filename` blends the canonical and deprecated-`binary` shapes. The Python SDK tolerates it (`extra="allow"`), but stricter SDKs (TypeScript zod, Rust) may reject it. Prefer source-only media parts.

5. **Align with the Python reference adapter.** `python/packages/ag-ui/.../_message_adapters.py` already handles this end to end (source-type aliases `uri`/`base64`, `mime_type` vs `mimeType`, `data:` URI passthrough, binary-id handling). Reusing its semantics avoids cross-language drift.

### Docs
ADR `0010-ag-ui-support.md` states the original scope was core text streaming with no custom content types. This PR expands the content model, so a short ADR update noting the multimodal content-block addition would be good.

Round-trip tests for `DataContent` and `HostedFileContent` (not just the url/image case) would catch issues 1 and 2.

### Nice to have
A small runnable end-to-end .NET sample showing a multimodal message (for example an image or document upload flowing through to the agent) would help users adopt this. Not a blocker, and note the Python package does not have one either, so this is a suggestion rather than a parity requirement.

ElderJames:
> AG-UI hosting structure is under big change, see:#6653

So, does this Pull Request still need to be pushed forward? @Kumima  @rogerbarreto

rogerbarreto:
@ElderJames, I don't think so, as well pointed by @Kumima, I think the official package now is the source of support for AGUI, feel free to close this PR. Thanks!

We will move forward bringing the change proposed in the PR below into MAF.

- https://github.com/microsoft/agent-framework/pull/6653

ElderJames:
Ok, thanks.

PR #6966, opened 2026-07-07
Python: bind policy-enforcement approvals to a single tool invocation

rogerbarreto:
### Motivation & Context

Hardens the FIDES `PolicyEnforcementFunctionMiddleware` approval flow (`approval_on_violation=True`). A granted approval was tracked only by `call_id` in a set that was never cleared, so a reused `call_id` could satisfy a later or different policy-violating call, and an approved response was accepted without checking its own `id` / embedded `function_call`.

- **Major changes:** an approval now binds to the exact invocation shown for review — `call_id` + function name + arguments + security label + session — the response `id` and embedded `function_call` are validated, and the approval is consumed on first use. Any reused `call_id`, different function, changed arguments, escalated label, different session, or mismatched response requires a fresh approval.
- **Impact:** Python `agent-framework-core` only; behaviour-preserving for legitimate approve→execute flows. No .NET change (the concept is Python-only; the hosted .NET approval path reconstructs calls from server-recorded state and is not affected).
- **Focus:** `PolicyEnforcementFunctionMiddleware` binding/consume-once logic and the added regression tests.

### Related Issue

N/A

### Contribution Checklist


python/packages/core/agent_framework/security.py
```diff
@@ -1678,64 +1693,165 @@ def __init__(
...
+        embedded = getattr(approval_response, "function_call", None)
+        if self._signature_from_function_call(embedded) != body_signature:
+            return False
+        # The response id (and the embedded call id, when present) must name the pending request.
+        response_id = getattr(approval_response, "id", None)
+        embedded_call_id = getattr(embedded, "call_id", None)
+        return not (
+            (response_id is not None and response_id != call_id)
+            or (embedded_call_id is not None and embedded_call_id != call_id)
+        )
```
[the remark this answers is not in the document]
rogerbarreto: Good catch. Now both response id and embedded call_id must exist AND equal pending call_id. None value no slip through no more, get rejected. Test added for the id-less response case.

python/packages/core/agent_framework/security.py
```diff
@@ -1847,7 +1969,8 @@ async def process(
 
                 self._log_violation(violation)
 
-                if self._is_policy_violation_approved(context):
+                if approved:
+                    self._consume_pending_approval(context)
```
moonbox3: Should this re-check approval after the integrity branch consumes it? Because `approved` is computed once before both checks, a call in an UNTRUSTED + PRIVATE context with `max_allowed_confidentiality="public"` gets an approval request only for `untrusted_context`, then the replay reaches this confidentiality branch with stale `approved=True` and executes despite the undisclosed data-exfiltration violation. 

Could we either surface all detected violations in the approval request or consume/recompute per violation so the confidentiality boundary needs explicit approval too?

TaoChenOSU (approved):

python/packages/core/agent_framework/security.py
rogerbarreto: Good catch. Fixed in 4521a5f. Root cause: `approved` computed once, stale True leaked into confidentiality branch, so a second undisclosed violation got waved on replay. Now both checks run up front, all violations collected, and one approval request discloses every one (multi-violation adds a `violations[]` list). Approval waves only what it disclosed; call_id+function+args+label+session binding and consume-once unchanged. Regression test `test_multiple_violations_disclosed_in_single_approval` covers UNTRUSTED+PRIVATE with max_allowed_confidentiality=public: red before, green now.

python/packages/core/agent_framework/security.py
```diff
@@ -1944,124 +1964,102 @@ async def process(
...
+            logger.debug(f"Policy check passed for tool '{function_name}'")
+            await call_next()
+            return
+
+        for violation in violations:
+            self._log_violation(violation["audit"])
+
+        disclosed = ", ".join(v["violation_type"] for v in violations)
+
+        if approved:
```
moonbox3, line 2065: I missed this last time. Could this also bind the approved violation set, not just the call body, label, and session? A pending approval created for one disclosed violation still matches after the tool's policy metadata changes, so replay can compute a new `max_allowed_confidentiality` violation here and execute it under the old approval even though the user never approved that data-exfiltration risk. Should `_PendingPolicyApproval` store the canonical disclosed violations and reject or re-request approval if replay computes a different set?

rogerbarreto: Yes, valid. Fixed in f9616c1. `_PendingPolicyApproval` now records the canonical disclosed violation fingerprint (violation type + reason), and replay must trip the same set or it re-requests approval disclosing the new set. Binding the reason (not just the type) also closes the same-type-worse-risk case: if the tool's `max_allowed_confidentiality` destination loosens between request and replay, the type stays `max_allowed_confidentiality` but the reason changes (PRIVATE->PUBLIC destination), so the old approval no longer matches. Ran a deeper audit while here and also tightened the approval gate to require `approved is True` (a truthy non-bool no longer counts). Tests added: new-violation-on-replay, same-type-worse-risk, and non-boolean approved flag. Considered but left out: requiring a policy-domain marker on the response object (the client round-trip does not guarantee `additional_properties` survives, so enforcing it risks the happy path); pending-map growth from never-answered requests is unchanged from before and orthogonal to the replay binding.

moonbox3 (approved):

Issue #6301, opened 2026-06-03
Agent Skills Release


semenshi:
- [x] [POC: MCP resource-template skills type integration](https://github.com/microsoft/agent-framework/issues/6358)

.NET: Items

- [x] https://github.com/microsoft/agent-framework/issues/6683
- [x] https://github.com/microsoft/agent-framework/issues/6304
- [x] https://github.com/microsoft/agent-framework/issues/6077
- [x] https://github.com/microsoft/agent-framework/issues/6727
- [x] https://github.com/microsoft/agent-framework/issues/6707
- [x] https://github.com/microsoft/agent-framework/issues/6765
- [x] https://github.com/microsoft/agent-framework/issues/6760
- [x] https://github.com/microsoft/agent-framework/issues/6710
- [x] [.NET: Support MCP skills of skill-md type](https://github.com/microsoft/agent-framework/issues/6076)
- [x] [[Feature]: InlineSkills. Allow to add scripts that are not added to the body of the skill](https://github.com/microsoft/agent-framework/issues/5405)
- [x] https://github.com/microsoft/agent-framework/issues/6844
- [x] https://github.com/microsoft/agent-framework/issues/6830
- [x] https://github.com/microsoft/agent-framework/issues/6833
- [x] https://github.com/microsoft/agent-framework/issues/6835
- [x] Remove the "experimental" attribute from the skills API, except for MCP ones
- [x] Update documentation
- [x] Announce via blog post
- [ ] ~https://github.com/microsoft/agent-framework/issues/6117~

Python: Items
- [x] https://github.com/microsoft/agent-framework/issues/6348
- [x] https://github.com/microsoft/agent-framework/issues/6728
- [x] https://github.com/microsoft/agent-framework/issues/6681
- [x] https://github.com/microsoft/agent-framework/issues/6845
- [x] https://github.com/microsoft/agent-framework/issues/6682
- [x] https://github.com/microsoft/agent-framework/issues/5672
- [x] https://github.com/microsoft/agent-framework/issues/6543
- [x] https://github.com/microsoft/agent-framework/issues/5282
- [x] https://github.com/microsoft/agent-framework/issues/6767
- [x] https://github.com/microsoft/agent-framework/issues/6711
- [x] https://github.com/microsoft/agent-framework/issues/6834
- [x] https://github.com/microsoft/agent-framework/issues/6836
- [x] https://github.com/microsoft/agent-framework/issues/6089
[...]

PR #6977, opened 2026-07-07
Python: Add refresh_interval (TTL) to CachingSkillsSource

giles17:
### Motivation & Context

The Agent Skills feature is being brought to parity between .NET and Python (epic #6301). One gap on the Python side is cache invalidation.

`CachingSkillsSource` wraps an expensive skill source (filesystem discovery or a remote `MCPSkillsSource`) and caches the discovered skill list so discovery does not run on every agent turn. In .NET, `CachingAgentSkillsSource` supports a `CachingAgentSkillsSourceOptions.RefreshInterval` that expires a cached list after a configured duration and re-fetches on the next request. Python had no equivalent: once populated, the cache was held for the lifetime of the process and was only cleared by a failed fetch. That meant callers had no built-in way to periodically re-discover skills for a source whose contents change at runtime — most importantly an `MCPSkillsSource` whose remote server updates its advertised skills. This PR closes that gap.

  - `CachingSkillsSource` gains an optional `refresh_interval: timedelta | None` (default `None`). A cached list older than the interval is treated as stale and re-fetched on the next call; `None` keeps the previous behavior (never expires). A zero/negative interval makes every result immediately stale (effectively disabling caching), and a failed refresh leaves the prior list in place and retries — mirroring the .NET semantics.
  - Freshness is measured with a monotonic clock via a small module-level `_monotonic()` helper (immune to wall-clock changes and monkeypatchable so tests are deterministic without real sleeps).
  - `SkillsProvider.__init__` and `SkillsProvider.from_paths` expose a `cache_refresh_interval` kwarg that is threaded into the built-in `CachingSkillsSource` (no effect on a caller-supplied source or when `disable_caching=True`). `from_paths` forwards the kwarg only when set, so subclasses with the previous `__init__` signature keep working.
  - New tests and updated `packages/core/AGENTS.md` guidance.

  - Purely additive and backward compatible: the new parameters default to `None`, so existing call sites and behavior are unchanged. This ports .NET's `RefreshInterval` to Python.

  - The freshness/expiry logic in `_get_fresh_cached` and its interaction with the existing per-key in-flight-fetch lock, and the choice of a monotonic clock.

### Related Issue

Contributes to the Agent Skills parity epic #6301. (No dedicated issue; intentionally not using a closing keyword so the epic is not auto-closed.)

### Contribution Checklist


moonbox3 (approved):

python/packages/core/agent_framework/_skills.py
```diff
@@ -3709,6 +3736,17 @@ async def get_skills(self, context: SkillsSourceContext) -> list[Skill]:
         return [s for s in skills if self._predicate(s, context)]
 
 
+def _monotonic() -> float:
```
eavanvalkenburg: I don't really like this, a function that is just there to make testing easier, we should just be able to monkeypatch time.monotonic.

eavanvalkenburg (approved):

TaoChenOSU (approved):

TaoChenOSU (approved):

PR #7000, opened 2026-07-08
.NET: [BREAKING] Hosting OpenAI Responses protocol helpers and optional execution state

rogerbarreto:
### Motivation & Context

[ADR-0027](https://github.com/microsoft/agent-framework/blob/main/docs/decisions/0027-hosting-channels.md) refocused the hosting design toward **protocol conversion helpers plus optional execution state**: Agent Framework owns protocol-native <-> run conversion, while the application owns HTTP routing, authentication, middleware, and storage. The Python slice landed in #6891.

This PR realizes that direction for .NET. A first-principles gap analysis showed .NET already covers most of the capability, and more richly, but bundled behind the route-owning `MapOpenAIResponses` server. The only genuine gap is the ownership model: an application cannot own its own route and call just the conversion primitives. So this change un-bundles the existing conversion rather than reinventing it. No new package, no OpenAI-SDK-typed reimplementation, and `MapOpenAIResponses` public behavior is unchanged.

  - New public facade `OpenAIResponses` in `Microsoft.Agents.AI.Hosting.OpenAI` (JsonElement/SSE boundary; wire DTOs stay internal), delegating to the existing internal converters:
    - `ToAgentRunRequest(JsonElement)` -> messages + `AgentRunOptions?`
    - `WriteResponse(AgentResponse, responseId, sessionId?)` -> Responses JSON
    - `WriteResponseStreamAsync(IAsyncEnumerable<AgentResponseUpdate>, responseId, ...)` -> Responses SSE frames
    - `GetSessionId(JsonElement)` -> `previous_response_id`/`conversation` id (kept separate so the trust boundary stays visible)
    - `CreateResponseId()` -> `resp_*`
  - Protocol-neutral execution state in `Microsoft.Agents.AI.Hosting`:
    - `AgentSessionStore.DeleteSessionAsync` (added as an abstract method so every store makes a conscious choice: implement deletion or throw `NotSupportedException` itself; the in-box stores implement it). Hosting is still preview, so this tightens the contract now rather than shipping a silent default.
    - No agent-side holder: applications use `AgentSessionStore` directly (`GetSessionAsync` creates on miss, `SaveSessionAsync` persists post-run including under a newly minted id, `DeleteSessionAsync` removes). ADR-0032 records why per-run instance and async target setup are left to the DI container in .NET.
    - `HostedWorkflowState` + `HostedWorkflowRunResult` (workflow instance or factory + `CheckpointManager` + an internal `sessionId -> CheckpointInfo` head cursor; `RunOrResumeAsync`/`RunOrResumeStreamingAsync`)
  - Two samples under `dotnet/samples/04-hosting/af-hosting/`: `local_responses` (app-owned route, sync + SSE) and `local_responses_workflow` (checkpoint resume).
  - ADR `0032-dotnet-hosting-protocol-helpers.md` and spec `003-dotnet-hosting-protocol-helpers.md`.
  - Breaking (preview package): adds a new abstract member (`AgentSessionStore.DeleteSessionAsync`) to a public abstract type and renames the store id parameter to `sessionStoreId`, so external subclasses must implement the new member. The net-new helpers are otherwise additive. `MapOpenAIResponses`, `IResponsesService`, ChatCompletions, and Conversations are untouched.
  - Applications can now expose an agent or workflow over OpenAI Responses while owning their own routing, auth, and storage.
  - The public helper shape and the JsonElement boundary decision (recorded in ADR-0032, alongside why not the OpenAI SDK Responses types).
  - The decision to reuse the richer existing `AgentSessionStore` instead of cloning Python's in-memory `SessionStore`, and the abstract (implement-or-throw) `DeleteSessionAsync`.
  - The `HostedWorkflowState` head-cursor approach for per-session checkpoint resume.

### Related Issue

N/A. This is the .NET realization of the accepted [ADR-0027](https://github.com/microsoft/agent-framework/blob/main/docs/decisions/0027-hosting-channels.md); it adds ADR-0032 and its spec. It supersedes the earlier channel-model draft in #6151 (closed). Opening as a draft for design feedback.

### Contribution Checklist

- [x] **This is a breaking change** (preview package): a new abstract member is added to `AgentSessionStore` and the store id parameter is renamed to `sessionStoreId`. The `breaking change` label is applied.


dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/OpenAIResponses.cs
```diff
@@ -0,0 +1,136 @@
...
+        AgentRunOptions? options = (mapOptions ?? new OpenAIResponsesMapOptions()).RunOptionsFactory(request.ToRequestInfo());
+
+        var messages = new List<ChatMessage>();
+        foreach (InputMessage inputMessage in request.Input.GetInputMessages())
+        {
+            messages.Add(inputMessage.ToChatMessage());
+        }
+
+        return new OpenAIResponsesRunRequest(messages, options);
+    }
```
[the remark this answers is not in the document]
rogerbarreto: good catch. bad json now caught and thrown as ArgumentException like the doc say. missing input guarded too so no more NRE. done in 2fb52b3e5.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/OpenAIResponses.cs
```diff
@@ -0,0 +1,136 @@
...
+    /// <remarks>
+    /// This is kept separate from <see cref="ToAgentRunRequest(JsonElement, OpenAIResponsesMapOptions?)"/> so the
+    /// trust boundary stays visible: using a request-derived key is an explicit application decision. The returned
+    /// value is an <strong>untrusted candidate key</strong> until the application has authorized it for the caller.
+    /// </remarks>
+    public static string? GetSessionId(JsonElement body)
+    {
+        CreateResponse? request = body.Deserialize(OpenAIHostingJsonContext.Default.CreateResponse);
+        return request?.PreviousResponseId ?? request?.Conversation?.Id;
+    }
```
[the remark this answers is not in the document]
rogerbarreto: yes. junk body now give back null instead of throwing. fixed in 2fb52b3e5.

docs/specs/003-dotnet-hosting-protocol-helpers.md
```diff
@@ -0,0 +1,219 @@
...
+var state = new HostedWorkflowState(workflow); // in-memory checkpoints + cursor
+
+app.MapPost("/responses", async (HttpContext http, CancellationToken ct) =>
+{
+    using var doc = await JsonDocument.ParseAsync(http.Request.Body, cancellationToken: ct);
+    JsonElement body = doc.RootElement;
+
+    string sessionId = Authorize(http.User, OpenAIResponses.GetSessionId(body))
+        ?? OpenAIResponses.CreateResponseId();
+
```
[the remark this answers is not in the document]
rogerbarreto: spec fixed. workflow need stable key. conversation id good, previous_response_id no. GetSessionId prefer previous_response_id so workflow read conversation straight. 2fb52b3e5.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/OpenAIResponses.cs
```diff
@@ -0,0 +1,136 @@
...
+    /// <param name="mapOptions">
+    /// Optional options controlling how request settings are mapped onto the run. By default no request
+    /// setting is mapped onto the run.
+    /// </param>
+    /// <returns>The parsed messages and mapped run options.</returns>
+    /// <exception cref="ArgumentException">The body could not be parsed as an OpenAI Responses request.</exception>
+    /// <exception cref="NotSupportedException">A request setting is not supported by the configured mapping.</exception>
+    public static OpenAIResponsesRunRequest ToAgentRunRequest(JsonElement body, OpenAIResponsesMapOptions? mapOptions = null)
+    {
+        CreateResponse request = body.Deserialize(OpenAIHostingJsonContext.Default.CreateResponse)
```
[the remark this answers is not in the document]
rogerbarreto: right, two are mutually exclusive per spec. GetSessionId now document it pick previous_response_id first when both present, and warn it not a stable key. fixed in 2fb52b3e5.

dotnet/samples/04-hosting/HostingResponsesWorkflow/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+Call the endpoint:
+
+```bash
+curl -s http://localhost:5000/responses -H "content-type: application/json" \
+  -d '{ "input": "Draft a short product announcement." }'
+```
+
+The response JSON includes the recorded checkpoint id in its summary text. Send a follow-up request with
+the same `previous_response_id` (or `conversation`) to resume from that checkpoint.
```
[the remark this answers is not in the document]
rogerbarreto: readme fixed. now say send same conversation id to resume. previous_response_id no good for that.

dotnet/samples/04-hosting/HostingResponsesWorkflow/Program.cs
```diff
@@ -0,0 +1,67 @@
...
+
+// The application owns this route.
+app.MapPost("/responses", async (HttpContext http, CancellationToken cancellationToken) =>
+{
+    using JsonDocument doc = await JsonDocument.ParseAsync(http.Request.Body, cancellationToken: cancellationToken).ConfigureAwait(false);
+    JsonElement body = doc.RootElement;
+
+    // The candidate continuation id is untrusted. A real app authenticates the caller and authorizes/binds
+    // this key before using it as the workflow's checkpoint session id. This sample falls back to a fresh id.
+    string sessionId = OpenAIResponses.GetSessionId(body) ?? OpenAIResponses.CreateResponseId();
```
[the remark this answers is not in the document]
rogerbarreto: true. previous_response_id change every turn so bad checkpoint key. sample now key on conversation id which stay same across turns. fixed in 2fb52b3e5.

dotnet/src/Microsoft.Agents.AI.Workflows/CheckpointManager.cs
```diff
@@ -58,4 +59,27 @@ ValueTask<Checkpoint> ICheckpointManager.LookupCheckpointAsync(string sessionId,
...
+    /// </summary>
+    /// <param name="sessionId">The session identifier whose latest checkpoint should be retrieved.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests.</param>
+    /// <returns>
+    /// The latest <see cref="CheckpointInfo"/> for <paramref name="sessionId"/>, or <see langword="null"/> when no
+    /// checkpoint has been committed for that session.
+    /// </returns>
+    public async ValueTask<CheckpointInfo?> GetLatestCheckpointAsync(string sessionId, CancellationToken cancellationToken = default)
+    {
+        IEnumerable<CheckpointInfo> index = await this._impl.RetrieveIndexAsync(sessionId, withParent: null).ConfigureAwait(false);
```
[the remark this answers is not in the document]
rogerbarreto: good catch. before, ICheckpointStore say nothing about order, so a custom store that give back newest-first or unsorted quietly break latest lookup. now RetrieveIndexAsync doc make commit order a hard contract: oldest first, newest last. both in-box store already do this (SessionCheckpointCache list, FileSystem OrderedCheckpointIndex list), so this just write down the invariant they follow. pick this over adding new store method to keep interface small and not break existing custom stores. GetLatestCheckpointAsync now lean on the documented order. fixed in bca52a2a5.

dotnet/samples/04-hosting/af-hosting/README.md
```diff
@@ -0,0 +1,38 @@
+# Agent Framework hosting helper samples
+
+End-to-end samples for exposing Agent Framework targets through app-owned hosting routes.
+
+The helper-first hosting packages provide protocol conversion and optional execution state. The
+application still owns the web framework, native SDK clients, authentication, response construction, and
+deployment shape.
```
westey-m: This intro feels a bit jargony.  E.g. what does 'targets', 'app-owned' and 'helper-first' mean.

dotnet/samples/04-hosting/af-hosting/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+| Sample | What it shows |
+|---|---|
+| [`local_responses/`](./local_responses) | One agent behind an app-owned ASP.NET Core route using the `OpenAIResponses` helper functions plus `HostedAgentState` / `AgentSessionStore`. Start here to learn the helper seam. |
```
westey-m: I'm not sure what this means: `Start here to learn the helper seam.`  Consider rewording.

dotnet/samples/04-hosting/af-hosting/README.md
```diff
@@ -0,0 +1,38 @@
...
+| [`local_responses_workflow/`](./local_responses_workflow) | A workflow target behind an app-owned ASP.NET Core route using the `OpenAIResponses` helper functions, `HostedWorkflowState`, an explicit `CheckpointManager`, and an app-owned checkpoint cursor. |
+
+Each sample is a **client/server pair**. Unlike the Python samples (which run a server `app.py` and calling
+scripts as loose files), .NET samples are projects, so each pair is split into two projects:
```
westey-m: Comparing with python is not interesting here, since users are here for the .net sample.

dotnet/samples/04-hosting/af-hosting/local_responses_workflow/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Server
+
+The server owns routing, authentication, and checkpoint storage. It uses the `OpenAIResponses` conversion
+helpers for the wire protocol and `HostedWorkflowState` for per-session checkpoint resume. The workflow is a
+brief adapter, a slogan-writer agent, and a formatter that renders one slogan line. The first turn runs the
+workflow forward; later turns restore the latest checkpoint and run forward with the new brief. It binds to
+`http://localhost:5001`.
+
+Like the Python sample, the server supports **`previous_response_id` continuation only** and **rejects
```
westey-m: Comparing with python samples is unnecessary here

dotnet/samples/04-hosting/af-hosting/local_responses_workflow/Server/Program.cs
```diff
@@ -0,0 +1,196 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+// This sample shows how an application can own its own ASP.NET Core route and expose a workflow over the
+// OpenAI Responses protocol. It uses the OpenAIResponses conversion helpers for the wire protocol and
+// HostedWorkflowState for per-session checkpoint resume. The application keeps control of routing, auth,
+// and checkpoint storage.
+//
+// Mirroring the Python local_responses_workflow sample, this server demonstrates previous_response_id
```
westey-m: More unnecessary comparison with the python sample

dotnet/samples/04-hosting/af-hosting/README.md
westey-m: Does app-owned mean not via MapOpenAIResponses?  If so, we should briefly explain and contrast the different options here so users understand what the options are, and preferably point them to MapOpenAIResponses samples.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/OpenAIResponses.cs
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    /// <summary>
+    /// Converts a final <see cref="AgentResponse"/> into an OpenAI Responses-shaped payload.
+    /// </summary>
+    /// <param name="response">The agent response to render.</param>
+    /// <param name="responseId">The id to assign to the rendered response (see <see cref="CreateResponseId"/>).</param>
+    /// <param name="sessionId">
+    /// The optional continuation/session id to surface as the response's conversation id.
+    /// </param>
```
westey-m: Why not just call this conversationId, if it should contain the responses conversation id?
There is no session id in AF or in OpenAI responses, so calling it that here, may be confusing.
Same for streaming.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/OpenAIResponses.cs
```diff
@@ -0,0 +1,165 @@
...
+    /// <c>previous_response_id</c> changes each turn and is therefore not a stable partition key; prefer the
+    /// <c>conversation</c> id when a stable key is required (for example a workflow checkpoint cursor key).
+    /// </para>
+    /// </remarks>
+    public static string? GetSessionId(JsonElement body)
+    {
+        CreateResponse? request;
+        try
+        {
+            request = body.Deserialize(OpenAIHostingJsonContext.Default.CreateResponse);
```
westey-m: Currently the code in the samples, parses the input twice.  First in GetSessionId and secondly in ToAgentRunRequest.  This is not the most efficient.
We should be able to flip the call ordering around and always do ToAgentRunRequest first, followed by GetSessionId, and then we can change GetSessionId to take OpenAIResponsesRunRequest as input rather than a JsonElement.
This is both more efficient, and removes the need for returning a null when the json isn't parse-able.  Instead, we know the reason for a null is always that there is no response id or conversation id.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/OpenAIResponses.cs
```diff
@@ -0,0 +1,165 @@
...
+    /// trust boundary stays visible: using a request-derived key is an explicit application decision. The returned
+    /// value is an <strong>untrusted candidate key</strong> until the application has authorized it for the caller.
+    /// <para>
+    /// The Responses protocol treats <c>previous_response_id</c> and <c>conversation</c> as mutually exclusive; if a
+    /// payload sets both, this helper prefers <c>previous_response_id</c> (the response-chain pointer). Note that
+    /// <c>previous_response_id</c> changes each turn and is therefore not a stable partition key; prefer the
+    /// <c>conversation</c> id when a stable key is required (for example a workflow checkpoint cursor key).
+    /// </para>
+    /// </remarks>
+    public static string? GetSessionId(JsonElement body)
```
westey-m: Also consider GetSessionStorageId, since it is the id under which the session is stored, rather than a more general identifier for a session.

dotnet/src/Microsoft.Agents.AI.Hosting/Local/InMemoryAgentSessionStore.cs
```diff
@@ -63,5 +63,12 @@ public override async ValueTask<AgentSession> GetSessionAsync(AIAgent agent, str
         };
     }
 
+    /// <inheritdoc/>
+    public override ValueTask DeleteSessionAsync(AIAgent agent, string conversationId, CancellationToken cancellationToken = default)
```
westey-m: We have some inconsistent naming for the id under which the session is stored.
Perhaps consider renaming all to sessionStorageId.

dotnet/src/Microsoft.Agents.AI.Hosting/AgentSessionStore.cs
```diff
@@ -76,6 +76,24 @@ public abstract ValueTask<AgentSession> GetSessionAsync(
...
+    /// <remarks>
+    /// The default implementation throws <see cref="NotSupportedException"/>. Stores that support removal
+    /// override this method. Deleting a missing session is a no-op for stores that override it.
+    /// </remarks>
+    /// <exception cref="NotSupportedException">The store does not support deletion.</exception>
+    public virtual ValueTask DeleteSessionAsync(
+        AIAgent agent,
+        string conversationId,
+        CancellationToken cancellationToken = default)
+        => throw new NotSupportedException($"{this.GetType().Name} does not support session deletion.");
```
westey-m: Since hosting is still preview, this may be a good time to make this abstract and require implementations to support or throw themselves if they really cannot implement.

dotnet/src/Microsoft.Agents.AI.Hosting/HostedAgentState.cs
```diff
@@ -0,0 +1,140 @@
...
+    public async ValueTask<IAsyncDisposable> LockSessionAsync(string sessionId, CancellationToken cancellationToken = default)
+    {
+        _ = Throw.IfNullOrEmpty(sessionId);
+
+        if (this._sessionLocks is null)
+        {
+            return NoopReleaser.Instance;
+        }
+
+        SemaphoreSlim gate = this._sessionLocks.GetOrAdd(sessionId, static _ => new SemaphoreSlim(1, 1));
```
westey-m: This is a potential memory leak, since nothing removes sessions from the _sessionLocks dictionary.
Should we be removing sessionId in the releaser?

dotnet/src/Microsoft.Agents.AI.Hosting/HostedAgentState.cs
```diff
@@ -0,0 +1,140 @@
...
+    /// Acquires an exclusive lock for <paramref name="sessionId"/> so concurrent requests for the same
+    /// session serialize their get-run-save cycle. Dispose the returned value to release the lock.
+    /// </summary>
+    /// <param name="sessionId">The application-selected session id.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests.</param>
+    /// <returns>An <see cref="IAsyncDisposable"/> that releases the lock when disposed.</returns>
+    /// <remarks>
+    /// When session locking is not enabled, this returns immediately with a no-op releaser.
+    /// </remarks>
+    public async ValueTask<IAsyncDisposable> LockSessionAsync(string sessionId, CancellationToken cancellationToken = default)
```
westey-m: Might be nice to show using this in the samples, since it's a really nice feature.

dotnet/src/Microsoft.Agents.AI.Hosting/HostedAgentState.cs
```diff
@@ -0,0 +1,140 @@
...
+/// </para>
+/// <para>
+/// <strong>Trust boundary.</strong> The <c>sessionId</c> values passed to these methods are
+/// application-selected partition keys. When a key originates from the wire (for example via
+/// <c>OpenAIResponses.GetSessionId(...)</c>), the application must authenticate the caller and authorize
+/// the key before using it here. For multi-user hosts, scope the underlying store per principal with
+/// <see cref="IsolationKeyScopedAgentSessionStore"/>.
+/// </para>
+/// </remarks>
+public sealed class HostedAgentState
```
westey-m: This class puts together a few things that I'm not convinced are necessary to bundle.
Specifically, there GetOrCreateSessionAsync, SaveSessionAsync and DeleteSessionAsync.  They just call through to the session store, and it doesn't seem super useful to have this class in the middle.  Users still have to have the session store and agent.  Why can't they just call the session store directly?

Separately there is also the locking functionality, which IMO doesn't really have that much to do with the storage piece.  I would probably just have a class for that (e.g. HostedAgentSessionMutex) and let users call AgentSessionStore directly for saving/loading/deleting.

dotnet/samples/04-hosting/af-hosting/README.md
rogerbarreto: removed python compare. reader here for dotnet. fixed in 6ec892773.

dotnet/samples/04-hosting/af-hosting/README.md
rogerbarreto: yes, app-owned mean not MapOpenAIResponses. added "two ways" section: MapOpenAIResponses batteries endpoint vs write own route and call OpenAIResponses helper. point reader to AgentWebChat sample that use MapOpenAIResponses. fixed in 6ec892773.

dotnet/samples/04-hosting/af-hosting/local_responses_workflow/Server/Program.cs
rogerbarreto: removed python compare from code comment as well. fixed in 6ec892773.

dotnet/samples/04-hosting/af-hosting/local_responses_workflow/README.md
rogerbarreto: removed python compare here too. fixed in 6ec892773.

dotnet/samples/04-hosting/af-hosting/README.md
rogerbarreto: good point. intro now plain words, no jargon. added section show two way to do this: MapOpenAIResponses batteries way, or write own route and call helper. fixed in 6ec892773.

dotnet/samples/04-hosting/af-hosting/README.md
rogerbarreto: reworded. no more "helper seam". now just say simplest sample to start with. fixed in 6ec892773.

dotnet/src/Microsoft.Agents.AI.Hosting/AgentSessionStore.cs
rogerbarreto: done. DeleteSessionAsync now abstract. in-box store implement it: InMemory remove real, Noop do nothing, Delegating/IsolationKeyScoped pass through. store that truly cannot delete throw NotSupportedException itself. fixed in 96a59b51d.

dotnet/src/Microsoft.Agents.AI.Hosting/Local/InMemoryAgentSessionStore.cs
rogerbarreto: agreed, id naming was inconsistent. renamed all to sessionStoreId across AgentSessionStore family + HostedAgentState. picked sessionStoreId not sessionStorageId because rest of repo use ...StoreId (vectorStoreId, VectorStoreId, AZURE_AI_MEMORY_STORE_ID), never storageId, and class is AgentSessionStore (Store). so sessionStoreId match both convention and class name. fixed in 96a59b51d.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/OpenAIResponses.cs
rogerbarreto: renamed. now GetSessionStoreId, it return the id the session store under. used Store not Storage to match sessionStoreId param we just renamed and rest of repo (vectorStoreId, AZURE_AI_MEMORY_STORE_ID) + class AgentSessionStore. fixed in 10531efce.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/OpenAIResponses.cs
rogerbarreto: nice catch, you right, sample parse body twice (GetSessionId + ToAgentRunRequest). flipped it: ToAgentRunRequest run first and now carry PreviousResponseId + ConversationId, then GetSessionStoreId(run) read them, no second parse. bad json throw ArgumentException in ToAgentRunRequest, so null now only mean no id present, no more null-on-unparseable. bonus: workflow sample drop the resp_ prefix check and just test run.ConversationId. fixed in 10531efce.

dotnet/src/Microsoft.Agents.AI.Hosting.OpenAI/OpenAIResponses.cs
rogerbarreto: good call. no session id concept in AF or responses. WriteResponse + WriteResponseStreamAsync param renamed sessionId -> conversationId, it surface the response conversation id. fixed in 10531efce.

dotnet/src/Microsoft.Agents.AI.Hosting/HostedAgentState.cs
rogerbarreto: good catch, real slow leak. dict grew one semaphore per session id forever. fixed with ref count: each lock now track how many hold or wait, and when last one release the entry get removed from dict and semaphore disposed. plain remove-in-releaser would race (new caller grab same instance while we remove, then next caller make a different semaphore for same id, two gate one session, lock broken), so ref count guard it under a lock. cancel during wait also drop the reservation. added tests: reclaim after release, entry stay while second holder wait, cancel reclaim. fixed in db98c7e25.

dotnet/src/Microsoft.Agents.AI.Hosting/HostedAgentState.cs
rogerbarreto: done. agent sample now turn on enableSessionLocking and wrap the get-run-save in LockSessionAsync(sessionStoreId) so two turn on same session serialize instead of clobber stored state. streaming path hold it for whole SSE, on purpose. workflow sample not need it, HostedWorkflowState already serialize with its own lock. fixed in db98c7e25.

dotnet/src/Microsoft.Agents.AI.Hosting/HostedAgentState.cs
rogerbarreto: revisited this after internalizing the locking. two parts:

locking: no longer public. it moved inside GetOrCreateSessionAsync, automatic and on by default, developer never manage it. so no separate HostedAgentSessionMutex needed, there is nothing public left to extract.

bundling: GetOrCreateSessionAsync is not a thin pass-through anymore. it now serialize concurrent first-touch of the same id, which calling the store straight does NOT give you. that is the holder reason to exist. SaveSessionAsync/DeleteSessionAsync stay thin on purpose: they bind the agent so route code hold one object (state) instead of juggling agent + store and repeating the agent arg every call. this shape lines up with the python holder (get_or_create_session value-add + thin set_session).

so keeping HostedAgentState as-is. fixed in fc0b19b2e.

dotnet/src/Microsoft.Agents.AI.Hosting/HostedAgentState.cs
```diff
@@ -0,0 +1,174 @@
...
+/// </para>
+/// <para>
+/// <strong>Trust boundary.</strong> The <c>sessionStoreId</c> values passed to these methods are
+/// application-selected partition keys. When a key originates from the wire (for example via
+/// <c>OpenAIResponses.GetSessionId(...)</c>), the application must authenticate the caller and authorize
+/// the key before using it here. For multi-user hosts, scope the underlying store per principal with
+/// <see cref="IsolationKeyScopedAgentSessionStore"/>.
+/// </para>
+/// </remarks>
+public sealed class HostedAgentState
```
westey-m: It would be nice to also explain the session lock functionality the class provides in the class xml docs.

dotnet/src/Microsoft.Agents.AI.Hosting/HostedAgentState.cs
```diff
@@ -0,0 +1,174 @@
...
+        bool acquired = false;
+        try
+        {
+            await entry.Semaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
+            acquired = true;
+            return await this._sessionStore.GetSessionAsync(this._agent, sessionStoreId, cancellationToken).ConfigureAwait(false);
+        }
+        finally
+        {
+            this.ReleaseSessionLock(sessionStoreId, entry, acquired);
```
westey-m: I'm confused about the releasing here.  I thought that the session should be locked until the request has been handled completely by the asp.net route handler.  Doesn't the session lock get relinquished immediately here though, so only getting/creating the session is serialized, not the full request handling?

westey-m (approved):

docs/decisions/0032-dotnet-hosting-protocol-helpers.md
```diff
@@ -0,0 +1,163 @@
...
+
+`Microsoft.Agents.AI.Hosting` gains:
+
+- `AgentSessionStore.DeleteSessionAsync(...)` (+ `InMemoryAgentSessionStore` implementation and
+  isolation-decorator passthrough): the one missing store operation.
+- No agent-side holder. Applications use `AgentSessionStore` directly: `GetSessionAsync(agent, id)`
+  already creates on miss and returns an independent session instance per call (so concurrent calls fork
+  the same stored state rather than sharing an instance), `SaveSessionAsync(agent, id, session)` persists
+  post-run (including under a newly minted id), and `DeleteSessionAsync(agent, id)` removes it. An earlier
+  draft added a `HostedAgentState` holder, but once create-on-miss lives in the store and the store does no
```
eavanvalkenburg, line 109: The AgentState does two things, one discussed here, create on miss. But the other is to allow you to have a callable as a agent, which is there for two reasons, 1) ensure a new instance of a agent for each run and 2) allow async setup for a agent, simplifying the overall server setup which can be sync because of this. So if the existing store already tackles create-on-miss then that is not a reason to skip the AgentState altogether.

rogerbarreto: good catch, you right. AgentState do three thing: create-on-miss, fresh instance per run, and async/deferred setup. old ADR only answer the first one. fixed both the table row and the bullet. now say: create-on-miss live in the store; the other two (per run instance, async setup) come from DI in .NET, AddScoped/AddTransient for fresh per run, async factory registration for deferred setup, so route handler just resolve a ready agent. also note AIAgent safe to call concurrent (state live in AgentSession not the agent), so fresh-per-run not needed like workflow. workflow keep the factory because one Workflow instance is stateful run engine and cannot run two runner at once, that is the real asymmetry. done in 57c4c9d0b.

dotnet/src/Microsoft.Agents.AI.Hosting/AgentSessionStore.cs
```diff
@@ -51,29 +51,58 @@ public abstract class AgentSessionStore
...
+    /// <param name="agent">The agent that owns this session.</param>
+    /// <param name="sessionStoreId">The id under which the session is stored.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/> to monitor for cancellation requests.</param>
+    /// <returns>A task that represents the asynchronous delete operation.</returns>
+    /// <remarks>
+    /// Implementations that support removal delete the session and treat a missing session as a no-op.
+    /// Implementations that genuinely cannot support deletion should throw <see cref="NotSupportedException"/>.
+    /// </remarks>
+    /// <exception cref="NotSupportedException">The store does not support deletion.</exception>
+    public abstract ValueTask DeleteSessionAsync(
```
peibekwe, line 103: The PR description mentioned that this was added as virtual to avoid it being a breaking change, but it's actually abstract. Was it intended to be virtual? Abstract would mean it's a breaking change as any exciting subclass would have to implement it.

dotnet/src/Microsoft.Agents.AI.Hosting/AgentSessionStore.cs
```diff
@@ -51,29 +51,58 @@ public abstract class AgentSessionStore
...
+    /// snapshot on every call; implementations that cache a live <see cref="AgentSession"/> must return an
+    /// independent copy (for example by round-tripping through
+    /// <see cref="AIAgent.SerializeSessionAsync(AgentSession, System.Text.Json.JsonSerializerOptions?, CancellationToken)"/>
+    /// and <see cref="AIAgent.DeserializeSessionAsync(System.Text.Json.JsonElement, System.Text.Json.JsonSerializerOptions?, CancellationToken)"/>)
+    /// rather than handing back the shared instance.
+    /// </remarks>
     public abstract ValueTask<AgentSession> GetSessionAsync(
         AIAgent agent,
-        string conversationId,
+        string sessionStoreId,
```
peibekwe, line 88: nit: This is also a breaking change, right? Perhaps the PR needs to be marked as breaking.

dotnet/src/Microsoft.Agents.AI.Hosting/HostedWorkflowState.cs
```diff
@@ -0,0 +1,373 @@
...
+        {
+            return await this._workflowFactory!(cancellationToken).ConfigureAwait(false);
+        }
+
+        // Cached factory: build the workflow once. The lock guards only the one-time task assignment (and the
+        // factory's synchronous prefix); the actual build is awaited outside the lock. CancellationToken.None is
+        // used because the single built instance is shared across runs and must not be tied to one request.
+        lock (this._cacheSync)
+        {
+            this._cachedWorkflowTask ??= this._workflowFactory!(CancellationToken.None).AsTask();
```
peibekwe: nit: If the factory task faults, is it acceptable for the faulted task to be cached and every later run re-await the same failure?

dotnet/src/Microsoft.Agents.AI.Workflows/StreamingRun.cs
```diff
@@ -77,7 +77,24 @@ public IAsyncEnumerable<WorkflowEvent> WatchStreamAsync(
...
+    /// stream instead ends at that halt, returning control to the caller — matching the non-streaming
+    /// <see cref="Run"/> behavior, which is useful for hosts that drive their own turn loop.
+    /// </remarks>
+    /// <param name="blockOnPendingRequest"><see langword="true"/> to block at a pending-request halt awaiting a
+    /// response; <see langword="false"/> to end the stream at that halt.</param>
+    /// <param name="cancellationToken">A <see cref="CancellationToken"/> that can be used to cancel the streaming
+    /// operation. If cancellation is requested, the stream will end and no further events will be yielded, but this
+    /// will not cancel the workflow execution.</param>
+    /// <returns>An asynchronous stream of <see cref="WorkflowEvent"/> objects representing significant workflow state changes.</returns>
+    public IAsyncEnumerable<WorkflowEvent> WatchStreamAsync(
```
peibekwe, line 97: nit: Is this intentionally made public? Just double checking because once it's public, we can't easily change that.

dotnet/src/Microsoft.Agents.AI.Workflows/Checkpointing/FileSystemJsonCheckpointStore.cs
```diff
@@ -34,7 +34,19 @@ public sealed class FileSystemJsonCheckpointStore : JsonCheckpointStore, IDispos
...
+    // RetrieveCheckpointAsync. It intentionally coexists with OrderedCheckpointIndex below: this set answers
+    // "does this checkpoint exist?" in O(1), while the list preserves commit order. A single HashSet cannot do
+    // both because HashSet enumeration order is not a contract.
     internal HashSet<CheckpointInfo> CheckpointIndex { get; }
+
+    // Insertion-ordered mirror of CheckpointIndex. HashSet enumeration order is not a contract (it can diverge
+    // from insertion order once a slot is freed by a rollback and reused), so RetrieveIndexAsync enumerates this
+    // list to return checkpoints in commit order, which callers such as CheckpointManager.GetLatestCheckpointAsync
+    // rely on to identify the head checkpoint.
+    private List<CheckpointInfo> OrderedCheckpointIndex { get; } = [];
```
peibekwe, line 48: nit: Probably not possible as checkpoints are unique, but there are no possible duplicate issues, right?

peibekwe (approved):

dotnet/src/Microsoft.Agents.AI.Hosting/AgentSessionStore.cs
rogerbarreto: yes, breaking. new abstract member on a public type plus the sessionStoreId param rename. updated the PR desc, dropped the non-breaking claim, added the breaking label.

dotnet/src/Microsoft.Agents.AI.Hosting/AgentSessionStore.cs
rogerbarreto: kept abstract on purpose. hosting still preview, so this is the window to force each store to make a conscious choice: implement delete or throw itself. virtual-with-throw would ship a footgun, a store inherits a method that compiles fine but blows up at runtime. you right it is breaking though, the PR desc was stale saying virtual. fixed the desc and added the breaking label.

dotnet/src/Microsoft.Agents.AI.Hosting/HostedWorkflowState.cs
rogerbarreto: good catch. before, a faulted build task stayed cached forever and every run re-awaited the same fail. fixed: reuse the cached build only when it is not faulted and not canceled, else the next run rebuilds fresh. added a test that faults the first build then succeeds. 17bc781a7.

dotnet/src/Microsoft.Agents.AI.Workflows/Checkpointing/FileSystemJsonCheckpointStore.cs
rogerbarreto: right, no dup possible. ids are unique by construction: GetUnusedCheckpointInfo loops until HashSet.Add succeeds and only then pushes to the ordered list, and the load path guards with if(CheckpointIndex.Add(...)). added tests: a duplicate index line on reload still yields one checkpoint, and many creates stay distinct in commit order. 17bc781a7.

dotnet/src/Microsoft.Agents.AI.Workflows/StreamingRun.cs
rogerbarreto: intentional. HostedWorkflowState lives in Hosting and needs this overload. exposing Workflows internals to Hosting via InternalsVisibleTo leaks Workflows own internal Throw helper and blows up with CS0436 (hundreds of dup errors), so the overload went public as the least-bad option. open to a cleaner split later.

westey-m (approved):

PR #7001, opened 2026-07-08
Python: HITL respond-URL addressing from inside workflows

ahmedmuhsin:
### Motivation & Context

The durable function HITL flow pauses a workflow with `request_info` and waits for a human to POST a response to the `/respond` endpoint. The gap was that nothing inside the workflow could build that respond URL. An executor only receives its message payload, not the orchestration `instanceId`, and the `requestId` does not exist until `request_info` runs. So to email someone an approval link you had to pull both ids from the HTTP run response or poll the status endpoint from outside the app.

This came out of a real POC where someone wanted Agent B to email a reviewer when it finished, then resume to Agent C once approved, but had no way to address the running workflow from within it. This change lets a workflow build its own respond URL and notify a human from inside the graph with no caller plumbing of `instanceId` or `requestId`, and it works for nested sub workflows too.

  - The orchestrator now surfaces orchestration metadata (the addressable root `instanceId`, the workflow name, and for nested workflows the request path prefix) onto each executor through the durable runner context. There is no new core API.
  - A new `WorkflowHitlContext` helper in `agent-framework-azurefunctions` builds the canonical respond and status URLs from that metadata. It returns `None` in process so the same executor degrades gracefully when it is not on the durable host.
  - `WorkflowHitlContext.pending_request_id(ctx)` reads the id `request_info` just generated back off the context, so the user never mints an id by hand. It works on any host through the existing core runner context protocol.
  - For nested sub workflows the address context propagates down `call_sub_orchestrator` via a new `SUBWORKFLOW_ADDRESS_KEY` marker, so an executor at any depth builds a URL that targets the top level instance with a qualified request id. That marker is stripped from untrusted input alongside the existing input marker, so a top level caller cannot forge it.
  - Samples 12 and 13 show the two step pattern. The review executor reads the request id back and sends it to a downstream notify executor that builds the URL, so a retried activity never emails a dead link because only the committed attempt reaches the notifier.

  - Additive. No core change, nothing public removed, and top level behavior is unchanged when the helper is not used.
  - New unit tests cover the metadata round trip, the id read back, the nested ordinal and prefix agreement with the read side, marker stripping, and URL building. Integration tests assert that the URL the helper builds equals the server `respondUrl` and that posting to it resumes the run, for both the flat (12) and nested (13) samples.

  - The per executor ordinal used to build the nested prefix at dispatch has to match the `enumerate` index the status and respond endpoints use when they resolve a nested request. That agreement is the one spot where a mistake would send an emailed URL to the wrong child or a 404.

### Related Issue

No separate issue. This builds on the Durable Task multi-workflow and sub-workflow hosting work that landed on `main` in #6696, and now targets `main` directly.

### Contribution Checklist


python/samples/04-hosting/azure_functions/12_workflow_hitl/README.md
```diff
@@ -37,6 +37,31 @@ async def handle_approval_response(
...
+# Pause the workflow. request_info generates the request id internally.
+await ctx.request_info(
+    request_data=HumanApprovalRequest(...),
+    response_type=HumanApprovalResponse,
+)
+
+# Read that id back, then build the respond URL with WorkflowHitlContext.
+request_id = await WorkflowHitlContext.pending_request_id(ctx)
+hitl = WorkflowHitlContext.from_context(ctx)
+respond_url = hitl.build_respond_url(request_id)  # email this to the reviewer
```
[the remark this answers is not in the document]
ahmedmuhsin: Addressed in a1030c7c4. Added an `if hitl and request_id:` guard before `build_respond_url` in the snippet so it matches the graceful-degradation behavior described in the surrounding text.

python/samples/04-hosting/azure_functions/13_subworkflow_hitl/README.md
```diff
@@ -45,6 +45,25 @@ child orchestration, so a nested `request_info` is recorded on the *child* insta
...
+
+```python
+# review_gate, inside the nested human_review workflow
+await ctx.request_info(request_data=HumanApprovalRequest(...), response_type=HumanApprovalResponse)
+request_id = await WorkflowHitlContext.pending_request_id(ctx)
+# ... send request_id to the notify executor ...
+
+# NotifyExecutor (also inside the inner workflow)
+hitl = WorkflowHitlContext.from_context(ctx)
+respond_url = hitl.build_respond_url(request_id)  # already qualified back to the root
```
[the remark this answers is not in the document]
ahmedmuhsin: Addressed in a1030c7c4. Added the same `if hitl and request_id:` guard before `build_respond_url` in the nested snippet.

python/packages/azurefunctions/tests/integration_tests/test_12_workflow_hitl.py
```diff
@@ -214,6 +217,68 @@ def test_hitl_workflow_with_neutral_content(self) -> None:
...
+        # Wait for the workflow to reach the HITL pause point
+        status = self._wait_for_hitl_request(instance_id)
+        pending_requests = status.get("pendingHumanInputRequests", [])
+        assert len(pending_requests) > 0, "Expected pending HITL request"
+        pending = pending_requests[0]
+        request_id = pending["requestId"]
+
+        # The review executor now generates an explicit uuid4 and passes it to
+        # request_info, so the pending id round-trips as a valid UUID (not an opaque
+        # framework default).
```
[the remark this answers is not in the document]
ahmedmuhsin: Addressed in a1030c7c4. Reworded the comment to say `request_info` generates the request id internally as a uuid4 when the caller does not pass one, so it no longer implies a behavior change.

python/packages/azurefunctions/tests/integration_tests/test_13_workflow_subworkflow_hitl.py
```diff
@@ -145,6 +148,52 @@ def test_nested_hitl_rejection(self) -> None:
...
+        })
+        instance_id = data["instanceId"]
+
+        status = self._wait_for_hitl_request(instance_id)
+        pending = status.get("pendingHumanInputRequests", [])
+        assert len(pending) == 1
+        qualified_id = pending[0]["requestId"]
+
+        # The qualified id is ``review_sub~0~{bare}``; the inner review gate now
+        # generates an explicit uuid4 as the bare id.
```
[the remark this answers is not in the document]
ahmedmuhsin: Addressed in a1030c7c4. Reworded the comment so it no longer implies a new behavior. The bare inner id comes from `request_info` default uuid4 generation.

python/packages/azurefunctions/agent_framework_azurefunctions/_hitl_context.py
```diff
@@ -0,0 +1,202 @@
...
+from typing import Any, cast
+
+# App setting carrying the function app's host (e.g. ``myapp.azurewebsites.net``).
+# Azure Functions sets this automatically in the cloud; for local ``func start`` runs
+# add it to the ``Values`` map in ``local.settings.json`` (e.g. ``localhost:7071``).
+WEBSITE_HOSTNAME_ENV = "WEBSITE_HOSTNAME"
+
+# Default Azure Functions HTTP route prefix; the workflow routes registered by
+# AgentFunctionApp live under ``/api/workflow/...`` to match it.
+_API_ROUTE_PREFIX = "api"
```
larohra: This hardcodes the `api` route prefix, but the prefix is configurable via `routePrefix` in `host.json` (and can be set to empty). If a customer customizes it, every respond/status URL built here will 404 on resume. We should derive the prefix from the host config / the same source the route registration uses rather than a literal. Given there's already a customer on this path, worth fixing before merge.

python/packages/azurefunctions/agent_framework_azurefunctions/_hitl_context.py
```diff
@@ -0,0 +1,202 @@
...
+                "Functions; add it to the 'Values' map in local.settings.json for local "
+                "`func start` runs, e.g. 'localhost:7071'), or pass base_url=... to "
+                "WorkflowHitlContext.from_context()."
+            )
+
+        # An override may already include a scheme; WEBSITE_HOSTNAME is host-only, so
+        # infer one (http for local loopback, https otherwise).
+        if hostname.startswith(("http://", "https://")):
+            return hostname.rstrip("/")
+        scheme = "http" if hostname.startswith(("localhost", "127.0.0.1")) else "https"
```
larohra: Loopback detection here only covers `localhost` and `127.0.0.1`. `func start` can also bind `0.0.0.0`, and IPv6 loopback is `::1` -- both fall through to `https`, producing an unreachable link locally. Let's broaden this (or pull it into a small `_is_loopback(host)` helper so the list lives in one place).

python/packages/azurefunctions/agent_framework_azurefunctions/_hitl_context.py
```diff
@@ -0,0 +1,202 @@
...
+                ``ctx.request_info``. Pass the **bare** id even from inside a nested
+                sub-workflow: any :attr:`request_path_prefix` is prepended for you to
+                qualify it (``{executor}~{ordinal}~{requestId}``) back to the root.
+
+        Returns:
+            The fully-qualified respond URL.
+        """
+        qualified_id = f"{self.request_path_prefix}{request_id}"
+        return (
+            f"{self.base_url}/{_API_ROUTE_PREFIX}/workflow/"
```
larohra: This respond-URL template is a second copy of the `respondUrl` the server builds in `_app.py`; `build_status_url` just below duplicates the status URL the same way. Today they're only kept aligned by the integration test asserting the two strings match -- that "synced by a test" smell is worth removing while the context is fresh. Both call sites are in this package, so let's extract one shared builder -- e.g. `build_respond_url(base, workflow_name, instance_id, qualified_id)` -- and call it from both the route/status endpoint and here. It also gives the route prefix a single home to reconcile with `host.json` (see my note on `_API_ROUTE_PREFIX`).

python/packages/durabletask/agent_framework_durabletask/_workflows/orchestrator.py
```diff
@@ -273,6 +280,17 @@ def _prepare_activity_task(
         "message": serialize_value(message),
         "shared_state_snapshot": shared_state_snapshot,
         "source_executor_ids": [source_executor_id],
+        # host_context addresses the *root* (HTTP-routable) orchestration so an executor
+        # can build a HITL respond URL (see CapturingRunnerContext.host_metadata):
+        # instance_id / workflow_name name the top-level instance, and
+        # request_path_prefix is the accumulated ``{executor}~{ordinal}~`` hops from the
+        # root down to this workflow level. For a top-level workflow the prefix is empty,
+        # so this reduces to addressing the instance directly.
+        "host_context": {
```
larohra, line 294: These keys (`instance_id`, `workflow_name`, `request_path_prefix`) are a cross-package contract: `WorkflowHitlContext.from_context` in the azurefunctions package reads them back by the same literal strings. A rename on either side degrades silently -- `from_context` just returns `None` and HITL URLs stop being built, with no error to trace. Let's pin the key names in shared constants (or a small TypedDict) so producer and consumer can't drift.

python/packages/durabletask/agent_framework_durabletask/_workflows/orchestrator.py
```diff
@@ -797,13 +863,25 @@ def _prepare_all_tasks(
...
 
     agent_messages_by_executor: dict[str, list[tuple[str, Any, str]]] = defaultdict(list)
 
+    # Per-executor, per-superstep ordinal for sub-workflow dispatch. This must match the
+    # read side's enumerate() index into the custom-status ``subworkflows[executorId]``
+    # list (built in this same dispatch order), so a nested pending request resolves
+    # back to the right child. It is deliberately distinct from ``subworkflow_counter``
+    # (a global, cross-superstep counter that only guarantees child-instance-id
+    # uniqueness, not addressing position).
+    per_executor_sub_ordinal: dict[str, int] = defaultdict(int)
```
larohra, line 888: Lower priority -- hardening, not a blocker. Your comment above nails the risk: this write-side ordinal has to match the read side's `enumerate()` index into `subworkflows[executorId]`, and they agree only because both walk the same dispatch order. It would desync silently if anyone ever sorts or filters `task_metadata_list`. Since you're already appending a `TaskMetadata` for each child right below, consider stamping the ordinal onto `TaskMetadata` at dispatch and having the read side consume that -- one source of truth instead of two independent derivations.

larohra (commented):
A few notes on the HITL addressing mechanism -- 2 likely must-fix URL-correctness bugs (hardcoded route prefix, loopback scheme), 2 cheap de-risking refactors (respond/status URL dedup, the cross-package `host_context` key contract), and 1 optional hardening note on the sub-workflow ordinal. Details inline. Nothing here blocks the durable use case; the first two just affect whether generated links resolve in non-default setups.

kshyju (approved):
Thorough review of all 18 files. The architecture is clean — address context propagation through the orchestration tree, security-conscious marker stripping, and the retry-safe two-step notify pattern are all well done. The critical per-executor ordinal / enumerate agreement is correct and well-tested. LGTM.

**Minor (non-blocking):** In `_hitl_context.py` ~L192, the comment says *"An override may already include a scheme"* but at that point `base_url_override` is already falsy (returned early). The comment should reference `WEBSITE_HOSTNAME` instead, e.g.: *"WEBSITE_HOSTNAME may include a scheme (unusual but possible); otherwise infer one (http for local loopback, https otherwise)."*

kshyju:
> Thorough review of all 18 files. The architecture is clean — address context propagation through the orchestration tree, security-conscious marker stripping, and the retry-safe two-step notify pattern are all well done. The critical per-executor ordinal / enumerate agreement is correct and well-tested. LGTM.
> 
> **Minor (non-blocking):** In `_hitl_context.py` ~L192, the comment says _"An override may already include a scheme"_ but at that point `base_url_override` is already falsy (returned early). The comment should reference `WEBSITE_HOSTNAME` instead, e.g.: _"WEBSITE_HOSTNAME may include a scheme (unusual but possible); otherwise infer one (http for local loopback, https otherwise)."_

From GHCP review ^

python/packages/azurefunctions/agent_framework_azurefunctions/_hitl_context.py
ahmedmuhsin: Fixed in 348da7826 and refined in 8d64f0a09. The server now derives the base and route prefix straight from the incoming request URL, the value the host actually routed, and the in-workflow helper reads the prefix from `host.json` since it has no request context. So a custom or empty `routePrefix` resolves instead of 404ing, and the caller-visible URLs come from the real request rather than a hardcoded `api`.

python/packages/azurefunctions/agent_framework_azurefunctions/_hitl_context.py
ahmedmuhsin: Fixed in 348da7826. Pulled it into a `_is_loopback` helper that also covers 0.0.0.0 and IPv6 ::1, including the bracketed [::1] form, with tests.

python/packages/azurefunctions/agent_framework_azurefunctions/_hitl_context.py
ahmedmuhsin: Done in 348da7826. Extracted the respond and status URL builders into the shared `_routes` module and call them from both `_app.py` and the helper, so they are no longer kept in sync only by the integration test.

python/packages/durabletask/agent_framework_durabletask/_workflows/orchestrator.py
ahmedmuhsin: Done in 348da7826. Pinned the key names as constants in durabletask (`HOST_METADATA_INSTANCE_ID`, `HOST_METADATA_WORKFLOW_NAME`, `HOST_METADATA_REQUEST_PATH_PREFIX`) and both the orchestrator and `from_context` use them, so they cannot drift.

python/packages/durabletask/agent_framework_durabletask/_workflows/orchestrator.py
ahmedmuhsin: Good call. Rather than leave this as an implicit invariant, I extracted the read-side grouping into a shared `_index_subworkflows` helper and added `test_readside_index_matches_dispatch_ordinal` in 33b4b10a2. It round-trips a fan-out through that helper and asserts `subworkflows[executor][ordinal]` resolves to the child the dispatch stamped that ordinal onto, so the write ordinal and read index agreement is now enforced in CI rather than just documented.

ahmedmuhsin:
Thanks for the thorough review. I fixed the `base_url` comment in 348da7826 to reference `WEBSITE_HOSTNAME` instead of an override, since the override returns early above.

python/packages/azurefunctions/agent_framework_azurefunctions/_routes.py
```diff
@@ -0,0 +1,117 @@
...
+    script_root = os.environ.get("AzureWebJobsScriptRoot") or os.getcwd()  # noqa: SIM112
+    host_json_path = os.path.join(script_root, "host.json")
+    try:
+        with open(host_json_path, encoding="utf-8") as f:
+            loaded = json.load(f)
+    except (OSError, ValueError):
+        logger.debug("Could not read '%s'; defaulting route prefix to '%s'.", host_json_path, DEFAULT_ROUTE_PREFIX)
+        return DEFAULT_ROUTE_PREFIX
+    if not isinstance(loaded, dict):
+        return DEFAULT_ROUTE_PREFIX
```
larohra, line 57: Follow-up on the (now resolved) route-prefix handling — flagging the `host.json` read as a coupling to note, not a blocker.

Reading `host.json` here is reasonable and I would not block on it: it is the app reading its own bundled config, read-only, cached per process, with a graceful `api` fallback, and `routePrefix` genuinely is not exposed via any env var or SDK API — so the in-executor helper (which has no request context) has no other way to get it. It is also well-isolated in this module. No security concern.

Two things worth calling out:
1. This makes worker code depend on a host-owned file schema and on-disk location (`AzureWebJobsScriptRoot`). `routePrefix` is stable, but the coupling is undocumented platform surface.
2. `host.json` gives the *declared* prefix, not the *effective public* one. Behind APIM / a custom domain the public prefix can differ, and there is a `base_url` override but no prefix override to correct it — so the emitted URL could be wrong in those setups.

Cleaner direction: the run/status HTTP handler already sees the authoritative prefix and base URL, and the orchestrator already threads `host_context` to executors. Capturing the prefix/base at that HTTP boundary and passing it through `host_context` would let the executor drop the `host.json` read entirely and guarantee server- and executor-built URLs are identical.

larohra (commented):
One follow-up note on the resolved route-prefix change (inline on _routes.py).

larohra (approved):

python/packages/azurefunctions/agent_framework_azurefunctions/_routes.py
ahmedmuhsin: Looked at this already and kept the hybrid on purpose. Threading the prefix into `host_context` means carrying it in on the durable run input, which is the hardened trust boundary, so it needs its own envelope and stripping instead of reusing the sub-workflow one that routes into `pickle`. That is a lot of surface to add just to drop one cached read-only read on the executor path. It also would not fully fix the APIM case, since the host sees the backend URL rather than the public one, so I will track a prefix override as a follow-up rather than block here.

Issue #6979, opened 2026-07-07
Python: [Bug]: OpenAI Chat Completions client buries plaintext `reasoning_details` as encrypted data


antsok:
### Description

`OpenAIChatCompletionClient` JSON-dumps the **entire** `reasoning_details` array into `Content.protected_data` and never sets `Content.text` — even when the entries are plaintext (`{"type": "reasoning.text", "text": "…"}`, as OpenRouter sends for e.g. `z-ai/glm-5.2`). `protected_data` is the channel for **encrypted** reasoning, so `agent-framework-ag-ui` emits a `ReasoningEncryptedValueEvent` (opaque) instead of a `ReasoningMessageContentEvent` (visible). Net effect on any AG-UI/CopilotKit frontend: a "Thought for N seconds" block with nothing inside,
although the readable reasoning is sitting right there inside `protected_data`.

The sibling `OpenAIResponsesClient` handles the same concept correctly: `Content.from_text_reasoning(id=item_id, text=delta)` — text populated, so it renders.

## Environment

- `agent-framework-core` 1.10.0, `agent-framework-openai` 1.10.0, `agent-framework-ag-ui` 1.0.0rc7
- Provider: OpenRouter (`base_url=https://openrouter.ai/api/v1`) via `OpenAIChatCompletionClient`, model `z-ai/glm-5.2`, streaming, `extra_body={"reasoning": {"enabled": True}}`.

## Reproduction

1. Point `OpenAIChatCompletionClient` at OpenRouter with a reasoning model and stream a prompt.
2. Inspect the parsed contents (or the AG-UI event stream). Each streamed `text_reasoning` `Content` has `text is None` and `protected_data = '[{"type":"reasoning.text","text":"…"}]'`, yielding `REASONING_ENCRYPTED_VALUE` events and no `REASONING_MESSAGE_CONTENT`.

### Code Sample

```markdown

```

### Error Messages / Stack Traces

```markdown

```

### Package Versions

agent-framework-core:1.10.0, agent-framework-openai:1.10.0, agent-framework-ag-ui:1.0.0rc7

### Python Version

Python 3.12

## Root cause

`python/packages/openai/agent_framework_openai/_chat_completion_client.py` — identical in the streaming (`_parse_response_update_from_openai`) and non-streaming paths:

```python
if reasoning_details := getattr(choice.delta, "reasoning_details", None):
    contents.append(Content.from_text_reasoning(protected_data=json.dumps(reasoning_details)))
```

[...]

Issue #6978, opened 2026-07-07
Python: [Bug]: OpenAI Chat Completions client crashes on (and drops) Mistral reasoning content chunks


antsok:
## Summary

Mistral's reasoning models (`mistral-medium-latest` / Medium 3.5, `mistral-small-latest`) return the assistant `content` as **a list of chunks** rather than a string when reasoning is enabled — a `ThinkChunk` (`{"type": "thinking", "thinking": [ {"type": "text", "text": …} ]}`) followed by a `TextChunk` (`{"type": "text", "text": …}`) — per https://docs.mistral.ai/studio-api/conversations/reasoning. The OpenAI SDK passes that list through unchanged (it does not coerce/reject it in streaming), but MAF's `OpenAIChatCompletionClient` assumes `content` is a string and wraps it verbatim in a single text `Content`. The run then **crashes** downstream in `agent-framework-ag-ui`:

```
pydantic_core.ValidationError: 1 validation error for TextMessageContentEvent
  Input should be a valid string [type=string_type,
   input_value=[{'type': 'thinking', ...}, {'type': 'text', 'text': 'Okay'}], input_type=list]
```

So Mistral reasoning models are **unusable** through the Chat Completions client, and even if they didn't crash, the `thinking` trace would be discarded. (The `OpenAIResponsesClient` path parses reasoning correctly via `Content.from_text_reasoning(text=…)`; only Chat Completions is affected.)

## Environment

- `agent-framework-core` 1.10.0, `agent-framework-openai` 1.10.0, `agent-framework-ag-ui` 1.0.0rc7
- Provider: Mistral (`base_url=https://api.mistral.ai/v1`) via `OpenAIChatCompletionClient`,
  model `mistral-medium-latest`, streaming, request body `reasoning_effort: "high"`.

## Reproduction

1. Point `OpenAIChatCompletionClient` at Mistral with `mistral-medium-latest` and send `extra_body={"reasoning_effort": "high"}` (or use `magistral-*`, which reasons by default).
2. Stream any prompt. `choice.delta.content` arrives as a list of chunk dicts; the run raises the `TextMessageContentEvent` `string_type` ValidationError above.

### Code Sample

```markdown

```

### Error Messages / Stack Traces

```markdown

```

### Package Versions

agent-framework-core:1.10.0, agent-framework-openai:1.10.0, agent-framework-ag-ui:1.0.0rc7

### Python Version

Python 3.12

## Root cause

[...]

PR #7028, opened 2026-07-09
Python: Add response/request customization hooks to OpenAIChatCompletionClient

giles17:
### Motivation & Context

`OpenAIChatCompletionClient` targets the OpenAI Chat Completions wire format, but it is widely
pointed at "almost OpenAI-compatible" endpoints (OpenRouter, vLLM, Mistral, DeepSeek, Ollama, …)
that diverge on the edges — most visibly in how they represent reasoning (`reasoning`,
`reasoning_content`, `reasoning_details`) and, for some Mistral models, by returning `content` as a
list of typed chunks instead of a string. Branching in the stock client for each provider quirk does
not scale and puts the client on the critical path of every third-party implementation.

Following reviewer feedback, this keeps the stock client provider-agnostic and instead exposes two
optional callables so callers can adapt parsing and request-building to their specific endpoint
without subclassing. It also hardens default parsing so structured (non-string) content can no longer
produce a malformed `Content` that crashes downstream.

  - Two optional constructor callables on `RawOpenAIChatCompletionClient` / `OpenAIChatCompletionClient`:
    - `response_parser` (`OpenAIChatResponseContentsParser`): `(choice, default_contents) -> contents`.
      Post-processes the `Content` list parsed from each response choice / streaming delta, e.g. to
      surface a provider's non-standard reasoning field for display.
    - `message_preparer` (`OpenAIChatMessagePreparer`): `(message, default_dicts) -> dicts`.
      Post-processes the outgoing request message dicts built from each framework `Message`, e.g. to
      echo a provider's reasoning field back on later turns (required by vLLM-style endpoints for
      multi-turn continuity).
  - Both default to `None` — a no-op, so stock OpenAI/Azure behavior is unchanged.
  - Default text parsing now skips non-string `content` instead of wrapping it verbatim into a
    malformed `text` `Content`; this prevents a downstream crash and gives a `response_parser` a clean
    slate to expand structured chunks.
  - The two callable type aliases are exported from the package and the core lazy
    `agent_framework.openai` namespace (and its `.pyi`).

  - No behavior change for stock OpenAI/Azure users; the hooks are opt-in and default to `None`.
  - Provider quirks (OpenRouter/vLLM reasoning surfacing and echo-back) are now addressable through the
    hooks rather than in-core branches; Mistral chunked content is handled by the dedicated
    `agent-framework-mistral` client.
  - New public API surface: two callable type aliases and two constructor parameters.

  - The parse and prepare hook seams (placement and per-choice granularity), and the non-string
    `content` guard in `_parse_text_from_openai`.

### Related Issue

Related to #6978 and #6979. Rather than fix these with provider-specific branches in the stock client,
this adds the supported extension path to resolve them: a `response_parser` surfaces plaintext
reasoning (#6979) and a `message_preparer` echoes it back for multi-turn continuity, while the
non-string-content guard removes the #6978 crash and the dedicated `agent-framework-mistral` client
handles Mistral chunked content.

### Contribution Checklist


python/packages/openai/agent_framework_openai/_chat_completion_client.py
```diff
@@ -709,12 +757,14 @@ def _parse_response_from_openai(self, response: ChatCompletion, options: Mapping
...
                 finish_reason = choice.finish_reason  # type: ignore[assignment]
             contents: list[Content] = []
-            if text_content := self._parse_text_from_openai(choice):
-                contents.append(text_content)
+            contents.extend(self._parse_text_from_openai(choice))
             if parsed_tool_calls := [tool for tool in self._parse_tool_calls_from_openai(choice)]:
                 contents.extend(parsed_tool_calls)
             if reasoning_details := getattr(choice.message, "reasoning_details", None):
-                contents.append(Content.from_text_reasoning(protected_data=json.dumps(reasoning_details)))
+                contents.append(Content.from_text_reasoning(
```
moonbox3: Should visible reasoning normalization cover the other documented OpenRouter plaintext forms here? `reasoning.summary` entries put the displayable value under `summary`, while raw-reasoning models return `message.reasoning` (or `reasoning_content`) without `reasoning_details`; on this head the first becomes `text=None` and the second is discarded entirely. Could we normalize those fields in both streaming and non-streaming paths so the fix covers the provider's supported response shapes?

python/packages/openai/agent_framework_openai/_chat_completion_client.py
```diff
@@ -795,14 +847,57 @@ def _parse_usage_from_openai(self, usage: CompletionUsage) -> UsageDetails:
...
+                elif isinstance(thinking, list):
+                    text = "".join(
+                        str(cast(dict[str, object], part).get("text", ""))
+                        for part in cast(list[object], thinking)
+                        if isinstance(part, dict)
+                    )
+                else:
+                    text = ""
+                if text:
+                    results.append(Content.from_text_reasoning(text=text, raw_representation=choice))
```
moonbox3: Should we preserve the original structured assistant content for the next turn? Converting the `ThinkChunk` here and the text chunk below into separate framework contents makes `_prepare_message_for_openai()` replay them as two plain assistant messages, so the original `[{"type":"thinking", ...}, {"type":"text", ...}]` is lost. Mistral requires the complete assistant message, including the `ThinkChunk`, for multi-turn reasoning and tool continuation; could we retain or reconstruct that list so it round-trips as one message?

python/packages/openai/agent_framework_openai/_chat_completion_client.py
giles17: Addressed in 55e8a115. Added:
- `reasoning.summary` entries now extract the `summary` field
- `message.reasoning` / `reasoning_content` top-level fields surfaced as `Content.from_text_reasoning(text=...)` in both streaming and non-streaming paths
- `reasoning_details` takes priority when both fields are present (to avoid duplication)

python/packages/openai/agent_framework_openai/_chat_completion_client.py
giles17: Addressed in 55e8a115. The original list is now preserved in `additional_properties["_source_content_list"]` on the first Content item. In `_prepare_message_for_openai`, when this key is detected, we emit a single message with the original list as `content` and skip sibling text/reasoning items from the same source, so the structured `[{"type":"thinking",...},{"type":"text",...}]` round-trips intact for Mistral multi-turn reasoning.

python/packages/openai/agent_framework_openai/_chat_completion_client.py
```diff
@@ -709,12 +773,22 @@ def _parse_response_from_openai(self, response: ChatCompletion, options: Mapping
...
+                    protected_data=json.dumps(reasoning_details),
+                ))
+            # Some OpenAI-compatible providers (e.g. OpenRouter) expose plaintext reasoning
+            # via a top-level "reasoning" or "reasoning_content" field instead of (or in
+            # addition to) "reasoning_details".  Surface it when no reasoning was already parsed.
+            elif (reasoning_str := (
+                getattr(choice.message, "reasoning", None)
+                or getattr(choice.message, "reasoning_content", None)
+            )) and isinstance(reasoning_str, str) and reasoning_str:
+                contents.append(Content.from_text_reasoning(text=reasoning_str))
```
Kimahriman: This is great for surfacing reasoning from things like vLLM, but a lot of open-source models also expect the reasoning to be passed back on subsequent requests. Would be nice if this could be propagated back in the request based on the key it was seen in the response. (i.e. vLLM returns `reasoning` and expects `reasoning` on associated message in the next request)

python/packages/openai/agent_framework_openai/_chat_completion_client.py
```diff
@@ -828,14 +925,74 @@ def _parse_usage_from_openai(self, usage: CompletionUsage) -> UsageDetails:
...
+            return [Content.from_text(text=content, raw_representation=choice)]
         if hasattr(message, "refusal") and message.refusal:
-            return Content.from_text(text=message.refusal, raw_representation=choice)
-        return None
+            return [Content.from_text(text=message.refusal, raw_representation=choice)]
+        return []
+
+    @staticmethod
+    def _parse_chunked_content(chunks: list[Any], choice: Choice | ChunkChoice) -> list[Content]:
+        """Parse structured content chunks (e.g. Mistral thinking/text format).
```
TaoChenOSU: nit: why would Mistral show up here in the OpenAI client?

TaoChenOSU (approved):

python/packages/openai/agent_framework_openai/_chat_completion_client.py
```diff
@@ -120,6 +121,106 @@ class PromptCacheOptions(TypedDict, total=False):
 ResponseModelT = TypeVar("ResponseModelT", bound=BaseModel | None, default=None)
 
 
+def _extract_reasoning_text(reasoning_details: Any) -> str | None:
+    """Extract visible plaintext from a reasoning_details payload.
+
+    OpenAI-compatible providers return reasoning_details in various forms:
```
eavanvalkenburg: this for me begs the question of what we want, is the `OpenAIChatCompletionClient` for OpenAI or for all "almost-compatible" OpenAI endpoints, because what we are solving for here is that the so-called compatible endpoints, are not actually compatible because they behave differently. So I hesitate to let this in, because at some point we might end up with a long tree of dedicated parsers that we have to go through for each of those, I would rather then see if we can add a callable that the user can supply that does that for their particular querk in the almost compatible surface or using subclassing to achieve this, I think that is both a more sustainable route and ultimately one that will make things easier for users, because they can then either create a `OpenRouterParser`/`MistralParser` or a `OpenRouterChatCompletionClient`/`MistralChatClient` (the latter we will probably want anyway, but that is besides the point) instead of this class bloating with all the possible exceptions and tweaks.

eavanvalkenburg (dismissed):

python/packages/openai/agent_framework_openai/_chat_completion_client.py
Kimahriman: FWIW we had to switch to pydantic ai because of this lack of reasoning support. I accidentally came across the fact that even the responses client doesn't send reasoning back to the server, even though it correctly populates it in the output (we use vLLM). I couldn't find any easy way to add custom hooks to support it, as the parsing of responses and building of requests were not very extensible, and by the time any middleware runs the original request was lost.

The main case for including this here is the shift toward the responses API for all actual openai things, where it seems the chat completion API is moreso used by "compatible" services. But making it possible to easily extend it with this type of functionality would go a long way toward making this useful for self-hosted users

eavanvalkenburg: @Kimahriman thanks for the background, we will take that into account

python/packages/openai/agent_framework_openai/_chat_completion_client.py
moonbox3, line 1063: I just noticed here: should `message_preparer` run for `system` and `developer` messages too? Both returns bypass the hook at `_chat_completion_client.py:1149`, despite the public contract saying it is called once per `Message`. A gateway that requires a provider-specific field on every outgoing message will silently receive an incomplete request, so could we route these returns through the common hook path?

python/packages/openai/agent_framework_openai/_chat_completion_client.py
```diff
@@ -139,6 +139,32 @@ def _sanitize_author_name(name: str | None) -> str | None:
 ResponseModelT = TypeVar("ResponseModelT", bound=BaseModel | None, default=None)
 
 
+OpenAIChatResponseContentsParser: TypeAlias = Callable[["Choice | ChunkChoice", list[Content]], list[Content]]
```
moonbox3: Could this seam receive the already-selected `message` or `delta` rather than `Choice | ChunkChoice`? The current public interface makes every parser depend on two OpenAI SDK types and duplicate this module's streaming dispatch, as the new tests do via `Any` and `hasattr`. Keeping that dispatch here would restore locality and narrow the interface.

python/packages/openai/agent_framework_openai/_chat_completion_client.py
```diff
@@ -1088,6 +1146,8 @@ def _prepare_message_for_openai(self, message: Message) -> list[dict[str, Any]]:
                         for text_item in text_items
                     )
 
+        if self.message_preparer is not None:
```
moonbox3: Would it make sense to preserve where each piece of content came from at this boundary instead of only exposing flattened message dictionaries? Right now, the round trip test has to remove auto generated reasoning by matching `dict["content"]` strings. That means callers have to reconstruct this module's internal ordering, and they could accidentally remove unrelated content with the same text. Keeping that information would allow the reasoning parsing and serialization logic to stay contained within this module.

moonbox3 (approved):

PR #7029, opened 2026-07-09
Python: Best effort to serialize tool def to Json for observability

TaoChenOSU:
### Motivation & Context

A bug was discovered while working on a task that uses an agent that has the code interpreter tool. The observability module reported "TypeError: Object of type AutoCodeInterpreterToolParam is not JSON serializable". This happens because the observability module tries to serialize the tool definition to json while the object is not serilizable.

  - Add logic to serialize Azure SDK type
  - Best effort to serialize such that serialization failures don't fail the application (warning will be given)
  - Bug fix
  - The serialization logic 

### Contribution Checklist

### DIFF
--- a/python/packages/core/agent_framework/observability.py
+++ b/python/packages/core/agent_framework/observability.py
@@ -39,8 +39,10 @@ from typing import (
 
 from dotenv import load_dotenv
 from opentelemetry import metrics, trace
+from typing_extensions import Sentinel
 
 from . import __version__ as version_info
+from ._serialization import SerializationProtocol
 from ._settings import load_settings
 
 if sys.version_info >= (3, 13):
@@ -2324,7 +2326,7 @@ def _get_instructions_from_options(options: Any) -> str | list[str] | None:
 # the cache.
 _TOOL_OTEL_JSON_CACHE: weakref.WeakKeyDictionary[Any, str | None] = weakref.WeakKeyDictionary()
 # Sentinel distinguishing "not cached" from a cached ``None`` (unparseable tool).
-_CACHE_MISS: Final = object()
+_CACHE_MISS: Final = Sentinel("CACHE_MISS")
 
 
 def _serialize_tool_definitions(tools: Any) -> str | None:
@@ -2442,7 +2444,7 @@ def _build_tool_otel_definition(tool_item: Any) -> dict[str, Any] | None:
     raw: Mapping[str, Any] | None = None
     if isinstance(tool_item, BaseModel):
         raw = tool_item.model_dump(exclude_none=True)
-    elif isinstance(tool_item, SerializationMixin):
+    elif isinstance(tool_item, (SerializationMixin, SerializationProtocol)):
         raw = tool_item.to_dict()
     elif isinstance(tool_item, Mapping):
         mapping_item = cast("Mapping[str, Any]", tool_item)
--- a/python/packages/core/agent_framework/observability.py
+++ b/python/packages/core/agent_framework/observability.py
@@ -2325,6 +2325,10 @@ def _get_instructions_from_options(options: Any) -> str | list[str] | None:
 _TOOL_OTEL_JSON_CACHE: weakref.WeakKeyDictionary[Any, str | None] = weakref.WeakKeyDictionary()
 # Sentinel distinguishing "not cached" from a cached ``None`` (unparseable tool).
 _CACHE_MISS: Final = object()
+# Tool-spec fields that can carry secrets (e.g. an OAuth access token on an MCP
+# tool, or injected auth headers) and must never be copied into emitted telemetry.
+# Matched by field name so any MCP tool spec is covered, not just a specific class.
+_SENSITIVE_TOOL_DEFINITION_KEYS: Final[frozenset[str]] = frozenset({"authorization", "headers"})
 
 
 def _serialize_tool_definitions(tools: Any) -> str | None:
@@ -2484,9 +2488,13 @@ def _otel_definition_from_mapping(raw: Mapping[str, Any]) -> dict[str, Any] | No
         if parameters:
             definition["parameters"] = parameters
         # Forward extra properties from both layers, preferring the inner spec.
+        # Sensitive fields (e.g. auth tokens/headers) are dropped so telemetry
+        # never carries secrets.
         for source in (nested, raw):
             for key, value in source.items():
                 if key in {"type", "function", "name", "description", "parameters"}:
+                    continue
+                if key in _SENSITIVE_TOOL_DEFINITION_KEYS:
                     continue
                 definition.setdefault(key, value)
         return definition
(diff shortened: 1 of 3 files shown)


python/packages/core/agent_framework/observability.py
```diff
@@ -2366,7 +2451,12 @@ def _build_tool_otel_definition(tool_item: Any) -> dict[str, Any] | None:
     elif isinstance(tool_item, SerializationMixin):
         raw = tool_item.to_dict()
     elif isinstance(tool_item, Mapping):
-        raw = cast("Mapping[str, Any]", tool_item)
+        mapping_item = cast("Mapping[str, Any]", tool_item)
+        # Azure SDK tool models expose ``as_dict()``, which recursively converts
+        # the whole model (including nested non-dict Mapping values that are not
+        # JSON-serializable) into plain dicts; plain mappings are used as-is.
+        as_dict: Callable[[], Mapping[str, Any]] | None = getattr(mapping_item, "as_dict", None)
+        raw = as_dict() if callable(as_dict) else mapping_item
```
moonbox3, line 2455: Could we avoid forwarding the complete `as_dict()` result here? For example, `MCPTool(authorization="Bearer ...", headers=..., require_approval=MCPToolRequireApproval(...))` previously failed on the nested model, but this recursively flattens it and `_otel_definition_from_mapping()` copies the OAuth token and auth headers into `gen_ai.tool.definitions`. `_get_span_attributes()` builds this attribute even when sensitive-data capture is disabled, so any configured OTLP or Azure Monitor exporter receives live MCP credentials; could we redact `authorization` and credential-bearing headers recursively, or allowlist OTel-safe fields, and cover this case with a regression test?

python/packages/core/agent_framework/observability.py
```diff
@@ -2281,6 +2325,47 @@ def _get_instructions_from_options(options: Any) -> str | list[str] | None:
...
+                json.dumps(definition, ensure_ascii=False)
+            except Exception:
+                logger.warning(
+                    "Failed to serialize tool definition for telemetry; skipping tool %r.",
+                    definition.get("name") or definition.get("type") or "<unknown>",
+                    exc_info=True,
+                )
+            else:
+                serializable.append(definition)
+        return json.dumps(serializable, ensure_ascii=False) if serializable else None
```
eavanvalkenburg: each part of this is try-serialized separately and then the whole list is also serialized, that is a bit inefficient, we (try to) serialize everything twice, and in general, we first loop through the list of tools in `_tools_to_dict` and then we try to do the same here, this is not a good pattern, we should touch each tool exactly once, and figure out if we can use it, and skip or fail otherwise, but not go through everything multiple times

eavanvalkenburg (dismissed):

python/packages/core/agent_framework/observability.py
TaoChenOSU: Added logic to filter secrets. Currently there are only "authorization" and "headers", so this is best effort. 

Note that they are always filtered regardless of the sensitive data setting because secrets should never be logged.

python/packages/core/agent_framework/observability.py
TaoChenOSU: Optimized. We are cache the json result.

python/packages/core/agent_framework/observability.py
```diff
@@ -2262,83 +2306,122 @@ def _get_instructions_from_options(options: Any) -> str | list[str] | None:
...
-    Args:
-        tools: The tools to parse. Can be a single tool or a sequence of tools.
-
-    Returns:
-        A list of OTel-conformant tool-definition dicts, or ``None`` when
-        ``tools`` is empty or no tool can be represented.
+# Tool-spec fields that can carry secrets (e.g. an OAuth access token on an MCP
+# tool, or injected auth headers) and must never be copied into emitted telemetry.
+# Matched by field name so any MCP tool spec is covered, not just a specific class.
+_SENSITIVE_TOOL_DEFINITION_KEYS: Final[frozenset[str]] = frozenset({"authorization", "headers"})
```
moonbox3: Could we avoid copying provider-specific configuration into telemetry and emit only OTel-safe metadata such as `type`, `name`, and `description`? For example, a user can give Code Interpreter outbound access to an allowlisted API and attach an API key for that domain. The Azure SDK nests that key inside the container's network policy, and forwarding the complete `container` exports it in `gen_ai.tool.definitions` even when sensitive-data capture is disabled. A shallow allowlist would preserve the single-pass serialization optimization while excluding current and future nested credential fields.

TaoChenOSU: Addressed

moonbox3 (approved):

python/packages/core/agent_framework/observability.py
```diff
@@ -2262,83 +2306,118 @@ def _get_instructions_from_options(options: Any) -> str | list[str] | None:
...
+# keyed by the tool object's identity. Tool objects (e.g. ``FunctionTool``,
+# ``MCPTool``) are often reused across runs, so caching the encoded string (rather
+# than just the definition dict) lets repeated runs reuse the fragment without
+# repeating the isinstance checks, schema generation, dict construction, and
+# ``json.dumps``. A ``WeakKeyDictionary`` lets entries be garbage collected with
+# their tools; unhashable / non-weak-referenceable specs (e.g. plain dicts) bypass
+# the cache.
+_TOOL_OTEL_JSON_CACHE: weakref.WeakKeyDictionary[Any, str | None] = weakref.WeakKeyDictionary()
 # Sentinel distinguishing "not cached" from a cached ``None`` (unparseable tool).
 _CACHE_MISS: Final = object()
```
eavanvalkenburg: the `sentinel` object type has been accepted, so use that from `typing_extensions` instead of this.

python/packages/core/agent_framework/observability.py
```diff
@@ -2366,7 +2445,12 @@ def _build_tool_otel_definition(tool_item: Any) -> dict[str, Any] | None:
     elif isinstance(tool_item, SerializationMixin):
         raw = tool_item.to_dict()
     elif isinstance(tool_item, Mapping):
-        raw = cast("Mapping[str, Any]", tool_item)
+        mapping_item = cast("Mapping[str, Any]", tool_item)
+        # Azure SDK tool models expose ``as_dict()``, which recursively converts
+        # the whole model (including nested non-dict Mapping values that are not
+        # JSON-serializable) into plain dicts; plain mappings are used as-is.
+        as_dict: Callable[[], Mapping[str, Any]] | None = getattr(mapping_item, "as_dict", None)
```
eavanvalkenburg, line 2454: our own classes adhere to the `SerializationProtocol`, which is just to_dict and from_dict, could we use a protocol for this, part as well, that is cleaner then this getattr setup.

alliscode (approved):

python/packages/core/agent_framework/observability.py
TaoChenOSU: Using that now

python/packages/core/agent_framework/observability.py
TaoChenOSU: Added support for `SerializationProtocol` but this is for Azure ADK models, which use `as_dict`.

alliscode (approved):

PR #7052, opened 2026-07-10
.NET: [BREAKING] Graduate todo and agent mode providers out of experimental

westey-m:
### Motivation & Context

The `HarnessAgent` is being prepared for release, but several of the features it
depends on are still marked experimental. This PR graduates the **Todo** and
**Agent Mode** providers (and their related types) out of experimental so they
can ship as part of the harness. These are features that are on by default when
using the harness, so they must be released rather than left behind an
experimental opt-in.

Part of the broader effort to release the harness dependencies.

  - Removed `[Experimental(MAAI001)]` from `TodoProvider`, `AgentModeProvider`,
    and their related todo/mode types. `TodoCompletionLoopEvaluator` remains
    experimental because it ships with the still-experimental Loop feature.
  - Renamed `AgentMode.Description` → `AgentMode.Instructions`, since the value is
    LLM instructions for the mode rather than a human-facing description.
  - Made the `AgentSession` parameters non-nullable (with `Throw.IfNull`) on
    `TodoProvider.GetAllTodosAsync`/`GetRemainingTodosAsync` and the mode helpers —
    a null session was effectively a no-op and only invited misuse.
  - Added per-session locking to `AgentModeProvider` (mirroring `TodoProvider`) so
    concurrent reads/mutations of the mode state on the same session are serialized.
    This made the public helpers async: `GetMode`/`SetMode` are now
    `GetModeAsync`/`SetModeAsync`.
  - Propagated the async change through the internal consumer
    (`TodoCompletionLoopEvaluator`) and the Harness console samples, and updated the
    affected unit tests.

  - The Todo and Agent Mode providers are now part of the stable (non-experimental)
    surface.
  - **Breaking (experimental surface):** `AgentModeProvider.GetMode`/`SetMode` are
    renamed to async `GetModeAsync`/`SetModeAsync`, and `AgentMode.Description` is
    renamed to `Instructions`. These types were experimental, but the change is
    flagged as breaking so preview users are notified.

  - The per-session locking pattern on `AgentModeProvider` matches `TodoProvider`.

### Related Issue

Related to #6964

### Contribution Checklist

### DIFF
--- a/dotnet/samples/02-agents/Harness/Harness_Shared_Console/Commands/ModeCommandHandler.cs
+++ b/dotnet/samples/02-agents/Harness/Harness_Shared_Console/Commands/ModeCommandHandler.cs
@@ -43,7 +43,7 @@ public sealed class ModeCommandHandler : CommandHandler
         string[] parts = input.Split(' ', 2, StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries);
         if (parts.Length < 2)
         {
-            string current = this._modeProvider.GetMode(session);
+            string current = await this._modeProvider.GetModeAsync(session).ConfigureAwait(false);
             await ux.WriteInfoLineAsync($"Current mode: {current}").ConfigureAwait(false);
             return true;
         }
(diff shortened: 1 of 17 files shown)


rogerbarreto (approved):
I see a big breaking change accompanied by a removal, should we keep it experimental on the new API before graduating with a big change right away?

westey-m:
> I see a big breaking change accompanied by a removal, should we keep it experimental on the new API before graduating with a big change right away?

For graduation it's good to consider the public surface changes, and the change I made to the public surface is just about making all the methods async so I can await SemaphoreSlims.  So while the change touches many files, it's conceptually not a significant change.  It's also arguably more future safe to have async methods, so generally just an improvement.

peibekwe (approved):

PR #7066, opened 2026-07-11
Python: feat: add Amazon Bedrock Knowledge Base tool and context provider

PVidyadhar:
# feat: add Amazon Bedrock Knowledge Base tool and context provider

### Motivation & Context

Enables Agent Framework agents to retrieve context from Amazon Bedrock Knowledge Bases. This adds RAG capabilities using AWS's managed infrastructure without requiring users to manage their own vector stores or embedding pipelines.

  - `BedrockKnowledgeBaseTool` — subclasses `FunctionTool` with agentic retrieval (AgenticRetrieveStream) and automatic fallback to standard Retrieve. Can be passed directly to any Agent or ChatClient.
  - `BedrockKnowledgeBaseProvider` — subclasses `ContextProvider` with `before_run()` for automatic context injection using standard managed retrieval on every agent invocation.
  - Samples demonstrating both patterns with Agent Framework primitives.

  - Additive only — new files in `python/packages/bedrock/`, no existing code modified
  - Users get two integration patterns: explicit tool calling (with agentic retrieval) or automatic context injection (standard retrieval)

  - Whether the `FunctionTool` subclass and `ContextProvider` subclass match the framework's conventions
  - The `before_run()` integration using `extend_messages()`

### Related Issue

N/A — new feature adding Amazon Bedrock Knowledge Base integration.

### Contribution Checklist

### Testing

- 49 unit tests pass (14 for KB tool/provider, 35 existing)
- E2E tested against live managed KB in us-west-2
- `tool.invoke()` returns formatted retrieval results with scores and sources
- `provider.before_run()` injects context via `extend_messages()` into `SessionContext`
- SDK requirement: `boto3 >= 1.43.32`
- Microsoft CLA: pre-approved (cla.opensource.microsoft.com)


PVidyadhar:
@microsoft-github-policy-service agree company="Amazon"

eavanvalkenburg (commented):
This is in the wrong place, please move this into the Bedrock package instead. I will look at the implementation then

python/packages/bedrock/agent_framework_bedrock/_knowledge_base_provider.py
```diff
@@ -0,0 +1,189 @@
...
+        min_score: Minimum relevance score to include a result. Defaults to 0.0.
+    """
+
+    def __init__(
+        self,
+        source_id: str = "bedrock-kb",
+        knowledge_base_id: Optional[str] = None,
+        region_name: Optional[str] = None,
+        number_of_results: int = 5,
+        min_score: float = 0.0,
```
eavanvalkenburg, line 58: this calls for a separation into args and keyword args...

python/packages/bedrock/agent_framework_bedrock/_knowledge_base_provider.py
```diff
@@ -0,0 +1,189 @@
...
+    ):
+        super().__init__(source_id=source_id)
+        self.knowledge_base_id = knowledge_base_id or os.environ.get("KNOWLEDGE_BASE_ID", "")
+        self.region_name = region_name or os.environ.get("AWS_REGION", "us-east-1")
+        self.number_of_results = number_of_results
+        self.min_score = min_score
+        self._client = None
```
eavanvalkenburg: in most of these kinds of integrations, we provide a way for the user to supply their own client, because there are often many knobs and params they can configure that we do not want to expose, so let's do that here as well, and then store that in `.client` or create it, I also prefer we raise here when boto3 is not installed, rather then when trying to use it.

python/packages/bedrock/agent_framework_bedrock/_knowledge_base.py
```diff
@@ -0,0 +1,183 @@
...
+    ):
+        self.knowledge_base_id = knowledge_base_id or os.environ.get("KNOWLEDGE_BASE_ID", "")
+        self.region_name = region_name or os.environ.get("AWS_REGION", "us-east-1")
+        self.number_of_results = number_of_results
+        self.use_agentic_retrieval = (
+            use_agentic_retrieval
+            if use_agentic_retrieval is not None
+            else os.environ.get("USE_AGENTIC_RETRIEVAL", "true").lower() != "false"
+        )
+        self._client = None
```
eavanvalkenburg: see my comment in the provider

python/packages/bedrock/agent_framework_bedrock/_knowledge_base_provider.py
```diff
@@ -0,0 +1,189 @@
...
+    def __init__(
+        self,
+        source_id: str = "bedrock-kb",
+        knowledge_base_id: Optional[str] = None,
+        region_name: Optional[str] = None,
+        number_of_results: int = 5,
+        min_score: float = 0.0,
+    ):
+        super().__init__(source_id=source_id)
+        self.knowledge_base_id = knowledge_base_id or os.environ.get("KNOWLEDGE_BASE_ID", "")
```
eavanvalkenburg: we have a setup for settings, that can load from env files and dotenvs, etc. please use that, `KNOWLEDGE_BASE_ID` is also not ideal from a separation perspective...

python/packages/bedrock/agent_framework_bedrock/_knowledge_base_provider.py
```diff
@@ -0,0 +1,189 @@
...
+                import boto3
+                from botocore.config import Config
+            except ImportError:
+                raise ImportError(
+                    "boto3 is required for BedrockKnowledgeBaseProvider. Install with: pip install boto3>=1.43.2"
+                )
+            self._client = boto3.client(
+                "bedrock-agent-runtime",
+                region_name=self.region_name,
+                config=Config(user_agent_extra="ms-agent-framework/bedrock-kb"),
```
eavanvalkenburg: we have a helper to get a user-agent string, that includes a bit more info, which is useful.

python/packages/bedrock/agent_framework_bedrock/_knowledge_base_provider.py
PVidyadhar: updated

python/packages/bedrock/agent_framework_bedrock/_knowledge_base_provider.py
PVidyadhar: updated

python/packages/bedrock/agent_framework_bedrock/_knowledge_base.py
PVidyadhar: updated

python/packages/bedrock/agent_framework_bedrock/_knowledge_base_provider.py
PVidyadhar: updated

python/packages/bedrock/agent_framework_bedrock/_knowledge_base_provider.py
PVidyadhar: updated

python/packages/bedrock/agent_framework_bedrock/_knowledge_base.py
```diff
@@ -0,0 +1,162 @@
...
+
+class BedrockKnowledgeBaseTool:
+    """Tool that retrieves documents from Amazon Bedrock Knowledge Bases.
+
+    Usage:
+        from agent_framework_bedrock import BedrockKnowledgeBaseTool
+
+        tool = BedrockKnowledgeBaseTool(knowledge_base_id="YOUR_KB_ID")
+        agent = Agent(tools=[tool])
+    """
```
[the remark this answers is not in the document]
PVidyadhar: Good callout. The current implementation exposes get_tool_definition() for schema introspection, but I'd like guidance from @eavanvalkenburg on the preferred pattern — should this be wrapped as a FunctionTool, or is there a different registration mechanism for tools in the bedrock package?
  Happy to adapt once the direction is clear.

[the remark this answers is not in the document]
eavanvalkenburg: So, indeed in order for this to be passed to any arbitrary ChatClient, this needs to be a FunctionTool, there are two ways to do that, either subclass FunctionTool, or set the `@tool` decorator on the `run` method of this class, and then instruct folks to pass `kb.run` to the agent, I would opt for the first, that will be simpler.

python/packages/bedrock/agent_framework_bedrock/_knowledge_base_provider.py
```diff
@@ -0,0 +1,92 @@
...
+        "boto3 is required for BedrockKnowledgeBaseProvider. "
+        "Install it with: pip install boto3>=1.43.2"
+    ) from e
+
+from agent_framework_bedrock._knowledge_base import _get_source_uri
+
+logger = logging.getLogger(__name__)
+
+
+class BedrockKnowledgeBaseProvider:
```
eavanvalkenburg: this does not implement a ContextProvider, you need to subclass that (we type check all objects passed to the `context_providers` param, and implement the `before_run` method, at least. The `after_run` method can also be used, depending on the scenario.

eavanvalkenburg (commented):
Would also be good to have some samples for this, and those will also help you ensure you build things that pass type checking etc. when used with MAF primitives.

python/packages/bedrock/agent_framework_bedrock/_knowledge_base.py
```diff
@@ -0,0 +1,164 @@
...
+                    results.append({
+                        "content": r.get("content", {}).get("text", ""),
+                        "source": _get_source_uri(r),
+                        "score": r.get("score", 0),
+                    })
+        return results
+
+    def _standard_retrieve(self, query: str) -> list[dict[str, Any]]:
+        """Use standard Retrieve API."""
+        response = self._client.retrieve(
```
moonbox3, line 167: Could we raise the boto3 and botocore lower bounds to 1.43.32? The package still allows 1.35.0, but `managedSearchConfiguration` and `AgenticRetrieveStream` first appear in botocore 1.43.32. In an otherwise supported older environment, the agentic call fails and its fallback then fails parameter validation, leaving both new entry points unusable.

moonbox3:
Are we still tracking this, @PVidyadhar ?

PVidyadhar:
> Are we still tracking this, @PVidyadhar ?

Yes, i will be looking into the comments soon and address them, thanks for reviewing @moonbox3

python/packages/bedrock/agent_framework_bedrock/_knowledge_base_provider.py
PVidyadhar: updated.

python/packages/bedrock/agent_framework_bedrock/_knowledge_base.py
PVidyadhar: updated

PVidyadhar:
Requesting re-review @moonbox3 , @eavanvalkenburg . thanks for your comments!

python/packages/bedrock/agent_framework_bedrock/_knowledge_base_provider.py
```diff
@@ -0,0 +1,152 @@
...
+            raise
+        except Exception:
+            logger.debug("KB retrieval failed, continuing without context", exc_info=True)
+            return
+
+        if not retrieved_context:
+            return
+
+        # Inject as a user message (untrusted external content) via extend_messages
+        context_message = Message(role="user", contents=[f"{self.context_prompt}\n\n{retrieved_context}"])
```
moonbox3, line 131: What happens when this provider is used with `BedrockChatClient`? `SessionContext.get_messages(include_input=True)` appends the real input after this context message, and `_prepare_bedrock_messages` preserves both, producing consecutive `user` roles that Bedrock Converse rejects because roles must alternate. Could we merge the retrieved block into the current user turn, or coalesce adjacent user messages before serialization?

moonbox3:
@PVidyadhar please re-open when you're ready to address the comment and move things forward. Thanks.

PVidyadhar:
Hi @moonbox3 @eavanvalkenburg — I addressed the consecutive-`user`-roles feedback but this PR couldn't be reopened after the force-push . Continued in #8173 - https://github.com/microsoft/agent-framework/pull/8173, with the fix: the provider now injects retrieved context via `extend_instructions()` so it lands in Bedrock's `system` field, separate from the `conversation` array — no consecutive user turns for any model. Verified E2E with BedrockChatClient against a live KB. Sorry for the new PR number!

PR #7071, opened 2026-07-12
.NET: Fix _CodeValidator os.* allow-list bypass for aliased imports (#7068)

### LINKED ISSUE #7068: Python: .NET: [Bug]: `_CodeValidator` does not enforce the documented `os.environ`/`os.path`-only policy for aliased imports
#### The documented contract

The validator *documents and intends* an `os`-restricted-to-`environ`/`path` policy, in three places in its own source:

- `ALLOWED_IMPORTS` admits `os` explicitly scoped:
  ```python
  "os",  # Limited to os.environ, os.path - validated via attribute access
  ```
- The attribute allow-list:
  ```python
  ALLOWED_OS_ATTRS: set[str] = {"environ", "path"}
  ```
- `visit_Attribute`’s own comment states the contract:
  > Enforce the `os` attribute allow-list. Anything outside `ALLOWED_OS_ATTRS` (file I/O, process control, mutating helpers, etc.) is rejected **so the validator matches the documented `os.environ` / `os.path`-only contract.**

So the intended, documented behaviour is unambiguous: through `os`, only `environ` and `path` should be reachable.

#### The inconsistency

Enforcement is keyed on the **literal receiver name `os`**. In `visit_Attribute`:

```python
if isinstance(node.value, ast.Name) and node.value.id == "os" and node.attr not in self._allowed_os_attrs:
    self._errors.append(f"Access to os.{node.attr} is not allowed")
```

`visit_Import` admits the module but never records its bound name (`asname`), and assignments aren’t tracked. So any name other than `os` bound to the `os` module escapes the check entirely, even though the accessed attribute (`system`, `popen`, `execvp`, `fork`, `unlink`, `chmod`, …) is identical.

Notably, `visit_ImportFrom` **already enforces this same policy** for the `from os import system` form:

```python
elif module_name == "os":
    for alias_node in node.names:
        if alias_node.name not in self._allowed_os_attrs:
            self._errors.append(f"Import from 'os' of '{alias_node.name}' is not allowed")
```

### Code Sample

```markdown
#### Reproduction

Standalone against `validator.py`, same commit:


# (a) direct — policy enforced ✔
$ echo '{"code":"import os\nos.system(\"id\")"}' | python -I validator.py
{"errors": ["Access to os.system is not allowed"]}          exit=1

# (b) from-import — policy enforced ✔  (existing visit_ImportFrom guard)
$ echo '{"code":"from os import system\nsystem(\"id\")"}' | python -I validator.py
{"errors": ["Import from 'os' of 'system' is not allowed"]} exit=1

# (c) module alias — SAME os.system, NOT enforced (X)
$ echo '{"code":"import os as x\nx.system(\"id\")"}' | python -I validator.py
(no errors)                                                 exit=0

# (d) assignment alias — NOT enforced ✗
$ echo '{"code":"import os\n_o = os\n_o.system(\"id\")"}' | python -I validator.py
(no errors)                                                 exit=0
```

```markdown

```

### Package Versions

Microsoft.Agents.AI.LocalCodeAct: source build @ main 68136ee081dbbee6983e6bb92a834f9ad30d20dc 

### .NET Version

.NET 8.0

#### Suggested fix

Track the names bound to `os` and check membership instead of a literal `== "os"` — mirroring the enforcement already present in `visit_ImportFrom`:

```python
def __init__(self, ...):
    ...
    self._os_aliases: set[str] = {"os"}

def visit_Import(self, node: ast.Import) -> None:
    for alias in node.names:
        if alias.name.split(".")[0] == "os":
            self._os_aliases.add(alias.asname or "os")
    self.generic_visit(node)

def visit_Assign(self, node: ast.Assign) -> None:
    if (isinstance(node.value, ast.Name) and node.value.id in self._os_aliases
            and len(node.targets) == 1 and isinstance(node.targets[0], ast.Name)):
        self._os_aliases.add(node.targets[0].id)
    self.generic_visit(node)

def visit_Attribute(self, node: ast.Attribute) -> None:
    if (isinstance(node.value, ast.Name) and node.value.id in self._os_aliases
            and node.attr not in self._allowed_os_attrs):
        self._errors.append(f"Access to os.{node.attr} is not allowed")
    ...
```

eajajhossain:
The code validator in `LocalCodeAct` is supposed to only let generated code touch `os.environ` and `os.path` — everything else on `os` should be blocked. But it only checked for the literal name `os`, so renaming the module got right past it:

import os; os.system("id")          # blocked (correct)
import os as x; x.system("id")      # NOT blocked (bug)
import os; _o = os; _o.system("id") # NOT blocked (bug)

Same dangerous call, just a different variable name. This is effectively a sandbox escape, and this PR closes the hole so the `os.environ`/`os.path`-only rule actually holds.

- **Major changes:** Instead of checking for the literal name `os`, the validator now tracks every name that points at the `os` module (via `import ... as`, `import os.path`, and plain assignments like `_o = os`) and checks against that set. This mirrors the approach `visit_ImportFrom` already uses. Changes are in `validator.py`, plus new regression tests.

- **Impact:** Aliased/renamed access to `os` is now blocked just like direct access. Normal usage (`os.environ`, `os.path`, and `import os.path as p; p.join(...)`) still works, and there are tests to make sure it isn't over-blocked. No public API changes — only the embedded Python script and its tests.

- **Focus for reviewers:** The edge cases - that `import os.path as p` is correctly *not* treated as an `os` alias, and that assignment tracking doesn't accidentally block legitimate code.

### Related Issue

Fixes #7068 

### Contribution Checklist

- [x] **This is not a breaking change.** If it _is_ a breaking change, add the `breaking change` label (or add "[BREAKING]" to the title prefix, before or after any language prefix) - a workflow keeps the label and title prefix in sync automatically.

### DIFF
--- a/dotnet/src/Microsoft.Agents.AI.LocalCodeAct/Resources/validator.py
+++ b/dotnet/src/Microsoft.Agents.AI.LocalCodeAct/Resources/validator.py
@@ -271,6 +271,12 @@ class _CodeValidator(ast.NodeVisitor):
         self._allowed_builtins = allowed_builtins if allowed_builtins is not None else ALLOWED_BUILTINS
         self._blocked_builtins = blocked_builtins if blocked_builtins is not None else BLOCKED_BUILTINS
         self._allowed_os_attrs = allowed_os_attrs if allowed_os_attrs is not None else ALLOWED_OS_ATTRS
+        # Names currently bound to the ``os`` module. Enforcement in
+        # ``visit_Attribute`` is keyed on membership here rather than the literal
+        # name ``os`` so that aliases (``import os as x``) and re-bindings
+        # (``_o = os``) cannot slip past the ``os.environ`` / ``os.path``-only
+        # contract with the very same attribute access (``x.system``, ``_o.system``).
+        self._os_aliases: set[str] = {"os"}
 
     def validate(self, code: str) -> None:
         """Validate code and raise CodeValidationError if it violates policy."""
@@ -303,6 +309,16 @@ class _CodeValidator(ast.NodeVisitor):
                 self._errors.append(f"Import of '{alias_node.name}' is not allowed (blocked: {module_name})")
             elif module_name not in self._allowed_imports:
                 self._errors.append(f"Import of '{alias_node.name}' is not allowed (not in allow-list)")
+            # Record every name bound to the ``os`` module so aliased access is
+            # still subject to the ``os`` attribute allow-list.
+            if alias_node.name == "os":
+                # ``import os`` / ``import os as x`` -> binds the ``os`` module.
+                self._os_aliases.add(alias_node.asname or "os")
+            elif alias_node.name.startswith("os.") and alias_node.asname is None:
+                # ``import os.path`` binds the top-level name ``os``.
+                # (``import os.path as p`` instead binds ``p`` to the submodule,
+                # not ``os``, so it is intentionally not tracked here.)
+                self._os_aliases.add("os")
         self.generic_visit(node)
 
     def visit_ImportFrom(self, node: ast.ImportFrom) -> None:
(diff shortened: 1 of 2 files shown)


eajajhossain:

eavanvalkenburg (approved):

eavanvalkenburg:
Thank you for investigating and implementing the validator hardening for #7068, including the alias and permitted-access regression coverage. PR #7138 incorporates this fix together with the LocalCodeAct package smoke-check change, so it supersedes this PR and avoids merging duplicate validator changes. Closing this PR in favor of #7138.

eavanvalkenburg:
Superseded by #7138.

